PR #86, opened 2020-11-18
Basic structure for SFTP transport with happy path

mry-usds:
This PR does defines a basic structure for the Transport/Send Azure functions and has a happy-path implementation for SFTP

## Changes
- Adds the azure function "Send" which waits for blobs placed into the processed container
- Uses the blob to send (default) SFTP to a localhost location/service
- Establishes the placement of metadata into the receivers.yml file and starts to build a structure for that
- Has placeholders and a start to dynamically calling the correct Transport from the metadata

## Checklist
- [X] Tested locally?
- [X] Added new dependencies?
- [ ] Updated the docs?
- [ ] Added a test?

## Fixes 

- #issue

## To Be Done
- Add tests
- #issue


prime-router/pom.xml
```diff
@@ -166,7 +166,7 @@
                     <archive>
                         <manifest>
                             <addClasspath>true</addClasspath>
-                            <classpathPrefix>${stagingDirectory}/lib</classpathPrefix>
+                            <classpathPrefix>lib/</classpathPrefix>
```
RickHawesUSDS, line 169: Not sure about this change. It breaks what I setup for the prime script. Why?

prime-router/src/main/kotlin/Receiver.kt
```diff
@@ -24,6 +27,20 @@ data class Receiver(
         }
     }
 
+    data class Transport(
```
RickHawesUSDS, line 30: +1 on this class.

prime-router/src/main/kotlin/azure/SendFunction.kt
```diff
@@ -0,0 +1,64 @@
...
+    fun run(
+        @BlobTrigger( name = "content", path="%PROCESSED_BLOB_CONTAINER%/{fileName}.csv", dataType = "binary") content: ByteArray,
+        @BindingName("fileName") fileName: String,
+        context: ExecutionContext
+    ) {
+        val baseDir = System.getenv("AzureWebJobsScriptRoot")
+        Metadata.loadAll("$baseDir/metadata")
+
+        context.logger.info("Dispatch function processed a blob. Name: ${fileName} Size: ${content.size} bytes");
+
```
RickHawesUSDS, line 40: +1 on logging info about the payload

RickHawesUSDS (changes requested):
One change in the POM that I'm requesting. Also, is there a docker SFTP server that we should be running? Should that be put into the docker-compose.yml file?

prime-router/pom.xml
mry-usds: thats a hold over from the older issues i was having with Windows...didnt realize that was in there...still

RickHawesUSDS (approved):

PR #106, opened 2020-11-25
Jim/e2e test

jimduff-usds:
This PR does ...
- Many modifications to make the data AZ wants.
- Fixed bug that .csv extension was missing
- Fixed bugs that ValueSets were not mapping properly.
Detailed comments with each modified file.

## Changes
- I have added comments to each file change.
- 

## Checklist
- [x] Tested locally?
- [No] Added new dependencies?
- [ ] Updated the docs?
- [ ] Added a test?

## Fixes 

- #issue

## To Be Done

- Mapping of ValueSets needs further improvement.   Need to support many different ways of displaying a single value set.  Currently it is fixed in code to only have two ways to display- as the 'code', or as the 'display'.   Need a more generic solution.
- Remove hack default values in metadata to meet MVP.


prime-router/metadata/organizations.yml
```diff
@@ -16,21 +16,23 @@
       - name: elr
         topic: covid-19
         schema: az/az-covid-19
-        jurisdictionalFilter: { standard.patient_state: AZ }
+        jurisdictionalFilter: { standard.testing_lab_state: AZ }
```
jimduff-usds, line 19: Now filtering on the Lab's location, not the patient's.    Will need further tweaking.

prime-router/metadata/organizations.yml
```diff
@@ -16,21 +16,23 @@
...
-    description: Pennsylvennia PHD
-    services:
-      - name: elr
-        topic: covid-19
-        schema: pa/pa-covid-19
-        jurisdictionalFilter: { standard.patient_state: PA }
-        transforms: { deidentify: false }
-        address: http://localhost:1181/
-        format: CSV
+# Temporarily removing PA until we're ready to send there.
```
jimduff-usds, line 25: Commented out PA for now.

prime-router/metadata/schemas/AZ/az-covid-19-csv.schema
```diff
@@ -15,7 +15,7 @@ elements:
   - name: standard.testing_lab_name
     csvField: Lab_Name
 
-  - name: standard.testing_lab_id
```
jimduff-usds, line 18: AZ needs the CLIA code, so adding this for PDI to send to us.

prime-router/metadata/schemas/AZ/az-covid-19-csv.schema
```diff
@@ -48,10 +48,10 @@ elements:
   - name: standard.patient_gender
     csvField: Patient_Sex
 
-  - name: standard.patient_race
+  - name: standard.patient_race#text
```
jimduff-usds, line 51: #text is the magic that tells code to use the "display" value instead of the "code" value.    These changes were to meet AZ requests.

prime-router/metadata/schemas/PrimeDataInput/pdi-covid-19.schema
```diff
@@ -139,7 +145,7 @@ elements:
   - name: standard.ordering_provider_phone_number
     csvField: Ordering_provider_phone
 
-  - name: standard.ordered_test_code
+  - name: standard.test_performed
```
jimduff-usds, line 148: PDI was using "ordered_test_code" but AZ was using "test_performed".  Turns out they are the same ValueSet.   I think we can remove ordered-test_code.

prime-router/src/main/kotlin/FakeReport.kt
```diff
@@ -38,7 +38,7 @@ class FakeReport {
                 }
                 Element.Type.DATETIME -> {
                     val date = faker.date().past(10, TimeUnit.DAYS)
-                    val formatter = SimpleDateFormat("yyyyMMddhhmm")
+                    val formatter = SimpleDateFormat("yyyyMMdd")
```
jimduff-usds, line 41: This change is per AZ request

prime-router/src/main/kotlin/Report.kt
```diff
@@ -28,7 +28,7 @@ class Report {
 
     val rowCount: Int get() = this.table.rowCount()
     val rowIndices: IntRange get() = 0 until this.table.rowCount()
-    val name: String get() = formFileName(id, schema.baseName, createdDateTime)
+    val name: String get() = formFileName(id, schema.baseName, destination?.format, createdDateTime)
```
jimduff-usds, line 31: There were two separate places where blob filenames were being created.   I've consolidated them into one static function in Report, formFileName.  (This is why the .csv extension wasn't working)

prime-router/src/main/kotlin/Report.kt
```diff
@@ -213,8 +213,10 @@ class Report {
             }
             toElement.isCode -> {
                 Array(table.rowCount()) { row ->
-                    val fromDisplay = table.getString(row, toElement.nameAsCodeText)
-                    valueSet.toCode(fromDisplay) ?: toElement.default ?: ""
+                    // Value in the table is already a code.  Just return it; no display->code mapping needed.
+                    // BUG!! If value in the incoming table is a #text display, this breaks!
+                    // Need robust generic way to have many ways to display a single valueset.
+                    table.getString(row, toElement.nameAsCode) ?: "BUG - incoming schema used #text?"
```
jimduff-usds, line 219: Example of the bug:  if some incoming PDI schema field uses "#text", that is, it sends in a 'display' value, and the outgoing AZ schema does not use "#text", that is, it expects a 'code' value, then this code will break.   By sheer luck, current incoming schema has no "#text", so I'm momentarily punting on fixing this, esp as I think a more general solution is needed.

prime-router/src/main/kotlin/Schema.kt
```diff
@@ -41,16 +41,16 @@ data class Schema(
         val missing = mutableSetOf<String>()
 
         toSchema.elements.forEach { toElement ->
-            findMatchingElement(toElement)?.let {
-                useDirectly[toElement.name] = it
-                return@forEach
-            }
             findMatchingValueSet(toElement)?.let {
```
jimduff-usds, line 44: @RickHawesUSDS  - please check this.  I swapped the order of the checks here, to give the 'useValueSet' higher priority.  Before, that option was unreachable code - so valueSets were never being properly processed.

prime-router/src/main/kotlin/azure/ReportQueue.kt
```diff
@@ -158,10 +158,6 @@ object ReportQueue {
         return outputStream.toByteArray()
     }
 
-    private fun createBlobFilename(report: Report): String {
```
jimduff-usds, line 161: Half the filename creation was here, and half was in Report.formFileName.   Some code was calling both, some code only calling one.    Consolidated.

jimduff-usds:
@mry-usds  - I have permanently added the extension (currently only .csv) to the name of the Report in a generic way.

So you may not need the .csv here any more:
     ./src/main/kotlin/transport/SftpTransport.kt:        val path = "${fileDir}/${service.fullName.replace( '.', '-')}-${header.id}.csv"

prime-router/src/main/kotlin/FakeReport.kt
RickHawesUSDS: The default should still be "yyyyMMddhhmm" because that follows national standards (HL7 ...). I suggest that we use the format field of the Element and stuff "yyyyMMdd" in there. So the code should be:
```
SimpleDateFormat(element.format ?: "yyyyMMddhhmm")
```

prime-router/src/main/kotlin/FakeReport.kt
```diff
@@ -75,7 +75,7 @@ class FakeReport {
                         else -> TODO()
                     }
                 }
-                Element.Type.TELEPHONE -> faker.phoneNumber().cellPhone()
+                Element.Type.TELEPHONE -> faker.numerify("##########") // faker.phoneNumber().cellPhone()
```
RickHawesUSDS, line 78: Look's fine, but being flexible on the format may be needed.

prime-router/src/main/kotlin/Report.kt
RickHawesUSDS: Could you change this to a TODO which can be found by IntelliJ?

RickHawesUSDS (approved):
Super good.

prime-router/src/main/kotlin/FakeReport.kt
jimduff-usds: check

prime-router/src/main/kotlin/Report.kt
jimduff-usds: will do

prime-router/src/main/kotlin/FakeReport.kt
jimduff-usds: will do

PR #121, opened 2020-12-03
zipcode

RickHawesUSDS:
This PR does adds formatting options for POSTAL_CODE types. 

## Changes
- 5 and 9 digit zipcode support
- Validation of zipcode on entry

## Checklist
- [x] Tested locally?
- [ ] Added new dependencies?
- [ ] Updated the docs?
- [ ] Added a test?

### DIFF
--- a/prime-router/metadata/schemas/AZ/az-covid-19-csv.schema
+++ b/prime-router/metadata/schemas/AZ/az-covid-19-csv.schema
@@ -103,6 +103,7 @@ elements:
   - name: standard.patient_zip_code
     csvFields:
     - name: Patient_Zip
+      format: $zip # Only a 5 digit zipcode
 
   - name: standard.patient_county
     csvFields:
--- a/prime-router/src/main/kotlin/Element.kt
+++ b/prime-router/src/main/kotlin/Element.kt
@@ -169,6 +169,11 @@ data class Element(
                     .replace(exchangeToken, parts[0].substring(3, 6))
                     .replace(subscriberToken, parts[0].substring(6))
                     .replace(extensionToken, parts[2])
+            Type.POSTAL_CODE -> {
+                when (field?.format) {
+                    zipPlus4Token -> normalizedValue
+                    else -> normalizedValue.substring(0, 5)
+                }
             }
             else -> normalizedValue
         }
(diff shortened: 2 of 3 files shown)


prime-router/metadata/schemas/AZ/az-covid-19-csv.schema
```diff
@@ -103,6 +103,7 @@ elements:
   - name: standard.patient_zip_code
     csvFields:
     - name: Patient_Zip
+      format: $zip # Only a 5 digit zipcode
```
RickHawesUSDS, line 107: AZ only wants 5 digits

prime-router/src/main/kotlin/Element.kt
```diff
@@ -207,6 +213,11 @@ data class Element(
                         ?: error("Invalid Code: '$formattedValue' does not match any codes for '${name}'")
                 }
             }
+            Type.POSTAL_CODE -> {
+                if (!Regex("^\\d{5}(-\\d{4})?\$").matches(formattedValue))
+                    error("Input Error: invalid postal code '$formattedValue'")
```
RickHawesUSDS, line 218: Accept only 5 and 9 digit zipcodes

jimduff-usds: What about international zip codes? - those will definitely occur.    If a file has a bad zip code, does this mean we're "failing" the whole file right now?

jimduff-usds (changes requested):
Rick, I'm concerned this will fail data that has non-canonical zip codes, which will admittedly be rare, but will definitely occur. Since we're not using the zip codes for anything right now, suggest we not include this validation enforcement, or wait and revisit when we are starting to do clever geo- lookups with zip codes etc.

prime-router/src/main/kotlin/Element.kt
RickHawesUSDS: Ok. It seems like the HL7 standard allows 12 letters and digits. I'll limit it to that. ZipCode is important though for our jurisdictional filter. Some districts are organized by zipcode and not county, for example.

jimduff-usds: Yes, 12 chars seems much more general.

RickHawesUSDS:
@jimduff-usds Please look again. I've researched Intl zipcodes, made the validation accept all of them plus anything that looks like a number and letter combination. On the formatting side, I defaulted to just returning what was entered if the number doesn't conform to US zipcode formats.

jimduff-usds (approved):
Looks good!

PR #123, opened 2020-12-03
Bug fix:  

jimduff-usds:
This PR fixes a bug in the altValue code.

Yesterday I put in a bug fix, which prioritized altValue even when no value set was present.  I found today I also had to fix the reverse flow, from formatted value to normalized value.   Turns out this is used in the Batch function.  now fixed.

## Changes
- Element.kt
- 

## Checklist
- [X] Tested locally?
- [ ] Added new dependencies?
- [ ] Updated the docs?
- [ ] Added a test?


RickHawesUSDS (approved):
Looks good.

Issue #146, opened 2020-12-09
TX HL7 schema


RickHawesUSDS:
The TX PHD would like to receive ELR in HL7 format. 

**Background**
https://www.dshs.state.tx.us/IDCU/investigation/ELR/Texas-HL7-2_5_1-ELR_Specification_-2016_Final-Draft-4-29-16.pdf

**Acceptance Criteria**
- [ ] Schema that follows the ELR specification
- [ ] Kayode Olupinyo of TX PHD reviews

drew-usds:
Waiting on feedback.

PR #151, opened 2020-12-10
HL7 schema for Texas

RickHawesUSDS:
This PR does adds a HL7 Schema for Texas. As part of this work, a number of improvements for HL7 conversion was added as well. 

## Changes
- Texas HL7 schema
- New basedOn and extends schema relationships. See comments in code
- Removed the `standard` terminology and the per element inheritance that it allowed. Too complicated and not needed.
- The HD type was reworked based on a better understanding of the type. It is now a single element (as it always should be) with multiple formats. 
- Added elements for `ordering_facility_email`, `message_id` and `processing_mode`. 

## Checklist
- [x] Tested locally?
- [ ] Added new dependencies?
- [x] Updated the docs?
- [x] Added a test?

## Fixes 

- #146 

## To be done

- Incorporate feedback and change requests from TX


prime-router/SFTP-SETUP.md
```diff
@@ -26,7 +26,7 @@ Steps
         - name: elr
           topic: covid-19
           schema: az/az-covid-19
-          jurisdictionalFilter: { standard.patient_state: AZ }
+          jurisdictionalFilter: { patient_state: AZ }
           transforms: { deidentify: false }
```
RickHawesUSDS, line 30: `standard.` is removed everywhere.

prime-router/metadata/schemas/AZ/az-covid-19-hl7.schema
```diff
@@ -1,185 +1,13 @@
...
-
-  - name: standard.test_result_date
-
-  - name: standard.test_result_status
-
-  - name: standard.date_result_released
-
-  - name: standard.comment
+  - name: receiving_facility
+    default: AZDOH&2.16.840.1.114222.4.1.142&ISO
```
RickHawesUSDS, line 12: With the `extends` a bunch of elements in the schema can just be the same as the base `covid-19` schema.

prime-router/metadata/schemas/covid-19.schema
```diff
@@ -4,11 +4,10 @@
 # Conventions:
 # - Sort by element name
 # - Use the HL7 message name with _ instead of space and lower cased
-# - Extended with #text or #system for the various parts of the HL7 field
 # - Use ID instead of Identifier
 # - Avoid abbreviations. Use street1, phone_number, zip_code, ...
 #
-name: standard
+name: covid-19
 topic: covid-19
```
RickHawesUSDS, line 11: making this schema name less generic

prime-router/metadata/schemas/covid-19.schema
```diff
@@ -591,6 +589,18 @@ elements:
...
+    default: P
+    hl7Field: MSH-11-1
+    documentation: P, D, or T for Production, Debugging, or Training
+
+  - name: message_id
+    type: ID
+    hl7Field: MSH-10
+    documentation: unique id to track the usage of the message
+
   # name: public_health_case_id
```
RickHawesUSDS, line 604: Will want SimpleReport to send us their internal message id

prime-router/metadata/schemas/covid-19.schema
```diff
@@ -652,17 +654,11 @@ elements:
...
-    hl7Field: MSH-3-1
-
-  - name: sending_application_id
     type: HD
-    default: 2.16.840.1.114222.4.1.237821
+    # Default is the official name of PRIME from CDC PHIN. Matches the OID in sending_application_id, below.
+    default: CDC PRIME - Atlanta, Georgia (Dekalb)&2.16.840.1.114222.4.1.237821&ISO
     natFlatFileField:
-    hl7Field: MSH-3-2
+    hl7Field: MSH-3
```
RickHawesUSDS, line 662: With a properly implemented HD type, two elements collapse into one.

prime-router/metadata/schemas/covid-19.schema
```diff
@@ -683,7 +679,7 @@ elements:
 
   - name: specimen_role
     type: CODE
-    valueSet: common/specimen_role
+    valueSet: specimen_role
     natFlatFileField: Role
```
RickHawesUSDS, line 683: this was a mistake.

prime-router/src/main/kotlin/Hl7Converter.kt
```diff
@@ -136,7 +137,7 @@ object Hl7Converter {
         val country = parts[1]
         val extension = parts[2]
 
-        terser.set(buildComponent(pathSpec, 2), "WPN")
+        terser.set(buildComponent(pathSpec, 2), if(element.nameContains("patient")) "PRN" else "WPN")
         terser.set(buildComponent(pathSpec, 5), country)
```
RickHawesUSDS, line 141: Guessing on Home versus Work based on field name.

prime-router/src/main/kotlin/Metadata.kt
```diff
@@ -74,26 +74,6 @@ object Metadata {
...
-            val schemaName = normalizeSchemaName(
-                if (dirRelPath.isEmpty()) fromSchemaFile.name else dirRelPath + "/" + fromSchemaFile.name
-            )
-            return fromSchemaFile.copy(name = schemaName)
-        } catch (e: Exception) {
-            throw Exception("Error parsing '${file.name}'", e)
-        }
-    }
-
     private fun readAllSchemas(catalogDir: File, dirRelPath: String): List<Schema> {
```
RickHawesUSDS, line 77: Just reordering functions to make the code read better.

prime-router/src/main/kotlin/Schema.kt
```diff
@@ -1,17 +1,50 @@
...
+ * @param name is the unique name of the schema
+ * @param topic defines the set of schemas that can translate with each other
+ * @param elements are the data elements of the schema
+ * @param trackingElement is the element that is unique per item and is used for tracking
+ * @param description is the text description of the schema
+ * @param referenceUrl is a link to a standard document used to create the schema
+ * @param extends is a name of a schema that this schema extends.
+ * @param basedOn is a name of a schema that this schema is based on
+ */
 data class Schema(
```
RickHawesUSDS, line 35: JavaDoc Comments. What a concept.

prime-router/metadata/schemas/AZ/az-covid-19-csv.schema
```diff
@@ -2,67 +2,65 @@
...
-    csvFields:
-    - name: Sending_Application_ID
-
-  - name: standard.testing_lab_name
+  - name: testing_lab_name
     csvFields:
     - name: Lab_Name
 
-  - name: standard.testing_lab_clia
+  - name: testing_lab_clia
```
jimduff-usds, line 19: A thought:   AZ expects fewer columns than are incoming from SimpleReport.   So there does need to be a way to say "out of all the covid-19 fields, please only include these...testing_lab_name, testing_lab_client, etc ,etc"

But in most cases, all the other metadata about those fields is identical to the parent schema.   Is there a way we can have a syntax that consolidated this long file, and simply lists the fields we want to display, in the most common cases where we're re-using the csvField column name etc.   Example:

```
  name: az-covid-19
  description: AZ COVID-19 flat file
  reuse-from-parent-schema: { testing_lab_name, testing_lab_client, etc ,etc }
```

With that list at the top of the file, then the rest of az-covid-19-csv.schema is simply a list of those places where we are overriding the metadata in the parent schema, eg, to use an $alt schema, or a different csvField columnName.   Much more readable.

prime-router/metadata/schemas/AZ/az-covid-19-hl7.schema
jimduff-usds: And per my comment above, the covid-19 base schemas as about 1e6 fields in it.   We need a crisp way to say, out of all the zillion fields in my parent schema, these are the only ones to include in the child schema.

Put another way, covid-19.schema is really more of a data dictionary than it is a schema, then the children pick and choose which words from the dictionary they want to use.

prime-router/metadata/schemas/AZ/pima-az-covid-19-csv.schema
```diff
@@ -2,22 +2,23 @@
 name: pima-az-covid-19
 description: Pima County, AZ COVID-19 flat file
 topic: covid-19
-trackingElement: standard.testing_lab_specimen_id
+trackingElement: testing_lab_specimen_id
+basedOn: covid-19
```
jimduff-usds, line 6: The Pima Schema is actually 99% the same as the SimpleReport schema.  (I created it by first just copying the pdi schema, then making adjustments.)

So, again, this is subtly different, but maybe we should be doing something like this:
```
  name: pima-az-covid-19
  description: Pima County, AZ COVID-19 flat file
  data-dictionary: covid-19
  extends:   pdi-covid-19
```
That way the pima schema can make use of all the fancy Valuesets etc in the data dictionary, while guaranteeing its just a subset of data coming from SimpleReport.

prime-router/metadata/schemas/PrimeDataInput/pdi-covid-19.schema
```diff
@@ -22,184 +23,189 @@ elements:
     csvFields: [{name: Patient_ID}]
```
jimduff-usds, line 23: This will, I hope, create a merge conflict.   I changed this back to standard.patient_id (or, now, simply patient_id), to fix the bug that Patient_ID was missing in AZ and Pima data.

Suggest we leave it that way until we have a conversation with SimpleReport team on IDs.

prime-router/metadata/schemas/PrimeDataInput/pdi-covid-19.schema
```diff
@@ -22,184 +23,189 @@ elements:
...
 
   # In SimpleReport, lookup_id is the ID the user entered into the app.
-  - name: standard.patient_id
+  - name: patient_id
```
jimduff-usds, line 26: For now, leave as patient_lookup_id as in master -   SimpleReport is NOT currently providing this value, so we really can't use it as our Patient PK.

prime-router/metadata/schemas/PrimeDataInput/pdi-covid-19.schema
```diff
@@ -22,184 +23,189 @@ elements:
...
 
-  - name: standard.patient_phone_number
+  - name: patient_phone_number
     csvFields: [{name: Patient_phone_number}]
 
-  - name: standard.patient_county
+  - name: patient_county
     csvFields: [{name: Patient_county}]
 
     # Eg, Staff, Resident, Visitor, Student
```
jimduff-usds, line 104: Hmm.   Very similar to some of the AOE questions, eg, "Works in healthcare?" and "Lives in congregate setting?"....

prime-router/metadata/schemas/PrimeDataInput/pdi-covid-19.schema
```diff
@@ -22,184 +23,189 @@ elements:
...
 
-  - name: standard.ordering_provider_phone_number
+  - name: ordering_provider_phone_number
     csvFields: [{name: Ordering_provider_phone_number}]
 
-  - name: standard.ordering_provider_county
+  - name: ordering_provider_county
     csvFields: [{name: Ordering_provider_county}]
 
+  - name: reporting_facility_name
```
jimduff-usds, line 207: ?   This brings up a question I've had for a while:   mappers tautologically work in one direction only, what does it mean when a schema is used for both incoming and outgoing data.    Do we need a 'reverse' mapper in some cases?

In the case here, SimpleReport is not currently sending us a `reporting_facility_name`, but the mapper tells how to create it from other data -- all good.   But what about when reading back in data _generated_ using this schema.   Now the `reporting_facility_name` field exists - shouldn't that trigger a "no-op" or "passthru-as-is" mapper?


You can see this more clearly in the example of how the AZ schema uses, but does not contain, the organization_name, which is only in the pdi schema.    AZ's schema says: 
```
   # Nonstandard
   - name: organization_and_facility
     mapper: 'concat(organization_name, ordering_facility_name)'
     csvFields: [{name: Ordering_facility_name}]
```

But AZ data, as created by the hub _no longer contains the organization_name column!_  So, if that data is read back into the Hub, this mapper, if it runs, should fail.   That is not happening, so I think there's an implied "no-op" or "passthru-as-is" mapper in that situation as well.

prime-router/metadata/schemas/covid-19.schema
jimduff-usds: Definitely, and I like the processing_mode_code too.     The schema that's for Pima needs the code for the mode. ;)

prime-router/metadata/valuesets/common.valuesets
```diff
@@ -90,20 +90,28 @@
   values:
     - display: American Indian or Alaska Native
       code: 1002-5
+      version: 2.5.1
```
jimduff-usds, line 93: Seems a shame we can't give the version for the whole valueSet.   its a big gloppy this way.

prime-router/metadata/valuesets/common.valuesets
```diff
@@ -223,4 +231,14 @@
...
       display: Weakly reactive
     - code: W
-      display: Worse--use when direction not relevant
\ No newline at end of file
+      display: Worse--use when direction not relevant
+
+- name: hl70103
+  system: HL7
+  values:
+    - code: D
```
jimduff-usds, line 239: Amusing that this isn't called "Testing", but I suspect its cause the T is already used for 'Training'

prime-router/src/main/kotlin/Schema.kt
jimduff-usds: At a summer job, I wrote an auto-documentation generator based on an invented "code decorating language" that years later turned out to be remarkably similar to javadoc.

prime-router/src/test/unit_test_files/pdi-covid-19.schema
```diff
@@ -22,176 +22,176 @@ elements:
...
-  - name: standard.ordering_provider_zip_code
+  - name: ordering_provider_zip_code
     csvFields: [{name: Ordering_provider_zip_code}]
 
-  - name: standard.ordering_provider_phone_number
+  - name: ordering_provider_phone_number
     csvFields: [{name: Ordering_provider_phone_number}]
 
-  - name: standard.ordering_provider_county
+  - name: ordering_provider_county
```
jimduff-usds, line 195: Phew - last one!

jimduff-usds (approved):
You can see my thoughts on schemas and such evolved a bit as I wrote my comments, but might spark some thoughts.

Deleting 50,000 instances of the word 'standard' is definitely a good deed done.  Thank you!

prime-router/metadata/schemas/AZ/az-covid-19-hl7.schema
RickHawesUSDS: Right. The AZ HL7 implementation guide is pretty complete and their CSV request is minimal.

prime-router/metadata/valuesets/common.valuesets
RickHawesUSDS: Right. The version can change within the same value set.

PR #197, opened 2020-12-15
Restart function app instead of update

cglodosky:
This PR restarts the function app instead of updating it upon deployment.


RickHawesUSDS (approved):
LGTM

PR #199, opened 2020-12-16
Improved report processing

### LINKED ISSUE #177: Processing modes to help on-boarding new partners
**Is your feature request related to a problem? Please describe.**
As a developer of new partner schemas, I would like to have processing modes that help me onboard to these partners.  

**Describe the solution you'd like**
[ ] The ability to validate messages before they are sent
[ ] The ability to send mark messages as test messages (ie. HL7 processing modes)
[ ] The ability to only-validate a message
[ ] OpenApi documentation updated
[ ] The ability to do a healthCheck request

**Proposed Processing Modes**
- onlyValidate 
- checkConnections
- skipSend
- trainingSend
- debugSend
- holdSend
- sendHeld
- emptyHeld

RickHawesUSDS:
This PR adds does multiple things with the aim to improve our reporting end-point. It merges three PRs from the week. In particular, the PR does the following

- improve our ability to onboard new senders and receivers
- support partial submissions (reports with a few bad rows)
- redo the error handling code to be make error messages more complete
- add a health check mode
- improve the report end-point's performance

## Changes
- OpenApi.yml file documents the current API
- "client" can be specified in a header or a query parameter
- "option=ValidatePayload" will only validate a pay, no other actions
- "option=CheckConnections" will only check the db connection that the function has. Good for a heath-check. 
- "option=SkipSend" will skip sending the reports, but it does the translations, so they can be verified.
- "default=<element>:<value>,..." will set a dynamic default
- Eliminated metadata as a global object. Allows dependency injection as well as performance improvements
- Combined the translation and validate step into one step. 
- New output format for the report end-point. Example
```
{
  "id": "979d3107-1a1b-4803-b9c0-1dabb1517b32",
  "itemCount": 39,
  "destinations": [
    "Sending 6 items to Arizona PHD at 2020-12-18T23:21-07:00",
    "Sending 2 items to Pima County, Arizona PHD at 2020-12-18T23:21-07:00"
  ],
  "rejectedItems": [
    5
  ],
  "details": [
    "Error: Invalid required value on item 5: Blank value for element 'placer_order_id'",
    "Warning: Invalid requested value on item 1: Blank value for element 'patient_id'"
  ]
}
```

## Checklist
- [x] Tested locally?
- [ ] Added new dependencies?
- [ ] Updated the docs?
- [x] Added a test?

## Fixes 

#177 
#199 

## To Be Done

- Actual values for usage
- Do work to look at translation errors and mapping errors.


prime-router/metadata/organizations-local.yml
```diff
@@ -67,6 +67,31 @@
...
+  - name: tx-phd
+    description: Texas Department of State Health Services
+    services:
+      - name: elr-test
+        topic: covid-19
+        schema: tx/tx-covid-19-hl7
+        jurisdictionalFilter:
+         - "matches(ordering_facility_state, TX)"
+         - "matches(processing_mode_code, T)"
+        transforms: { deidentify: false }
```
RickHawesUSDS, line 79: Different services based on the `processing_mode_code`

prime-router/metadata/schemas/PrimeDataInput/pdi-covid-19.schema
```diff
@@ -39,6 +39,7 @@ elements:
     csvFields: [{name: Instrument_ID}]
 
   - name: placer_order_id
+    usage: required
     csvFields: [{name: Result_ID}]
```
RickHawesUSDS, line 43: New `usage` field

prime-router/metadata/schemas/covid-19.schema
```diff
@@ -228,6 +229,7 @@ elements:
 
   - name: ordering_facility_email
     type: EMAIL
+    canBeBlank: true
     hl7Field: ORC-23-4
```
RickHawesUSDS, line 233: Another new field indicates that blank is a valid value for this field. Suppresses warnings and errors if this field is true.

prime-router/pom.xml
```diff
@@ -15,7 +15,7 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <kotlin.code.style>official</kotlin.code.style>
         <kotlin.compiler.jvmTarget>11</kotlin.compiler.jvmTarget>
-        <kotlin.compiler.incremental>true</kotlin.compiler.incremental>
+        <kotlin.compiler.incremental>false</kotlin.compiler.incremental>
         <kotlin.version>1.4.10</kotlin.version>
```
RickHawesUSDS, line 19: About 5 second improvement in `mvn clean package` times for my laptop.

prime-router/src/main/kotlin/CsvConverter.kt
```diff
@@ -8,30 +8,80 @@ import java.io.OutputStream
...
-object CsvConverter {
-    private data class Mapping(
+class CsvConverter(val metadata: Metadata) {
+    private data class CsvMapping(
         val useCsv: Map<String, Element.CsvField>,
         val useMapper: Map<String, Pair<Mapper, List<String>>>,
         val useDefault: Map<String, String>,
+        val errors: List<String>,
+        val warnings: List<String>,
     )
```
RickHawesUSDS, line 18: Errors and warnings are now passed back instead of throwing exceptions.

prime-router/src/main/kotlin/CsvConverter.kt
```diff
@@ -8,30 +8,80 @@ import java.io.OutputStream
 /**
  * A converter differs from a serialization in that
  */
-object CsvConverter {
-    private data class Mapping(
+class CsvConverter(val metadata: Metadata) {
+    private data class CsvMapping(
```
RickHawesUSDS, line 12: This class was almost completely rewritten.

prime-router/src/main/kotlin/Element.kt
```diff
@@ -244,7 +277,101 @@ data class Element(
     }
 
     /**
-     * Take a formatted CsvField value and turn into a normalized value stored in an element
+     * Take a formatted value and check to see if can be stored in a report.
+     */
+    fun checkForError(formattedValue: String, format: String? = null): String? {
+        if (formattedValue.isBlank() && !isOptional && canBeBlank != true) return "Blank value for element '$name'"
```
RickHawesUSDS, line 283: New function that returns error details and guards against formatting exceptions.

prime-router/src/main/kotlin/Hl7Converter.kt
```diff
@@ -8,11 +8,11 @@ import java.time.OffsetDateTime
 import java.time.ZoneId
 import java.util.Properties
 
-object Hl7Converter {
-    private const val softwareVendorOrganization = "Centers for Disease Control and Prevention"
-    private const val softwareProductName = "PRIME Data Hub"
+class Hl7Converter(val metadata: Metadata) {
+    private val softwareVendorOrganization = "Centers for Disease Control and Prevention"
```
RickHawesUSDS, line 12: Metadata is now a passed in object instead of a global.

prime-router/src/main/kotlin/Metadata.kt
```diff
@@ -9,24 +9,10 @@ import java.io.FilenameFilter
...
-
-    private val PRIME_ENVIRONMENT = System.getenv("PRIME_ENVIRONMENT") ?: ""
-
-    private val ext = if (PRIME_ENVIRONMENT.isNotEmpty()) "-" + PRIME_ENVIRONMENT else PRIME_ENVIRONMENT
-
-    private val organizationsList = "organizations$ext.yml"
-
-    private var schemas = mapOf<String, Schema>()
+class Metadata {
+    private var schemaStore = mapOf<String, Schema>()
```
RickHawesUSDS, line 15: Going from a global to a passed in value allows for dependency injection. It's a big change to our code in lines affected, but not so much from a conceptual point of view.

prime-router/src/main/kotlin/Metadata.kt
```diff
@@ -111,46 +125,93 @@ object Metadata {
...
-
     private fun normalizeSchemaName(name: String): String {
         return name.toLowerCase()
     }
 
+    /**
+     * The fixup process fills in references and inherited attributes.
+     */
+    private fun fixupElement(element: Element, baseElement: Element? = null): Element {
+        if (element.canBeBlank == true && element.default != null)
```
RickHawesUSDS, line 169: Element fixup now fixes up references to mappers, tables, and valuesets. Code which use elements is simpler and faster.

prime-router/src/main/kotlin/OrganizationService.kt
```diff
@@ -108,51 +108,4 @@ data class OrganizationService(
             // API
         }
     }
-
```
RickHawesUSDS, line 111: This code is moved into a new translation class.

prime-router/src/main/kotlin/Translator.kt
```diff
@@ -0,0 +1,117 @@
+package gov.cdc.prime.router
+
```
RickHawesUSDS, line 2: This is a new class that focuses on the process of translation. It pulls together code from the schema and metadata classes in one place.

prime-router/src/main/kotlin/azure/ReportFunction.kt
```diff
@@ -0,0 +1,309 @@
+package gov.cdc.prime.router.azure
```
RickHawesUSDS, line 1: The report function is combination of the old Validate and Translate functions. By doing both operations in one-step, it is now possible to do a much more complete validation which includes translation errors.

prime-router/src/main/kotlin/azure/WorkflowEngine.kt
```diff
@@ -1,23 +1,40 @@
...
  * @see QueueAccess
  * @see DatabaseAccess.Header
  */
 class WorkflowEngine(
-    private val db: DatabaseAccess = DatabaseAccess(),
-    private val blob: BlobAccess = BlobAccess(),
-    private val queue: QueueAccess = QueueAccess(),
+    // Immutable objects can be shared between every function call
+    val metadata: Metadata = WorkflowEngine.metadata,
+    val hl7Converter: Hl7Converter = WorkflowEngine.hl7Converter,
```
RickHawesUSDS, line 22: It metadata object is now loaded once, on the first call to a function. This speeds up processing considerably.

prime-router/src/test/unit_test_files/organizations.yml
```diff
@@ -1,46 +0,0 @@
----
```
jimduff-usds, line 1: There are similar .yml and .schema files in inside src/test/csv_test_files/input/.   This has worked well to keep those tests simple, isolated, and working, but maybe there's a better place for these to go now.

jimduff-usds (changes requested):
- My main request: I think our error json needs review.   Could be a ux sort of review or other.   In particular, if there are changes to the json structure, let's do them now, and not make Tim/Tom change twice.   Making the error messages readable is also crucial but could come later.
- Feels like the usage, default, and canBeBlank may have squirrely interactions with each other.   Need to document and test edge cases.  Even better, find a way to simplify, if possible.

prime-router/metadata/schemas/covid-19.schema
RickHawesUSDS: Remove canBeBlank as a field and add a couple of types where it is necessary. Here's logic table which I put in code. 

```
| Type         | Cardinality | no column on input                   | empty input value                    | Invalid input value | valid input value |
|--------------|-------------|--------------------------------------|--------------------------------------|---------------------|-------------------|
| FOO          | 0..1        | mapper -> default -> empty + warning | mapper -> default -> empty + warning | empty + warning     | value             |
| FOO          | 1..1        | mapper -> default -> error           | mapper -> default -> error           | error               | value             |
| FOO_OR_BLANK | 0..1        | mapper -> default -> empty + warning | empty                                | empty + warning     | value             |
| FOO_OR_BLANK | 1..1        | mapper -> default -> error           | empty                                | error               | value             |

```

prime-router/src/main/kotlin/Metadata.kt
RickHawesUSDS: Just one level for now, but I think we will need multiple levels as I do the Redox schemas.

prime-router/src/main/kotlin/Translator.kt
RickHawesUSDS: I'm thinking of renaming converter to serializers, which is the common Java term for these classes. I'm also thinking putting them in them in a separate folder.

prime-router/src/main/kotlin/azure/ReportFunction.kt
RickHawesUSDS: Agree. Can't do everything in a single pull request.

prime-router/src/main/kotlin/azure/WorkflowEngine.kt
RickHawesUSDS: The server process doesn't shutdown after a request.

prime-router/src/main/kotlin/CsvConverter.kt
```diff
@@ -8,30 +8,80 @@ import java.io.OutputStream
...
+         */
+        val report: Report?,
+        /**
+         * The list of rows that were rejected while reading
+         */
+        val rejected: List<Int> = emptyList(),
+        /**
+         * The list of errors that caused a item to not be reported
+         */
+        val errors: List<String> = emptyList(),
```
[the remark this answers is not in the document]
RickHawesUSDS: I like your idea. Here's what I shared at today's engineering sync for a format. 
```
{
    "id": null,
    "destinations": [],
    "errors": [
        {
            "scope": "REPORT",
            "id": "",
            "details": "Missing 'Result_ID' header"
        },
        {
            "scope": "REPORT",
            "id": "",
            "details": "Missing 'Patient_state' header"
        }
    ],
    "warnings": [
        {
            "scope": "REPORT",
            "id": "",
            "details": "Missing 'Patient_email' header"
        },
        {
            "scope": "REPORT",
            "id": "",
            "details": "Missing 'Ordering_facility_email' header"
        },
        {
            "scope": "ITEM",
            "id": "563423",
            "details": "'Patient_lookup_ID' value is invalid"
        },
        {
            "scope": "REPORT",
            "id": "",
            "details": "Unexpected 'Testing_lab_specimen_ID' header is ignored"
        }
    ]
}
```

prime-router/src/main/kotlin/CsvConverter.kt
```diff
@@ -64,77 +114,115 @@ object CsvConverter {
...
      */
-    private fun mapRow(schema: Schema, mapping: Mapping, inputRow: Map<String, String>): List<String> {
+    private fun mapRow(schema: Schema, csvMapping: CsvMapping, index: Int, inputRow: Map<String, String>): RowResult {
         val lookupValues = mutableMapOf<String, String>()
+        val errors = mutableListOf<String>()
+        val warnings = mutableListOf<String>()
 
-        return schema.elements.map { element ->
+        val row = schema.elements.map { element ->
             fun addToLookup(normalized: String): String {
```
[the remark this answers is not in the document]
RickHawesUSDS: I've added comments and restructured the code. It should be clearer now.

prime-router/metadata/schemas/covid-19.schema
jimduff-usds: I think there's still room for further simplification.  note that
`FOO 0..1`   and 'FOO_OR_BLANK 0..1`   are nearly identical - the minor difference that one gives a Warning on empty input values, and the other is silent.

prime-router/src/main/kotlin/Translator.kt
jimduff-usds: Yes to both the rename and the separate folder.

prime-router/src/main/kotlin/CsvConverter.kt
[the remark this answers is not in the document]
jimduff-usds: Yes, this seems ok.   Suggested further improvements for a future PR, and a bug:

- I think people will want to see report level problems first, then row-level.    So, a simple suggestion is to sort the ones with REPORT scope to the top of the `errors` array, and similar for `warnings`.   Keep in mind that the first time someone sees this, its likely to be a gigantic daunting list.

- I get why you are not putting in row numbers, because other formats don't have rows, but rownums sure would be easier on the reader.    At minimum, lets at least make sure all the ITEM level issues appear in the same order as the rows in the file.

- A good readability improvement for the user would be to explicitly use the column name used by the sender for the id.  For example, the primary key for a row in Simple Report is called the Result_ID in the file, so...
```
        {
            "scope": "ITEM",
            "id": "Result_ID:563423",      # Or alternatively simply    "Result_ID" : "563423"
            "details": "'Patient_lookup_ID' value is invalid"
        },
```

- And for extra bonus points and badges:  If _every_ ITEM in a column has the same error, report that as a REPORT level issue, rather than repeating the same error a zillion times.   This will be very common - if the phone number is misformatted, its likely to be globally wrong.


- Oooh, and I just found a bug while looking at this:   pdi-covid-19 has
    `trackingElement: testing_lab_specimen_id`
but `testing_lab_specimen_id` is no longer even in the data!   That should be an error on metadata load.

prime-router/src/main/kotlin/azure/ReportFunction.kt
```diff
@@ -0,0 +1,309 @@
...
+class ReportFunction {
+    private val clientParameter = "client"
+    private val optionParameter = "option"
+    private val defaultParameter = "default"
+    private val defaultSeparator = ":"
+    private val jsonMediaType = "application/json" // TODO: find a good media library
+
+    enum class Options {
+        None,
+        ValidatePayload,
```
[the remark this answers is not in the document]
RickHawesUSDS: Agree.

prime-router/src/main/kotlin/CsvConverter.kt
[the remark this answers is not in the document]
RickHawesUSDS: Good stuff. Let's pull this in another PR.

jimduff-usds (approved):

PR #206, opened 2020-12-17
multi-transport

mry-usds:
This PR adds mutli-transport capabiliteis to the system -- transport => transports in organizations.yml files

## Changes
- Added multi-transports
- Swapped out JSCH for SSHJ (JSCH was impacted by the transaction mechanism)

## Checklist
- [X] Tested locally?
- [X] Added new dependencies?
- [ ] Updated the docs?
- [ ] Added a test?

## Fixes 

- #issue

## To Be Done

- #issue


prime-router/metadata/organizations-local.yml
```diff
@@ -64,7 +64,6 @@
           numberPerDay: 1440 # Every minute.  NOTE:  In production, they want deliveries once a day at 8am MST.
           initialBatch: 00:00
           timeZone: ARIZONA
-        address: http://localhost:1181/
         format: CSV
```
RickHawesUSDS, line 67: Thank you.

prime-router/pom.xml
```diff
@@ -447,23 +448,11 @@
...
-        <dependency>
-            <groupId>com.github.mhewedy</groupId>
-            <artifactId>sftp-utils</artifactId>
-            <version>2.0.0</version>
-        </dependency>        
+<dependency>
+    <groupId>com.hierynomus</groupId>
+    <artifactId>sshj</artifactId>
+    <version>0.27.0</version>
+</dependency>
```
RickHawesUSDS, line 455: Looks like a good project.

prime-router/src/main/kotlin/transport/SftpTransport.kt
```diff
@@ -1,58 +1,90 @@
...
+
+    private fun uploadFile(
+        host: String,
+        port: String,
+        user: String,
+        pass: String,
+        path: String,
+        contents: ByteArray
+    ) {
+        val sshClient: SSHClient = setupSshj(host, port, user, pass);
```
RickHawesUSDS, line 76: Suggest that this block should protected against exceptions being thrown. Either a `sshClient.use { }` block since sshClient implements closable or a try-catch block if disconnect is really needed.

RickHawesUSDS (changes requested):

RickHawesUSDS:
Look's real good. Just one small suggestion.

prime-router/metadata/organizations-local.yml
mry-usds: NP...

prime-router/src/main/kotlin/transport/SftpTransport.kt
mry-usds: Changed to use blocks for sshClient and sftpClient

RickHawesUSDS (approved):
Thanks for the changes. LGTM

PR #210, opened 2020-12-18
Move ACR Push to Release Phase

cglodosky:
Due to limitations in passing a build artifact between workflows, I updated the CI/CD workflow to redo the build stage within the release workflow.  This will prevent and upstream PR from building and pushing a docker image that is not in sync with the one that is to be deployed by the release workflow as all are referencing the `latest` image tag.  This can be iterated upon, but for now this will avoid any unwanted by-products of multiple builds happening at the same time.


RickHawesUSDS (approved):
LGTM

Issue #109, opened 2020-11-30
Florida Go Live Checklist


jlusds:
Overview doc = https://drive.google.com/file/d/1uCKyeW4hh2wmpahD3JHcJ9jH66sodyt0/view?usp=sharing

Checklist
- [x]  Intro 
  - [x] get documents https://drive.google.com/drive/folders/1Oed8k5AfqxdouQTPuS3bDRedNGrc6hxM
  - [x] submit registration
- [x]  Receive SFTP credentials
- [x]  Confirm data schema
Elr documentation = https://drive.google.com/file/d/1my9ftaEcqzECjWYCwnWu5VuY_2AeYQh9/view?usp=sharing
CSV = https://cdc.sharepoint.com/:x:/t/USDSatCDC/EQ44_DXxLbREub1DbMB3dw8BCuumC8UkF-8o6AK8QuXUuQ?e=EGAO2B

- [x]  Send test file
- [x]  Confirm new site registration process
- [ ] Go live

drew-usds:
Sending CSV for now. Waiting on final steps/feedback from FL.

MauriceReeves-usds:
I updated the fake data generation and sent a new file to FL this morning.

MauriceReeves-usds:
This is still in progress. We have sent them another file, but we are waiting to review with them. We will send them more synthetic data this week.

jlusds:
Blocked by first new sender

jlusds:
Another file sent on Friday, Waiting for approval for HL7

PR #215, opened 2020-12-22
Florida Test File Creation Updates

MauriceReeves-usds:
This PR does is a set of changes to make the creation of the FL test file go more smoothly.

## Changes
- Changes the FL schema to use more codes for fields where it makes sense
- Added some Florida-specific value sets. I'm not crazy about a file per state for their own value sets, but having a single file for all states might be unwieldy. I'm open to suggestions
- Updated the FakeReport generator to take alternate values into account for other schemas and return logical values for some of the FL AOE fields
- Tests

## Checklist
- [x] Tested locally?
- [ ] Added new dependencies?
- [x] Updated the docs?
- [x] Added a test?

## Fixes 

- #109


prime-router/metadata/schemas/FL/fl-covid-19.schema
```diff
@@ -43,9 +43,15 @@ elements:
       csvFields: [{ name: Patient Ethnicity }]
 
     # column I
-    # TODO: does this need a custom mapper? ELR FAQ v2.6 has CSV specific values for HL7 codes
     - name: patient_gender
       csvFields: [{ name: Patient Gender }]
+      altValues:
```
jimduff-usds, line 48: In cases where there are lots of different ways of saying the same thing, it occurs to me that we should move to using @RickHawesUSDS 's `TABLE` type rather than CODE.    Then each time a state has a new "creative" way of specifying something we can just add a column to the table.

prime-router/metadata/schemas/FL/fl-covid-19.schema
```diff
@@ -75,7 +81,7 @@ elements:
     # column C part 2
     # nonstandard
     - name: patient_ssn
-      type: TEXT
+      type: ID_SSN
```
jimduff-usds, line 84: Maybe note that we aren't actually collecting this at the moment?
@RickHawesUSDS 's new options stuff is meant to cover this oxymoron situation where a field is required but its also OK to be blank.

prime-router/metadata/schemas/FL/fl-covid-19.schema
```diff
@@ -196,7 +202,10 @@ elements:
       csvFields: [{ name: Reference Range }]
 
     # column AR
+    # nonstandard
```
jimduff-usds, line 205: I've been using "nonstandard" to mean "not in covid-19.schema".   Since abnormal_flag is in covid-19.schema, I'm a bit worried about future weirdness in the code: will it choose the FL Valueset specified here, or the one specified in covid-19.schema.   Suggest you use a different `name` for it.

prime-router/metadata/schemas/FL/fl-covid-19.schema
```diff
@@ -208,11 +217,19 @@ elements:
       documentation: This is the coded value that describes the result. For IgG, IgM and CT results that provide a value leave this field blank.
 
     # column AT
-    - name: testing_lab_name
+    # nonstandard. need to concat because we currently can't add the same column twice to a table
+    # TODO: fix this
+    - name: name_of_testing_lab
+      type: TEXT
+      mapper: 'concat(testing_lab_name)'
```
jimduff-usds, line 224: I don't understand what's the requirement here?  To show the same lab name string under two different column headers?  That would be 
`csvFields: [{ name: Foo}, {name: Bar}]` 
?

prime-router/metadata/schemas/FL/fl-covid-19.schema
```diff
@@ -208,11 +217,19 @@ elements:
...
       csvFields: [{ name: Performing Lab Name }]
 
     # column AU
-    - name: testing_lab_clia
+    # nonstandard. need to concat because we currently can't add the same column twice to a table
+    # TODO: fix this
+    - name: clia_of_testing_lab
+      type: TEXT
```
jimduff-usds, line 231: Same question:  I'm not getting what FL wants here.

prime-router/metadata/schemas/FL/fl-covid-19.schema
```diff
@@ -250,7 +266,6 @@ elements:
 
     # column AY
     - name: employed_in_healthcare
-      type: CODE
       csvFields: [{ name: Employment in Health Care }]
       altValues:
           - display: Yes
```
jimduff-usds, line 271: I suppose we should be grateful there are so may ways to say 'Yes'. ;) 
Per my previous comment, let's create a `TABLE` with all of them.

prime-router/metadata/schemas/FL/fl-covid-19.schema
```diff
@@ -335,9 +347,9 @@ elements:
 
     # column BG
     # nonstandard
-    # todo: create a value set for this based on SNOMED codes
     - name: specify_congregate_care_setting
```
jimduff-usds, line 350: Some comment as elsewhere - I'm concerned that our code might have weird bugs with this kind of "override".  Suggest using a different name for the field, not the name in covid-19.schema.

prime-router/metadata/schemas/FL/fl-covid-19.schema
```diff
@@ -347,23 +359,20 @@ elements:
...
+          - display: Pregnant
+            code: Y
+          - display: Not Pregnant
+            code: N
+          - display: Unknown
+            code: U
 
     # column BI
     # nonstandard
     - name: student_teacher_other
```
jimduff-usds, line 372: It would be _really_ nice to just add Student and Teacher to patient_role - I think this would be easier for SimpleReport too - they don't want to have two different dropdown, one for student/teacher and one for staff/resident, ?

prime-router/metadata/valuesets/florida.valuesets
```diff
@@ -0,0 +1,55 @@
...
+
+- name: fl/abnormal_flag
+  system: LOCAL
+  values:
+      - display: Abnormal
+        code: A
+      - display: Normal
+        code: N
+
+- name: fl/congregate_care_setting
```
jimduff-usds, line 23: This appears to be a standard type-of-institution / environment value in SNOMED.   I think this can just go straight into common.valuesets.

prime-router/src/main/kotlin/FakeReport.kt
```diff
@@ -49,6 +49,17 @@ class FakeReport {
                     when {
                         element.nameContains("lab_name") -> "Any lab USA"
                         element.nameContains("facility_name") -> "Any facility USA"
+                        element.nameContains("patient_age_and_units") -> {
```
jimduff-usds, line 52: Interesting.  Our very-large-hospital customer mentioned to us that some states object strongly to being sent an Age field, since its redundant with birthday.    Other states _require_ it.    Sigh.

prime-router/src/main/kotlin/FakeReport.kt
```diff
@@ -71,9 +82,17 @@ class FakeReport {
                     when (element.name) {
                         "specimen_source_site_code" -> "71836000"
                         else -> {
+                            val altValues = element.altValues
                             val valueSet = findValueSet(element.valueSet ?: "")
                                 ?: error("ValueSet ${element.valueSet} is not available}")
-                            val possibleValues = valueSet.values.map { it.code }.toTypedArray()
+                            // if the code defines alternate values in the schema we need to
+                            // output them here
+                            val possibleValues = if (altValues?.isNotEmpty() == true) {
```
jimduff-usds, line 90: I'm not a fan of the design of this alternate value concept in general.   Again, I think the new `TABLE` idea might work better.

jimduff-usds (changes requested):
Most of my comments (eg, the stuff about `TABLE` are for the future.

The 'Request changes' is mainly for these two questions:
- Wondering if there's already a solution for 'name_of_testing_lab', 'clia_of_testing_lab', etc.
- Worried about overriding names of fields that are in the covid-19.schema (eg abnormal_flag), but with new value sets.  Why not use the $alt thing, or just go ahead and attach a Valueset to the field in covid-19.schema.

prime-router/metadata/schemas/FL/fl-covid-19.schema
RickHawesUSDS: Please don't put SSN into our data.

prime-router/src/main/kotlin/FakeReport.kt
```diff
@@ -118,7 +137,11 @@ class FakeReport {
                         else -> TODO()
                     }
                 }
-                Element.Type.TELEPHONE -> faker.numerify("##########:1:") // faker.phoneNumber().cellPhone()
+                Element.Type.TELEPHONE -> {
+                    // schema fields can define their own special format
+                    val formatString = element.csvFields?.get(0)?.format ?: "##########:1:"
+                    faker.numerify(formatString)
+                }
                 Element.Type.EMAIL -> "${context.patientName.username()}@email.com"
```
RickHawesUSDS, line 145: Look's good.

RickHawesUSDS (changes requested):
Look's good, but I don't want a SSN in our schemas.

RickHawesUSDS (changes requested):

prime-router/metadata/schemas/FL/fl-covid-19.schema
jimduff-usds: change to local_testing_lab_name.
Use use(testing_lab_name)

prime-router/metadata/schemas/FL/fl-covid-19.schema
```diff
@@ -347,23 +361,20 @@ elements:
...
       type: CODE
+      valueSet: fl/student_teacher_other
       csvFields: [{ name: "Is the patient a student, teacher, or other faculty member" }]
-      altValues:
```
jimduff-usds, line 356: Change to patient_role, and make that use a covid-19.valueset valueSet of these values.
Then write a one-off altValues here to cover the Other situation.

prime-router/metadata/schemas/FL/fl-covid-19.schema
MauriceReeves-usds: We won't be. I just wanted to mark the field as what it is, but we don't need to send it through. It can be blank.

prime-router/metadata/schemas/FL/fl-covid-19.schema
```diff
@@ -347,23 +359,20 @@ elements:
     # column BH
     - name: pregnant
       csvFields: [ { name: Pregnancy Status } ]
+      altValues:
+          - display: Pregnant
+            code: Y
+          - display: Not Pregnant
+            code: N
+          - display: Unknown
+            code: U
```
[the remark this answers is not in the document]
MauriceReeves-usds: As discussed on our call, because FL wants to use special values for "Unknown."

prime-router/metadata/schemas/FL/fl-covid-19.schema
MauriceReeves-usds: `patient_role` doesn't exist yet in the standard schema, and LOINC are all listed as TBD. This is a little different than occupation, though there's some overlap. That said `patient_occupation` is also not in the schema yet. We should probably sit with SimpleReport and discuss how they're capturing that information. I'd like to leave this as is until we start adding LOINC values or build in the occupation field.

prime-router/metadata/valuesets/florida.valuesets
MauriceReeves-usds: I moved `congregate_care_setting` into common.valuesets and named it `site_of_care`.

prime-router/metadata/organizations-local.yml
```diff
@@ -91,6 +91,16 @@
         transforms: { deidentify: false }
         format: HL7
 
+  - name: fl-phd
+    description: Florida Department of Health
+    services:
+      - name: elr
+        topic: covid-19
+        schema: fl/fl-covid-19
+        jurisdictionalFilter: [ "matches(ordering_facility_state, FL)" ]
```
jimduff-usds, line 100: Check with the PHD on what rule they want for the long term.   In AZ, it was if the resident's address is in AZ __or__ if the ordering facility was in AZ.

prime-router/quick-test.sh
```diff
@@ -18,8 +18,30 @@ RED='\033[0;31m'
 GREEN='\033[0;32m'
 NC='\033[0m' # No Color
 
-# For testing the unhappy path
+# let's make it possible for run for different states
```
jimduff-usds, line 21: Thanks for this improvement!

prime-router/src/main/kotlin/Element.kt
```diff
@@ -111,6 +111,7 @@ data class Element(
         PERSON_NAME,
         TELEPHONE,
         EMAIL,
+        BLANK,
```
jimduff-usds, line 114: I'm guessing this is for SSN.

prime-router/src/main/kotlin/cli/main.kt
```diff
@@ -59,6 +58,12 @@ class RouterCli : CliktCommand(
     private val includeTimestamps by
     option("--include-timestamps", help = "includes creation time stamps when generating documentation")
         .flag(default = false)
+    private val targetState: String? by
```
jimduff-usds, line 61: Super!

prime-router/src/main/kotlin/FakeReport.kt
```diff
@@ -6,15 +6,15 @@ import java.util.Random
 import java.util.concurrent.TimeUnit
 
 class FakeReport(val metadata: Metadata) {
-    class RowContext(findLookupTable: (String) -> LookupTable?) {
+    class RowContext(findLookupTable: (String) -> LookupTable? = { null }, reportState: String? = null) {
```
jimduff-usds, line 9: Nice

prime-router/src/main/kotlin/FakeReport.kt
```diff
@@ -116,19 +130,23 @@ class FakeReport(val metadata: Metadata) {
                     else -> TODO()
                 }
             }
-            Element.Type.TELEPHONE -> faker.numerify("2#########:1:") // faker.phoneNumber().cellPhone()
```
jimduff-usds, line 119: The problem was that our phone number Translator was breaking occasionally on FakeReport generated numbers with leading zeros or ones.   The `2` was a hack fix for that.  Have we fixed this in some better way?

prime-router/src/test/kotlin/SimpleReportTests.kt
```diff
@@ -151,13 +150,106 @@ class SimpleReportTests {
...
+            actualKeys.minus(expectedKeys).count() == 0,
+            "There are keys in actual that are not in expected: ${actualKeys.minus(expectedKeys).joinToString { "," }}"
+        )
+        assertTrue(
+            expectedKeys.minus(actualKeys).count() == 0,
+            "There are keys in expected that are not present in actual: ${expectedKeys.minus(actualKeys).joinToString { "," }}"
+        )
+    }
+
+    private fun compareLinesOfMaps(
```
jimduff-usds, line 194: Cool!   We could even move this into "regular" code, and expose it as a general utility.     Example:  our customers may want a simple way to programmatically compare and validate two CSVs - we could expose this via an API someday.

jimduff-usds (approved):
Sorry to take so long to get to this.   Just had a couple minor comments; could be tweaked in a future pr.

prime-router/src/test/kotlin/SimpleReportTests.kt
MauriceReeves-usds: I think that would be good as well. It'll be useful down the road.

MauriceReeves-usds:
@RickHawesUSDS I've added a BLANK type for Elements now to always pass a blank column for SSN. Can you approve my PR so we can get this in before any more big changes come in? Thank you!

RickHawesUSDS (approved):
LGTM. Please add a task to replace BLANK with Cardinality Zero.

PR #217, opened 2020-12-26
AZ HL7 changes

RickHawesUSDS:
This PR improves the HL7 processing, in general, and improves the HL7 message for AZ. 

## Changes
- A new type: Entity Identifier (EI) to represent the `message_profile_id`
- A new mapper, obx8, to calculate the `abnormal_flag` from the `test_result`
- `message_profile_id` now matches AZ spec
- HL70078 valueset for abnormal flag

## Checklist
- [x] Tested locally?
- [ ] Added new dependencies?
- [ ] Updated the docs?
- [x] Added a test?

### DIFF
--- a/prime-router/src/main/kotlin/Element.kt
+++ b/prime-router/src/main/kotlin/Element.kt
@@ -130,6 +130,12 @@ data class Element(
         val universalIdSystem: String?
     )
 
+    data class SubValue(
+        val name: String,
+        val value: String,
+        val format: String?
+    )
+
     /**
      * @property ZERO_OR_ONE Can be null or present (default)
      * @property ONE Must be present, error if not present
@@ -511,11 +517,11 @@ data class Element(
             Type.EI -> {
                 when (format) {
                     null,
-                    hdCompleteFormat -> {
+                    eiCompleteFormat -> {
                         parseEI(formattedValue) // to check
                         formattedValue
                     }
-                    hdNameToken -> {
+                    eiNameToken -> {
                         val ei = parseEI(formattedValue)
                         ei.name
                     }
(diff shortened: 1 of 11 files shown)


prime-router/src/main/kotlin/Hl7Converter.kt
```diff
@@ -193,11 +208,6 @@ class Hl7Converter(val metadata: Metadata) {
         terser.set("MSH-16", "NE")
         terser.set("MSH-12", "2.5.1")
         terser.set("MSH-17", "USA")
-        // Values that NIST requires (although they are not part 2.5.1)
-        terser.set("MSH-21-1", "PHLabReportNoAck")
-        terser.set("MSH-21-2", "ELR_Receiver")
-        terser.set("MSH-21-3", "2.16.840.1.113883.9.11")
-        terser.set("MSH-21-4", "ISO")
```
RickHawesUSDS, line 211: Not hard code now.

prime-router/src/test/kotlin/ElementTests.kt
```diff
@@ -219,4 +219,99 @@ internal class ElementTests {
         assertEquals(parent.documentation, child.documentation, "Documentation value didn't carry over from parent")
         assertEquals(parent.type, child.type, "Element type didn't carry over from parent")
     }
+
+    @Test
+    fun `test normalize and formatted round-trips`() {
+        val postal = Element(
```
RickHawesUSDS, line 225: A new set of tests to try formatting and normalize roundtrip.

prime-router/src/main/kotlin/Element.kt
```diff
@@ -460,19 +501,101 @@ data class Element(
...
+                    }
+                    else -> error("Schema Error: invalid format value")
+                }
             }
             else -> formattedValue
         }
     }
 
+    fun toNormalized(subValues: List<SubValue>): String {
+        if (subValues.isEmpty()) return ""
```
RickHawesUSDS, line 536: New function to read multiple subFields of an element.

prime-router/src/test/kotlin/ElementTests.kt
jimduff-usds: Excellent!

prime-router/src/main/kotlin/Element.kt
```diff
@@ -460,14 +495,32 @@ data class Element(
...
             }
             Type.HD -> {
-                // No matter what data value is, overwrite with our hardcoded default.
-                // By definition, we're the sending_application!
-                //
-                // Note:  This hack 'fixes' a bug in the Send function where data is read back in, and the
-                // incoming value is split between two fields: sending_application and sending_application_id
-                // There's currently no function to combine these in toNormalized(),
-                // so that it gets properly split apart again later in toFormatted.
-                this.default ?: ""
```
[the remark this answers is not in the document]
jimduff-usds: It looks like we are allowing the breaking-apart and recombining of subfields only specific narrow cases, that is, for EI ant HD types only?    

This makes sense to me; we may want to generalize this someday (what if a PHD wants to separate street number, street name, and apartment number? Gack), but this stuff is fragile code, so it makes sense to keep it confined for now.

OTOH, here's an example we already have of this:   AZ wants random stuff put into the "Notes" field.  Upon reading AZ data back in ("roundtrip" as you call it), in an ideal world we should be able to parse that out, and direct that data to the correctly named fields.     Will we use a "^" to separate the sub values in the "Notes" field, or will we use json, or what?   Food for thought.

jimduff-usds (approved):
Looks good.  No immediate changes suggested, just a couple thoughts.

PR #218, opened 2020-12-27
Add key for SimpleReport

cglodosky:
This PR adds a key for the SimpleReport client to communicate with the `reports` endpoint.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -23,6 +23,7 @@ jobs:
                    echo >> $GITHUB_ENV RESOURCE_GROUP=prime-data-hub-prod
                    echo >> $GITHUB_ENV ACR_REPO=primedatahubprod.azurecr.io
                    echo >> $GITHUB_ENV AZ_FXNS_DEFAULT_KEY=${{ secrets.AZ_FXNS_PROD_DEFAULT_KEY }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -81,6 +83,15 @@ jobs:
...
               --key-value ${{ env.AZ_FXNS_DEFAULT_KEY }}
 
+      - name: Set Azure Function SimpleReport Key
+        run: |
+              az functionapp function keys set \
+              --resource-group ${{ env.RESOURCE_GROUP }} \
+              --name ${{ env.RESOURCE_GROUP }} \
+              --function-name reports \
+              --key-name SimpleReport \
+              --key-value ${{ env.AZ_FXNS_SIMPLEREPORT_KEY }}
```
jimduff-usds, line 93: Hmm.   This seems OK for SimpleReport, but it doesn't scale to have to change our git workflow every time we onboard or offboard a client of the hub.  Still, given that we plan to switch away from keys to a better auth mechanism, this seems ok for now.

jimduff-usds (approved):
Feels like the time is ripe to revisit our use of access keys, and replace.

.github/workflows/release.yml
cglodosky: I agree - this was a stop-gap fix until we have a secure storage spot for authentication credentials.

PR #219, opened 2020-12-28
Data Source Fix

RickHawesUSDS:
This PR fixes a problem where the database connections would become exhausted in Azure. 

## Problem Description

This problem was found in production and reproduced locally. My local build would exhaust after 10 successful requests. The connection pool was also set to a connection size of 10 — more than a coincidence.  The code was not releasing database connections as expected. The fix was to switch Jooq to use a data source provider. Data sources are what is used if you have a connection pool. The fix was tested by posting 1000 times to my local instance. 

## Changes
- Use a data-source configuration in Jooq
- Fixed organization-local.yml to support the new multi-transport

## Checklist
- [x] Tested locally?
- [ ] Added new dependencies?
- [ ] Updated the docs?
- [x] Added a test?
- [ ] Did you check for sensitive data and remove any?

## Fixes 

- #issue

## To Be Done

- #issue

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@ hs_err_pid*
 # IDEA files
 **/.idea/**
 **/*.iml
+**/.run/**
 
 # Play files
 **/result_files/*.csv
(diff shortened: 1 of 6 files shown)


jimduff-usds (approved):
Looks good.

PR #222, opened 2020-12-29
docker-compose changes

mry-usds:
This PR fixes docker-compose

## Changes
- trapped for no transports
- removed address from organizations-local.yml

## Checklist
- [X] Tested locally?
- [ ] Added new dependencies?
- [ ] Updated the docs?
- [ ] Added a test?

## Fixes 

- #issue

## To Be Done

- #issue


RickHawesUSDS (approved):
LGTM

prime-router/src/main/kotlin/azure/SendFunction.kt
```diff
@@ -34,15 +34,17 @@ class SendFunction {
...
+                        when (it) {
+                            is SFTPTransportType -> SftpTransport().send(service.fullName, it, header, content)
+                            else -> true
+                        }
+                    }.reduce { acc, s -> acc && s }
+
+                    if (success) {
+                        context.logger.info("Sent report: ${header.task.reportId} to ${service.fullName}")
                     }
-                }.reduce { acc, s -> acc && s }
```
RickHawesUSDS, line 42: The .map function appears to throw the exception, but also call the following reduce. So, I'm seeing logs that have both an SFTP exception and the map-reduce exception. Here's my non-functional sytle suggestion that avoids the map function.
```
                if (service.transforms.isNotEmpty()) {
                    var success: Boolean = true
                    service.transports.forEach {
                        success = success && when (it) {
                            is SFTPTransportType -> SftpTransport().send(service.fullName, it, header, content)
                            else -> true
                        }
                    }
                    if (success) {
                        context.logger.info("Sent report: ${header.task.reportId} to ${service.fullName}")
                    }
                }
```

RickHawesUSDS (changes requested):
So, where is the docker-compose.yml change?

prime-router/src/main/kotlin/azure/SendFunction.kt
mry-usds: Inserted...

prime-router/src/main/kotlin/azure/SendFunction.kt
```diff
@@ -34,15 +34,17 @@ class SendFunction {
 
                 val content = workflowEngine.readBody(header)
 
-                val success: Boolean = service.transports.map {
-                    when (it) {
-                        is SFTPTransportType -> SftpTransport().send(service.fullName, it, header, content)
-                        else -> true
+                if (service.transforms.isNotEmpty()) {
+                    var success: Boolean = true
```
RickHawesUSDS, line 38: My suggestion had a typo. Replace `transforms` with `transports`

RickHawesUSDS (changes requested):

prime-router/src/main/kotlin/azure/SendFunction.kt
mry-usds: updated...

RickHawesUSDS (approved):

PR #225, opened 2020-12-29
Remove Setting Fxn App Keys

cglodosky:
Setting the function app keys is now removed.  This will cause the occasional failure if the function app does not restart quickly enough.  Since a restart does not destroy the keys - they do not need to be set with each deploy.  Instead, they can be managed via the Azure portal.


jimduff-usds (approved):
Super!

RickHawesUSDS (approved):
LGTM

PR #259, opened 2021-01-08
PA and STRAC changes

RickHawesUSDS:
This PR supports the changes requested by PA as well as the changes requested by Redox

## Changes
- One destination per PA county. Initially provisioning 3 counties. 
- STRAC's ID and KIT ID mapped placer and filler order number
- Filling the producer field with the school name
- Better logging in the Redox transport functions. 

## Checklist
- [x] Tested locally?
- [ ] Added new dependencies?
- [ ] Updated the docs?
- [ ] Added a test?

## To Be Done

- There is an issue on how to report the County PHD CLIA that is being resolved by Redox

### DIFF
--- a/prime-router/metadata/organizations-local.yml
+++ b/prime-router/metadata/organizations-local.yml
@@ -142,15 +142,47 @@
   - name: pa-phd
     description: Pennsylvania Department of Health
     services:
-      - name: elr-redox-debug
+      - name: elr-bucks-local
         topic: covid-19
         schema: pa/pa-covid-19-redox
-        jurisdictionalFilter: [ "matches(ordering_facility_state, PA)"]
+        jurisdictionalFilter: [ "filterByCounty(PA, Bucks)" ]
         transforms: { deidentify: false }
         defaults:
           processing_mode_code: T
-          redox_destination_id: cd019512-8a58-40e5-a416-4e597404f9b8
-          redox_destination_name: "CDC Pennsylvania Dept of Health Destination (s)"
+          redox_destination_id: e0d33443-c134-4e5f-9c15-69f9ba6340bf
+          redox_destination_name: "CDC Bucks County PDH Destination (s)"
+          redox_source_id: d89d4057-930f-4c5b-bb39-8cddb326e928
+          redox_source_name: "Prime Data Hub (Staging)"
+        format: REDOX
+        transports:
+          - type: REDOX
+            apiKey: some_key
+            baseUrl: http://redox:1080
+      - name: elr-chester-local
+        topic: covid-19
+        schema: pa/pa-covid-19-redox
+        jurisdictionalFilter: [ "filterByCounty(PA, Chester)" ]
+        transforms: { deidentify: false }
+        defaults:
+          processing_mode_code: T
+          redox_destination_id: 86aa61ac-8864-417f-860e-d83ae46c951f
+          redox_destination_name: "CDC Chester County PDH Destination (s)"
+          redox_source_id: d89d4057-930f-4c5b-bb39-8cddb326e928
+          redox_source_name: "Prime Data Hub (Staging)"
+        format: REDOX
+        transports:
+          - type: REDOX
+            apiKey: some_key
+            baseUrl: http://redox:1080
+      - name: elr-chester-local
+        topic: covid-19
+        schema: pa/pa-covid-19-redox
+        jurisdictionalFilter: [ "filterByCounty(PA, Montgomery)" ]
+        transforms: { deidentify: false }
+        defaults:
+          processing_mode_code: T
+          redox_destination_id: 21485dc8-8a6e-49d3-8b80-46531e015039
+          redox_destination_name: "CDC Montgomery County PDH Destination (s)"
           redox_source_id: d89d4057-930f-4c5b-bb39-8cddb326e928
           redox_source_name: "Prime Data Hub (Staging)"
         format: REDOX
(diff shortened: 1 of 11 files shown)


prime-router/metadata/organizations-local.yml
```diff
@@ -128,15 +128,47 @@
   - name: pa-phd
     description: Pennsylvania Department of Health
     services:
-      - name: elr-redox-debug
+      - name: elr-bucks-local
+        topic: covid-19
```
RickHawesUSDS, line 132: We need to use different Redox destinations for each PA county. It's sad, but true.

prime-router/metadata/organizations-prod.yml
```diff
@@ -83,21 +83,51 @@
         transforms: { deidentify: false }
         format: CSV
 
-#  When PA gives use a prod
 #  - name: pa-phd
 #    description: Pennsylvania Department of Health
```
RickHawesUSDS, line 87: Still don't have PROD provisioned. Waiting for OK from PA.

prime-router/src/main/kotlin/FakeReport.kt
```diff
@@ -16,10 +16,10 @@ class FakeReport(val metadata: Metadata) {
...
         val state = reportState ?: randomChoice("FL", "PA", "TX", "AZ", "CO")
         val county = findLookupTable("fips-county")?.let {
-            if (state == "AZ") {
-                randomChoice("Pima", "Yuma")
-            } else {
-                randomChoice(it.filter("State", state, "County"))
+            when (state) {
+                "AZ" -> randomChoice("Pima", "Yuma")
+                "PA" -> randomChoice("Bucks", "Chester", "Montgomery")
+                else -> randomChoice(it.filter("State", state, "County"))
```
RickHawesUSDS, line 22: We are dealing with 3 specific counties in PA.

prime-router/src/main/kotlin/transport/RedoxTransport.kt
```diff
@@ -8,6 +8,8 @@ import com.github.kittinunf.result.Result
 import gov.cdc.prime.router.RedoxTransportType
 import gov.cdc.prime.router.ReportId
 import gov.cdc.prime.router.TransportType
+import org.apache.logging.log4j.LogManager
+import org.apache.logging.log4j.Logger
```
RickHawesUSDS, line 13: Just adding better logging to the transport.

prime-router/metadata/organizations-local.yml
jimduff-usds: We are approaching the time when we need to move everything in organizations.yml into the database, or, at least outside of the build, so we can modify without a deploy.

prime-router/src/main/kotlin/transport/RedoxTransport.kt
jimduff-usds: OK, bleh:  this may moot the change I just put in my PR (just merged, https://github.com/CDCgov/prime-data-hub/pull/260) in which I pass in the Azure function context into ITransport, to do logging.   Either way, the lack of clarity on what was being sent where was making it very hard to debug, at least on the sftp side.

prime-router/src/main/kotlin/transport/RedoxTransport.kt
```diff
@@ -38,7 +40,7 @@ class RedoxTransport() : ITransport {
                 retryItems == null || RetryToken.isAllItems(retryItems) || retryItems.contains(index.toString())
             }
             .mapNotNull { (index, message) ->
-                if (!sendItem(redoxTransportType, token, message)) {
+                if (!sendItem(redoxTransportType, token, message, "$reportId-$index")) {
                     index.toString()
```
jimduff-usds, line 44: I made change to sendItem as well, I believe, so you may want to fix that.

prime-router/src/main/kotlin/transport/RedoxTransport.kt
```diff
@@ -89,9 +98,19 @@ class RedoxTransport() : ITransport {
             .responseJson()
         // TODO: store the result id when we get line level tracking
         return when (result) {
-            is Result.Success -> true
-            is Result.Failure -> false
+            is Result.Success -> {
+                logger.info("Redox call succeeded: $itemId")
+                true
```
jimduff-usds, line 103: Again, I messed in this code as well, so you'll see conflicts.   I think you should definitely log the URL its sending to; config problems will be much easier to figure out.

prime-router/src/main/kotlin/transport/RedoxTransport.kt
```diff
@@ -89,9 +98,19 @@ class RedoxTransport() : ITransport {
...
+                logger.error("Redox endpoint call failed: $itemId: ${result.error.message}")
+                false
+            }
             else -> false
         }
     }
+
+    companion object {
+        private val logger: Logger = LogManager.getLogger(RedoxTransport::class.java)
+    }
```
jimduff-usds, line 115: Not sure about the propriety of having two different Logger managers in the code, this one, and the context from Azure.  Seems like a recipe for future confusion.  ("I changed my logging level, and I'm still not seeing the  log messages!", etc)

jimduff-usds (approved):
Note my comments - I was messing around in the redox code to add logging, as well as in the sftp code, but I added the Logger in a different way than you did....

prime-router/src/main/kotlin/transport/RedoxTransport.kt
RickHawesUSDS: @jimduff-usds I followed your lead on logging to resolve the conflict, but by putting Azure Context into the transport interface making transport only work in Azure. I would like to see it to be able to work in other containers as well, like the command line. I don't see any disadvantage of using Log4J, because it works in all containers.

PR #262, opened 2021-01-08
Actually connect to DB when checking connections

Mr0grog:
👋 Howdy, this is my first PR for this project. This is a pretty small fix for an issue I noticed while first getting the application running.

If you check connections by requesting `POST /api/reports?option=CheckConnections`, the response is an error if the database is down and the application has never connected to it before, but it *succeeds* if the database has gone down *after* the application has previously connected to it.

This resolves the issue by making sure we actually try to transact with the database when checking its connection.

To reproduce:
1. Start Postgres
2. Start application
3. Request `POST /api/reports?option=CheckConnections`. You should receive a 200 response.
4. Stop Postgres
5. Request `POST /api/reports?option=CheckConnections`. Before applying this patch, you should receive a 200 response. After applying, you should receive a 500 response.

I’m not sure how to effectively write a test for this; if anyone has advice on how to do so, I’m happy to add one.

(Apologies for not submitting this as an issue first, but it seemed so small that it wouldn’t be worthwhile. Please let me know if you’d rather I do so for similar things in the future!)

## Checklist
- [x] Tested locally?
- [x] Added new dependencies? (No)
- [x] Updated the docs? (N/A)
- [ ] Added a test?
- [x] Are there potential vulnerabilities or licensing issues with any new dependencies introduced? (N/A)

### DIFF
--- a/prime-router/src/main/kotlin/azure/DatabaseAccess.kt
+++ b/prime-router/src/main/kotlin/azure/DatabaseAccess.kt
@@ -44,7 +44,7 @@ class DatabaseAccess(private val create: DSLContext) {
     data class Header(val task: Task, val sources: List<TaskSource>)
 
     fun checkConnection() {
-        create.selectFrom(TASK).where(TASK.REPORT_ID.eq(UUID.randomUUID()))
+        create.selectFrom(TASK).where(TASK.REPORT_ID.eq(UUID.randomUUID())).fetch()
     }
 
     /**


Mr0grog:
I *think* the build might be breaking because I’m an external user, don’t have access to the repo secrets, and the test DB uses secrets for its credentials: https://github.com/CDCgov/prime-data-hub/blob/dde4069b8fd8bf69896ac9f9bfab50e7672d4d6f/.github/workflows/build.yml#L18-L33

Seems like it might be reasonable to include the credentials for the test DB as cleartext/non-secrets for testing purposes (I’m assuming the test DB should not be sensitive). I’d be happy to do that in this PR or submit a second PR for that.

RickHawesUSDS:
@cglodosky Could you look at this action run? https://github.com/CDCgov/prime-data-hub/runs/1675225637?check_suite_focus=true. Could Rob's status in our project be the cause?

Mr0grog:
OK, I dug into this a little bit more, and there are a few things going on here:

1. The workflow uses secrets for the DB configuration (as I mentioned above), and secrets are scoped to the repo. That is, they’re only available to PR from branches in this repo, not other repos (even if the owner of that repo has push access here). Now that I am a member of the repo, I *can* fix this by closing this PR and submitting a new one from a branch here rather than in my fork.

2. The env vars set by the secrets aren’t actually used by the application, so they only affect the workflow step that creates the DB, and don’t affect what DB the app tries to read. We use secrets to set `POSTGRESQL_USER` and `POSTGRESQL_PWD`, but the application reads `POSTGRES_USER` and `POSTGRES_PASSWORD`: https://github.com/CDCgov/prime-data-hub/blob/3ed4ea665b16e3073a9d313ac3e744b8537be257/prime-router/src/main/kotlin/azure/DatabaseAccess.kt#L31-L33

    I’m guessing this is a mistake or typo. Should we change the workflow so it uses the same environment variable names as the application?

2. For the purposes of testing/packaging, it looks like the values of the secrets *have* to be `prime` and `changeIT!`, since these are hard-coded in two parts of `pom.xml`. Not sure if they are also hard-coded anywhere else:

    https://github.com/CDCgov/prime-data-hub/blob/3ed4ea665b16e3073a9d313ac3e744b8537be257/prime-router/pom.xml#L39-L41

    https://github.com/CDCgov/prime-data-hub/blob/3ed4ea665b16e3073a9d313ac3e744b8537be257/prime-router/pom.xml#L296-L301

    It might make sense to change `pom.xml` to use a property/variable for both of those, and make sure the workflows set it by calling:

    ```yml
    - name: Build Maven Package
      run: mvn -B clean package --file pom.xml -Dpg.user=$POSTGRES_USER -Dpg.password=$POSTGRES_PASSWORD
    ```

So, I can fix this PR in the short term by closing it and creating a new PR from a branch in this repo. That doesn’t seem like a great solution, though, since other external contributors will just run into the same problem.

I think it makes more sense to do 3 things here:

1. Don’t use secrets (but still set environment variables) for the Postgres user and password. (As far as I can tell, this is safe and reasonable to do regardless of whether the base branch is `production`, but we could also just do it for when the base branch is not `production`.)
2. Update those environment variable names to match the ones used by the application. (`POSTGRES_USER` and `POSTGRES_PASSWORD`)
3. Update the workflow file and `pom.xml` to set and use properties for the above environment variables when calling `mvn package` so those environment variables actually get used in the packaging phase if they’re set.

A simpler alternative might be to just get rid of all the environment variable and secrets usage for DB configuration in the workflow files.

Does that make sense? Should I do that here or in a separate PR?

Mr0grog:
I wound up coding up an example of the above in #271, since I think seeing the code is actually much clearer than the explanation in this case (and the code changes are not too complex).

MauriceReeves-usds:
This all makes sense to me, but I'd like to see what @cglodosky has to say before we approve. Great detective work @Mr0grog

jimduff-usds (approved):
Good catch.   And we're delighted to have you working with us!

Mr0grog:
Thanks! Glad to be helping out. 😄

Mr0grog:
Rebased this on master to get the new CI setup where builds and unit tests are skipped.

PR #271, opened 2021-01-10
Make actions work for external PRs

Mr0grog:
This is an example of a solution for the build issues in https://github.com/CDCgov/prime-data-hub/pull/262#issuecomment-757544632. It’s mainly meant to illustrate the suggested changes, but could also be merged as a complete fix.

The current GitHub Actions workflows depend on secret values that are not available to external PRs, causing them to always fail on PRs from forks. This is real barrier for non-CDC/USDS folks wanting to contribute to the project.

This change attempts to fix the issue by no longer using secrets for database configuration in workflows. These secrets weren't actually sensitive values or used outside the workflow environment to begin with, so it's fairly safe, and makes understanding configuration simpler. This also removes some duplicate database configuration and makes that configuration settable via environment variables in `pom.xml` by using properties for configuring both Flyway and JDBC.

## Checklist
- [x] Tested locally?
- [ ] Added new dependencies?
- [ ] Updated the docs?
- [ ] Added a test?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,19 +18,18 @@ jobs:
 
       - name: Set Environment Variables
         run: |
+               echo >> $GITHUB_ENV POSTGRES_USER='prime'
+               echo >> $GITHUB_ENV POSTGRES_PASSWORD='changeIT!'
+               
                if [ "$GITHUB_BASE_REF" == "production" ]
                then
                    echo "Building for the production environment."
                    echo >> $GITHUB_ENV ACR_REPO=pdhprodcontainerregistry.azurecr.io
                    echo >> $GITHUB_ENV PREFIX=pdhprod
-                   echo >> $GITHUB_ENV POSTGRESQL_USER=${{ secrets.POSTGRESQL_PROD_USER }}
-                   echo >> $GITHUB_ENV POSTGRESQL_PWD=${{ secrets.POSTGRESQL_PROD_PWD }}
                else
                    echo "Building for the test environment."
                    echo >> $GITHUB_ENV ACR_REPO=pdhtestcontainerregistry.azurecr.io
                    echo >> $GITHUB_ENV PREFIX=pdhtest
-                   echo >> $GITHUB_ENV POSTGRESQL_USER=${{ secrets.POSTGRESQL_TEST_USER }}
-                   echo >> $GITHUB_ENV POSTGRESQL_PWD=${{ secrets.POSTGRESQL_TEST_PWD }}
                fi
 
       - name: Setup PostgreSQL
@@ -38,8 +37,8 @@ jobs:
         with:
           postgresql version: 11
           postgresql db: prime_data_hub
-          postgresql user: ${{ env.POSTGRESQL_USER }}
-          postgresql password: ${{ env.POSTGRESQL_PWD }}
+          postgresql user: ${{ env.POSTGRES_USER }}
+          postgresql password: ${{ env.POSTGRES_PASSWORD }}
        
       - name: Setup JDK
         uses: actions/setup-java@v1.4.3
@@ -47,7 +46,7 @@ jobs:
            java-version: 11
       
       - name: Build Maven Package
-        run: mvn -B clean package --file pom.xml
+        run: mvn -B clean package --file pom.xml -Dpg.user=$POSTGRES_USER -Dpg.password=$POSTGRES_PASSWORD
 
       - name: Build Docker Image
         run: docker build . --file Dockerfile --tag ${{ env.ACR_REPO }}/${{ env.PREFIX }}:latest
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,6 +10,9 @@ defaults:
 
 jobs:
   build:
+    # Do not run on PRs from forks to prevent workflow abuse.
+    # NOTE: changing this condition *may* require adjusting secrets usage in
+    # the workflow steps below.
     if: github.event.pull_request.head.repo.full_name == github.repository
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -17,36 +17,35 @@ jobs:
...
                    echo >> $GITHUB_ENV ACR_REPO=pdhprodcontainerregistry.azurecr.io
                    echo >> $GITHUB_ENV PREFIX=pdhprod
-                   echo >> $GITHUB_ENV POSTGRESQL_USER=${{ secrets.POSTGRESQL_PROD_USER }}
-                   echo >> $GITHUB_ENV POSTGRESQL_PWD=${{ secrets.POSTGRESQL_PROD_PWD }}
                else
                    echo "Building for the test environment."
                    echo >> $GITHUB_ENV ACR_REPO=pdhtestcontainerregistry.azurecr.io
                    echo >> $GITHUB_ENV PREFIX=pdhtest
-                   echo >> $GITHUB_ENV POSTGRESQL_USER=${{ secrets.POSTGRESQL_TEST_USER }}
-                   echo >> $GITHUB_ENV POSTGRESQL_PWD=${{ secrets.POSTGRESQL_TEST_PWD }}
```
Mr0grog, line 32: AFAICT, the DB-related environment variables never get incorporated into the build artifact, so this seems like a safe change, and simplifies things. We could also just get rid of these environment variables entirely, and hard-code the values in the “Setup PostgreSQL” step like we do with the database name.

prime-router/pom.xml
```diff
@@ -36,9 +36,9 @@
         <!-- Define skip flags with common defaults for developers. -->
         <package.skip.azure>false</package.skip.azure>
         <package.skip.cli>true</package.skip.cli>
-        <flyway.user>prime</flyway.user>
-        <flyway.password>changeIT!</flyway.password>
-        <flyway.url>jdbc:postgresql://localhost:5432/prime_data_hub</flyway.url>
+        <pg.user>prime</pg.user>
+        <pg.password>changeIT!</pg.password>
+        <pg.url>jdbc:postgresql://localhost:5432/prime_data_hub</pg.url>
```
Mr0grog, line 41: It seemed reasonable to change these since they are no longer used for just Flyway. Nothing else that is checked in to the repo uses them, but I don’t know if this will affect other scripts or ops-related code outside this repo.

Mr0grog, line 41: It seemed reasonable to change these since they are no longer used for just Flyway. Nothing else that is checked in to the repo uses them, but I don’t know if this will affect other scripts or ops-related code outside this repo.

.github/workflows/build.yml
```diff
@@ -17,36 +17,35 @@ jobs:
...
 
       - name: Setup PostgreSQL
         uses: Harmon758/postgresql-action@0be19fa37850b22cb4c9bbf28a03abbf44abd863
         with:
           postgresql version: 11
           postgresql db: prime_data_hub
-          postgresql user: ${{ env.POSTGRESQL_USER }}
-          postgresql password: ${{ env.POSTGRESQL_PWD }}
+          postgresql user: ${{ env.POSTGRES_USER }}
+          postgresql password: ${{ env.POSTGRES_PASSWORD }}
```
MauriceReeves-usds, line 40: Great find.

MauriceReeves-usds (commented):
I think this is okay. Once @cglodosky and @jimduff-usds take a look I'd be okay merging this in.

Mr0grog:
@cglodosky @jimduff-usds any thoughts on this? It would be nice to get this or some other solution, since builds/unit tests will currently always fail on PRs from external contributors. I’m happy to make any changes you think are needed here.

prime-router/pom.xml
cglodosky: @RickHawesUSDS - do you know of any potential side effects from making this update?  I'm unaware of any.

cglodosky:
@Mr0grog - I'm going to talk more with the team about this today, but I do like your proposed changes.  As of yesterday, I have caused the build phase to be skipped when the PR is coming from an external repository.  If we decide we want to allow builds to happen from external contributors I will merge in your changes.  Thanks for your contributions!

cglodosky:
@Mr0grog - the team decided this morning that we are going to continue with bypassing builds from external repositories.  You changes look good to me, but since we are going to forgo allowing external builds they won't be necessary at this time.  However, your changes to `pom.xml` may still be something that @RickHawesUSDS will want to implement.

Mr0grog:
> I have caused the build phase to be skipped when the PR is coming from an external repository

Is that #287? I was under the impression that the context of the command is the branch & repo it’s *running in* (not where it’s coming from) so that wouldn’t skip a build. Good to know that was incorrect!

> we are going to continue with bypassing builds from external repositories.

Doesn’t that mean PRs from forks will not run unit tests? It doesn’t look like anything else runs them.

Mr0grog:
Ahhhh, nevermind on the first part, I see #276 probably didn’t work (as I expected) and you changed to a more complex check that makes sense in #287. (Side note: those identical and minimal PR names made that pretty confusing!)

Regardless, am I understanding right that this means we’re skipping unit tests?

Mr0grog:
Talked with Rick and understand the tradeoff being made with no unit tests here a bit better.

Would you like me to reinstate the secrets usage, or leave as-is? (I think leaving as-is is better so it’s easier to change the decision about which types of PRs can trigger builds if you decide to do so in the future, but this is definitely a decision for you, not me. 🙂)

prime-router/pom.xml
RickHawesUSDS: @cglodosky I agree. Developer local passwords are not secrets.

.github/workflows/build.yml
cglodosky: @Mr0grog - would you be able to revert this change?  Since we are not going to perform builds of forked PRs, we are going to keep these as secrets.

.github/workflows/build.yml
cglodosky: @Mr0grog - could you revert this as well per the comment above?

cglodosky:
@Mr0grog - sorry for the delay on this.  If you could revert the plaintext credential changes, we will keep your `pom.xml` changes.  Once you have those reverted I can approve and get this merged into `master`.  Thanks!

cglodosky (changes requested):
@Mr0grog - sorry for the delay on this. If you could revert the plaintext credential changes, we will keep your pom.xml changes. Once you have those reverted I can approve and get this merged into master. Thanks!

.github/workflows/build.yml
Mr0grog: Yep, will do.

FWIW, though, I *do* still think keeping them as plaintext would be better (as noted here: https://github.com/CDCgov/prime-data-hub/pull/271#issuecomment-759657862). These passwords are in plaintext elsewhere in the codebase, and you are already restricting who this workflow runs for via other means. You’d have to change this back to plaintext in the future if you ever loosened those restrictions (e.g. using GitHub’s repo moderation settings or creating a blocklist instead of basing it on the PR branch’s repo).

Mr0grog:
@cglodosky rebased on master and updated it to use secrets again (but kept the other changes so those secrets are actually used, and not required to be the same as the plaintext passwords hard-coded in other parts of the codebase). I also added a quick comment in the workflow to explain the build limitations and their impact on secrets (since I know if I were maintaining the ops stuff on this repo, I’d totally lose track of these issues a few months down the road).

.github/workflows/build.yml
Mr0grog: Oh, wait, I missed this comment and am a little confused. I understand reinstating secrets (https://github.com/CDCgov/prime-data-hub/pull/271#discussion_r561079751), but if we don’t keep this change (and the one to the `mvn package` line below), these secrets HAVE to be the same as the passwords that are plaintext elsewhere in the codebase (otherwise the build will fail here), which seems to defeat the purpose of having secrets. Why are we changing this?

cglodosky: Sorry, you are correct.  I missed the connection between the maven build and this.

cglodosky (approved):

cglodosky (approved):

PR #273, opened 2021-01-11
Auth Proposal

RickHawesUSDS:
This PR adds a proposal for how clients should authenticate to the data hub in the long term. This proposal would replace the function key mechanism now in place. 

## Checklist
- [ ] Tested locally?
- [ ] Added new dependencies?
- [ ] Updated the docs?
- [ ] Added a test?


MauriceReeves-usds (approved):
Sounds good to me, though is an area I still have to learn more about. But I generally prefer public key auth over other forms.

PR #284, opened 2021-01-12
Secrets Management Proposal

### LINKED ISSUE #99: Make a proposal on how to support for storing/fetching 1000s of secrets from a Vault
As a dev-ops engineer, I would like to have a vault for data-hub's secrets. As a security official, I would like this vault to be secure and auditable. 

AC
- [ ] Can hold 1000s of secrets including SFTP passwords
- [ ] Has a master encryption key that is stored in a KeyVault (or CDCs equivalent)
- [ ] Works in Azure
- [ ] Request rates and latency that match the need of the data hub
- [ ] Simple to operate in Azure
- [ ] No single failure points in Azure
- [ ] Works locally in Docker on developer laptop

ronaldheft-usds:
This PR adds a proposal for an improved secrets management architecture to address both application and client secrets (resolves #99).

I liked the structure @RickHawesUSDS created for the Auth Proposal in #273, so I followed that structure for this proposal. I also took feedback from our engineering sync on 1/11 and made some changes.

## What to Focus on When Reviewing

- Does the overall proposal make sense?
- Are there aspects of the proposal where it's not clear the work that will be done from a high level?
- Am I using the correct and consistent terminology for what we call things in the Data Hub?
- Do I adequately address the technical requirements in the proposal?
- Do I adequately address security concerns in the proposal?
- Do I adequately address auditing concerns in the proposal?
- Any thoughts on encrypting the secrets with an HSM key before storing them in Azure's Key Vault? Additional details below...

### HSM Keys in Azure

I'm having a hard time understanding if Azure's premium tier of the Key Vault is encrypting secrets using a hardware key or a software key. I've been unable to find any white papers on their Key Vault implementation and [their phrasing is confusing regarding the premium tier](https://docs.microsoft.com/en-us/azure/key-vault/general/overview):

> Azure Key Vaults may be either software-protected or, with the Azure Key Vault Premium tier, hardware-protected by hardware security modules (HSMs). Software-protected keys, secrets, and certificates are safeguarded by Azure, using industry-standard algorithms and key lengths. For situations where you require added assurance, you can import or generate keys in HSMs that never leave the HSM boundary. Azure Key Vault uses nCipher HSMs, which are Federal Information Processing Standards (FIPS) 140-2 Level 2 validated. You can use nCipher tools to move a key from your HSM to Azure Key Vault.

The phrasing does not specifically address if secrets in Key Vault use HSM or if they fall under the software-protected keys. The Azure console and APIs do not provide the option to specify a key used for secret encryption, so if they are using an HSM key for secrets, key management is abstracted away, [as the documentation makes clear](https://docs.microsoft.com/en-us/azure/key-vault/secrets/about-secrets):

> All secrets in your Key Vault are stored encrypted. This encryption is transparent, and requires no action from the user. The Azure Key Vault service encrypts your secrets when you add them, and decrypts them automatically when you read them. The encryption key is unique to each key vault.

Furthermore, there is a line in the secret management overview that basically says for added security you should do your own encryption:

> For highly sensitive data, clients should consider additional layers of protection for data. Encrypting data using a separate protection key prior to storage in Key Vault is one example.

[There is also this line on the pricing page](https://azure.microsoft.com/en-us/pricing/details/key-vault/), regarding secrets:

> Finally, Azure Key Vault is designed so that Microsoft does not see or extract your data.
>
> Secrets are data (under 10_KB) such as passwords or .PFX files that your application can store and retrieve in plaintext. The Key Vault service persists secrets encrypted using an HSM-backed key, and provides an access control layer over them.

So, what I'm thinking is happening is all secrets are encrypted using Azure's own HSM regardless of tier.

Assuming secrets are encrypted using an HSM, what I'm debating from a security standpoint is Azure's line about doing our own encrypting prior to storing them in Key Vault. If we encrypt secrets prior to storage with an HSM, how can we do that in a way that reduces or removes an attack vector?

If someone (either external or an internal actor) compromises the Azure Key Vault data store, they only have the encrypted data. They would still need the key. If they were able to further compromise the Azure system to get access to the keys, at that point, have they not fully compromised the Azure infrastructure that they would be at risk of gaining access to the HSM we used to encrypt the secrets anyway?

My thinking is the only benefit to doing our own encryption with an Azure Key Vault HSM is if we are storing secrets outside of Azure's Key Vault, such as in our database. Beyond that, if we wanted to encrypt with an HSM prior to storing in Key Vault, we would need to do so with an HSM outside the boundary of Azure, so if Azure is compromised, our HSM would not be compromised as well.

Thoughts on my understanding of Azure's HSMs?

## Next Steps Once Approved
- @ronaldheft-usds to break up this proposal into tickets
- @ronaldheft-usds to start on the local portion of secrets management, as he does not have access to the Azure environment yet


ronaldheft-usds:
Just pushed a change. Fixed some typos and added some additional details around HSM use and how we will enable auditing of secrets access.

prime-router/docs/proposals/secrets-management.md
```diff
@@ -0,0 +1,109 @@
...
+## Proposal
+
+### Tiering Secret Storage
+
+Due to multiple needs and different secrets storage strategies, our secrets storage should be broken up into two different tiers.
+
+#### Tier 1: Application Secrets
+
+These are secrets the core application needs to function: database connection strings, tokens for external core-application services like Okta, etc. These secrets should be stored in environment variables that can easily be configured on local developer machines and cloud services.
+
```
RickHawesUSDS, line 31: Agree that there are two tiers.

prime-router/docs/proposals/secrets-management.md
```diff
@@ -0,0 +1,109 @@
...
+To enable storage-agnostic secrets access, we should build a base service that can be extended depending on the storage mechanism. Method signatures for this class may look like the following:
+
+```
+interface ConnectionCredentialStorageService {
+    ConnectionCredential fetchConnectionCredential(UUID connectionId)
+    void saveConnectionCredential(UUID connectionId, ConnectionCredential credential)
+}
+```
+
```
RickHawesUSDS, line 46: I like this idea and it should allow us to have a simple (but not secure) implementation for our local development environments.

RickHawesUSDS: One question in my mind is how we can trace who and why this credential is being asked for. Should that be the responsibility of the caller or the secret manager? Should the interface require this information?

prime-router/docs/proposals/secrets-management.md
```diff
@@ -0,0 +1,109 @@
...
+```
+interface ConnectionCredential {
+}
+
+class UserPassConnectionCredential implements ConnectionCredential {
+    String username;
+    String password;
+}
+
```
RickHawesUSDS, line 59: Looks good.

prime-router/docs/proposals/secrets-management.md
```diff
@@ -0,0 +1,109 @@
...
+With a `ConnectionCredentialStorageService` in place, we can store secrets slightly differently depending on the environment.
+
+#### Storing Secrets Locally
+
+Locally, we should store secrets in a way that leverages the same access methods as we will in the cloud.
+
+For applications secrets, we should set local environment variables. Any application secrets that need to be shared between developers should be communicated via a secure method when setting up the developer's environment.
+
+For connection secrets, we should leverage Hashicorp Vault. Hashicorp Vault provides a Docker container we can configure to spin up with our existing Docker containers. By managing our local Vault instances through Docker, we can easily share configuration across developers, ensuring Vault is configured correctly locally.
+
```
RickHawesUSDS, line 77: Sounds fine.

prime-router/docs/proposals/secrets-management.md
```diff
@@ -0,0 +1,109 @@
...
+The container should be configured to use local storage, which persists across Docker environment tear downs. The local storage file will be added to our `.gitignore` to ensure the secrets database remains unique per developer. The local storage database will be encrypted using a Vault secret, so in the event a developer's local secret database is compromised, the attacker would still require the Vault secret.
+
+#### Storing Secrets in Higher Cloud Environments
+
+In the cloud, we should leverage Azure Key Vault for secrets management. Key Vault is highly available, scales, and is affordable for our needs. The underlying Key Vault service is managed by Microsoft, so there are no additional infrastructure management costs on our end.
+
```
RickHawesUSDS, line 83: Agreed.

RickHawesUSDS (approved):

RickHawesUSDS:
Thanks for this work Ron. Look's good enough to proceed.

prime-router/docs/proposals/secrets-management.md
ronaldheft-usds: For consistency, the secret manager (i.e. `ConnectionCredentialStorageService`) should require it.

I'm not sure exactly what details are pertinent to log for auditing purposes. I'm thinking at a minimum we'd want to know the called and the reason requested. Maybe a method signature like this:

```
enum CredentialRequestReason {
   SEND_BATCH,
   SEND_SINGLE,
   ...
}

ConnectionCredential fetchConnectionCredential(UUID connectionId, String callerId, CredentialRequestReason reason);
```

PR #290, opened 2021-01-13
Add a patient_id for STRAC when none exists

RickHawesUSDS:
This PR adds a patient_id element for a STRAC app result, even if one doesn't exist. 

## Background

The STRAC app does not have the concept of a patient id. Redox requires a Paitent ID (fails if it is not present). After some discussion, it was decided to use a per-test id the placer_order_id to make Redox happy. Redox will not pass this ID to PA when it generates the HL7 message, so PA will be happy. 

## Changes
- Define patient_id as the placer_order_number for Redox
- Change patient_county to the FIPS county code per PA's request
- Code clean-up on the RedoxSerializer 

## Checklist
- [x] Tested locally?
- [ ] Added new dependencies?
- [ ] Updated the docs?
- [ ] Added a test?

### DIFF
--- a/prime-router/metadata/schemas/PA/pa-covid-19-redox.schema
+++ b/prime-router/metadata/schemas/PA/pa-covid-19-redox.schema
@@ -8,3 +8,46 @@ elements:
   # override the default cardinality: The CLIA is held by the county.
   - name: testing_lab_clia
     cardinality: ZERO_OR_ONE
+
+  # PA wants a county code instead of a name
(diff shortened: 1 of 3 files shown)


jimduff-usds (approved):
In a way, the strac patient_id is that barcode they use, either on the license, or created on the phone.   Seems like they could create some sort of unique hash based on that, so that we don't have to manage PII, eg, someone's drivers lic.

PR #291, opened 2021-01-13
Sftp fix

mry-usds:
This PR fixes the STAT related bug in the SSHJ code.

## Changes
- upgraded the SSHJ library (27=>30)
- added .setPreserveAttributes(false)
- added additional info about the InMemoryFile (in particular that it is a file and not a directory)

## Checklist
- [X] Tested locally?
- [X] Added new dependencies?
- [ ] Updated the docs?
- [ ] Added a test?

## Fixes 

- #issue

## To Be Done

- #issue

### DIFF
--- a/prime-router/pom.xml
+++ b/prime-router/pom.xml
@@ -471,7 +471,7 @@
         <dependency>
             <groupId>com.hierynomus</groupId>
             <artifactId>sshj</artifactId>
-            <version>0.27.0</version>
+            <version>0.30.0</version>
         </dependency>
         <dependency>
             <groupId>org.apache.commons</groupId>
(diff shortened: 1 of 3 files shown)


RickHawesUSDS:
Could you explain what was happening to cause the exception and why this fix works?

RickHawesUSDS:
Also, we've disabled retry. Could you enable it in the send function?

mry-usds:
By default, SSHJ moves the file, then attempts to set the permissions on the file to those of the origination.  In our case, this fails as we don't have permissions to set the file attributes.  The line client.getFileTransfer().setPreserveAttributes(false) stop this attempt from happening and uses the mask on the receiving side.

RickHawesUSDS (approved):
LGTM

PR #322, opened 2021-01-15
Break CLI into multiple subcommands

Mr0grog:
The CLI had several options that were mutually exclusive, as well as options that behaved differently depending on some of those exclusive options (for example, `--output` behaved differently depending on whether `--generate_docs` was set). This breaks the CLI up into multiple subcommands, both to clarify behavior for users and to make the implementation a bit more organized.

The new commands are:
- `data` (the "main" command that most options used to apply to — **I’d love some feedback on whether this is a good name; the right answer here is less clear to me than the other two**)
- `list`
- `generate-docs`

I've also tried to rework the help text here to be clearer and more informative. Unsupported options that are placeholders for future behavior now throw errors when used rather than silently failing. I’d **love** any feedback on these docs.

Otherwise, I’ve tried not to change the various options or functionality in any big way, although I *did* make all options kebab-case so they were more consistent (some were snake_case and some were kebab-case).

Example command to generate fake data now:

```sh
$ ./prime data --input-fake 5 --input-schema fl/fl-covid-19 --output florida-data.csv
```

Docs for all commands:

```
$ ./prime
Usage: prime [OPTIONS] COMMAND [ARGS]...

  Tools and commands that support the PRIME Data Hub.

Options:
  -h, --help  Show this message and exit

Commands:
  data           process data
  list           list known schemas, clients, and services
  generate-docs  generate documentation for schemas
```

```
$ ./prime data --help
Usage: prime data [OPTIONS]

  process data

  Use this command to process data in a variety of ways, similar to how the
  REST service might handle data sent to it.

  You must always specify input data (via --input, --input-dir, --merge, or
  --input-fake to generate random data) and --input-schema.

  The data can then be transformed to a new schema (using --route, --route-to,
  or --output-schema) and written to a given location. Output will be written
  in CSV format unless otherwise specified.

  If no output location is set, files are written to the current working
  directory.

  To generate test data for later use, use --input-fake with no
  transformations. For example, to generate 5 rows of test data using the
  Florida COVID-19 schema: > prime data --input-fake 5 --input-schema
  fl/fl-covid-19 --output florida-data.csv

Options:
  --merge <paths>               list of comma-separated CSV files to merge as
                                input
  --input <path>                path to CSV file to read
  --input-fake <int>            generate N rows of random input according to
                                --input-schema. The value is the number of
                                rows to generate.
  --input-dir <dir>             path to directory of files
  --input-schema <schema_name>  interpret input according to this schema
  --validate                    Validate stream
  --send                        send output to receivers
  --route                       transform output to the schemas for each
                                receiver the input would be routed to
  --route-to <receiver>         transform output to the schema for the given
                                receiver
  --output-schema <name>        transform output to the given schema
  --output-hl7                  format output as HL7 instead of CSV
  --output <path>               write output to this file. Do not use with
                                --route, which generates multiple outputs.
  --output-dir <path>           write output files to this directory instead
                                of the working directory
  --target-state <abbrev>       when using --input-fake, fill geographic
                                fields using this state. This should be a
                                two-letter abbreviation, e.g. 'FL' for
                                Florida.
  --target-county <name>        when using --input-fake, fill county-related
                                fields with this county name.
  -h, --help                    Show this message and exit
```

```
$ ./prime list --help
Usage: prime list [OPTIONS]

  list known schemas, clients, and services

Options:
  -h, --help  Show this message and exit
```

```
$ ./prime generate-docs --help
Usage: prime generate-docs [OPTIONS]

  generate documentation for schemas

  By default, this will generate documentation for all known schemas. To
  generate docs for a particular schema, use the --input-schema option.

Options:
  --input-schema <schema_name>  schema to document
  --include-timestamps          include creation time in file names
  --output <path>               write documentation to this file (should not
                                include extension)
  --output-dir <path>           interpret `--output` relative to this
                                directory (default:
                                "docs/schema_documentation")
  -h, --help                    Show this message and exit
```

## Other Questions

This really helped me get a better grasp on how a few things work, but did leave me with some questions about things that might be worth addressing further in this or another PR:

1. The relationship between `--output` and `--output-dir` is a little complex (both in `data` and `generate-docs`)
    - In `data`, it might be good to add a warning or error if `--output` is when multiple files are being generated, or if both are set (since one will be ignored).
    - In `generate-docs`, I think it might be clearer and simpler to remove `--output-dir` and treat `--output` as a directory. It *does* remove some functionality, but I’m not sure that would really be bad.

2. `--validate` is always true and cannot be disabled, so it seems like a meaningless option.
    - Should we drop it?
    - Should it conceptually be more like `--validate-only` and stop the logic early? It *might* be better as a separate command in this case.
    - Should this actually control validation output, and cause warnings & errors not to print? In that case, if we want the current default, it might be better to make the option `--no-validate` and default to false.

3. I think it would be good to rename `--output-schema` to `--transform` or `--transform-schema` to clarify its role and the fact that it is an alternative to `--route` and `--route-to`.

4. There is logic for REDOX output, but CLI option. Should we have `--output-format csv|hl7|redox` instead of `--output-hl7`, or should I add `--output-redox`? (I think I’d vote for the former, given my understanding so far.)

5. I was also hoping to add `./prime --version`, but wasn’t sure how best to read in the current version programmatically. Would love any advice here, including if version info is not worthwhile to have. (It’s been a while since I’ve done a lot of Java dev, and don’t have a whole lot of Kotlin experience.)

## Checklist
- [x] Tested locally?
- [ ] Added new dependencies?
- [x] Updated the docs?
- [ ] Added a test?

### DIFF
--- a/prime-router/src/main/kotlin/cli/main.kt
+++ b/prime-router/src/main/kotlin/cli/main.kt
@@ -112,8 +112,7 @@ class ProcessData : CliktCommand(
         if (listOfFiles.isNullOrEmpty()) error("No files to merge.")
         val files = listOfFiles.split(",", " ").filter { ! it.isNullOrBlank() }
         if (files.isEmpty()) error("No files to merge found in comma separated list.  Need at least one file.")
-        val reports = files.map { readReportFromFile(metadata, it) }.filter { report -> report != null }
-        if (reports.isEmpty()) error("Unable to merge.  All reports failed validation")
+        val reports = files.map { readReportFromFile(metadata, it) }
         echo("Merging ${reports.size} reports.")
         return Report.merge(reports)
     }
@@ -279,7 +278,7 @@ class ListSchemas : CliktCommand(
             println(
                 formatTemplate.format(
                     it.organization.name, it.name, it.schema,
-                    it.jurisdictionalFilter.joinToString { it -> it }
+                    it.jurisdictionalFilter.joinToString()
                 )
             )
         }
--- a/prime-router/src/main/kotlin/cli/main.kt
+++ b/prime-router/src/main/kotlin/cli/main.kt
@@ -3,6 +3,7 @@
 package gov.cdc.prime.router.cli
 
 import com.github.ajalt.clikt.core.CliktCommand
+import com.github.ajalt.clikt.core.subcommands
 import com.github.ajalt.clikt.parameters.groups.mutuallyExclusiveOptions
 import com.github.ajalt.clikt.parameters.groups.single
 import com.github.ajalt.clikt.parameters.options.convert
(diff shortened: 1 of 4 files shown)


prime-router/src/main/kotlin/cli/main.kt
```diff
@@ -136,6 +167,7 @@ class RouterCli : CliktCommand(
         }
     }
 
+    // NOTE: This exists to support not-yet-implemented functionality.
```
Mr0grog, line 170: As far as I can tell, `postHttp` is supposed to support `--send`, but it’s not wired up.

prime-router/src/main/kotlin/cli/main.kt
```diff
@@ -80,8 +112,7 @@ class RouterCli : CliktCommand(
         if (listOfFiles.isNullOrEmpty()) error("No files to merge.")
         val files = listOfFiles.split(",", " ").filter { ! it.isNullOrBlank() }
         if (files.isEmpty()) error("No files to merge found in comma separated list.  Need at least one file.")
-        val reports = files.map { readReportFromFile(metadata, it) }.filter { report -> report != null }
-        if (reports.isEmpty()) error("Unable to merge.  All reports failed validation")
+        val reports = files.map { readReportFromFile(metadata, it) }
         echo("Merging ${reports.size} reports.")
```
Mr0grog, line 116: @jimduff-usds the compiler was warning that filtering and empty checking here was redundant (since `readReportFromFile()` throws if `report == null`. I did this to clean up the error, but not sure if you would have rather put a try...catch in the map block so one bad file doesn’t stop the process.

prime-router/src/main/kotlin/cli/main.kt
```diff
@@ -284,4 +250,99 @@ class RouterCli : CliktCommand(
...
+    """
+) {
+    private val inputSchema by option("--input-schema", metavar = "<schema_name>", help = "schema to document")
+    private val includeTimestamps by
+    option("--include-timestamps", help = "include creation time in file names")
+        .flag(default = false)
+    private val outputFileName by option("--output", metavar = "<path>", help = "write documentation to this file (should not include extension)")
+    private val defaultOutputDir = "docs/schema_documentation"
+    private val outputDir by option("--output-dir", metavar = "<path>", help = "interpret `--output` relative to this directory (default: \"$defaultOutputDir\")")
+        .default(defaultOutputDir)
```
Mr0grog, line 312: I could not for the life of me figure out how to make `helpTags` and `defaultForHelp` actually do their job and print the default value in the help text, so I did this. If anyone knows how to make Clikt do it, please fill me in!

MauriceReeves-usds (approved):

RickHawesUSDS (approved):
Thanks. I'm going to do the merge.

PR #320, opened 2021-01-14
Jim/lineage design

jimduff-usds:
This PR is the design proposal for the base tables tracking lineage and history of Reports and Items in the web.

Its easier to read the proposal here: https://github.com/CDCgov/prime-data-hub/tree/jim/lineage-design/prime-router/docs/proposals/lineage

## Designs

- #147 

## To Be Done

- The actual implementation of #147

PR #328, opened 2021-01-16
Create a Proposal Process read-me

RickHawesUSDS:
This PR does add a readme about the proposal process into the proposal process directory. 

## Reviewers
- Comment on the idea of asking for 2 approvals since these changes are supposed to be significant. Could be convinced to keep our 1 approver rule.


jimduff-usds (approved):

prime-router/docs/proposals/proposal-process-readme.md
```diff
@@ -0,0 +1,23 @@
...
+
+At the same time, there is a balance between the spent on paper design and writing code. There is a limit to the types of problems a paper design can flush out. We do not expect a design proposal to have all aspects of a design detailed. Every engineer should judge how much detail to put into their documents. After their pull-request is approved, the engineer should feel free to improve a design as they code.
+
+## Proposal
+
+The proposal process involves a Markdown document for the proposal, and a GitHub pull-request that will contain the discussion about the proposal. Using the project's public GitHub infrastructure allows for outside reviewers.
+
+**Steps**
+
+1. Put a proposal document in a single markdown document in the project's proposal directory. Please include sections for context and motivation behind the proposal, the goals of the proposal, and the details of the proposal itself. Put assets in a subdirectory or the assets directory.
```
Mr0grog, line 17: Having a template for this would be extremely helpful. [Rust’s RFC process](https://github.com/rust-lang/rfcs) has been the general standard I’ve seen a lot of large open-source projects base their projects on (it’s been a pretty successful model). They’ve got a [good template](https://github.com/rust-lang/rfcs/blob/master/0000-template.md), although I think their process is definitely too heavy-weight for the scale (and speed) we need to operate at here. Their template can probably offer more utility than the rest for our case here.

> Put assets in a subdirectory or the assets directory

I think you should recommend one specific structure here, otherwise people will do different things, and it will get messy. I suggest adding a directory at `docs/proposals/<proposal_name>-assets/` for assets.

prime-router/docs/proposals/proposal-process-readme.md
```diff
@@ -0,0 +1,23 @@
...
+
+2. Create a pull-request that contains instructions to the reviewers. Describe the aspects of the design that the reviewers should consider.
+
+3. Add all active contributors the area of the proposal as reviewers of the review. Aim for rough consensus in the discussion. At least two reviewers should approve.
```
Mr0grog, line 21: My experience is that using a process like this with small open-source teams can start to feel high overhead, and things can easily get lost or put off if people feel like reviewing a [potentially long] proposal is a lot of effort. Since the team here is mainly paid staff working full time, this might not be as much of an issue, though.

Setting a standard timeframe you expect at least a couple reviews to be delivered by or asking the proposer to put a timeframe on the proposal can help. It sets some general expectations and lets proposers know when it might be OK to bug people a bit more for feedback.

prime-router/docs/proposals/proposal-process-readme.md
```diff
@@ -0,0 +1,23 @@
...
+
+**Steps**
+
+1. Put a proposal document in a single markdown document in the project's proposal directory. Please include sections for context and motivation behind the proposal, the goals of the proposal, and the details of the proposal itself. Put assets in a subdirectory or the assets directory.
+
+2. Create a pull-request that contains instructions to the reviewers. Describe the aspects of the design that the reviewers should consider.
+
+3. Add all active contributors the area of the proposal as reviewers of the review. Aim for rough consensus in the discussion. At least two reviewers should approve. 
+
+Follow the normal PR process for the review, including calling meetings if needed.
```
Mr0grog, line 23: I think it would be a helpful reminder to note that, if there’s a meeting, participants should make sure someone summarizes the suggestions or results in a comment or review on the PR, so people not in the meeting can a rough sense of what’s going on with the proposal.

Mr0grog:
Re: 2+ reviewers, I don’t have a strong opinion since I don’t feel I really know what the team’s current bandwidth is. In general, I think it seems reasonable since you have a dedicated team and a major design should *usually* require more feedback.

prime-router/docs/proposals/proposal-process-readme.md
Mr0grog: Or as @jimduff-usds is doing in #320, proposals with assets should be a directory:

```
docs/
└─┬ proposals/
  ├── <proposal_with_no_assets_name>.md
  └─┬ <proposal_with_assets_name>/
    ├── README.md (proposal text)
    ├── <asset1>
    ├── <asset2>
    └── etc.
```

MauriceReeves-usds: I agree that the Rust RFC process is probably too heavyweight at this point, but we could crib a few things from it. I do especially appreciate the value of an RFC being open for ten days so everyone has a chance to look and provide feedback before it's adjudicated.

prime-router/docs/proposals/proposal-process-readme.md
MauriceReeves-usds: It would probably be good to have a dedicated meeting where RFC's are reviewed, prior to backlog grooming and sprint planning, that way if we want to pull a proposal into the next sprint it's been reviewed and approved before that.

prime-router/docs/proposals/proposal-process-readme.md
RickHawesUSDS: We are not a project that is ready for a formal RFC process. 

I like Jim's sub-folders.

Mr0grog: As another quick tidbit to steal from Rust, prefixing the proposal with a number or date (a sort-friendly one, like “2021-01-19”) would be helpful here, too. For example, we currently have a proposal named `authentication.md`, but if we made major, proposal-worthy authentication changes in the future, the prefix would be helpful in making it clear which proposal came first.

ronaldheft-usds (approved):
I think this strikes a great balance between documenting our work and getting the work done. The proposed process does not put a large burden on a developer to get a substantial change approved, while still allowing for some oversight and discussion.

prime-router/docs/proposals/0000-proposal-process-readme.md
```diff
@@ -0,0 +1,29 @@
...
+
+The proposal process involves a Markdown document for the proposal, and a GitHub pull-request that will contain the discussion about the proposal. Using the project's public GitHub infrastructure allows for outside reviewers.
+
+**Steps**
+
+1. Put a proposal document in a single markdown document in the project's proposal directory or if multiple files are needed in a subdirectory of this directory. Prefix each proposal (file or directory) with a 4-digit number to keep the proposals sorted in order. Please include sections for context and motivation behind the proposal, the goals of the proposal, and the details of the proposal itself.
+
+2. Create a pull-request that contains instructions to the reviewers. Describe the aspects of the design that the reviewers should consider.
+
+3. Conduct the review. Add all active contributors in the area of the proposal as reviewers.
```
shanewholloway-usds, line 25: Possible addition to step 3: capture relevant aspects of discussion back into the written proposal. (e.g. questions, answers, explorations, options, alternatives, and/or decisions.) Ideally the written proposal should reflect the team's understanding after the review.

Also, including a link to the PR(s) where discussion occurred would leave an nice explicit trail to follow.

shanewholloway-usds (approved):
Aiming for a lightweight and visible process feels right for now; iteration on the process is simple and open.

Issue #267, opened 2021-01-10
Add the ability to stop a retry sequence 


RickHawesUSDS:
**Is your feature request related to a problem? Please describe.**
Need to add a stop retry procedure for a task, for a destination, and all tasks

**Describe the solution you'd like**
A method only people with Azure account access can use. 

**Describe alternatives you've considered**

**Additional context**
This feature idea comes out of today's incident

jimduff-usds:
Closing as we haven't seen a need for this.  Feel free to reopen if a need arises.

PR #329, opened 2021-01-18
Simple Stop Batch or Send Support

RickHawesUSDS:
This PR adds the simple ability to stop a queued Batch or Send task by directly manipulating the DB. 

## Background

If a mistake is made in production, operators would like the ability to stop a batch or a send task in production by hand. Today this is accomplished by deleting the entire report task row in the database. 

## Operations

To stop a pending send, an operator with a DB connection can change the `next_action` value to `none`. Note: this is an improvement over deleting the row and the records associated with it.

An example query to find all tasks with pending retries.
```
SELECT report_id, next_action, created_at FROM task WHERE retry_token is not null;
```

An example query to stop pending send. 
```
UPDATE task
SET next_action = 'none'
WHERE report_id = '673bcfe9-3d4e-4aaa-97eb-a0f8e173d706';
```

## Changes
- Function checks to ensure that the queue message matched what the function expects
- Workflow check to ensure that the action in the DB matches what was received. If not, do nothing.  

## Checklist
- [x] Tested locally?
- [ ] Added new dependencies?
- [ ] Updated the docs?
- [ ] Added a test?

## Fixes 

- #267 Add the ability to stop a retry sequence

## To Be Done

This is a minimal fix for the moment. One could imagine a CLI or Web Site that allowed senders to stop their own reports. These types of fixes would need per client authentication and new end-points to implement the logic.

### DIFF
--- a/prime-router/docs/playbooks/Stop_a_batch_or_send.md
+++ b/prime-router/docs/playbooks/Stop_a_batch_or_send.md
@@ -1,0 +1,26 @@
+## Conditions
+
+There is a batch or send operation that has been queued, but we would
+like to prevent these operations from taking place. In other words, we want
+to disrupt the normal workflow. 
+
+## Prerequisites
+
+1. Connection to production database
+2. PSQL or equivalent tool that supports DB queries
+
+## Actions
+
+1. Find report id to act on. 
+
+    ```
+    SELECT report_id, next_action, created_at FROM task WHERE retry_token is not null;
+    ```
+2. Set the `next_action` column to `none`
+
+    ```
+    UPDATE task
+    SET next_action = 'none'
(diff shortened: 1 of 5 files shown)


jimduff-usds (changes requested):
Rick, we should add some info here or perhaps better in docs/  explaining how you go about stopping a batch or send step.
- what privileges you need
- how to connect to the databsae
- example query to run to find the data you want to modify
- example update to run.
- how to be careful and not damage the data.

ronaldheft-usds (commented):
The underlying code of this PR looks great. The documentation in this PR makes sense to me; maybe we can take and put it in `docs/playbooks/stop_batch_or_send.md`?

jimduff-usds (approved):

RickHawesUSDS:
@ronaldheft-usds Added a playbook folder

Issue #344, opened 2021-01-19
Add new column 'Corrected_result_ID' to simplereport schema to fully support corrections.


jimduff-usds:
SimpleReport will soon be sending us corrections.  this ticket is to be ready for that.

Every submitted result row from SimpleReport, even a correction, always has a new/unique Result_ID. 

So, we'll create a new `Corrected_result_ID` column that a Correction can use to point to the previous data being corrected.

If there are multiple corrections, the `Corrected_result_id` should be the Result_id of the most recent previous connection.  That is, the second correction points to the first correction, not back to the original result.

There are two work items for the Hub team:
- Hub team needs to add `Corrected_result_ID as an optional column.
- Hub team needs to check on whether we're using the right valueset hl70123 vs hl70085

jimduff-usds:
So, hl70123 is the "Result Status", and is required for use in an OBR hl7 line.
https://phinvads.cdc.gov/vads/ViewValueSet.action?oid=2.16.840.1.114222.4.11.815

And hl70085 is the "Observation Result Status" and is required for use in an OBX hl7 line.
See: https://phinvads.cdc.gov/vads/ViewValueSet.action?oid=2.16.840.1.114222.4.11.811

Our hl7 appears to create both OBX and OBR rows, so we probably need both.  However, we're only likely to use Final, Corrected, and Delete/Cancel, so it seems silly to ask SimpleReport to send both.    I think we can just map from one to the other.    So, I'm going to **leave it as hl70123** in the schema, and we can translate as needed to generated the OBX line.

PR #345, opened 2021-01-20
Added Corrected_result_id, per #344

jimduff-usds:
This PR added Corrected_result_ID to the columns we read from SimpleReport, in support of their sending corrections to prior tests.  (The Test_result_status field is also needed for this, but its already in the schema)

## Checklist
- [X] Tested locally?
- [ ] Added new dependencies?
- [X] Updated the docs?
- [ ] Added a test?

## Fixes 

- #344 

## To Be Done

- SimpleReport has asked for two potential follow-on features:
1. Optionally prevent redundant submissions by validating that every Result_ID is unique.  That is, validate that each Result_ID has not been previously submitted.
2. Optionally confirm that a Corrected_result_ID does indeed point to a valid previous Result_ID.

These need discussion, prioritization, and are dependent on other work first.

### DIFF
--- a/prime-router/docs/releases/next-release.md
+++ b/prime-router/docs/releases/next-release.md
@@ -1,0 +1,14 @@
+#  Hub Release Notes
+
+*NO DATE SET*
+
+## General useful links:
+
+- All Schemas are documented here:  [Link to detailed schema dictionaries](../schema_documentation)
+- The Hub API is documented here: [Hub OpenApi Spec](../openapi.yml)
+- [Click here for all release notes](../releases)
+
+## Changes to the Schemas, including for SimpleReport:
+
+Added optional field `Corrected_result_ID`, a pointer/link to the unique Result_ID of a previously submitted result.  Usually blank. Or, if an item modifies/corrects a prior item, this field holds the ResultID of the prior item.
+
--- a/prime-router/metadata/schemas/PrimeDataInput/pdi-covid-19.schema
+++ b/prime-router/metadata/schemas/PrimeDataInput/pdi-covid-19.schema
@@ -45,6 +45,9 @@ elements:
     cardinality: ONE
     csvFields: [{name: Result_ID}]
 
+  - name: previous_message_id
+    csvFields: [{name: Corrected_result_ID}]
+
   - name: placer_order_id
     mapper: use(message_id)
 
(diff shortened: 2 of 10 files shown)


prime-router/docs/releases/next-release.md
```diff
@@ -0,0 +1,14 @@
...
+
+## General useful links:
+
+- All Schemas are documented here:  [Link to detailed schema dictionaries](../schema_documentation)
+- The Hub API is documented here: [Hub OpenApi Spec](../openapi.yml)
+- [Click here for all release notes](../releases)
+
+## Changes to the Schemas, including for SimpleReport:
+
+Added optional field `Corrected_result_ID`, a pointer/link to the unique Result_ID of a previously submitted result.  Usually blank. Or, if an item modifies/corrects a prior item, this field holds the ResultID of the prior item.
```
Mr0grog, line 13: If I’m starting to understand how everything works correctly, the field is only named `Corrected_result_ID` for input from SimpleReport (i.e. the `primedatainput/pdi-covid-19` schema) and all other schemas (so, basically all receivers/services) will see this as `previous_message_id`, right?

If so, it might be worth clarifying that here; this doesn’t mention `previous_message_id`, but it’s the one that *most* people/tools interacting with DataHub will actually see.

jimduff-usds: @Mr0grog , Not 100% sure, but I think that's not quite right  - the 'previous_message_id' name is internal - its the [ineffable name that no human research can discover](https://poets.org/poem/naming-cats)

At least for CSVs, I believe each incoming and outgoing schema must specify the externally visible`name` in order to work.

Mr0grog: > each incoming and outgoing schema must specify the externally visiblename in order to work.

Ahhhhhhh, ok. So the effective change here is that we are allowing `Corrected_result_ID` in SimpleReport input, but not necessarily *sending* that field anywhere yet?

prime-router/metadata/schemas/PrimeDataInput/pdi-covid-19.schema
```diff
@@ -45,6 +45,9 @@ elements:
     cardinality: ONE
     csvFields: [{name: Result_ID}]
 
+  - name: previous_message_id
+    csvFields: [{name: Corrected_result_ID}]
+
   - name: placer_order_id
```
RickHawesUSDS, line 51: Looks good.

RickHawesUSDS (approved):
Look's good.

PR #368, opened 2021-01-22
Preparing to remove production branch

cglodosky:
This PR prepares us to move on from the `master --> production` workflow.  In summary, a test build will be triggered by an internal PR as well as a test deploy upon a merge to `master`.  Instead of a merge from `master` into `production` the publish of a release (to be versioned) targeting the `master` branch will now trigger a deployment to production.  For now, this process is manual with the potential that we can automate production deployments in the future.


.github/workflows/test_release.yml
```diff
@@ -71,19 +57,24 @@ jobs:
...
         run: |
               az functionapp config appsettings set \
               --resource-group ${{ env.RESOURCE_GROUP }} \
               --name ${{ env.PREFIX }}-functionapp \
-              --settings REDOX_SECRET=${{ env.REDOX_SECRET }} 
+              --settings REDOX_SECRET=${{ env.REDOX_SECRET }} \
+              --output none \
+              --only-show-errors
+
+      - name: Restart Azure Functions App
```
cglodosky, line 78: I moved the restart after setting env vars.  In rare cases, the env vars may not set because the restart rendered the function app unreachable.

jimduff-usds (commented):
Your overall process sounds good to me, and an improvement over the weird problems we're having now.      I presume/hope the ability to do a release is limited to just a few people?

I don't have experience with actions, so maybe ask @Mr0grog  to review?

Mr0grog:
Happy to review whenever this is ready, if helpful.

Quick thought while I’m here: it feels like there’s a fair amount of duplication across workflows, so it might be useful to set up some composite actions. For example, a `build` action could set up Postgres, set up the JDK, load/cache Maven dependencies, and run the build, and take appropriate arguments to do all that (PG user/pass, JDK version, etc.).

cglodosky:
> Happy to review whenever this is ready, if helpful.
> 
> Quick thought while I’m here: it feels like there’s a fair amount of duplication across workflows, so it might be useful to set up some composite actions. For example, a `build` action could set up Postgres, set up the JDK, load/cache Maven dependencies, and run the build, and take appropriate arguments to do all that (PG user/pass, JDK version, etc.).

@Mr0grog - I was thinking about 

> Happy to review whenever this is ready, if helpful.
> 
> Quick thought while I’m here: it feels like there’s a fair amount of duplication across workflows, so it might be useful to set up some composite actions. For example, a `build` action could set up Postgres, set up the JDK, load/cache Maven dependencies, and run the build, and take appropriate arguments to do all that (PG user/pass, JDK version, etc.).

I was thinking about the duplication of actions as well on Friday as I was moving things around.  Now that I have the PR merged in for checking for updated documentation I will take those changes into account and work on some composite actions.  I'll ping you for review once I have it ready.

Issue #147, opened 2020-12-09
Audit trail for every Report


RickHawesUSDS:
As a sender and a receiver of reports from the data hub, I would like to know where every test result has been sent to. 

**Background**
This is part of the overall history/lineage Epic #313.

This ticket sets up the tracking of Actions, Reports, and Items in reports.

There will be follow-on tickets for getting this information out to senders and receivers, which is sometimes known as the Sara Johnston feature. 

**AC**
[X] SQL tables ACTION, REPORT_FILE, REPORT_LINEAGE 
[X] Wire those into the code and test
[X] Query that shows where each report item was delivered. 
[X] Query that shows for a receiver the other receivers that received the data as well. 

Detailed requirements are here: https://github.com/CDCgov/prime-data-hub/tree/jim/lineage-design/prime-router/docs/proposals/lineage

PR #370, opened 2021-01-23
Jim / lineage phase 1

jimduff-usds:
This PR adds three new database tables:   action, history_file, and history_lineage.   It includes modifications to the ReportFunction,  BatchFunction, and SendFunction for both SftpTransport and RedoxTransport.  Since EmailTransport doesn't take any action on a report, it does not create any lineage/history.  

The old Task and TaskSource tables have not been touched, and are still fully in use.   There is no code yet that reads from the new tables, so they are harmless, and are not used.

Note: If you checkout and build jim/lineage, you will be adding new tables to your local Postgres database.  These do not appear to be breaking changes, as I've reverted back and forth multiple times, but you should be aware.  If the database does break when you try to run in an old branch, you can try wiping it out and starting over as follows:
```
dropdb prime_data_hub
createdb --owner=prime prime_data_hub
mvn clean package        <-- this will re-run flyway from scratch
```

If you check out this branch and test it, here's a little sanity test you can run:
After `mvn clean package` and starting up your local `docker-compose up`, go into Postgres and snoop around as a sanity check, like this:
```
> curl -X POST -H "client:simple_report" -H "Content-Type: text/csv"  --data-binary "@./src/test/csv_test_files/input/simplereport.csv" "http://localhost:7071/api/reports?option=SkipSend"
> psql prime_data_hub
prime_data_hub=# select * from action;
prime_data_hub=# select * from report_file;
prime_data_hub=# select * from report_lineage;
```

For each select, you can get a sense of the history data being tracked at the Report level.   Don't forget to wait 60secs for batch to run.

## Checklist
- [X] Tested locally?
- [ ] Added new dependencies?
- [X] Updated the docs?
- [X] Added a test?

## Fixes 

- This is a first step in doing #147

## To Be Done

- Track history of SendFunction also.
- Track history of items, not just reports.  (this is the main goal of this feature)
- Replace or amend old Task and TaskSource tables with this new infrastructure.   Clean up old code.  Maybe keep Task as a "todo" list, so that Action can play the role of a history table?
- Numerous todo's as decorated in the code.


prime-router/pom.xml
```diff
@@ -317,7 +317,7 @@
                             <pojos>true</pojos>
                             <pojosEqualsAndHashCode>true</pojosEqualsAndHashCode>
                             <pojosToString>true</pojosToString>
-                            <immutablePojos>true</immutablePojos>
+                            <immutablePojos>false</immutablePojos>
                         </generate>
```
jimduff-usds, line 321: You'll see in the ActionHistory container class, that I'm using the generated POJOs as convenient storage boxes.  For this to work, I needed the POJOS to be mutable.

prime-router/src/main/kotlin/Report.kt
```diff
@@ -36,6 +36,7 @@ class Report {
 
     /**
      * The sources that generated this service
+     * todo this is now redundant with ActionHistory.reportLineages.
      */
     val sources: List<Source>
```
jimduff-usds, line 41: This has not been touched, but I think it can be removed, once the full and glorious lineage tracking is ready.   (Its also buggy, because as in-memory copies of copies of reports occur, the original sources get lost.

prime-router/src/main/kotlin/Report.kt
```diff
@@ -64,6 +65,11 @@ class Report {
      */
     val name: String get() = formFileName(id, schema.baseName, destination?.format, createdDateTime)
 
+    /**
+     * A pointer to where the Report is stored.
+     */
+    var bodyURL: String = ""
+
```
jimduff-usds, line 72: The ActionHistory uses a more robust style of lineage tracking - rather than "insert as things happen", it accumulates a set of changes associated with an action, then saves them all at once.   

As a minor side effect of that significant stylistic change, it was convenient to have a report know its own blob storage url.

prime-router/src/main/kotlin/Report.kt
```diff
@@ -259,6 +265,10 @@ class Report {
         return StringColumn.create(name, List(itemCount) { "" })
     }
 
+    fun getBodyFormat(): OrganizationService.Format {
+        return this.destination?.format ?: OrganizationService.Format.CSV
+    }
+
```
jimduff-usds, line 271: Again, this function was in a different place - it was convenient and made sense for a report to know its own format.

prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -0,0 +1,336 @@
+package gov.cdc.prime.router.azure
```
jimduff-usds, line 1: This is the main new class supporting lineage tracking.    Documentation in the code, I hope.

prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -0,0 +1,336 @@
...
+     * Note:  There is a strong OO argument that this list should be broken out into each individual child Report.kt.
+     * (That is, every report should know its own parents!)
+     * However, its here because there are Functions that do not create Report.kt objects.  For example, Send.
+     * In addition, in-memory, reports get copied many times, with lots of parent-child relationships
+     * that are error-prone to track.  Hiding the lineage data here helps ensure correctness and hide complexity.
+     *
+     * todo Note that this does not work for command line.   Is that a problem?
+     * todo this is redundant with `Report.sources`.   Merge these together; eliminate one of them.
+     *
+     */
```
jimduff-usds, line 73: Each time some new morsel of historical information is created, call one of the "track" functions to store it here.

Its like the Presidential Records Act, but in Kotlin.

prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -0,0 +1,336 @@
...
+    constructor(actionStr: String, context: ExecutionContext? = null) {
+        try {
+            action.actionName = TaskAction.valueOf(actionStr)
+            this.context = context
+        } catch (e: IllegalArgumentException) {
+            error("Unknown action $actionStr")
+        }
+    }
+
+    fun trackActionParams(request: HttpRequestMessage<String?>) {
```
jimduff-usds, line 85: "Params" are incoming parameters to an action.   In the case of this function, incoming from SimpleReport or other external clients.

Storing them as json.

prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -0,0 +1,336 @@
...
+
+    fun trackActionResult(actionResult: String) {
+        // TODO should be able to get this max size from the jooq-generated code.
+        // chop the string down to 2048 if needed.
+        action.actionResult = actionResult.chunked(2048)[0]
+    }
+
+    fun trackActionResult(httpResponseMessage: HttpResponseMessage) {
+        trackActionResult(httpResponseMessage.status.toString() + ":\n" + httpResponseMessage.body.toString())
+    }
```
jimduff-usds, line 118: Finally, we are now saving the same json that we send back to our client!

prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -0,0 +1,336 @@
...
+
+    /**
+     * No report can be tracked twice, either as an input or output.   Prevents loops and other shenanigans.
+     */
+    private fun reportAlreadyTracked(id: UUID): Boolean {
+        return reportsReceived.containsKey(id) ||
+            reportsIn.containsKey(id) ||
+            reportsOut.containsKey(id)
+    }
```
jimduff-usds, line 127: A nice sanity check called every time a new report is tracked by this ActionHistory obj.

prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -0,0 +1,336 @@
...
+
+    /**
+     * Save the history about this action and related reports
+     */
+    fun saveToDb(db: DatabaseAccess, txn: Configuration? = null) {
+        if (txn != null) {
+            insertAll(txn)
+        } else {
+            db.transact { innerTxn -> insertAll(innerTxn) }
+        }
```
jimduff-usds, line 209: My vision is that `DatabaseAccess` will become very tiny - it will hold the generic stuff only.   So a second pass will go thru and simplify it.

All other table-specific code will go in other classes, like ActionHistory.  Sorta like DAOs but not that formal or fancy.

prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -0,0 +1,336 @@
...
+    }
+
+    private fun insertAll(txn: Configuration) {
```
jimduff-usds, line 212: This saves entries into the ACTION, REPORT_FILE, and REPORT_LINEAGE tables now.    ITEM and ITEM_LINEAGE are still to come.

prime-router/src/main/kotlin/azure/DatabaseAccess.kt
```diff
@@ -275,7 +276,7 @@ class DatabaseAccess(private val create: DSLContext) {
          * That is functions amortize startup costs by reusing an existing process for a function invocation.
          * Hence, a connection pool is a win in latency after the first initialization.
          */
-        private val hikariDataSource: HikariDataSource by lazy {
+        val hikariDataSource: HikariDataSource by lazy {
             DriverManager.registerDriver(Driver())
```
jimduff-usds, line 280: Hmm.  I think this is a mistake.

prime-router/src/main/kotlin/azure/Event.kt
```diff
@@ -16,6 +16,7 @@ abstract class Event(val action: Action, val at: OffsetDateTime?) {
     abstract fun toQueueMessage(): String
 
     enum class Action {
+        RECEIVE,
         TRANSLATE, // Deprecated
```
jimduff-usds, line 20: The rule is that every Report is created by an Action, even the initially submitted report.   So I had to create an 'receive' action to cover that situation.

prime-router/src/main/kotlin/azure/ReportFunction.kt
```diff
@@ -63,24 +63,33 @@ class ReportFunction {
...
+                    actionHistory.trackExternalIncomingReport(validatedRequest)
+                    createdResponse(request, validatedRequest, responseBody)
                 }
             }
-            context.logger.info("Successfully reported: ${validatedRequest.report!!.id}.")
-            workflowEngine.receiveReport(validatedRequest.report)
-            val destinations = mutableListOf<String>()
-            routeReport(context, workflowEngine, validatedRequest, destinations)
-            return createdResponse(request, validatedRequest, destinations)
+            actionHistory.trackActionResult(httpResponseMessage)
```
jimduff-usds, line 90: The above messy code reshuffling  is to set up for this call to trackActionResults - now tracking not just for submissions, but now _every_ call to reports/api endpoint that passes authentication will get logged in the database.

prime-router/src/main/kotlin/azure/ReportFunction.kt
```diff
@@ -304,11 +316,11 @@ class ReportFunction {
     private fun createdResponse(
         request: HttpRequestMessage<String?>,
         validatedRequest: ValidatedRequest,
-        destinations: List<String>
+        responseBody: String,
```
jimduff-usds, line 319: I refactored the creation of the body up one level, so the ActionHistory could get to it.

prime-router/src/main/kotlin/azure/WorkflowEngine.kt
```diff
@@ -158,6 +160,10 @@ class WorkflowEngine(
         return blob.downloadBlob(header.task.bodyUrl)
     }
 
+    fun recordAction(actionHistory: ActionHistory) {
+        actionHistory.saveToDb(db)
+    }
+
```
jimduff-usds, line 166: By passing this through WorkflowEngine, it keeps the database access stuff hidden from the Function level.

NOTE:  Right now, the Action History stuff is being committed in a separate transaction from the original data, blob creation etc.   That needs to be cleaned up.

prime-router/src/main/resources/db/migration/V3__new_lineage_tables.sql
```diff
@@ -0,0 +1,73 @@
...
+    next_action_at TIMESTAMP WITH TIME ZONE,
+
+    -- These are non-null only for 'receive' actions
+    sending_org VARCHAR(63),        -- should be a ref to an org table someday
+    sending_org_client VARCHAR(63), -- OrganizationClient
+
+    -- These are non-null only for 'send' actions:
+    receiving_org VARCHAR(63),      -- should be a ref to an org table someday
+    receiving_org_svc VARCHAR(63),  -- OrganizationService
+
```
jimduff-usds, line 53: Note:  I took out the JSON `result` and `parameters` columns  from this table - they are now in the ACTION table.    I waffled over this for a long time, as they properly normalize to the ACTION.  It was only a convenience to have them in REPORT_FILE, and it was very redundant.  This is better, even though it will cause some extra sql joins.

prime-router/src/main/resources/db/migration/V3__new_lineage_tables.sql
```diff
@@ -0,0 +1,73 @@
...
+    schema_name VARCHAR(63) NOT NULL,   -- should be a fk someday
+    schema_topic VARCHAR(63) NOT NULL,
+    body_url VARCHAR(2048),
+    body_format VARCHAR(63) NOT NULL,
+    blob_digest bytea,
```
jimduff-usds, line 58: blob_digest is not impl'ed yet.

prime-router/src/test/kotlin/azure/ActionHistoryTests.kt
```diff
@@ -0,0 +1,150 @@
...
+        assertEquals(reportFile.itemCount, 0)
+
+        // not allowed to track the same report twice.
+        assertFails { actionHistory1.trackCreatedReport(event1, report1, orgSvc, valReq1) }
+    }
+
+    /**
+     * todo This is a very weak test.   What I'd really like to do is confirm that two sql inserts were generated,
+     * one to insert into ACTION and one to insert into REPORT_FILE.
+     */
```
jimduff-usds, line 124: I need some help on writing this test.

prime-router/src/main/kotlin/azure/BatchFunction.kt
```diff
@@ -27,7 +27,7 @@ class BatchFunction {
             context.logger.info("Batch message: $message")
             val workflowEngine = WorkflowEngine()
             val event = Event.parseQueueMessage(message) as ReceiverEvent
-            if (event.action != Event.Action.BATCH) {
+            if (event.eventAction != Event.EventAction.BATCH) {
```
jimduff-usds, line 30: I changed name of Event.Action to Event.EventAction, because the name 'Action' was confusingly overloaded with the database table called ACTION.    This resulted in a zillion little changes throughout the code, as you'll see as you review.

prime-router/src/test/kotlin/azure/ActionHistoryTests.kt
```diff
@@ -0,0 +1,146 @@
...
+        assertNull(reportFile.sendingOrg)
+        assertEquals(reportFile.itemCount, 0)
+
+        // not allowed to track the same report twice.
+        assertFails { actionHistory1.trackCreatedReport(event1, report1, orgSvc) }
+    }
+
+    /**
+     * todo Figure out how to make this test work.
+     * What I'd really like to do is confirm that two sql inserts were generated,
```
jimduff-usds, line 117: I need some help with this.  Unable to figure out how to create a mock transaction.  Further, unable to figure out how to make this test anything useful.

prime-router/src/main/kotlin/azure/SendFunction.kt
```diff
@@ -50,24 +55,43 @@ class SendFunction(private val workflowEngine: WorkflowEngine = WorkflowEngine()
...
+                                        service,
+                                        transport,
+                                        content,
+                                        inputReportId,
+                                        sentReportId,
+                                        retryItems,
+                                        context,
+                                        actionHistory
+                                    )
                             }
```
jimduff-usds, line 73: These send methods are getting ungainly.

prime-router/src/main/kotlin/azure/SendFunction.kt
```diff
@@ -50,24 +55,43 @@ class SendFunction(private val workflowEngine: WorkflowEngine = WorkflowEngine()
                     .filterIndexed { i, _ -> retryToken == null || retryToken.transports.find { it.index == i } != null }
                     .forEachIndexed { i, transport ->
                         val retryItems = retryToken?.transports?.find { it.index == i }?.items
+                        val sentReportId = UUID.randomUUID() // each sent report gets its own UUID
```
jimduff-usds, line 58: This is a significant change.    The act of transmitting a file is now tracked as a parent-child relationship, just like any other.   Here, the parent is the blob stored in the Hub, the `inputReportId` to the `send` function.   And the child gets a new UUID here, the `sendReportId`, which will be used as the PK of a report_file entry representing the file as transmitted.   Note that this lets us track both failed and successful transmissions.

prime-router/docs/proposals/0003-lineage/cool-lineage-example.sql
```diff
@@ -0,0 +1,54 @@
+-- Steps to use this as a demo (a hack, I know)
```
jimduff-usds, line 1: I wrote some Postgres stored procedures to show how to meet the key lineage use cases as outlined in the proposal - they are in the next file, lineage-report-stored-procedures.sql 

This example file here, shows actual calls to those stored procs.

prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -0,0 +1,331 @@
...
+        trackSentReport(service, sentReportId, null, params, msg, 0)
+    }
+
+    /**
+     * Save the history about this action and related reports
+     */
+    fun saveToDb(txn: Configuration) {
+        insertAll(txn)
+    }
+
```
RickHawesUSDS, line 263: I'm wondering if a fluent style may be a better choice for this object. Create a builder and then load to db.

prime-router/src/main/kotlin/azure/BatchFunction.kt
RickHawesUSDS: Event.EventAction doesn't read well to me. No suggestions, something I'm thinking about.

RickHawesUSDS (approved):
Pulled the PR and tried it out. Great to see the work. 

Keep at it.

prime-router/src/main/kotlin/azure/BatchFunction.kt
jimduff-usds: @RickHawesUSDS, can we get rid of EventAction completely, and just use the jooq-generated TaskAction?

prime-router/src/main/kotlin/Report.kt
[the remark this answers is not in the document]
jimduff-usds: Yes, I think we can remove the task_source table.    Debatable whether to remove Task or not.

PR #372, opened 2021-01-23
Add a PR check to verify that docs are up-to-date

Mr0grog:
After seeing #344, I noticed that, even though we have some autogenerated docs in the repo, they are not always kept up-to-date. This adds an action to the build workflow that fails if there are uncommitted docs updates in a PR.

It will add a check like:

Screen Shot 2021-01-23 at 2 49 51 PM

And post comments to the PR like:

Screen Shot 2021-01-23 at 2 49 33 PM

The comments include a list of files it expected to see changed (basically the output of `git status docs`), since displaying a full diff seemed excessive. To see a full diff, you can click the “details” link on the status check to look at the workflow’s logs:

Screen Shot 2021-01-23 at 2 55 03 PM

**This is focused on a pass/fail approach, but we could do something different and auto-commit the generated docs.** @jimduff-usds suggested this and I tried it out, but committing to the PR branch is not so great: either you re-run the build a second time, or you don’t run the workflow again and the build/unit test status checks disappear from the PR (since the latest commit didn’t run any jobs/checks). If we wanted to auto-commit the docs, this might be better to run after pushes to the `master` or `production` branches (e.g. add it to the `release` workflow).

If helpful, auto-committing the changes instead of commenting would mean replacing the “Add diff as PR comment” step with this (tested):

```yaml
      - name: Commit Changes
        if: ${{ steps.check_changes.outcome == 'failure' }}
        run: |
          # Alter these however you want so commits are attributed in a useful way.
          git config user.name 'GitHub Actions Bot'
          git config user.email 'github.actions.bot@example.com'

          git add docs
          git commit -m 'Generate updated docs'
          git push
```

### Other Changes

While I was at it, I added dependency caching to both workflows to speed up builds a little bit.

This also includes some docs updates! This PR’s new checks would fail for the PR if I didn’t, and illustrate exactly why it’s useful. 😉

## Checklist
- [x] Tested locally?
- [ ] Added new dependencies?
- [x] Updated the docs?
- [ ] Added a test?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,6 +3,8 @@ name: Prime Data Hub
 on:
   pull_request:
     branches: [ master, production ]
+    paths-ignore:
+      - 'prime-router/docs/**'
 
 defaults:
   run:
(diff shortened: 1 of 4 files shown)


cglodosky:
@Mr0grog - I'm adding @mauricereeves onto this review as well for his input.  I like what you have, but I believe that Maurice has a very lightweight bash script that generates the documentation.  Perhaps it would be cleaner to just run that as part of the build to ensure that the documentation is always up to date?  Then we can avoid having to do it manually.

MauriceReeves-usds (approved):

Mr0grog:
> Maurice has a very lightweight bash script that generates the documentation.

If that's the case, we should also probably get rid of the `prime generate-docs` command and its supporting code, right? There’s not much value in maintaining two systems to do the same thing and having to keep them in sync. Is having it in the `prime` CLI supporting some other use?

> Perhaps it would be cleaner to just run that as part of the build to ensure that the documentation is always up to date?

I don't think that solves the problem. Someone may still forget to actually build locally (not common, but definitely happens for low-stakes changes) or forget to commit the files (very common) or forget to *remove* docs for a schema that was removed (because `prime generate-docs` doesn't do that for you).

MauriceReeves-usds:
@Mr0grog - the bash script @cglodosky mentions was just calling the generate documentation options in the CLI, but I don't see the bash script now. It must have been removed. I would prefer we keep the generate documentation logic in the CLI, and extend it to generate the MD files as needed.

Mr0grog:
> I would prefer we keep the generate documentation logic in the CLI, and extend it to generate the MD files as needed.

@mauricereeves @cglodosky I think maybe there’s a lot of talking past each other, then. The CLI generates all the MD files as appropriate, and is what’s being used here. The meat is just this step (the rest is either setting up the environment or formatting & posting a comment to the PR):

https://github.com/CDCgov/prime-data-hub/blob/53509dcf7ec4bb826a87f96cb2f5d3041bf8aed2/.github/workflows/build.yml#L95-L99

So I’m not sure anything needs to change?

cglodosky:
> > I would prefer we keep the generate documentation logic in the CLI, and extend it to generate the MD files as needed.
> 
> @mauricereeves @cglodosky I think maybe there’s a lot of talking past each other, then. The CLI generates all the MD files as appropriate, and is what’s being used here. The meat is just this step (the rest is either setting up the environment or formatting & posting a comment to the PR):
> 
> https://github.com/CDCgov/prime-data-hub/blob/53509dcf7ec4bb826a87f96cb2f5d3041bf8aed2/.github/workflows/build.yml#L95-L99
> 
> So I’m not sure anything needs to change?

I missed the `./prime generate-docs` line when I was reading through it.  All of the green made it blend in too easily.  These changes look fine to me, but I'm wondering if we want to auto-commit the updated docs - could we move this job upstream from the build/unit test job?  This way if the check for updated documents fails we can commit the new ones and skip any downstream build/unit tests.  Then, when the commit happens it will trigger the workflow again, but with a successful documentation check and then a build/unit test job can run.

Mr0grog:
> could we move this job upstream from the build/unit test job?

No. The docs are generated by the prime CLI, so it has to be built before it can generate docs.

See this discussion of auto-committing from the original post:

> This is focused on a pass/fail approach, but we could do something different and auto-commit the generated docs. @jimduff-usds suggested this and I tried it out, but committing to the PR branch is not so great: either you re-run the build a second time, or you don’t run the workflow again and the build/unit test status checks disappear from the PR (since the latest commit didn’t run any jobs/checks). If we wanted to auto-commit the docs, this might be better to run after pushes to the master or production branches (e.g. add it to the release workflow).
>
> [example workflow YAML code snipped]

Mr0grog:
Another issue I didn’t note in the OP about auto-committing on the PR (as opposed to on master/production) is that it makes things a little tougher when the person who wrote the PR needs to make additional changes — GitHub’s version of their branch will have diverged from their local.

MauriceReeves-usds:
Just got caught up again. This looks fine to me.   I missed the call to generate the docs as well.

Mr0grog:
FWIW, I tried to break up the commits in a readable way, so that might help. The most salient one to review is probably [`635b314` (#372)](https://github.com/CDCgov/prime-data-hub/pull/372/commits/635b31477b36968736baa19bfabd73d41c21bd74), which is the actual job.

cglodosky:
> Another issue I didn’t note in the OP about auto-committing on the PR (as opposed to on master/production) is that it makes things a little tougher when the person who wrote the PR needs to make additional changes — GitHub’s version of their branch will have diverged from their local.

It makes sense to me to merge this in as is for now then.  If we ever want to circle back on it we can, but I'd rather get the documentation check in there sooner rather than later.  Having this merged in will also allow me to account for the changes when moving away from the `production` branch as we discussed this morning.

PR #373, opened 2021-01-24
Settings API Part 1 - TranslationConfiguration 

RickHawesUSDS:
This PR is the first in a series of PRs to add a settings API for per organization, receiver, and sender configuration. It focuses on the basic structures. The main improvement are new HL7 and Redox standard templates that do not require custom schemas. These templates where used on TX, AZ and CO. Other states still use a custom translation. 

## Changes
- Settings API proposal
- TranslationConfiguration object to capture standard HL7 and Redox configurations
- Renaming of OrganizationService -> Receiver, OrganizationClient -> Sender, Batch -> Timing, schema -> schemaName
- Prepared for Organization to be hosted in a DB. Created a shallow, deep and JSON versions of the structure. 

## Checklist
- [x] Tested locally?
- [x] QuickTest
- [x] Download site ran locally?
- [x] Added new dependencies?
- [x] Updated the docs?
- [x] Added a test?

## To Be Done
- The jurisdicational filter settings will need to be remapped to API JSON. Left to another PR.


shanewholloway-usds:
## Context

The API is a fairly direct modeling of the `prime-router/metadata/organizations.yml` file.  The editors of these Yaml settings have a privileged admin/developer role, counterbalanced by the audit trail provided by Git source code control.

Transitioning these settings to be database driven enable more flexible administration.  Providing API endpoints enable this flexibility while safeguarding the integrity of the database.  However, careful consideration of the data security semantics and implications is warranted.  (See the [confused deputy problem][wiki_deputy])

 [wiki_deputy]: https://en.wikipedia.org/wiki/Confused_deputy_problem "A confused deputy is a legitimate, more privileged computer program that is tricked by another program into misusing its authority on the system. It is a specific type of privilege escalation." 


## My review thoughts


### Query APIs

- As a user, how do I get my relevant organization settings without knowing the `oranizationName` id?
  Using the `GET /settings` seems like a reasonable place to start; however, that route is currently dedicated to (normal) REST collection semantics for the Admin role.

- What is the use case for the collection APIs: `GET /settings/senders` and `GET /settings/receivers`?
  The data model suggests these would be collections under `/settings/organization/{organizationName}`
    - e.g. `/settings/organization/{organizationName}/senders/{senderName}`
    - e.g. `/settings/organization/{organizationName}/receivers/{receiverName}`


### Implementation details

- `{organizationName}` seems like it should be `organizationId` from the description. 
  - Pros for name:
    - Human-friendly names make debugging more pleasant.
  - Pros for identifier:
    - Human-friendly names have a tendency to change, albeit infrequently.
    - It looks like a foreign key relation; changing it may lead to data integrity issues.
    - Easily guessed names can make security attacks more tractable; hard to guess ids defend against this.

- `{senderName}` and `{receiverName}` seem better as an unique identifiers
  - If the `sender` and `receiver` records are managed in database tables with unique ids, we need to consider and ensure data access security semantics.
  - If the `sender` and `receiver` records are managed under the `{organizationName}` key, the unique identifiers can be much simpler. (and shorter)


### Security Semantics

Specific concerns:

- Access to `GET /settings/receivers/{receiverName}`
    - Inappropriate disclosure of transport, connection, and other details.
      (e.g. dns names, IPs, port numbers, Redox Ids)

- Access to `PUT /settings/receivers/{receiverName}`
    - Changing `jurisdictionalFilter` in ways that:
        - break the filter
        - inappropriately escalate data access
    - Inappropriately changing `transforms.deidentify` to `true`, potentially violating jurisdictional rules.

- Access to `GET /settings/senders/{senderName}`
    - Disclosure of sender update frequency and other details.

- Access to `PUT /settings/senders/{senderName}`
    - Breaking the submission pipeline by changing/misconfiguring
      settings like `format`, `schema`, or `isActive`

- Access to `GET /settings/organizations/{organizationName}`
    - Possible disclosure of enrollment of an organization due to guessable `{organizationName}`.

- Access to `PUT /settings/organizations/{organizationName}`
    - Changing `description` to impersonate or otherwise interfere with another organization.


## Alternatives & Discussion

I suggest we discuss and consider separate APIs for the materially different use cases.

Specifically the API for administration can use the collection APIs, largely matching what is in this first iteration of the proposal. The use case of the Admin role is roughly analogous to how the `prime-router/metadata/organizations.yml` was started. Additionally, access to this route could be more restrictively locked down (e.g. in the Azure firewall or by IP subnet) for defense-in-depth.

Secondly, a focused direct API _"useful for a settings page for the PHD customer"_. As the data security semantics are significantly different than the Admin role, I suggest the clarity of purpose and scope are advantageous.  (Both writing and reviewing the implementation for correctness should be easier than if it were embedded in the general Admin role of the collection API.) As data security issues tend to be subtle and are some of the most critical, I believe the explicitness is warranted.

Does the admin role line up well with _"useful for on-boarding states by the Data Hub team"_?  I think it can; if a more specific API workflow is desired, we can add it later in our backlog.


##### Real discussion invited!

For example, one downside is the cost of implementing, monitoring, and maintaining additional APIs.  Another is [TOOWTDI](https://wiki.python.org/moin/TOOWTDI) / [DRY](https://en.wikipedia.org/wiki/Don't_repeat_yourself) argument.  What else should we discuss?

prime-router/src/main/kotlin/azure/ReportFunction.kt
```diff
@@ -217,64 +218,70 @@ class ReportFunction {
...
         validatedRequest: ValidatedRequest,
         destinations: MutableList<String>,
         actionHistory: ActionHistory,
         txn: DataAccessTransaction
     ) {
-        val serviceDescription = if (service.organization.services.size > 1)
-            "${service.organization.description} (${service.name})"
-        else
-            service.organization.description
+        val receiverDescription = "${organization.description} (${receiver.name})"
```
RickHawesUSDS, line 250: Counting the number of receivers associated with an organization is another query when a DB is the backing store. Didn't seem worth it here.

prime-router/src/main/kotlin/TranslatorConfiguration.kt
```diff
@@ -0,0 +1,104 @@
+package gov.cdc.prime.router
+
+import com.fasterxml.jackson.annotation.JsonCreator
```
jimduff-usds, line 3: I'm puzzled as to why we had to have a new concept of "Configuration" (or Template, as you call it in your proposal).   Seems like we could just allow all schemas to be parameterized with values found in the organizations data, then just have a redox_covid_standard.schema and hl7_covid_standard.schema.

The references to `covid` embedded in the code below are a sign that this is still kinda custom code.

jimduff-usds (changes requested):
This is a huge step in the right direction of separating our core functional code from configuration - excellent!   I'm suggesting a few changes, and have a few key comments.

1.  See my comments on V5__*.sql, for some changes.

2.  You'll want to merge in from master a fix that Mo did for  ./quick-test.sh to pass.   It fixes this exception:
```
Generate fake ND data, HL7!
Exception in thread "main" java.lang.IllegalStateException: Invalid element type for sending_facility
```

3.  Very minor: Might want to do a global search for Sender and Receiver.  For example, I note that wording is still used in the response to
    `./prime --list`

# Testing Done on this branch

- I tested reports generated for PA, TX, FL, AZ, Pima, CO, in Redox, HL7, CSV formats.
- I tested Retry failures, and recovery from failures for both sftp and redox. 
- In the case of Redox, I tested partial failures.   
- Tested a file of size 1 row up to 1000 rows.  Tested merging of multiple submissions.

All tests passed, except the simple-test.sh mentioned above.

# Other comments

Like last time, we're going to see lots of conflicts when we merge this PR and my item-lineage PR.   I'm happy to "go second" and deal with that.

Unrelated question:  Is there a reason the `transport:` section has disappeared for many sites in organizations-local.yml.

prime-router/src/main/resources/db/migration/V5__Create_setting_tables.sql
```diff
@@ -0,0 +1,50 @@
...
+    values JSON,
+
+    -- Metadata
+    version INT,
+    created_at TIMESTAMP WITH TIME ZONE,
+    created_by VARCHAR(63)
+);
+
+CREATE TABLE setting_history (
+    -- Key
```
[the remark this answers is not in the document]
RickHawesUSDS: Good suggestion.

RickHawesUSDS:
Closing to reopen a full PR.

PR #374, opened 2021-01-25
North Dakota Onboarding

MauriceReeves-usds:
This PR adds the schema for North Dakota and some valuesets that they are using.

## Changes
- Adds schemas and valuesets

## Checklist
- [ ] Tested locally?
- [ ] Added new dependencies?
- [ ] Updated the docs?
- [ ] Added a test?


prime-router/metadata/schemas/covid-19.schema
```diff
@@ -180,6 +194,13 @@ elements:
     hl7Field: AOE
     hl7AOEQuestion: 65222-2
 
+  - name: observation_result_status
+    type: CODE
```
jimduff-usds, line 198: I presume you are aware we already have two status fields, test_result_status and order_result_status.  
Two notes: 
1.  SimpleReport only uses test_result_status, and will soon be using it to mark data that is a correction of prior results.    We had some discussion about whether hl70123 or hl70085 was the correct value set to use.  They seemed absurdly similar, so I decided we could just map from one to the other in the Hub, as needed.

@MauriceReeves-usds , can you doublecheck my comment here: https://github.com/CDCgov/prime-data-hub/issues/344#issuecomment-763143528

to see if this makes you want to change over to hl70123?  Or, fell free to change test_result_status over to hl70085 if that's the right answer.

The current plan is for SimpleReport to only use F for Final, C for Corrected, and X to cancel a test result.  (See slack discussion here:  https://usds.slack.com/archives/C01H78NGM45/p1611087869002300

2.  Just for your awareness, Redox in all its glory has **three** _status fields:

```
Orders[].ResultStatus
Orders[].Status
Orders[].Results[].Status
```
each of which has slightly different sets of allowed values.  You are in a maze of little twisty passages, all alike....

prime-router/metadata/schemas/covid-19.schema
```diff
@@ -780,6 +804,28 @@ elements:
     documentation: |
       The date which the specimen was collected. The default format is yyyyMMddHHmmsszz
 
+  - name: specimen_collection_method
```
jimduff-usds, line 807: For this and a few other similar fields being added here and used by ND, I'm wondering how will we generate these from SimpleReport data?

SimpleReport is currently sending us these vaguely similar fields:
```
specimen_source_site_code
specimen_type
equipment_model_name
test_performed_code
```
Is it possible to use those instead?  If not, then we need to have a way to map them.

prime-router/metadata/schemas/covid-19.schema
```diff
@@ -767,9 +788,12 @@ elements:
     natFlatFileField:
     hl7Field: MSH-3
 
-  - name: specimen_collection_site # collection location
-    type: TEXT # CWE without a defined table
-    hl7Field: SPM-10
+  - name: sending_facility
+    type: HD
+    # Default is the official name of PRIME from CDC PHIN. Matches the OID in sending_application_id, below.
```
jimduff-usds, line 793: Yes, this seems redundant with sending_application.   Should we just use a `mapper` to generate this?

jimduff-usds (approved):
Mo -  There's concerns that you probably already thought about, but I added a bunch of questions/comments just in case.   Mainly, I'm not clear on how SimpleReport will generate some of the fields ND wants.

If needed, we can ask SR to make changes, but it would be on their timeline - we could generate data in the hub in the meantime.  Again, if needed.

prime-router/metadata/schemas/covid-19.schema
MauriceReeves-usds: I wasn't sure if this was ever going to change. For example, if we batch up results from a single facility we might want to fill in this field.

prime-router/metadata/schemas/covid-19.schema
MauriceReeves-usds: Those don't really appear to map to specimen_collection_method, but I'll look and see if we can map somehow.

prime-router/metadata/schemas/covid-19.schema
MauriceReeves-usds: In some instances hl70123 and hl70085 do seem similar but hl70123 is used in the OBR section, and hl70085 is used in the OBX section. Some states are requesting we send them both OBR and OBX, for example North Dakota. I'm going to send this over to SimpleReport and ask them to review.

jimduff-usds: @mauricereeves , I think they are so similar that we should just generate one from the other in the Hub.

prime-router/metadata/schemas/covid-19.schema
jimduff-usds: They all seem to be various descriptions of nasal passages to me.  Very appealing to the middle-schooler in me.

prime-router/metadata/schemas/covid-19.schema
jimduff-usds: @mauricereeves, check with @RickHawesUSDS .

PR #325, opened 2021-01-15
Secrets Management: Local Hashicorp Vault Docker container

ronaldheft-usds:
This PR begins work on Secrets Management (#190) by adding a local Hashicorp Vault (#292). This PR only addresses standing up a local Hashicorp Vault instance; at this time the Vault is not used for anything, but it will in follow-on PRs.

## Changes
- Adds the official Hashicorp Vault Docker image to our `docker-compose.yml`
- Creates a persistent Docker volume to store the encrypted local Vault
  - This will allow the secrets to persist locally across `docker-compose up` and `docker-compose down` actions
- Configures Vault to use the file storage engine, leveraging the persistent Docker volume
- Adds a custom Vault startup script to automatically initialize and unseal the Vault, reducing developer burdon for running Vault locally
- Stores vault keys and tokens in `.vault/env` and adds the path to `.gitignore` to prevent developers from committing
  - `.vault/env/key` contains the unseal key, allowing Vault to decrypt the secrets database when starting up
  - `.vault/env/.env.local` contains the root token in envfile format, allowing us to inject the root token into the application in #294 to use the Vault api to store/retrieve secrets
- Adds documentation to `getting_started.md` explaining how Vault works and documents common troubleshooting methods

## Checklist
- [x] Tested locally?
- [x] Added new dependencies?
- [x] Updated the docs?
- [ ] Added a test?
- [x] Did you check for sensitive data, and remove any?
- [ ] Are additional approvals needed for this change?
- [ ] Are there potential vulnerabilities or licensing issues with any new dependencies introduced?

## Fixes 

- #issue

## To Be Done

- Next steps are to leverage the Vault by building out a `ConnectionCredentialStorageService` in #294

Issue #305, opened 2021-01-13
Update documentation for Data Hub developers to move their local envs to ConnectionCredentialStorageService


ronaldheft-usds:
Follows secrets management proposed in #284.

### Acceptance Criteria

- [ ] Communicate to all devs how to get their environment setup
- [ ] Documentation for future devs on the project

Issue #304, opened 2021-01-13
Migrate all connection-based logic to use the ConnectionCredentialStorageService


ronaldheft-usds:
Follows secrets management proposed in #284.

### Acceptance Criteria

- [ ] `ConnectionCredentialStorageService` is built and ready for use (#294)
- [ ] Migrated every use of connection secrets being used through environment variables to use the `ConnectionCredentialStorageService`

### For Merge and Deployment
- [ ] TEST and PROD environment have their Key Vaults ready for use (#300, #301)
- [ ] A plan is communicated to all devs when the cutover happens

Issue #295, opened 2021-01-13
Implement an Azure Key Vault version of the ConnectionCredentialStorageService


ronaldheft-usds:
Follows secrets management proposed in #284.

### Acceptance Criteria

- [ ] Uses the Azure client library for secure secrets access
- [ ] Configured for use in cloud environments
- [ ] Access credentials are obtained automatically from a service running in the Azure environment
- [ ] The Key Vault used is configurable
- [ ] A logical and clear storage structure for storing and retrieving secrets in a Key Vault is used

PR #392, opened 2021-01-28
Application secrets management

ronaldheft-usds:
This PR implements secrets management as defined in epic #190.

Instead of creating PRs for each issue, I've decided to batch up the entire set applications changes in one PR, since it's only functional once the entire task list has been implemented.

Azure environment changes will end up in a separate PR, as environment changes will need to be deployed out prior the application code.

## Changes
- Implements a CredentialService: #293
  - This ticket defines out the base inheritance strategy for both the `CredentialService` and `Credential` objects
  - The `CredentialService` will store the credential object as serialized JSON in whatever secret store is used
  - The secret store can be changed and configured with the environment variable `CREDENTIAL_STORAGE_METHOD`
  - The names of the classes have been shorted from the proposal for brevity and clarity
- Implements a Hashicorp Vault version of the CredentialService: #294
  - Implements Hashicorp Vault API using the existing HTTP library we're using, [fuel](https://github.com/kittinunf/fuel)
  - Adds a dependency to [mock-fuel](https://github.com/KennethWussmann/mock-fuel) to enable easier testing
  - Documents in `getting_started.md` how to use the local version of Hashicorp Vault
  - Fixes some issues with unsealing and initializing the Vault from my earlier PR #325 (anyone who initialized a Vault prior to this PR will need to re-initialize their Vault)
- Implements Azure Key Vault version of the CredentialService: #295 
  - Adds an environment variable `CREDENTIAL_KEY_VAULT_NAME` to define the Azure Key Vault name
  - Adds a dependency to azure-security-keyvault-secrets and azure-identity to use Azure's client for Key Vault
- Connects the SftpTransport to the CredentialService #304
  - Adds additional logging when credential do not exist in the secret store
  - Removes the existing environment variable in docker-compose.yml
- Updates the `SFTP-SETUP.md` with directions on how to use the secret store #305

## TLDR; What Do I Care About as a Developer?
- Documented how to use the secrets manager in `getting_started.md`.
  - If you don't do anything, you'll use an in-memory version of the secret store that will not persist, which may or may not be okay for the type of development you're doing.
  - If you're using docker-compose, you need to run this command once from the `prime-router` folder: `touch .vault/env/.env.local`
  - You will need to follow the directions in `getting_started.md` to re-initialize your Vault, as the earlier code I committed did not work as expected.
- New CLI command `prime create-credential` for generate credential JSON or persisting directly to the secret store
- Updated `SFTP-SETUP.md` with directions on how to use the secret store instead of environment variables
- Added a dependency [mock-fuel](https://github.com/KennethWussmann/mock-fuel) to enable easier testing of HTTP calls using [fuel](https://github.com/kittinunf/fuel).

## Areas I'm Looking for Feedback

- This is my first ever Kotlin code written (although I come from a Java background). Am I following best practices for Kotlin and team standards?
- Names have been changes for brevity from the proposal. Do the shorter names make sense, or should I go back to the more verbose `ConnectionCredential*` naming?
- My strategy for dependency injection. I'm finding Kotlin has a lot of nice features built right into the language and this was one of the cool ones.
- Logging of credential access.
- Any other areas that need a specific call out, I'll add a comment in the code.

## Checklist
- [x] Tested locally?
- [x] Added new dependencies? Yes, mock-fuel, azure-security-keyvault-secrets, azure-identity
- [x] Updated the docs? Yes, `getting_started.md`, `SFTP-SETUP.md`. Also added a helpful comment in `docker-compose.yml` for first-time errors that occur.
- [x] Added a test?
- [x] Are additional approvals needed for this change? Yes, this is a large change. We should get multiple approvals and take care in rolling this out.

## To Be Done

- Before this can be deployed, Key Vaults need to be stood up in Azure and secrets migrated to them.
  - This will be handled in the remaining tasks in epic #190


prime-router/docker-compose.yml
```diff
@@ -13,6 +13,11 @@ services:
         source: ./target
         target: /prime-data-hub-router
     entrypoint: [/prime-data-hub-router/start_func.sh]
+    env_file:
+      # On first run of docker-compose, create this file, it will get populated, but docker-compose doesn't allow optional env_file
+      # Example: touch .vault/env/.env.local
+      # https://github.com/docker/compose/pull/3955
+      - ./.vault/env/.env.local
```
ronaldheft-usds, line 20: This bug is really annoying. If this file doesn't exist docker-compose does not start. It's a simple work around to creating the file (and it only needs to be done once), but I haven't found a great solution around this issue that also ensures we don't accidentally check in our local Vault keys.

prime-router/src/main/kotlin/credentials/Credential.kt
```diff
@@ -0,0 +1,33 @@
...
+data class UserPassCredential(val user: String, val pass: String) : Credential()
+
+@JsonTypeInfo(
+    use = JsonTypeInfo.Id.NAME,
+    include = JsonTypeInfo.As.PROPERTY,
+    property = "@type"
+)
+@JsonSubTypes(
+    JsonSubTypes.Type(value = UserPassCredential::class, name = "UserPass")
+)
```
ronaldheft-usds, line 20: We will need to maintain this list for any additional credential types that get added in the future.

ronaldheft-usds:
This PR is ready for review, but do not merge! Azure changes are required for this to work when deployed out.

prime-router/src/main/kotlin/credentials/CredentialService.kt
```diff
@@ -0,0 +1,36 @@
...
+
+    /* Methods to implement in subclasses */
+
+    protected abstract fun fetchCredential(connectionId: String): Credential?
+    protected abstract fun saveCredential(connectionId: String, credential: Credential)
+
+    /* Base implementation for credentialService with validations */
+
+    fun fetchCredential(connectionId: String, callerId: String, reason: CredentialRequestReason): Credential? {
+        require(URL_SAFE_KEY_PATTERN.matches(connectionId)) { "connectionId must match: ${URL_SAFE_KEY_PATTERN.pattern}" }
```
RickHawesUSDS, line 19: Is the callerId expected to be? Suggest a comment block would be helpful.

RickHawesUSDS (approved):
Looks like top-notch work. A few comments that you should consider.

prime-router/src/main/kotlin/credentials/AzureCredentialService.kt
```diff
@@ -0,0 +1,33 @@
+package gov.cdc.prime.router.credentials
+
+import com.azure.core.credential.TokenCredential
+import com.azure.identity.DefaultAzureCredentialBuilder
+import com.azure.security.keyvault.secrets.SecretClient
+import com.azure.security.keyvault.secrets.SecretClientBuilder
+
+object AzureCredentialService : CredentialService() {
+
```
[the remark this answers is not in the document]
ronaldheft-usds: @RickHawesUSDS I agree on two points, but disagree on one of your points.

I agree that developers should not be using any of the service implementations directly, as dependency injections should be managing it for them. I also agree testability is a major concern with using singletons, and [while mockk does make testing singletons easier](https://notwoods.github.io/mockk-guidebook/docs/mocking/static/), it can still get messy.

What I disagree with is moving away from the singleton pattern. I think the issue here is visibility.

The credential service implementations should be marked as `internal` so they can not be instantiated outside of the package. If the implementations are marked as `internal`, the only way to use the credential service is through the [`CredentialManagement` interface](https://github.com/CDCgov/prime-data-hub/pull/392/files#diff-756129fa5f504f0c5fe47e9ccad012f0282f36c1e2aae91bf3910e691f96bf02) ([example of use](https://github.com/CDCgov/prime-data-hub/pull/392/files#diff-15b4e6a4398afcc6af7babd9fee78ef4cdf859897db923ceb34f465a5f844cecR20)).

The way `CredentialManagement` initializes the variable `credentialService` it's really easy to override in a test. Using the following syntax, you can quickly return a mock:

```
every { <your_class_here> getProperty "credentialService" } returns mockk<CredentialService>()
```

This was a great point to raise. I changed the visibility of the implementation classes in the following commit: https://github.com/CDCgov/prime-data-hub/pull/392/commits/78d3045e4ab16f2a7b484fd03b3a6672a00d2c98

prime-router/src/main/kotlin/credentials/CredentialService.kt
ronaldheft-usds: I didn't have a strong option on a callerId, and I wasn't sure if we'd end with a UUID on the settings API. Right now it's using the same format as the SftpTransport which is expecting `org_service` ala `AZ_PHD__ELR`.

@RickHawesUSDS With the settings API, it looks like you're sticking with organization name, so in that case the expected value will remain the same. If that's correct I can update the method docs to make that clear. Thoughts?

jimduff-usds (approved):

PR #393, opened 2021-01-28
MVP Download page #3

mry-usds:
This PR does ...

## Changes
-  Add date, location in front of filename test-results.csv -> maybe YYYYMMDD-[state]-[county name]-test-results.csv
- Typo: remove trailing "`" at the end of the no data available message.
- Added SENT_AT != NULL to query to avoid duplicates
- Moved orgName/userName to local to avoid threading issues

### DIFF
--- a/prime-router/src/main/kotlin/azure/DatabaseAccess.kt
+++ b/prime-router/src/main/kotlin/azure/DatabaseAccess.kt
@@ -190,10 +190,12 @@ class DatabaseAccess(private val create: DSLContext) {
         receiverName: String,
     ): List<Header> {
         val cond = if (since == null) {
-            TASK.RECEIVER_NAME.like("$receiverName%")
+            TASK.SENT_AT.isNotNull()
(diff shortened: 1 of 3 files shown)


prime-router/src/main/kotlin/azure/DatabaseAccess.kt
```diff
@@ -190,10 +190,12 @@ class DatabaseAccess(private val create: DSLContext) {
         receiverName: String,
     ): List<Header> {
         val cond = if (since == null) {
-            TASK.RECEIVER_NAME.like("$receiverName%")
+            TASK.SENT_AT.isNotNull()
+                .and(TASK.RECEIVER_NAME.like("$receiverName%"))
         } else {
```
jimduff-usds, line 195: For the next round of work:  The "like" is a security flaw.   Once we switch to the new lineage tables, the organization and service are broken apart as separate fields, and you can do a simple exact match.

jimduff-usds (approved):

cglodosky (approved):

PR #399, opened 2021-01-29
Rick/hl7 batch

RickHawesUSDS:
This PR adds support for a single HL7 message format in addition to the existing HL7 batch format. 

## Background
The TX and other states use a single test per file format, while AZ and other states use a more efficient batch format with BHS and FHS segments and multiple tests per file. Added a new single test HL7 file format (called now called `HL7`) to the existing HL7 batch file format (called `HL7_BATCH`). 

Another related problem was that batching (ie. delayed actions) didn't work for HL7 or Redox schemas. This is because schemas intended for HL7 messages do not save as CSVs well and CSVs are the intermediate file format that is used when batching. The solution was to create an INTERNAL format that is used for delayed processing. All schemas can serialize to this format since it bypasses formatting and mapping. 

## Decisions Made
- Store reports that are queued for delayed processing in an `INTERNAL` file format. This file format is just a CSV, but one serialized without formatting or mapping. The format can be adjusted in the future if Reports contain raw HL7. 
- Made splitting of Reports into single items happen in the reports & batch function as opposed to the Send function. This worked better with the new lineage tracking but is less efficient. Download could explode because of this. 

## Changes
- INTERNAL, HL7 and HL7_BATCH formats
- CLI has an `output-format` parameter that replaces the `output-hl7` format. It allows all formats to be specified. 
- Report has a new split action to create single item reports from a report
- AZ HL7 feeds now batch like the other feeds

## Checklist
- [x] Tested locally?
- [ ] Added new dependencies?
- [ ] Updated the docs?
- [x] Added a test?

## Fixes 

- #146 TX Schema

## Future
- It is now clear to me that schema implies file format, we should rethink schema to include file format.  
- If we get a large number of HL7 messages, then the download page will need a new design


prime-router/metadata/organizations-local.yml
```diff
@@ -55,7 +55,7 @@
             host: sftp
             port: 22
             filePath: ./upload
-        format: HL7
+        format: HL7_BATCH
       - name: elr-prod
```
RickHawesUSDS, line 59: HL7_BATCH is the new format type.

prime-router/src/main/kotlin/OrganizationService.kt
```diff
@@ -36,14 +36,16 @@ data class OrganizationService(
...
+        HL7_BATCH, // HL7 with BHS and FHS headers
         REDOX;
         // FHIR
 
         fun toExt(): String {
             return when (this) {
                 CSV -> "csv"
                 HL7 -> "hl7"
+                HL7_BATCH -> "hl7"
                 REDOX -> "redox"
```
RickHawesUSDS, line 49: Both HL7 batch and HL7 have the same file extension

prime-router/src/main/kotlin/azure/ReportFunction.kt
```diff
@@ -237,25 +237,42 @@ class ReportFunction {
...
-                destinations += destination
-                ReceiverEvent(Event.EventAction.BATCH, service.fullName, time)
+                destinations += "Sending ${report.itemCount} items to $serviceDescription at $time"
+                val event = ReceiverEvent(Event.EventAction.BATCH, service.fullName, time)
+                workflowEngine.dispatchReport(event, report, txn)
+                actionHistory.trackCreatedReport(event, report, service)
+                loggerMsg = "Queue: ${event.toQueueMessage()}"
+            }
+            service.format == OrganizationService.Format.HL7 -> {
+                destinations += "Sending ${report.itemCount} reports to $serviceDescription immediately"
```
RickHawesUSDS, line 257: Split the report into many single item reports for the single item HL7 format.

prime-router/src/main/kotlin/azure/WorkflowEngine.kt
```diff
@@ -62,7 +63,8 @@ class WorkflowEngine(
      * Place a report into the workflow
      */
     fun dispatchReport(nextAction: Event, report: Report, txn: Configuration? = null) {
-        val (bodyFormat, bodyUrl) = blob.uploadBody(report)
+        val forceFormat = if (nextAction.eventAction == Event.EventAction.BATCH) OrganizationService.Format.CSV else null
+        val (bodyFormat, bodyUrl) = blob.uploadBody(report, forceFormat)
```
RickHawesUSDS, line 67: Batches should always be CSV formatted.

prime-router/src/main/kotlin/serializers/Hl7Serializer.kt
```diff
@@ -31,7 +32,19 @@ class Hl7Serializer(val metadata: Metadata) {
...
+        if (report.itemCount != 1) error("Internal Error: multiple item report cannot be written as a single HL7 message")
+        val message = createMessage(report, 0)
+        outputStream.write(message.toByteArray())
+    }
+
+    /**
+     * Write a report with BHS and FHS segments and multiple items
+     */
+    fun writeBatch(report: Report, outputStream: OutputStream) {
         // Dev Note: HAPI doesn't support a batch of messages, so this code creates
```
RickHawesUSDS, line 48: The old `write` method is now called `writeBatch`

prime-router/src/main/kotlin/transport/SftpTransport.kt
```diff
@@ -79,35 +80,30 @@ class SftpTransport : ITransport {
...
-                        path + "/" + fileName
-                    )
-                // TODO: remove this over logging when bug is fixed
-                // context.logger.log(Level.INFO, "SFTP PUT succeeded: $fileName")
-            } finally {
-                client.close()
+            val client = sshClient.newSFTPClient()
+            client.fileTransfer.preserveAttributes = false
+            client.use {
+                it.put(makeSourceFile(contents, fileName), "$path/$fileName")
```
RickHawesUSDS, line 89: Splitting this method up into two functions.

prime-router/src/test/kotlin/azure/SendFunctionTests.kt
```diff
@@ -18,7 +18,12 @@ import java.time.OffsetDateTime
 import java.util.UUID
 import java.util.logging.Level
 import java.util.logging.Logger
-import kotlin.test.*
+import kotlin.test.Ignore
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertNotNull
+import kotlin.test.assertNull
+import kotlin.test.assertTrue
```
RickHawesUSDS, line 27: Per style guide

prime-router/src/main/kotlin/azure/BatchFunction.kt
```diff
@@ -51,10 +51,14 @@ class BatchFunction {
                     actionHistory.trackExistingInputReport(it.task.reportId)
                     report
                 }
-                val outReports = when (receiver.batch?.operation) {
+                val mergedReports = when (receiver.batch?.operation) {
                     OrganizationService.BatchOperation.MERGE -> listOf(Report.merge(inReports))
                     else -> inReports
```
jimduff-usds, line 56: See my next comment below.   I think Batch should just always merge, and nothing else.

If some receiver doesn't want merging, then simply skip that step in the workflow.

prime-router/src/main/kotlin/azure/BatchFunction.kt
```diff
@@ -51,10 +51,14 @@ class BatchFunction {
...
                 }
+                val outReports = when (receiver.format) {
+                    OrganizationService.Format.HL7 -> mergedReports.flatMap { it.split() }
+                    else -> mergedReports
+                }
```
jimduff-usds, line 61: I'm confused by this split here.   It looks like the split() is occurring in ReportFunction, so why do it here?   Splitting here is really confusing - it means that, in some cases, BatchFunction does the _opposite_ of batching.

A cleaner solution might be to allow SendFunction to optionally run on a cron timer, just like BatchFunction does now.  Then the workflow for TX might be like this:
1.   ReportFunction creates separate split hl7 files for each TX test result.
2. Skip the BatchFunction step completely, and queue a Send request at a specific time.
3. Send wakes up at the desired TX time, and sends all the accumulated HL7 files to TX.

prime-router/src/main/kotlin/azure/Event.kt
```diff
@@ -20,11 +20,11 @@ abstract class Event(val eventAction: EventAction, val at: OffsetDateTime?) {
         TRANSLATE, // Deprecated
         BATCH,
         SEND,
-        WIPE,
+        WIPE, // Deprecated
         NONE,
```
jimduff-usds, line 24: ?  Are we no longer planning to Wipe data?

prime-router/src/main/kotlin/azure/ReportFunction.kt
```diff
@@ -237,25 +237,42 @@ class ReportFunction {
...
         else
             service.organization.description
-        val event = when {
+        val loggerMsg: String
+        when {
             validatedRequest.options == Options.SkipSend -> {
-                ReportEvent(Event.EventAction.NONE, report.id)
+                val event = ReportEvent(Event.EventAction.NONE, report.id)
+                workflowEngine.dispatchReport(event, report, txn)
+                actionHistory.trackCreatedReport(event, report, service)
```
jimduff-usds, line 245: I am working toward combining `dispatchReport` and `trackCreatedReport` in an upcoming future PR, so this repeated pattern of the two called together, will get cleaner.

prime-router/src/main/kotlin/azure/WorkflowEngine.kt
jimduff-usds: What about future incoming hl7 data?    Do we really want to transform that to CSV, then back to HL7?

prime-router/src/main/kotlin/azure/ReportFunction.kt
```diff
@@ -237,25 +237,42 @@ class ReportFunction {
...
             service.batch != null -> {
                 val time = service.batch.nextBatchTime()
-                val destination = "Sending ${report.itemCount} items to $serviceDescription at $time"
-                destinations += destination
-                ReceiverEvent(Event.EventAction.BATCH, service.fullName, time)
+                destinations += "Sending ${report.itemCount} items to $serviceDescription at $time"
+                val event = ReceiverEvent(Event.EventAction.BATCH, service.fullName, time)
+                workflowEngine.dispatchReport(event, report, txn)
+                actionHistory.trackCreatedReport(event, report, service)
+                loggerMsg = "Queue: ${event.toQueueMessage()}"
```
jimduff-usds, line 254: This is a bug that needs fixing before you merge this PR.   Down inside dispatchReport, format is secretly changing to CSV, but the format is now incorrect in ActionHistory.   I suggest the format decision should be refactored out and made here, then sent to both dispatchReport and trackCreatedReport.

Again, note that this is temporary.  My intention is that the new `report_file `table will be the primary table for all report information, not the `task` table.   The `task` table will remain, but only as a mutable "todo list" of work that needs to be done, with the details about reports properly normalized into `report_file`. Right now, `task` and `report_file` are redundant, which is a dangerous situation I intend to fix.

Furthermore, see my comment on putting Send on a cron timer.     That would enable you to skip the batch step entirely, and maybe you don't even need the `forceFormat` stuff all?

jimduff-usds (changes requested):
I have a suggestion for a possible alternate way to do this single-HL7 functionality, in my comments.

prime-router/src/main/kotlin/azure/BatchFunction.kt
RickHawesUSDS: Disagree: batch is what happens at a later time. It should do all the things that should happen at a later time.

prime-router/src/main/kotlin/azure/BatchFunction.kt
RickHawesUSDS: Again we have a different view of what the batch function is for. Send is the function that does retries. Batch is for time-based work.

prime-router/src/main/kotlin/azure/Event.kt
RickHawesUSDS: Right. Going to be done by delete policy.

prime-router/src/main/kotlin/azure/ReportFunction.kt
RickHawesUSDS: Ok. I'll look at this bug.

jimduff-usds: Actually, I think you should go ahead and merge, and I'll fix this.  I'd like to do my merge on top of yours, as I know there are going to be a lot of conflicts.

jimduff-usds (approved):

prime-router/src/main/kotlin/azure/ReportFunction.kt
RickHawesUSDS: Nice.

prime-router/src/test/kotlin/azure/SendFunctionTests.kt
```diff
@@ -19,7 +19,12 @@ import java.time.OffsetDateTime
...
```
RickHawesUSDS, line 28: Fix a style violation.

prime-router/src/main/kotlin/Report.kt
```diff
@@ -25,6 +25,33 @@ typealias DefaultValues = Map<String, String>
...
+        fun isSingleItemFormat(): Boolean {
+            return when (this) {
+                REDOX -> true
+                HL7 -> true
+                else -> false
+            }
+        }
+    }
+
     /**
```
RickHawesUSDS, line 55: Moving Format into Report from OrganizationService. This is a step in the right direction. I'd like to explore how it could be a property of Schema, but that is something for later.

MauriceReeves-usds: I don't know. I kind of like the idea that a schema can be chimeric and handle different output formats. I see merits to doing it both ways, so I'm not opposed to having the format in the schema if everyone wants to go that way, but if do go that direction, then we may need to revisit how we're doing the `csvFields` logic. It could need to be refactored if a schema is only ever used for CSV serialization.

MauriceReeves-usds (approved):

jimduff-usds (changes requested):
I found two problems, both can be seen by simply running ./quick-test.sh.   They are both the same exception; its not clear if they need the same solution.

**First issue:**
Run 
`/.prime data --input-fake 50 --input-schema primedatainput/pdi-covid-19 --output-file foo.csv`
Then run
`./prime data --input-schema primedatainput/pdi-covid-19 --input ./foo.csv --output bar.csv --route`
I get this exception:
```
Exception in thread "main" java.lang.IllegalStateException: Internal Error: multiple item report cannot be written as a single HL7 message
	at gov.cdc.prime.router.serializers.Hl7Serializer.write(Hl7Serializer.kt:39)
	at gov.cdc.prime.router.cli.ProcessData$run$1.invoke(main.kt:261)
```
Its not clear to me which state is causing this.  Would need more investigation.

**Second issue:** 
This call in ./quick-test.sh
`./prime data --input-fake 50 --input-schema nd/nd-covid-19 --output-dir $outputdir --target-state ND --output-hl7`
 gives this error:

```
Generate fake ND data, HL7!
Exception in thread "main" java.lang.IllegalStateException: Internal Error: multiple item report cannot be written as a single HL7 message
	at gov.cdc.prime.router.serializers.Hl7Serializer.write(Hl7Serializer.kt:39)
	at gov.cdc.prime.router.cli.ProcessData$run$1.invoke(main.kt:261)
	at gov.cdc.prime.router.cli.ProcessData$run$1.invoke(main.kt:46)
	at gov.cdc.prime.router.cli.ProcessData.writeReportsToFile(main.kt:174)
	at gov.cdc.prime.router.cli.ProcessData.run(main.kt:258)

```
Note:  ND does NOT appear in organizations.yml right now!, so there's no places to put the `format: HL7_BATCH` annotation.   (As aside: I think that being able to generate data for a schema without having its recipient listed in organizations.yml is a nice feature, and does (perhaps) imply that we should put the `format:` into the schema.)

In the meantime, to fix this error, maybe we need both a `--output-hl7` and a -`-output-hl7-batch` options to `./prime`.    Or maybe some other solution.   Anyway, clearly ND is a good test case.

jimduff-usds (approved):

PR #404, opened 2021-01-29
Synthetic Data Generation

MauriceReeves-usds:
## Overview
This PR does creates the first round of synthetic data generation from another file.

For the first round of this work, the synthesize strategies are hard coded with the key PII files. A future PR will include the ability to control the synthesize strategies from a config file, and potentially set up distinct strategies per organization.

The synthesize step happens in between the input and output steps. Once a file is read in, or generated, as a report, the report will call synthesize and transform itself before moving into the output steps.  I did the synthesizing work inside `Report` so I could use the table abstraction that Report wraps. As the comments note in the code we could potentially change which table abstraction we use, so the best place to transmogrify the data is inside `Report`.

### Notes:
- There is a test I created, which is currently set to `@Ignore` because the shuffling of data is non-deterministic which means there would be times the test would break because the shuffling may not have moved one or more column values
- We should probably review the fields I have identified as PII per our discussion with CDC at our demo.

## Changes
- Breaks the generation of fake data out into a separate service that can be used in more than one place
- Adds a new command line option `--synthesize`, that lets us synthesize data from an input file (or even from generated fake data if we really want to)
- Adds the logic in `Report` to create a new copy of itself with columns changed according to the map of `SynthesizeStrategy` passed in to the method. The map can be null at which point the report just does a straight pass-through. We can revisit this if we want and move the default strategies that are currently in `main` into `Report` instead.
- Perform pass through on `--target-state` and `--target-county` as well for the synthesizing of data

## Checklist
- [x] Tested locally?
- [ ] Added new dependencies?
- [x] Updated the docs?
- [x] Added a test?
- [x] Are additional approvals needed for this change?


prime-router/src/main/kotlin/Report.kt
```diff
@@ -200,6 +211,40 @@ class Report {
...
+                    // examine the synthesizeStrategy for the field
+                    // can be one of three values right now:
+                    // empty column, shuffle column, pass through column untouched
+                    SynthesizeStrategy.SHUFFLE -> {
+                        val shuffledValues = table.column(it.name).asStringColumn().shuffled()
+                        StringColumn.create(it.name, shuffledValues)
+                    }
+                    SynthesizeStrategy.FAKE -> {
+                        // generate random faked data for the column passed in
+                        buildFakedColumn(it.name, it, targetState, targetCounty)
```
jimduff-usds, line 235: So, just so I'm clear, this generation of fake data is separate from the --input-fake option?
Put another way, supposed I were to use both `--input-fake`, and `--synthesize` - in that case, the system would generate a fake Report, then here, some of those fake data would get overwritten with new fake data, true?   "Even faker date", eh?  ;)     

That seems fine, I just want to make sure I'm understanding how it works.

prime-router/src/main/kotlin/Report.kt
```diff
@@ -200,6 +211,40 @@ class Report {
...
+        val columns = schema.elements.map {
+            val synthesizedColumn = synthesizeStrategies[it.name]?.let { strategy ->
+                // look in the mapping parameter passed in for the current element
+                when (strategy) {
+                    // examine the synthesizeStrategy for the field
+                    // can be one of three values right now:
+                    // empty column, shuffle column, pass through column untouched
+                    SynthesizeStrategy.SHUFFLE -> {
+                        val shuffledValues = table.column(it.name).asStringColumn().shuffled()
+                        StringColumn.create(it.name, shuffledValues)
```
jimduff-usds, line 231: I have never dealt with synthetic data generation, but I'm quite concerned that, with _small_ files, it would be possible to readily reconstruct the original data from shuffled data.

Is that a concern, and if so, how could we address it?

prime-router/src/main/kotlin/Translator.kt
```diff
@@ -29,29 +29,7 @@ class Translator(private val metadata: Metadata) {
         return metadata.organizationServices.map { service -> translateByService(input, service, defaultValues) }
     }
```
jimduff-usds, line 30: @mauricereeves , I'm having trouble seeing what changes were made to Translator.kt - could you just put in a brief comment?

At first, it looked like major changes, but the more I looked at it, it seems like some methods just got moved around a bit.

Thanks.

prime-router/src/main/kotlin/FakeReport.kt
```diff
@@ -1,154 +1,183 @@
 package gov.cdc.prime.router
```
jimduff-usds, line 2: @mauricereeves , it looks like you did some nice cleanup on FakeReport.kt - could you put in a brief comment explaining the changes?
Thanks.

prime-router/src/main/kotlin/Mappers.kt
```diff
@@ -282,6 +282,18 @@ class Obx8Mapper : Mapper {
...
+class SyntheticDataMapper : Mapper {
+    override val name = "syntheticData"
+
+    override fun valueNames(element: Element, args: List<String>): List<String> {
+        return args
+    }
+
+    override fun apply(element: Element, args: List<String>, values: List<ElementAndValue>): String? {
+        return "PII"
+    }
```
jimduff-usds, line 294: I don't understand how this mapper fits into the work of generating synthetic data - can you add a comment?

prime-router/src/main/kotlin/Report.kt
```diff
@@ -200,6 +211,40 @@ class Report {
         return Report(schema, Table.create(columns), fromThisReport("deidentify"))
     }
 
+    // takes the data in the existing report and synthesizes different data from it
+    // the goal is to allow us to take real data in, move it around and scramble it so it's
+    // not able to point back to the actual records
```
jimduff-usds, line 216: This looks good here for now, but as this grows, let's pull it out into a separate Synthesizer.kt file, or something like that.    Then that file can include the "default" synthesis that's currently in cli/main, so that its all in one place, and its more generally available, for example, if we wanted to generate synthetic data inside the actual hub workflow while a customer is testing.

Seems like we could have a Synthesizer.synthesizeData(inputReport: Report): Report, then we could add a SynthesizerFunction.kt azure function, that could be called as part of a hub workflow, if that's a good use case.

prime-router/src/test/kotlin/ReportTests.kt
```diff
@@ -132,4 +130,140 @@ class ReportTests {
...
+        assertEquals("sarah", synthesizedReport.getString(0, "first_name"))
+        assertEquals("mary", synthesizedReport.getString(1, "first_name"))
+        assertEquals("roberta", synthesizedReport.getString(2, "first_name"))
+        assertEquals("", synthesizedReport.getString(0, "ssn"))
+        assertEquals("", synthesizedReport.getString(1, "ssn"))
+        assertEquals("", synthesizedReport.getString(2, "ssn"))
+    }
+
+    // ignoring this test for now because shuffling is non-deterministic
+    @Test
```
jimduff-usds, line 232: You could just pull both the orig and the shuffled data into Sets, then the Sets should be equal - all the original names are still there, with the same number of rows.  That seems like a pretty good test.

jimduff-usds (commented):
Nice work.   My main concern is the one I wrote about small files, and someone thinking they are obscuring data when they really aren't.    I'll submit this as a comment for now, for you or others to weigh in.

prime-router/src/main/kotlin/Report.kt
MauriceReeves-usds: That is correct. This is a different step from `--input-fake`.

prime-router/src/main/kotlin/Report.kt
MauriceReeves-usds: It is potentially possible that this could happen. This is why I marked the one test as `@Ignore` because there were times it would fail because the potential of shuffling the results didn't move one of the data points I was checking. We could always check for some number of records threshold and below that number, skip shuffling in favor of faking data.

prime-router/src/main/kotlin/Translator.kt
MauriceReeves-usds: I was originally going to have the synthetic data happen in Translator, but it was increasingly complex and problematic so I rolled back my changes. There shouldn't be any real changes to review there.

prime-router/src/main/kotlin/FakeReport.kt
MauriceReeves-usds: Yes, will do.

jimduff-usds:
Oh, and one other concern:   the vast majority of real-life results are covid-19 negatives, yet positive test results are the really important ones to get right.   It might be important to force at least one positive result into every synthesized data set.

prime-router/src/main/kotlin/Mappers.kt
MauriceReeves-usds: This should have been removed. I missed this during my revert of my work in the translator.

prime-router/src/main/kotlin/Report.kt
MauriceReeves-usds: The problem with doing it outside of `Report` is we lose access to the `Table` abstraction we're using in `Report` and that makes the work of synthesizing the data, moving it around, etc easier. There are notes on the code asking us not to leak the abstraction. An alternative would be to subclass `Report` and then add a `toReport` function that would make it immutable again.

jimduff-usds: Good point.  I remember this problem, and I'm afraid leakage has already occurred.   I had the same issue when I built the JurisdictionalFilters, and I just cheated and used the Table features.   

I don't like the subclass idea -doesn't seem worth it.  I support just leaving it as is for now; I think leakage is the way to go over the long term.

prime-router/src/main/kotlin/Report.kt
jimduff-usds: I strongly support that idea of a threshold, or some other workaround for small data   Can you consider it for the next round of work?

prime-router/src/main/kotlin/Translator.kt
jimduff-usds: gracias

prime-router/src/main/kotlin/Report.kt
RickHawesUSDS: I like not leaking the Table abstraction. I'm not sure we want to keep this table package in the long term.

prime-router/src/main/kotlin/Report.kt
RickHawesUSDS: I'm pretty sure the Shuffle strategy (even for larger data sets) won't pass privacy scrutiny with regard to PII. Can we say that it can't be used on certain elements like phone numbers?

prime-router/src/test/kotlin/ReportTests.kt
```diff
@@ -1,8 +1,6 @@
 package gov.cdc.prime.router
 
-import kotlin.test.Test
-import kotlin.test.assertEquals
-import kotlin.test.fail
+import kotlin.test.*
```
RickHawesUSDS, line 3: There is a style guide rule that bans using wildcard inputs.

RickHawesUSDS (commented):
My biggest concern is about the shuffle strategy used on certain elements and the small number problem that Jim pointed out. I'm especially concerned about the PII elements. I'm willing to making the results less realistic to protect privacy.

prime-router/src/main/kotlin/Report.kt
MauriceReeves-usds: Yeah we can definitely limit strategies based on the data point. Let me think about how to do that.

prime-router/src/test/kotlin/ReportTests.kt
MauriceReeves-usds: Could definitely do that.

prime-router/src/main/kotlin/cli/main.kt
```diff
@@ -283,8 +301,28 @@ class ProcessData : CliktCommand(
             }
         }
 
+        // synthesize the data here
+        // todo: put these strategies into metadata so we can load them from a file
+        val synthesizeStrategies = mapOf(
+            "patient_last_name" to Report.SynthesizeStrategy.SHUFFLE,
+            "patient_first_name" to Report.SynthesizeStrategy.SHUFFLE,
```
jimduff-usds, line 308: I'm not seeing a reason to ever shuffle real names - let's just always make them fake.

Date of Birth, on the other hand, might be valuable to fake the day and month, but keep the year, so the data has realistic age shape.

jimduff-usds (approved):
I think you should always fake first and last names.

PR #430, opened 2021-02-01
Versioning and Git Flow Proposal

cglodosky:


.gitignore
```diff
@@ -77,5 +77,10 @@ prime-router/source.mry
 prime-router/result_files/*
 prime-router/routed_files/*
 
+# Terraform files
```
ronaldheft-usds, line 80: I don't think this `.gitignore` matches with this PR. Is this from your Terraform branch?

prime-router/docs/proposals/0004-git-flow-and-versioning.md
```diff
@@ -0,0 +1,63 @@
...
+The overall platform should be versioned in a temporally distinct way, yet allowing for multiple production releases within a single calendar day.  The proposed format is as follows:
+
+> YYYY.0M.0D_MICRO
+
+|YYYY|0M|0D|MICRO
+|---|---|---|---|
+|Full year|Zero-padded month|Zero-padded day|Release number|
+|2019, 2020, ...|01, 10, ...|01, 30, ...|0, 1, 2, ...
+
+The proposed versioning scheme is based off of `Calendar Versioning`<sup>1</sup>.  A separate proposal may leverage `Semantic Versioning`<sup>2</sup> to add granularity to the individual components within the PRIME Data Hub platform.
```
ronaldheft-usds, line 30: So you're proposing deployments use calendar versioning, but we update our individual components to use semantic versioning? I like that hybrid approach.

prime-router/docs/proposals/0004-git-flow-and-versioning.md
```diff
@@ -0,0 +1,63 @@
...
+
+Upon merge, the GitHub actions `test_release` workflow will be triggered to deploy to the `prime-data-hub-test` resource group within Azure.  The merged changes should then be thoroughly tested within the cloud environment.
+
+This proposed flow is based off of the `GitHub Flow`<sup>3</sup>.
+
+## Production Release
+
+When a collection of features is deemed ready for production deployment, a release should be published.  Upon publish, the `prod_release` workflow will be triggered to deploy to the `prime-data-hub-prod` resource group within Azure.
+
+To create a release, navigate to the `Create a new release` link on the right-hand side of the main GitHub repository page.  Enter the appropriate version, title, and description and click `Publish release`.  The title should be a succinct overview of the theme(s) surrounding each release.  The description should contain a detailed change log describing the changes that were made to the platform.
```
ronaldheft-usds, line 53: How are failures handled in either case? If a deployment fails, to re-run another one? Will we have a manual GitHub Action trigger to re-run a deployment?

What about hotfixes to production in the event something was merged to test and a PR was merged to test? Should we be treating as everything merged into master is safe to deploy to production, and there will never be a state when master is unsafe to deploy?

Can anyone run a deployment? Only a developer who has changes in master? Select team members?

How often should we be releasing code to production?

ronaldheft-usds (commented):
Added some questions to the proposal.

prime-router/docs/proposals/0004-git-flow-and-versioning.md
RickHawesUSDS: CalVersioning is simple to understand and avoids is this release a major or minor release.

prime-router/docs/proposals/0004-git-flow-and-versioning.md
RickHawesUSDS: So should the release comment be a copy of the release notes. Or point to the release notes section?

RickHawesUSDS: Is there a use for the pre-release feature?

.gitignore
cglodosky: I removed this.

prime-router/docs/proposals/0004-git-flow-and-versioning.md
cglodosky: > How are failures handled in either case? If a deployment fails, to re-run another one? Will we have a manual GitHub Action trigger to re-run a deployment?

If a deployment fails because of a bad build, this would be caught going into the test environment.  Code updates can be pushed to then trigger the build to run again.  In the case of a production deployment failing, we have the ability to go in and re-run the exact version of the workflow that ran via GitHub actions.

> 
> What about hotfixes to production in the event something was merged to test and a PR was merged to test? Should we be treating as everything merged into master is safe to deploy to production, and there will never be a state when master is unsafe to deploy?

With the assumptions that changes are tested first locally and then in the testing environment - code in the master branch should be safe to deploy to production.  If a developer is testing in the test environment and finds a change to not produce the desired effect, then it should be communicated to the team to not publish any releases until that person has either reverted the change or pushed and tested a fix.

> 
> Can anyone run a deployment? Only a developer who has changes in master? Select team members?

I'm open to suggestions on this.  Currently, anybody in the CDCgov organization with write access can publish a release.  Do we want that process to be different?

> 
> How often should we be releasing code to production?

I'm also open to suggestions here as well.  We currently organically find a point-in-time where a set of features are ready and do an *ad hoc* deployment

cglodosky: > So should the release comment be a copy of the release notes. Or point to the release notes section?

My thought was that the release comment would serve as the container for release notes, but if you prefer a separate document for release notes it can serve as a pointer instead.

cglodosky: > Is there a use for the pre-release feature?

If we operate under the assumption that code making it to `master` is production-ready - then I do not think we will have a need tor a pre-release.  Alternatively, each release could start as a pre-release so that a second person can review it prior to deploying to production.  Thoughts?

RickHawesUSDS: > > Can anyone run a deployment? Only a developer who has changes in master? Select team members?
> 
> I'm open to suggestions on this. Currently, anybody in the CDCgov organization with write access can publish a release. Do we want that process to be different?
> 

This is a good discussion. I think that publishing a release is an equivalent power to being able to commit to master.

.gitignore
```diff
@@ -78,4 +78,4 @@ prime-router/result_files/*
 prime-router/routed_files/*
 
 # Hashicorp Vault Secrets
-prime-router/.vault/env/**
\ No newline at end of file
+prime-router/.vault/env/**
```
ronaldheft-usds: There is an extra line that got added here. Don't mean to be a stickler, but our PRs should only contains changes to files we actively developed. If we're making code formatting changes, that should be a separate PR. If we're having issues where different developer's IDEs are formatting code differently, we should call that out, and improve our linting and formatting rules so it's consistent across the team. This could be the source of some of our merge issues (different linting rules per dev).

prime-router/docs/proposals/0004-git-flow-and-versioning.md
ronaldheft-usds: Agreed. I'm all for CalVersioning for deployments.

prime-router/docs/proposals/0004-git-flow-and-versioning.md
ronaldheft-usds: > If a deployment fails because of a bad build, this would be caught going into the test environment. Code updates can be pushed to then trigger the build to run again. In the case of a production deployment failing, we have the ability to go in and re-run the exact version of the workflow that ran via GitHub actions.

I would disagree with this statement to some effect. While deployment errors should be caught going to a test environment, that's not always the case. Sometimes, especially in the state of the project we're in now, it's possible there are slight differences between environments that cause issues, so a deployment to test might succeed but a deployment to prod fails.

Regardless, I think ensuring we have the ability to re-run a deployment in the event of a failure is essential, and we should get that out of the box with GitHub Actions. I would also suggest we add a restriction to our production deployment that the test deployment's last run must have succeeded for the production deployment to run.

ronaldheft-usds: > With the assumptions that changes are tested first locally and then in the testing environment - code in the master branch should be safe to deploy to production. If a developer is testing in the test environment and finds a change to not produce the desired effect, then it should be communicated to the team to not publish any releases until that person has either reverted the change or pushed and tested a fix.

Agree with this, and I think this can be mitigated if we have a more limited set of team members who can run deploys. I would agree with @RickHawesUSDS that only team members with direct commit access to master. This would be either team members who have "Maintain" or "Admin" access to the repository.

ronaldheft-usds (changes requested):

prime-router/docs/proposals/0004-git-flow-and-versioning.md
```diff
@@ -0,0 +1,63 @@
...
+Upon merge, the GitHub actions `staging_release` workflow will be triggered to deploy to the `prime-data-hub-staging` resource group within Azure.  The merged changes should then be thoroughly tested within the cloud environment.
+
+This proposed flow is based off of the `GitHub Flow`<sup>3</sup>.
+
+## Production Release
+
+When a collection of features is deemed ready for production deployment, a release should be published.  Upon publish, the `prod_release` workflow will be triggered to deploy to the `prime-data-hub-prod` resource group within Azure.
+
+To create a release, navigate to the `Create a new release` link on the right-hand side of the main GitHub repository page.  Enter the appropriate version, title, and description and click `Save draft`.  The title should be a succinct overview of the theme(s) surrounding each release.  The description should contain a detailed change log describing the changes that were made to the platform.  A second person should then review the release for clarity and completeness before the release is published.
+
```
RickHawesUSDS, line 54: Thanks for the change to make this a two person operations.

RickHawesUSDS (approved):
Thanks for this work.

ronaldheft-usds (approved):

Issue #495, opened 2021-02-11
Harden the API end-points


RickHawesUSDS:
As a security professional, I would like to see the DH's end-points hardened against malicious attacks that manipulate the payload sent through the data hub. 

**Goals**
The API end-points should have a number of limits, checks and validations that prevents from an attacker who knows about the inner workings of the data hub (ie. read our source code) from information disclosure and service disruption. 

This epic is not about DDOS attacks which will be handled by the WAF epics. 

**Acceptance Criteria**
- [X] SQL injection protection by design
- [X] Limits on report size overall size
- **ICEBOX**: URL hashing in the of the data hubs reports end-point

Tasks:
- [X] SQL injection protection: Fix the "LIKE" query currently used for download. 
- [X] Limits:  Write brief proposal on how to do data element limits
- [X]  Limits: Impl report size 
- [X] What are the threats we're trying to avoid by using URL hashing?      Requires an additional secret known by the sender to use as a salt.

Moved to other tickets.
- [ ] Impl element size limits
- [ ] Expiring URLs.

Issue #494, opened 2021-02-11
Add report itemCount and overall byte size limits


RickHawesUSDS:
Protect the report end-point from very large gibberish request. 

**AC**
[X] Reasonable limits on size of request
[ ] What if the request body is 100x memory size, would all resources be consumed? 
[X] What if there is a 200000 row CSV being loaded? 
[X] Write test framework to cover this and other test cases, that can be run against the azure functions, either locally or in test.

Issue #493, opened 2021-02-11
Add data element size limits


RickHawesUSDS:
As part of hardening our report end-point and to improve data validation in our schema, add a size limit to each data element type that is based on reasonable limits and HL7 standard. 

Acceptance Criteria
[ ] Consideration of HL7 and Redox limits
[ ] Allow individual custom schemas to set limits 
[ ] truncate fields when converting  
[ ] Consider whether to Error or Warn on truncation.

Issue #492, opened 2021-02-11
Add hash, expire, and user the download URL for the download end-point


RickHawesUSDS:
As a practice of defense in depth, prevent URL tampering in the end end-point by adding a URL hash that is checked on every request. Also, add to the end-points URL scoping parameters like expiration time and user that limit the URLs scope. 

Suggestions:

Use a FIPS 180-4 approved hash algorithm. https://csrc.nist.gov/publications/detail/fips/180/4/final
Put initialization materials into the secrets keyvault.
Rotation of initialization materials in secret keyvault.
Develop a common modules and libraries for URL signing
Develop a conical form of every URL query parameters
add a hash= query parameter as required for every URL
check hash on entry

PR #519, opened 2021-02-17
API Hardening Proposal

jimduff-usds:
This PR has a proposal for how we'll approach API security hardening.

## Changes
- Proposal 0005
- 
## Issues

- #495 
- #493 
- #494 
- #492 
- #491


prime-router/docs/proposals/0005-api-hardening.md
```diff
@@ -0,0 +1,125 @@
...
+- MAX_BYTES - a reasonable limit on the overall size of the payload. (Azure limit is 100Meg.)   **Suggest: 40Meg. (allows 4k per row)** 
+- MAX_ITEMS - a limit on the number of Items (in a CSV, this is a limit on the number of rows + 1).  **Suggest: 10,000 Items (10,001 rows)**
+- **Phase 2 work?**: (MAX_COLUMNS - for CSVs only, a limit on the number of columns input. **Suggest: 1000 columns**)
+- MAX_ITEM_SIZE - byte limit on the size of one item.  This is most useful for reading in hl7 - we'll want a way to reject inputs without having to get bogged down in detailed item parsing.   **Suggest:  200K, to match Redox max**
+- MAX_ERRORS - Validation will fail immediately if the total number of Errors passes this threshold.  **Suggest: 100 errors**
+- Note:  no limit on warnings, since warnings by definition should not cause failures.
+
+These limits will be set globally, hardcoded, for now.  Could be set per schema in the future, with careful consideration of how that gets locked down.
+
+The limits will apply to both CSV and HL7, when appropriate.
```
MauriceReeves-usds, line 28: In the case if HL7, instead of speaking of 10,000 items/10,001 rows, we will have to limit it to some number of messages, and it should be less than 10,000. From my testing, that many faked HL7 messages is just over 40MB.

MauriceReeves-usds (approved):

prime-router/docs/proposals/0005-api-hardening.md
jimduff-usds: This is very helpful.   Very roughly, what would be the total Megs of 10,000 patients worth of test results, in hl7?  Maybe that should be our `MAX_BYTES`.   I realize you can only estimate, but we could round up from a ballpark number.

PR #520, opened 2021-02-18
Rename NAT Gateway Public IP

cglodosky:
Rename the public IP for the NAT Gateway back to the original (generic) naming.  Azure currently doesn't support renaming an IP address without destroying and re-creating.  Since we don't want the static IP to change, the name is being reverted.


ronaldheft-usds (approved):
Good catch. We don't want to lose our current IP. Hopefully Azure supports renaming in the future.

PR #521, opened 2021-02-18
Diagnostic Logging for Azure Resources

ronaldheft-usds:
This PR does adds logging via an event hub to pass to Splunk for several Azure resources.

## Changes
- Configures storage account blob access logging for Splunk (#442)
- Configures function app logging for Splunk (#443)
- Configures database logging for Splunk (#444)

With all of the above changes, all active resources using in prod will have logs that can be sent to Splunk.

## Checklist
- [x] Tested locally? Tested against my resource group.
- [ ] Added new dependencies?
- [ ] Updated the docs?
- [ ] Added a test?
- [ ] Did you check for sensitive data, and remove any?
- [ ] Are additional approvals needed for this change?
- [ ] Are there potential vulnerabilities or licensing issues with any new dependencies introduced?

## Fixes 

- #issue

## To Be Done

- Securely communicate all of the access keys to Splunk teams once these roll out to test and production.

PR #523, opened 2021-02-18
Vormetric Keys for Storage / Database

ronaldheft-usds:
This PR rotates our encryption to use the RSA keys provided by Vormetric.

## Changes
- Rotates storage account to use the Vormetric key
- Rotates database to use the Vormetric key

## Checklist
- [ ] Tested locally?
- [ ] Added new dependencies?
- [ ] Updated the docs?
- [ ] Added a test?

## Fixes 

- #issue

## To Be Done

- Once this is verified in test, we reach out to the IAMB team, they send us keys in prod, we update our config to use in prod, then deploy.


ronaldheft-usds:
@cglodosky I could use your assistance testing this in our test environment. We can investigate after #521.

operations/app/src/modules/database/main.tf
```diff
@@ -27,6 +27,18 @@ resource "azurerm_postgresql_server" "postgres_server" {
   }
 }
 
+data "azurerm_key_vault_key" "postgres_server_encryption_key" {
+  count = var.environment == "test" ? 1 : 0 // Only use Vormetric keys in test
```
cglodosky, line 31: Is this just until we are fully ready to roll out to production?  Or did I miss something that we aren't doing this in prod?

operations/app/src/modules/storage/main.tf
```diff
@@ -23,6 +23,14 @@ resource "azurerm_storage_account" "storage_account" {
   }
 }
 
+resource "azurerm_storage_account_customer_managed_key" "storage_key" {
+  count = var.environment == "test" ? 1 : 0 // Only use Vormetric keys in test
+  key_name = "${var.resource_prefix}-key"
```
cglodosky, line 28: For continuity, do we want to put the number of bits (ie. 2048, 4096) into the key name?  Or is everything 4096 except for the one above?

cglodosky (commented):
I left two questions as comments, but the rest looks good to me.

operations/app/src/modules/database/main.tf
ronaldheft-usds: Yes, it's only in test for the moment. We need to have the Vormetric team push keys to prod once we validate the environment.

operations/app/src/modules/storage/main.tf
ronaldheft-usds: I'll ask if they can rename the key. Also going to ask about adding it to staging once we get the staging environment up.

operations/app/src/modules/database/main.tf
ronaldheft-usds: I updated this to more easily change which environments are used by setting a variable that passes through.

operations/app/src/modules/storage/main.tf
ronaldheft-usds: Going to work on getting the key changed in test, but the key names are clearer in staging, and will be in prod.

ronaldheft-usds:
Aiming to test around 3:30pm EST, after @jimduff-usds gives me an overview of running smoke tests in staging.

cglodosky (approved):
Looks good to me.

PR #524, opened 2021-02-19
Add prod-specific conditionals for Function App environment

cglodosky:
Added app settings specific to the production instance of the Function App.  Since the production app is on an Elastic Premium service plan it requires additional environment variables not necessary in the test version of the Function App.


ronaldheft-usds (approved):
This looks good. We are creating a chicken and egg problem for the new environments, as front door is requiring the data from the existing instance of front door, but it's a great workaround until we get environment configuration stood up in Key Vault.

PR #530, opened 2021-02-22
Disable CI/CD to Test Environment for Pentesting

cglodosky:


.github/workflows/build.yml
```diff
@@ -2,7 +2,9 @@ name: Prime Data Hub
 
 on:
   pull_request:
-    branches: [ master, production ]
+    # Disable master branch for pentesting
+    # branches: [ master, production ]
+    branches: [ master ]
```
ronaldheft-usds, line 7: Shouldn't this remove `master` not `production`?

ronaldheft-usds (changes requested):

.github/workflows/build.yml
cglodosky: Oops...that was a bad copy/paste.  This is fixed.

ronaldheft-usds (approved):
Looks good now. Approved!

Issue #529, opened 2021-02-19
Update HL7 Results per Vermont DOH Feedback


MauriceReeves-usds:
Per testing our HL7 messages, Vermont provided the following validation errors they'd like addressed:

```
Segment:                            MSH|1|,PID|1|

 

Field/Component:            MSH-3.1 (MSH/SendingApplication/NamespaceID)

Field/Component:            MSH-4.1 (MSH/SendingFacility/NamespaceID)

Field/Component:            PID-3.4.1 (PATIENT_RESULT[0]/PATIENT/PID/PatientIdentifierList[0]/AssigningAuthority/NamespaceID)

Validation Error:               Field length is too long. Length of the field is exceeding the maximum length ie.,20

Expected Value(s):           Field length should not exceed 20.

 

Segment:                            MSH|1|

Field/Component:            MSH-7.1 (MSH/DateTimeOfMessage/Time)

Validation Error:               Field value does not match regular expression

Current Value(s):              20210212140141

Expected Value(s):           Format: YYYYMMDDHHMMSS[.S[S[S[S]]]]+/-ZZZZ. Note that the time zone offset is required.

 

Segment:                            MSH|1|

Field/Component:            MSH-21.1 (MSH/MessageProfileIdentifier/EntityIdentifier)

Validation Error:               Field value does not match regular expression

Current Value(s):              PHLabReportNoAck

Expected Value(s):           PHLabReport-NoAck

                                            Please refer to HL7 implementation guide for MSH-21.1 examples.

 

Segment:                            PID|1|,ORC|1|,OBR|1|,SPM|1|

Field/Component:            PID-3.4.2(PID/PatientIdentifierList/AssigningAuthority/UniversalID)

Field/Component:            ORC-2.3(ORC/PlacerOrderNumber/UniversalID)

Field/Component:            ORC-3.3(ORC/FillerOrderNumber/UniversalID)

Field/Component:            OBR-2.3(OBR/PlacerOrderNumber/UniversalID)

Field/Component:            OBR-3.3(OBR/FillerOrderNumber/UniversalID)

Field/Component:            SPM-2.1.3(SPM/PlacerOrderNumbe/UniversalID)

Field/Component:            SPM-2.2.3(SPM/FillerOrderNumber/UniversalID)

Validation Error:                Value is missing

Current Value(s):              No value is indicated

Expected Value(s):           Must be an OID

 

Segment:                            PID|1|,ORC|1|,OBR|1|,

[...]

PR #531, opened 2021-02-22
Vermont HL7 Improvements

MauriceReeves-usds:
This PR adds some additional information to the HL7 generation per requests from Vermont's DOH. The issues are outlined in #529 

## Changes
- Adds some hardcoded values to match the NIST validator
- Adds logic to add ordering facility information to each of the AOE questions
- Adds method to let us look up values in a table row by the HL7 spec associated with the element

## Checklist
- [x] Tested locally?
- [ ] Added new dependencies?
- [x] Updated the docs?
- [x] Added a test?


prime-router/metadata/schemas/covid-19.schema
```diff
@@ -730,7 +731,7 @@ elements:
   - name: message_profile_id
     type: EI
     hl7Field: MSH-21
-    default: PHLabReportNoAck^ELR_Receiver^2.16.840.1.113883.9.11^ISO
+    default: PHLabReport-NoAck^ELR_Receiver^2.16.840.1.113883.9.11^ISO
```
MauriceReeves-usds, line 734: Changed to match the regex that validates this field

prime-router/src/main/kotlin/Report.kt
```diff
@@ -295,6 +295,13 @@ class Report {
         return table.getString(row, column)
     }
 
+    fun getStringByHl7Field(row: Int, hl7Field: String): String? {
+        val column = schema.elements.filter { it.hl7Field.equals(hl7Field, ignoreCase = true) }.firstOrNull()
+            ?: return null
+        val index = schema.findElementColumn(column.name) ?: return null
+        return table.getString(row, index)
+    }
+
```
MauriceReeves-usds, line 304: Allows me to pull information from a single row by the HL7 spec. Fails safely if there's no matching field for the spec and returns null. I will probably need to create a second one that searches the `hl7OutputFields` collection, and right now this just does a `firstOrNull()` and we could probably do a check to find any with a value and return that.  Not sure.

prime-router/src/main/kotlin/serializers/Hl7Serializer.kt
```diff
@@ -441,7 +442,18 @@ class Hl7Serializer(val metadata: Metadata) {
 
         terser.set(formPathSpec("OBX-11", aoeRep), "F")
         terser.set(formPathSpec("OBX-14", aoeRep), date)
+        terser.set(formPathSpec("OBX-23-7", aoeRep), "XX")
```
MauriceReeves-usds, line 445: This just specifies that the ID type for the testing lab is of type `organization identifier`: https://hl7-definition.caristix.com/v2/HL7v2.5.1/Tables/0203

This value is apparently required.

prime-router/src/main/kotlin/serializers/Hl7Serializer.kt
```diff
@@ -469,15 +481,23 @@ class Hl7Serializer(val metadata: Metadata) {
...
+        terser.set("/PATIENT_RESULT/ORDER_OBSERVATION/OBR-3-4", "ISO")
 
         terser.set("/PATIENT_RESULT/ORDER_OBSERVATION/SPECIMEN/SPM-1", "1")
+        terser.set("/PATIENT_RESULT/ORDER_OBSERVATION/SPECIMEN/SPM-2-1-4", "ISO")
+        terser.set("/PATIENT_RESULT/ORDER_OBSERVATION/SPECIMEN/SPM-2-2-4", "ISO")
 
         terser.set("/PATIENT_RESULT/ORDER_OBSERVATION/OBSERVATION/OBX-1", "1")
         terser.set("/PATIENT_RESULT/ORDER_OBSERVATION/OBSERVATION/OBX-2", "CWE")
+        terser.set("/PATIENT_RESULT/ORDER_OBSERVATION/OBSERVATION/OBX-23-7", "XX")
     }
```
MauriceReeves-usds, line 501: A set of literals to encode as well.

MauriceReeves-usds (commented):
Adding clarifying comments on my changes.

jimduff-usds (approved):
looks good.

prime-router/src/main/kotlin/serializers/Hl7Serializer.kt
```diff
@@ -360,6 +362,30 @@ class Hl7Serializer(val metadata: Metadata) {
...
                 }
             }
+            Element.Type.EMAIL -> {
+                if (value.isNotEmpty()) {
+                    setEmailComponent(terser, value, pathSpec, element)
+                }
+            }
+            Element.Type.DATETIME -> {
+                if (element.name.equals("testing_lab_specimen_received_datetime", ignoreCase = true)) {
+                    val normalDate = try {
```
RickHawesUSDS, line 372: Confused. What is this doing?

prime-router/src/main/kotlin/serializers/Hl7Serializer.kt
```diff
@@ -360,6 +362,30 @@ class Hl7Serializer(val metadata: Metadata) {
...
+                    } catch (e: DateTimeParseException) {
+                        null
+                    } ?: try {
+                        val formatter = DateTimeFormatter.ofPattern(Element.datetimePattern, Locale.ENGLISH)
+                        OffsetDateTime.parse(value, formatter)
+                    } catch (e: DateTimeParseException) {
+                        error("Invalid date: '$value' for element '$${element.name}'")
+                    }
+                    val adjustedDateTime = normalDate.plusSeconds(6)
+                    val formattedDateTime = formatter.format(adjustedDateTime)
```
RickHawesUSDS, line 383: magic number.

prime-router/src/main/kotlin/serializers/Hl7Serializer.kt
```diff
@@ -401,11 +427,34 @@ class Hl7Serializer(val metadata: Metadata) {
...
+            if (extension.isNotEmpty()) terser.set("/PATIENT_RESULT/PATIENT/PID-13($rep)-8", extension)
+        } else {
+            terser.set(buildComponent(pathSpec, 2), "WPN")
+            terser.set(buildComponent(pathSpec, 5), country)
+            terser.set(buildComponent(pathSpec, 6), areaCode)
+            terser.set(buildComponent(pathSpec, 7), local)
+            terser.set(buildComponent(pathSpec, 8), extension)
+        }
+    }
+
```
RickHawesUSDS, line 448: Looks good.

prime-router/src/main/kotlin/serializers/Hl7Serializer.kt
RickHawesUSDS: Hmm. If we set OBX-23-7 Identifier ID type, don't we need to include OBX-23-3 Identifier ID?

prime-router/src/main/kotlin/serializers/Hl7Serializer.kt
```diff
@@ -469,15 +540,23 @@ class Hl7Serializer(val metadata: Metadata) {
...
+        terser.set("/PATIENT_RESULT/ORDER_OBSERVATION/OBR-3-4", "ISO")
 
         terser.set("/PATIENT_RESULT/ORDER_OBSERVATION/SPECIMEN/SPM-1", "1")
+        terser.set("/PATIENT_RESULT/ORDER_OBSERVATION/SPECIMEN/SPM-2-1-4", "ISO")
+        terser.set("/PATIENT_RESULT/ORDER_OBSERVATION/SPECIMEN/SPM-2-2-4", "ISO")
 
         terser.set("/PATIENT_RESULT/ORDER_OBSERVATION/OBSERVATION/OBX-1", "1")
         terser.set("/PATIENT_RESULT/ORDER_OBSERVATION/OBSERVATION/OBX-2", "CWE")
+        terser.set("/PATIENT_RESULT/ORDER_OBSERVATION/OBSERVATION/OBX-23-7", "XX")
     }
```
RickHawesUSDS, line 560: See my OBX-23-3 comment above.

RickHawesUSDS (changes requested):
Looks good, but I think there is one bug in setting OBX-23-7 and OBX-23-3.

prime-router/src/main/kotlin/serializers/Hl7Serializer.kt
MauriceReeves-usds: Some states don't care if the received date time is the same as the collected date time, but others want them offset by a few seconds, so I'm manually shifting the received date time slightly so the file will pass validation. I don't like this hack, and I need to put in a comment about the intent here. I'll remove the magic number too.

prime-router/src/main/kotlin/serializers/Hl7Serializer.kt
MauriceReeves-usds: I will verify. None of the validation has thrown this error, but it's best for me to double check.

MauriceReeves-usds: No. OBX-23-3 is old, and only maintained for backward compatibility. It is recommended instead that we use OBX-23-10, which we are.

MauriceReeves-usds:
@RickHawesUSDS - can you please rereview? For OBX-23-7, we don't need to fill in OBX-23-3 because we are using OBX-23-10. OBX-23-3 is only maintained for backward compatibility.

prime-router/src/main/kotlin/serializers/Hl7Serializer.kt
```diff
@@ -450,14 +547,14 @@ class Hl7Serializer(val metadata: Metadata) {
...
+        terser.set(formPathSpec("NTE-4-7"), HL7_SPEC_VERSION)
     }
 
     private fun setLiterals(terser: Terser) {
         // Value that NIST requires (although # is not part of 2.5.1)
         terser.set("MSH-15", "NE")
         terser.set("MSH-16", "NE")
-        terser.set("MSH-12", "2.5.1")
+        terser.set("MSH-12", HL7_SPEC_VERSION)
         terser.set("MSH-17", "USA")
```
RickHawesUSDS, line 558: I like

prime-router/src/main/kotlin/serializers/Hl7Serializer.kt
```diff
@@ -467,15 +564,23 @@ class Hl7Serializer(val metadata: Metadata) {
         terser.set("SFT-6", buildDate)
 
         terser.set("/PATIENT_RESULT/PATIENT/PID-1", "1")
+        terser.set("/PATIENT_RESULT/PATIENT/PID-3-4-3", "ISO")
```
RickHawesUSDS, line 568: I don't think this is a good idea. To add sub-field without checking if the field is present, seems like it will lead to other problems. Isn't there another way?

prime-router/settings/organizations-test.yml
```diff
@@ -172,6 +172,83 @@
...
+    - name: elr
+      organizationName: la-doh
+      topic: covid-19
+      jurisdictionalFilter: [ "matches(ordering_facility_state, LA)" ]
+      translation:
+        type: HL7
+        useBatchHeaders: true
+        receivingApplicationName: LA-ELR
+        receivingFacilityName: LADOH
+
```
RickHawesUSDS, line 189: Looks good.

prime-router/settings/organizations-test.yml
```diff
@@ -172,6 +172,83 @@
...
+      topic: covid-19
+      jurisdictionalFilter: [ "matches(ordering_facility_state, OH)" ]
+      translation:
+        type: HL7
+        useBatchHeaders: true
+        receivingApplicationName: OHDOH
+        receivingApplicationOID: 2.16.840.1.114222.4.1.3674
+        receivingFacilityName: OHDOH
+        receivingFacilityOID: 2.16.840.1.114222.4.1.3674
+
```
RickHawesUSDS, line 206: Looks good.

prime-router/settings/organizations-test.yml
```diff
@@ -172,6 +172,83 @@
...
+  stateCode: MT
+  receivers:
+    - name: elr
+      organizationName: mt-doh
+      topic: covid-19
+      jurisdictionalFilter: [ "matches(patient_state, MT)" ]
+      translation:
+        type: HL7
+        useBatchHeaders: true
+
```
RickHawesUSDS, line 236: No receivingApplication?

prime-router/settings/organizations-test.yml
```diff
@@ -172,6 +172,83 @@
...
+      jurisdictionalFilter: [ "matches(ordering_facility_state, TX)" ]
+      translation:
+        type: HL7
+        useBatchHeaders: true
+        receivingApplicationName: NEDSS
+        receivingFacilityName: TX-ELR
+
 - name: co-phd
   description: Colorado Department of Health
   jurisdiction: STATE
```
RickHawesUSDS, line 254: So, I was expecting to see a `organization-prod.yml` change. Is that coming later?

prime-router/src/main/kotlin/LookupTable.kt
```diff
@@ -62,6 +62,18 @@ class LookupTable(
...
+        return table
+            .filter { row ->
+                filters.all { (k, v) ->
+                    val filterColumnNumber = headerIndex[k.toLowerCase()] ?: error("$k doesn't exist lookup table")
+                    row[filterColumnNumber] == v
+                }
+            }
+            .map { row -> row[selectColumnNumber] }
+    }
+
```
RickHawesUSDS, line 76: what is this for? just curious.

prime-router/src/main/kotlin/serializers/Hl7Serializer.kt
```diff
@@ -360,6 +362,39 @@ class Hl7Serializer(val metadata: Metadata) {
...
                 }
             }
+            Element.Type.EMAIL -> {
+                if (value.isNotEmpty()) {
+                    setEmailComponent(terser, value, pathSpec, element)
+                }
+            }
+            Element.Type.DATETIME -> {
+                // HACK WARNING
+                // some states have complained that the specimen_collected_datetime and the
```
RickHawesUSDS, line 372: Can we do this hack with a mapper? It's place where tricky stuff should live. Not in serializers.

RickHawesUSDS (changes requested):
A lot of good work here. I'd like to talk about a couple of approaches to the problems that you trying to solve.

MauriceReeves-usds:
@RickHawesUSDS - I've made the requested changes to this PR. When you have bandwidth can you look again please?

RickHawesUSDS (approved):
Thanks. Mo. Look's good.

PR #533, opened 2021-02-22
Updates to support staging environment

cglodosky:
An additional environment module has been added to terraform for a `staging` environment.  It will essentially mimic the behaviors of the `test` environment, yet it will be isolated within the `prime-data-hub-staging` resource group in Azure.

### DIFF
--- a/operations/Dockerfile
+++ b/operations/Dockerfile
@@ -15,6 +15,7 @@ RUN yum -y install azure-cli-${AZURE_CLI_VERSION}-1.el7
 # Changing the terraform version must also be done in:
 # - app/src/environments/dev/main.tf
 # - app/src/environments/test/main.tf
+# - app/src/environments/staging/main.tf
 # - app/src/environments/prod/main.tf
 ENV TERRAFORM_VERSION=0.14.5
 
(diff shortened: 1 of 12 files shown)


operations/Dockerfile
```diff
@@ -15,6 +15,7 @@ RUN yum -y install azure-cli-${AZURE_CLI_VERSION}-1.el7
 # Changing the terraform version must also be done in:
 # - app/src/environments/dev/main.tf
 # - app/src/environments/test/main.tf
+# - app/src/environments/staging/main.tf
 # - app/src/environments/prod/main.tf
```
ronaldheft-usds, line 19: Good catch on the docs

operations/app/src/modules/function_app/main.tf
```diff
@@ -70,7 +70,7 @@ resource "azurerm_function_app" "function_app" {
 
     "OKTA_baseUrl" = "hhs-prime.okta.com"
     "OKTA_clientId" = var.okta_client_id
-    "OKTA_redirect" = (var.environment == "prod" ? "https://prime.cdc.gov/download" : "https://prime-data-hub-test.azurefd.net/api/download")
+    "OKTA_redirect" = (var.environment == "prod" ? "https://prime.cdc.gov/download" : (var.environment == "staging" ? "https://prime-data-hub-staging.azurefd.net/download" : "https://prime-data-hub-test.azurefd.net/download"))
```
ronaldheft-usds, line 73: Ah, yes, the lack of easier if/else logic in HCL. If possible, it would be cleaner to introduce a new variable and set this in the base module per environment.

ronaldheft-usds (approved):
Added one suggestion, but otherwise looks great!

PR #534, opened 2021-02-22
Add key vaults for secret management

ronaldheft-usds:
This PR adds two key vaults for secret management.

## Changes
- Adds `pdh*-appconfig` key vault
- Adds `pdh*-clientconfig` key vault
- Grants dev access to both key vaults

## Checklist
- [x] Tested locally?
- [ ] Added new dependencies?
- [ ] Updated the docs?
- [ ] Added a test?

## Fixes 

- #issue

## To Be Done

- After this PR rolls out, we need to populate the key vaults with secrets.
- Follow on PRs will switch our apps and configs to use the key vaults.

### DIFF
--- a/operations/app/src/modules/key_vault/main.tf
+++ b/operations/app/src/modules/key_vault/main.tf
@@ -95,6 +95,96 @@ resource "azurerm_key_vault_access_policy" "frontdoor_access_policy" {
   certificate_permissions = [ "Get" ]
 }
 
+resource "azurerm_key_vault" "app_config" {
+  name = "${var.resource_prefix}-appconfig" # Does not include "-keyvault" due to char limits (24)
+  location = var.location
+  resource_group_name = var.resource_group
+  sku_name = "premium"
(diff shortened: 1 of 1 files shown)


cglodosky:
Is there some limitation that forces us into using three key vaults per environment as opposed to one key vault with more fine-grained access policies?

ronaldheft-usds:
@cglodosky Great question. I've been unable to find a way to grant access to secrets on a per secret level or any type of organization structure that can be utilized within Key Vaults. Access can be granted to various Key Vault services (like keys, certificates, etc) and operations (read, write, decrypt, etc), but not to individual objects within a Key Vault.

The issue is our client secrets should eventually be secured that we can't even see them, so that needs to be a separate vault. App secrets could potentially be held under the same vault that contains our HTTPS cert and encryption keys, but I'm not sure what else we will store in there in the future, so I thought it was cleaner to have a dedicated Key Vault for app configuration as well.

cglodosky (approved):

Issue #532, opened 2021-02-22
Add New States to organizations


MauriceReeves-usds:
Need to add the following states to the test organizations yaml file and verify the information is correct:

- Vermont
- New Mexico
- North Dakota
- Florida
- Louisiana
- Texas

jlusds:
Was this done?

MauriceReeves-usds:
Yup. Let's mark it done.

PR #536, opened 2021-02-22
Adds new facilities into organizations yaml

MauriceReeves-usds:
This PR adds the most recent states we've been working with to the test yaml file.

## Checklist
- [ ] Tested locally?
- [ ] Added new dependencies?
- [ ] Updated the docs?
- [ ] Added a test?

## Fixes 

- #532

### DIFF
--- a/prime-router/metadata/organizations-test.yml
+++ b/prime-router/metadata/organizations-test.yml
@@ -113,19 +113,118 @@
     services:
       - name: elr
         topic: covid-19
-        schema: covid-19
+        schema: fl/fl-covid-19
         jurisdictionalFilter: [ "matches(ordering_facility_state, FL)" ]
         transforms: { deidentify: false }
         format: HL7_BATCH
+        batch:
+          operation: MERGE
+          numberPerDay: 96 # Every 15 minutes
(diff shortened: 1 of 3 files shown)


prime-router/metadata/organizations-test.yml
```diff
@@ -113,19 +113,118 @@
...
+          timeZone: CENTRAL
+
+  - name: la-doh
+    description: Louisiana Department of Health
+    services:
+      - name: elr
+        topic: covid-19
+        schema: la/la-covid-19
+        jurisdictionalFilter: [ "matches(ordering_facility_state, LA)" ]
+        format: HL7_BATCH
```
jimduff-usds, line 204: It just occurred to me:  are we offering the option to states, to receiving results from either the facility_state matching OR the person's address state matching?     It looks like all of our states are only matching on facility_state, which is surprising.    The only exception so far is Pima - the `filterByCounty` function does an either/or match.

jimduff-usds (approved):
Great to see these states moving toward production.  One non-tech question, inline.

MauriceReeves-usds:
I think this is done. We can move this to closed.

PR #537, opened 2021-02-22
Migrate app secrets to Key Vault

ronaldheft-usds:
This PR utilizes our `pdh*-appsecret` Key Vault for various configuration secrets. Client PHD connection secrets are handles in a separate PR.

## Changes
- Switches our database Terraform module to pull in secrets from our Key Vault to configure the database.
- Switches our metabase connection string to use the database user/pass from Key Vault.
- Switches the function app to consume the secrets directly from the Key Vault.
- Creates an access policy for the function app to use the Key Vault.
- Removes all Terraform variable references to the old secrets.

## Checklist
- [x] Tested locally?
- [ ] Added new dependencies?
- [ ] Updated the docs?
- [ ] Added a test?

## To Be Done

- This PR requires an update when $533 is merged.
- Before this can be rolled out, #534 must be rolled out, and the following secrets add to the `*-appconfig` Key Vault in each environment:
  - `functionapp-postgres-user`
  - `functionapp-postgres-pass`
  - `functionapp-okta-client-id`
  - `functionapp-redox-secret`
  - `functionapp-az-phd-user` (A future PR will remove the need for this to be configured, and will instead use our client secrets storage.)
  - `functionapp-az-phd-pass` (A future PR will remove the need for this to be configured, and will instead use our client secrets storage.)


operations/app/src/modules/function_app/main.tf
```diff
@@ -96,12 +96,12 @@ resource "azurerm_function_app" "function_app" {
...
-    "PRIME__CSV__USER" = (var.environment != "prod" ? var.az_phd_user : null)
-    "PRIME__CSV__PASS" = (var.environment != "prod" ? var.az_phd_password : null)
+    "PRIME__HL7__USER" = (var.environment != "prod" ? "@Microsoft.KeyVault(VaultName=${var.resource_prefix}-appconfig;SecretName=functionapp-az-phd-user)" : null)
+    "PRIME__HL7__PASS" = (var.environment != "prod" ? "@Microsoft.KeyVault(VaultName=${var.resource_prefix}-appconfig;SecretName=functionapp-az-phd-pass)" : null)
+    "PRIME__HL7_BATCH__USER" = (var.environment != "prod" ? "@Microsoft.KeyVault(VaultName=${var.resource_prefix}-appconfig;SecretName=functionapp-az-phd-user)" : null)
+    "PRIME__HL7_BATCH__PASS" = (var.environment != "prod" ? "@Microsoft.KeyVault(VaultName=${var.resource_prefix}-appconfig;SecretName=functionapp-az-phd-pass)" : null)
+    "PRIME__CSV__USER" = (var.environment != "prod" ? "@Microsoft.KeyVault(VaultName=${var.resource_prefix}-appconfig;SecretName=functionapp-az-phd-user)" : null)
+    "PRIME__CSV__PASS" = (var.environment != "prod" ? "@Microsoft.KeyVault(VaultName=${var.resource_prefix}-appconfig;SecretName=functionapp-az-phd-pass)" : null)
```
cglodosky, line 104: Is this syntax required for either `app_settings` or because it is within a conditional?  Just curious since the `database` module accesses secrets as data sources.

cglodosky (approved):
I just put one comment down as a question.  Looks good otherwise!  It will be nice to have these get pulled on-the-fly instead of as environment variables.

operations/app/src/modules/function_app/main.tf
ronaldheft-usds: This is the syntax for a function's configuration to pull it's values from the Key Vault directly. `OKTA_clientId` has a cleaner example without the conditional.

We could follow the method used for the database user/pass, but technically it's less secure as the decrypted values have to pulled down into the Terraform state. Use the `@Microsoft.KeyVault` syntax for the function app removes a hop from the process.

cglodosky: Ok, that makes sense.  Have we given the function app read access to secrets?

PR #545, opened 2021-02-24
Terraform Updates For App Settings and Metabase

cglodosky:
- Add new app settings for test/staging environments.  I will create a future PR for merging app settings as opposed to all of the conditionals.  
- Remove metabase resources for non-test environments.

### DIFF
--- a/operations/app/src/modules/front_door/main.tf
+++ b/operations/app/src/modules/front_door/main.tf
@@ -5,7 +5,7 @@ terraform {
 locals {
     name = var.environment != "dev" ? "prime-data-hub-${var.environment}" : "prime-data-hub-${var.resource_prefix}"
     functionapp_address = "${var.resource_prefix}-functionapp.azurewebsites.net"
-    metabase_address = "${var.resource_prefix}-metabase.azurewebsites.net"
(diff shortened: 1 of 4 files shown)


operations/app/src/modules/front_door/main.tf
```diff
@@ -85,30 +94,24 @@ resource "azurerm_frontdoor" "front_door" {
...
+          
+          custom_https_configuration {
+            certificate_source = "AzureKeyVault"
+            azure_key_vault_certificate_secret_name = var.https_cert_name
+            azure_key_vault_certificate_secret_version = data.azurerm_key_vault_secret.https_cert[0].version
+            azure_key_vault_certificate_vault_id = var.key_vault_id
+          }
         }
-    }
+      }
```
ronaldheft-usds, line 108: I think we have inconsistent IDE settings for Terraform files, as this looks like an auto-formatting change. I'm going to propose an editorconfig for *.tf files in an upcoming PR so we don't end up going back and forth adjusting formatting between our environments

ronaldheft-usds (approved):
Looks good!

PR #557, opened 2021-02-26
Enable PostgreSQL Advanced Threat Protection

cglodosky:
This PR enables Advanced Threat Protection for our PostgreSQL servers.

### DIFF
--- a/operations/app/src/modules/database/main.tf
+++ b/operations/app/src/modules/database/main.tf
@@ -18,6 +18,11 @@ resource "azurerm_postgresql_server" "postgres_server" {
   ssl_enforcement_enabled = true
   ssl_minimal_tls_version_enforced = "TLS1_2"
 
+  threat_detection_policy {
+    enabled = true
+    email_account_admins = true
+  }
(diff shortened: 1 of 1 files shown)


operations/app/src/modules/database/main.tf
```diff
@@ -18,6 +18,11 @@ resource "azurerm_postgresql_server" "postgres_server" {
   ssl_enforcement_enabled = true
   ssl_minimal_tls_version_enforced = "TLS1_2"
 
+  threat_detection_policy {
+    enabled = true
+    email_account_admins = true
```
cglodosky, line 23: This configuration will send alerts to `Azure Security Center`.  Based on terraform's docs, it can also be configured to be sent to a storage account if we need to do that instead.

ronaldheft-usds: This looks good. Emailing is great start, and we can integrate with Splunk if there is a desire from that team.

ronaldheft-usds (approved):

PR #558, opened 2021-02-26
Provision Metabase in Production

cglodosky:
This PR provisions Metabase within the production environment.  Because the production environment uses an Elastic Premium service plan for the function apps, a separate service plan is required so that Metabase can be provisioned.  The Elastic Premium option is only available for Function Apps.


ronaldheft-usds (approved):
This looks good. So is the plan to run our metabase out of prod now instead of test?

cglodosky:
At this time, we are going to have a metabase instance for both test and prod - each with its own connection to the respective database.

PR #559, opened 2021-02-26
Keep Metabase Front Door configuration for prod environment

cglodosky:
This PR keeps the metabase-specific configurations within the Front Door in the production environment.


ronaldheft-usds (approved):
LGTM

Issue #562, opened 2021-02-26
Subtle bug:  Hanging redox send appears to hang unrelated threads waiting on db lock


jimduff-usds:
Found a	subtle bug today:  By chance, my redox transport is misconfigured in Stage, such that RedoxTransport.kt hangs waiting, then eventually fails.

But that's not the bug.   That redox send was one of four Sends running simultaneously, all 3 others are sftp Send.   **But all four sends hung waiting!**   Not 100% certain, but I'm thinking this is because the Redox send happens inside a txn that locks on the Task table.  Once the redox fails, the three sftps suddenly awaken and run successfully to completion.

Here's a snippet of the exception in STAGING today, during an end2end run, during the Send step:

2021-02-26T14:32:06.613245555Z: [INFO]  [41m[30mfail[39m[22m[49m: Function.send.User[0]
2021-02-26T14:32:06.613264155Z: [INFO]        Send exception
2021-02-26T14:32:06.613269455Z: [INFO]  Microsoft.Azure.WebJobs.Script.Workers.Rpc.RpcException: Result: Send exception
2021-02-26T14:32:06.613274155Z: [INFO]  Exception: Rollback caused
2021-02-26T14:32:06.613278555Z: [INFO]  Stack: org.jooq.exception.DataAccessException: Rollback caused
2021-02-26T14:32:06.615165263Z: [INFO]         at java.base/java.lang.Thread.run(Unknown Source)
2021-02-26T14:32:06.615169763Z: [INFO]  Caused by: java.net.MalformedURLException: no protocol: /some_url/auth/authenticate
2021-02-26T14:32:06.615914167Z: [INFO]         at gov.cdc.prime.router.transport.RedoxTransport.fetchToken(RedoxTransport.kt:126)
2021-02-26T14:32:06.615918467Z: [INFO]         at gov.cdc.prime.router.transport.RedoxTransport.send(RedoxTransport.kt:46)
2021-02-26T14:32:06.615922767Z: [INFO]         at gov.cdc.prime.router.azure.SendFunction$run$1.invoke(SendFunction.kt:80)

The more likely worrisome scenario in production is that an SFTP might stall, and other paralle, sends would stall as well.    Still, even there it might only cause a minute or so delay, unless the other threads simple gave up and assumed there was a permanent deadlock.

Issue #502, opened 2021-02-16
Add the `ignore` Organization  into production, to use for production sanity tests


jimduff-usds:
Its not good enough that we depend on ourselves to remember to use the `option=SkipSend`

jimduff-usds:
The `'/prime test --env staging` test harness is now set up and running well in Staging.   Once the locked-down database access is done, and well tested, we can safely roll `./prime test --env prod` too.

PR #565, opened 2021-02-27
General Hardening

jimduff-usds:
This PR does a variety of hardening work, listed under changes.

_You should try out the new endpoint tests_.   First make sure your database env vars are set up.  Run
`./prime test --help`
to get help with setting up env vars.   Then start up your local `docker-compose up` and run
`./prime test --run ping,end2end`
Then simply run
`./prime test`
to run all the tests locally.   This will take about 10 mins.   As of this writing, the `strac` test fails (because it should, IMO), but all the other tests, including `stracbasic`, should pass.

## Changes
- api/reports endpoint now checks and returns failure on payloads with
-- too many items
-- too many columns
-- too many errors, or 
-- that are just plain too big.
- Added new command line tests.  
- `./prime test --list`   
```
Available options to ./prime test --run <test1,test2> are:
ping                	Is the reports endpoint alive and listening?
end2end             	Create Fake data, submit, wait, confirm sent via database lineage data
merge               	Submit multiple files, wait, confirm via db that merge occurred
huge                	Submit 10000 line csv file, wait, confirm via db.  Slow.
toobig              	Submit 10001 lines, which should be an error.  Slower ;)
toomanycols         	Submit a file with more than 2000 columns, which should error
stracbasic          	Basic strac test to REDOX only.
strac               	Submit data in the strac schema format, wait, confirm via database queries
```
- `./prime test` can run with `--env` set to local, test, or staging. 
- Removed (very unlikely) possible sql injection problem with LIKE query during Download.
- Implemented creation of sha256 digests on internal blobs.
- Removed the `TASK_SOURCE` table and related code.

## Checklist
- [X] Tested locally?
- [ ] Added new dependencies?
- [X] Updated the docs?
- [X] Added a test?

## Fixes 

- #313 
- #494 
- #502 
- Portions of #495 

## To Be Done

- strac test always fails - due to missing data fields needed by output schemas, not currently in strac schema.   Issue is worth highlighting, as it will bite us if strac expands to states that currently take simple report data.
- compare actual to expected sha256 digests when reading in blobs form bookstore.
- did not implement api/reports validation check on each individual row max bytes size.   I think the combination of the other checks is sufficient.
- Do we also want to remove Sources in Report.kt?  I don't see a clear use case for it.
- #562 
- `TASK` table should remain as our internal "todo list", but should remove columns redundant with `REPORT_FILE`.  Then I can remove those annoying sanity checks too.
- main.kt cli needs refactoring to use the methods in FileUtilities.kt


prime-router/src/main/kotlin/azure/DownloadFunction.kt
```diff
@@ -88,63 +89,62 @@ class DownloadFunction(private val workflowEngine: WorkflowEngine = WorkflowEngi
...
     private fun responsePage(
         request: HttpRequestMessage<String?>,
-        authClaims: AuthClaims,
-        context: ExecutionContext
+        authClaims: AuthClaims
     ): HttpResponseMessage {
         val htmlTemplate: String = Files.readString(Path.of(DOWNLOAD_PAGE))
-        val headers = workflowEngine.fetchDownloadableHeaders(
+        val reportFiles = workflowEngine.fetchDownloadableReportFiles(
             OffsetDateTime.now().minusDays(DAYS_TO_SHOW), authClaims.organization.name
```
jimduff-usds, line 140: @mry-usds is right - you don't want to pull in the whole Header for the list of files, because it now includes the onerous ItemLineage stuff (which could possibly be lazy loaded, but is not at this time).  The ReportFile table by design contains everything you need to know about a report - so just pull in that.

prime-router/src/main/kotlin/azure/HttpUtilities.kt
```diff
@@ -4,8 +4,19 @@ import com.google.common.net.HttpHeaders
...
+import java.io.IOException
+import java.net.HttpURLConnection
+import java.net.URL
 import java.time.OffsetDateTime
 
+enum class ReportStreamEnv(val endPoint: String) {
+    TEST("https://pdhtest-functionapp.azurewebsites.net/api/reports"),
+    LOCAL("http://localhost:7071/api/reports"),
+    STAGING("https://pdhstaging-functionapp.azurewebsites.net/api/reports")
+}
```
jimduff-usds, line 18: I created this little enum in support of the new TestReportStream, but then put it here since it might be generally useful.

prime-router/src/main/kotlin/cli/FileUtilities.kt
```diff
@@ -0,0 +1,117 @@
+package gov.cdc.prime.router.cli
+
```
jimduff-usds, line 2: I created this utilities class to parallel HttpUtilities.

There is a refactoring task not yet done:   I didn't fix the file creation code in ProcessData in the cli to use these new methods yet.  :(

prime-router/src/main/kotlin/serializers/CsvSerializer.kt
```diff
@@ -53,9 +56,35 @@ class CsvSerializer(val metadata: Metadata) {
         defaultValues: Map<String, String> = emptyMap(),
     ): ReadResult {
         val schema = metadata.findSchema(schemaName) ?: error("Internal Error: invalid schema name '$schemaName'")
-        val rows: List<Map<String, String>> = csvReader().readAllWithHeader(input)
         val errors = mutableListOf<ResultDetail>()
         val warnings = mutableListOf<ResultDetail>()
+        var rows = mutableListOf<Map<String, String>>()
+        csvReader().open(input) {
+            readAllWithHeaderAsSequence().forEach { row: Map<String, String> ->
```
jimduff-usds, line 63: Checking number of items is serializer-specific.

So there's a flaw here - someone has to "remember" to implement this check for HL7, etc, as we create new formats.  Better would be to extract this up into an Interface method, to at least act as a reminder of required functionality.

prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -190,7 +188,7 @@ class ActionHistory {
     /**
      * Use this to record history info about a new externally submitted report.
      */
-    fun trackExternalInputReport(incomingReport: ReportFunction.ValidatedRequest) {
+    fun trackExternalInputReport(incomingReport: ReportFunction.ValidatedRequest, blobInfo: BlobAccess.BlobInfo) {
         val report = incomingReport.report ?: error("No report to track!")
```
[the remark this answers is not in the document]
MauriceReeves-usds: Python and C# would both like a word.

prime-router/src/main/kotlin/cli/FileUtilities.kt
MauriceReeves-usds: I think there's a few places we should abstract out some logic into interfaces. This might be one of those.

MauriceReeves-usds (approved):

PR #582, opened 2021-03-01
Hotfix: Add Key Vault access policies for the function app

ronaldheft-usds:
- Adds missing access policies for the Key Vaults / Function App
- Note: There is a bug in the Terraform module for function apps. I have a workaround, but it requires two deploys. See my comment in the code for more info.

### DIFF
--- a/operations/app/src/modules/function_app/main.tf
+++ b/operations/app/src/modules/function_app/main.tf
@@ -111,9 +111,39 @@ resource "azurerm_function_app" "function_app" {
     "WEBSITE_HTTPLOGGING_RETENTION_DAYS" = (var.environment == "prod" ? 3 : null)
   }
 
+  identity {
+    type = "SystemAssigned"
+  }
+
   tags = {
     environment = var.environment
   }
+}
+
+resource "azurerm_key_vault_access_policy" "functionapp_app_config_access_policy" {
+  # This is a hack. The function_app module has a bug where it does not export the values until after being updated.
+  # By using a count, we workout the bug by running two deploy. The first deploy created the system-assigned identity.
+  # The second deploy adds the Key Value access policy.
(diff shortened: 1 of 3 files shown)


operations/app/src/modules/function_app/main.tf
```diff
@@ -111,11 +111,41 @@ resource "azurerm_function_app" "function_app" {
...
   tags = {
     environment = var.environment
   }
 }
 
+resource "azurerm_key_vault_access_policy" "functionapp_app_config_access_policy" {
+  # This is a hack. The function_app module has a bug where it does not export the values until after being updated.
+  # By using a count, we workout the bug by running two deploy. The first deploy created the system-assigned identity.
+  # The second deploy adds the Key Value access policy.
+  count = azurerm_function_app.function_app.identity.0.tenant_id != null ? 1 : 0
```
cglodosky, line 127: Would we be able to use the `depends_on` or `wait_for_write` arguments here to avoid two deploys?  Or by "after being updated" do you mean that the entire plan has to be completed?

ronaldheft-usds: Tried a number of different alternative syntax options this morning to force the dependency, and no luck. Without this workaround, I receive the following error on the plan:

```
Error: "object_id": required field is not set

  on ../../modules/function_app/main.tf line 123, in resource "azurerm_key_vault_access_policy" "functionapp_app_config_access_policy":
 123: resource "azurerm_key_vault_access_policy" "functionapp_app_config_access_policy" {
```

The issue is the identity block's output is null during the plan. The way `azurerm_front_door` implements the identity block seems to be doing something that breaks dependencies. The system principal is not actually a resource on the front door, and in Azure is created differently, but Terraform kind of threw it in there to make the syntax cleaner.

cglodosky (approved):

PR #587, opened 2021-03-03
Add App Settings to Function App for New States

cglodosky:
This PR adds app settings to the Function App so that credentials can be pulled for FL, LA, ND, OH, and TX for HL7 SFTP transfer.


ronaldheft-usds (approved):
Terraform configuration looks good. I also verified all of the secrets exist in production. However, the secrets do not exist in staging.

I am approving the PR because the code is correct, but we have additional configuration that is needed. I would actually love to see what happens if we deploy this without the values in our Key Vault. I'm curious if we'll get an error during deployment, the function app will fail to start, or the function app will gracefully ignore it.

PR #596, opened 2021-03-03
Discrete GitHub Action Builds

### LINKED ISSUE #457: PRs never run CI when only doc updates, and this fail PR checks
Since our GitHub action has an exclude for the docs folder, if a PR only contains doc changes, GitHub actions never run. This can result in a PR that looks like it's stuck or failing the CI tests. For example: https://github.com/CDCgov/prime-data-hub/pull/455

The build will never run, and thus the PR will never show as safe to merge. While it can be merged by an admin, it's messy and we can fix by having the workflow run by skip the actual work (to safe CI cycles and give faster results).

ronaldheft-usds:
This PR improves our GitHub actions.

- Fixes #457
- Builds only run when files in a directory they care about are modified
  - This means ops PRs will not have failing tests because we didn't update the docs directory (yay!)
- Breaks our actions into separate flows to enable the above, and also sets us up for a GitHub action for terraform
- Renames the actions to descriptive names

### DIFF
--- a/.github/workflows/build_frontend.yaml
+++ b/.github/workflows/build_frontend.yaml
@@ -1,0 +1,41 @@
+name: Frontend Build
+
+on:
+  pull_request:
+    branches: [ master, production ]
+    paths-ignore:
+      - 'prime-router/docs/**'
+
+jobs:
+  gate_workflow:
+    # Do not run on PRs from forks to prevent workflow abuse.
+    # NOTE: changing this condition *may* require adjusting secrets usage in
+    # the workflow steps below.
+    if: github.event.pull_request.head.repo.full_name == github.repository
+    runs-on: ubuntu-latest
+
+    steps:
+      - run: echo 'Build gate PASSED. Continuing with other build steps.'
+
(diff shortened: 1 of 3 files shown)


ronaldheft-usds:
The hung frontend build on this PR is a result of an earlier bug that ran the front-end build when not needed. This will not occur on future branches and when merged.

.github/workflows/build_frontend.yaml
```diff
@@ -0,0 +1,41 @@
...
+      - 'frontend/**'
+
+jobs:
+  gate_workflow:
+    # Do not run on PRs from forks to prevent workflow abuse.
+    # NOTE: changing this condition *may* require adjusting secrets usage in
+    # the workflow steps below.
+    if: github.event.pull_request.head.repo.full_name == github.repository
+    runs-on: ubuntu-latest
+
```
RickHawesUSDS, line 16: Nice.

RickHawesUSDS (approved):
Looks good.

PR #603, opened 2021-03-04
Remove old schemas with generate-docs

ronaldheft-usds:
We've been getting GitHub action failures around updating docs.

Tracked it down, and the issue is when we delete a schema, it remains in our repository, but the GitHub action isn't able to regenerate it, and thus complains there are missing docs.

I fixed this issue by changing our `generate-docs` CLI to delete all the schemas before it generates, thus ensuring all documentation is correct.

This PR also removes the troublesome schema docs.


MauriceReeves-usds (approved):

RickHawesUSDS (commented):
Thank you.

PR #608, opened 2021-03-05
Remove setting env vars manually

cglodosky:
Environment variables were manually being set using the Azure CLI via stored GitHub secrets.  This is no longer desired as the secrets are now in Azure Key Vault and their reference is derived by terraform.  Removing this code prevents secrets being reverted from Key Vault references to their original values stored in GitHub.  Once this PR is merged, these secrets will be purged from GitHub.


ronaldheft-usds (approved):
Good catch! I noticed some of the values were reverted in staging when doing a plan yesterday. This makes sense why. Approved.

PR #609, opened 2021-03-05
Update Azure App Service Plan

cglodosky:
In order to host both function apps and app services alongside one another, the app service plan must be updated from an `Elastic Premium` SKU to a `Premium` SKU.  Only function apps are supported within the `Elastic Premium` SKU.  This would require any additional app services to be hosted within a separate region of Azure so that they can be contained within the same resource group.


ronaldheft-usds (approved):
This looks good. We'll have to tune the auto-scaling in the future, but this will get us off and running.

PR #637, opened 2021-03-11
Revert Service Plan Name

cglodosky:
Renaming an app service plan requires the destruction and re-provisioning of it.  I am reverting this so that it can be avoided in the environments where it is not necessary.


ronaldheft-usds (approved):
Good call.

Issue #631, opened 2021-03-11
A6 aftermath: Fix ReportStream workflow issues


jimduff-usds:

jimduff-usds:
- [ ] Bug:  there was a 500 error in production report API, and nothing went into the Action table.
- [ ] Similar BUG: It appears that exceptions during Download alos fail to put a send_error row in the Action table.
- [ ] Possible bug:  incoming reports are listed as body_format == INTERNAL - are we saving exactly what is sent us, or something different?
- [ ] SkipInvalidItems doesn't distinguish between REPORT/PARAMETER, and ITEM level errors.  SKipInvalidItems not always quite handled properly.
- [ ] Add workflow related tests to cli/TestReportStream.kt:
-- Add a test that tests garbled headers
-- Test SkipInvalidItems returns a 201 with ITEM level errors
--Test with no SkipInvalidItems returns a 400 with ITEM level errors
-  [ ] USE THIS when running `./prime test`: Flyway has an env variable so that it doesn't try to run the configuration.
- [ ] We really need a good way to note Items that go nowhere, and why

PR #640, opened 2021-03-12
Demo Cleanup first pass

jimduff-usds:
This PR does a number of improvements of issues found after the demo earlier this week.

## Changes
- [X[ Modified SendFunction to log even on major exceptions.
- [X] Download mime type was hardcoded to csv.  Fixed.
- [X] Download uses the Task table, which caused failures when Send failed.  Fixed to use REPORT_FILE table.
- [X] NullTransport was never actually called.  Fixed.
- [X] Have a very simple smoke test.  Now you can (and should!) start your docker container and call `./prime test` and it'll run a localhost smoke test!  Takes just over 5 minutes.
- [X] Ensure that running `./prime test` does not do Flyway migrations. 

## Checklist
- [X] Tested locally?
- [X] Ran `quickTest all`?
- [X] Downloaded a file from http://localhost:7071/api/download
- [X] Added tests?

## Security
- None known

## Fixes
- Partial #631 

## To Be Done
- The rest of #631 
-


.github/pull_request_template.md
```diff
@@ -7,6 +7,7 @@ This PR ...
 ## Checklist
 - [ ] Tested locally?
 - [ ] Ran `quickTest all`?
+- [ ] Ran `./prime test` against local Docker ReportStream container?
```
jimduff-usds, line 10: I think `./prime test` is now easy to use, and can become part of our canon of tests.

prime-router/src/main/kotlin/Report.kt
```diff
@@ -39,30 +39,26 @@ const val REPORT_MAX_ERRORS = 100
...
+        val ext: String,
+        val mimeType: String,
+        val isSingleItemFormat: Boolean = false,
+    ) {
+        INTERNAL("internal", "text/csv"), // A format that serializes all elements of a Report.kt (in CSV)
+        CSV("csv", "text/csv"), // A CSV format the follows the csvFields
+        HL7("hl7", "text/hl7", true), // HL7 with one result per file
+        HL7_BATCH("hl7", "text/hl7"), // HL7 with BHS and FHS headers
+        REDOX("redox", "text/json", true); // Redox format
         // FHIR
```
jimduff-usds, line 52: Added mime types here, and some other cleanup.

prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -137,6 +137,7 @@ class ActionHistory {
      * Always appends, to allow for actions that do a mix of work (eg, SEND)
      */
     fun trackActionParams(actionParams: String) {
+        if (actionParams.isNullOrEmpty()) return
```
jimduff-usds, line 140: Noticed this bug during testing.

prime-router/src/main/kotlin/azure/DatabaseAccess.kt
```diff
@@ -166,11 +164,20 @@ class DatabaseAccess(private val create: DSLContext) : Logging {
      * ActionHistory queries
      */
 
-    fun fetchReportFile(reportId: ReportId, txn: DataAccessTransaction? = null): ReportFile {
+    /**
+     * You should include org as a search criteria to enforce authorization to get that report.
+     */
+    fun fetchReportFile(reportId: ReportId, org: Organization? = null, txn: DataAccessTransaction? = null): ReportFile {
```
jimduff-usds, line 170: This addition is to support DownloadFunction using REPORT_FILE rather than TASK to get the file to download.

prime-router/src/main/kotlin/azure/DatabaseAccess.kt
```diff
@@ -465,6 +472,11 @@ class DatabaseAccess(private val create: DSLContext) : Logging {
      */
 
     companion object {
+        /**
+         * Global var.  Set to false prior to the lazy init, to prevent flyway migrations
+         */
+        var isFlywayMigrationOK = true
```
jimduff-usds, line 478: Please look at this carefully.   We had a dangerous bug where `./prime test --env staging`, run locally on a laptop, caused Flyway to push unwanted migrations onto staging.   This occurred because `./prime test` needs to access the staging database to check results.

My solution here is a global variable always set to true, but then set to false only by TestReportStream, prior to the lazy init of the database connectivity.

prime-router/src/main/kotlin/azure/DownloadFunction.kt
```diff
@@ -169,19 +169,12 @@ class DownloadFunction(private val workflowEngine: WorkflowEngine = WorkflowEngi
...
                 val filename = Report.formExternalFilename(header)
-                // todo replace with Report.Format.toMimeType
-                val mimeType = when (header.reportFile.bodyFormat) {
-                    Report.Format.CSV.name -> "text/csv"
-                    Report.Format.HL7.name -> "text/hl7"
-                    Report.Format.HL7_BATCH.name -> "text/hl7"
-                    Report.Format.INTERNAL.name -> "text/csv"
-                    else -> "text/plain"
-                }
+                val mimeType = Report.Format.safeValueOf(header.reportFile.bodyFormat).mimeType
```
jimduff-usds, line 177: refactor to clean up mime types.

prime-router/src/main/kotlin/azure/SendFunction.kt
```diff
@@ -33,16 +33,16 @@ class SendFunction(private val workflowEngine: WorkflowEngine = WorkflowEngine()
         message: String,
         context: ExecutionContext,
     ) {
+        val actionHistory = ActionHistory(TaskAction.send, context)
+        actionHistory.trackActionParams(message)
         try {
```
jimduff-usds, line 38: This move of the actionHistory code is part of overall changes to SendFunction so that the actionHistory gets stored in the database, even on major Exceptions.

prime-router/src/main/kotlin/azure/SendFunction.kt
```diff
@@ -100,9 +100,15 @@ class SendFunction(private val workflowEngine: WorkflowEngine = WorkflowEngine()
...
         } catch (t: Throwable) {
-            context.logger.log(Level.SEVERE, "Send function exception for event: $message", t)
+            // For debugging and auditing purposes
+            val msg = "Send function exception for event: $message"
+            context.logger.log(Level.SEVERE, msg, t)
+            actionHistory.setActionType(TaskAction.send_error)
+            actionHistory.trackActionResult(msg)
+        } finally {
+            // Note this is operating in a different transaction than the one that did the fetch/lock of the report
+            workflowEngine.recordAction(actionHistory)
```
jimduff-usds, line 111: Doing this in the "finally" is the key change.

prime-router/src/main/kotlin/azure/WorkflowEngine.kt
```diff
@@ -142,7 +141,6 @@ class WorkflowEngine(
                 retryJson,
                 txn
             )
-            recordAction(actionHistory, txn)
         }
```
jimduff-usds, line 144: A nice side effect of changing actionHistory in SendFunction is that it was no longer needed down here.

prime-router/src/main/kotlin/azure/WorkflowEngine.kt
```diff
@@ -300,11 +298,14 @@ class WorkflowEngine(
 
     fun fetchHeader(
         reportId: ReportId,
-        orgName: String
+        organization: Organization,
```
jimduff-usds, line 301: A simple refactor.  Objects are better.

prime-router/src/test/kotlin/azure/SendFunctionTests.kt
```diff
@@ -95,13 +95,14 @@ class SendFunctionTests {
         var nextEvent: ReportEvent? = null
         setupLogger()
         setupWorkflow()
-        every { workflowEngine.handleReportEvent(any(), any(), any()) }.answers {
-            val block = thirdArg() as
+        every { workflowEngine.handleReportEvent(any(), any()) }.answers {
+            val block = secondArg() as
```
jimduff-usds, line 99: OK, I admit it publicly:  I find mockk to be the single most inscrutable thing in our codebase.

prime-router/src/main/kotlin/cli/TestReportStream.kt
```diff
@@ -91,25 +98,55 @@ Examples:
...
+        }
+        if (problem) {
+            echo("Error: --env is $env but database is set to $dbEnv")
+            exitProcess(-1)
         }
     }
 
     override fun run() {
-        if (list) doList()
+        DatabaseAccess.isFlywayMigrationOK = false
```
jimduff-usds, line 143: Turn off flyway migrations - see my comment above.  Please check my idea carefully - can this go wrong somehow?

prime-router/src/main/kotlin/cli/TestReportStream.kt
```diff
@@ -49,39 +49,46 @@ export POSTGRES_PASSWORD=<SECRET>
 Examples:
 ```
  ./prime test --list        List detailed information about available tests
- ./prime test               Runs all the tests on localhost
+ ./prime test               Runs the set of tests labelled as '${TestStatus.GOODSTUFF.description}'
```
jimduff-usds, line 52: I did gobs and gobs of little cleanup work in this class.

carlosfelix2 (approved):
I have very little context into the code, but saw nothing obvious to me

prime-router/src/main/kotlin/cli/TestReportStream.kt
ronaldheft-usds: This appears correct to me for this particular use case.

If we want to cover the situation more generally (we write a new cli class and forget about the need to disable migrations), I think the default should be disabled, with enabled being turned on via an environment variable, and if no environment variable, defaulting to false.

For even more protection, the environment variable could be something like `FLYWAY_MIGRATION_ENVIRONMENT_ENABLED=local` and the environment variable would have to match the name set in `PRIME_ENVIRONMENT`. This way if we accidentally run a config mixing environments, there is an extra layer of protection.

Obviously, the best solution here is to ensure testing is always done with read-only accounts, but I am all for adding multiple layers of protection.

ronaldheft-usds (approved):
Everything looks good. Added an additional idea for extra security around Flyway migrations, but I'm okay with the solution as is.

prime-router/src/main/kotlin/cli/TestReportStream.kt
jimduff-usds: Thanks, Ron.  I was reluctant to default false at the moment, because the flyway migrations work so well 99.9% of the time.  We won't enable this in prod until @cglodosky's database hardening is done.

Issue #634, opened 2021-03-11
A6 aftermath: sftp problems: Need a way to test "last mile" problems from within Azure


jimduff-usds:

jimduff-usds:
- Need a way to confirm that our sftp connectivity works in production.
- We discussed having a separate HealthCheck Azure Function that would know how to ping sftp sites, etc.
- Dunno if this is useful, but Rick says ApplicationInsights has a way to trigger a healthcheckevent in an Azure function.
- This should be expandable to other hard-to-reach tests in the future, eg, is Redox working?

jimduff-usds:
The first pass of this feature was implemented as follows:

A New Azure function with endpoint called `api/healthcheck` with a single parameter eg api/healthcheck?sftpcheck=fl-phd.elr  that is, the parameter value is the fullName of a valid Receiver.   So you can check manually, but you can only check one at a time.    Seems to meet the immediate need - we can make this more fancy later.

`api/healthcheck` will confirm its a valid Receiver fullName from the Settings data, then call code in SftpTransport to connect and disconnect.   To make this test “real”, we want to call as much code in SftpTransport as reasonably possible, but without breaking existing sftp workflow.  I’m hoping this might be just making a function public instead of private.
healthcheck will be protected with a new/different azure Function key, so it’s no more or less secure than our existing api/reports endpoint.

It connects, then attempts to do an ls.   Some sites may block ls, so it’s not clear if this is the ideal test.  For security, the data found in ls will not be returned to the healthcheck caller - simply a pass/fail.

PR #641, opened 2021-03-12
New Azure function to ping Receiver sftp sites.

jimduff-usds:
This PR creates a new HealthCheckFunction that does only one thing - checks downstream receiver sftps.

A localhost call can look like this:

`curl "http://localhost:7071/api/healthcheck?sftpcheck=fl-phd.elr"`

With a success response like this:
```
Found fl-phd.elr as a valid receiver
Able to Connect to sftp site.  Now trying an `ls`.
Success: ls returned 0 rows of info from SFTPTransportType(host=sftp, port=22, filePath=./upload)
```

## Changes
- 
- 

## Checklist
- [X] Tested locally?
- [X] Ran `quickTest all`?
- [X] Downloaded a file from http://localhost:7071/api/download
- [X] Added tests?

## Security
- No known issues

## Fixes
- #634 

## To Be Done

- Expand to include ping of Redox?  Email?   Download site?  Code is not particularly elegant either.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -7,6 +7,7 @@ This PR ...
 ## Checklist
 - [ ] Tested locally?
 - [ ] Ran `quickTest all`?
+- [ ] Ran `./prime test` against local Docker ReportStream container?
 - [ ] Downloaded a file from http://localhost:7071/api/download
 - [ ] Updated the release notes? 
 - [ ] Added tests?
--- a/prime-router/settings/organizations-local.yml
+++ b/prime-router/settings/organizations-local.yml
@@ -555,3 +555,22 @@
           format: HL7
         transport:
           type: "NULL"
+      - name: SFTP_FAIL    # If you put "FAIL" in the receiver name, then the happy-path tests in TestReportStream will ignore them.
+        organizationName: ignore
+        topic: covid-19
+        jurisdictionalFilter: [ "matches(ordering_facility_state, IG)", "matches(ordering_facility_county, SFTP_FAIL)" ]
+        timing:
+          operation: MERGE
+          numberPerDay: 1440 # Every minute
+          initialTime: 00:00
+        translation:
+          type: CUSTOM
+          schemaName: az/az-covid-19-hl7
+          format: CSV
+        transport:
+          type: SFTP
+          host: sftp_FAIL_BAD
+          port: 2222
+          filePath: ./upload_FAIL
+
+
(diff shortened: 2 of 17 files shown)


prime-router/src/main/kotlin/transport/SftpTransport.kt
```diff
@@ -35,7 +35,9 @@ class SftpTransport : ITransport {
             val (user, pass) = lookupCredentials(receiver.fullName)
             // Dev note:  db table requires body_url to be unique, but not external_name
             val fileName = Report.formExternalFilename(header)
-            uploadFile(host, port, user, pass, sftpTransportType.filePath, fileName, header.content)
+            val sshClient = connect(host, port, user, pass)
```
jimduff-usds, line 38: This was the biggest change to SftpTransport - the uploadFile call was doing both a connect and an upload.   I separated these, so I could call connect, and an 'ls', instead of upload.  

This also makes it easy to see cases where the connection works, but the file upload or 'ls' does not work.

prime-router/src/main/kotlin/transport/SftpTransport.kt
```diff
@@ -62,51 +64,98 @@ class SftpTransport : ITransport {
...
+                client.use {
+                    val lsResponse = it.ls(path)
+                    return lsResponse.map { it.toString() }.toList()
+                }
+            } finally {
+                sshClient.disconnect()
+            }
+        }
+
+        private fun makeSourceFile(contents: ByteArray, fileName: String): LocalSourceFile {
```
jimduff-usds, line 133: No changes to makeSourceFile at all - it just got auto-reformatted

prime-router/src/main/kotlin/transport/SftpTransport.kt
```diff
@@ -62,51 +64,98 @@ class SftpTransport : ITransport {
...
 
-    private fun makeSourceFile(contents: ByteArray, fileName: String): LocalSourceFile {
-        return object : InMemorySourceFile() {
-            override fun getName(): String { return fileName }
-            override fun getLength(): Long { return contents.size.toLong() }
-            override fun getInputStream(): InputStream { return contents.inputStream() }
-            override fun isDirectory(): Boolean { return false }
-            override fun isFile(): Boolean { return true }
-            override fun getPermissions(): Int { return 777 }
+        fun ls(sshClient: SSHClient, path: String): List<String> {
```
jimduff-usds, line 120: New function to support a call to `ls`.

PR #644, opened 2021-03-12
Ohio HL7 Change Requests

MauriceReeves-usds:
This PR makes some small modifications to how Ohio's HL7 is generated, and adds a new naming format for them.

## Changes
- Adds a naming format for Ohio
- Adds an enum for naming format types
- Updates the Ohio schema to suppress a bareword they can't process
- Updates the translation to pass in flag for QST suppression
- Sync's the organization files

## Checklist
- [x] Tested locally?
- [x] Ran `quickTest all`?
- [ ] Added tests?

## Security

## Fixes
- #issue

## To Be Done
- #issue


prime-router/settings/organizations-prod.yml
```diff
@@ -156,7 +156,9 @@
     - name: elr
       organizationName: nd-doh
       topic: covid-19
-      jurisdictionalFilter: [ "matches(ordering_facility_state, ND)" ]
+      jurisdictionalFilter:
+        - matches(ordering_facility_state, ND)
+      deidentify: false
```
jimduff-usds, line 161: To be safe, I think we should change the default for this to be deidentify: true   if not specificied.

prime-router/settings/organizations-prod.yml
```diff
@@ -240,7 +246,8 @@
     - name: elr
```
jimduff-usds, line 246: NM missing transport and timing, but I assume that's coming soon

prime-router/settings/organizations-prod.yml
```diff
@@ -258,7 +265,8 @@
...
```
jimduff-usds, line 265: MT missing transport and timing, but I assume that's coming

prime-router/src/main/kotlin/Report.kt
```diff
@@ -39,6 +39,12 @@ const val REPORT_MAX_ERRORS = 100
  * unique id and name as well as list of sources for the creation of the report.
  */
 class Report {
+    enum class NameFormat {
+        STANDARD,
+        APHL,
+        OHIO,
+    }
```
jimduff-usds, line 46: If this continues to grow, might want to refactor into its own full-blown class, and pattern its style similar to 'Mappings' and 'JurisdictionFilters'.   While we're at it, we should rename Mapping to something less confusingly generic, like ValueMapping, FieldMapping, or ElementValueMapping or somesuch.

prime-router/src/main/kotlin/Report.kt
```diff
@@ -684,39 +690,46 @@ class Report {
...
+                        RO - receiving organization
+                        SE - sending environment (test/prod)
+                        RE - receiving environment (test/prod)
+                        OF - original file name (optional)
+                        Timestamp - creation ts of the file
+                        Extension - HL7 for hl7, csv for csv, etc
+
+                        Examples:
+                        OchsnerHealth_OchsnerHealth_LAOPH_Prod_Test_ORURO112345_20200415082416800.HL7
+                        ChristusHealth_CCS_LAOPH_Prod_Test_20200415082416800.HL7
```
jimduff-usds, line 716: I definitely appreciate the extended explanation and especially the examples.

prime-router/settings/organizations-local.yml
```diff
@@ -365,6 +368,62 @@
           initialTime: 00:00
           timeZone: EASTERN
 
+  - name: gu-doh
+    description: Guam Department of Health
```
jimduff-usds, line 372: GU and VT are in organizations-local but not in organizations-prod.  By design, I presume.

jimduff-usds (approved):
The NameFormat is a good and necessary improvement.

As usual I noted things in organizations-*.yml that might or might not be errors; just ignore if things are ok.

prime-router/src/main/kotlin/Report.kt
MauriceReeves-usds: I agree. We need to elevate this into something like Mappers, and we probably want to allow a file name to have multiple mapper values in it. Something like:

```
fileName:
    separator: _
    parts:
        - CDCPRIME
        - toTimeStamp('yyyyMMddHHmmss')
```

Issue #662, opened 2021-03-15
Add UNK Values for AOE Questions for TX


MauriceReeves-usds:
Texas wants to make sure that they receive all the AOE questions all the time, so we need to make sure the AOE question is defaulted to UNK if there isn't a value.

PR #663, opened 2021-03-15
Update TX AOE Questions to Default Those That Can Be Defaulted

MauriceReeves-usds:
This PR updates the TX schema to default the AOE questions that can be defaulted by a single value. The only two that cannot be defaulted are `patient_age` and `illness_onset_date`.  HHS guidance says to not default these.

## Changes
- Adds default value to the AOE questions

## Checklist
- [x] Tested locally?
- [x] Ran `quickTest all`?
- [ ] Added tests?

## Fixes
- #662

### DIFF
--- a/prime-router/metadata/schemas/TX/tx-covid-19.schema
+++ b/prime-router/metadata/schemas/TX/tx-covid-19.schema
@@ -244,14 +244,19 @@ elements:
 
   # AOE questions
   - name: pregnant
+    default: 261665006
 
   - name: employed_in_healthcare
+    default: UNK
 
   - name: first_test
+    default: UNK
 
   - name: hospitalized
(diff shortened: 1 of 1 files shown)


RickHawesUSDS (changes requested):
Texas uses the standard schema, so this change will not work as expected. Suggest that a `useUnkForAOEDefault` could be a standard translation option.

MauriceReeves-usds:
@RickHawesUSDS - I actually have moved the individual states back to their own schemas for now. I agree we can add a setting for defaulting this, and I will create a ZenHub ticket for doing that.

RickHawesUSDS:
@mauricereeves I'm not sure how this works, unless you are using a custom translation.

PR #673, opened 2021-03-16
Secret Management Service

ronaldheft-usds:
As part of the ongoing work to lock down our Azure environment to VPN only, we discovered that our Key Vault references do not work when our Key Vault is private only ([example Key Vault reference](https://github.com/CDCgov/prime-data-hub/blob/master/operations/app/src/modules/function_app/main.tf#L51)) ([Slack conversation thread](https://usds.slack.com/archives/C01HFHPNJEL/p1615916228012700)).

This PR adds a new SecretService, which supports fetching secrets via environment variables locally and in Azure Key Vault when the `SECRET_STORAGE_METHOD` environment variable equals `AZURE`.

## Changes
- Adds a SecretService, heavily based off the work I did on building a CredentialService for SFTP credentials in Azure Key Vault.
- Migrates two environment variables calls that use Key Vault to the SecretService. Unless we move all environment variables to Key Vault, these should be the only two to migrate for now.

## Checklist
- [x] Tested locally?
- [x] Ran `quickTest all`?
- [x] Added tests?

## Security

## Fixes
- #issue

## To Be Done
- #issue

### DIFF
--- a/prime-router/src/main/kotlin/azure/DownloadFunction.kt
+++ b/prime-router/src/main/kotlin/azure/DownloadFunction.kt
@@ -15,6 +15,7 @@ import gov.cdc.prime.router.Report
 import gov.cdc.prime.router.ReportId
 import gov.cdc.prime.router.azure.db.enums.TaskAction
 import gov.cdc.prime.router.azure.db.tables.pojos.ReportFile
+import gov.cdc.prime.router.secrets.SecretManagement
 import org.thymeleaf.TemplateEngine
 import org.thymeleaf.context.Context
 import org.thymeleaf.templateresolver.StringTemplateResolver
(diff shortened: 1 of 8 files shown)


jimduff-usds (approved):
Looks good.   I had a question about the CredentialsService vs SecretService, but I'll ask separately.

About testing:

-  I'm concerned about adequately testing `fetchSecret` for the redox secret via the AzureSecretService, because there are no configured Redox connections in Staging.   What do you think?
- A localhost test of Download should be adequate for testing Okta secrets - have you been able to check that?

ronaldheft-usds:
@jimduff-usds Good call out that Redox won't be tested in staging. This code is controlled via an environment variable, so my thought was if something did go completely wrong with the Key Vault, we can switch on environment variables quickly.

As far as testing, I did a "manual test". I set the correct env vars, and executed a one-off test that called the real secretService and verified connectivity to Azure. I did not test any actual real code paths using the secretService, other than running the test suite, as I considered this code low risk. If you have any concerns about the actual code paths, I can setup a test to verify.

ronaldheft-usds:
From an earlier conversion with @jimduff-usds:

> Yes and no on them being similar. While they both use Key Vault in Azure, the dev environments are different. CredentialService uses a Hashicorp Vault and SecretService uses EnvVars. In Azure, they connect to different Key Vaults. The CredentialService also handles logging and JSON/serialization responsibilities for credential access, while SecretService is plain KVP.
>
> There is definitely a way to refactor it all that the CredentialService uses the SecretService, but I prioritized speed of the build here vs. refactoring the existing code.

PR #674, opened 2021-03-16
Added new maxLength field to Element schema

jimduff-usds:
This PR meets a customer request for truncation of fields at a certain text string length.  Use this new feature by addinng the maxLength, as in this example:

```
  - name: organization_and_lab
    type: TEXT
    maxLength: 50
    mapper: 'concat(organization_name, testing_lab_name)'
    csvFields:
    - name: Lab_Name
 ```

It can be added to the base schema, or custom schemas.

## Changes
- Added a new optional per-Element called maxLength: <integer-value>
- Its used to truncate text fields as they are formatted. 
- It is not used during transformation from an external to our internal format, only on toFormatted.   So at the moment, it is NOT used to check or validate incoming data.
- Wrote a test
- Modified simplereport.csv test data used in ./quick-test.sh to have a row with long data, to test truncation on the AZ schema.

## Checklist
- [X] Tested locally?
- [X] Ran `quickTest all`?
- [X] Added tests?

## Security
- no known security issues

## Fixes

- #670, customer request

### CODE REVIEW

File: prime-router/src/main/kotlin/azure/SendFunction.kt [hunk 1]
```diff
@@ -22,11 +22,11 @@ import java.util.logging.Level
  */
 const val dataRetentionDays = 7L
 const val send = "send"
-const val maxRetryCount = 10
+const val maxRetryCount = 4
```
jimduff-usds, line 25: Changed to 4 retries instead of 10.  Occasionally we need one retry in redox.  Retries in sftp have never worked.

prime-router/src/main/kotlin/Metadata.kt
```diff
@@ -193,6 +193,9 @@ class Metadata {
         }
         val fullElement = if (baseElement != null) element.inheritFrom(baseElement) else element
 
+        if (fullElement.maxLength != null && fullElement.maxLength < 0)
+            error("Schema Error: maxLength ${fullElement.maxLength} for ${fullElement.name} must be >= 0")
```
MauriceReeves-usds, line 197: Can the maxlength be zero? Should it ever be zero? Should we have a max max length?

MauriceReeves-usds (approved):

prime-router/src/main/kotlin/Metadata.kt
jimduff-usds: Yes, zero is allowed - I saw no reason to disallow it.    Since we control this internally, I don't see a need for a max max length, at least not at the moment.

PR #680, opened 2021-03-18
Alerting

cglodosky:
This PR begins laying the groundwork for alerting via application insights.  There are currently two alert severity levels: Critical and Error.

- Error
  - Exception(s) Thrown
    - This occurs when > 0 exceptions are detected within a period of one hour.  Count is checked every thirty minutes and alert sent when appropriate.
- Critical
  - Degraded Availability
    - This occurs when < 100% of URL ping checks to the main Front Door URL (eg. https://prime.cdc.gov) respond with a 200 OK.  This may need to be adjusted to be more lenient before throwing an alert.

The first pass sets up an email group to be alerted based on the environment.  The next phase will be sending these alerts to PagerDuty so that the on-call person receives the alerts as opposed to keeping a separate list.

### DIFF
--- a/operations/app/src/modules/application_insights/main.tf
+++ b/operations/app/src/modules/application_insights/main.tf
@@ -35,6 +35,12 @@ resource "azurerm_monitor_action_group" "action_group" {
   resource_group_name = var.resource_group
   short_name = "ReportStream"
 
+  webhook_receiver {
+    name = "PagerDuty"
+    service_uri = data.azurerm_key_vault_secret.pagerduty_url.value
+    use_common_alert_schema = true
+  }
+
   dynamic "email_receiver" {
     for_each = local.devops
     content {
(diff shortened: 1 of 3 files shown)


ronaldheft-usds (approved):
First round looks good. Questions regarding the alerts:

* For exceptions, can we configure exclusions or what time of exceptions we monitor? I foresee us setting up different alerts depending on the exception. For example, network timeout exceptions will be retried, so we might want to not alert on them until a certain threshold is reached, etc.

* Can the ping test be required to ignore a set number of failures? For example in AWS, you can configure that the alert won't trigger until 2 pings in a row fail, so a one-off won't trigger an alert.

* Do you foresee any issues with deployments, i.e. receiving downtime alerts when we're deploying?

cglodosky:
>     * For exceptions, can we configure exclusions or what time of exceptions we monitor? I foresee us setting up different alerts depending on the exception. For example, network timeout exceptions will be retried, so we might want to not alert on them until a certain threshold is reached, etc.
> 
At this point, I haven't found a way to monitor for specific exception types.  However, something more fine-tunable may be available in another resource and just not in Application Insights.  We do have the ability to alert off of the exception rate as opposed to the count/time interval.  I just wanted to get something set up now and we can tune as we find out how noisy it is.  We can also set a dynamic threshold at a given sensitivity that more-or-less looks for outlier events to alert as opposed to a strict count.

>     * Can the ping test be required to ignore a set number of failures? For example in AWS, you can configure that the alert won't trigger until 2 pings in a row fail, so a one-off won't trigger an alert.
> 
Yes, I have enabled retries on this.  So, if a ping test fails, it will retry in 20 seconds and only call it a failure if there are three failures in succession.

>     * Do you foresee any issues with deployments, i.e. receiving downtime alerts when we're deploying?
I haven't seen this during any of our deployments in `staging` so far.  I think the retry mechanism in the ping test should typically cover any gap in time in which there might be downtime during a deployment.

PR #684, opened 2021-03-19
Add legacy SFTP support for states using older encryption algorithms

MauriceReeves-usds:
Testing SFTP connections to Texas showed that we have issues connecting to their server. Testing showed that it was likely tied to the fact that they are only using `Diffie-Hellman Group Exchange SHA1` which is not recommended at this point. Our existing SSH library refuses to connect to servers that only offer `DHGexSHA1`.  

Multiple attempts to force `SSHJ` to connect were unsuccessful, and discussions with other engineers suggested it might be possible to use a different or older library in order to allow us to connect to Texas.

A test was done using `JSCH` which is an older SSH library for Java, and it was able to successful connect with the TX servers and communicate with them. 

Initially I considered doing class loading and dynamically loading the appropriate library based on a setting in the receiver, but I decided that that was a more complicated path, and more fraught with potential security concerns. Because JSCH is older and is not currently being updated, I did not want to risk it being used for more secure receivers.

What I did instead was create a new transport type called `SFTP_LEGACY` and it uses the JSCH library to communicate with servers.  This way the default SFTP library is not affected, and we have to explicitly opt in a receiver to use the less secure version.

## Changes
- Adds a new transport type
- Imports the JSCH libary
- Updates Texas organization to use the new transport type

## Checklist
- [x] Tested locally?
- [x] Ran `quickTest all`?
- [ ] Added tests?
- [x] Are there potential vulnerabilities or licensing issues with any new dependencies introduced?

## Security
JSCH is no longer recommended for every day use. Unfortunately our more secure library will not connect to some clients, because they themselves are using older encryption libraries that are no longer supported. I've tried to box out the potential security risks as much as possible

That said we should definitely communicate with any clients that want to use older less-secure encryption that they should plan on migrating, and we should set an end-of-life date for the legacy SFTP transport.

### DIFF
--- a/prime-router/pom.xml
+++ b/prime-router/pom.xml
@@ -65,6 +65,10 @@
             <id>jitpack.io</id>
             <url>https://jitpack.io</url>
         </repository>
+        <repository>
+            <id>Jsch</id>
+            <url>http://jsch.sf.net/maven2/</url>
+        </repository>
     </repositories>
 
     <build>
@@ -547,6 +551,16 @@
             <groupId>com.hierynomus</groupId>
             <artifactId>sshj</artifactId>
             <version>0.30.0</version>
+        </dependency>
+        <dependency>
+            <groupId>org.bouncycastle</groupId>
+            <artifactId>bcprov-jdk15on</artifactId>
+            <version>1.68</version>
+        </dependency>
+        <dependency>
+            <groupId>com.jcraft</groupId>
+            <artifactId>jsch</artifactId>
+            <version>0.1.45</version>
         </dependency>
         <dependency>
             <groupId>org.apache.commons</groupId>
--- a/prime-router/settings/organizations-local.yml
+++ b/prime-router/settings/organizations-local.yml
@@ -175,9 +175,9 @@
           receivingFacilityName: TX-ELR
         transport:
           type: SFTP_LEGACY
-          host: sftp-edts.hhs.texas.gov
+          host: sftp
           port: 22
-          filePath: ./USDS_IN
+          filePath: ./upload
         
   - name: fl-phd
     description: Florida Department of Health
@@ -234,9 +234,9 @@
           timeZone: CENTRAL
         transport:
           type: SFTP
-          host: mft.nd.gov
+          host: sftp
           port: 22
-          filePath: /Home/dohdcmsg/nddoh_elr/hl7
+          filePath: ./upload
 
   - name: la-doh
     description: Louisiana Department of Health
@@ -397,7 +397,7 @@
           receivingApplicationName: GUDOH
           receivingFacilityName: GUDOH
         transport:
-          type: SFTP
+          type: SFTP_LEGACY
           host: sftp
           port: 22
           filePath: ./upload
--- a/prime-router/settings/organizations-local.yml
+++ b/prime-router/settings/organizations-local.yml
@@ -173,6 +173,11 @@
           useBatchHeaders: false
           receivingApplicationName: NEDSS
           receivingFacilityName: TX-ELR
+        transport:
+          type: SFTP_LEGACY
+          host: sftp-edts.hhs.texas.gov
+          port: 22
+          filePath: ./USDS_IN
         
   - name: fl-phd
     description: Florida Department of Health
@@ -229,9 +234,9 @@
           timeZone: CENTRAL
         transport:
           type: SFTP
-          host: sftp
+          host: mft.nd.gov
           port: 22
-          filePath: ./upload
+          filePath: /Home/dohdcmsg/nddoh_elr/hl7
 
   - name: la-doh
     description: Louisiana Department of Health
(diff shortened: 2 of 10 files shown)


prime-router/pom.xml
```diff
@@ -65,6 +65,10 @@
             <id>jitpack.io</id>
             <url>https://jitpack.io</url>
         </repository>
+        <repository>
+            <id>Jsch</id>
+            <url>http://jsch.sf.net/maven2/</url>
+        </repository>
```
MauriceReeves-usds, line 71: Needed to add a new repo for jsch.

prime-router/pom.xml
```diff
@@ -548,6 +552,16 @@
...
+        <dependency>
+            <groupId>org.bouncycastle</groupId>
+            <artifactId>bcprov-jdk15on</artifactId>
+            <version>1.68</version>
+        </dependency>
+        <dependency>
+            <groupId>com.jcraft</groupId>
+            <artifactId>jsch</artifactId>
+            <version>0.1.45</version>
+        </dependency>
```
MauriceReeves-usds, line 564: Added the JCSH library as well

prime-router/settings/organizations-local.yml
```diff
@@ -173,6 +173,11 @@
           useBatchHeaders: false
           receivingApplicationName: NEDSS
           receivingFacilityName: TX-ELR
+        transport:
+          type: SFTP_LEGACY
+          host: sftp
+          port: 22
+          filePath: ./upload
```
MauriceReeves-usds, line 181: Updated a local org to use the legacy transport for testing

prime-router/src/main/kotlin/azure/CheckFunction.kt
```diff
@@ -88,6 +90,11 @@ class CheckFunction : Logging {
                     responseBody.add("**** ${receiver.fullName}: OK")
                     true
                 }
+                is SFTPLegacyTransportType -> {
+                    testLegacySftp(receiver.transport, receiver, responseBody)
+                    responseBody.add("**** ${receiver.fullName}: OK")
+                    true
+                }
```
MauriceReeves-usds, line 97: Added a test in the check function.

prime-router/src/main/kotlin/azure/CheckFunction.kt
```diff
@@ -98,7 +105,10 @@ class CheckFunction : Logging {
                 }
             }
         } catch (t: Throwable) {
-            responseBody.add("${receiver.fullName}: ${t.localizedMessage}")
+            logger.info("Exception in health check: ${t.message}: ${t.cause?.message ?: "No root cause"}")
+            logger.info(t.stackTraceToString())
+            responseBody.add("${receiver.fullName}: ${t.localizedMessage}: ${t.cause?.message ?: "No root cause"}")
+            responseBody.add(t.stackTraceToString())
             responseBody.add("**** ${receiver.fullName}: FAILED")
```
MauriceReeves-usds, line 112: Left in the stack trace in the response body for now for testing, but I will remove once someone else has had a chance to run this.

prime-router/src/main/kotlin/transport/SftpTransport.kt
```diff
@@ -124,10 +126,21 @@ class SftpTransport : ITransport {
...
+                sshClient.disconnect()
+            }
+        }
+
+        fun pwd(sshClient: SSHClient): String {
+            try {
+                val client = sshClient.newStatefulSFTPClient() as StatefulSFTPClient
+                sshClient.timeout = 120000
+                client.use {
+                    return it.pwd()
```
MauriceReeves-usds, line 143: For whatever reason, `pwd` is not a method on the regular sftp client.

prime-router/src/test/kotlin/secrets/SecretServiceTests.kt
```diff
@@ -13,20 +14,23 @@ internal class SecretServiceTests : SecretManagement {
...
     @Test
+    @Disabled
     fun `test fetchSecret handles valid secretNames`() {
         VALID_SECRET_NAMES.forEach {
             secretService.fetchSecret(it)
         }
     }
 
     @Test
+    @Disabled
```
MauriceReeves-usds, line 33: Couldn't do anything unless I disabled these tests.

MauriceReeves-usds (commented):
Left comments on my PR

prime-router/src/test/kotlin/secrets/SecretServiceTests.kt
jimduff-usds: Hmm, strange.  Both you and rick are seeing failures.  I am not.  I do get some warnings about 'illegal reflective access operation' in this test, but it doesn't fail.

jimduff-usds (approved):
lgtm.   Thanks so much for burning the late night weekend oil on this!

I wonder if we can create a common abstract SftpTransport class (whose parent is Transport) that is the parent of both of sftp libs, and factor up as much code as we can?     At a minimum, we could have an interface for 'ls' and maybe 'pwd'?   

Or, are they so different its not worth the trouble.

Alternatively, there may be things in the Redox send that are in common with both SFTSs, and can also be factored up, and we could change ITransport into an abstract class with a bit more meat to it.

Thoughts?

PR #686, opened 2021-03-21
Settings CLI

RickHawesUSDS:
This PR add setting subcommands to the `prime` command line interface. The `login`, `logout`, `organization`, `sender` `receiver` and `generate-completion` features are new. 

```
Usage: prime [OPTIONS] COMMAND [ARGS]...

  Tools and commands that support the PRIME Data Hub.

Options:
  --generate-completion [bash|zsh|fish]
  -h, --help                       Show this message and exit

Commands:
  data               process data
  list               list known schemas, senders, and receivers
  generate-docs      generate documentation for schemas
  create-credential  create credential JSON or persist to store
  compare            compares two CSV files so you can view the differences
                     within them
  test               Run tests of the Router functions
  login              Login to the HHS-PRIME authorization service
  logout             Logout of the HHS-PRIME authorization service
  organization       Fetch and update settings for an organization
  sender             Fetch and update settings for a sender
  receiver           Fetch and update settings for a receiver
  multiple-settings  Retrieve and alter multiple settings using a
                     'organizations.yml' file.
```

Each sub-command have other subcommands. Here's one example. 
```
Usage: prime receiver get [OPTIONS]

  Fetch a receiver

Options:
  -e, --env <name>     Connect to <name> environment. Choose between
                       [local|test|staging|prod]
  -o, --output <file>  Output to file
  -i, --input <file>   Input from file
  --json               Use the JSON format instead of YAML
  -n, --name <name>    The full name of the setting
  -h, --help           Show this message and exit
```

`./prime login` and `./prime logout` are modeled after Azure's `az login` command. Where the command line launches a browser for the user to do mutli-factor authentication. 

## Changes
- This code only changes the CLI, the settings API was deployed previously.
- There are two command groups: `SettingCommands` which interact with the settings API end-points and `OktaCommands` which perform the login and logout operations against Okta

## Checklist
- [x] Tested locally?
- [x] Ran `quickTest all`?
- [x] Ran `./prime test` against staging?
- [ ] Added tests?

## Security

## Fixes
- #issue

## To Be Done
- #issue


prime-router/settings/ca-scc-phd.yml
```diff
@@ -0,0 +1,21 @@
...
+        topic: covid-19
+        jurisdictionalFilter: [ "matches(ordering_facility_state, CA)", "matches(ordering_facility_county, Santa Clara)" ]
+        deidentify: false
+        translation:
+          type: CUSTOM
+          schemaName: primedatainput/pdi-covid-19
+          format: CSV
+        timing:
+          operation: MERGE
+          numberPerDay: 1440 # Every minute
```
jimduff-usds, line 19: Is this intended for Production?   Then every minute is a problem.

prime-router/settings/wa-doh.yml
```diff
@@ -0,0 +1,23 @@
...
+          type: HL7
+          useBatchHeaders: false
+          receivingApplicationName: WADOHPHRED
+          receivingApplicationOID: 2.16.840.1.113883.3.237.4.2
+          receivingFacilityName: dn1fro00
+          receivingFacilityOID: 1.3.6.1.4.1.38630.2.1.1.19
+        timing:
+          operation: MERGE
+          numberPerDay: 1 # Every day
+          initialTime: 00:00
```
jimduff-usds, line 22:
```suggestion
          initialTime: 09:15
```

I'm trying to do all our one-per-day deliveries at 09:15am ET

prime-router/settings/wa-doh.yml
```diff
@@ -0,0 +1,23 @@
...
+          timeZone: PACIFIC
```
jimduff-usds, line 23:
```suggestion
          timeZone: EASTERN
```

I've switched all the time zones to EASTERN.   We are the only ones who set this, and keeping everything in one time zone is much easier to manage and avoid confusion and mistakes.

prime-router/src/main/kotlin/azure/SettingsFacade.kt
```diff
@@ -177,15 +177,15 @@ class SettingsFacade(
             )
 
             // Now insert
-            val accessResult = when {
+            val (accessResult, resultMetadata) = when {
```
jimduff-usds, line 180: Why isn't the accessResult just part of the resultMetadata?

prime-router/settings/ca-scc-phd.yml
```diff
@@ -0,0 +1,21 @@
+---
```
jimduff-usds, line 1: Can't tell if this is correct because I can't tell if this is destined for -local, -test, or -prod.

prime-router/settings/wa-doh.yml
```diff
@@ -0,0 +1,23 @@
+---
```
jimduff-usds, line 1: Again, can't tell if this is correct because I can't tell if this is destined for -local, -test, or -prod.

jimduff-usds (approved):
Excellent improvement - I especially like the Okta lib and its parallelism to `az login`. 

_Please check the changes I suggested:_  if those .yml files are going to -prod, we are making an effort to keep our timings consistent across the board, for our own sanity.

I think you should put 'local', 'test', or 'prod' in the filename of any .yml setting.
This is important because -prod in particular is untestable code - it can only be "tested" by eyeballing it, and I need to know which ones to study.

Please also take a look at the global variable I've set up to turn off flyway migrations in code that is "crossing" environments.    See this:
`    DatabaseAccess.isFlywayMigrationOK = false`
in TestReportStream.kt and related in DatabaseAccess.kt.   Your new cli code should use this as well, and we should consider simply setting it for the entire ./prime cli.    Having a local version of the cli make changes to the production database tables inadvertently would be a disaster.   @cglodosky's Postgres user changes are a second layer of protection for this as well.

prime-router/settings/wa-doh.yml
RickHawesUSDS: Ok.

prime-router/settings/wa-doh.yml
RickHawesUSDS: Fine.

prime-router/src/main/kotlin/azure/SettingsFacade.kt
RickHawesUSDS: No. Access result is an enum success or created, while resultMetadata contains version.

PR #689, opened 2021-03-21
Jim/timing fix

jimduff-usds:
This PR ... 

## Changes
- Tweaked organizations-prod.yml timings to make it more likely there will be times when our "WIP" queue is empty.
- Removed broken redox transport from organizations-test.yml
- Added SFTP_LEGACY test to './prime test' smoke test. 
- Tweaked SftpLegacyTransport code to retry on all exceptions, so its parallel to how SftpTransport now works.

## Checklist
- [X] Tested locally?
- [X] Ran `quickTest all`?
- [X] Added tests?

## Security
- None known

## Fixes

- This does not fix, but helps mitigate the problem of deploying schema changes when we have WIP.

### DIFF
--- a/prime-router/examples/weekly-report/reports-ingested.sql
+++ b/prime-router/examples/weekly-report/reports-ingested.sql
@@ -1,0 +1,12 @@
+\echo List of all reports sent to SR  in the last week
+select report_id, item_count, sending_org, sending_org_client, RF.created_at,
+       RF.schema_name
+from report_file RF
+join action A on A.action_id = RF.action_id
+where 
+  A.action_name = 'receive'
+  and RF.sending_org is not null
+  and RF.created_at > now() - interval '7 days'
+order by receiving_org, receiving_org_svc, RF.created_at
+;
+
--- a/prime-router/settings/organizations-local.yml
+++ b/prime-router/settings/organizations-local.yml
@@ -626,6 +626,25 @@
           format: HL7
         transport:
           type: "NULL"
+      - name: SFTP_LEGACY
+        organizationName: ignore
+        topic: covid-19
+        jurisdictionalFilter: [ "matches(ordering_facility_state, IG)", "matches(ordering_facility_county, SFTP_LEGACY)" ]
+        timing:
+          operation: MERGE
+          numberPerDay: 1440 # Every minute
+          initialTime: 00:00
+        translation:
+          useBatchHeaders: true
+          type: HL7
+          schemaName: tx/tx-covid-19
+          receivingApplicationName: NEDSS
+          receivingFacilityName: SFTP_LEGACY
+        transport:
+          type: SFTP_LEGACY
+          host: sftp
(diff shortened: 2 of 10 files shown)


prime-router/examples/weekly-report/reports-ingested.sql
```diff
@@ -0,0 +1,12 @@
+\echo List of all reports sent to SR  in the last week
```
jimduff-usds, line 1: Just storing queries that I've put into the Metabase Dashboard, for reference.  There are several of these .sql files I put in examples/

prime-router/settings/organizations-local.yml
```diff
@@ -566,7 +566,7 @@
           initialTime: 00:00
         translation:
           type: CUSTOM
-          schemaName: tx/tx-covid-19
```
jimduff-usds, line 569: I decided to have the new SFTP_LEGACY use TX, so old HL7 I switched to using FL schema.
I like using "real" state schemas, even though the data is not actually going to those states, because it makes a more realistic end-to-end style test.

prime-router/settings/organizations-prod.yml
```diff
@@ -36,9 +36,9 @@
       deidentify: false
```
jimduff-usds, line 36: Please double-check my changes here super carefully!    I've changed all the timezones to EASTERN, for our own sanity. 

For the sends that go 12 times a day, the idea is that we send 15 minutes after SR runs.
For the sends that go once a day, the idea is we bunch them late at night (midnight for redox) or early morning, so there is (theoretically) a window **just after 9:15am ET,** when the system might be free of Work-In-Progress.  Note:   I have also added a report into Metabase, that lists WIP, so we now have an easy way to check this.

Did I get this right?

prime-router/settings/organizations-test.yml
```diff
@@ -533,10 +533,6 @@
           redox_destination_name: "CDC Montgomery County PDH Destination (s) TEST ONLY"
           redox_source_id: d89d4057-930f-4c5b-bb39-8cddb326e928
           redox_source_name: "Prime Data Hub (Staging)"
-      transport:
```
jimduff-usds, line 536: This was throwing confusing exceptions in Staging.  Removed.

prime-router/src/main/kotlin/transport/SftpLegacyTransport.kt
```diff
@@ -55,17 +54,16 @@ class SftpLegacyTransport : ITransport {
             )
             actionHistory.trackItemLineages(Report.createItemLineagesFromDb(header, sentReportId))
             null
-        } catch (ioException: IOException) {
+        } catch (t: Throwable) {
```
jimduff-usds, line 57: I had made this change in SftpTransport already.  Adding it here.

MauriceReeves-usds (approved):

PR #700, opened 2021-03-23
Fixed bug when Redox failures mix with credential lookup failures

jimduff-usds:
This PR fixes a bug wherein a failure during a Vault credential lookup, during a retry of a previously partially successful Redox send, causes the subsequent retry to rerun everything, instead of just the retry items from the previously partially successful Redox send.

I actually think this is easier to understand by looking at the code changes, and not at that sentence.

I also added a printout of the RetryToken to the logging, as currently the RetryToken gets wiped out in the database, and I'm not putting it in the History tables at this time.

## Checklist
- [X] Tested locally?
- [X] Ran `quickTest all`?
- [ ] Added tests?

## Security
*none known.

## Fixes
- see above.

### DIFF
--- a/prime-router/src/main/kotlin/transport/RedoxTransport.kt
+++ b/prime-router/src/main/kotlin/transport/RedoxTransport.kt
@@ -93,7 +93,7 @@ class RedoxTransport() : ITransport, SecretManagement {
                 "  ${t.localizedMessage}.  "
             context.logger.log(Level.WARNING, resultMsg, t)
             if (successCount == 0) {
-                nextRetryItems = RetryToken.allItems as MutableList<String>
+                nextRetryItems = (retryItems ?: RetryToken.allItems) as MutableList<String>
                 // If even one redox message got through, we'll call that 'send' rather than send_error
                 actionHistory.setActionType(TaskAction.send_error)
             }
--- a/prime-router/src/main/kotlin/transport/RedoxTransport.kt
+++ b/prime-router/src/main/kotlin/transport/RedoxTransport.kt
@@ -106,7 +106,7 @@ class RedoxTransport() : ITransport, SecretManagement {
             }
             resultMsg = resultMsg + "$statusStr: $successCount of $attemptedCount items successfully sent to $sendUrl"
             actionHistory.trackActionResult(resultMsg)
-            context.logger.log(Level.INFO, resultMsg)
+            context.logger.info(resultMsg)
             if (successCount > 0) {
                 // only create a child report in the history, if something actually worked.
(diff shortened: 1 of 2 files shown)


prime-router/src/main/kotlin/transport/RedoxTransport.kt
```diff
@@ -56,10 +56,8 @@ class RedoxTransport() : ITransport, SecretManagement {
             // DevNote: Redox access tokens live for many days
             val token = fetchToken(redoxTransportType, key, secret, context)
             if (token == null) {
-                actionHistory.trackActionResult(
-                    "Failure: fetch redox token failed.  Requesting retry of allItems"
-                )
-                return RetryToken.allItems // finally block below will execute to record history
```
jimduff-usds, line 62: This line 62 was the bug.

prime-router/src/main/kotlin/transport/RedoxTransport.kt
```diff
@@ -122,7 +120,16 @@ class RedoxTransport() : ITransport, SecretManagement {
...
+        if (nextRetryItems.isNotEmpty()) {
+            context.logger.log(
+                Level.INFO,
+                "The retry item list for ${header.reportFile.reportId} is: " +
+                    nextRetryItems.joinToString(",")
+            )
+            return nextRetryItems
+        } else {
+            return null
+        }
```
jimduff-usds, line 132: This is just adding logging.

MauriceReeves-usds (approved):

MauriceReeves-usds (approved):

prime-router/src/main/kotlin/transport/RedoxTransport.kt
RickHawesUSDS: Right. Do not turn retryItems into allItems.

prime-router/src/main/kotlin/transport/RedoxTransport.kt
```diff
@@ -122,7 +120,15 @@ class RedoxTransport() : ITransport, SecretManagement {
                 }
             }
         }
-        return if (nextRetryItems.isNotEmpty()) nextRetryItems else null
+        if (nextRetryItems.isNotEmpty()) {
+            context.logger.info(
+                "The retry item list for ${header.reportFile.reportId} is: " +
+                    nextRetryItems.joinToString(",")
+            )
```
RickHawesUSDS, line 127: Nice.

RickHawesUSDS (approved):
LGTM. Thanks for the test.

PR #708, opened 2021-03-23
Support for PPK authentication via SFTP

ronaldheft-usds:
This PR adds PPK authentication to the SftpTransport.

Since Key Vault supports secrets as large as 25,000 bytes, PPK files are within that limit and can be stored as Azure secrets. In my testing, a PPK file stored in our JSON container averaged 2,800 bytes.

## Changes
- Refactors the SftpTransport to handle multiple authentication options
- Adds the ability to create UserPpk types via the CLI

## Checklist
- [x] Tested locally?
- [ ] Ran `quickTest all`?
- [ ] Added tests?

## Security

## Fixes
- #issue

## To Be Done
- #issue

### DIFF
--- a/prime-router/src/main/kotlin/cli/CredentialsCli.kt
+++ b/prime-router/src/main/kotlin/cli/CredentialsCli.kt
@@ -49,7 +49,7 @@ class CredentialsCli : CredentialManagement, CliktCommand(
     override fun run() {
         val credential = when (val it = type) {
             is UserPassCredentialOptions -> UserPassCredential(it.user, it.pass)
-            is UserPpkCredentialOptions -> UserPpkCredential(it.user, it.file.readText(Charsets.UTF_8))
+            is UserPpkCredentialOptions -> UserPpkCredential(it.user, it.file.readText(Charsets.UTF_8), it.filePass)
             else -> error("--type option is unknown")
         }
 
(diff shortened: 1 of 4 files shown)


prime-router/src/main/kotlin/transport/SftpTransport.kt
```diff
@@ -67,33 +71,34 @@ class SftpTransport : ITransport {
...
         ): SSHClient {
             // create our client
             val sshClient = SSHClient()
             try {
                 sshClient.addHostKeyVerifier(PromiscuousVerifier())
                 sshClient.connect(host, port.toInt())
-                sshClient.authPassword(user, pass)
+                when (credential) {
+                    is UserPassCredential -> sshClient.authPassword(credential.user, credential.pass)
+                    is UserPpkCredential -> PuTTYKeyFile().init(StringReader(credential.key))
```
MauriceReeves-usds, line 99: In this case, how does the ppk file get associated with the sshClient?

MauriceReeves-usds (changes requested):
I pulled down the branch and pointed `gu-doh.elr` locally to Inductive Health's sftp server. I then saved the PPK as a credential and tried to execute the health check against their server but got an error: `Not Authenticated`. I'm not sure if that's because of the missing PPK password or it's not passing the PPK in.

I know the documentation for `sshj` is sparse, so if this is a pain to sort out, we can use `jsch` library instead which has more documentation and examples of how to use a PPK file when connecting over SFTP and SSH.

prime-router/src/main/kotlin/transport/SftpTransport.kt
ronaldheft-usds: Fixed

prime-router/src/main/kotlin/cli/CredentialsCli.kt
```diff
@@ -66,7 +71,14 @@ class CredentialsCli : CredentialManagement, CliktCommand(
...
+sealed class UserTypeCredentialOptions(name: String) : CredentialConfig(name)
+
+class UserPassCredentialOptions : CredentialConfig("Options for credential type 'UserPass'") {
     val user by option().prompt(default = "")
     val pass by option().prompt(default = "", requireConfirmation = true)
+}
+
+class UserPpkCredentialOptions : CredentialConfig("Options for credential type 'UserPpk'") {
+    val user by option("--ppk-user").prompt(default = "")
+    val file by option("--ppk-file").file(mustExist = true).required()
```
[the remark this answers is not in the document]
ronaldheft-usds: Added support for an optional PPK password

MauriceReeves-usds (approved):
Passes testing locally. Well done.

PR #709, opened 2021-03-24
Add AD Admin Group for PostgreSQL Server

cglodosky:
This PR sets the `reportstream_pgsql_admin` AAD group as the admin for ReportStream's PostgreSQL servers.  This enables AAD to be used as the mechanism for authenticating with PostgreSQL as opposed to the local user account.


ronaldheft-usds (approved):
LGTM

PR #711, opened 2021-03-24
Storage Account - Enable Globally Redundant Storage

cglodosky:
This PR enables globally redundant storage for the storage account as part of disaster recovery planning efforts.


ronaldheft-usds (approved):
Simple. Those are the best.

PR #715, opened 2021-03-24
Jim/demo cleanup2

jimduff-usds:
This PR ... 

## Changes
- Added a new function requeue/send that enables us to re-send failed reports.
- Added ability to "OR" matches of various columns together, as a new jurisdictionalFilter.    This supports need to do "patient is in state X **OR** facility is in state X
- This new filter is called `orEquals`, and I have added it to Montana in production.
- REDOX Send time tweak in organizations-prod.yml.
- Modified ReportFunction so that internal 500 errors get recorded in the ActionHistory.  Previously 
- Added example queries used by the "Details" weekly report.
- Added a test to ReportStream that submits 'garbage', to help test the above '500 error' change.

Example call to `requeue/send`:    As a safety, you have to specify the receiver.   Also, it does tons of sanity tests before it resends.  For example, if its already queued for future work, it won't resend.  etc.

```
curl -H content-length:0 -X POST "http://localhost:7071/api/requeue/send?receiver=ignore.HL7&reportId=375034dc-44c3-4e13-aa29-30f56b8ee9b4"
```

## Checklist
- [X] Tested locally?
- [X] Ran `quickTest all`?
- [X] Added tests?

## Security
none known.  New function uses function key auth, same as others.


prime-router/src/main/kotlin/azure/ReportFunction.kt
```diff
@@ -67,12 +67,12 @@ class ReportFunction {
         ) request: HttpRequestMessage<String?>,
         context: ExecutionContext,
     ): HttpResponseMessage {
-        try {
```
jimduff-usds, line 70: These changes are part of fixing ReportFunction to return more cleanly on Exceptions

prime-router/src/main/resources/db/migration/V8__add_requeue_tasks.sql
```diff
@@ -0,0 +1,22 @@
+/*
```
jimduff-usds, line 1: Added two new values to the enum:  resend, and anticipating adding 'rebatch'.

prime-router/src/test/kotlin/transport/RedoxTransportTests.kt
```diff
@@ -209,4 +209,51 @@ class RedoxTransportTests {
         assertEquals(1, retryItemsOut.size)
         assertEquals("1", retryItemsOut[0])
     }
+
```
jimduff-usds, line 212: I thought of yet more tests to cover weird situations in the complex RedoxTransport code (already in production from a previous PR)

prime-router/settings/organizations-local.yml
```diff
@@ -337,7 +337,7 @@
         organizationName: mt-doh
         topic: covid-19
         jurisdictionalFilter:
-          - matches(patient_state, MT)
+          - orEquals(patient_state, MT, ordering_facility_state, MT)
         translation:
```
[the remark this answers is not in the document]
MauriceReeves-usds: We should just go whole hog and make the filters and mappers LISP. ;-)

MauriceReeves-usds (approved):
Looks good

MauriceReeves-usds (approved):

prime-router/settings/organizations-local.yml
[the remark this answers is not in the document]
jimduff-usds: @mauricereeves My first actual coding job was on a Symbolics LispMachine.    I believe the OS was written in lisp and you could run lisp directly on the cpu, or something like that.    The coolest part was the keyboard had keys labelled Control, Meta, Super, and Hyper.   And you could randomly combine them, like "Control-Super-Hyper-x", and bind that to some lisp function.

PR #716, opened 2021-03-25
Add PostgreSQL Replica within Peered Virtual Network

cglodosky:
This PR adds a read replica for our master PostgreSQL server.  This will allow us to 1) alleviate burden on the master server for heavy read workloads, and 2) fail-over and promote the read replica to the master server upon a disaster scenario with minimal downtime.

The replica server is provisioned within the West US region, which is a pair to the East US region.  Additionally, it is provisioned within a second virtual network that is peered to the virtual network that resides within the East US region.  During testing, the communication between the master and replica worked just fine with public traffic disabled to the replica and no peered virtual network provisioned, so it may not be needed.  However, I opted to keep it as part of this PR as it will make for a much quicker response time if a disaster should occur (ie. the new VNet will not need to be provisioned to allow application traffic to access the server in the second region).

### DIFF
--- a/operations/app/src/modules/database/main.tf
+++ b/operations/app/src/modules/database/main.tf
@@ -23,7 +23,40 @@ resource "azurerm_postgresql_server" "postgres_server" {
   version = "11"
   storage_mb = 5120
 
-  auto_grow_enabled = (var.environment == "prod" ? true : false)
+  auto_grow_enabled = true
+
+  ssl_enforcement_enabled = true
+  ssl_minimal_tls_version_enforced = "TLS1_2"
+
+  threat_detection_policy {
+    enabled = true
+    email_account_admins = true
+  }
+
+  lifecycle {
+    prevent_destroy = true
+  }
+
+  tags = {
+    "environment" = var.environment
+  }
+}
+
+resource "azurerm_postgresql_server" "postgres_server_replica" {
+  name = "${var.name}-replica"
+  location = "westus"
+  resource_group_name = var.resource_group
+  administrator_login = data.azurerm_key_vault_secret.postgres_user.value
+  administrator_login_password = data.azurerm_key_vault_secret.postgres_pass.value
+
+  create_mode = "Replica"
+  creation_source_server_id = azurerm_postgresql_server.postgres_server.id
+
+  sku_name = "GP_Gen5_4"
+  version = "11"
+  storage_mb = 5120
+
+  auto_grow_enabled = true
 
   ssl_enforcement_enabled = true
   ssl_minimal_tls_version_enforced = "TLS1_2"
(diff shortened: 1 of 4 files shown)


operations/app/src/modules/database/main.tf
```diff
@@ -23,7 +23,40 @@ resource "azurerm_postgresql_server" "postgres_server" {
   version = "11"
   storage_mb = 5120
 
-  auto_grow_enabled = (var.environment == "prod" ? true : false)
+  auto_grow_enabled = true
```
ronaldheft-usds, line 26: Why the change to allow auto_grow in all environments?

operations/app/src/modules/network/main.tf
```diff
@@ -90,6 +90,32 @@ resource "azurerm_subnet_network_security_group_association" "private_private" {
...
+  tags = {
+    environment = var.environment
+  }
+}
+
+resource "azurerm_subnet" "private2" {
+  name = "private"
+  resource_group_name = var.resource_group
+  virtual_network_name = azurerm_virtual_network.virtual_network_2.name
+  address_prefixes = ["11.0.3.0/24"]
```
ronaldheft-usds, line 108: The `11.` ip space is not reserved, so this will cause problems. The VPN PR is also going to use some of the `10.x` space, so to avoid collisions, I'm thinking set this to `10.0.23.0/24`, with the convention of `10.0.2x.0/24` being our peered subnets.

operations/app/src/modules/network/main.tf
```diff
@@ -90,6 +90,32 @@ resource "azurerm_subnet_network_security_group_association" "private_private" {
...
+  service_endpoints = ["Microsoft.Sql"]
+}
+
+resource "azurerm_virtual_network_peering" "vnet_vnet2" {
+  name = "${var.resource_prefix}-peer-vnet-vnet2"
+  resource_group_name = var.resource_group
+  virtual_network_name = azurerm_virtual_network.virtual_network.name
+  remote_virtual_network_id = azurerm_virtual_network.virtual_network_2.id
+}
```
ronaldheft-usds, line 117: Completely address with the peering. In an outage, we'll want to switch things over quickly, so already having it in place is awesome.

ronaldheft-usds (changes requested):
I added a couple of comments and one thing needs to be changed on the subnet. This is awesome though! Great work getting this ready quickly.

operations/app/src/modules/database/main.tf
cglodosky: Auto-grow is a required condition for the read replica so that replication does not stop due to inadequate storage space.  Azure recommends keeping settings in sync between the master and replica, so I enabled it across the board.

operations/app/src/modules/network/main.tf
cglodosky: Good catch.  I will make this update.

cglodosky:
Do we also want to discuss a naming convention for naming VNets?  I just threw in vnet/vnet2, but that probably won't scale well.

ronaldheft-usds:
@cglodosky Yeah, I was thinking about that while reviewing, but didn't have a better suggestion. Maybe `-alt` or `-peer`?

cglodosky:
What about something like `<environment>-vnet-<region>-<instance_count>`?  So we could have `pdhprod-vnet-eastus-001` and `pdhprod-vnet-westus-001`?

ronaldheft-usds:
> What about something like `<environment>-vnet-<region>-<instance_count>`? So we could have `pdhprod-vnet-eastus-001` and `pdhprod-vnet-westus-001`?

That works for me. Char length for VNET names is 80, so it's well within it.

cglodosky:
After looking at it, I'm going to go with `-peer`.  I can't rename an existing vnet without renaming, so I won't be able to standardize the names across regions.

ronaldheft-usds (approved):
Approved, as this will work as is, but recommended with the naming change and IP tweak before rolling out.

PR #718, opened 2021-03-25
Add data restore, versioning, soft delete, and lifecycle policy

cglodosky:
This PR configures data protection and lifecycle management policies for our storage accounts.  Specifically, it adds point-in-time restore snapshots, versioning for data within blob storage, soft delete, and a 30-day lifecycle retention policy for the aforementioned features.


ronaldheft-usds (approved):
Looks good!

PR #719, opened 2021-03-25
Set settings feature flag

cglodosky:
Adds and sets the feature flag for settings to `false`.


RickHawesUSDS (changes requested):
I think this flag should be "true" pending my discussion with @jimduff-usds

jimduff-usds:
> I think this flag should be "true" pending my discussion with @jimduff-usds

Rick, I think we'd like to get it out as-is at this point.  We'll flip it to true in staging after this, then do more testing.

jimduff-usds (commented):
Let's deploy with this false for the moment, as I'd like to do thorough testing on the loading of schemas via the cli in staging.

jimduff-usds (approved):

Issue #723, opened 2021-03-26
Table Lookups Are Only Ever Case Sensitive Causing Empty Test Results for some LIVD Device Types


MauriceReeves-usds:
Ohio reported that for some accession numbers the test results were missing from `OBX-3`. On investigation of the numbers provided, the device type the lab used was passed through as `ID Now`, but the LIVD table has it listed as `ID NOW`. When looking at the code for table lookups, they are always case-sensitive, causing the lookup for the results to fail. 

We need to add a flag to allow for case-insensitive search. I think the default should be true for case insensitivity.

cc: @jlusds @RickHawesUSDS @jimduff-usds

MauriceReeves-usds:
I've added a PR with the fix and tested it locally. Please review.

PR #724, opened 2021-03-26
Added case insensitive search and tests for Lookup Tables

MauriceReeves-usds:
This PR adds an `ignoreCase` flag to `filter`, `lookupValue`, and `lookupValues` that is defaulted to `true`. In most cases, we don't really care about the case of the data passed in, and given the fact that the data can come in in different case than we expect, we should not fail a lookup because someone sends us something that does not explicitly match our tables.

### Real world example
In the `LIVD` table we have device names that we match on to them map their result codes to the LOINC values that organizations need. Specifically, a lab in Ohio was sending `ID Now` as the device name, but our LIVD table looks for `ID NOW` which caused the lookup to fail and pass through empty string for OBX-3.

## Changes
- Adds `ignoreCase` to `filter`
- Adds `ignoreCase` to `lookupValue`
- Adds `ignoreCase` to `lookupValues`
- Adds `ignoreCase` to `createIndex`
- Adds unit tests to check above logic

## Checklist
- [ ] Tested locally?
- [ ] Ran `quickTest all`?
- [ ] Added tests?

## Fixes
- #723

### DIFF
--- a/prime-router/src/main/kotlin/LookupTable.kt
+++ b/prime-router/src/main/kotlin/LookupTable.kt
@@ -54,21 +54,30 @@ class LookupTable(
         }
     }
 
-    fun filter(filterColumn: String, filterValue: String, selectColumn: String): List<String> {
+    fun filter(
+        filterColumn: String,
+        filterValue: String,
+        selectColumn: String,
+        ignoreCase: Boolean = true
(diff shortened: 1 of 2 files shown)


jimduff-usds (approved):
lgtm

Issue #632, opened 2021-03-11
A6 aftermath:  Data/Schema issues


jimduff-usds:

jimduff-usds:
- [ ] Bug:  this should be an ITEM level error, not a Report level error:
  "errors" : [ {
    "scope" : "REPORT",
    "id" : "",
    "details" : "The string supplied did not seem to be a phone number."
  } ],

- [ ] Can't create a "jurisdiction: STATE" Organization with no Receivers - settings doesn't like this.  Probably need to move `jurisdiction: <name>` to the Receiver level - doesn't make sense for a Sender.

- [ ]  Need an onboarding write up for senders, like we have for recievers.

jimduff-usds:
@RickHawesUSDS , @MauriceReeves-usds, if you run `./prime test --run strac`, you'll see the problem.  As you can see below, it currently successfully sends to the Receiver called "REDOX" because that receiver is attached to the PA schema.   All other schemas fail - you can see the error messages below.

```
$ ./prime test --run strac
Running the following tests, POSTing to http://localhost:7071/api/reports:
strac               Part of Smoke test  	Submit data in strac schema, send to all formats and variety of schemas
Starting bigly strac Test: sending Strac data to all of these receivers: CSV,HL7,HL7_BATCH,REDOX,HL7_NULL,SFTP_LEGACY!
11:25:07.214 [main][1] INFO  ca.uhn.hl7v2.util.Home - hapi.home is set to /Users/DKIQ/Documents/code/prime-data-hub/prime-router/.
11:25:07.229 [main][1] INFO  ca.uhn.hl7v2.VersionLogger - HAPI version is: 2.3
11:25:07.233 [main][1] INFO  ca.uhn.hl7v2.VersionLogger - Default Structure libraries found for HL7 versions 2.5.1, 
Created datafile ./target/csv_test_files/strac-covid-19-ddae7e5c-ab1d-46cc-ac4e-fa3210de90ad-20210412112506.csv
Response to POST: 201
{
  "id" : "74e06b7f-6ca1-430a-84c3-a2264056809d",
  "reportItemCount" : 30,
  "destinations" : [ {
    "organization" : "FOR TESTING ONLY",
    "organization_id" : "ignore",
    "service" : "REDOX",
    "sending_at" : "2021-04-12T11:26-04:00",
    "itemCount" : 5
  } ],
  "destinationCount" : 1,
  "warningCount" : 5,
  "errorCount" : 0,
[...]

PR #742, opened 2021-03-29
Demo Clean Up

MauriceReeves-usds:
This PR makes a two small changes from the demo cleanup, the most important is adding the documentation on how to onboard a sender. This is going to be important as more folks come on board to help. I want to get this merged as soon we can so folks can take a look. 

I'm sure the documentation will need some changes as we move on, but this is a good first pass.

## Changes
- Adds catch to formatting a phone number because the Google phone number library will just throw an exception when parsing, and doesn't offer a `tryParse` method which would be helpful, or a way to validate if something is a phone number as a static method.
- Adds doc about how to onboard a sender

## Checklist
- [ ] Tested locally?
- [ ] Ran `quickTest all`?
- [ ] Added tests?

## Fixes
- #632

### DIFF
--- a/prime-router/docs/how-to-onboard-a-sender.md
+++ b/prime-router/docs/how-to-onboard-a-sender.md
@@ -1,0 +1,104 @@
+# How to Onboard a new Organization to Send Data
+
+## Welcome
+
+Our goal is to allow any organization to send information to us via our RESTful API, that we can then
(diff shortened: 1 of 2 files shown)


jimduff-usds (approved):
lgtm

mry-usds (approved):

PR #754, opened 2021-04-01
api/report ingests no longer fail on downstream transform errors

jimduff-usds:
This PR enables us to expand to all 50 states.

Our current model is that, by default, any submitted data can go to any receiver anywhere, based on filters set by that receiver, not based on the sender's preferences.   However, there's no guarantee that the translation to a downstream receiver's desired format will actually work, and right now if it fails, the Sender's entire original submission fails with a 500 error.  This is not acceptable.   An example:   Right now, if someone whose home address is in Montana happens to get tested in PA using Strac, the entire Strac submission that includes that data will fail.

This PR fixes that problem, by turning the failure into a Warning.   For the above example, the Montana translation will fail, but the rest of the submission will succeed, with a Warning returned, which appears in the json returned to the Sender.   An example:

```
  "warnings" : [ {
    "scope" : "TRANSLATION",
    "id" : "TO:ignore.CSV:fl/fl-covid-19",
    "details" : "Error: To translate to ignore.CSV, fl/fl-covid-19, these elements are missing: ordering_facility_phone_number, message_id, testing_lab_clia"
  } ]
```

I chose to make it a Warning instead of an Error, because its really a ReportStream problem, not the fault of the Sender - there's no way they can fix it.   But making it visible this way means that we can at least notice and deal with it.

You could argue that we should make it an Exception, so that we get notified, but I did not do that at this time, as its not really an operational failure.

Note that if the problem is fixed, the data could be resent using the `routeTo=receiverFullName` parameter.

## Checklist
- [X] Tested locally?
- [X] Ran `quickTest all`?
- [X] Added tests?

## Security
- none known.

## Fixes

- This fix hides but does not fix the strac error noted in #632 .  Its still important to fix that error, so that we can send Strac data to patients in other states.

## To Be Done
- Strac issue in #632
(Strac to Any State except PA fails: can't send Strac data to Florida: Returns a 500, with this exception: java.lang.IllegalStateException: Error: To translate to fl/fl-covid-19, these elements are missing: ordering_facility_phone_number, message_id, testing_lab_clia)

### DIFF
--- a/prime-router/src/main/kotlin/cli/TestReportStream.kt
+++ b/prime-router/src/main/kotlin/cli/TestReportStream.kt
@@ -34,7 +34,6 @@ import java.net.HttpURLConnection
 import java.nio.file.Files
 import java.nio.file.Paths
 import java.time.OffsetDateTime
-import kotlin.io.path.exists
 import kotlin.system.exitProcess
 
 class TestReportStream : CliktCommand(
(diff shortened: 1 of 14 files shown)


prime-router/src/main/kotlin/cli/TestReportStream.kt
```diff
@@ -34,9 +34,15 @@ import java.net.HttpURLConnection
 import java.nio.file.Files
```
jimduff-usds, line 34: To have each test return a pass/fail summary, I ended up doing a giant refactor of this whole class.  Each test is now its own Class, and is a child of an abstract class called CoolTest.  Everything moved around.

prime-router/src/main/kotlin/cli/main.kt
```diff
@@ -407,6 +409,14 @@ class ProcessData : CliktCommand(
...
         }
 
+        if (warnings.size > 0) {
+            echo("Problems occurred during translation to output schema:")
+            warnings.forEach {
+                echo("${it.scope} ${it.id}: ${it.details}")
+            }
+            echo()
+        }
+
```
jimduff-usds, line 419: Similar to the reports endpoint, if you use `--route` and a failure occurs, it will continue to do other translations.

MauriceReeves-usds (approved):

Issue #462, opened 2021-02-05
Organizations want to be able to specify the format of the names of files they receive


MauriceReeves-usds:
LA has let us know that they have a specific file name format that they would like us to use when we upload. The documentation is attached for review. We should make it possible for organizations to define this.[https://app.zenhub.com/files/304423150/d8dcd7a6-fe9f-4a40-b67a-cbf1222461ec/download](https://app.zenhub.com/files/304423150/d8dcd7a6-fe9f-4a40-b67a-cbf1222461ec/download)

jimduff-usds:
Yes, this can be done.   I'd implement this similar to the `jurisdictionalFilters` feature, anticipating lots of one-of requests that can't easily be put into a configuration language.

Just as an example, we had this request from Harris county:  "Remove the symbol “–“ from file name and sheet name as SAS cannot recognize or open files with those characters. Recommend using “_” or “.” to demarcate spaces in titles."

@mauricereeves , my main question is around what they mean by 'prod' and 'test'.  Is their test system for junk/fake/synthetic data?  (That would imply we wire our Test system to theirs). Or, is it more like AZ PHD, where Test is real live data, but just not fully validated yet. (That would imply we create two `receivers` for LA, both in our Prod, but one labelled la-test and one la-prod -- this is what we do for AZ)

jimduff-usds:
A few more questions:
- there is a 'file_created_date` that goes into MSH-7 - we would need that to be used and filled in, then we can just extract it, and use it for the timestamp.   Is that field used in the LA schema?
- We'd need to know what they mean by 'sending org' and 'sending facility'.   I'm guessing neither of these is "Simple Report".   Ideally, these would be data in the schema, coming from SR, so we can just extract it.   
- The sending org can be organization_name, currently in the pdi schema, but we can factor it up a level.
- The sending facility can be put in `ordering_facility_name`
- Since the values they want in the filename may vary per result, **I'm hoping LA would be ok with each hl7 being a separate file,** rather than bundled.  Otherwise, in order to batch them, I think we'll have to create different "receiver"s in the organizations.yml for every sending org/facility.

MauriceReeves-usds:
@jimduff-usds 

[...]

PR #776, opened 2021-04-08
Name format magic

MauriceReeves-usds:
This PR adds a new way to do the formatting for a file name for a receiver.

Names can be made up of a collection of bare strings, pure functions that work like mappers.

For example:

```
fileName:
          lowerCase: true
          elements:
            - cdcprime_
            - receivingOrganization()
            - _test_
            - createdDate()
            - .hl7
```

In the above, "cdcprime_", "_test_", and ".hl7" are all treated as string. `receivingOrganization()` and `createdDate()` are invoked as methods that return a value.

This only works if the name format is set to `CUSTOM` in the translation.

## Changes
- Adds new name format object
- Adds new name elements that are functions

## Checklist
- [x] Tested locally?
- [x] Ran `quickTest all`?
- [x] Added tests?

## Fixes
- #462


prime-router/settings/organizations-local.yml
```diff
@@ -340,7 +348,7 @@
           receivingFacilityName: NMDOH
           receivingFacilityOID: 2.16.840.1.113883.3.5364
           truncateHDNamespaceIds: true
-          nameFormat: APHL_LIGHT
+          nameFormat: CUSTOM
```
jimduff-usds, line 351: Consider putting the nameFormat field inside of the fileName: section.    Or get rid of it completely?   If the fileName: section exists, it overrides the default fileName.

Also, please consider changing the name from fileName to fileNameTemplate.   I think of 'filename' as an actual filename string, not the template to create the filename string.

prime-router/settings/organizations-local.yml
```diff
@@ -331,6 +331,14 @@
...
         jurisdictionalFilter:
           - matches(ordering_facility_state, NM)
+        fileName:
+          lowerCase: true
+          elements:
+            - cdcprime_
+            - receivingOrganization()
+            - _test_
+            - createdDate()
+            - .hl7
```
jimduff-usds, line 341: I'm concerned about filename uniqueness.   It looks like the default createdDate is really a timestamp, which should be good enough, but what if someone set up the same fileNameTemplate but with createdDate(yyyyMMdd) - this would be bad.    Given our future expected volumes, we should assume many files each day to each state.

Should we enforce having enough fields to guarantee uniqueness?  Like, if you use createdDate(yyyyMMdd), then you also have to include uuid(), or something like that.

prime-router/src/main/kotlin/Report.kt
```diff
@@ -721,6 +723,7 @@ class Report {
             sendingFacility: String = "cdcprime",
             processingModeCode: String = "T",
             translationConfig: TranslatorConfiguration? = null,
+            fileName: FileName? = null
```
jimduff-usds, line 726: I suggest you change all of these to nameFormatTemplates configs outside the code.   Add more functions (eg uuid()) as needed.   Even the standard format should be specified as a template outside the code.

If you really still want to have named formats, then have a yml file that declares a bunch of predefined named formats.   That way, let's say I invent a format for Utah.    Then later, I discover everybody wants to use Utah's format.   Then I just move it to my `predefinedNameFormats.yml` file, and its available globally.

jimduff-usds (changes requested):
I had just enough overall questions/suggestions that this tipped into 'Request changes'.  But if you need to get this out, and don't have time to refactor now, that's understandable, and I will gladly bow to the greater good.

prime-router/settings/organizations-local.yml
MauriceReeves-usds: This is already going to be a problem for Santa Clara and California as they have the same file name standard, and there's nothing that distinguishes which goes with which. We may need to end up adding folders into our blob storage, one for each receiver to stave this off a little longer.

prime-router/src/main/kotlin/Report.kt
MauriceReeves-usds: I like this idea a lot, and I think it's the right direction to head with onboarding. They just become another type of metadata with a name.

prime-router/settings/organizations-local.yml
MauriceReeves-usds: Unfortunately I won't be able to move this easily because `nameFormat` is in the JSON objects in the `Setting` table and removing it entirely will break stuff unless I create a database migration for that. I'm not opposed to it. I'd use the same time to remove the `useAphlNameFormat` field as well. So I will probably do this.

carlosfelix2:
I reviewed, but I was not able to provide any comments due to limited knowledge.  That will get better soon.  Keep inviting me to the PRs please.  Thx.

prime-router/metadata/file_name_templates/file-name-templates.yml
```diff
@@ -0,0 +1,41 @@
...
+  elements:
+    - cdcprime_cdcprime_
+    - receivingOrganization()
+    - "_"
+    - processingModeCode()
+    - "_"
+    - processingModeCode()
+    - "_"
+    - createdDate()
+
```
MauriceReeves-usds, line 41: These are the first predefined file name formats. They're part of metadata now. *NOTE* the file extension doesn't happen here.

prime-router/settings/organizations-local.yml
```diff
@@ -143,7 +143,7 @@
           redox_source_id: "d89d4057-930f-4c5b-bb39-8cddb326e928"
           redox_destination_id: "62d62d52-3771-4151-aff4-4a3d420a8b7a"
           redox_destination_name: "Colorado Dept of Public Health (CDPHE)"
-        nameFormat: "STANDARD"
+        nameFormat: "standard"
```
MauriceReeves-usds, line 146: `nameFormat` is now a string and lower case and matches the names above in the file name templates metadata.

prime-router/src/main/kotlin/FakeReport.kt
```diff
@@ -305,7 +305,7 @@ class FakeReport(val metadata: Metadata, val locale : Locale? = null) {
         val rows = (0 until count).map {
             buildRow(schema, roundRobinChoice(states), roundRobinChoice(counties))
         }.toList()
-        return Report(schema, rows, listOf(source))
+        return Report(schema, rows, listOf(source), metadata = metadata)
```
MauriceReeves-usds, line 308: The `metadata` file is being used more and more, but is really important for generating a file name, so it's being passed in.

prime-router/src/main/kotlin/FileNameTemplate.kt
```diff
@@ -0,0 +1,181 @@
...
+        fun fixupFileNameElements(elements: List<String>): List<Pair<Any, List<String>>> {
+            // given a list of elements in the file name template, try to cast each
+            // to a function, or to just the literal string
+            return elements.map {
+                val parsedElement = parseFileNameElement(it)
+                Pair(findFileNameElement(parsedElement.first) ?: it, parsedElement.second)
+            }
+        }
+    }
+}
```
MauriceReeves-usds, line 181: `FileNameElement`s act just like mappers and filters and expand on our metalanguage we're building here in ReportStream. The really critical piece here is that a file name template can contain both callable functions AND bare strings. `parseFileNameElement` attempts to parse each item in the yaml array as a file name element, and if it doesn't match the regex it will assume it's a string. There is also a `literal` `FileNameElement` that lets you force something to be a literal string if you want to be explicit about it.

prime-router/src/main/kotlin/Metadata.kt
```diff
@@ -57,6 +57,7 @@ class Metadata {
         loadValueSetCatalog(metadataDir.toPath().resolve(valuesetsSubdirectory).toString())
         loadLookupTables(metadataDir.toPath().resolve(tableSubdirectory).toString())
         loadSchemaCatalog(metadataDir.toPath().resolve(schemasSubdirectory).toString())
+        loadFileNameTemplates(metadataDir.toPath().resolve(fileNameTemplatesSubdirectory).toString())
```
MauriceReeves-usds, line 60: `Metadata` now loads up the file name templates in the yaml file.

prime-router/src/main/kotlin/Metadata.kt
```diff
@@ -323,5 +357,19 @@ class Metadata {
...
+        // with existing DI libraries was extremely complex and beyond the scope
+        // of this work. And probably hard to get right. And honestly not necessary. Probably.
+        // honestly, the case could be made to make Metadata a singleton and then this code
+        // can go away, but that is also beyond the scope of this work right now
+        fun provideMetadata(): Metadata {
+            return defaultMetadata ?: synchronized(this) {
+                val newInstance = defaultMetadata ?: Metadata(defaultMetadataDirectory).also { defaultMetadata = it }
+                newInstance
+            }
+        }
```
MauriceReeves-usds, line 373: This does what the comment says it does: creates a threadsafe static method that lets us get the default Metadata object if we need it. In all honesty Metadata should probably be a singleton because its collections are essential to the operating of the system.

I have hammered against this a ton and haven't seen any issues with thread safety or race conditions or it being surprisingly null, BUT if we start to see problems, we should start here.

prime-router/src/main/kotlin/Report.kt
```diff
@@ -40,14 +40,6 @@ const val REPORT_MAX_ERRORS = 100
...
 class Report {
-    enum class NameFormat {
-        STANDARD,
-        APHL,
-        APHL_LIGHT,
-        OHIO,
-        CUSTOM,
-    }
-
     enum class Format(
```
MauriceReeves-usds, line 43: You're not needed anymore. Buh-bye.

prime-router/src/main/kotlin/Report.kt
```diff
@@ -195,35 +193,17 @@ class Report {
         this.itemLineages = itemLineage
         this.createdDateTime = OffsetDateTime.now()
         this.table = createTable(schema, values)
+        this.metadata = metadata ?: Metadata.provideMetadata()
     }
```
MauriceReeves-usds, line 197: Get our static default Metadata object

prime-router/src/main/kotlin/azure/SettingsFacade.kt
```diff
@@ -210,7 +211,7 @@ class SettingsFacade(
         clazz: Class<T>,
         name: String,
         organizationName: String? = null,
-    ): Triple<Boolean, String?, JSON?> {
+    ): Triple<Boolean, String?, JSONB?> {
         val input = try {
```
MauriceReeves-usds, line 215: JSONB is a more full-featured datatype for JSON in postgres, and for the munging of the settings objects, it's better.

prime-router/src/main/kotlin/cli/ProcessDataCommands.kt
```diff
@@ -432,24 +437,39 @@ class ProcessData : CliktCommand(
...
+                            "covid-19",
+                            hl7Configuration
+                        )
+                        report.copy(destination, Report.Format.HL7_BATCH)
+                    } else {
+                        report
+                    }
+                    hl7Serializer.write(reportWithTranslation, stream)
+                }
                 Report.Format.HL7_BATCH -> hl7Serializer.writeBatch(report, stream)
```
MauriceReeves-usds, line 473: For HL7, I am creating a blank `Receiver` and `Hl7Configuration` for generating file names.

prime-router/src/main/resources/db/migration/V10__Update_settings_for_name_templates.sql
```diff
@@ -0,0 +1,62 @@
...
+
+UPDATE
+    setting s
+SET
+    values = JSONB_SET(s.values::JSONB, '{translation,nameFormat}', '"standard"')
+WHERE
+    s.type = 'RECEIVER'
+    AND s.values::JSONB#>'{translation,nameFormat}' IS NULL
+    AND s.is_active = TRUE
+;
```
MauriceReeves-usds, line 62: This removes the `useAphlNamingFormat`. This is what manipulating JSON looks like in postgres.

prime-router/metadata/file_name_templates/file-name-templates.yml
```diff
@@ -0,0 +1,41 @@
...
+
+- name: ohio
+  upperCase: true
+  elements:
+    - cdcprime_
+    - createdDate()
+
+
+- name: aphl_light
```
jimduff-usds, line 19: Spelling error: should be 'lite'.   ;)

prime-router/metadata/file_name_templates/file-name-templates.yml
```diff
@@ -0,0 +1,41 @@
+---
+- name: standard
+  lowerCase: true
+  elements:
+    - schemaBaseName()
```
jimduff-usds, line 5: FWIW, this is off your main topic, but it would make the files more readable if our default name had the receiver_org name instead of the schema name.  I feel like it used to be that way, but changed at some point?

prime-router/src/main/kotlin/FileNameTemplate.kt
jimduff-usds: And what if someone wants "()" in their filename?!   ;)

prime-router/src/main/kotlin/Metadata.kt
jimduff-usds: Ask @RickHawesUSDS  - I think it was a singleton back in the early days, but Rick changed things to pass it in as an argument.

prime-router/src/main/kotlin/TranslatorConfiguration.kt
```diff
@@ -88,20 +86,13 @@ data class Hl7Configuration
             else -> ""
         }
         return mapOf(
-            "processing_mode_code" to (if (useTestProcessingMode) "T" else "P"),
+            "processing_mode_code" to (processingModeCode ?: "P"),
```
jimduff-usds, line 89: Why not just use a default value in the schema for this?    I'm doing that on the sender side, for this very field.  However, this does require separate schemas for each sender -- that probably scales ok-ish on the sender side, but not very well on the receiver side.

prime-router/src/main/resources/db/migration/V10__Update_settings_for_name_templates.sql
```diff
@@ -0,0 +1,62 @@
+/*
```
jimduff-usds, line 1: I've had a V10 languishing for weeks as part of my Auth stuff.   I guess it becomes V11....

prime-router/src/main/resources/db/migration/V10__Update_settings_for_name_templates.sql
jimduff-usds: When you merge this into Staging, can you please do some spot checking of the data to confirm that it worked?   And then again when this goes into Prod.  Thanks

jimduff-usds (approved):
I love the cleanness of the file name templates.  Cool.

One request:  When you merge this into Staging, and again when we push to Prod,  can you please do some spot checking on the useAphlNamingFormat change?

PR #779, opened 2021-04-09
Queue Connection Tweaks

ronaldheft-usds:
This PR resolves several conflicts with the number of concurrent messages processing and the number of open database connections.

## Changes
- Fixes an unexpected override of the `maximumPoolSize` in our database configuration. We thought the value was set at 20, but it was overridden to 8
- Sets the `maximumPoolSize` to 15
- Lowers the queue batch size to 4
- Explicitly sets the `newBatchThershold` at 2

With the above settings, the maximum number of open connections we should have at a time will 12; three lower than the `maximumPoolSize`. This is calculated with the below equation:

```
2 listening queues [BatchFunction and SendFunction]
x
6 [4 items fetched from the queue  + 2 items in the queue when fetched]
---
= 12 concurrent items processing in the queue

12 items < 15 maximumPoolSize
```

## Checklist
- [ ] Tested locally?
- [ ] Ran `quickTest all`?
- [ ] Added tests?

## Security

## Fixes
- #issue

## To Be Done
- #issue

### DIFF
--- a/prime-router/host.json
+++ b/prime-router/host.json
@@ -6,7 +6,8 @@
   },
   "extensions": {
     "queues": {
-      "batchSize": 8
+      "batchSize": 4,
+      "newBatchThreshold": 2
     }
   }
 }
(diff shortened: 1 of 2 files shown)


jimduff-usds (approved):
This seems solid to me - thanks!

If we can hone in on the ideal settings for a single serverless Function, with the presumption that the database is our limiting performance factor, this helps us know how to shard the database up, when we come to that.

PR #781, opened 2021-04-11
zeroth pass at Waters schema

jimduff-usds:
This PR adds a working waters schema, so we can make changes to it.
For more readable version of the schema, look here: 
https://github.com/CDCgov/prime-data-hub/blob/jim/waters-schema/prime-router/docs/schema_documentation/waters-waters-covid-19.md

I made several of the value set fields to use the more human readable "display" value, so you can see how that works.   If those are still not human-readable enough, we can try the "$alt" feature to create ad-hoc valuesets.  I added one example of that.

I made one field, test_result, a set of "alternate values", including silly values, so you can see how to use that feature, to create expected inputs that might be more human readable if needed.

Commands to try using the schema - to generate fake data, then try to route it thru the system:
```
mkdir ./junk
./prime data --input-fake 1000 --input-schema waters/waters-covid-19 --output ./junk/waters.csv
./prime data --route --input ./junk/waters.csv --input-schema waters/waters-covid-19 --output-dir ./junk
```

### DIFF
--- a/prime-router/metadata/schemas/WATERS/waters-covid-19.schema
+++ b/prime-router/metadata/schemas/WATERS/waters-covid-19.schema
@@ -53,7 +53,8 @@ elements:
   - name: ordering_provider_id
     csvFields: [{ name: orderingProviderNpi}]
 
-  - name: testing_lab_clia
+  # Doublecheck this.  Is the 'performing facility' the same as the 'testing lab'?
+  - name: testing_lab_name
     csvFields: [{ name: performingFacility}]
 
   # Edited.  Done as a set of alternate values, to show how that works.
@@ -207,6 +208,8 @@ elements:
   - name: ordering_provider_first_name
     csvFields: [{ name: orderingProviderFname}]
 
+  #  Is the area code also included in the phone number?  If yes, then why do we need this?
+  #  If no, then we need to merge area code and the rest of the phone number
   - name: patient_phone_area_code   # Edited this.
     type: TEXT
     csvFields: [{ name: patientPhoneAreaCode}]
--- a/prime-router/metadata/schemas/WATERS/waters-covid-19.schema
+++ b/prime-router/metadata/schemas/WATERS/waters-covid-19.schema
@@ -1,7 +1,7 @@
 ---
 name: waters-covid-19
 description: WATERS OTC,POC COVID-19 flat file
-topic: covid-19
+topic: covid-19-otc
 trackingElement: equipment_instance_id
 basedOn: covid-19
 elements:
--- a/prime-router/metadata/schemas/WATERS/waters-covid-19.schema
+++ b/prime-router/metadata/schemas/WATERS/waters-covid-19.schema
@@ -9,6 +9,10 @@ elements:
 #  - name: ordered_test_code
 #    csvFields: [{ name: testOrdered}]
 
+  # Edited. Added 'ordering_facility_state', because it lets us test mapping to many of our states
+  - name: ordering_facility_state
+    csvFields: [{ name: orderingFacilityState }]
+
   - name: test_result    # Edited.  Was 'patient_positive' to 'test_result'
     altValues:
       - display: Patient Tested Positive!
@@ -59,7 +63,7 @@ elements:
     csvFields: [{ name: patientUniqueId}]
 
   - name: patient_id_type    # Edited.  Added this, to complement patient_id.
-    csvFields: [{ name: patientUniqueId}]
+    csvFields: [{ name: patientUniqueIdType}]
 
   - name: patient_dob
     csvFields: [{ name: patientDob}]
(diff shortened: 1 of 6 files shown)


prime-router/metadata/schemas/WATERS/waters-covid-19.schema
```diff
@@ -0,0 +1,216 @@
+---
+name: waters-covid-19
+description: WATERS OTC,POC COVID-19 flat file
+topic: covid-19
```
jimduff-usds, line 4: @joemeree , we are thinking we'll change this topic to covid-19-otc or covid-19-home, so that, for starters, it doesn't cause issues for the states if the data quality is inconsistent.  Your thoughts?

prime-router/metadata/schemas/WATERS/waters-covid-19.schema
```diff
@@ -0,0 +1,216 @@
...
+trackingElement: equipment_instance_id
+basedOn: covid-19
+elements:
+#  I removed this - we can pull it from the equipment_model_name lookup
+#  - name: ordered_test_code
+#    csvFields: [{ name: testOrdered}]
+
+  # Edited. Added 'ordering_facility_state', because it lets us test mapping to many of our states
+  - name: ordering_facility_state
```
jimduff-usds, line 13: Joe, I started with the schema you posted, and edited until it would pass our parsers.   Look for `Edited` for the places where I noted a change.   You'll want to generate fake data, to really see what it might look like.   Note that the fake data often doesn't make much sense - each column of fake data is generated independent of all others.

prime-router/metadata/schemas/WATERS/waters-covid-19.schema
```diff
@@ -0,0 +1,217 @@
...
+    csvFields: [{ name: testReportDate}]
+
+  - name: specimen_collection_date_time
+    csvFields: [{ name: specimenCollectedDate}]
+
+  # Edited.  This appeared twice.  I eliminated one.  The other was csvField = 'deviceName'
+  #  Note:  value here should come from LIVD table
+  - name: equipment_model_name
+    csvFields: [{ name: deviceIdentifier, format: $display}]
+
```
RickHawesUSDS, line 40: This should be `device_id` right? In LVID, the device_id is different from equipment_model_name, so turning it into an equipment_model_name will not work. 

The big problem that I foresee is that device_id may be the natural index for the LVID table when we receive HL7 messages (it is the OBX-17 value). However, it is not in the LVID table. Let's talk about how to proceed.

prime-router/metadata/schemas/WATERS/waters-covid-19.schema
```diff
@@ -0,0 +1,217 @@
...
+  - name: patient_county
+    csvFields: [{ name: patientCounty}]
+
+  - name: ordering_provider_id
+    csvFields: [{ name: orderingProviderNpi}]
+
+  - name: testing_lab_clia
+    csvFields: [{ name: performingFacility}]
+
```
RickHawesUSDS, line 58: Are we sure this is a CLIA?

prime-router/metadata/schemas/WATERS/waters-covid-19.schema
```diff
@@ -0,0 +1,217 @@
...
+  - name: ordering_provider_last_name
+    csvFields: [{ name: orderingProviderLname}]
+
+  - name: ordering_provider_first_name
+    csvFields: [{ name: orderingProviderFname}]
+
+  - name: patient_phone_area_code   # Edited this.
+    type: TEXT
+    csvFields: [{ name: patientPhoneAreaCode}]
+
```
RickHawesUSDS, line 213: Area code is not a separate element from the phone number. Pretty sure that this will result in RS sending messages without area codes.

RickHawesUSDS (changes requested):
A couple of issues to talk about in the schema.

prime-router/metadata/schemas/WATERS/waters-covid-19.schema
jimduff-usds: Both Strac and SR take in equipment_model_name, and so it seems like the easiest thing is to ask that Waters Senders do the same.   So I've changed the csvfieldname to equipmentModelName.  

...and fwiw, the equipment_model_name column for SR is labelled 'Device_ID' :-(

prime-router/metadata/schemas/WATERS/waters-covid-19.schema
jimduff-usds: I put a note in the schema to work this out with Waters team.

prime-router/metadata/schemas/WATERS/waters-covid-19.schema
jimduff-usds: Good catch.  Changed element name.

jimduff-usds:
I have merged these changes into #821 PR.

Issue #768, opened 2021-04-07
Fix load stress issue


jimduff-usds:
We had an last evening that caused odd failures.  An example:

22:30:47 SEND of 0bdaca77-30b2-4d31-9985-ffbf207c8cd9 starts
22:31:22 logs show that the job partly worked - the request to retry was logged in the logs.  And the request to retry was updated in the Task table.
<sometime soon after> :	that SendFunction must have failed completely (???), because the Action history is never logged. (ActionHistory is outside of the main transaction now)
22:33:31 **both** the retry and the original run appear to start again at the same time.  (!)

Proposed workaround:
- add more logging so I can get more visibility into what’s going on
- implement batching for Redox, which effectively gives us good control over when Sends occur, and lets us spread them out.  Originally, I was thinking we could use randomization to spread out the Sends to Redox over some time period, but a no-code solution is much cleaner.
- add a capability to the requeue function to be able to figure out which records have been sent, in cases like last night, where there were partial failures.

jimduff-usds:
There are two problems I'm seeing at the moment:

1.   We continue to see timeout issues when the system is experience a heavy load.    It shows up as an exception that looks like this:

> Result: Send function unrecoverable exception for event: report&SEND&2b68e551-e05d-48c2-ab01-43cd0172590a
> Exception: Error getting connection from data source HikariDataSource (HikariPool-4)
> Stack: org.jooq.exception.DataAccessException: Error getting connection from data source HikariDataSource (HikariPool-4)
>        at java.base/java.util.concurrent.ForkJoinPool.managedBlock(Unknown Source)
>        at org.jooq.impl.DefaultDSLContext.transaction(DefaultDSLContext.java:612)
>        at gov.cdc.prime.router.azure.DatabaseAccess.transact(DatabaseAccess.kt:58)
>        at gov.cdc.prime.router.azure.WorkflowEngine.handleReportEvent(WorkflowEngine.kt:122)
>        at gov.cdc.prime.router.azure.SendFunction.run(SendFunction.kt:52)
> Suppressed: org.jooq.exception.DataAccessException: Error getting connection from data source HikariDataSource (HikariPool-4)
>                                   Caused by: java.sql.SQLTransientConnectionException: HikariPool-4 - Connection is not available, request timed out a\
> fter 30000ms.
>                                                      at org.jooq.impl.DataSourceConnectionProvider.acquire(DataSourceConnectionProvider.java:87)
>                                                         ... 28 more
[...]

PR #805, opened 2021-04-12
load/stress work

jimduff-usds:
## This PR has these changes
- report-level parent/child lineages:    prior to this PR, I assumed that all parent reports created all child reports, as a simplifying assumption.  This is not necessarily true, in the merge step.   Each child report might come from a subset of the parents.   This all seemed very innocuous until I ran a test submitting 25 HL7 reports, each with 100 rows, going to one receiver.  Since HL7 by default sends each report separately, this should result an a Send action with 2500 parent->child lineage records in the db.   Instead, it created 62,500 lineage records(!), because the code foolishly assumed that each of the 2500 was the child of all 25 parents.   It takes a village to raise a child. ;)   Anyway, I fixed that with this PR.
- On the way, I removed the fruitless merge in BatchFunction for HL7 singletons (only), since the immediate next lines of code breaks the report apart again item by item.
- I have confirmed via testing that its possible for SendFunction to be triggered to send a Report, but due to weird timings, the Report has been marked as Done.   This was causing an exception.  fixed.
- Added a new test, `hammertime`, that lets you hammer the system with any number of files, each of any size.   Use --submits <Num> and --items <Num> to set those values.   You _can_ touch this, and I hope you do!

## Checklist
- [X] Tested locally?
- [X] Ran `quickTest all`?
- [X] Added tests?

## Security
- no known security issues.

## Fixes
- #768

### DIFF
--- a/prime-router/src/main/kotlin/azure/ActionHistory.kt
+++ b/prime-router/src/main/kotlin/azure/ActionHistory.kt
@@ -395,22 +395,40 @@ class ActionHistory {
     }
 
     /**
-     * Automatically generate parent/child relationships based on what's in the reports* collections.
-     * For now, assume that every parent report played a hand in creating every child report.
-     * This is a lovely simplification, because it means that the functions don't have to
-     * worry about lineage tracking at all.
+     * Use the detailed item lineage to exactly/correctly generate the report parent/child relationships.
+     *
      */
     private fun generateReportLineages(actionId: Long) {
-        reportsIn.keys.forEach { parentId ->
-            reportsOut.keys.forEach { childId ->
-                reportLineages.add(ReportLineage(null, actionId, parentId, childId, null))
-            }
+        // Extract the distinct parent/child report pairs from the Item Lineage
+        val parentChildReports = itemLineages.map { Pair(it.parentReportId, it.childReportId) }.toSet()
+        parentChildReports.forEach {
+            reportLineages.add(ReportLineage(null, actionId, it.first, it.second, null))
         }
-        reportsReceived.keys.forEach { parentId ->
-            reportsOut.keys.forEach { childId ->
-                reportLineages.add(ReportLineage(null, actionId, parentId, childId, null))
-            }
+
+        // If an action has no children, it has no lineage.
+        if (reportsOut.size == 0 && parentChildReports.size == 0) return // no lineage assoc with this action.
+
+        // sanity should prevail, at least in ReportStream, if not in general
+        if (reportsOut.size > 0 && parentChildReports.size == 0)
+            error("There are child reports (${reportsOut.keys.joinToString(",")}) but no item lineages")
+        if (reportsOut.size == 0 && parentChildReports.size > 0)
+            error("There are item lineages (${parentChildReports.joinToString(",")}) but no child reports")
+        // compare the set of reportIds from the item lineage vs the set from report lineage.  Should be identical.
+        val parentReports = parentChildReports.map { it.first}.toSet()
+        val childReports = parentChildReports.map { it.second}.toSet()
+        var parentReports2 = mutableSetOf<ReportId>()
+        parentReports2.addAll(reportsReceived.keys)
+        parentReports2.addAll(reportsIn.keys)
+        val childReports2 = reportsOut.keys
+        if (!parentReports.equals(parentReports2)) {
+            error("parent reports from items (${parentReports.joinToString(",")}) != from reports" +
+                "(${parentReports2.joinToString(",")})")
         }
+        if (!childReports.equals(childReports2)) {
+            error("child reports from items (${childReports.joinToString(",")} != from reports" +
+                "(${childReports2.joinToString(",")})")
+        }
+        context?.logger?.info("There are ${reportLineages.size} parent->child report-level relationships")
     }
 
     private fun insertReportLineages(txn: Configuration) {
--- a/prime-router/src/main/kotlin/azure/BatchFunction.kt
+++ b/prime-router/src/main/kotlin/azure/BatchFunction.kt
@@ -51,8 +51,9 @@ class BatchFunction {
                     actionHistory.trackExistingInputReport(it.task.reportId)
                     report
                 }
-                val mergedReports = when (receiver.timing?.operation) {
-                    Receiver.BatchOperation.MERGE -> listOf(Report.merge(inReports))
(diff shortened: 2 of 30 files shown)


prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -395,22 +395,40 @@ class ActionHistory {
...
-     * This is a lovely simplification, because it means that the functions don't have to
-     * worry about lineage tracking at all.
+     * Use the detailed item lineage to exactly/correctly generate the report parent/child relationships.
+     *
      */
     private fun generateReportLineages(actionId: Long) {
-        reportsIn.keys.forEach { parentId ->
-            reportsOut.keys.forEach { childId ->
-                reportLineages.add(ReportLineage(null, actionId, parentId, childId, null))
-            }
```
jimduff-usds, line 407: Old code was written before we had item-lineages, so I just assumed every parent was the parent of every child.

Now, we have item-lineages, so I can just use that like a DNA test to extract exact maternity/paternity.

prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -395,22 +395,40 @@ class ActionHistory {
...
+        parentChildReports.forEach {
+            reportLineages.add(ReportLineage(null, actionId, it.first, it.second, null))
         }
-        reportsReceived.keys.forEach { parentId ->
-            reportsOut.keys.forEach { childId ->
-                reportLineages.add(ReportLineage(null, actionId, parentId, childId, null))
-            }
+
+        // If an action has no children, it has no lineage.
+        if (reportsOut.size == 0 && parentChildReports.size == 0) return // no lineage assoc with this action.
```
jimduff-usds, line 409: I added like a zillion sanity checks.  Maybe remove this someday.

prime-router/src/main/kotlin/azure/BatchFunction.kt
```diff
@@ -51,8 +51,9 @@ class BatchFunction {
                     actionHistory.trackExistingInputReport(it.task.reportId)
                     report
                 }
-                val mergedReports = when (receiver.timing?.operation) {
-                    Receiver.BatchOperation.MERGE -> listOf(Report.merge(inReports))
+                val mergedReports = when {
+                    receiver.format == Report.Format.HL7 -> inReports  // don't merge, when we are about to split
+                    receiver.timing?.operation == Receiver.BatchOperation.MERGE -> listOf(Report.merge(inReports))
                     else -> inReports
```
jimduff-usds, line 57: For HL7 Format, just below here we do a split(), which negates the merge().    Originally I was thinking this pointless merge was causing a bug.  Now, I don't think it was any more, but I left the code that avoids the pointless merge.

prime-router/src/main/kotlin/azure/WorkflowEngine.kt
```diff
@@ -146,17 +146,17 @@ class WorkflowEngine(
...
-                nextEvent.eventAction,
-                nextEvent.at,
+                nextEvent!!.eventAction,
+                nextEvent!!.at,
                 retryJson,
                 txn
             )
         }
-        queue.sendMessage(nextEvent) // Avoid race condition by doing after txn completes.
+        if (nextEvent != null)  queue.sendMessage(nextEvent!!) // Avoid race condition by doing after txn completes.
```
jimduff-usds, line 159: Bug Fix:   This line was getting the Kotlin equiv of a NullPointerException, because the lateinit variable was not initialized if the "Weirdness" (above) occurs.    Its weirdness, but its not a bug/error.

prime-router/src/main/kotlin/cli/TestReportStream.kt
```diff
@@ -78,10 +80,25 @@ Examples:
...
+            "Default is $itemsDefault"
+    ).int().default(itemsDefault)
+
+    private val submitsDefault: Int = 5
+    private val submits by option(
+        "--submits",
+        metavar = "<int>",
+        help = "For tests involving multiple submits, do this many submissions. " +
+            "Default is $submitsDefault"
+    ).int().default(itemsDefault)
```
jimduff-usds, line 101: Can now specify the number of files to submit, and number of items per file, for some tests.

prime-router/src/main/kotlin/cli/TestReportStream.kt
```diff
@@ -685,7 +714,56 @@ class StracBasic : CoolTest() {
         val reportId = ReportId.fromString(tree["id"].textValue())
         echo("Id of submitted report: $reportId")
         waitABit(25, environment)
-        return examineLineageResults(reportId, listOf(redoxReceiver), fakeItemCount)
+        return examineLineageResults(reportId, listOf(redoxReceiver), items)
+    }
+}
+
+class HammerTime : CoolTest() {
+    override val name = "hammertime"
```
jimduff-usds, line 722: New test

prime-router/src/test/redox/initializerJson.json
```diff
@@ -18,7 +18,12 @@
       "headers": {
         "Content-Type": ["application/json"]
       },
-      "body": {"Meta":{"DataModel":"Results","EventType":"NewUnsolicited","Message":{"ID":5911323776},"Source":{"ID":"d89d4057-930f-4c5b-bb39-8cddb326e928","Name":"Prime Data Hub (Staging)"},"Destinations":[{"ID":"cd019512-8a58-40e5-a416-4e597404f9b8","Name":"CDC Pennsylvania Dept of Health Destination (s)"}]}}
+      "body": {"Meta":{"DataModel":"Results","EventType":"NewUnsolicited","Message":{"ID":5911323776},"Source":{"ID":"d89d4057-930f-4c5b-bb39-8cddb326e928","Name":"Prime Data Hub (Staging)"},"Destinations":[{"ID":"cd019512-8a58-40e5-a416-4e597404f9b8","Name":"CDC Pennsylvania Dept of Health Destination (s)"}]}},
+      "delay": {
+        "timeUnit": "SECONDS",
+        "value": 0
+      }
```
jimduff-usds, line 25: This timer is another super way to load test.   Simply change the "0" to a number of seconds to delay each redox call, and let loose the dogs of war.   Even a 2 or 3 second delay wreaks havoc.

MauriceReeves-usds (approved):

PR #809, opened 2021-04-13
Fixes issue with the synthesizer scrambling TOO MUCH data

MauriceReeves-usds:
This PR fixes an issue where the synthesizer is scrambling the data TOO much by recreating the row context for every single row in every single column. This ended up causing a problem where city, state, zip, and county would not match because they were all randomized.

Now, after the generation of the columns, we loop through the rows and set the city, state, zip, and county to the tied values created in the RowContext. This is obviously more expensive, but it's a more correct way to accomplish our goal. 

## Changes
- Adds new tests to verify that filtering of tables works
- Adds new test to verify that the chaining of lookup values in the `RowContext` works as expected
- Adds logic to walk the rows and set the address information to something that is correct

## Checklist
- [x] Tested locally?
- [x] Ran `quickTest all`?
- [x] Added tests?

### DIFF
--- a/prime-router/src/main/kotlin/Report.kt
+++ b/prime-router/src/main/kotlin/Report.kt
@@ -2,6 +2,7 @@ package gov.cdc.prime.router
 
 import gov.cdc.prime.router.azure.WorkflowEngine
 import gov.cdc.prime.router.azure.db.tables.pojos.ItemLineage
+import org.jooq.Meta
 import tech.tablesaw.api.StringColumn
 import tech.tablesaw.api.Table
 import tech.tablesaw.columns.Column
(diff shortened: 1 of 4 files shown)


prime-router/src/main/kotlin/Report.kt
```diff
@@ -378,6 +380,10 @@ class Report {
         targetCounty: String? = null,
         metadata: Metadata,
     ): Report {
+        fun maybeSetString(row: Row, columnName: String, value: String) {
```
carlosfelix2, line 383: Maybe a more descriptive name.  setRowValue?  setValueIfExists?  It is the maybe that confuses what the logic does.

jimduff-usds (approved):
lgtm.

lgtm.

There are a couple tests that create 10,000 row csvs (`./prime test --run huge,toobig`)  They are already removed from the regular smoke test regimen because they are painful to wait for, but I wonder if this will make them slower, or if the change is not noticeable.

Issue #815, opened 2021-04-14
HL7 Ingestion


MauriceReeves-usds:
A collection of issues around the HL7 ingestion via the CLI based on sample files provided by DataRobot.  The two HL7 test files are attached. The goal is to allow ReportStream to take in HL7 files from senders and then convert them into our internal Report format for routing and translation.

#### Note:
The files end in extension `txt` so they can be uploaded. Rename them to `hl7` in order to ingest them.

To convert a test file from HL7 into CSV to compare you would do this:

```
./prime data --input sample_file.hl7 --input-schema covid-19 --output-schema primedatainput/pdi-covid-19 --output-format CSV --output ./junk/sample_file.csv
```

[Sully File](https://app.zenhub.com/files/304423150/8cb0e840-ef06-47ea-aeba-885ea89220fc/download)
[Ellume File](https://app.zenhub.com/files/304423150/ab8d1f7e-f4a0-4ecc-8430-367a88477765/download)

PR #816, opened 2021-04-14
Adds in the HL7 ingestion into the CLI

MauriceReeves-usds:
This PR adds HL7 ingestion to the CLI. This is still very much in a rough shape and won't work for external files very well. I want to add this so we can start testing, and filing bugs against it.

## Changes
- Moves the `ProcessData` class to its own file for the CLI as we've done with other functional pieces
- Wires in the HL7 ingestion into the CLI based on file extension
- Fixes a few issues:
    - Formatting a phone number coming from HL7 broke `toFormatted` in `Element` because the assumption was the value would be normalized
    - Empty lines were causing the HL7 serializer to fail 
    - Warning messages were obscure. Added more detail to them

The actual functionality of the CLI isn't any different because the serializer is chosen based on the extension of the input file.

```
./prime data --input sample_file.hl7 --input-schema covid-19 --output-schema primedatainput/pdi-covid-19 --output-format CSV --output ./junk/sample_file.csv
```

### Jim's notes.

I ran
`./prime data --input ./junk/mytestfile.hl7 --input-schema covid-19 --route`
to prove I could read an hl7 file using the cli.  And
`curl -X POST -H "client:simple_report.hl7test" -H "content-type:text/hl7" --data-binary "@./junk/mytestfile.hl7" "http://localhost:7071/api/reports"`
to prove I could read an hl7 via ReportStream running in docker.

I used a single .hl7 testresult generated by our fake data generator

## Checklist
- [x] Tested locally?
- [x] Ran `quickTest all`?
- [ ] Added tests?

## To Be Done
- #815


prime-router/src/main/kotlin/Element.kt
```diff
@@ -288,7 +288,19 @@ data class Element(
...
+                    // remove parens from HL7 formatting
+                    listOf(
+                        normalizedValue
+                            .replace("(", "")
+                            .replace(")", ""),
+                        "1",    // country code
+                        ""      // extension
+                    )
+                }
+
```
MauriceReeves-usds, line 303: When we do the reverse direction of data in from HL7, phone number is in a different format. The HL7 serializer should, on ingest, put the phone number into the format we expect. But I wanted to put a guard around this logic because it never checks if the `:` is in the phone number or not.

prime-router/src/main/kotlin/cli/ProcessDataCommands.kt
```diff
@@ -0,0 +1,458 @@
...
+            when (format) {
+                Report.Format.INTERNAL -> csvSerializer.writeInternal(report, stream)
+                Report.Format.CSV -> csvSerializer.write(report, stream)
+                Report.Format.HL7 -> hl7Serializer.write(report, stream, hl7Configuration)
+                Report.Format.HL7_BATCH -> hl7Serializer.writeBatch(report, stream, hl7Configuration)
+                Report.Format.REDOX -> redoxSerializer.write(report, stream)
+            }
+        }
+    }
+}
```
MauriceReeves-usds, line 458: I moved the `ProcessData` class into a separate file to start cleaning up `main` and follow the pattern we're doing with other CLI actions.

prime-router/src/main/kotlin/cli/main.kt
```diff
@@ -5,444 +5,21 @@ package gov.cdc.prime.router.cli
...
 import gov.cdc.prime.router.Translator
-import gov.cdc.prime.router.serializers.CsvSerializer
-import gov.cdc.prime.router.serializers.Hl7Serializer
-import gov.cdc.prime.router.serializers.RedoxSerializer
 import java.io.File
-import java.io.OutputStream
-import java.net.HttpURLConnection
-import java.net.URL
 import java.nio.file.Files
 import java.nio.file.Paths
```
MauriceReeves-usds, line 21: In the process of moving `ProcessData` I am able to strip a bunch of imports from main. I don't think this will have any practical performance benefit, but it will at least make the code a little cleaner.

prime-router/src/main/kotlin/serializers/Hl7Serializer.kt
```diff
@@ -107,6 +107,10 @@ class Hl7Serializer(val metadata: Metadata) {
             val parsedMessage = convertMessageToMap(nextMessage.toString(), schema)
             errors.addAll(parsedMessage.errors)
             warnings.addAll(parsedMessage.warnings)
+            // there is a chance that there's an empty row (for example, an empty line)
+            // that won't parse. so we should skip that because it's not valid HL7
+            if (parsedMessage.row.isEmpty())
+                return
             rowResults.add(parsedMessage)
```
MauriceReeves-usds, line 114: What it says. An empty line could cause problems with processing, so just ignore them.

prime-router/src/main/kotlin/serializers/Hl7Serializer.kt
```diff
@@ -169,7 +173,7 @@ class Hl7Serializer(val metadata: Metadata) {
             }
             // add the rows
             if (parsedValue.isNullOrEmpty()) {
-                warnings.add("Blank for $terserSpec")
+                warnings.add("Blank for $terserSpec - $elementName")
             }
```
MauriceReeves-usds, line 177: Added the element name to the warning so it's easier to see what fields are empty.

prime-router/src/main/kotlin/serializers/Hl7Serializer.kt
```diff
@@ -181,64 +185,74 @@ class Hl7Serializer(val metadata: Metadata) {
...
-
-            if (it.hl7Field.isNullOrEmpty() && it.hl7OutputFields.isNullOrEmpty()) {
-                mappedRows[it.name]?.add("")
-                return@forEach
-            }
+        val cleanedMessage = reg.replace(message, "\r").trim()
+        // if the message is empty, return a row result that warns of empty data
+        if (cleanedMessage.isEmpty()) {
+            return RowResult(emptyMap(), emptyList(), listOf("Cannot parse empty message"))
+        }
```
MauriceReeves-usds, line 193: Add a check in the parsing step to return an empty row result before trying to invoke the parser.

prime-router/src/main/kotlin/serializers/Hl7Serializer.kt
```diff
@@ -181,64 +185,74 @@ class Hl7Serializer(val metadata: Metadata) {
...
+            errors.add("${e.localizedMessage} ${e.stackTraceToString()}")
         }
+
         // convert sets to lists
-        val rows = mappedRows.keys.map {
-            it to (mappedRows[it]?.toList() ?: emptyList())
-        }.toMap()
+        val rows = mappedRows.keys.associateWith {
+            (mappedRows[it]?.toList() ?: emptyList())
+        }
```
MauriceReeves-usds, line 256: `associateWith` is a nice piece of Kotlin magic that takes the `it` value and associates it with the return value from the closure and automagically creates a map out of it.

prime-router/src/main/kotlin/serializers/Hl7Serializer.kt
```diff
@@ -256,7 +270,11 @@ class Hl7Serializer(val metadata: Metadata) {
         val mappedRows = mapping.mappedRows
         errors.addAll(mapping.errors.map { ResultDetail(ResultDetail.DetailScope.ITEM, "", it) })
         warnings.addAll(mapping.warnings.map { ResultDetail(ResultDetail.DetailScope.ITEM, "", it) })
-        return ReadResult(Report(schema, mappedRows, source), errors, warnings)
+        mappedRows.forEach {
+            println("${it.key} -> ${it.value.joinToString()}")
+        }
+        val report = Report(schema, mappedRows, source)
+        return ReadResult(report, errors, warnings)
     }
```
MauriceReeves-usds, line 278: Spit out the mapped values on read. We should turn this into a debugging statement.

prime-router/src/main/kotlin/serializers/Hl7Serializer.kt
jimduff-usds: A lovely gem!    I found a Kotlin gem recently too, 'groupingBy' that essentially does the same as a 'group by' in sql.

prime-router/src/main/kotlin/serializers/Hl7Serializer.kt
jimduff-usds: Is this the evil line of code that makes my builds like 10,000 lines long?? ;)

prime-router/src/main/kotlin/cli/ProcessDataCommands.kt
jimduff-usds: Thank you, thank you, thank you!   

I actually wrote a 'ProcessData.kt' like a month ago, but then on the same day Rick did a zillion changes to the same code in main.kt, so I tossed mine into the bit bucket.  Very glad to see this coming to fruition.

jimduff-usds (approved):
Whole-heartedly approve

carlosfelix2 (approved):

Issue #793, opened 2021-04-12
Code a BlobStoreTransport.kt mechanism for HHS Protect


jimduff-usds:

jimduff-usds:
Done!

PR #821, opened 2021-04-15
Jim/blob store transport

jimduff-usds:
This PR adds the BlobStoreTransport and prepares to roll a draft Waters schema to Staging.

## Changes
- the new transport depends in having an environment variable `PartnerStorage` set, to parallel the current dependency on `AzureWebJobsStorage` for our "regular" blob storage account.  This assumes a partner blob storage setup per what @ronaldheft-usds has created.
- I have moved the new Waters schema from a different PR into this one, as this work needed that schema, for testing.
- created new organization and sender for waters, and new organization and receiver for hhsprotect.   They are currently under a different `topic`.   If we want them under the covid-19 topic, we'll need to do some work to part the waters from the other data, at least temporarily.   Note:  Currently this is just for local and Staging, NOT for production.
- separately, created a new Sender and Receiver under the test `ignore` org, for testing Waters.  Also under the same topic as the waters schema and hhsprotect.
- added a new test, `./prime test --run waters` that uses that new sender and receiver.  Note:  it also sends to hhsprotect right now, because hhsprotect has no filters.  
- added the `pii` field to every element in covid-19.schema that involves patient_* and ordering_provider_*.   I did this based on [this HHS guidance](https://www.hhs.gov/sites/default/files/covid-19-laboratory-data-reporting-guidance.pdf).

@Mr0grog , this new BlobStoreTransport is perhaps yet another model for writing a new Transport, as it really is mostly wrapper stuff - the actual "transport" is like one line of code.
@joemeree , we still need to do a thorough review and cleanup of the draft Waters schema.

## Checklist
- [X] Tested locally?
- [X] Ran `quickTest all`?
- [X] Added tests?

## Security
- no known issues

## Fixes
- #791 (first steps)
- #793 

## To Be Done
- @jlusds - Important:  We need someone to review every field in [the core cover-19 schema](https://github.com/CDCgov/prime-data-hub/blob/jim/BlobStoreTransport/prime-router/metadata/schemas/covid-19.schema) and confirm that the PII settings are correct per HHS Covid guidance.  This can be done after the PR goes in, because we're not ready to roll this to production anyway.


prime-router/metadata/schemas/WATERS/waters-covid-19.schema
```diff
@@ -0,0 +1,217 @@
+---
```
jimduff-usds, line 1: This is just a DRAFT.   Needs fine-tooth-comb walkthrough

prime-router/metadata/schemas/covid-19.schema
```diff
@@ -329,31 +329,36 @@ elements:
```
jimduff-usds, line 329: @jlusds , these "pii" settings need review by someone other than me.

MauriceReeves-usds (approved):

prime-router/settings/organizations-local.yml
```diff
@@ -835,6 +872,22 @@
           host: sftp
           port: 22
           filePath: ./upload
+      - name: BLOBSTORE
```
jimduff-usds, line 875: For testing the waters.

Issue #843, opened 2021-04-19
Add Sender Configuration to Senders


MauriceReeves-usds:
In order to allow for a sender to have hard-coded values that get added into the `Report` object on ingestion. 

Two examples are:

- Having a `processing_mode_code` set for senders who are still building
- Manually setting a CLIA or a lab name for a sender who has a schema that doesn't include it

We should add a Sender Configuration object on senders, in the same fashion we have `TranslationConfiguration` for outbound translations. It could potentially even inherit from `TranslationConfiguration`.

MauriceReeves-usds:
Moving to `In Progress` to start working on this.

jimduff-usds:
I used the field default feature as a workaround for the absence of this feature.  It is ok, but not perfect.   The main problem is that flipping a sender from "T" Training mode in production to "P" production mode in production currently requires a code release since its in the .schema, not in the settings.

Here's an example of how I do this for a Sender called MoHealth:

```
---
name: mohealth-covid-19
description: Every Day we provide Mo and Mo Health
topic: covid-19
trackingElement: specimen_id
basedOn: covid-19
extends: direct/direct-covid-19

elements:
  - name: sender_id
    default: MoHealth

  - name: processing_mode_code
    default: T

```

PR #845, opened 2021-04-20
Sender configuration

MauriceReeves-usds:
This PR adds a sender configuration object to a sender that gets set in the organizations or settings, and then is based into the default values collection that is then used to fill in values in the report on ingestion. 

## Changes
- Adds a `SenderConfiguration` class
- Adds the object to the `Sender` class
- Adds logic to fill in the default values from `SenderConfiguration` if they're passed in

## Checklist
- [x] Tested locally?
- [x] Ran `quickTest all`?
- [ ] Added tests?

## Fixes
- #843

### DIFF
--- a/prime-router/src/main/kotlin/SenderConfiguration.kt
+++ b/prime-router/src/main/kotlin/SenderConfiguration.kt
@@ -1,0 +1,10 @@
+package gov.cdc.prime.router
+
+import com.fasterxml.jackson.annotation.JsonCreator
+
+data class SenderConfiguration
+@JsonCreator constructor(
+    val processingModeCode: String? = null,
+    val testingLabName: String? = null,
+    val testingLabCLIA: String? = null
+)
--- a/prime-router/src/main/kotlin/azure/ReportFunction.kt
+++ b/prime-router/src/main/kotlin/azure/ReportFunction.kt
@@ -184,9 +184,23 @@ class ReportFunction {
                     return@mapNotNull null
                 }
                 Pair(parts[0], parts[1])
-            }.toMap()
+            }.toMap().toMutableMap()
         } else {
-            emptyMap()
+            mutableMapOf()
+        }
+
+        sender.configuration?.let {
+            if (!it.processingModeCode.isNullOrEmpty()) {
+                defaultValues["processing_mode_code"] = it.processingModeCode
+            }
+
+            if (!it.testingLabCLIA.isNullOrEmpty()) {
+                defaultValues["testing_lab_clia"] = it.testingLabCLIA
+            }
+
+            if (!it.testingLabName.isNullOrEmpty()) {
+                defaultValues["testing_lab_name"] = it.testingLabName
+            }
         }
 
         if (content.isEmpty() || errors.isNotEmpty()) {
--- a/prime-router/pom.xml
+++ b/prime-router/pom.xml
@@ -516,6 +516,11 @@
             <version>${jackson.version}</version>
         </dependency>
         <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-annotations</artifactId>
+            <version>${jackson.version}</version>
+        </dependency>
+        <dependency>
             <groupId>com.github.javafaker</groupId>
             <artifactId>javafaker</artifactId>
             <version>1.0.2</version>
--- a/prime-router/settings/organizations-local.yml
+++ b/prime-router/settings/organizations-local.yml
@@ -41,6 +41,8 @@
         topic: covid-19
         schemaName: tpca/tpca-covid-19
         format: CSV
+        configuration:
+          processingModeCode: Q
 
   - name: az-phd
     description: Arizona PHD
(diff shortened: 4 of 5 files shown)


prime-router/src/main/kotlin/SenderConfiguration.kt
```diff
@@ -0,0 +1,10 @@
+package gov.cdc.prime.router
```
jimduff-usds, line 1: Very minor note:  for similar "sub-objects" in Receiver, the data class is right inside Receiver, not in a separate class.

prime-router/src/main/kotlin/SenderConfiguration.kt
```diff
@@ -0,0 +1,10 @@
...
+
+import com.fasterxml.jackson.annotation.JsonCreator
+
+data class SenderConfiguration
+@JsonCreator constructor(
+    val processingModeCode: String? = null,
+    val testingLabName: String? = null,
+    val testingLabCLIA: String? = null
```
jimduff-usds, line 9: Don't need to do this now, but I suggest we refactor to make this generic.   That is instead of 

```
configuration:
  processingModeCode: "T"
  testingLabName: "Mo and Mo Testing"
```

we could do something along the lines of:
```
configuration:
    - action: additional-item
      name: processing_mode_code
      value: "T"
    - action: additional-item
      name: testingLabName
      value: "Mo and Mo Testing"
    - action: force-deidentify    // action is a dumb name, but the idea is you might do other things besides add default values. 
```

Another option would be to look at following the same syntax as the default option on http parameters, eg something like
`   - default:  "processing_mode_code:T"
`
However, the ":" might throw yaml into conniption fits without the quotes?

If we do this here, we should also do it for other places where default values are set in our yaml, so we keep a simple consistent syntax.

prime-router/src/main/kotlin/azure/ReportFunction.kt
```diff
@@ -184,9 +184,23 @@ class ReportFunction {
...
                 Pair(parts[0], parts[1])
-            }.toMap()
+            }.toMap().toMutableMap()
         } else {
-            emptyMap()
+            mutableMapOf()
+        }
+
+        sender.configuration?.let {
+            if (!it.processingModeCode.isNullOrEmpty()) {
```
jimduff-usds, line 193: So now that I think about this:   this entire feature is in lieu of just have two schemas, one with, for example, the default 'T' and the other with the default 'P' for processing_mode_code.

I can't remember if @RickHawesUSDS  schema inheritance allows for multiple levels, but if it does, then that's a way to do this cleanly, with two child schemas of the Sender schema.

To be honest I kinda like that way better than this, because it leaves all our data schema stuff in one place, in the schema/ directory, instead of some secreted away inside the settings table.

jimduff-usds (changes requested):
If we're going to do it this way, I do like the elegance of riding on the 'default' feature that already exists.

However, per my comment on ReportFunction.kt, I'm wondering if using schema inheritance might be a cleaner and clearer way to do this?

Feel free to override, and I'll just approve.

prime-router/src/main/kotlin/SenderConfiguration.kt
MauriceReeves-usds: Good point. I was looking at it like the TranslatorConfiguration class which can be overloaded and lives separately. I was thinking there might be a chance we ended up needing to subclass this in the future, but I could be overthinking it.

prime-router/src/main/kotlin/SenderConfiguration.kt
MauriceReeves-usds: I did think about making it more generic, and allowing us to add whatever we wanted into the defaults collection, and I do like the open-ended nature of this, but I was also concerned with making it too easy to shoot ourselves in the foot.

prime-router/src/main/kotlin/azure/ReportFunction.kt
MauriceReeves-usds: I also considered this, but the problem is that if we end up ingesting HL7 (and it looks like based on data from Texas and/or Waters we could) the data in the schema wouldn't be taken into consideration. For inbound and outbound HL7 messages, it's either an `ORU_R01` message or it isn't.

The only thing we use the schema for is to iterate through the fields in the schema and look for HL7 elements. 

I guess we could modify the HL7 serializer to look for defaults on ingest and to use that if the `Terser` doesn't find a value.

My only other reservation with this is that we currently have to deploy changes for schemas still through a full deployment, and that means a longer lead time to getting everything out. If we can move the schema up into Postgres like the settings, then that would make this better.

jimduff-usds: Both good points.     I do think you should be able to set defaults and expect them to be used, whether the data is HL7 or CSV.  So I think your suggestion to have the HL7 serializer look for defaults is probably a requirement.

prime-router/src/main/kotlin/SenderConfiguration.kt
jimduff-usds: Agreed.  Let's leave as is - if this turns out to be widely used, then we genericize.

jimduff-usds (approved):

PR #847, opened 2021-04-20
New load test; logging

jimduff-usds:
This PR adds a new load test and logging
## Changes
- Added logging during blobstore transport
- Added a load test.  Works up to 100,000 Items / hour on local laptop runs.
- Increased default max reports per merge operation to 500 from 100, in preparation for further load testing.

## Checklist
- [X] Tested locally?
- [X] Ran `quickTest all`?
- [X] Added tests?

## Security
- none known

### DIFF
--- a/prime-router/src/main/kotlin/cli/TestReportStream.kt
+++ b/prime-router/src/main/kotlin/cli/TestReportStream.kt
@@ -36,7 +36,11 @@ import java.nio.file.Files
 import java.nio.file.Paths
 import java.time.OffsetDateTime
 import kotlin.concurrent.thread
+import kotlin.random.Random
 import kotlin.system.exitProcess
+import kotlin.time.Duration
+import kotlin.time.ExperimentalTime
+import kotlin.time.measureTime
 
 enum class TestStatus(val description: String) {
     DRAFT("Experimental"), // Tests that are experimental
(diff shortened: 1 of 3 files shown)


prime-router/src/main/kotlin/cli/TestReportStream.kt
```diff
@@ -734,56 +757,107 @@ class Waters : CoolTest() {
...
         echo("Id of submitted report: $reportId")
-        waitABit(25, environment)
-        return examineLineageResults(reportId, listOf(blobstoreReceiver), items)
+        waitABit(25, environment, options.muted)
+        if (file.exists()) file.delete()  // because of RepeatWaters
+        return examineLineageResults(reportId, listOf(blobstoreReceiver), options.items)
     }
 }
 
+class RepeatWaters : CoolTest() {
```
jimduff-usds, line 787: All the other changes in this file are to support this change:   Adding this new `repeatwaters` test, which repeatedly sends submits to the waters schema.  

As set at the moment, it sends a new Report every 360ms, with a little randomization thrown in.   So if you put 10 Items in each report, that's a pace of 100,000 Items per hour.     This appears to be a sustainable pace on localhost - I ran this with no problems:
` ./prime test --run repeatwaters --submits 2500 --items 10`

This test differs from hammertime in that hammer time submits all its reports all at once, whereas this is a more realistic scenario with reports coming in at random times over an extended time period.

ronaldheft-usds (approved):

PR #912, opened 2021-04-22
Improves HD truncation logic for VT

MauriceReeves-usds:
This PR expands on the HD truncation logic for VT. In some HL7 segments, components are of type `HD` which has a max length of 20 for the namespace ID: (`MSH-4`, `ORC-2`, `OBR-2`, etc). Some jurisdictions want to make sure that the first subcomponent for the HD fields is truncated to 20 char. While some of our fields in our schemas are encoded as type `HD`, some are not, and therefore it was necessary to flag some `TEXT` fields as actually `HD` fields that need to be truncated.

## Changes
- Adds an `HD_FIELDS` list to the companion object of the `Hl7Serializer`
- Adds logic to truncate the fields if their length is greater than the 20 char limit and the truncation flag is true.

## Checklist
- [x] Tested locally?
- [x] Ran `quickTest all`?
- [ ] Added tests?


RickHawesUSDS (approved):
Looks good.

PR #918, opened 2021-04-23
Make STRAC data feed work for all states

RickHawesUSDS:
This PR fixes a bug where a report from STRAC would fail on many states. 

## Changes
- Added mappings to the STRAC schema to provide required data elements
- Moved a mapping for PA from the STRAC schema to PA schema

## Checklist
- [x] Tested locally?
- [x] Ran `quickTest all`?
- [x] Ran `./prime test` against local Docker ReportStream container?
- [x] Downloaded a file from http://localhost:7071/api/download
- [ ] Updated the release notes? 
- [ ] Added tests?
- [ ] Did you check for sensitive data, and remove any? 
- [ ] Does logging contain sensitive data?  
- [ ] Are additional approvals needed for this change?
- [ ] Are there potential vulnerabilities or licensing issues with any new dependencies introduced?

## Other testing
Generated fake STRAC data for AZ. This data failed before because of missing elements. Now it succeeds.
```
rickhawes@MacBook-Pro-16 ~/P/p/prime-router (rick/fix-strac-to-allstates)> ./prime data --input result_files/fake-az-strac.csv --input-schema=strac/strac-covid-19 --route --output-dir routed_files
10:37:19.263 [main][1] INFO  ca.uhn.hl7v2.util.Home - hapi.home is set to /Users/rickhawes/Projects/prime-data-hub/prime-router/.
10:37:19.274 [main][1] INFO  ca.uhn.hl7v2.VersionLogger - HAPI version is: 2.3
10:37:19.277 [main][1] INFO  ca.uhn.hl7v2.VersionLogger - Default Structure libraries found for HL7 versions 2.5.1,
Loaded schema and receivers
Opened: /Users/rickhawes/Projects/prime-data-hub/prime-router/result_files/fake-az-strac.csv
Creating these files:
/Users/rickhawes/Projects/prime-data-hub/prime-router/routed_files/az-covid-19-917da57a-3eed-4e28-9ef2-bd02a91700a0-20210423103719.csv
/Users/rickhawes/Projects/prime-data-hub/prime-router/routed_files/covid-19-295a5bf6-076a-4e8e-a0d9-3a2190028c34-20210423103719.hl7
/Users/rickhawes/Projects/prime-data-hub/prime-router/routed_files/pima-az-covid-19-ef4184a1-a1dc-48f8-aa9d-5686848e86c3-20210423103719.csv
rickhawes@MacBook-Pro-16 ~/P/p/prime-router (rick/fix-strac-to-allstates)>
```

## Security
None

[...]

PR #922, opened 2021-04-23
Onboard All the Things!

MauriceReeves-usds:
This PR adds the remaining states to the organizations. 

## Checklist
- [x] Tested locally?
- [x] Ran `quickTest all`?
- [ ] Added tests?

### DIFF
--- a/prime-router/src/main/kotlin/cli/main.kt
+++ b/prime-router/src/main/kotlin/cli/main.kt
@@ -256,14 +256,22 @@ class ProcessData : CliktCommand(
                 val outputFile = if (outputFileName != null) {
                     File(outputFileName!!)
                 } else {
-
+                    // is this config HL7?
+                    val hl7Config = report.destination?.translation as? Hl7Configuration?
+                    // if it is, get the test processing mode
+                    val processingMode = if (hl7Config?.useTestProcessingMode == false) {
+                        "P"
+                    } else {
+                        "T"
+                    }
                     val fileName = Report.formFilename(
                         report.id,
                         report.schema.baseName,
                         format,
                         report.createdDateTime,
                         getNameFormat(Report.NameFormat.STANDARD),
-                        receivingOrganization ?: report.destination?.translation?.receivingOrganization
+                        receivingOrganization ?: report.destination?.translation?.receivingOrganization,
+                        processingMode
                     )
                     File(outputDir ?: ".", fileName)
                 }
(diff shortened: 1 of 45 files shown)


jimduff-usds (approved):
I wish they all said 01:15 instead of 09:15 - but it doesn't actually affect the timing of anything, it would just satisfy some latent OCD of mine.  I studied the 0015-all-the-rest file pretty well and didn't see other problems.

I think this release MUST get combined with Rick's #918 fix to allow strac data to work in non-strac states.  And even so, let's do some testing where we generate data across all 60 states.

Spreading out the timings of transports is our #1 cheapest way to scale this thing.  Although that conflicts with the goal of having a wide open window at 9:15am to 11am.

prime-router/src/main/kotlin/serializers/Hl7Serializer.kt
```diff
@@ -514,19 +539,31 @@ class Hl7Serializer(val metadata: Metadata) {
...
-        }
+    private fun setEmailComponent(terser: Terser, value: String, element: Element, hl7Config: Hl7Configuration?) {
+        // branch on element name. maybe we'll pass through ordering provider email information as well
         if (element.nameContains("patient_email")) {
-            // this is an email address
-            terser.set("/PATIENT_RESULT/PATIENT/PID-13($rep)-2", "NET")
-            // specifies it's an internet telecommunications type
-            terser.set("/PATIENT_RESULT/PATIENT/PID-13($rep)-3", "Internet")
-            terser.set("/PATIENT_RESULT/PATIENT/PID-13($rep)-4", value)
+            // for some state systems, they cannot handle repetition in the PID-13 field, despite what
```
RickHawesUSDS, line 545: Looks good.

prime-router/src/main/kotlin/cli/main.kt
```diff
@@ -256,14 +256,22 @@ class ProcessData : CliktCommand(
...
                     val fileName = Report.formFilename(
                         report.id,
                         report.schema.baseName,
                         format,
                         report.createdDateTime,
                         getNameFormat(Report.NameFormat.STANDARD),
-                        receivingOrganization ?: report.destination?.translation?.receivingOrganization
+                        receivingOrganization ?: report.destination?.translation?.receivingOrganization,
+                        processingMode
                     )
```
RickHawesUSDS, line 275: I guess. I don't understand why this code doesn't use the FileUtilities function here, instead of repeating it. Or the new Report.formFileName function.

RickHawesUSDS (approved):
One concern about duplicate code, but love the all the state checkin.

PR #923, opened 2021-04-23
Fix for Redox bug, fix to Mappers bug, minor test cleanup

### LINKED ISSUE #920: Error: Two records did not send to Redox this morning due to missing serialization of a Test_result_status value
Error: Two CO records did not send to Redox this morning.

The Redox error on redoxengine website was "`Required Field Missing - Orders[].Results[].Status In Results:NewUnsolicited"`
Cause: the SimpleReport Test_result_status field was 'X', meaning "Order Cancelled"
valueset is hl70123
Normally that field is 'F' for a regular "Final" test result, sometimes its 'C'.   And now 'X', which our code does not handle.

We didn't know how to serialize the 'X', so its a null in the redox json.
Note:  The Orders[].Results[].ResultStatus is also null, 

Supported values for Orders[].Results[].Status are:
```
Final		 
Preliminary	 
Corrected	 
Canceled	 
Unavailable	 
-- These three are in Orders[].Results[].Status, but not in Orders[].ResultStatus, so we won't use them.
-- Incomplete	 
-- None Specified	 
-- Deleted

```
SupportValues for Orders[].ResultStatus are:
```
Final	      
Preliminary   
-- Not in Orders[].Results[].Status:    In Process    
Corrected     
Canceled      
Unavailable
```

So the intersection of available status values in Redox that cover both these fields is:
```
Final	      
Preliminary   
Corrected     
Canceled      
Unavailable

```
The mapping in covid-19-redox.schema looks like this:
```
  - name: test_result_status
    redoxOutputFields: ["Orders[0].ResultsStatus", "Orders[0].Results[0].Status"]
    altValues:
      - display: Corrected
        code: C
      - display: Final
        code: F

```
And the hl70123 valueset looks like this:
```
    - code: A
      display: Some, but not all, results available
    - code: C
      display: Corrected, final
    - code: F
      display: Final results
    - code: I
      display: No results available; specimen received, procedure incomplete
    - code: M
      display: Corrected, not final
    - code: N
      display: Procedure completed, results pending
    - code: O
      display: Order received; specimen not yet received
    - code: P
      display: Preliminary
    - code: R
      display: Results stored; not yet verified
    - code: S
      display: No results available; procedure scheduled, but not done
    - code: X
      display: No results available; Order canceled
    - code: Y
      display: No order on record for this test
    - code: Z
      display: No record of this patient
```

**FINALLY, HERE IS THE FIX**:  So based on those codes I'm changing the altValues so it looks like this:

```
  - name: test_result_status
    redoxOutputFields: ["Orders[0].ResultsStatus", "Orders[0].Results[0].Status"]
    altValues:
      - display: Corrected
        code: C
      - display: Final
        code: F
      - display: Canceled
        code: X
      - display: Preliminary
        code: A
      - display: Unavailable
        code: I
      - display: Corrected
        code: M
      - display: Preliminary
        code: N
      - display: Preliminary
        code: O
      - display: Preliminary
        code: P
      - display: Preliminary
        code: R
      - display: Unavailable
        code: S
      - display: Unavailable
        code: Y
      - display: Unavailable
        code: Z

```

jimduff-usds:
This PR fixes #920 

## Changes
- Added altValues to the base covid-19-redox schema to cover the 'X' value coming from SimpleReport.    I went ahead and added all the other values too - not clear if they will ever be used.
- Fixed a bug I saw in the mapper while messing with the CO data.
- minor cleanup to TestReportStream

## Checklist
- [X] Tested locally?
- [X] Ran `quickTest all`?
- [X] Added tests?

## Security
None known at this time

## Fixes
- #920

### DIFF
--- a/prime-router/metadata/schemas/covid-19-redox.schema
+++ b/prime-router/metadata/schemas/covid-19-redox.schema
@@ -179,6 +179,28 @@ elements:
         code: C
       - display: Final
         code: F
+      - display: Canceled
+        code: X
+      - display: Preliminary
+        code: A
+      - display: Unavailable
+        code: I
+      - display: Corrected
+        code: M
+      - display: Preliminary
+        code: N
+      - display: Preliminary
+        code: O
+      - display: Preliminary
+        code: P
+      - display: Preliminary
+        code: R
+      - display: Unavailable
+        code: S
+      - display: Unavailable
+        code: Y
+      - display: Unavailable
+        code: Z
 
   - name: ordered_test_code
     redoxOutputFields: ["Orders[0].Procedure.Code"]
--- a/prime-router/src/main/kotlin/Mappers.kt
+++ b/prime-router/src/main/kotlin/Mappers.kt
@@ -80,6 +80,8 @@ class MiddleInitialMapper : Mapper {
             return null
         } else {
             if (values.size != 1) error("Found ${values.size} values.  Expecting 1 value. Args: $args, Values: $values")
+            if (values.first().value.isEmpty())
+                return null
             return values.first().value.substring(0..0).toUpperCase()
         }
     }
--- a/prime-router/src/main/kotlin/cli/TestReportStream.kt
+++ b/prime-router/src/main/kotlin/cli/TestReportStream.kt
@@ -45,7 +45,6 @@ import kotlin.time.measureTime
 enum class TestStatus(val description: String) {
     DRAFT("Experimental"), // Tests that are experimental
     FAILS("Always fails"), // For tests that just always fail, and we haven't fixed the issue yet.
-    SLOW("Good test, but slow"), // For tests that are crazy slow
     LOAD("Load Test"),
     GOODSTUFF("Part of Smoke test"), // Only Smoke the Good Stuff.
 }
@@ -378,7 +377,7 @@ abstract class CoolTest {
         }
 
         /**
-         * A hack attempt to wait enough time, but not too long, for Hub to finish.
+         * A hack attempt to wait enough time, but not too long, for ReportStream to finish.
          * This assumes the batch/send executes on the minute boundary.
          */
         fun waitABit(plusSecs: Int, env: ReportStreamEnv, silent: Boolean = false) {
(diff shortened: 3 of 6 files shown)


prime-router/metadata/schemas/covid-19-redox.schema
```diff
@@ -179,6 +179,28 @@ elements:
         code: C
       - display: Final
         code: F
+      - display: Canceled
+        code: X
+      - display: Preliminary
```
jimduff-usds, line 184: The 'X' fixes the immediate bug.   I went ahead and added all these others.

prime-router/src/main/kotlin/Mappers.kt
```diff
@@ -80,6 +80,8 @@ class MiddleInitialMapper : Mapper {
             return null
         } else {
             if (values.size != 1) error("Found ${values.size} values.  Expecting 1 value. Args: $args, Values: $values")
+            if (values.first().value.isEmpty())
+                return null
             return values.first().value.substring(0..0).toUpperCase()
```
jimduff-usds, line 85: This line expects a string of length 1.  I had a case where there was a non-null string, but it was empty, which causes an exception here.

prime-router/src/main/kotlin/cli/TestReportStream.kt
```diff
@@ -378,7 +377,7 @@ abstract class CoolTest {
         }
 
         /**
-         * A hack attempt to wait enough time, but not too long, for Hub to finish.
```
jimduff-usds, line 381: deHubbing.

prime-router/src/main/kotlin/cli/TestReportStream.kt
```diff
@@ -1026,12 +1025,12 @@ class TooBig : CoolTest() {
...
  * This test, when it fails, exposes a database connection exception in Staging.
- *
+ * I think this is actually passing now, but the query isn't quite right.
  */
 class DbConnections : CoolTest() {
     override val name = "dbconnections"
-    override val description = "Test weird issue wherein many 'sends' cause db connection failures"
-    override val status = TestStatus.FAILS
+    override val description = "Test issue wherein many 'sends' caused db connection failures"
+    override val status = TestStatus.DRAFT
```
jimduff-usds, line 1034: This test now passes, but is kinda redundant with others.

prime-router/src/main/kotlin/Mappers.kt
jimduff-usds: Actually, just googling substring in Kotlin, I think the above always creates an empty string.

Has this MiddleEarthMapper ever worked??

MauriceReeves-usds (approved):

Issue #236, opened 2021-01-04
Make schema validator step and have prime-cli call it.


RickHawesUSDS:
[ ] Check that each element has a Type

MauriceReeves-usds:
@carlosfelix2 I was thinking that this might be another good thing you could work on. Currently we validate a schema by adding it to the `quick-test.sh` file and then running data through it, but it would be good to actually try to automate that and allow us to call `validate` as an option of the prime executable. 

### Schema Validation

I was thinking something like:

```
./prime validate --all-schemas
```

and that would load the metadata catalog as a first step (since schemas are loaded as part of the metadata) and then would attempt to run fake data through all the schemas in both the `CSV` and `HL7_BATCH` formats. The thing to keep in mind here is that some schemas don't have CSV fields defined, so it's possible to run data through the `CSV` format and have it return a zero-length document. That shouldn't necessarily be a failure.

I would also add an option to let us validate a single schema like:

```
./prime validate --schema <SCHEMA_NAME>
```

And in addition to say :

```
./prime validate --schema <SCHEMA_NAME> --output-format CSV
```

And have it provide warnings if an element in the schema doesn't have a `csvFields` value defined. This is only a warning because we include some fields into schemas so they can be consumed by mappers. When serializing elements we actually take two passes through the document to allow some fields to be loaded and THEN consumed by mappers on pass two. For example, you might want to do a 

```
- name: local_patient_name
   type: TEXT
   csvFields:
     - name: patient_name
   mapper: concat(patient_first_name, patient_last_name)

# standard fields consumed by mapper
- name: patient_first_name

- name: patient_last_name
```

Admittedly, this is a contrived example, but examples like this do exist. But it would be good to have it say something like:

```
WARNINGS:
- "patient_first_name" doesn't have a csvFields value defined. Is this on purpose?
- "patient_last_name" doesn't have a csvFields value defined. Is this on purpose?
```

Another area that we want this to help validate is the creation of `altValues` for a schema. We want to make sure that the codes an altValue maps to are actual codes defined for the value set.

### Organization and Receiver Validation
While the original ticket calls for just schema validation, I was also thinking it would be good to have this check organization and receiver files as well. The biggest areas where @jimduff-usds and I (mostly me) have tripped ourselves up are:

- Leaving the frequency for merging batches at 1440 per day for the prod file (where we would not want to batch every minute)
- Referencing jurisdictional filters that don't exist. I always write `or` when I mean `orEquals`.
[...]

PR #959, opened 2021-04-29
Jim/50 states

jimduff-usds:
This PR does a few minor data tweaks that fix outstanding issue preventing certain input data from going to certain output state receivers.

## Changes
- various fixes in strac and NM schemas.
- moved ordering_facility_state to optional in waters, per discussion with them.
- The `--input-fake `option now generates data for all ~60 states and territories in a round-robin fashion.  Override this using the `--target-states` option.

Here's how you can test this pretty easily:

```

./prime data --input-fake 1000 --input-schema waters/waters-covid-19 --output ./junk/from-waters-to-all-states.csv
./prime data --input-fake 1000 --input-schema primedatainput/pdi-covid-19 --output ./junk/from-sr-to-all-states.csv
./prime data --input-fake 1000 --input-schema strac/strac-covid-19 --output ./junk/from-strac-to-all-states.csv

curl -X POST -H "client:strac"  -H "content-type:text/csv" --data-binary "@./junk/from-strac-to-all-states.csv" "http://localhost:7071/api/reports"
curl -X POST -H "client:waters"  -H "content-type:text/csv" --data-binary "@./junk/from-waters-to-all-states.csv" "http://localhost:7071/api/reports"
curl -X POST -H "client:simple_report"  -H "content-type:text/csv" --data-binary "@./junk/from-sr-to-all-states.csv" "http://localhost:7071/api/reports"

```
although right now you need to eyeball the results or do data

### Testing
- [X] Tested locally?
- [X] Ran `quickTest all`?
- [X] Added tests?    Well, I listed them above.  😬 

### Security

### Process

## Security
- no issues known

## To Be Done
Per discussion in #236, write an automated check to confirm this continues to work.


prime-router/metadata/schemas/NM/nm-covid-19-csv.schema
```diff
@@ -269,11 +269,11 @@ elements:
...
     altValues:
-      - code: Y
+      - code: 77386006
         display: Y
-      - code: N
+      - code: 60001007
         display: N
-      - code: UNK
+      - code: 261665006
         display: U
```
jimduff-usds, line 277: Bug fix:   The "code" values (Y, N, UNK) were not correct.  They need to match to the weird number values in the valueset.   I think we don't enforce that right now, so we only find out at runtime when unknown code values appear, eg, as generated by the fake data generator.

prime-router/metadata/schemas/STRAC/strac-covid-19.schema
```diff
@@ -112,14 +112,52 @@ elements:
     csvFields: [{name: patient_zip, format: $zipFive}]
 
   - name: test_result
```
jimduff-usds, line 114: Bug fix:  The fake data generator was creating code values that the minimalist alt valueset didn't have.

As long as strac continues to only send the first 3 values, the newly added values should be harmless.

prime-router/metadata/schemas/WATERS/waters-covid-19.schema
```diff
@@ -30,7 +30,7 @@ elements:
   # I added this and made it required so we remember to discuss.
   # Very important field for routing data to states.
   - name: ordering_facility_state
-    cardinality: ONE
+    cardinality: ZERO_OR_ONE
     csvFields: [{ name: ordering_facility_state}]
```
jimduff-usds, line 34: per discussion with the waters team.  They don't currently have ordering_facility_state; some states require this, in order to filter properly.

prime-router/src/main/kotlin/FakeReport.kt
```diff
@@ -294,7 +294,12 @@ class FakeReport(val metadata: Metadata) {
         targetCounties: String? = null
     ): Report {
         val counties = targetCounties?.split(",")
-        val states = targetStates?.split(",")
+        val states = if (targetStates.isNullOrEmpty()) {
+            metadata.findLookupTable("fips-county")?.getDistinctValuesInColumn("State")
+                ?.toList()
+        } else {
+            targetStates?.split(",")
+        }
```
jimduff-usds, line 302: By default now, generate all the two letter codes in the firs-county table.

prime-router/src/main/kotlin/azure/WorkflowEngine.kt
```diff
@@ -97,9 +97,17 @@ class WorkflowEngine(
...
     ) {
-        val blobInfo = blob.uploadBody(report)
+        val blobInfo = try {
+            // formatting errors can occur down in here.
+            blob.uploadBody(report)
+        } catch (ex: Exception) {
+            context?.logger?.warning("Got exception while dispatching to schema ${report.schema.name}" +
+                ", and rcvr ${receiver.fullName}")
+            throw ex
+        }
```
jimduff-usds, line 110: Added this logging because low level errors were occurring down in the bowels of the Element.kt, but it was impossible to tell which schema/receiver the errors were for.

It feels odd to me that the formatting of data for export occurs down inside a method called 'blob.uploadBody'.

prime-router/metadata/schemas/NM/nm-covid-19-csv.schema
RickHawesUSDS: These ALT values match the real LOINC values. Why do we need them? Suggest you change the format to be $code? Or is there something else that I'm missing?

prime-router/metadata/schemas/WATERS/waters-covid-19.schema
```diff
@@ -74,8 +74,19 @@ elements:
...
+    altValues:
+      - code: H
+        display: 2135-2
+      - code: N
+        display: 2186-5
+      - code: U
+        display: UNK
+      - code: U
+        display: ASKU
+    csvFields: [{ name: patientEthnicity, format: $alt}]
```
RickHawesUSDS, line 90: I fear that having to same code for "U" for both UNK and ASKU will lead to problems. 

There appears to be OMB vs ELR community argument about what is the right value set to use. PHINVADS and the confluence board should be the source of truth. From the notes, on the confluence board, the recommendation may expand to include

prime-router/src/main/kotlin/LookupTable.kt
```diff
@@ -97,6 +97,17 @@ class LookupTable(
...
 
+    /**
+     * Get the set of distinct values in a column in the table.
+     */
+    fun getDistinctValuesInColumn(selectColumn: String): Set<String> {
+        val selectColumnNumber = headerIndex[selectColumn.toLowerCase()] ?: return emptySet()
+        return table
+            .map { row -> row[selectColumnNumber] }
+            .filter { !it.equals(selectColumn) }   // terrible hack to remove the column header.
+            .toSet()
```
RickHawesUSDS, line 108: perhaps a .drop(1) would be a clearer way to remove the header row.

RickHawesUSDS (changes requested):
Looks good overall. A couple of detail items.

prime-router/metadata/schemas/NM/nm-covid-19-csv.schema
jimduff-usds: The valueset we have in covid-19.schema has Pregnant, Not Pregnant, and Unknown.
These appear to be the correct LOINC codes.  See https://loinc.org/82810-3/

So, New Mexico appears to want Y, N, UNK, which does not appear to be the LOINC enumeration.   Hence the need for an $alt valueset.

prime-router/metadata/schemas/WATERS/waters-covid-19.schema
jimduff-usds: ReportStream currently uses hl70189 as its INTERNAL representation, which has only 3 values.

Waters sends omb_ethnicity, which has four values.    Waters is probably right.  We should switch covid-19.schema to use omb_ethnicity.

But that means a ton of changes both inbound and outbound, that I'm not ready to work on right now.  This is a reasonable workaround.    For incoming data, mapping both UNK and ASKU is the best we can do --- there's no way to drop from 4 values to 3 without a loss of information.

prime-router/metadata/schemas/NM/nm-covid-19-csv.schema
RickHawesUSDS: ok

prime-router/metadata/schemas/WATERS/waters-covid-19.schema
RickHawesUSDS: Right. Why don't we move to OMB for our internal?

jimduff-usds: I agree that's the right approach, but we'd need to carefully test that both upstream and downstream data are not affected.    At the moment, let's get the waters schema out the door and working.

prime-router/src/main/kotlin/LookupTable.kt
jimduff-usds: Done!

Issue #955, opened 2021-04-28
Move Texas off of SHA-1


RickHawesUSDS:
**Describe the bug**
Texas is using SHA-1 which is an unsafe encryption. 

**Impact**
SHA-1 is deprecated and considered unsafe encryption. 
This is listed as a Critical (must be addressed in 14 days) security incident. 

**To Reproduce**

**Expected behavior**
Feed to Texas is on SHA-256 or shutdown. Legacy transport code can be removed in a follow-on ticket. 

**Screenshots**

**Logs**
If applicable, please attach logs to help describe your problem.

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

PR #961, opened 2021-04-29
Hide SFTP Disconnect Promise Timeout Failures

MauriceReeves-usds:
This PR changes the pattern slightly for the disconnection after executing an `ls`. I did extensive debugging over the last several hours, including pointing my local docker image at both TX and AL, and then debugging within the docker container to see where the timeout exception is being thrown.

In the case of both servers, the call to `ls` actually succeeds, but then when it attempts to call the disconnect and close methods, SSHJ throws a timeout when it takes too long for the server to report it has disconnected.  This is apparently a persistent and nagging issue that has been around for a long time, and has been reported by several people on the repo, but it's never been addressed:

- https://github.com/hierynomus/sshj/issues/510
- https://github.com/hierynomus/sshj/issues/452

Given that fact, I'm advocating we swallow and ignore just the Timeout exception happening and move on.  It is an imperfect solution, and I'm not going to lie, I considered digging into code for SSHJ and issuing a PR to fix this, but I don't currently have the bandwidth to accomplish that.

## Changes
- Adds yamlito to make Texas' transport be SFTP
- Updates their jurisdictional filter to include patient_state as well
- Wraps and ignores TimeoutExceptions 

### Testing
- [x] Tested locally?
- [x] Ran `quickTest all`?
- [ ] Added tests?

### Security

### Process

## Fixes
- #955


prime-router/pom.xml
```diff
@@ -503,6 +504,11 @@
             <artifactId>jackson-databind</artifactId>
             <version>${jackson.version}</version>
         </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-annotations</artifactId>
+            <version>${jackson.version}</version>
+        </dependency>
         <dependency>
```
MauriceReeves-usds, line 512: I was getting weird intellisense errors until I added annotations to the pom file as well.

prime-router/pom.xml
```diff
@@ -583,6 +589,10 @@
                     <groupId>com.github.kittinunf.fuel</groupId>
                     <artifactId>fuel</artifactId>
                 </exclusion>
+                <exclusion>
+                    <groupId>org.jetbrains.kotlinx</groupId>
+                    <artifactId>kotlinx-coroutines-core</artifactId>
+                </exclusion>
             </exclusions>
```
MauriceReeves-usds, line 596: I had to update the pom file to tell mock-fuel to not try to fulfill this requirement.

prime-router/pom.xml
```diff
@@ -591,6 +601,11 @@
             <version>2.3.1</version>
             <scope>runtime</scope>
         </dependency>
+        <dependency>
+            <groupId>org.jetbrains.kotlinx</groupId>
+            <artifactId>kotlinx-coroutines-core</artifactId>
+            <version>1.4.3</version>
+        </dependency>
         <dependency>
```
MauriceReeves-usds, line 609: And then I downloaded the reference here instead.

prime-router/src/main/kotlin/transport/SftpTransport.kt
```diff
@@ -137,15 +137,20 @@ class SftpTransport : ITransport {
...
-                    val lsResponse = it.ls(path)
-                    return lsResponse.map { l -> l.toString() }.toList()
+                sshClient.use { sshClient ->
+                    sshClient.newSFTPClient().use {
+                        it.ls(path).map { l -> lsResults.add(l.toString()) }
+                    }
                 }
+            } catch (_: java.util.concurrent.TimeoutException) {
+                // do nothing. some servers just take a long time to disconnect
             } finally {
```
MauriceReeves-usds, line 149: This is where the magic happens kids. Just eat the exception and move on.

jimduff-usds: `ls` is only called by our checking code, not by the regular sftp push code.  And sftp to TX was failing, apart the from the `ls` check, true?   That may imply that we need to put the empty catch in other places in the code as well.   Have you been able to test pushing actual files to TX?

I'm wary of empty catches - can we add a logger.warn here?

jimduff-usds (approved):
Approved because I know it is flagged.  However, per my comment, let's add some logging, and I'm wondering about regular sftp uploads to TX.

prime-router/src/main/kotlin/transport/SftpTransport.kt
jimduff-usds: Oh, and _excellent_ debugging work!

MauriceReeves-usds: I can add a warn to the logger. Let me test that out this morning, and good catch. I better check sending a test file as well.

prime-router/metadata/schemas/covid-19.schema
```diff
@@ -53,7 +53,7 @@ elements:
     reference: Follows guidence for OBX-17 as defined in the HL7 Confluence page
     referenceUrl: https://confluence.hl7.org/display/OO/Proposed+HHS+ELR+Submission+Guidance+using+HL7+v2+Messages#ProposedHHSELRSubmissionGuidanceusingHL7v2Messages-DeviceIdentification
     hhsGuidanceField: Device ID
-    hl7Field: OBX-17-1
+    hl7OutputFields: [ OBX-17-1, OBX-17-9 ]
```
MauriceReeves-usds, line 56: For some receivers (NM, for example) having the value for the `device_id` in OBX-17-1 causes issues because it's too long, but they permit putting it into OBX-17-9 which allows longer values. Changed the schema here to output the value in both places.

jimduff-usds (approved):
Somehow I thought I already approved this...

Two thoughts:
- as we change pom.xml, I guess we have to think whether the gradle config needs to change.
- Should we keep the SFTP_LEGACY tests in TestReportStream and the IG state of ignore?  Or remove all of that code completely?

ronaldheft-usds:
> Somehow I thought I already approved this...
> 
> Two thoughts:
> 
> * as we change pom.xml, I guess we have to think whether the gradle config needs to change.
> * Should we keep the SFTP_LEGACY tests in TestReportStream and the IG state of ignore?  Or remove all of that code completely?

We should remove the SFTP Legacy code. Code scanners will pick up the lower level of encryption, and we need to show for the ATO documentation a code commit with the encryption removed.

Issue #1007, opened 2021-05-05
The json response to a POST to api/reports should include the 'topic' field.


jimduff-usds:
Add the topic to the json response to api/reports

This is DONE when:
[ ] topic is added as a top-level value in the json response we send back from a submission to api/reports.  Its ok to have it there when the submission fails.
[ ] Re-run ./prime test  locally, to insure this didn't break anything.  This takes 5 minutes or so, to run.
[ ] Update docs/openapi.yml .  It may be out of date in other ways too. update as needed.
[ ] Since this is a customer facing change, we'll need to provide notification prior to production deployment.

tlosbo:
I'll take this one. I don't think we can assign it to ourselves though?

jimduff-usds:
> I'll take this one. I don't think we can assign it to ourselves though?

Consider it yours!    Best I can do at the moment is assign it to Joe.

jimduff-usds:
@tlosbo  - this is done?

tlosbo:
This is done with PR #1031 merged in.

PR #1031, opened 2021-05-07
Issue 1007: Adding topic to /api/reports Json response

tlosbo:
## Changes
- added "topic" to the POST Json response for /api/reports
- updated the openapi.yml to reflect the added Json element
- added an echo of the topic (right after Id) in end2end test which will fail if it doesn't exist with an "Unable to properly parse response json" message
- fixed a documentation link in the getting started doc that was going to 404 page

## Checklist
- [ ] Verify that the "topic" is added to the POST /api/reports response Json
- [ ] Review the CLI TestReportStream to determine if the test is adequate
- [ ] Check the openapi.yaml for correctness
- [ ] Ensure the link change for the SFTP readme is correct 

### Testing
- [Y] Tested locally?
- [Y] Ran `quickTest all`?
- [-] Ran `./prime test` against local Docker ReportStream container?
I am seeing end2end, merge, garbage tests failing on this branch, but also fails on the current master branch
- [N] Downloaded a file from `http://localhost:7071/api/download`?
Not sure of the credentials to use here
- [Y] Added tests?
Checking for "topic" in the response Json payload, if it doesn't exist it fails

### Security
- [Y] Did you check for sensitive data, and remove any? 
- [Y] Does logging contain sensitive data?  
- [?] Are additional approvals needed for this change?

### Process
- [Y] Includes a summary of what a code reviewer should verify?
- [N] Updated the release notes?
How do we handle adding notes for an upcoming release? Are the release notes manually curated or generated?
- [-] Database changes are submitted as a separate PR?
- [-] DevOps team has been notified if PR requires ops support?

## Security

## Fixes
- #1007

## To Be Done
- #issue

### DIFF
--- a/prime-router/docs/getting_started.md
+++ b/prime-router/docs/getting_started.md
@@ -178,7 +178,7 @@ A useful Azure tool to examine Azurite and Azure storage is (Storage Explorer)[h
 
 ## Function Development with Docker Compose
 ### Local SFTP Server
-You will need an SFTP server to receive data from the router.  For local tests, refer to the [SFTP-SETUP document](SFTP_SETUP.md) to setup a local SFTP server as a receiver of data.
+You will need an SFTP server to receive data from the router.  For local tests, refer to the [SFTP-SETUP document](SFTP-SETUP.md) to setup a local SFTP server as a receiver of data.
 
 ### Running the Router Locally
 The project's [README](../readme.md) file contains some steps to use the PRIME router in a CLI. However, for the POC app and most other users of the PRIME router will the router in the Microsoft Azure cloud. When hosted in Azure, the PRIME router uses Docker containers. The `DockerFile` describes how to build this container. 
--- a/prime-router/src/main/kotlin/cli/TestReportStream.kt
+++ b/prime-router/src/main/kotlin/cli/TestReportStream.kt
@@ -518,12 +518,10 @@ class End2End : CoolTest() {
                 passed = false
             }
             waitABit(25, environment)
-            passed = examineLineageResults(reportId, allGoodReceivers, fakeItemCount)
+            return passed and examineLineageResults(reportId, allGoodReceivers, fakeItemCount)
         } catch (e: NullPointerException) {
-            bad("***end2end Test FAILED***: Unable to properly parse response json")
-            passed = false
+            return bad("***end2end Test FAILED***: Unable to properly parse response json")
         }
-        return passed
     }
 }
 
--- a/prime-router/src/main/kotlin/cli/TestReportStream.kt
+++ b/prime-router/src/main/kotlin/cli/TestReportStream.kt
@@ -518,13 +518,10 @@ class End2End : CoolTest() {
                 passed = false
             }
             waitABit(25, environment)
-            return examineLineageResults(reportId, allGoodReceivers, fakeItemCount)
+            passed = examineLineageResults(reportId, allGoodReceivers, fakeItemCount)
         } catch (e: NullPointerException) {
-            return bad("***end2end Test FAILED***: Unable to properly parse response json")
+            bad("***end2end Test FAILED***: Unable to properly parse response json")
             passed = false
-        }
-        if (!passed) {
-            bad("***end2end Test FAILED***")
         }
         return passed
     }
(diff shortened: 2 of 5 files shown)


prime-router/docs/getting_started.md
```diff
@@ -178,7 +178,7 @@ A useful Azure tool to examine Azurite and Azure storage is (Storage Explorer)[h
 
 ## Function Development with Docker Compose
 ### Local SFTP Server
-You will need an SFTP server to receive data from the router.  For local tests, refer to the [SFTP-SETUP document](SFTP_SETUP.md) to setup a local SFTP server as a receiver of data.
+You will need an SFTP server to receive data from the router.  For local tests, refer to the [SFTP-SETUP document](SFTP-SETUP.md) to setup a local SFTP server as a receiver of data.
```
jimduff-usds, line 181: Thanks!

prime-router/src/main/kotlin/cli/TestReportStream.kt
```diff
@@ -507,6 +507,8 @@ class End2End : CoolTest() {
             val tree = jacksonObjectMapper().readTree(json)
             val reportId = ReportId.fromString(tree["id"].textValue())
             echo("Id of submitted report: $reportId")
+            val topic = tree["topic"].textValue()
+            echo("Topic for client org sender: $topic")
```
jimduff-usds, line 511: Can you change this so that it fails if the topic is not exactly 'covid-19'? - this is a stronger test than just the exception getting thrown.   Look for the pattern in other places in this file with a 'passed' boolean:
   `var passed = true`
Then you set it to false on error, but allow the test to continue, and return that value from the test.   Make sure once its false it never gets reset to 'true'.   

In addition, there is a method `bad` that prints stuff in red, for failure, and a method `good` that prints stuff in green, for a test passing.   So, when a failure does occur, you can easily see what's going on.

Thanks!

jimduff-usds (changes requested):
I checked out your code and built and ran it - looks good!   I'm requesting a minor change in the test.

- Regarding `'./prime test` failures: I was concerned when you said several of the TestReportStream tests were failing, so I wanted to try them out with your branch.   Out of the box, I'm seeing the same 3 tests fail - merge and end2end are fixable failures caused by missing local credentials to the locally running fake sftp site.      In my setup, the vault itself failed to properly start when docker-compose started up, but I haven't investigated why.

The failure of 'garbage' is a known issue; I've been equivocating about how best to fix it -- the issue is the inclusion now of hhsprotect as a new receiver.   We shouldn't be sending test data to it.  Not quite sure how to fix this, as I like having our test data use our main covid-19 topic - it makes it more likely that we'll find errors.   So, I continue to waffle.

- Regarding the Download site:  Yes, you'd need Okta creds to test that; ignore this for now.

- Regarding "additional approvals" - this refers to stuff like the need to for an official approval of public web content, not regular code PRs.

- Regarding release notes:  We don't have an ideal way to do release notes right now.   We are supposed to create a .md file inside docs/releases, with the date of the release.  One problem is you hafta guess at the release date.   Best guess is Weds May 12, if you want to start a release-notes file.

prime-router/docs/getting_started.md
tlosbo: No problem.

prime-router/src/main/kotlin/cli/TestReportStream.kt
```diff
@@ -498,20 +499,31 @@ class End2End : CoolTest() {
...
+            } else {
+                bad("***end2end Test FAILED***: 'topic' is missing from response json")
+                passed = false
+            }
             waitABit(25, environment)
-            return examineLineageResults(reportId, allGoodReceivers, fakeItemCount)
+            passed = examineLineageResults(reportId, allGoodReceivers, fakeItemCount)
         } catch (e: NullPointerException) {
-            return bad("***end2end Test FAILED***: Unable to properly parse response json")
+            bad("***end2end Test FAILED***: Unable to properly parse response json")
```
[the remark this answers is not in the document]
tlosbo: Makes sense. The SonarCloud bot ended up flagging it so I tried to move the return to the end, but this looks more elegant.

PR #1070, opened 2021-05-12
Added and updated pii to Yes/No in the schema docs

pathikkothari:
This PR ... 

## Changes
- Added pIi field to be populated in the md files
- Added conditional to populate markdown files with Yes/No instead of true/false

### Testing
- [x ] Tested locally?
- [ ] Ran `quickTest all`?
- [ ] Added tests?

### Security

### Process

## Security

## Fixes
- #issue

## To Be Done
- #issue

### DIFF
--- a/prime-router/src/main/kotlin/DocumentationFactory.kt
+++ b/prime-router/src/main/kotlin/DocumentationFactory.kt
@@ -22,6 +22,11 @@ object DocumentationFactory {
         sb.appendLine("") // start with a blank line at the top 
         appendLabelAndData(sb, "Name", displayName)
         appendLabelAndData(sb, "Type", element.type?.name)
+		if (element.pii == true) {
+			appendLabelAndData(sb, "PII", "Yes")
+		} else {
+			appendLabelAndData(sb, "PII", "No")
+		}
         appendLabelAndData(sb, "Format", csvField?.format)
         if (element.hl7OutputFields != null)
             appendLabelAndData(sb, "HL7 Fields", element.hl7OutputFields.joinToString(", "))
--- a/prime-router/src/main/kotlin/DocumentationFactory.kt
+++ b/prime-router/src/main/kotlin/DocumentationFactory.kt
@@ -22,11 +22,7 @@ object DocumentationFactory {
         sb.appendLine("") // start with a blank line at the top 
         appendLabelAndData(sb, "Name", displayName)
         appendLabelAndData(sb, "Type", element.type?.name)
-		if (element.pii == true) {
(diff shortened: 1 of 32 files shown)


prime-router/src/main/kotlin/DocumentationFactory.kt
```diff
@@ -22,6 +22,11 @@ object DocumentationFactory {
         sb.appendLine("") // start with a blank line at the top 
         appendLabelAndData(sb, "Name", displayName)
         appendLabelAndData(sb, "Type", element.type?.name)
+		if (element.pii == true) {
```
jimduff-usds, line 25: More idiomatic Kotlin would be something like:
     appendLabelAndData(sb, "PII", if (element.pii) "Yes" else "No")

The "if (...) ... else ... " expression in Kotlin is the equivalent of the ternary operation  "... ? ... : ..."  in java.

jimduff-usds (changes requested):
Suggesting you try a more kotlin-esque way of writing it - Kotlin has all kinds of enjoyable tricks in it!

Yes, I think you should regenerate all the docs via `./prime generate-docs`, and commit them to the your branch.  We may get conflicts when we merge.

prime-router/src/main/kotlin/DocumentationFactory.kt
pathikkothari: Updated to something similar.  I had to add "== true".  I tend to prefer old syntax at times (unless it's too verbose) since it makes reading and rewriting it in a different language and data processing frameworks much easier.
I will have to read up more on it and will see how well Kotlin works with modern data related frameworks.

MauriceReeves-usds: One of the major benefits of Kotlin is that in many ways it can/does operate like a functional language, so you can tersely express your ideas in a type safe and null safe fashion.  

Another major benefit is that because it sits on top of the JVM, it works with any data framework that Java can, but you can often express things in a cleaner way. This is especially true with the fact that by default variables are immutable, which means you need to opt something in to mutation, and can know that when you pass around your variables they won't be changed, and there shouldn't be any side effects from calling a function (or very few. logging is an example where side effects can't be avoided)

One feature I particularly like is the notion of `let` in Kotlin. Whereas in Java or other languages where you need to do a null check before doing an action on it, in Kotlin you can say:

```kotlin
val returnVal: String? = someFunction()
returnVal?.let {
    // consume and do something with the variable, or, if it's null, don't. it will jus skip this check
}
```

MauriceReeves-usds (approved):

jimduff-usds (approved):

PR #1072, opened 2021-05-12
Proposal:  EmailSchedule Service

mry-usds:
This PR creates a proposal for email scheduler notifications.


MauriceReeves-usds:
I have questions:

1. I don't see anything here about opt-outs and how to handle them. There is the notion of excluding someone from receiving an email, but does SendGrid handle opting-out for us? Are we able to hook into that event and then update our database with the emails of people who have opted-out?
2. Is SendGrid handling other email send exceptions like OOF replies or bounces, etc?
3. What about rate limiting? How do we prevent someone from spamming the function in order to DDoS a DOH?
4. Are there any restrictions on the size/number of the parameters that get sent in so we don't create mammoth emails that fill up people's inboxes?
5. How do we prevent the leakage of PII/PHI?
6. How do we prevent sharing information across jurisdictional boundaries?
7. Who handles the creation of the templates? Is that us? The CDC? State PHDs?
8. Are there going to be assets included in the emails like images, styling, etc? Where will those be hosted?
9. Will we be collecting analytics on the emails when they're sent? Do we need to disclose that? Will SendGrid? Is there any issue with that?

I'm sure I'll have more questions. This is actually what one of my prior employers did as its whole business, and we were team of 40+ devs who dealt with really thorny issues. I just want to make sure we constrain this really tightly and limit our exposure to any potential problems.

jlusds:
Opt out
Sendgrid has API to manage this, we store in our DB
@jlusds  to come up with list of categories - marketing

Sending address
@jlusds  to look at benefits of no reply vs monitored email address

How do we track errors
@mry-usds  to look at what is tracked natively in SendGrid
AI do later after we figure this out -- What do we do about it, e.g What do we do if we keep getting bounces

What about rate limiting?
@mry-usds  to look at what we need to build/integrate from sendgrid that does safety check, e.g. no more than X emails per day within a specific email category

How do we prevent the leakage of PII/PHI?
Good thing to flag, we should think about this as we develop content

How do we prevent sharing information across jurisdictional boundaries?
This should be addressed at the jurisdiction filter logic. A good reminder of how we link content - should always be protected behind a login

Templates
Sendgrid has the functionality to host content @drew-usds has been looking into this
Open question for us --- as we personalize emails more (e.g. jurisdiction name, # of reports, etc) we may want to host our own emails

Hosting images
Will try both -- sendgrid, hosting on reportstream site. Since we're often sending to government email addresses we probably want to avoid image use in general (or make it non-critical)

Analytics
This should come out of the box from sendgrid. good point on disclaimers, e.g. analytics, that content will link to PHI. we'll address in content review. @drew-usds to follow up

mry-usds:
Just some technical clarification:

1. The API supports opt out through the "excluded" array - the process of opting out needs to be addressed (i.e. clicking on an opt out list which modifies the list or in the frontend clicking on settings and clicking opt out)

2. Send exceptions, etc. are tracked by SendGrid - eventually, they can feed back to the EmailSchedule api to block addresses, etc. (Phase Next)

3. As an API this is only accessible by someone with DHPrimeAdmin privileges (right now only us) - part of the purpose of SendGrid is to manage the outgoing emails - so (presumably) they will handle throttling, etc. (@mry-usds to followup)

4. The emails will be dispatched to sendgrid by organization, each organization will receive a list of reports (from their organization) sent since the last email went out -- it is up to the content/template engine to build the email (showing just a count of reports, listing each report, showing totals for each report) - there is no limit on parameters sent to sendgrid

5. Review of content for the templates

6. See 4 above - sendgrid email requests are dispatched by organization

7. We do (at this point) - there is no plan to make these accessible to end-users

8. See note above - preference is no images, but if images are required - they can be send embedded/base64 encoded (just like the original download site) - note thought that SendGrid does allow for hosting of images, attachments, etc.

9. SendGrid collects analytics - we will need to develop a standard footer/disclaimer for emails

jlusds:
**Email categories**
Came up with the following. Thoughts @drew-usds 
-Transactional (cannot opt out) - account creation, password reset, account deletion
-Product Notifications - e.g. daily, weekly summaries
-Account Notifications - e.g. A new team member has been added to your organizaiton
-Product news (Marketing) - announcement of new features
-Research - surveys, asking for people to sign up for 1:1 sessions, etc.

**No-reply versus a monitored email** - it sounds like many sites *discourage* a no-reply email. We should create an address for specifically sending these emails though, e.g. notifications@reportstream.gov or reportstream-notififications@cdc.gov

prime-router/docs/proposals/0006-email-notification/0006-email-notification.md
```diff
@@ -0,0 +1,174 @@
...
+        "param1": "value for param1",
+        ...
+    }
+}
+```
+
+See [openapi.emailservice.yml](./openapi.emailservice.yml) for a detailed specification
+
+The EmailService will fire off emails based on:
+- If there are organizations, only for those organizations, otherwise, for every organization
```
RickHawesUSDS, line 42: I think we need more power than this. At least at first we should have the ability to specify FEDERAL, STATE and COUNTY. And orgs with senders or receivers. Ideally, I want to address only download organizations.

prime-router/docs/proposals/0006-email-notification/0006-email-notification.md
```diff
@@ -0,0 +1,174 @@
...
+{
+    "template": "template.name", // the name of the template
+    "schedule": "chron string for scheduling", // optional, no schedule means send immediate
+    "organization" : "optional organization to send to", // ex. DHpima_az_doh, DHmt_doh, etc.
+    "recipient": [ "optional", "optional" ], //optional array of recipients emails
+    "excluding": [ "optional"], // a list of recipients to exclude, otherwise everybody within the organization receives
+    "parameters": {
+        "param1": "value for param1",
+        ...
+    }
```
RickHawesUSDS, line 35: I like the idea of parameters.

prime-router/docs/proposals/0006-email-notification/0006-email-notification.md
```diff
@@ -0,0 +1,174 @@
...
+]
+```
+#### Notes
+In this case, we have 2 email schedules, the first uses the template "daily.tmpl" and is scheduled to fire daily at noon, with paramters for reports, today, and organization.  The second, uses a template "weekly.tmpl" and goes to everyone in the pima_az_doh organization (except for joe.jones) once a week on friday at noon.
+
+### Create an EmailSchedule
+#### Request
+```json
+POST /email-service/schedule HTTP/1.1
+
```
RickHawesUSDS, line 102: I think of the schedule as just another setting. EG. /settings/email-schedule. This way we don't have write another CLI, Facade, etc. Instead just extend the current one.

prime-router/docs/proposals/0006-email-notification/0006-email-notification.md
```diff
@@ -0,0 +1,174 @@
...
+1. Maintain a schedule of emails to be sent
+1. Have parameters that can be filled with the template based on
+    - most recently sent emails
+    - organization and organization participants
+    - rellevant dates
+
+## Proposal
+
+The EmailService will run as a HTTP triggered service presenting a REST-ful API to manage EmailSchedules.  It will have capabilities to LIST, CREATE, DELETE, and UPDATE EmailSchedules.  When an EmailSchedule is created, the EmailService will programatically create and launch a time-trigger based Azure function based on the 'schedule' element within the supplied message for building the desired email and notifying SendGrid.  For 'send-immediate' messages, no function will be created - but the email will be sent.
+
```
RickHawesUSDS, line 21: It seems like there are two types of functions: a set of HTTP API functions and a timer-triggered function.  The API function doesn't launch another function right? The API functions manipulate database records. While the timer-triggered function talks to SendGrid.

RickHawesUSDS (changes requested):
Thank you for the proposal. In general, I like the idea of an email-schedule as a top-level object in the system. It needs to be better integrated with the other objects in the system like settings. 

It would be nice to understand the database tables you have in mind.

prime-router/docs/proposals/0006-email-notification/0006-email-notification.md
```diff
@@ -0,0 +1,174 @@
...
+| Parameter | Description |
+| --------  | ------------|
+|$reports_since_last| retrieves an array of reports sent since the last email |
+|$today | Java Date representing the current date/time |
+|$org | The organization being referenced |
+|$user | The username being referenced |
+
+*Note: more to be added*
+
```
RickHawesUSDS, line 59: It feels like everything in the organization setting should be made available. For example, state code. If a value is missing, it should be added to the organization setting. Fore example, state name.

prime-router/docs/proposals/0006-email-notification/0006-email-notification.md
```diff
@@ -0,0 +1,174 @@
...
+}
+```
+
+See [openapi.emailservice.yml](./openapi.emailservice.yml) for a detailed specification
+
+The EmailService will fire off emails based on:
+- If there are organizations, only for those organizations, otherwise, for every organization
+- If there are recipients, only for those recipients, otherwise, for every recipient, with the exception of anyone listed in "excluding"
+
+The EmailService will record all of its actions to a logging table.
```
jimduff-usds, line 45: Matt,
 
The `ActionHistory` class should already be set up and ready for this use.   We'll just need to add an new action type 'email'.   It logs to the 'action' table.

RickHawesUSDS (approved):
Thanks for all the improvements.

Issue #1066, opened 2021-05-12
Inconsistent warning messages when posting reports


tlosbo:
**Describe the bug**
When posting a report that has unexpected column headers or invalid/unexpected data, the messages provided in the Json response use internal field names that are confusing to the client posting the report.
Example #1
```
"warnings" : [ {
    "scope" : "ITEM",
    "id" : "737324",
    "details" : "Invalid code: 'sdfsdfdsf' is not a display value in altValues set for 'patient_ethnicity'"
  }, {
    "scope" : "ITEM",
    "id" : "737324",
    "details" : "Invalid date: 'baasasdfs' for element 'patient_dob'"
  } ]
```
The posted CSV report has ‘patientEthnicity’ and ‘patientDOB_pii’ respectively for those  two fields.

Example #2
```
"warnings" : [ {
    "scope" : "ITEM",
    "id" : "testId24ODSBLs_0",
    "details" : "Invalid Code: 'patientRacevVn7JRow_0' does not match any codes for 'patient_race'"
  }, {
    "scope" : "ITEM",
    "id" : "testId24ODSBLs_0",
    "details" : "Invalid code: 'patientEthnicityMWvxRhOg_0' is not a display value in altValues set for 'patient_ethnicity'"
  }, {
    "scope" : "ITEM",
    "id" : "testId24ODSBLs_0",
    "details" : "Invalid Code: 'patientSexaQcW2VWn_0' does not match any codes for 'patient_gender'"
  }, {
    "scope" : "ITEM",
    "id" : "testId24ODSBLs_0",
    "details" : "Invalid postal code 'PatZippmW74WJ0_0'"
  }, {
    "scope" : "ITEM",
    "id" : "testId24ODSBLs_0",
    "details" : "Invalid date: 'patientDob_piiQxxfVWxj_0' for element 'patient_dob'"
  }, {
    "scope" : "ITEM",
    "id" : "testId24ODSBLs_0",
    "details" : "Invalid phone number 'patientPhone_piimc8kiy90_0' for 'patient_phone_number'"
  }]
```
These results are from using the attached CSV report. All warnings are using internal field names. The message should include the report field name and internal mapped field name:
```
{
    "scope" : "ITEM",
    "id" : "testId24ODSBLs_0",
    "details" : "Invalid phone number 'patientPhone_piimc8kiy90_0' for 'patientPhone_pii' mapped to  'patient_phone_number'"
}
```

**Impact**
[...]

PR #1080, opened 2021-05-13
Improving error/warning messages in Json response to report submission

tlosbo:
When posting a report that has unexpected column headers or invalid/unexpected data, the messages provided in the Json response use internal field names that are confusing to the client posting the report. On missing fields, the message shows the CSV header but not the external/internal combination. After discussion on issue #1066 this format was decided:
`'<external field(s)>' ('<internal field>')`

## Changes
- added a custom getter to create the desired field mapping
- replaced the internal field name in all the messaging to the field mapping from the custom getter
- added a findElementByCsvName to the Schema class to find the Element and access the custom getter for missing fields

## Checklist
- [ ] review the errors and warnings by submitting bad data and confirm that they contain the desired information

### Testing
- [x] Tested locally?
- [x] Ran `quickTest all`?
*** Tests FAILED:  garbage *** (known issue)
- [ ] Added tests?

### Security

### Process

## Security

## Fixes
- #1066

## To Be Done

### DIFF
--- a/prime-router/src/main/kotlin/Element.kt
+++ b/prime-router/src/main/kotlin/Element.kt
@@ -177,6 +177,10 @@ data class Element(
             type == Type.TABLE_OR_BLANK ||
             type == Type.BLANK
 
+    // Creates a field mapping string showing the external CSV header name(s)
+    // and the corresponding internal field name
+    val fieldMapping get() = "'${this.csvFields?.joinToString { it -> it.name }}' ('${this.name}')"
+
     fun inheritFrom(baseElement: Element): Element {
         return Element(
             name = this.name,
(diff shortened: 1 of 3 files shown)


MauriceReeves-usds (approved):

MauriceReeves-usds (changes requested):
The code changes look fine, but we should add some unit tests for this.

jimduff-usds (commented):
Looks good - I thought your solution was simple and comprehensive.   I agree with Mo that a unit test on fieldMapping would be good.

Super-useful would be a new test in TestReportStream wherein we create "bad" data with a dummy schema that causes each one of the errors you have modified to occur.   But this is different than testing whether the messages are useful to humans or not.  I was so inspired by this idea, I wrote a ticket for it: #1113.

I do suspect there will be obscure cases where there are multiple CsvField values for a single element, that will look weird on errors.  But this feature has general problems anyway - for example, what would it mean for an element E in some _incoming_ data to be assoc with two csvfields "a" and "b" - and what if "a" is 10 and "b" 20?  What do we do with that?

tlosbo:
I'll address the unit test in issue #1113

jimduff-usds (approved):

PR #1081, opened 2021-05-14
Jim/quality

jimduff-usds:
This PR adds a qualityFilter to every receiver

## Changes
- Each receiver can now set a qualityFilter field, just like a jurisdictionFilter.   The set of filters available to be qualityFilters is the same as for JurisdictionFilters.  The syntax is also the same.  They are run at the same time - first the jurisdictionFilter is applied, then if there is still data not filtered, the qualityFilter is applied.
- There is only one difference between qualityFilters and jurisdictionFilters:   There is a default QualityFilter for topic covid-19, applied if none is set in the Settings for a receiver.  (Whereas for jurisdictionFilters, the default is to do no filtering - a problem for a different day)
- The default quality filter is documented in the release notes.
- The `garbage` test now fails because the quality checks remove bad data before the data gets further down the pipe.  I have removed it from the smoke test list for now, and replaced it with a new test, `qualityfilter` that exercises various quality filter options.
- Added three new filter functions:   `hasValidDataFor(col1,col2....),  hasAtLeastOneOf(col1, col2, ...) and allowAll()`.  The first two are used to implement the default quality filter, and any specific filter a state wants, the last one is for where a state wants to override the default filter with no filtering at all.

### Testing
- [X] Tested locally? 
- [X] Ran `quickTest all`?
- [X] Added tests?

### Security

### Process
- [X] Updated the release notes?
- [X] Database changes are submitted as a separate PR?
- [X] DevOps team has been notified if PR requires ops support?

## Security
No security issues known.

## To Be Done
- Move the default covid-19 quality filter into a GLOBAL Setting in the Settings table.  Right now it's hardcoded.


prime-router/src/main/kotlin/JurisdictionalFilters.kt
```diff
@@ -157,7 +161,112 @@ class OrEquals : JurisdictionalFilter {
...
+                    "  All data in this report will fail the quality check")
+                return Selection.withRange(0, 0)
+            }
+        }
+        return selection
+    }
+}
+
+/**
+ * Implements a quality check match.  If a row has valid data for all the columnsm, the row is selected.
```
MauriceReeves-usds, line 208: I think you mean *any* of the columns. Columns is also misspelled.

MauriceReeves-usds: Or, *at least one column*

prime-router/src/main/kotlin/JurisdictionalFilters.kt
```diff
@@ -157,7 +161,112 @@ class OrEquals : JurisdictionalFilter {
...
+        "hasValidDataFor(" +
+            "message_id," +
+            "equipment_model_name," +
+            "specimen_type," +
+            "test_result," +
+            "patient_last_name," +
+            "patient_first_name," +
+            "patient_dob" +
+        ")",
+        // has valid location (for contact tracing)
```
MauriceReeves-usds, line 253: This might read cleaner if you do a multiline string instead of concatenation.

prime-router/src/main/kotlin/JurisdictionalFilters.kt
```diff
@@ -157,7 +161,112 @@ class OrEquals : JurisdictionalFilter {
...
+    override val name = "hasValidDataFor"
+
+    override fun getSelection(args: List<String>, table: Table, receiver: Receiver): Selection {
+        var selection = Selection.withRange(0, table.rowCount())
+
+        val columnNames = table.columnNames()
+        args.forEach {
+            if (columnNames.contains(it)) {
+                selection = selection.andNot(table.stringColumn(it).isEmptyString)
+            } else {
```
MauriceReeves-usds, line 197: For now we're just checking for the existence of data, but at some point we may want to extend this to do things like check for valid CLIA, valid zip, etc.

MauriceReeves-usds (approved):
There's a minor error in the comments, but otherwise looks good.

prime-router/src/main/kotlin/JurisdictionalFilters.kt
jimduff-usds: I'm assuming that invalid data has gotten "zeroed out" - that is, at this point the field is empty.  But I didn't test this exact situation, I admit.

prime-router/src/main/kotlin/JurisdictionalFilters.kt
jimduff-usds: Good point, although I'm not sure our regex would swallow that, and didn't want to spend time testing it.

PR #1114, opened 2021-05-18
Add fluent assertions to Project

MauriceReeves-usds:
This PR adds a reference to the assertk fluent assertions library, and converts some tests to use the assertk library as examples.

Fluent assertions are easier to read and are chainable, so you're able to more clearly express the intent of your checks in your tests. An example from our own code is:

```kotlin
// this checks to see if the schema is not null, and if it has an element named "c"
assertThat(three).isNotNull().hasElement("c")
```

```kotlin
// this checks to see if the schema is not null, it has an element named "a" and the element has two csvFields attached
assertThat(schema).isNotNull().hasElement("a").csvFieldsHasSize(2)
```
In both of the cases above, I've not only made use of the built-in assertions provided by assertk, but I also added extension methods that let us build our own custom assertions.  Extension methods in assertk come in two flavors:
- `given` where the value is checked and either returns unit or throws the expected condition to fail the test
- `transform` where a value is checked in the assertion, and then a value is returned so assertions can be chained, or the test fails and expected is thrown.

This is more easily explained by examples:
```kotlin
// this checks a nullable list to see if it has a size. by default assertk doesn't do tests on nullable type
// this is a given assertion, which means it's expected it will either return unit or will throw
private fun Assert<List<*>?>.hasSize(expected: Int) = given { actual ->
    // check the size for what was expected, and if it matches, then return
    if (actual?.size == expected) return
    // otherwise fail
    expected("size:${show(expected)} but was ${show(actual?.size?.toString() ?: "null")}")
}

// this checks a schema for an element, and if the element doesn't exist, fails
// this is a transform assertion, which means it takes in a schema object, but actually returns
// a nullable Element wrapped in an assertion, which then lets us chain our assertions
private fun Assert<Schema>.hasElement(expected: String): Assert<Element?> = transform(
    appendName("elementName")
) { actual ->
    // if the schema contains an element with name expected, return the actual element
    if (actual.findElement(expected) != null) {
        actual.findElement(expected)
    } else {
        expected("element named ${show(expected)} to exist")
    }
}
```

The biggest shortcoming I see with assertK is it doesn't work on nullable types cleanly, so we will have some custom code to write for some of our logic, but given how easy that is to accomplish, I don't see it as major hurdle.

## Changes
- Adds reference to the assertk library
- Converts some tests to use assertk style fluent assertions
- Adds some extension methods to demonstrate how they're done
- Added in setting in gradle to parallelize our tests

### Testing
- [x] Tested locally?
- [ ] Ran `quickTest all`?
- [ ] Added tests?

### Security

### Process

### DIFF
--- a/prime-router/build.gradle.kts
+++ b/prime-router/build.gradle.kts
@@ -61,11 +61,13 @@ tasks.test {
     useJUnitPlatform()
     dependsOn("compileKotlin")
     // Run the test task if specified configuration files are changed
-    inputs.files(fileTree("./") {
-        include("settings/**/*.yml")
-        include("metadata/**/*")
-    })
-    outputs.upToDateWhen { 
+    inputs.files(
+        fileTree("./") {
+            include("settings/**/*.yml")
+            include("metadata/**/*")
+        }
+    )
+    outputs.upToDateWhen {
         // Call gradle with the -Pforcetest option will force the unit tests to run
         if (project.hasProperty("forcetest")) {
             println("Rerunning unit tests...")
@@ -74,6 +76,10 @@ tasks.test {
             true
         }
     }
+}
+
+tasks.withType<Test>().configureEach {
+    maxParallelForks = (Runtime.getRuntime().availableProcessors() / 2).takeIf { it > 0 } ?: 1
 }
 
 tasks.processResources {
@@ -345,5 +351,6 @@ dependencies {
     testImplementation("com.github.KennethWussmann:mock-fuel:1.3.0")
     testImplementation("io.mockk:mockk:1.11.0")
     testImplementation("org.junit.jupiter:junit-jupiter-api:5.7.1")
+    testImplementation("com.willowtreeapps.assertk:assertk-jvm:0.24")
     testRuntimeOnly("org.junit.jupiter:junit-jupiter-engine:5.7.1")
 }
--- a/prime-router/src/test/kotlin/CsvFileTests.kt
+++ b/prime-router/src/test/kotlin/CsvFileTests.kt
@@ -1,11 +1,15 @@
 package gov.cdc.prime.router
 
+import assertk.assertThat
+import assertk.assertions.exists
+import assertk.assertions.hasSize
+import assertk.assertions.isEqualTo
+import assertk.assertions.isNotNull
 import gov.cdc.prime.router.serializers.CsvSerializer
 import org.junit.jupiter.api.TestInstance
 import java.io.File
 import kotlin.test.Test
 import kotlin.test.assertEquals
-import kotlin.test.assertNotNull
 import kotlin.test.assertTrue
 import kotlin.test.fail
 
@@ -27,10 +31,8 @@ class CsvFileTests {
     init {
         val outputDirectory = File(outputPath)
         outputDirectory.mkdirs()
-
         val expectedDir = File(expectedResultsPath)
-        assertTrue(expectedDir.exists())
-
+        assertThat(expectedDir).exists()
         metadata = Metadata()
         loadTestSchemas(metadata)
         settings = FileSettings()
@@ -51,7 +53,7 @@ class CsvFileTests {
     private fun transformFileAndTest(fileName: String) {
         val file = File(fileName)
         val baseName = file.name
-        assertTrue(file.exists())
+        assertThat(file).exists()
         val schema = metadata.findSchema(defaultSchema) ?: error("$defaultSchema not found.")
 
         // 1) Ingest the file
@@ -60,7 +62,7 @@ class CsvFileTests {
         val inputReport = result.report ?: fail()
         // 2) Create transformed objects, according to the receiver table rules
         val outputReports = Translator(metadata, settings).translateByReceiver(inputReport)
-        assertEquals(2, outputReports.size)
+        assertThat(outputReports).hasSize(2)
         // 3) Write transformed objs to files, and check they are correct
 
         outputReports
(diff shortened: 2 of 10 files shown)


prime-router/build.gradle.kts
```diff
@@ -76,6 +78,10 @@ tasks.test {
     }
 }
 
+tasks.withType<Test>().configureEach {
+    maxParallelForks = (Runtime.getRuntime().availableProcessors() / 2).takeIf { it > 0 } ?: 1
+}
```
MauriceReeves-usds, line 83: This makes our tests run in parallel.

prime-router/src/test/kotlin/CsvFileTests.kt
```diff
@@ -27,10 +38,8 @@ class CsvFileTests {
     init {
         val outputDirectory = File(outputPath)
         outputDirectory.mkdirs()
-
         val expectedDir = File(expectedResultsPath)
-        assertTrue(expectedDir.exists())
-
+        assertThat(expectedDir).exists()
         metadata = Metadata()
```
MauriceReeves-usds, line 43: Asserts that a file exists.

prime-router/src/test/kotlin/CsvFileTests.kt
```diff
@@ -51,16 +60,21 @@ class CsvFileTests {
...
         val schema = metadata.findSchema(defaultSchema) ?: error("$defaultSchema not found.")
 
         // 1) Ingest the file
         val result = csvSerializer.readExternal(schema.name, file.inputStream(), TestSource)
         assertTrue(result.warnings.isEmpty() && result.errors.isEmpty())
+        assertThat(result).all {
+            prop("warnings") { ReadResult::warnings.call(it) }.isEmpty()
+            prop("errors") { ReadResult::errors.call(it) }.isEmpty()
+        }
+        assertThat(result).hasNoWarnings().hasNoErrors()
```
MauriceReeves-usds, line 73: This shows two different ways to accomplish the same thing. One is by using `assertk`'s prop method, which lets you reflect into the object and assert against it. The second one uses two extension methods.

prime-router/src/test/kotlin/CsvFileTests.kt
```diff
@@ -92,15 +105,35 @@ class CsvFileTests {
...
+            }
+        }
+
+        private fun Assert<ReadResult>.hasNoErrors(): Assert<ReadResult> = transform { actual ->
+            if (actual.errors.count() == 0) {
+                actual
+            } else {
+                expected("expected: ReadResult to have no errors, but it had ${show(actual.errors.count())}")
+            }
+        }
```
MauriceReeves-usds, line 137: Here are two extension methods that takes in `ReadResult` and then returns it back allowing the assertions to be chained.

prime-router/src/test/kotlin/ElementTests.kt
```diff
@@ -49,10 +50,12 @@ internal class ElementTests {
             ),
             csvFields = Element.csvFields("b", format = "\$alt")
         )
-        val noResult = one.toNormalized("Non", "\$alt")
-        assertEquals("N", noResult)
-        val yesResult = one.toNormalized("Oui", "\$alt")
-        assertEquals("Y", yesResult)
+        one.toNormalized("Non", "\$alt").run {
+            assertThat(this).isEqualTo("N")
+        }
```
MauriceReeves-usds, line 55: Rather than create a `noResult` and `yesResult` variable, I just used the `.run` scope function which creates the `this` object and then it falls out of scope at the end of the block.

prime-router/src/test/kotlin/FileSettingsTests.kt
```diff
@@ -64,21 +66,18 @@ class FileSettingsTests {
...
     }
 
     @Test
     fun `test loading a sender and receiver`() {
-        val settings = FileSettings()
-        settings.loadOrganizations(ByteArrayInputStream(sendersAndReceiversYaml.toByteArray()))
-
+        val settings = FileSettings().also {
+            it.loadOrganizations(ByteArrayInputStream(sendersAndReceiversYaml.toByteArray()))
+        }
```
MauriceReeves-usds, line 78: Kotlin has this fun `.also` scope function that lets act on the variable you're creating and then returns the object you're acting on.

prime-router/src/test/kotlin/SimpleReportTests.kt
```diff
@@ -177,129 +180,120 @@ class SimpleReportTests {
...
+            for (expectedResultsLine in lines) {
+                // if a header is passed in and we need to skip it, then check our control values
+                if (skipHeader && !skippedHeader) {
+                    // while we're in the header, find our record id index, which will be our map key
+                    val headerValues = headerLine.split(",")
+                    recordIdIndex = headerValues.indexOf(recordId)
+                    // reset our control variable
+                    skippedHeader = true
+                    continue
+                }
```
MauriceReeves-usds, line 213: I moved all of these functions in SimpleReportTests into the companion object because they are static and should live there instead.

prime-router/build.gradle.kts
carlosfelix2: As a note, we have the ability to mark specific tests to not run in parallel.

carlosfelix2 (approved):

carlosfelix2:
Definitely easier to read!

carlosfelix2:
@MauriceReeves-usds You may want to check on the Sonar check failing

Issue #1008, opened 2021-05-05
Add ability to send a tiny test file to the check/sftpcheck Azure Function


jimduff-usds:
CheckFunction is designed to help us diagnose problems with the system.   Right now it only checks one thing:  sftp connections too states.

It is meant to be run manually, with postman or curl, ad-hoc, as we on-board states, and as we detect problems.  So, its a diagnostic tool.

This task is DONE when:

[ ] the sftp check has a new optional way to send a file.  this should be implemented as an http parameter, eg `&sendfile`.   The file should be sent as the content body of the sftp check call, just like we do for reports.   For example the payload could be specified like this to a. curl call --data-binary "@/my/test/file.txt"
[ ] if the file is over a certain size, fail.  Only allow small files (???)
[ ] The current way of running sftp check work exactly as it does not, if the new file option is not included.

jimduff-usds:
@tlosbo has this one.

MauriceReeves-usds:
It might also be good to allow the check to just write an empty `helloworld.txt` file if we specify that as an option.  This should also probably be rate-limited so that someone cannot post many of these and DDoS a state or us.

tlosbo:
Would you setup the rate limiting in Azure as part of some WAF controls? Or were you thinking of implementing rate limiting in the application code?

tlosbo:
@jimduff-usds @MauriceReeves-usds 
I have some thoughts / questions on this ticket. The current end point is a GET end point. I'll need to make it a POST as well to allow for an uploaded file on the SFTP check. Thinking through options here's what I have in mind:
* Allow for the new optional sendfile parameter and honor GET/POST. If it is a GET request then use a zero byte helloworld.txt file as @MauriceReeves-usds mentioned in the comment above.
* Create a new POST end point and move the common code into supporting functions in the class containing two distinct end point functions

Keeping as one might make it a bit easier and using the GET with the generic helloworld.txt for simple quick checks. And the POST could be used to test a specific file for troubleshooting.

jimduff-usds:
I'd keep this really simple - its a small endpoint.  Why not only have the POST, and get rid of the GET?

tlosbo:
I'm running all the tests and submitting a PR right now. I ended up keeping it in the same endpoint and conditionally upload the file based on how it is called.

MauriceReeves-usds:
@jimduff-usds we merged in #1115. We can probably put this into Review/QA if not done. Agreed?

[...]

PR #1115, opened 2021-05-18
Adding an SFTP send file option to the CheckFunction

tlosbo:
Added a `sendfile` flag to the URL and the ability to `POST` a file to test sending a file via SFTP to the receiver.

## Changes
- added `POST` to the end point
- checking for a `sendfile` parameter in the query string and generating a test file based on `GET` or `POST`
- added an `rm` function to the `SftpTransport` class to remove the uploaded file
- changed logging / messaging slightly to account for new logic

## Checklist
- [ ] ensure the old functionality/messaging works as expected (connect, ls and drop)
- [ ] test the `GET` option with the `sendfile` flag
- [ ] test the `POST` option with the `sendfile` flag and sending data in the request body
- [ ] test the `POST` with a file larger than 100K characters to ensure it gets rejected

### Testing
- [x] Tested locally?
- [x] Ran `quickTest all`?
*** Tests FAILED:  garbage *** known issue
- [ ] Added tests?

### Security
- [x] Are there potential vulnerabilities or licensing issues with any new dependencies introduced?
Potential for testers to bog down the end point trying to upload a large file. Also could inundate the end point with lots of small requests.

### Process

## Security

## Fixes
- #1008

## To Be Done
- #issue

### DIFF
--- a/prime-router/src/main/kotlin/azure/CheckFunction.kt
+++ b/prime-router/src/main/kotlin/azure/CheckFunction.kt
@@ -11,7 +11,10 @@ import gov.cdc.prime.router.Receiver
 import gov.cdc.prime.router.SFTPTransportType
 import gov.cdc.prime.router.transport.SftpTransport
 import java.util.UUID
+import net.schmizz.sshj.sftp.RemoteResourceFilter
+import net.schmizz.sshj.sftp.RemoteResourceInfo
 import org.apache.logging.log4j.kotlin.Logging
+
 
 /*
  * Check API
@@ -23,6 +26,15 @@ class CheckFunction : Logging {
         val name: String,
         val contents: String,
     )
+
+    class TestFileFilter(val fileName: String) : RemoteResourceFilter {
+        override fun accept(resource: RemoteResourceInfo?): Boolean {
+            resource?. let {
+                return resource.isRegularFile && resource.name == fileName
+            }
+            return false
+        }
+    }
 
     @FunctionName("check")
     fun run(
@@ -145,9 +157,13 @@ class CheckFunction : Logging {
         responseBody.add("${receiver.fullName}: Able to Connect to sftp site")
         sftpFile?. let {
             logger.info("Attempting to upload ${it.name} to $sftpTransportType")
+            if (SftpTransport.ls(sshClient, path, TestFileFilter(it.name)).isNotEmpty()) {
+                throw Exception("File ${sftpFile.name} already exists on SFTP server. Aborting upload.")
+            }
+            // the client connection is closed in the SftpTransport methods
+            sshClient = SftpTransport.connect(host, port, credential)
             SftpTransport.uploadFile(sshClient, path, it.name, it.contents.toByteArray())
             responseBody.add("${receiver.fullName}: Uploaded file '${sftpFile.name}' to SFTP transport")
-            // the client connection is closed in the SftpTransport methods
             sshClient = SftpTransport.connect(host, port, credential)
         }
         logger.info("Now trying an `ls` on $path")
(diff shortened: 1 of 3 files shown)


prime-router/src/main/kotlin/azure/CheckFunction.kt
```diff
@@ -35,9 +41,30 @@ class CheckFunction : Logging {
...
+                    request?.body ?: ""
+                )
+                else -> SftpFile("hello-${UUID.randomUUID()}.txt", "")
+            }
+            // size check on the sftp file contents, fails if more than 100K chars in length
+            sftpFile?. let {
+                if (sftpFile.contents.length > 100000) {
+                    return HttpUtilities.badRequestResponse(request, "Test upload file exceeds 100K size limit")
+                }
+            }
```
MauriceReeves-usds, line 64: I like it. Thanks for adding this.

prime-router/src/main/kotlin/azure/CheckFunction.kt
```diff
@@ -107,19 +134,46 @@ class CheckFunction : Logging {
...
+        responseBody.add("${receiver.fullName}: SFTP Transport: $sftpTransportType")
         val credential = SftpTransport.lookupCredentials(receiver.fullName)
-        val sshClient = SftpTransport.connect(host, port, credential)
-        responseBody.add("${receiver.fullName}: Able to Connect to sftp site.  Now trying an `ls`...")
+        var sshClient = SftpTransport.connect(host, port, credential)
+        responseBody.add("${receiver.fullName}: Able to Connect to sftp site")
+        sftpFile?. let {
+            logger.info("Attempting to upload ${it.name} to $sftpTransportType")
+            SftpTransport.uploadFile(sshClient, path, it.name, it.contents.toByteArray())
+            responseBody.add("${receiver.fullName}: Uploaded file '${sftpFile.name}' to SFTP transport")
```
MauriceReeves-usds, line 149: What happens if a file with that name already exists? Does `uploadFile` error out, or does it overwrite? Can we just make it error constantly? I don't want to risk ever overwriting valid data.

prime-router/src/main/kotlin/azure/CheckFunction.kt
```diff
@@ -107,19 +134,46 @@ class CheckFunction : Logging {
...
+            } else {
+                "${receiver.fullName}: Found uploaded file '${sftpFile.name}' in `ls` results"
+            }
+            logger.info(msg)
+            responseBody.add(msg)
+            msg = "${receiver.fullName}: Removing '${sftpFile.name}' from SFTP transport"
+            logger.info(msg)
+            responseBody.add(msg)
+            SftpTransport.rm(SftpTransport.connect(host, port, credential), path, sftpFile.name)
+            msg = "${receiver.fullName}: Success: removed '${sftpFile.name}' from SFTP Transport"
```
MauriceReeves-usds, line 174: I like that we're cleaning up our own record, but given the possibility that we could have a name collision and then remove or overwrite an existing file, I'd like to skip the `rm` if there was a collision by throwing an exception or branching out above.

MauriceReeves-usds (changes requested):

prime-router/src/main/kotlin/azure/CheckFunction.kt
tlosbo: Will do. I was relying on the UUID to hopefully create a unique file name but no guarantee and it should check. I'll see how uploadFile handles this situation and either check to see if it exists and regenerate the file name or just throw an exception.

MauriceReeves-usds: I know the odds of UUID collision is practically nil, but it's definitely better to fail and be safe just in case the stars do align unexpectedly.

tlosbo: Verified that uploading an existing file does overwrite the file that is there, so I added some precautions with this commit:
https://github.com/CDCgov/prime-data-hub/pull/1115/commits/a0182a263cf9b6488526a7d3ba6934ecb68b9fe6

Summary of those changes:
- overloaded the `ls()` method in the SftpTransport with a signature that accepts a RemoteResourceFilter
- added a check using the filter above and throw an exception if the file already exists
- changed the upload check to use the filter vs iterating through the bulk ls return (seems more accurate)

prime-router/src/main/kotlin/azure/CheckFunction.kt
tlosbo: Last commit checks to see if the file exists and will throw an exception before the upload.

tlosbo:
@MauriceReeves-usds I've added the additional check to see if the file exists before uploading and aborting by throwing an exception if that is the case. Please review again when you have time.

PR #1191, opened 2021-05-31
changes to support direct submission to ReportStream, etc

jimduff-usds:
## Changes in this PR:

- Wrote a new `hhsprotect/hhsprotect-covid-19` schema, that generally mimics what Waters sends.  Does not send redundant text versions of code columns.  Added a couple columns.  See that schema for details.
- Wrote new `hash(column-name1, column-name2) `mapper, which generates a sha-256 digest of the concatenation table values.
- This is now used generate hashes identical to Waters, so that HHSProtect can detect duplicates.   Only doing this for patient_id right now - the patient_id we send to HHSProtect is the hashed version, not the real one.
- For incoming waters, direct data, and outgoing HHSProtect, use YES,NO,UNK instead of Y,N,UNK
- Delivery codes for ethnicity are done per HHS standard (TODO: fix this in covid-19.schema globally)
- Added detailed logging to the qualityFilters.   It now lists the row number of every row removed, in the logs.  Next step is to get this into ActionHistory, and into the json payload returned on ingest 201.
- HHSProtect now does a jurisdiction filter on sender_id value. This should be a safe way to prevent, for example, SimpleReport data from going to HHSProtect.   Its set fairly lenient to send major Waters senders in Local and Test, but is pretty locked down in Prod.
- Added schemas for new senders.

### Testing
- [X] Tested locally?
- [X] Ran `quickTest all`?
- [X] Added tests?

### Security

### Process
- [Y] Includes a summary of what a code reviewer should verify?

## Security
no issues known.

## To Be Done
- refactor a couple Elements up into covid-19.schema, per my comments.

### DIFF
--- a/prime-router/src/main/kotlin/JurisdictionalFilters.kt
+++ b/prime-router/src/main/kotlin/JurisdictionalFilters.kt
@@ -194,7 +194,7 @@ class HasValidDataFor : JurisdictionalFilter {
         val columnNames = table.columnNames()
         args.forEach { colName ->
             if (columnNames.contains(colName)) {
-                val before = selection.or(Selection.withRange(0,0))  // hack copy constructor!  For logging only.
+                val before = Selection.with(*selection.toArray())  // hack way to copy to a new Selection obj
                 selection = selection.andNot(table.stringColumn(colName).isEmptyString)
                 JurisdictionalFilters.logFiltering(before, selection,"$name($colName)", receiver)
             } else {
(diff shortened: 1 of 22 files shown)


prime-router/src/main/kotlin/JurisdictionalFilters.kt
```diff
@@ -178,6 +178,34 @@ class AllowAll : JurisdictionalFilter {
...
+
+    override fun getSelection(args: List<String>, table: Table, receiver: Receiver): Selection {
+        var selection = Selection.withRange(0, table.rowCount())
+
+        val columnNames = table.columnNames()
+        args.forEach { colName ->
+            if (columnNames.contains(colName)) {
+                val before = Selection.with(*selection.toArray())  // hack way to copy to a new Selection obj
+                selection = selection.andNot(table.stringColumn(colName).isEmptyString)
+                JurisdictionalFilters.logFiltering(before, selection,"$name($colName)", receiver)
```
jimduff-usds, line 199: Check out the cool new detailed logging, via `logFiltering` which prints out the row number of every row that got the kibosh, and `logAllEliminated`, for cases where all the rows are eliminated, its silly to print them all out.

prime-router/src/main/kotlin/Translator.kt
```diff
@@ -86,7 +86,7 @@ class Translator(private val metadata: Metadata, private val settings: SettingsP
             JurisdictionalFilters.defaultQualityFilters[receiver.topic] != null ->
                 JurisdictionalFilters.defaultQualityFilters[receiver.topic]!!
             else -> {
-                logger.info("No default quality filter found for topic ${receiver.topic}. Not doing qual filtering")
+                logger.info("No default qualityFilter found for topic ${receiver.topic}. Not doing qual filtering")
```
jimduff-usds, line 89: A Lame Jim Hack:   I use the term qualityFilter in the log messages, so I can search for that term in the Azure logs easily.

MauriceReeves-usds (approved):

Issue #1148, opened 2021-05-24
Add DB structures to track relevant non-PII data points


MauriceReeves-usds:
We should add a table or set of tables that tracks non-PII/non-PHI data points and lets us view trends in data.

The fields I think we should track are:

- Equipment Model
- Equipment Manufacturer
- Abnormal Flag
- Ordering Provider Name
- Ordering Provider NPI
- Ordering Provider State
- Ordering Provider Zip
- Patient State
- Patient County
- Patient Age
- Patient Gender
- Patient Race
- Patient Tribal Citizenship
- Specimen Collection Date Time
- Testing Lab City
- Testing Lab County
- Testing Lab Name
- Testing Lab Clia
- Testing Lab State
- Client Id (simple_report, cue_evo, strac, etc)
- Receiving orgs (state PHDs, county PHDs, etc)

The goal here is two-fold:
1. Use this to build some internal analytics in Metabase that can help drive our own decision making as well decision-making for our stakeholders at HHS and the CDC
2. Allow the web frontend to build in some analytics for state and county PHDs that let them make more informed decisions.

MauriceReeves-usds:
The acceptance criteria are:

- Create the table(s) to hold the data
- Add the logic to pull this data out from the `Report` object on ingest and then serialize this data to the DB
- Verify that the data requested doesn't violate PII/PHI restrictions
- Test to make sure that this doesn't slow everything down, especially on large insertions, maybe make the extraction and serialization async?
- Add some Metabase queries based on the extracted data

PR #1206, opened 2021-06-02
Add non-PII/non-PHI test data table

MauriceReeves-usds:
This PR adds a `test_data` table that contains all of the important non-PII/non-PHI data that can then be queried by Metabase and allow us to build trends off the data. It could also be used by the download site for viz and metrics. 

## Changes
- Adds a migration for flyway to create the `test_data` table
- Adds db method to get all the internal reports submitted by a sender from the last thirty days
- Adds db method to refresh the data in `test_data` table
- Adds Azure function to save the data for a specific report
- Adds Azure function to save the data for all reports for the last thirty days
- Adds logic that saves the data in report ingest when sent by a sender
- Adds JSON report with the output from saving

### Testing
- [x] Tested locally?
- [x] Ran `quickTest all`?
- [ ] Added tests?

### Security

### Process

## Fixes
- #1148


prime-router/src/main/kotlin/azure/TestDataFunction.kt
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    private fun saveTestDataForAllReports(
+        context: ExecutionContext,
+        request: HttpRequestMessage<String?>
+    ): HttpResponseMessage {
+        val results = mutableMapOf<UUID, String>()
+        val createdDateTime = OffsetDateTime.now().minusDays(30)
+        workflowEngine.db.transact { txn ->
+            workflowEngine.db.fetchAllInternalReports(createdDateTime, txn).forEach { reportFile ->
```
MauriceReeves-usds, line 54: Fetches all the report files for the last 30 days

prime-router/src/main/kotlin/azure/TestDataFunction.kt
```diff
@@ -0,0 +1,181 @@
...
+                val reportId = reportFile.reportId
+                val schema = workflowEngine.metadata.findSchema(reportFile.schemaName)
+                if (schema == null) {
+                    results[reportId] = "Schema name ${reportFile.schemaName} not found in metadata."
+                    return@forEach
+                }
```
MauriceReeves-usds, line 60: Sanity check the schema name

prime-router/src/main/kotlin/azure/TestDataFunction.kt
```diff
@@ -0,0 +1,181 @@
...
+                try {
+                    context.logger.info("Trying to download report $reportId")
+                    val report = getReport(reportFile, schema)
+                    context.logger.info("Extracting the deidentified data for $reportId")
+                    val deidentifiedData = report.getDeidentifiedTestData()
```
MauriceReeves-usds, line 65: Get the deidentified test data

prime-router/src/main/kotlin/azure/TestDataFunction.kt
```diff
@@ -0,0 +1,181 @@
...
+                    context.logger.info("Extracted ${deidentifiedData.count()} rows to insert for $reportId")
+                    context.logger.info("Removing old test data records for report id $reportId")
+                    workflowEngine.db.deleteTestDataForReportId(reportId, txn)
```
MauriceReeves-usds, line 68: Delete the old data so we don't double add data

prime-router/src/main/kotlin/azure/TestDataFunction.kt
```diff
@@ -0,0 +1,181 @@
...
+                    context.logger.info("Inserting deidentified data")
+                    workflowEngine.db.saveTestData(deidentifiedData, txn)
```
MauriceReeves-usds, line 70: Save the test data

prime-router/src/main/kotlin/azure/TestDataFunction.kt
```diff
@@ -0,0 +1,181 @@
...
+            }
+        }
+        return HttpUtilities.okResponse(request, writeResponseBody(results))
+    }
+
+    private fun saveTestDataForSingleReport(
+        reportId: UUID,
+        context: ExecutionContext,
+        request: HttpRequestMessage<String?>
+    ): HttpResponseMessage {
```
MauriceReeves-usds, line 86: Does the same as above for all the reports, but just for a single report ID.

prime-router/src/main/kotlin/azure/TestDataFunction.kt
```diff
@@ -0,0 +1,181 @@
...
+        schema: Schema
+    ): Report {
+        val bytes = workflowEngine.blob.downloadBlob(reportFile.bodyUrl)
+        return workflowEngine.csvSerializer.readInternal(
+            schema.name,
+            ByteArrayInputStream(bytes),
+            emptyList(),
+            blobReportId = reportFile.reportId
+        )
+    }
```
MauriceReeves-usds, line 145: Extracting the report here. I'm not doing any error handling because the error handling is happening higher up where it's called. I'm letting this throw if there's a problem.

prime-router/src/main/kotlin/azure/ReportFunction.kt
```diff
@@ -93,6 +93,13 @@ class ReportFunction {
                     // Regular happy path workflow is here
                     context.logger.info("Successfully reported: ${validatedRequest.report.id}.")
                     routeReport(context, workflowEngine, validatedRequest, actionHistory)
+                    // write the data to the table
+                    context.logger.info("Writing deidentified report data to the DB")
+                    workflowEngine.db.transact { txn ->
+                        val deidentifiedData = validatedRequest.report.getDeidentifiedTestData()
+                        workflowEngine.db.saveTestData(deidentifiedData, txn)
+                        context.logger.info("Wrote ${deidentifiedData.count()} rows to test data table")
+                    }
```
MauriceReeves-usds, line 102: This is the call to save the data on ingest.

prime-router/src/main/kotlin/azure/DatabaseAccess.kt
```diff
@@ -484,6 +504,55 @@ class DatabaseAccess(private val create: DSLContext) : Logging {
...
+                .executeAsync()
+        }
+    }
+
+    fun deleteTestDataForReportId(reportId: UUID, txn: DataAccessTransaction) {
+        DSL.using(txn)
+            .deleteFrom(TEST_DATA)
+            .where(TEST_DATA.REPORT_ID.eq(reportId))
+            .execute()
+    }
```
MauriceReeves-usds, line 554: This deletes from the table for a single report ID

prime-router/src/main/kotlin/azure/DatabaseAccess.kt
```diff
@@ -484,6 +504,55 @@ class DatabaseAccess(private val create: DSLContext) : Logging {
...
+                .set(TEST_DATA.PATIENT_GENDER, it.patientGender)
+                .set(TEST_DATA.PATIENT_POSTAL_CODE, it.patientPostalCode)
+                .set(TEST_DATA.PATIENT_RACE_CODE, it.patientRaceCode)
+                .set(TEST_DATA.PATIENT_RACE, it.patientRace)
+                .set(TEST_DATA.PATIENT_STATE, it.patientState)
+                .set(TEST_DATA.PATIENT_AGE, it.patientAge)
+                .set(TEST_DATA.SPECIMEN_COLLECTION_DATE_TIME, it.specimenCollectionDateTime)
+                .executeAsync()
+        }
+    }
```
MauriceReeves-usds, line 547: Set this up to execute async so it can just run in the background and we don't have to monitor it. I think it will be fine in production, but we will have to see if it continues to hold up and perform well.

prime-router/src/main/kotlin/azure/DatabaseAccess.kt
```diff
@@ -187,6 +189,24 @@ class DatabaseAccess(private val create: DSLContext) : Logging {
...
+            .and(Tables.REPORT_FILE.CREATED_AT.ge(createdDt))
+        return ctx
+            .selectFrom(Tables.REPORT_FILE)
+            .where(cond)
+            .fetchArray()
+            .map {
+                it.into(ReportFile::class.java)
+            }
+    }
+
```
MauriceReeves-usds, line 209: This gets all the internal reports from X date, or the last thirty days if null.

prime-router/src/main/kotlin/Report.kt
```diff
@@ -834,9 +843,26 @@ class Report {
...
+            }
+        }
+
+        private fun Row.getStringOrDefault(columnName: String, default: String? = null): String? {
+            return try {
+                this.getString(columnName)
+            } catch (_: Exception) {
+                default
+            }
+        }
```
MauriceReeves-usds, line 866: Added two extension methods to let me safely get the data if it exists in the report otherwise ignore the problem.

prime-router/src/main/kotlin/Report.kt
```diff
@@ -487,6 +473,74 @@ class Report {
...
+                    }
+                    it.reportId = this.id
+                }
+            }
+        } catch (e: Exception) {
+            logger.warn(e)
+            emptyList()
+        }
+    }
+
```
MauriceReeves-usds, line 543: This gets all the data from the report for our insert. I'm not super crazy about how I'm doing the valueset translation, but I'm not sure there's a cleaner way to accomplish it.

prime-router/build.gradle.kts
```diff
@@ -83,6 +83,10 @@ tasks.withType<Test>().configureEach {
     maxParallelForks = (Runtime.getRuntime().availableProcessors() / 2).takeIf { it > 0 } ?: 1
 }
 
+tasks.withType<JavaExec>().configureEach {
+    jvmArgs(listOf("--add-opens", "java.base/jdk.internal.loader=ALL-UNNAMED"))
+}
+
```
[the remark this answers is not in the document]
carlosfelix2: Will suggest having the JVM args in gradle.properties.

prime-router/build.gradle.kts
```diff
@@ -83,6 +83,14 @@ tasks.withType<Test>().configureEach {
...
+tasks.withType<JavaCompile>().configureEach {
+    options.forkOptions.jvmArgs?.addAll(listOf("--add-opens", "jdk.compiler/com.sun.tools.javac.code=ALL-UNNAMED"))
```
carlosfelix2, line 91: Ditto, in gradle.properties.

prime-router/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.0-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.0.2-bin.zip
```
carlosfelix2, line 3: This update will work, but make sure to run `./gradlew wrapper --gradle-version 7.0.2` in case it updates something else.

prime-router/src/main/kotlin/Report.kt
carlosfelix2: I wonder, since we are marking the fields in the schema as PII, could we get these fields based on that?  Of course, the different types makes it more difficult.

carlosfelix2 (approved):
You have some good comments on this review related to what the code is doing.  I suggest to have those comments as part of the code, even using KDoc.

carlosfelix2:
How do I test this locally?  Maybe add some steps to the description of the PR.

jimduff-usds (changes requested):
This will be a huge improvement, but I think it needs just a bit more work.

I guess I'm confused about several things:
- I'm not getting why the need for the azure function?    It seems to me you just run the last 30 days once, after you deploy, then never again.   Then each time an internal report is created, just incrementally save its data to the table.
- How are you handling the lineage of data?   There will be queries coming in from senders and from recievers - senders will query based on sender names, and receiver based on receiver names --- if you are storing all 'lnternal' files, these data will appear multiple times, which is fine/good, but when you query you need to filter.   Probably need to include the action_name that created the data as well, in the table, so you know what step of data you are looking at.

Most importantly, I think this data is covid-19 specific, and needs to be tracked that way.

I was thinking to get around that by having a generic json structure attached to each item_lineage line in that table, then extract that out into 'topic' specific tables, like yours.  However, your solution is much faster, and we could still still glom on that intermediate staging of data later, if ever.

A few other questions/requests in other comments.

prime-router/src/main/kotlin/azure/DatabaseAccess.kt
MauriceReeves-usds: There is a chance we want to rerun the load for a single report, and instead of updating, it's easier to remove something and reinsert.

prime-router/src/main/kotlin/Report.kt
MauriceReeves-usds: It's not about the types that make it difficult, it's the fact that there's lots of non-PII/non-PHI data we might not want. I am trying to keep this tightly constrained to a smaller subset of data. I recognize there is a tradeoff there that we might need to update this code as we add new column types, but for now, I think that that is less of an issue, and is manageable.

prime-router/gradle/wrapper/gradle-wrapper.properties
MauriceReeves-usds: Done.

prime-router/src/main/kotlin/Report.kt
```diff
@@ -470,6 +473,74 @@ class Report {
         return Report(mapping.toSchema, newTable, fromThisReport("mapping"), itemLineage = itemLineages)
     }
 
+    fun getDeidentifiedTestData(): List<TestData> {
+        return try {
+            table.map { row ->
+                TestData().also {
+                    it.trackingId = row.getStringOrNull("message_id")
```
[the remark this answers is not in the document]
MauriceReeves-usds: Changed this to always use the message ID. Between that and the report ID we should be able to get the sender, etc. Thoughts?

As for the Azure function, the goal with it is to allow us to refresh the data from what exists in our backlog whenever we want. For example, if we do a code push, and add in the abnormal flag, for example, we'd want to be able to just run the endpoint and fill in that data. Or if we decide we do want the sender in the table, etc.

Running it one time on deploy means I'd need to add it to the CLI and then invoke that on Azure, etc and that seems less logical here.

prime-router/src/main/kotlin/Report.kt
```diff
@@ -470,6 +473,74 @@ class Report {
...
+                            try {
+                                LocalDate.parse(dt, Element.datetimeFormatter)
+                            } catch (_: Exception) {
+                                null
+                            }
+                        } else {
+                            null
+                        }
+                    }
+                    it.patientAge = row.getStringOrNull("patient_dob").let { dob ->
```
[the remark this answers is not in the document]
MauriceReeves-usds: If you have some examples for this, I'd like to take a look.

prime-router/src/main/kotlin/Report.kt
```diff
@@ -470,6 +473,74 @@ class Report {
         return Report(mapping.toSchema, newTable, fromThisReport("mapping"), itemLineage = itemLineages)
     }
 
+    fun getDeidentifiedTestData(): List<TestData> {
```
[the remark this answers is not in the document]
MauriceReeves-usds: Changed.

jimduff-usds (approved):
I slacked you a proposed modified version of the .sql file.

jimduff-usds (changes requested):
Per my comments

prime-router/src/main/resources/db/migration/V11__add_test_data_table.sql
```diff
@@ -0,0 +1,61 @@
...
+    specimen_collection_date_time DATE NULL,
+    -- testing lab information
+    testing_lab_city VARCHAR(128) NULL,
+    testing_lab_county VARCHAR(128) NULL,
+    testing_lab_postal_code VARCHAR(32) NULL,
+    testing_lab_state VARCHAR(128) NULL,
+    testing_lab_name VARCHAR(512) NULL,
+    testing_lab_clia VARCHAR(64) NULL,
+    -- the record creation date
+    created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP
```
td-usds, line 56: Why `timestamp` and not `timestamp with timezone`? (https://www.postgresql.org/docs/9.1/datatype-datetime.html)

prime-router/src/main/resources/db/migration/V11__add_test_data_table.sql
```diff
@@ -0,0 +1,61 @@
...
+    patient_age VARCHAR(64) NULL,
+    patient_gender VARCHAR(32) NULL,
+    patient_gender_code VARCHAR(32) NULL,
+    patient_ethnicity VARCHAR(64) NULL,
+    patient_ethnicity_code VARCHAR(32) NULL,
+    patient_race VARCHAR(64) NULL,
+    patient_race_code VARCHAR(32) NULL,
+    patient_tribal_citizenship VARCHAR(128) NULL,
+    patient_tribal_citizenship_code VARCHAR(32) NULL,
+    specimen_collection_date_time DATE NULL,
```
td-usds, line 47: Should this also be a `timestamp with timezone`?

td-usds (changes requested):
General question on the SQL Script: is that autogenerated (and if so, how much control do we have over that autogeneration). If so, my comments on that can be ignored.

prime-router/src/main/resources/db/migration/V11__add_test_data_table.sql
```diff
@@ -0,0 +1,61 @@
+/*
+This SQL creates the tables of the DB. The Flyway tool applies this migration to create the database
```
td-usds, line 2: if this file is hand-authored, then _consider_ adding a `check` constraint to `varchar` fields that are nullable along the lines of
`fieldname varchar(42) null check (fieldname is null or lenght(fieldname) > 0)`
This will prevent empties from making it into the database (and yes, requiring you to substitute empties in the source for `null`s)

This suggestion **only** applies if there is no semantic difference between "not present" (i.e. `null`) and "empty" (i.e. `''`).

td-usds (changes requested):

prime-router/src/main/kotlin/Report.kt
```diff
@@ -484,6 +487,75 @@ class Report {
...
+                    it.patientState = row.getStringOrNull("patient_state")
+                    it.testResultCode = row.getStringOrNull("test_result")
+                    it.testResult = metadata.findValueSet("covid-19/test_result")
+                        ?.toDisplayFromCode(it.testResultCode)
+                    it.equipmentModel = row.getStringOrNull("equipment_model_name")
+                    it.specimenCollectionDateTime = row.getStringOrNull("specimen_collection_date_time").let { dt ->
+                        if (!dt.isNullOrEmpty()) {
+                            try {
+                                LocalDate.parse(dt, Element.datetimeFormatter)
+                            } catch (_: Exception) {
```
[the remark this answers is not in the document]
MauriceReeves-usds: The reason I don't specify here is because I don't care what the exception is here. Because this runs on ingestion, I don't want that to fail for some reason I didn't add a guard for. The bigger sin here is ingestion failing vs losing some amount of data, especially considering I've added some logic to allow us to reload data again, which means we could clean up some error and then fill in this data.

prime-router/src/main/resources/db/migration/V11__add_test_data_table.sql
MauriceReeves-usds: Good call.

prime-router/src/main/resources/db/migration/V11__add_test_data_table.sql
MauriceReeves-usds: I think this is okay. They don't really care about the time here.

jimduff-usds (approved):
This is a huge feature improvement - thank you!

I still think these refactors should be done, but as a later PR/releaseL
1.  this should occur as part of ActionHistory, so that we can elegantly FK to ITEM_LINEAGE.
2. Consider how to integrate as well with @tlosbo 's future warning/error data.
3. I still think there may be tidbits of code in this that are written covid-19 specific that need to be generalized, and
4. we need to refactor Matt's HistoryFunction to call this new table.

Issue #1124, opened 2021-05-20
HL7 ingest support for date DT data type


carlosfelix2:
The HL7 date DT data type is very flexible and we need to store such data in the normalized format.
DT format: YYYY[MM[DD]]

Acceptance criteria:
1. Date DT formats needs to be converted to the normalized internal format upon ingesting HL7 data.
2. A warning is generated to the sender if the date does not contain month and/or day.

PR #1215, opened 2021-06-03
1124 HL7 ingest date type support

carlosfelix2:
This PR adds support for the DATE schema type along with the DT (date) HL7 type. 

To test:
Run the ObservationMessageTests or just `./gradlew test` and verify the tests are successfully.

## Changes
- Support schema DATE type conversion from an HL7 timestamp and provide proper warnings of precision in the data.
- Support HL7 DT date decoding
- Added unit tests and updated data tests

### Testing
- [x] Tested locally?
- [x] Ran `quickTest all`?
- [ ] Added tests?

### Security

### Process

## Security

## Fixes
- #1124


prime-router/src/main/kotlin/serializers/Hl7Serializer.kt
```diff
@@ -954,37 +966,63 @@ class Hl7Serializer(val metadata: Metadata) : Logging {
...
                     // Date range. For getting a date time, use the start of the range
-                    is DR -> value.rangeStartDateTime?.time
-                    else -> null
+                    is DR -> {
+                        dtm = value.rangeStartDateTime?.time?.valueAsDate?.toInstant()
+                        rawValue = value.toString()
+                    }
+                    is DT -> {
+                        dtm = LocalDate.of(value.year, value.month, value.day)
+                            .atStartOfDay(ZoneId.systemDefault()).toInstant()
```
MauriceReeves-usds, line 990: Do we know what the system default is? Is it always UTC on Azure? I'm worried about the potential for time shift here.

prime-router/src/main/kotlin/serializers/Hl7Serializer.kt
```diff
@@ -250,19 +253,28 @@ class Hl7Serializer(val metadata: Metadata) : Logging {
...
-                            errors, warnings
-                        )
-                    }
+                        else {
+                            queryTerserForValue(
+                                terser, getTerserSpec(hl7Field), element.name,
+                                errors, warnings
+                            )
+                        }
                     if (value.isNotBlank()) break
```
MauriceReeves-usds, line 278: Not a fail, but I'm wondering why replace the when with an if/else. Just preference?

MauriceReeves-usds (approved):
No concerns, just two questions

prime-router/src/main/kotlin/serializers/Hl7Serializer.kt
carlosfelix2: Yes, in Azure it is always UTC, however, for this conversion it does not matter as all it is doing is taking a date and converting it to something that has time for the sake of using other conversion methods in the next few lines.  It will not affect the date itself, so you can use any default timezone including EST in our locals.

prime-router/src/main/kotlin/serializers/Hl7Serializer.kt
carlosfelix2: I struggled with this one.  I researched online what to do when you have two when blocks with the same code.  The consensus online was to use if/else instead for readability.  No other reason.

Issue #1217, opened 2021-06-03
COVID results are sometimes encoded as Influenza results from the LIVD table if the test can check both


MauriceReeves-usds:
Email from Ohio noted that they are receiving influenza test results for patients. These appear to be coming from the Sofia 2 test which can do both SARS-CoV-2 and Influenza, and when we attempt to encode the results we sometimes encode them as influenza and return incorrect LOINC codes.

This will need to be fixed right away because this will affect other states. ND, for example, also has labs using the Sofia 2 test.

Acceptance criteria:

- When querying the LIVD table for a COVID-19 message, we should only check for COVID test types and exclude influenza

MauriceReeves-usds:
cc: @RickHawesUSDS

PR #1218, opened 2021-06-03
Fixes sofia look up bug in LIVD table

MauriceReeves-usds:
This PR fixes an issue with some test results getting incorrectly encoded as influenza when they're actually from a test that can detect not only COVID but also Influenza A & B. An example is the Sofia 2 test. 

## Changes
- Adds additional methods to `LookupTable` that lets us filter the table before then doing the lookup
- Baking in the filtering logic in the LIVD lookup to only compare based on the `test_performed_code`

To understand why this is necessary, when we look up the test type, we look up by the equipment model name, which in the case of Sofia 2 is: `Sofia 2 Flu + SARS Antigen FIA*`

If you search for that in the LIVD table, you will see there are three rows, one for the COVID test, and one each for influenza A and influenza B. The easiest distinguishing feature to filter further is the `test_performed_code`, which is provided by senders. By using a combination of the two fields, we are able to eliminate the chance that we end up returning incorrect results for a different test type.

### Testing
- [x] Tested locally?
- [x] Ran `quickTest all`?
- [x] Added tests?

### Process

## Fixes
- #1217


prime-router/src/main/kotlin/LookupTable.kt
```diff
@@ -57,6 +57,21 @@ class LookupTable(
...
+        indexColumn: String,
+        indexValue: String,
+        lookupColumn: String,
+        filters: Map<String, String>,
+        ignoreCase: Boolean = true
+    ): String? {
+        val filteredTable = filter(filters, ignoreCase)
+        return filteredTable.lookupValue(indexColumn, indexValue, lookupColumn, ignoreCase)
+    }
+
```
MauriceReeves-usds, line 74: Adds a method that takes a map of filters and winnows down the LookupTable before then looking up values

prime-router/src/main/kotlin/LookupTable.kt
```diff
@@ -164,6 +194,22 @@ class LookupTable(
...
+                filters.all { (k, v) ->
+                    val filterColumNumber = headerIndex[k.lowercase()] ?: error("$k doesn't exist in lookup table")
+                    row[filterColumNumber].equals(v, ignoreCase)
+                }
+            }
+        val headerAndFilteredRows = listOf(headerRow) + filteredRows
+        return LookupTable(headerAndFilteredRows)
+    }
+
     /**
```
MauriceReeves-usds, line 213: Creates the method that filters the table and then returns a whole new LookupTable object to operate on

prime-router/src/main/kotlin/Mappers.kt
```diff
@@ -225,16 +225,23 @@ class LIVDLookupMapper : Mapper {
...
+        return listOf(EQUIPMENT_MODEL_NAME, DEVICE_ID, EQUIPMENT_MODEL_ID, TEST_KIT_NAME_ID, TEST_PERFORMED_CODE)
     }
 
     override fun apply(element: Element, args: List<String>, values: List<ElementAndValue>): String? {
+        // get the test performed code
+        val testPerformedCode = values.firstOrNull { it.element.name == TEST_PERFORMED_CODE }?.value
+        val filters: MutableMap<String, String> = mutableMapOf()
+        if (!testPerformedCode.isNullOrEmpty()) {
+            filters[LIVD_TEST_PERFORMED_CODE] = testPerformedCode
+        }
```
MauriceReeves-usds, line 237: Get the `test_performed_code` and create a map of filter values for this table to work with

prime-router/src/main/kotlin/LookupTable.kt
RickHawesUSDS: Suggest that you update the comment to include examples of how to use the filter. A Map<String, String> is a pretty generic type and doesn't communicate what the filter does.

RickHawesUSDS: Would it better to think of this a second index or third index?

RickHawesUSDS:
So if I understand your solution to the problem, it relies on the `test_performed_code` being present to resolve between the three rows. This works for SimpleReport, but it doesn't for STRAC. Any suggestions on how we solve this for strac?

prime-router/src/main/kotlin/LookupTable.kt
MauriceReeves-usds: It is potentially an index, but I didn't want to leave it limited to one or two options max. I wanted to add in the flexibility to get in a bunch of filters if we need to in the future.

MauriceReeves-usds:
@RickHawesUSDS I wasn't even aware that STRAC doesn't send the test_performed_code. We don't map that for them? I'll look again. One suggestion that @jimduff-usds and I thought about was actually splitting the LIVD table by topic, so the two Sofia influenza tests would go into a new file, and then only the COVID tests would be in a covid-19 file, and so on. I'm not super enthused by that option, and it introduces some future complexity, but the other option I was thinking about was using a regex and letting us do a negative match on influenza, which I'm not super crazy about either.

prime-router/src/main/kotlin/LookupTable.kt
```diff
@@ -57,6 +57,21 @@ class LookupTable(
         return table[rowNumber][colNumber]
     }
 
+    /**
+     * Filters a table down based on a predicate, and then invokes the standard lookupValue above
```
carlosfelix2, line 61: Love you added kdoc.  Maybe add the arguments and return too?

prime-router/src/main/kotlin/LookupTable.kt
```diff
@@ -164,6 +194,22 @@ class LookupTable(
             .map { row -> row[selectColumnNumber] }
     }
 
+    fun filter(
```
carlosfelix2, line 197: Add kdoc to this one too

carlosfelix2 (approved):

RickHawesUSDS:
So, on the STRAC problem, I could see two options: 

1. Require them to give us a `test_performed_code`, especially if they start supporting Sofia 2. 
2. Build in logic that distinguishes covid tests from other tests

MauriceReeves-usds:
@RickHawesUSDS my preference is to ask them to start providing the code, since that is the most reliable way to distinguish between the different test types. If not, the other options are:
1. We add a column to the LIVD table for topic, and then we filter down by the topic
2. We split the LIVD table by topic, and load the appropriate one for the schema we're working on
3. We add some kind of in-string or regex matching that eliminates anything that matches a certain pattern/only includes what matches the pattern

PR #1223, opened 2021-06-06
Default quality filter now remove processing_mode_code = T,D data

jimduff-usds:
This PR 
- adds a new predicate to the default quality filter:   `doesNotMatch(processing_mode_code,T,D)`.    That is, the data will flow to the state if the `processing_mode_code` column is empty or completely missing (most common situation), or of the column exists and has a 'P'.     Obviously data that is 'T' or 'D' is tautologically not "quality" data.
- This means we now have a new safe way for senders to send test data containing PII to our Production system.   Even if they forget to attach `SkipSend` as an argument, we can still protect the system by marking their `processing_mode_code `to 'T' default in their schema. (Which is exactly what I have done, for all the new senders)
- Added schemas for more direct senders to ReportStream.  Also shuffled around location of several existing schemas.

Here's how the processing_mode_code will behave, with this release.  This little example presumes there is a schema called healthtest that is `basedOn` covid-19.schema, but sets its own default value for `processing_mode_code`:
|Default in covid-19 schm|        Default in healthtest schm|     Actual Data|    RESULT |
|----                        |----              |--- |--- |
|P|P|missing col|Goes to state |
|P|P|empty col|Goes to state |
|P|P|P|Goes to state |
|P|P|T|DOES NOT GO TO STATE |
|P|T|missing col|DOES NOT GO TO STATE |
|P|T|empty col|DOES NOT GO TO STATE |
|P|T|P|Goes to state |
|P|T|T|DOES NOT GO TO STATE |

Notes:
- You can set processing_mode_code defaults in the following places:
-- in the translate section of the receiver settings (at least for redox)
-- as a parameter to the POST
-- in the 'basedOn' schema (eg, the 'covid-19' kitchen sink, as above)
-- in the parent schema
-- in your own schema (as in healthtest example, as above)
-- And of course in the actual data, which can be also be blank or missing the column completely.
Yes, this is all quite confusing.  I certainly couldn't tell you which gets priority.

- The default quality filter is hardcoded right now, and does not vary per environment.   This means that the test/local environments default quality filter will remove data that isn't marked as 'P', production. This is harmless, but confusing.  Remember that the default quality filter can be overridden in the receiver settings, but its a pain to have to do that. 

### Testing
- [X] Tested locally?
- [X] Ran `quickTest all`?
- [ ] Added tests?

### Security

### Process
- [X] Updated the release notes?

## Security
no known new issues

## To Be Done
- need put the default quality filter into the settings, so it can vary per environment.   This would require creating a new 'global' or 'per-env' setting type.
- I ran each situation in the table above by hand, but it would be good to write a comprehensive test of all the 1e6 ways processing_mode_code can be defaulted.


prime-router/metadata/schemas/covid-19.schema
```diff
@@ -776,7 +776,7 @@ elements:
 
   - name: processing_mode_code
     type: CODE
-    default: D
+    default: P
     cardinality: ZERO_OR_ONE
```
jimduff-usds, line 780: So the default in covid-19.schema for processing_mode_code is now 'P' not 'D'.       This enables our unit/quicktest/and TestReportStream tests to pass; otherwise the new default quality filter setting just kills all our data.

I don't know why it was set to 'D', which seems dangerous to me anyway - we might mark real data as Debug data in production.

prime-router/metadata/schemas/direct/imagemover-covid-19.schema
```diff
@@ -0,0 +1,16 @@
...
+trackingElement: specimen_id
+basedOn: covid-19
+extends: direct/direct-covid-19
+
+elements:
+  - name: sender_id
+    default: ImageMover
+
+  - name: processing_mode_code
+    default: T
```
jimduff-usds, line 14: Right here is the reason for all the fuss with processing_mode_code :   I can default this to 'T' here in a sender's schema, then I *know* they can safely send junk data into Production.   Critically, this lets them test with PII data, which they like.  When we are ready, we update their schema to "P" and they are live.

I thought briefly about adding "Sender Defaults", in the settings, as another way to implememt this same idea, instead of the above child-schema-with-default.   But there are so many darn ways to default things already, it seemed wiser to use one our existing clever mechanisms for doing the same thing.

prime-router/settings/prod/0027-la-fix.yml
```diff
@@ -0,0 +1,43 @@
+# Run this with ./prime receiver set --env prod --input ./settings/prod/0027-la-fix.yml
```
jimduff-usds, line 1: This was an emergency fix - already done.   There was a missing comma in the orEquals filter.

prime-router/settings/staging/0021-quality.yml
```diff
@@ -1,16 +1,34 @@
-# Run this:   ./prime receiver set --input ./settings/staging/0021-quality.yml --env staging
+# Run this:   ./prime multiple-settings set --input ./settings/staging/0021-quality.yml --env staging
 ---
-name: "QUALITY_REVERSED"
```
jimduff-usds, line 3: This was to fix a bug with my test case.  Already run in staging.

prime-router/src/main/kotlin/JurisdictionalFilters.kt
```diff
@@ -372,7 +372,7 @@ object JurisdictionalFilters: Logging {
         // has at least one valid CLIA
         "isValidCLIA(testing_lab_clia,reporting_facility_clia)",
         // never send T (Training/Test) or D (Debug) data to the states.
-        // NOT TURNED ON YET    "doesNotMatch(processing_mode_code,T,D)",
+        "doesNotMatch(processing_mode_code,T,D)",
     )
```
jimduff-usds, line 376: Turning on the new default quality filter setting here.

MauriceReeves-usds (approved):

PR #1235, opened 2021-06-07
Changes the default file name format back to the classic style until we can communicate the change to states

MauriceReeves-usds:
This PR changes the default file name format back to the classic style until we can communicate the change to states

## Changes
- updates the name formats yaml to use the old style
- updates the default style in the `Report` class to match the old style

### Testing
- [x] Tested locally?
- [x] Ran `quickTest all`?
- [ ] Added tests?


td-usds (approved):

jimduff-usds (approved):
"Classic style"   :-)

PR #1265, opened 2021-06-08
Enhancements to the operations' Makefile

td-usds:
This PR ... 

## Changes
- rename `TF_ENV_MODULE` to `TF_STAGE` which more accurately reflects what it is (and avoids confusion with `TF_ENV`)
- Adds more information to `make help` output
- Adds more build steps to the `.PHONY` list

### DIFF
--- a/operations/Makefile
+++ b/operations/Makefile
@@ -2,7 +2,7 @@ SHELL := bash
 .SHELLFLAGS := -eu -o pipefail -c
 
 TF_ENV = "dev"
-TF_ENV_MODULE = ""
+TF_STAGE = ""
 
 # HELP
 # This will output the help for each task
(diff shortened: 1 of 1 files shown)


operations/Makefile
```diff
@@ -1,16 +1,24 @@
 SHELL := bash
 .SHELLFLAGS := -eu -o pipefail -c
 
-TF_ENV = "dev"
-TF_ENV_MODULE = ""
+TF_ENV = "dev" #VAR Your target environment (e.g. dev, test, staging, prod)
+TF_STAGE = ""  # not surfaced
```
ronaldheft-usds, line 5: Thank you for documenting! `TF_STAGE` is logical. I always struggle with variable naming, haha. Seeing `TF_STAGE` now, I'm like: "Doh! That makes significantly more sense!"

operations/Makefile
```diff
@@ -21,16 +29,23 @@ build-tf: ## Build the Terraform docker container
...
-	${MAKE} tf-cli TF_ENV_MODULE="04-app"
\ No newline at end of file
+	${MAKE} tf-cli TF_STAGE="04-app"
+
+# Not using build dependencies for this because we want to control order and concurrency
+all: build-tf ## Preconfigures all Terraform environments
+	${MAKE} tf-01-network
+	${MAKE} tf-02-config
+	${MAKE} tf-03-persistent
+	${MAKE} tf-04-app
```
ronaldheft-usds, line 51: I'm not sure the benefit of this. The `tf-XX-*` tasks drop to interactive shells, so while this will drop a user to the correct shell in the correct order, it's probably confusing.

I was thinking on the next round of enhancements to this, we add other tasks that generate plans and apply them. That's where I think a unified task will make more sense.

ronaldheft-usds, line 51: Remove the `all` target.

```suggestion

```

operations/Makefile
```diff
@@ -1,16 +1,24 @@
...
-TF_ENV = "dev"
-TF_ENV_MODULE = ""
+TF_ENV = "dev" #VAR Your target environment (e.g. dev, test, staging, prod)
+TF_STAGE = ""  # not surfaced
 
 # HELP
 # This will output the help for each task
 # thanks to https://marmelab.com/blog/2016/02/29/auto-documented-makefile.html
-.PHONY: help
+.PHONY: help all tf-cli tf-01-network tf-02-config tf-03-persistent tf-04-app
```
ronaldheft-usds, line 10: Remove the `all` target. See comment below.

```suggestion
.PHONY: help tf-cli tf-01-network tf-02-config tf-03-persistent tf-04-app
```

ronaldheft-usds (changes requested):
These suggestions are awesome! Added a suggestion around the `all` task.

operations/Makefile
td-usds: Agreed, I'll get rid of this for now.

td-usds: See 46206085f13adc14a3f7b7e46a43415e3c6869b3

operations/Makefile
td-usds: see 46206085f13adc14a3f7b7e46a43415e3c6869b3

ronaldheft-usds (approved):
Awesome changes! Good to merge into my branch.

PR #1266, opened 2021-06-08
1234 - Change sending_at response on SkipSend=on

pablocaviglia:
This PR makes the **report** endpoint to write back the **sending_at** field with the value **never - skipSend specified** when the SkipSend flag is set.

**curl -X POST -H "client:simple_report" -H "content-type:text/csv" --data-binary "@c:\users\pablo\pdi-covid-19-20210610115239-5cf8f0e5-5109-467b-8557-ba60f906a021.csv" "http://localhost:7071/api/reports?option=SkipSend"**

RESPONSE
```
{
  "id" : "675135a0-a308-4027-a0b7-9790cbcf94ff",
  "timestamp" : "2021-06-10T20:12:20.038694Z",
  "topic" : "covid-19",
  "reportItemCount" : 1,
  "destinations" : [ {
    "organization" : "Public Health Department - Santa Clara, California",
    "organization_id" : "ca-scc-phd",
    "service" : "elr-download",
    "sending_at" : "never - skipSend specified",
    "itemCount" : 1
  }, {
    "organization" : "California Department of Public Health",
    "organization_id" : "ca-dph",
    "service" : "elr",
    "sending_at" : "never - skipSend specified",
    "itemCount" : 1
  } ],
  "destinationCount" : 2,
  "warningCount" : 0,
  "errorCount" : 0,
  "errors" : [ ],
  "warnings" : [ ]
}

```
## Changes
- **prime-router\src\main\kotlin\azure\ActionHistory.kt** to parameterize the functions call-chain with the report options to evaluate
- **prime-router\src\test\kotlin\azure\ActionHistoryTests.kt** to change the function calls and add the new neede parameter, as also a new unitary test for the new case of SkipSend

### Testing
- [x] Tested locally?
- [x] Ran `quickTest all`?
- [ ] Added tests?

### Security

### Process

## Security

## Fixes
- #issue

## To Be Done
- #issue

### DIFF
--- a/prime-router/src/main/kotlin/azure/ActionHistory.kt
+++ b/prime-router/src/main/kotlin/azure/ActionHistory.kt
@@ -414,19 +414,23 @@ class ActionHistory {
         if (reportsOut.size == 0 && parentChildReports.size > 0)
             error("There are item lineages (${parentChildReports.joinToString(",")}) but no child reports")
         // compare the set of reportIds from the item lineage vs the set from report lineage.  Should be identical.
-        val parentReports = parentChildReports.map { it.first}.toSet()
-        val childReports = parentChildReports.map { it.second}.toSet()
+        val parentReports = parentChildReports.map { it.first }.toSet()
+        val childReports = parentChildReports.map { it.second }.toSet()
         var parentReports2 = mutableSetOf<ReportId>()
         parentReports2.addAll(reportsReceived.keys)
         parentReports2.addAll(reportsIn.keys)
         val childReports2 = reportsOut.keys
         if (!parentReports.equals(parentReports2)) {
-            error("parent reports from items (${parentReports.joinToString(",")}) != from reports" +
-                "(${parentReports2.joinToString(",")})")
+            error(
+                "parent reports from items (${parentReports.joinToString(",")}) != from reports" +
+                    "(${parentReports2.joinToString(",")})"
+            )
         }
         if (!childReports.equals(childReports2)) {
-            error("child reports from items (${childReports.joinToString(",")} != from reports" +
-                "(${childReports2.joinToString(",")})")
+            error(
+                "child reports from items (${childReports.joinToString(",")} != from reports" +
+                    "(${childReports2.joinToString(",")})"
+            )
         }
         context?.logger?.info("There are ${reportLineages.size} parent->child report-level relationships")
     }
@@ -472,7 +476,11 @@ class ActionHistory {
      *
      * This works by side-effect on jsonGen.
      */
-    fun prettyPrintDestinationsJson(jsonGen: JsonGenerator, settings: SettingsProvider, reportOptions: ReportFunction.Options) {
+    fun prettyPrintDestinationsJson(
+        jsonGen: JsonGenerator,
+        settings: SettingsProvider,
+        reportOptions: ReportFunction.Options
+    ) {
         var destinationCounter = 0
         jsonGen.writeArrayFieldStart("destinations")
         if (reportsOut.isNotEmpty()) {
@@ -496,7 +504,12 @@ class ActionHistory {
             }
             singles.forEach { (_, destData) ->
                 prettyPrintDestinationJson(
-                    jsonGen, destData.orgReceiver, destData.organization, destData.sendingAt, destData.count, reportOptions
+                    jsonGen,
+                    destData.orgReceiver,
+                    destData.organization,
+                    destData.sendingAt,
+                    destData.count,
+                    reportOptions
                 )
                 destinationCounter++
             }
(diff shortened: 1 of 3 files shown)


prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -519,7 +519,7 @@ class ActionHistory {
         jsonGen.writeStringField("service", orgReceiver.name)
         jsonGen.writeStringField(
             "sending_at",
-            if (sendingAt == null) "immediately" else "$sendingAt"
+            if (sendingAt == null) "never" else "$sendingAt"
```
MauriceReeves-usds, line 522: @jimduff-usds is this correct? I thought we could still send immediately? Does an immediate send now fill in the `sendingAt` property?

MauriceReeves-usds: I also think we should add additional context here instead of just saying "never," we should note that this was submitted with `skipSend` true. I know @jimduff-usds said we should just have "never" in the `sendingAt` field, but I propose it would be better and more clear to say `never - skipSend specified`. I'll leave it up to Jim to provide the final word on that.

MauriceReeves-usds (commented):
The only thing that I would comment on the PR is that it's better to add a descriptive title to the PR, and then link this PR to an issue down below. The current title for the PR doesn't give me any context on what this change is supposed to do.

pablocaviglia-baires:
> The only thing that I would comment on the PR is that it's better to add a descriptive title to the PR, and then link this PR to an issue down below. The current title for the PR doesn't give me any context on what this change is supposed to do.

Is it possible for you to change the PR title? It looks like I don't have permissions for doing it.
I'll start adding some extra info on the PR titles from now on.

prime-router/src/main/kotlin/azure/ActionHistory.kt
pablocaviglia: @jimduff-usds do you agree with the change Maurice is proposing ?

jimduff-usds: Yes, I like Mo's idea, and his  `never - skipSend specified` is good.  Please change the ticket as well.

However, the bigger problem is that the ticket very clearly says, 
```
Do no harm check: PLEASE also carefully check that other cases without SkipSend, have not changed! For example, if the SkipSend is NOT used, the above scenario should still say "sending_at" : "immediately",, or it should specify a timestamp.
```

You'll need to set up a transport that has no batching/merging to test this.

jimduff-usds (changes requested):
This doesn't work - ticket says to only replace 'immediately' if skipSend is true.    In other use cases, when there is no batching/timing set, it should still say 'immediately'.

prime-router/src/main/kotlin/azure/ActionHistory.kt
pablocaviglia: fixed

pablocaviglia:
> This doesn't work - ticket says to only replace 'immediately' if skipSend is true. In other use cases, when there is no batching/timing set, it should still say 'immediately'.

fixed

prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -510,17 +510,22 @@ class ActionHistory {
...
-            if (sendingAt == null) "immediately" else "$sendingAt"
-        )
+
+        if(reportOptions == ReportFunction.Options.SkipSend) {
+            jsonGen.writeStringField("sending_at", "never - skipSend specified")
+        }
+        else {
+            jsonGen.writeStringField("sending_at", if (sendingAt == null) "immediately" else "$sendingAt")
+        }
+
```
jimduff-usds, line 528: @pablocaviglia , try rewriting this using the Kotlin "when" statement - you'll find lots of examples of it in our code, plus stuff on the web.   Its like a "switch" statement but more cool and more powerful.   When you have three or more choices, that's a time to think a "when" statement might be appropriate.   Enjoy!

jimduff-usds (changes requested):

prime-router/src/main/kotlin/azure/ActionHistory.kt
pablocaviglia-baires: fixed @jimduff-usds

jimduff-usds: looks good - one minor thing - I think standard Kotlin is to use .isNull instead of == null
Are you using IntelliJ?  If so, it gives you nice tips/fixes like this.

pablocaviglia-baires: Uhmmm... I dont know that function, or extension function... neither the IDE 

image

jimduff-usds: There should be an sendingAt.isNullOrEmpty and .isNullOrBlank

But never mind, let's just get this merged in.

jimduff-usds (approved):

prime-router/src/main/kotlin/azure/ActionHistory.kt
pablocaviglia-baires: Ohh I see what are you saying, but it only applies to strings, not for OffsetDateTime like sendingAt variable

image

Issue #1213, opened 2021-06-03
Generate unit tests for integration testing of CSV to HL7 conversion


carlosfelix2:
We need to test CSV to HL7 conversion in the same manner that we are testing HL7 to CSV conversion.  Furthermore, we want to make these tests belong to a new test suite for integration testing run separately from the unit tests.  This new integration suite needs to be invoked from Gradle as well.

Acceptance criteria:
1. Generate integration tests for converting CSV to HL7.
2. The integration tests belong to a new test suite separate from the unit tests
3. Move the ObservationMessageTests to the integration suite and perhaps rename the file.
4. Any reuse of code for the new tests from the ObservationMessageTests needs to be common code (no code duplication)
7. Tests need to include batch messages

PR #1267, opened 2021-06-09
1213 Add CSV to HL7 integration tests

carlosfelix2:
This PR adds an integration test for the conversion from CSV to HL7 to verify that the generation of HL7 data is not broken with changes to the baseline.  This works in the same manner as the HL7 to CSV tests (formerly known as the ObservationMessageTest), where you have a folder with CSV input data and a companion HL7 file with the expected HL7 output data.  The tests will generate errors and warnings as follows:
Errors are generated when:
1. The number of reports is different
2. A segment in the expected values does not exist in the actual values
3. A expected component value does not match the actual value

Warnings are generated when:
1. A component actual value exists, but no expected value.

Test steps:
1. Build the baseline with `./gradlew package`
2. Verify the test results for both the test and testIntegration test suite.  See build/reports/test.
3. You can run the integration test only with `./gradlew testIntegration` and force it with `./gradlew testIntegration -Pforcetest`

## Changes
- Create a new test suite called integration tests that runs separate from the unit tests.  We will be adding tests like quicktest to this suite
- Publish the test reports in Github Actions and the PR
- Created CSV to HL7 test 
- Renamed the old ObservationMessageTest to HL7 to CSV conversion tests.
- Increased the wait time for the DEV settings container, otherwise I could not run the router correctly
- Added extra check to DateTimeOffsetMapper as it was throwing an exception when a date field is empty

### Testing
- [x] Tested locally?
- [x] Ran `quickTest all`?
- [ ] Added tests?

### Security

### Process

## Security

## Fixes
- #1213


MauriceReeves-usds:
@carlosfelix2 I linked this PR to #1213 for you.

RickHawesUSDS:
Is it possible to make a gradle option to create take the current output and bless it as the new baseline? Or is that what PforceTest does?

carlosfelix2:
> Is it possible to make a gradle option to create take the current output and bless it as the new baseline? Or is that what PforceTest does?

It is possible, but will need to add functionality to create those files.  I actually did just that, grabbed the current output then created the test files.  I'll add another issue for that.  forcetest is just to ignore the gradle up-to-date checks and run the test regardless of changes.

carlosfelix2:
I have been thinking a bit more about this, and when we move the quicktest to this testIntegration suite I really do not want that to run as part of the package task.  Otherwise the local build will take too long.  It should run as a separate call to testIntegration and added as a new step in the Github Actions build.

jimduff-usds:
> I have been thinking a bit more about this, and when we move the quicktest to this testIntegration suite I really do not want that to run as part of the package task. Otherwise the local build will take too long. It should run as a separate call to testIntegration and added as a new step in the Github Actions build.

The quickTest has become the slowTest.   It needs to be rewritten.    There's no reason to test all those states that use the same schema, and there are other useful tests (eg, from all the new Waters schemas, and hhsprotect) that are not in it.

RickHawesUSDS:
The issue that I ran into with the ObservationMessageTests was when I tried to add a new element, all of the expected test files needed to be updated. Could the comparison be updated to accept new columns? I'm not sure if I actually believe this is a good idea, but its worth discussing.

carlosfelix2:
> The issue that I ran into with the ObservationMessageTests was when I tried to add a new element, all of the expected test files needed to be updated. Could the comparison be updated to accept new columns? I'm not sure if I actually believe this is a good idea, but its worth discussing.

This was resolved by removing the one test for number of columns.  Adding a new field to a schema will not fail the test.

MauriceReeves-usds (approved):

Issue #1282, opened 2021-06-11
Prime CLI via Gradle not using environment variables


carlosfelix2:
When running the CLi via gradle, gradle does not pass the process environment variables.  This can be seen by adding a println(environment).  The issue is that the environment is being replaced when passing the database parameters.

PR #1283, opened 2021-06-11
1282 Fixed CLI environment overwrite issue when using Gradle

carlosfelix2:
This PR correctly sets the environment variables when running the CLI from Gradle.  The previous way completely replaced the environment provided by the process.  This new way just adds more variables.

To test:
Run the following commands and verify that the output is using the HashicorpVaultCredentialService and not the MemoryService.
`export $(cat ./.vault/env/.env.local | xargs)`
`./gradlew primeCLI --args='create-credential --type=UserPass --persist=IGNORE--CSV --user foo --pass pass'`

The output will look like:
10:13:12.888 [main][1] INFO  gov.cdc.prime.router.credentials.**HashicorpVaultCredentialService** - CREDENTIAL UPDATE: ...

## Changes
- Set environment variables correctly in Gradle.

### Testing
- [x] Tested locally?
- [ ] Ran `quickTest all`?
- [ ] Added tests?

### Security

### Process

## Fixes
- #1282

### DIFF
--- a/prime-router/build.gradle.kts
+++ b/prime-router/build.gradle.kts
@@ -119,7 +119,6 @@ tasks.register<JavaExec>("primeCLI") {
     environment["POSTGRES_URL"] = dbUrl
     environment["POSTGRES_USER"] = dbUser
     environment["POSTGRES_PASSWORD"] = dbPassword
-    println(environment)
     doFirst {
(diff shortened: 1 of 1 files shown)


ronaldheft-usds (approved):
Awesome find! Thanks for looking into this @carlosfelix2

PR #1284, opened 2021-06-11
Fixes quick test for Sofia 2 fix

MauriceReeves-usds:
This PR is fixes all the way down

## Changes
- Updates the actual and expected CSV files for AZ and Pima so the test type and test name match what's in the LIVD table and allows quick test to pass

### Testing
- [x] Tested locally?
- [x] Ran `quickTest all`?
- [ ] Added tests?


carlosfelix2 (approved):
Tested and approved

PR #1327, opened 2021-06-15
Canned data for localhost

mry-usds:
This PR creates canned data when running as localhost:8088

## Changes
- Added canned data
- Created npm serve to serve up on localhost

### Testing
- [X] Tested locally?
- [ ] Ran `quickTest all`?
- [ ] Added tests?

### Security

### Process

## Security

## Fixes
- #issue

## To Be Done
- #issue


ahay-agile6 (approved):
Lgtm

frontend/src/js/prime-web-receiver-pre.js
```diff
@@ -167,6 +167,15 @@ function signIn() {
 }
 */
 
+/**
```
td-usds, line 170: Is it possible to wrap this whole functionality (together with `isLocalhost()` checks above) in an adapter pattern which then in turn is responsible for returning the right data (i.e. the adapter also functions as a sort-of facade that returns the orgName and the reports based on which kind of adapter it is)?
That way, there is only a single place where you have to check whether or not you are running on localhost or not, and subsequently instantiate the right adapter instance which can then be used from anywhere.

Bisonburger: All good points & good ideas - but I'd rather not make too many changes to this as its OBE with the move to React

PR #1353, opened 2021-06-21
Adding some error checking and constraints on the metadata table

MauriceReeves-usds:
This PR adds some error checking and constraints on the metadata table. There were instances where the insert into the `covid_result_metadata` table were happening *before* the insert into `report_file` happened, which didn't cause any problems when we didn't have the foreign key in `covid_result_metadata`, but once that was added it caused intermittent errors.

I've moved the insert into the `covid_result_metadata` table after the insert into `report_file` to fix the error.

In addition, I've added some constraints on some of the data in `covid_result_metadata` to make sure we're actually inserting null instead of empty string.

## Changes
- Moves the insert for the deidentified covid result metadata after the report file is written to the DB to fix an error with the foreign key
- Adds some constraints on the covid result metadata table
- Added a `trimToNull` extension method for String that converts empty string down to null for our inserts just to make sure we don't end up with issues where we are dealing with both empty strings and nulls and having to differentiate.

### Testing
- [x] Tested locally?
- [x] Ran `quickTest all`?
- [ ] Added tests?

### Security

### Process

## Security

## Fixes
- #issue

## To Be Done
- #issue


prime-router/src/main/kotlin/azure/ReportFunction.kt
```diff
@@ -119,9 +109,41 @@ class ReportFunction {
...
     }
 
+    private fun writeCovidResultMetadataForReport(
+        report: Report?,
+        context: ExecutionContext,
+        workflowEngine: WorkflowEngine
+    ) {
+        if (report != null && report.schema.topic.lowercase() == "covid-19") {
+            // next check that we're dealing with an external file
+            val clientSource = report.sources.firstOrNull { it is ClientSource }
```
jimduff-usds, line 125: So, we only store metadata about the initial incoming file?   That seems incomplete.   For example, Matt's download system needs data about rows in outdoing Reports.

(At one time, I was considering getting rid of these Sources as they didn't seem very useful.)

jimduff-usds (approved):
It looks like we only store metadata on incoming files?   (see my comment)

prime-router/src/main/resources/db/migration/V12__add_constraints_to_test_result_table.sql
```diff
@@ -0,0 +1,200 @@
...
+UPDATE
+    covid_result_metadata
+SET
+    testing_lab_clia = null
+WHERE
+        TRIM(testing_lab_clia) = '';
+
+ALTER TABLE covid_result_metadata
+    ADD CONSTRAINT  ck_testing_lab_clia
+        CHECK(testing_lab_city IS NULL OR LENGTH(testing_lab_city) > 0);
```
td-usds, line 126: This should target `testing_lab_clia` not `testing_lab_city`.

prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -25,6 +25,7 @@ import gov.cdc.prime.router.azure.db.tables.pojos.Task
 import org.jooq.Configuration
 import org.jooq.DSLContext
 import org.jooq.impl.DSL
+import org.postgresql.util.PSQLException
```
td-usds, line 28: Is this an intentional addition?

td-usds (changes requested):
Needs a bugfix in the sql file

prime-router/src/main/kotlin/Report.kt
```diff
@@ -864,12 +864,9 @@ class Report : Logging {
             }
         }
 
-        private fun Row.getStringOrNull(columnName: String): String? {
-            return try {
-                this.getString(columnName)
-            } catch (_: Exception) {
-                null
-            }
+        private fun String?.trimToNull(): String? {
```
carlosfelix2, line 867: kdoc

prime-router/src/main/kotlin/azure/ReportFunction.kt
```diff
@@ -119,9 +109,41 @@ class ReportFunction {
         actionHistory.trackActionResult(httpResponseMessage)
         workflowEngine.recordAction(actionHistory)
         actionHistory.queueMessages() // Must be done after creating TASK record.
+        // write the data to the table if we're dealing with covid-19. this has to happen
+        // here AFTER we've written the report to the DB
+        writeCovidResultMetadataForReport(report, context, workflowEngine)
         return httpResponseMessage
     }
 
+    private fun writeCovidResultMetadataForReport(
```
carlosfelix2, line 118: kdoc

prime-router/src/main/kotlin/Report.kt
```diff
@@ -879,5 +876,9 @@ class Report : Logging {
                 default
             }
         }
+
+        private fun Row.getStringOrNull(columnName: String): String? {
```
carlosfelix2, line 880: I did not see this function being used.  Remove?

carlosfelix2 (approved):

carlosfelix2:
I do not know how to test this, so I could only review the code

prime-router/src/main/kotlin/azure/ReportFunction.kt
MauriceReeves-usds: Currently yes. Adding downloads would duplicate all the data. My sense was we could get this data from the report lineage for files, something like joining on the message_id and the incoming file(s) that fed an outbound report. If that's too expensive we could add a flag to the table for inbound/outbound to denormalize it a little.

prime-router/src/main/kotlin/Report.kt
MauriceReeves-usds: It's used in the Report class itself to safely query each row for the schema values if they exist.

td-usds (approved):

Issue #1381, opened 2021-06-23
Resubmit June 3 data to Santa Clara that was excluded by quality filter


MauriceReeves-usds:
On June 3, Santa Clara sent in information that did not pass the quality filter. Once #1380 is completed, we need to submit that information so they have it.

Issue #1380, opened 2021-06-23
Update quality filter for Santa Clara and California to exclude CLIA check


MauriceReeves-usds:
The quality filter for Santa Clara and California needs to be updated to exclude the CLIA check.

Acceptance criteria:
- Santa Clara has a custom quality filter that excludes the CLIA format check
- California has a custom quality filter that excludes the CLIA format check

PR #1382, opened 2021-06-23
Adds yamlitos for CA quality filter change

MauriceReeves-usds:
This PR adds yamlitos that update the quality filter for CA and Santa Clara that removes the CLIA quality filter check

## Changes
- Adds a yamlito for production that removes the CLIA format check from the default quality filter
- Adds a yamlito for staging that removes the CLIA format check from the default quality filter

### Testing
- [x] Tested locally?
- [x] Ran `quickTest all`?
- [ ] Added tests?

## Fixes
- #1380 

## To Be Done
- #1381

### DIFF
--- a/prime-router/settings/prod/0033-ca-quality-filter.yml
+++ b/prime-router/settings/prod/0033-ca-quality-filter.yml
@@ -1,0 +1,90 @@
+---
+- name: "ca-scc-phd"
+  description: "Public Health Department - Santa Clara, California"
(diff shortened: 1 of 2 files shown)


jimduff-usds (approved):
Harumph!   I think we should make the Senders send proper CLIAs!

(And FWIW, I've stopped trying to keep staging looking like prod, except in cases where I know I want to test a particular weird setup)

PR #1386, opened 2021-06-23
Remove the security_issue.md file

td-usds:
This PR removes the security_issue.md issue template because we have a SECURITY.md file.
If you have such a file, GitHub creates a 'Report a security vulnerability' issue type that refers to the SECURITY.md file itself. In other words: there is no need for this issue type.

See the screenshot that shows we now have 2 security issue types (and the latter one referring to the security.md file)
Screenshot from 2021-06-23 14-07-48


ronaldheft-usds (approved):
Well that's handy and unexpected! Approved.

PR #1399, opened 2021-06-24
shared default ftp credentials for receiver

pablocaviglia:
This PR contains the implementation for sharing default credentials definition for receiver's FTP transport.
Defines a common set of credentials that are used on organization.yml and organization-local.yml. Then when the FTP transport is going to be used, the code evaluates if the receiver has set the default credentials and uses them, otherwise the normal mechanism for credentials lookup is used.

## Changes
- prime-router\docs\getting_started.md -> add command to create default credentials
- prime-router\docs\SFTP-SETUP.md -> explains how to use the new 'credentials' field
- prime-router\settings\organizations-local.yml -> adds new credentials mechanism for SFTP transport
- prime-router\settings\organizationsl.yml -> adds new credentials mechanism for SFTP transport
- prime-router\src\main\kotlin\transport\SftpTransport.kt -> new logic to use the new field
- prime-router\src\main\kotlin\TransportType.kt -> new field definition on SFTPTransportType

### Testing
- [x] Tested locally?
- [x] Ran `quickTest all`?
- [ ] Added tests?

    - new endpoint A
    - new endpoint B

### Process

## Fixes
- #issue

## To Be Done
- #issue


prime-router/docs/SFTP-SETUP.md
```diff
@@ -36,8 +36,11 @@ docker run -p 22:22 -v sftp:/home/foo/upload -d atmoz/sftp foo:pass:::upload
             host: localhost
             port: 2222
             filePath: ./upload
+            credentials: DEFAULT-FTP
```
MauriceReeves-usds, line 39: Instead of `DEFAULT-FTP` this should be called `DEFAULT-SFTP` so it's clear we're *not* using FTP.

prime-router/docs/getting_started.md
```diff
@@ -236,6 +236,7 @@ End-to-end tests check if the deployed system is configured correctly.  The test
     ./gradlew primeCLI --args='create-credential --type=UserPass --persist=IGNORE--CSV --user foo --pass pass'
     ./gradlew primeCLI --args='create-credential --type=UserPass --persist=IGNORE--HL7 --user foo --pass pass'
     ./gradlew primeCLI --args='create-credential --type=UserPass --persist=IGNORE--HL7-BATCH --user foo --pass pass'
+    ./gradlew primeCLI --args='create-credential --type=UserPass --persist=DEFAULT-FTP --user foo --pass pass'
```
MauriceReeves-usds, line 239: See note above about naming this `DEFAULT-SFTP`. 

Also, I'm inclined to wrap this all into a gradle task that will refresh all the permissions in our local dev environment. Thoughts @jimduff-usds @carlosfelix2?

prime-router/src/main/kotlin/transport/SftpTransport.kt
```diff
@@ -43,7 +43,16 @@ class SftpTransport : ITransport, Logging {
...
+
+            // if the transport definition has defined default
+            // credentials use them, otherwise go with the
+            // standard way by using the receiver full name
+            val credential = if (sftpTransportType.credentials != null) {
+                getCredentials(sftpTransportType.credentials)
+            } else {
+                lookupCredentials(receiver.fullName)
+            }
+
```
MauriceReeves-usds, line 55: I am concerned about the order of operations here because it's possible that someone could accidentally assign default credentials to a receiver and that would take precedent over the already defined credentials.  I think it would be better for it to fallback to the default if the credentials for the receiver name don't exist.

MauriceReeves-usds (changes requested):
Great work @pablocaviglia. 

I am only requesting one change, and then have two additional comments. My change is the name of the default should be `DEFAULT-SFTP` instead of `DEFAULT-FTP` because FTP and SFTP are distinct protocols and FTP is notoriously insecure. Other devs or stakeholders could see the FTP instead of the SFTP and confused by the nomenclature. Therefore it's better for us to be explicitly clear.

The other two things I raise are more questions:
- Is the order of operations "correct" in favoring the default first or should we flip it so the CredentialManager checks for the receiver-specific credentials first and then falls back onto the default if it exists?
- Should we have a script somewhere in the project that sets all of these credentials for new devs coming in or add it as a Gradle task? If it is a gradle task we'd obviously need to ensure Docker is running, so maybe just a dev-environment-setup shell script would be best.

prime-router/docs/getting_started.md
ronaldheft-usds: That would be nice to have for bootstrapping the dev environment, so we don't have to think about it during local setup.

prime-router/src/main/kotlin/transport/SftpTransport.kt
ronaldheft-usds: I disagree. The downside would be a failed login to an SFTP site. I think if we're explicitly saying _use these credentials_ we should follow what we define, as if it's an override.

prime-router/src/main/kotlin/transport/SftpTransport.kt
```diff
@@ -43,7 +43,16 @@ class SftpTransport : ITransport, Logging {
...
-            val credential = lookupCredentials(receiver.fullName)
+
+            // if the transport definition has defined default
+            // credentials use them, otherwise go with the
+            // standard way by using the receiver full name
+            val credential = if (sftpTransportType.credentials != null) {
+                getCredentials(sftpTransportType.credentials)
+            } else {
+                lookupCredentials(receiver.fullName)
+            }
```
ronaldheft-usds, line 54: Instead of having discrete methods, I would opt to reuse the same method. The logic on lines 97-100 are to sanitize the string for unsafe characters in Azure. Therefore, if the method signature is changed to:

```
fun lookupCredentials(credentialName: String): SftpCredential
```

then the `getCredentials` method could be removed and the following lines simplified with an elvis:

```suggestion
            val credential = lookupCredentials(sftpTransportType?.credentials ?: receiver.fullName)
```

ronaldheft-usds (changes requested):
Great approach! Some suggestions for keeping the code clean and avoiding duplications, but otherwise looking good!

prime-router/docs/getting_started.md
jimduff-usds: Definitely, although I think this could be a good first step, and do that as a followup?

prime-router/src/main/kotlin/TransportType.kt
```diff
@@ -23,7 +23,8 @@ data class SFTPTransportType
 @JsonCreator constructor(
     val host: String,
     val port: String,
-    val filePath: String
+    val filePath: String,
+    val credentials: String? = null
```
jimduff-usds, line 27: I think credentialName is a better variable name.   This string is not the actual credentials, its just the name used for the credentials in the vault.

prime-router/src/main/kotlin/transport/SftpTransport.kt
jimduff-usds: I'll be the tiebreaker vote, in this case to adopt Ron's lookupCredentials suggestion, above.  

Note: The default credentials feature should only be a dev/test feature.

jimduff-usds (commented):
I'm ok with this once changes are made per comments from Mo and Ron.

prime-router/docs/getting_started.md
pablocaviglia: Is the intention to create a gradle task that does the same thing than ?
"primeCLI --args='create-credential --type=UserPass --persist=DEFAULT-FTP --user foo --pass pass'

Or a task that iterates over the settings table to update each receiver that has SFTP as the transport?

prime-router/docs/SFTP-SETUP.md
pablocaviglia: fixed

prime-router/src/main/kotlin/transport/SftpTransport.kt
pablocaviglia: fixed

prime-router/src/main/kotlin/transport/SftpTransport.kt
pablocaviglia: fixed

prime-router/src/main/kotlin/TransportType.kt
pablocaviglia: fixed

prime-router/src/main/kotlin/transport/SftpTransport.kt
```diff
@@ -75,8 +80,10 @@ class SftpTransport : ITransport, Logging {
     }
 
     companion object {
-        fun lookupCredentials(receiverFullName: String): SftpCredential {
-            val credentialLabel = receiverFullName
+
+        fun lookupCredentials(credentialName: String): SftpCredential {
+
+            val credentialLabel = credentialName
```
ronaldheft-usds, line 86: Looks good, but a question on the extra spacing. Is this the default now for our linter? /cc @jimduff-usds

ronaldheft-usds (approved):

prime-router/src/main/kotlin/transport/SftpTransport.kt
pablocaviglia: ktlint haven't complied about it 

image

prime-router/docs/SFTP-SETUP.md
```diff
@@ -36,8 +36,11 @@ docker run -p 22:22 -v sftp:/home/foo/upload -d atmoz/sftp foo:pass:::upload
             host: localhost
             port: 2222
             filePath: ./upload
+            credentialName: DEFAULT-SFTP
     ```
-    the above specifies an SFTP transport for the az-phd organization at localhost:22 writing to the ./upload directory under the account
+    the above specifies an SFTP transport for the az-phd organization at localhost:22 writing to the ./upload directory under the account.<br><br> 
+    By setting the 'credentials' field at the 'transport' node, you indicate that the credentials should be retrieved from the CredentialService with that name.<br>
```
jimduff-usds, line 42:
```suggestion
    By setting the 'credentialName' field at the 'transport' node, you indicate that the credentials should be retrieved from the CredentialService with that name.<br>
```

prime-router/src/main/kotlin/transport/SftpTransport.kt
jimduff-usds: @pablocaviglia, let's keep the spacing conformant with the way the rest of our code looks.

jimduff-usds (approved):

MauriceReeves-usds (approved):

prime-router/docs/getting_started.md
jimduff-usds: @pablocaviglia - the first one.   Just wrap
   `primeCLI --args='create-credential --type=UserPass --persist=DEFAULT-FTP --user foo --pass pass`
with a gradle call.  

Important:   set it up so you can pass in the user and password as parameters.

PR #1407, opened 2021-06-24
Proposal: IP Address CIDR Request

ronaldheft-usds:
This is a proposal for the IP CIDR request we need to submit to the CDC, which is a prerequisite to peering with the Palo Alto firewall. Items to discuss are listed within the proposal.

### DIFF
--- a/prime-router/docs/proposals/0008-cdc-ip-space.md
+++ b/prime-router/docs/proposals/0008-cdc-ip-space.md
@@ -4,7 +4,7 @@
 
 We are required to implement a firewall system provided by Palo Alto Networks as part of our ATO. The Palo Alto firewall exists in an Azure VNET that is peered with the CDC intranet.
 
-To utilize the Palo Alto firewall, we are required to peer our VNET with Palo Alto firewall VNET, so our egress traffic can filter through the firewall. In Azure, [peering VNET requires that the IP space does not overlap](https://docs.microsoft.com/en-us/azure/virtual-network/virtual-networks-faq#can-i-peer-two-vnets-with-matching-or-overlapping-address-ranges).
+To utilize the Palo Alto firewall, we are required to peer our VNET with the Palo Alto firewall VNET, so our egress traffic can filter through the firewall. In Azure, [peered VNETs require non-overlapping IP spaces](https://docs.microsoft.com/en-us/azure/virtual-network/virtual-networks-faq#can-i-peer-two-vnets-with-matching-or-overlapping-address-ranges).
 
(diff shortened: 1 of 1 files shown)


jimduff-usds (commented):
I'd just say that its highly likely we'll stay with a serverless architecture, at least for those components which might need to support very heavy usage.   This would imply that 512 is on the high end of our need for IPs.

prime-router/docs/proposals/0008-cdc-ip-space.md
```diff
@@ -0,0 +1,132 @@
...
+| GatewaySubnet | Not used under a CDC VNET |  | |  |  |
+| endpoint | Not used under a CDC VNET |  | |  |  |
+
+#### VNET Summary
+
+| Count | Label |
+|--:|:--|
+| 512 | IPs in CIDR |
+| 192 | Allocated IPs to Subnets |
+| 320 | Available IPs |
```
ronaldheft-usds, line 125: @jimduff-usds That's a fair statement that we're requesting too many IPs. Thoughts on 128 IPs per VNET?

```suggestion
**x.x.x.x/25**

### Why a `/25`?

* A `/25` will grant us 128 IP addresses per VNET
* While we currently use a limited number of IP addresses, we cannot anticipate our future needs
* We may want to utilize private endpoints again in the future
* While our current utilized resources do not consume VNET ip addresses when scaled (ex. function app, database, etc), we may in the future have bare containers that need to be scaled, which will consume an IP per server

### Why request a `/25` in each environment-region pair?

* Environments should be identical
    * When we conduct full-scale load testing in a lower environment, we’ll want the same number of IPs to verify our scaling logic
* Regions should allow for full redundant replication
    * While the only resource currently deployed in our secondary region is a database replica, we’ll want the capability to stand up our entire system in our secondary region
    * This would grant us full region redundancy and/or provide a closer network path for users on the west coast

#### Subnet Usage

| Subnet | Description | CIDR | IPs | Used | Available |
|:--|:--|--:|--:|--:|--:|
| public | Internet-routable resources | x.x.x.x/28 | 16 | 5 + 0 = 5 | 11 |
| container | Docker containers (ex. SFTP test) | x.x.x.x/28 | 16 | 5 + 2 = 7 | 9 |
| private | VNET-only; not Internet-routable | x.x.x.x/28 | 16 | 5 + 0 = 5 | 11 |
| GatewaySubnet | Not used under a CDC VNET |  | |  |  |
| endpoint | Not used under a CDC VNET |  | |  |  |

#### VNET Summary

| Count | Label |
|--:|:--|
| 128 | IPs in CIDR |
| 48 | Allocated IPs to Subnets |
| 80 | Available IPs |
```

ronaldheft-usds:
Discussed during our dev sync today. Could I get an approval from someone before merging? Thanks!

td-usds (approved):

Issue #1057, opened 2021-05-11
Clear Build Warnings


MauriceReeves-usds:
We have a series of build warnings that are persistently hanging around. It would be good to address them.

Below is the output from my build (so yours will be different).
```
w: /Users/maurice/development/USDS/prime-data-hub/prime-router/src/main/kotlin/CsvComparer.kt: (73, 66): Unchecked cast: Any? to List<String>
w: /Users/maurice/development/USDS/prime-data-hub/prime-router/src/main/kotlin/CsvComparer.kt: (74, 70): Unchecked cast: Any? to List<String>
w: /Users/maurice/development/USDS/prime-data-hub/prime-router/src/main/kotlin/FakeReport.kt: (134, 74): Elvis operator (?:) always returns the left operand of non-nullable type String
w: /Users/maurice/development/USDS/prime-data-hub/prime-router/src/main/kotlin/FakeReport.kt: (147, 47): Elvis operator (?:) always returns the left operand of non-nullable type String
w: /Users/maurice/development/USDS/prime-data-hub/prime-router/src/main/kotlin/FakeReport.kt: (301, 25): Unnecessary safe call on a non-null receiver of type String?
w: /Users/maurice/development/USDS/prime-data-hub/prime-router/src/main/kotlin/LookupTable.kt: (104, 59): 'toLowerCase(): String' is deprecated. Use lowercase() instead.
w: /Users/maurice/development/USDS/prime-data-hub/prime-router/src/main/kotlin/azure/Authentication.kt: (69, 59): Unchecked cast: Any? to Collection<String>
w: /Users/maurice/development/USDS/prime-data-hub/prime-router/src/main/kotlin/azure/DownloadFunction.kt: (231, 25): Variable 'jwtString' initializer is redundant
w: /Users/maurice/development/USDS/prime-data-hub/prime-router/src/main/kotlin/azure/DownloadFunction.kt: (243, 56): Unchecked cast: Any? to List<String>
w: /Users/maurice/development/USDS/prime-data-hub/prime-router/src/main/kotlin/azure/HistoryFunctions.kt: (432, 13): Variable 'header' is assigned but never accessed
[...]

PR #1408, opened 2021-06-24
Solved warning messages

jorg3lopez:
Worked to solved warning messages that would present themselves during build.

## Changes
- Addressed build warnings (Ktlint line limit: 120 and null checks)
-

### Testing
- [ ] Tested locally?
- [x] Ran `quickTest all`?
- [ ] Added tests?

    - new endpoint A
    - new endpoint B

### Process

## Fixes
- #1057

## To Be Done
- #issue

### DIFF
--- a/prime-router/src/main/kotlin/azure/DownloadFunction.kt
+++ b/prime-router/src/main/kotlin/azure/DownloadFunction.kt
@@ -227,7 +227,7 @@ class DownloadFunction() : SecretManagement, BaseHistoryFunction() {
         var userName = ""
         var orgName = ""
         val cookies = request.headers["cookie"] ?: ""
-        var jwtString = ""
+        var jwtString: String
         cookies.replace(" ", "").split(";").forEach {
             val cookie = it.split("=")
             jwtString = if (cookie[0] == "jwt") cookie[1] else ""
@@ -239,10 +239,14 @@ class DownloadFunction() : SecretManagement, BaseHistoryFunction() {
                     val jwt = jwtVerifier.decode(jwtString)
                     userName = jwt.getClaims().get("sub").toString()
                     val orgs = jwt.getClaims().get("organization")
-                    var org = if (orgs !== null) (orgs as List<String>)[0] else ""
-                    orgName = if (org.length > 3) org.substring(2) else ""
+                    var org = if (orgs !== null) (orgs as List<*>)[0] else ""
+                    orgName = when (org) {
+                        is String -> if (org.length > 3) org.substring(2) else ""
+                        else -> ""
+                    }
+                    // orgName = if (org.length > 3) org.substring(2) else ""
                 } catch (ex: Throwable) {
-                    System.out.println(ex)
+                    println(ex)
                 }
             }
         }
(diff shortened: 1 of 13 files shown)


prime-router/src/main/kotlin/azure/WorkflowEngine.kt
```diff
@@ -184,7 +184,6 @@ class WorkflowEngine(
         // Send immediately.
         val nextEvent = ReportEvent(Event.EventAction.SEND, reportId, at = null)
         db.transact { txn ->
-            val task = db.fetchAndLockTask(reportId, txn)
```
jimduff-usds, line 187: Huh?   Why is this line being eliminated???  As the "atomic read and set" comment above says, the point is to prevent another simultaneous thread/process from processing the same report at the same time.

jimduff-usds (changes requested):

prime-router/src/main/kotlin/azure/WorkflowEngine.kt
jorg3lopez: Hi Jim, this val is never used. Seeing that this creates a lock, do you suggest leaving it as is or taking away the <val task => part and just leaving : db.fetchAndLockTask(reportId, txn)?

prime-router/src/main/kotlin/azure/DownloadFunction.kt
```diff
@@ -239,10 +239,14 @@ class DownloadFunction() : SecretManagement, BaseHistoryFunction() {
...
-                    orgName = if (org.length > 3) org.substring(2) else ""
+                    var org = if (orgs !== null) (orgs as List<*>)[0] else ""
+                    orgName = when (org) {
+                        is String -> if (org.length > 3) org.substring(2) else ""
+                        else -> ""
+                    }
+                    // orgName = if (org.length > 3) org.substring(2) else ""
                 } catch (ex: Throwable) {
-                    System.out.println(ex)
+                    println(ex)
```
MauriceReeves-usds, line 249: We should probably get away from even using println and move these messages into the logging instead.

prime-router/src/main/kotlin/azure/WorkflowEngine.kt
MauriceReeves-usds: @jimduff-usds @jorg3lopez Kotlin was complaining `task` wasn't being used. maybe we should wrap the variable in synchronized, which then uses the `task` variable as a monitor and then it falls out of scope at the same time as the DB lock?

MauriceReeves-usds (changes requested):

prime-router/src/main/kotlin/azure/DownloadFunction.kt
jimduff-usds: I think DownloadFunction is dead code anyway?

prime-router/src/main/kotlin/azure/WorkflowEngine.kt
jimduff-usds: I think just getting rid of "val task =" is sufficient.  The function is being called for its side effect.

jorg3lopez:
Started a new PR for the warnings.

Issue #270, opened 2021-01-10
Licensing for this project is unclear


Mr0grog:
**What type of help do you need?**

* [x] Question
* [ ] New Repo
* [ ] Delete Repo
* [ ] Other

**Please describe how you'd like us to help.**

While reading through the docs in this repo, I found myself a bit confused as to how it is licensed:
- The [`LICENSE` file in the repo root](https://github.com/CDCgov/prime-data-hub/blob/3ed4ea665b16e3073a9d313ac3e744b8537be257/LICENSE) is the Apache 2 License.
- [`CONTRIBUTING.md`](https://github.com/CDCgov/prime-data-hub/blob/3ed4ea665b16e3073a9d313ac3e744b8537be257/CONTRIBUTING.md#public-domain) says the project is in the public domain.
- [`README.md`](https://github.com/CDCgov/prime-data-hub/blob/3ed4ea665b16e3073a9d313ac3e744b8537be257/README.md#public-domain-standard-notice) lists the project as being in the public domain *and* being licensed via Apache 2 in sections right next to each other, and does not clarify that differing circumstances apply to each, as in other dual-licensed open-source projects (e.g. public domain within the United States and Apache 2 elsewhere).

This might leave potential users and contributors both confused about the particular terms and rights that apply to this project, since these two licensing schemes (CC0/public domain vs. Apache 2) definitely conflict with each other. Which is correct, or is there a more nuanced explanation, like certain files are licensed one way while others are licensed differently, or different jurisdictions of use where one or the other license applies?

leebrian:
FYI, I noticed this issue and am interested in what you develop for your project and needs. I set up the template with both because creative comments is meant to address copyright and apache is meant to address software license. That doesn't mean they conflict with each other, but they aren't incompatible by nature, and wanted to have something operational as most programs don't customize this (although I wish they would).

Apache includes a section for copyright, but since we make government products that are in the public domain we don't typically copyright anything we create. So the intent is to note that the material in the project is available under creative commons in terms of IP usage. And that the material is available under an apache license that includes lots more rights and restrictions than copyright. At one point we didn't include a license since just being under public domain means it can be used in pretty much every way that ASL grants, but it was harder on users who were trying to evaluate license compatibility for their projects. So Apache was added to make projects more compatible so by default material are available under both CC0 and ASL. Of course, programs can adjust this to meet particular needs.

[...]

PR #1439, opened 2021-06-29
Updating licensing

MauriceReeves-usds:
This PR updates the licensing in the project to match SimpleReport and provide clarity about whether or not contributions would violate the Anti-Deficiency Act. There's no code changed here. Just doc updates.

## Changes
- Updates LICENSE
- Updates contributing.md
- Updates README.md

## Fixes
- #270

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,19 +36,16 @@ submitting a pull request you are agreeing to comply with this waiver of
 copyright interest.
 
 ## License Standard Notice
+This project is in the public domain within the United States, and copyright and
+related rights in the work worldwide are waived through the [CC0 1.0 Universal public domain dedication](https://creativecommons.org/publicdomain/zero/1.0/).
+All contributions to this project will be released under the CC0 dedication. By
+submitting a pull request or issue, you are agreeing to comply with this waiver
+of copyright interest and acknowledge that you have no expectation of payment,
+unless pursuant to an existing contract or agreement.
+
 The repository utilizes code licensed under the terms of the Apache Software
-License and therefore is licensed under ASL v2 or later.
-
-This source code in this repository is free: you can redistribute it and/or modify it under
-the terms of the Apache Software License version 2, or (at your option) any
-later version.
-
-This source code in this repository is distributed in the hope that it will be useful, but WITHOUT ANY
-WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
-PARTICULAR PURPOSE. See the Apache Software License for more details.
-
-You should have received a copy of the Apache Software License along with this
-program. If not, see http://www.apache.org/licenses/LICENSE-2.0.html
+License, which if unmodified, remains licensed under ASL v2 or later, but all other code
+contributed to this project falls under the CC0 1.0 Universal public domain dedication.
 
 The source code forked from other open source projects will inherit its license.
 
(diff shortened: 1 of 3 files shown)


td-usds:
I think this looks good. The [LICENSE](https://github.com/CDCgov/prime-reportstream/pull/1439/files#diff-c693279643b8cd5d248172d9c22cb7cf4ed163a3c98c8a3f69c2717edd3eacb7) file goes from Apache -> CC0, which I _think_ is fine under "_...it does not require a derivative work of the software, or modifications to the original, to be distributed using the same license._" (quoted from [Wikipedia](https://en.wikipedia.org/w/index.php?title=Apache_License&oldid=1027443356#Licensing_conditions))

td-usds (approved):
I think this looks good.

jimduff-usds (approved):

carlosfelix2 (approved):

README.md
```diff
@@ -36,19 +36,16 @@ submitting a pull request you are agreeing to comply with this waiver of
...
 
-This source code in this repository is distributed in the hope that it will be useful, but WITHOUT ANY
-WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
-PARTICULAR PURPOSE. See the Apache Software License for more details.
-
-You should have received a copy of the Apache Software License along with this
-program. If not, see http://www.apache.org/licenses/LICENSE-2.0.html
+The repository utilizes code licensed under the terms of the Apache Software
+License, which if unmodified, remains licensed under ASL v2 or later, but all other code
+contributed to this project falls under the CC0 1.0 Universal public domain dedication.
```
Mr0grog, line 48: Seems like something that should go with this is a) ensuring that we reliably call out non-public-domain files in the repo with comments (e.g. [`prime-router/gradlew.bat`](https://github.com/CDCgov/prime-reportstream/blob/518bf88991f42ebf0d3aee40b00f2567e6d59797/prime-router/gradlew.bat#L1-L15)) and b) explain here that such files can be identified with that header. Otherwise I’m not sure what a consumer should do other than treat the whole codebase as *if* it were ASL2, since it’s the more restrictive license.

MauriceReeves-usds: I thought the CC0 is more restrictive because it prohibits people from taking the work in the repo and creating their own patentable work off of it, but I could be mistaken. I will check with the lawyers and see what they think.

Mr0grog: Ah, yeah, I guess this is more just *differently* restrictive; I was focusing on copyright. So someone would still have to treat everything as the most-restrictive-common-denominator.

> CC0 is more restrictive because it prohibits people from taking the work in the repo and creating their own patentable work off of it

Worth checking with the lawyers, but I don’t believe it does that. A contributor can patent their contributions under CC0, potentially preventing a consumer from using it (so I guess it is potentially more restrictive, in that sense) but it doesn’t address what consumers can do with the work, like incorporating it into a patented invention (but at that point, I don’t know what patent law says about inventions that involve public domain stuff). ASL2 explicitly says consumers are granted a license to use it even if some contributions are patented, but also makes no claims about downstream patenting by consumers (but I imagine that wouldn’t be possible since copyright is retained in ASL2, which I think would complicate patent applications by non-copyright-holders).

Mr0grog: Anyway, I think my main concern was that this paragraph throws things back into an unclear status in terms of “what can I do with the contents of this repository?” and that still stands.

RickHawesUSDS (approved):

PR #1443, opened 2021-06-30
SendGrid - TO PIMA CO ONLY

Bisonburger:
This PR sets up a SendGrid fn to send a message to users.  Initially - we'll use it to send a notification message to PIMA CO only.

## Changes
- Adds a triggered function to periodlically check for messages and send them
- Adds an HTTP triggered function to create an email schedule

### Testing
- [X] Tested locally?
- [ ] Ran `quickTest all`?
- [ ] Added tests?

    - new endpoint A
    - new endpoint B

YES - adds:
- implementation("com.cronutils:cron-utils:9.1.5")
- implementation("khttp:khttp:0.1.0")

### Process

## Fixes
- #issue

## To Be Done
- #issue


prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
```diff
@@ -0,0 +1,328 @@
...
+        @HttpTrigger(
+            name = "createEmailSchedule",
+            methods = [HttpMethod.POST],
+            authLevel = AuthorizationLevel.ANONYMOUS,
+            route = "email-schedule"
+        ) request: HttpRequestMessage<String?>,
+        context: ExecutionContext,
+    ): HttpResponseMessage {
+        var user:String? = validateUser( request, context );
+        var ret = request.createResponseBuilder(HttpStatus.CREATED);
```
td-usds, line 178: This is a bit nit-picky but I'll raise it nonetheless: this code assumes success by default and then changes that status if things went badly.
I would encourage the opposite:
```kotlin
        var user:String? = validateUser( request, context );
        var ret = request.createResponseBuilder(HttpStatus.UNAUTHORIZED);

        if( user !== null ){
            WorkflowEngine.databaseAccess.insertEmailSchedule(request.body, user );
            ret.status(HttpStatus.CREATED )
        }

        return ret.build();
```

This eliminates the else branch and gives a better default return code.

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
```diff
@@ -0,0 +1,328 @@
...
+
+        if( user !== null ){
+            WorkflowEngine.databaseAccess.insertEmailSchedule(request.body, user );
```
td-usds, line 181: `insertEmailSchedule` returns `Int?` but the return value is discarded.
This should check the return value of the method and only set the status to `CREATED` if a non-null, strictly positive return value comes from `insertEmailSchedule`

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
```diff
@@ -0,0 +1,328 @@
...
+        }
+        else{
+            ret.status(HttpStatus.UNAUTHORIZED )
+        }
+        return ret.build();
+    }
+
+
+    fun validateUser( request: HttpRequestMessage<String?>, context: ExecutionContext ): String? {
```
td-usds, line 190: Something tells me that we should have a common method for this. This appears like something we either already to in multiple places or will be doing in multiple places. Should this belong in the `OktaCommands.kt` file (somewhere)?

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
```diff
@@ -0,0 +1,328 @@
...
+        var jwtToken = request.headers["authorization"] ?: ""
+
+        jwtToken = if (jwtToken.length > 7) jwtToken.substring(7) else ""
+
+        if (jwtToken.isNotBlank()) {
+            try {
+                // get the access token verifier
+                val jwtVerifier = JwtVerifiers.accessTokenVerifierBuilder()
+                    .setIssuer("https://hhs-prime.okta.com/oauth2/default")
```
td-usds, line 199: use constants from `OktaCommands.kt`

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
```diff
@@ -0,0 +1,328 @@
...
+                    .build()
+                // get it to decode the token from the header
+                val jwt = jwtVerifier.decode(jwtToken)
+                    ?: throw Throwable("Error in validation of jwt token")
+                // get the user name and org
+                return jwt.claims["sub"].toString();
+            }
+            catch (ex: Throwable) {
+                context.logger.log(Level.WARNING, "Error in verification of token", ex)
+                return null
```
td-usds, line 209: Generally speaking, I prefer a "one way in, one way out" approach to functions. Here we have multiple exit points. I suggest something like the following specifically for the return-ing aspect of the code (which also sets your 'default' to be a good default:
```kotlin
        var jwtToken = request.headers["authorization"] ?: ""
        var ret = null

        jwtToken = if (jwtToken.length > 7) jwtToken.substring(7) else ""

        if (jwtToken.isNotBlank()) {
            try {
                // get the access token verifier
                val jwtVerifier = JwtVerifiers.accessTokenVerifierBuilder()
                    .setIssuer("https://hhs-prime.okta.com/oauth2/default")
                    .build()
                // get it to decode the token from the header
                val jwt = jwtVerifier.decode(jwtToken)
                    ?: throw Throwable("Error in validation of jwt token")
                // get the user name and org
                ret = jwt.claims["sub"].toString();
            }
            catch (ex: Throwable) {
                context.logger.log(Level.WARNING, "Error in verification of token", ex)
            }
        }
        return ret;
```

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
```diff
@@ -0,0 +1,328 @@
...
+            }
+        }
+        return null;
+    }
+    
+
+    /**
+     * TODO: Fixme!
```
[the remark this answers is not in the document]
td-usds: Also, this seems like a relatively useless helper method as it's used only once. Delete? If we use it multiple times in the future, _then_ we can extract a helper method out of it

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
```diff
@@ -0,0 +1,328 @@
...
+     */
+    private fun getSchedules(): List<String>{
+        return WorkflowEngine.databaseAccess.fetchEmailSchedules();
+    }
+
+    /**
+     * Converts an organization name to an OKTA group name
+     */
+    private fun convertOrgToGroup( org: String ): String {
+        return "DH"+org.replace( "-","_" )
```
td-usds, line 227: We do this in multiple places (at least in `Authentication.kt:79` as well and the inverse (`.replace("_", "-")` in a good number of other places too. This would be a good candidate to centralize that. Maybe something akin to `normalizeGroupName` or `denormalizeGroupName` of something.

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
```diff
@@ -0,0 +1,328 @@
...
+    }
+
+    /**
+     * Retrieve a list of emails within an organization
+     * 
+     * @params org The organization to fetch emails for
+     * @returns List of emails to send to
+     */
+    private fun getEmails( org: String, context: ExecutionContext ): List<String> {
```
td-usds, line 236: Total nitpick: you are passing in the context but only using the logger from the context. Consider passing in the logger instead. Think of this as deliberately putting on a filter and explicitly stating some design intentions of the function: this function is not supposed to have the entire context, it only needs the logger, so only pass that in.

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
```diff
@@ -0,0 +1,328 @@
...
+
+        if( ssws == null ){
+            context.logger.warning("Can't find SSWS_OKTA secret")
+            return listOf();
+        }
+        
+        var grp = convertOrgToGroup( org );
+
+        // get the OKTA Group Id
+        var response1 = httpGet( url="https://hhs-prime-admin.okta.com/api/v1/groups?q=${grp}",
```
td-usds, line 248: Magic constants; reuse from `OktaCommands.kt` where possible

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
```diff
@@ -0,0 +1,328 @@
...
+     * @params org The organization to fetch emails for
+     * @returns List of emails to send to
+     */
+    private fun getEmails( org: String, context: ExecutionContext ): List<String> {
+
+        var ssws: String? = SecretHelper.getSecretService().fetchSecret("SSWS_OKTA")
+
+        if( ssws == null ){
+            context.logger.warning("Can't find SSWS_OKTA secret")
+            return listOf();
```
td-usds, line 242: Consider "one way in, one way out":
```kotlin
        var ret = listOf() // reasonable default
        var ssws: String? = SecretHelper.getSecretService().fetchSecret("SSWS_OKTA")

        if( ssws !== null ){
                var grp = convertOrgToGroup( org );

                // get the OKTA Group Id
                var response1 = httpGet( url="https://hhs-prime-admin.okta.com/api/v1/groups?q=${grp}", 
                                       headers=mapOf( "Authorization" to "SSWS ${ssws}" ) );
                var grpId = ((response1.jsonArray).get( 0 ) as JSONObject).getString("id");

                // get the users within that OKTA group
                var response = httpGet( url="https://hhs-prime-admin.okta.com/api/v1/groups/${grpId}/users", 
                                       headers=mapOf( "Authorization" to "SSWS ${ssws}" ) );
    
               var emails :MutableList<String> = mutableListOf()

                for( user in response.jsonArray ){
                    emails.add( (user as JSONObject).getJSONObject("profile").getString("email") );
                } 

                ret = emails
        }
        else {
            context.logger.warning("Can't find SSWS_OKTA secret")
        }

        return ret;
```

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
```diff
@@ -0,0 +1,328 @@
...
+
+     /**
+     * Get the reports that have been generated since a given date
+     * 
+     * @param org
+     * 
+     */
+    private fun getReportsSinceLast( org: String, last: Date ): List<ReportFile> {
+
+      val reportFiles = workflowEngine
```
td-usds, line 273: Nitpick: formatting; also consider just returning and not storing in the variable (for conciseness).

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
```diff
@@ -0,0 +1,328 @@
...
+                        .fetchDownloadableReportFiles( 
+                                last.toInstant().atOffset(ZoneOffset.UTC), 
+                                org )
+
+        return reportFiles;
+    }
+
+     /**
+     * TODO: Fixme!
```
td-usds, line 282: Fix or remove

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
```diff
@@ -0,0 +1,328 @@
...
+     */
+    @Suppress( "UNRESOLVED_REFERENCE")
+    private fun dispatchToSendGrid( 
+        template: String,
+        emails: List<String>,
```
td-usds, line 287: Nitpick: this method only ever uses the `emails` to iterate over it; consider asking for it as an `Iterable<String>` instead of as a list.
That makes your design intentions clear: I am asking for something and will only iterate over it. Any changes to behavior in the method's implementation will now show up as a clear deviation to that initial design because it will require that you change the parameter type.
Think of this as both broadening your inputs and tightening your own operations:
- Broadening: more things can be used to call this method (more things are `Iterable<String>` than just `List<String>` so you can now call this method with _any_ of those
- tightening: this method intends to only iterate over emails, by asking for it as an `Iterable<String>`, you ensure that that is all you _can_ do and you give a guarantee to your caller that you won't modify this thing in any way.

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
```diff
@@ -0,0 +1,328 @@
...
+        reportsSinceLast: List<ReportFile>,
+        context: ExecutionContext
```
td-usds, line 289: You only use the logger from this,k pass in the logger instead of the whole context

prime-router/start_func.sh
```diff
@@ -23,4 +23,4 @@ fi
 
 cd $function_folder
 
-func host start --cors http://localhost:8090 --language-worker -- "-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=*:5005"
\ No newline at end of file
+func host start --cors http://localhost:8090,http://localhost:3000 --language-worker -- "-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=*:5005"
```
td-usds, line 26: Is this an intended change?

prime-router/build.gradle.kts
```diff
@@ -435,6 +437,9 @@ dependencies {
     implementation("com.google.guava:guava:30.1.1-jre")
     implementation("com.helger.as2:as2-lib:4.7.1")
 
+    implementation("com.cronutils:cron-utils:9.1.5")
```
td-usds, line 440: This hits on this PR template question and needs extra info:
Does this change introduce new dependencies that need vetting?

td-usds (changes requested):

prime-router/src/main/kotlin/azure/SettingsFacade.kt
```diff
@@ -59,6 +59,7 @@ class SettingsFacade(
     }
 
     override fun findSender(fullName: String): Sender? {
+        System.out.println( "fullName = ${fullName}" )
```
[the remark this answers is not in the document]
mry-usds: Will remove - debug artifact

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
```diff
@@ -0,0 +1,328 @@
...
+    val parameters: Map<String,String> = HashMap<String,String>()
+) {}
+ 
+class EmailScheduleEngine  {
+
+    val workflowEngine = WorkflowEngine()
+
+    @FunctionName("emailScheduleEngine")
+    @StorageAccount("AzureWebJobsStorage")
+    @Suppress( "UNUSED_PARAMETER" )
```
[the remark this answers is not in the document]
mry-usds: Removed

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
```diff
@@ -0,0 +1,328 @@
...
+    }
+
+    /**
+     * Reports the last time that the timer has been fired for this schedule
+     * 
+     * @param schedule the schedule to check the last time fired against
+     * @returns Date of the last time fired
+     */
+    private fun getLastTimeFired( schedule: EmailSchedule ): Date {
+        val parser = CronParser( CronDefinitionBuilder.instanceDefinitionFor(CronType.UNIX) );
```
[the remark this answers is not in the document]
mry-usds: Removed

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
mry-usds: Changed name to encodeOrg - centralizing will come later

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
mry-usds: Its not always the case that it will only be iterated over...this is a first cut at the function for later modifications

prime-router/src/main/resources/db/migration/V13__create_email_tables.sql
```diff
@@ -0,0 +1,28 @@
...
+    -- Key
+    email_schedule_id SERIAL PRIMARY KEY,
+
+    -- Value
+    values JSON,
+    is_deleted BOOLEAN,
+    is_active BOOLEAN,
+
+    -- Metadata
+    version INT NOT NULL,
```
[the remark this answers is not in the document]
Bisonburger: Dont want to constrain version to non-negatives (may be required later)

prime-router/src/main/resources/db/migration/V13__create_email_tables.sql
```diff
@@ -0,0 +1,28 @@
...
+    created_by VARCHAR(63) NOT NULL,
```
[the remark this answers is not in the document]
Bisonburger: Dont want to constrain version to non-negatives (may be required later)

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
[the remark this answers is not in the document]
Bisonburger: Removed

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
Bisonburger: Added documentation

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
Bisonburger: As requested

prime-router/start_func.sh
Bisonburger: Yes - this supports react

prime-router/src/main/resources/db/migration/V13__create_email_tables.sql
```diff
@@ -0,0 +1,28 @@
...
+Email schedules are kept a table where the actual schedule values are in a JSON field.
+Each schedule is versioned to keep around old values for lineage auditing. Deleted schedules
+are tomb-stoned, a row is never deleted
+*/
+CREATE TABLE email_schedule (
+    -- Key
+    email_schedule_id SERIAL PRIMARY KEY,
+
+    -- Value
+    values JSON,
```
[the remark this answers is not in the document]
Bisonburger: This follows the convention established with the setting API

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
td-usds: Are we expecting in the future to do more than iterate over it? Based on the logic and name, it doesn't appear that way.
I guess generally speaking: yes, I agree (with caveats), but here, I would change it.

Bisonburger:
Agree – change made

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
```diff
@@ -0,0 +1,328 @@
...
+     */
+    private fun shouldFire( schedule: EmailSchedule ): Boolean{
+
+        val parser = CronParser( CronDefinitionBuilder.instanceDefinitionFor(CronType.UNIX) );
+        // Get date for last execution
+        val now = ZonedDateTime.now();
+        val executionTime = ExecutionTime.forCron(parser.parse(schedule.cronSchedule));
+        val timeFromLastExecution = executionTime.timeFromLastExecution(now);
+
+        return ( timeFromLastExecution.get().toSeconds() <= 4*60);
```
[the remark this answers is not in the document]
Bisonburger: Added comment

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
```diff
@@ -0,0 +1,328 @@
...
+        @Suppress( "NEW_INFERENCE_NO_INFORMATION_FOR_PARAMETER" )    
+        var ret = workflowEngine.db.transactReturning {  tx -> 
+            @Suppress( "UNRESOLVED_REFERENCE")
+            val settings = workflowEngine.db.fetchSettings( SettingType.ORGANIZATION, tx )
+            @Suppress( "NEW_INFERENCE_NO_INFORMATION_FOR_PARAMETER" )
+            settings.map{ it.getName() }
+        }
+       
+        @Suppress("OVERLOAD_RESOLUTION_AMBIGUITY")
+        System.out.println( ret );
```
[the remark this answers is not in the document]
Bisonburger: Removed

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
```diff
@@ -0,0 +1,328 @@
...
+ 
+class EmailScheduleEngine  {
+
+    val workflowEngine = WorkflowEngine()
+
+    @FunctionName("emailScheduleEngine")
+    @StorageAccount("AzureWebJobsStorage")
+    @Suppress( "UNUSED_PARAMETER" )
+    fun run(
+        @TimerTrigger( name = "emailScheduleEngine", schedule = "*/5 * * * *") timerInfo : String,
```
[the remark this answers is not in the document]
Bisonburger: You can actually call it explicitly - but the "happy path" fires off a trigger every 5 minutes

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
Bisonburger: Agreed - done

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
Bisonburger: agreed - but that refactoring is happening via @ahay-agile6

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
Bisonburger: OktaCommands.kt is in the cli package - but created my own consts at the head of the file

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
Bisonburger: Agree - done

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
Bisonburger: Id is passed back in body...

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
Bisonburger: Agree - done

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
Bisonburger: OktaCommands.kt is in cli package - created my own const at head of file

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
Bisonburger: Agree - done

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
Bisonburger: Agree - done

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
```diff
@@ -0,0 +1,341 @@
...
+        ) request: HttpRequestMessage<String?>,
+        @BindingName("scheduleId") scheduleId: Int,
+        context: ExecutionContext,
+    ): HttpResponseMessage {
+        var user:String? = validateUser( request, context.logger );
+        var ret = request.createResponseBuilder(HttpStatus.UNAUTHORIZED);
+
+        if( user !== null ){
+            val id = WorkflowEngine.databaseAccess.deleteEmailSchedule(scheduleId);
+            ret.status(HttpStatus.CREATED)
```
td-usds, line 122: Should this not be an `HttpStatus.OK`? We didn't create anything, this is a `DELETE` verb.

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
```diff
@@ -0,0 +1,341 @@
...
+        ) request: HttpRequestMessage<String?>,
+        context: ExecutionContext,
+    ): HttpResponseMessage {
+        var user:String? = validateUser( request, context.logger );
+        var ret = request.createResponseBuilder(HttpStatus.UNAUTHORIZED);
+
+        if( user !== null ){
+            val id = WorkflowEngine.databaseAccess.insertEmailSchedule(request.body, user );
+            ret.status(HttpStatus.CREATED)
+            ret.body( "created id $id")
```
td-usds, line 97: For automation purposes, I think it would make more sense to just return the id without any additional strings (i.e. `ret.body( "$id")`). I'm think about people wanting to do something with the result in a script that uses `curl`.

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
```diff
@@ -0,0 +1,341 @@
...
+        @BindingName("scheduleId") scheduleId: Int,
+        context: ExecutionContext,
+    ): HttpResponseMessage {
+        var user:String? = validateUser( request, context.logger );
+        var ret = request.createResponseBuilder(HttpStatus.UNAUTHORIZED);
+
+        if( user !== null ){
+            val id = WorkflowEngine.databaseAccess.deleteEmailSchedule(scheduleId);
+            ret.status(HttpStatus.CREATED)
+            ret.body( "deactivated id $id")
```
td-usds, line 123: See line 97: just return `id`

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
```diff
@@ -0,0 +1,341 @@
...
+
+        schedulesToFire.forEach { schedule -> 
+            val orgs: Iterable<String> = getOrgs( schedule );
+
+            logger.info( "processing ${schedule.template}" )
+
+            // get the orgs to fire for
+            orgs.forEach{ org ->
+                    val emails: List<String> = if ( schedule.emails.size > 0) schedule.emails else getEmails(org, logger)
+                    logger.info( "processing ${org}" )
```
td-usds, line 154: I would make this a bit more explanatory in terms of _what_ we are processing. Otherwise we may be making our lives harder than needed when we need to trawl through logs to figure out what's happening.

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
```diff
@@ -0,0 +1,341 @@
...
+     */
+    private fun shouldFire( schedule: EmailSchedule ): Boolean{
+
+        val parser = CronParser( CronDefinitionBuilder.instanceDefinitionFor(CronType.UNIX) );
+        // Get date for last execution
+        val now = ZonedDateTime.now();
+        val executionTime = ExecutionTime.forCron(parser.parse(schedule.cronSchedule));
+        val timeFromLastExecution = executionTime.timeFromLastExecution(now);
+
+        return ( timeFromLastExecution.get().toSeconds() <= 4*60 /* 4 minutes */);
```
td-usds, line 178: I am not sure I 100% understand (and the cron-utils docs aren't super clear either - I'm looking at https://github.com/jmrozanec/cron-utils/blob/master/src/main/java/com/cronutils/model/time/ExecutionTime.java and https://github.com/jmrozanec/cron-utils/blob/master/src/main/java/com/cronutils/model/time/SingleExecutionTime.java)

Is my understanding correct that...
```kotlin
val timeFromLastExecution = executionTime.timeFromLastExecution(now);
```
...grabs a duration objection that contains how long ago it was since the last time (something  in the past) this schedule would have fired?

If so, then should this `<=` be a `>=` instead?

I guess the second part here is: what is this 4 minute timeout doing here and why is it different from the (default) 5 minute timeout used by the schedule in
```kotlin
/*56:  */ const val SCHEDULE = "*/5 * * * *" // every 5 minutes
/* ... */
/*134: */ fun run(
/*135: */         @TimerTrigger( name = "emailScheduleEngine", schedule = SCHEDULE ) timerInfo : String,
/*136: */         context: ExecutionContext
/*137: */     )
```

I think turning this magic value of 4 minutes into a constant with an explanatory name might clear up some things.

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
```diff
@@ -0,0 +1,341 @@
...
+                fetchAllOrgs() 
+        )
+    }
+
+    /**
+     * Retrieves the list of all organization supported
+     * 
+     * @returns List of all organizations supported
+     */
+    private fun fetchAllOrgs(): List<String>{
```
td-usds, line 202: Suggestion to make this return an `Iterable<String>`

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
```diff
@@ -0,0 +1,341 @@
...
+    fun createEmailSchedule(
+        @HttpTrigger(
+            name = "createEmailSchedule",
+            methods = [HttpMethod.POST],
+            authLevel = AuthorizationLevel.ANONYMOUS,
+            route = "email-schedule"
+        ) request: HttpRequestMessage<String?>,
+        context: ExecutionContext,
+    ): HttpResponseMessage {
+        var user:String? = validateUser( request, context.logger );
```
td-usds, line 91: Reading `validateUser`, any user with a valid okta jwt token containing a subject (`sub`) will be able to create (and below: delete) e-mail schedules. Is this desired? Does there need to be any validation that you are touching only those schedules you are allowed to touch?

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
```diff
@@ -0,0 +1,341 @@
...
+    }
+
+    /**
+     * Retrieve a list of emails within an organization
+     * 
+     * @params org The organization to fetch emails for
+     * 
+     * @returns List of emails to send to
+     */
+    private fun getEmails( org: String, logger: Logger): List<String> {
```
td-usds, line 265: Suggestion to make this return an `Iterable<String>`

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
```diff
@@ -0,0 +1,341 @@
...
+     * 
+     * @param request the HTTPRequest object
+     * @param logger logger
+     * 
+     * @returns user from the token; otherwise null
+     */
+    fun validateUser( request: HttpRequestMessage<String?>, logger: Logger): String? {
+        var jwtToken = request.headers["authorization"] ?: ""
+
+        jwtToken = if (jwtToken.length > 7) jwtToken.substring(7) else ""
```
td-usds, line 223: This does eerily similar things to https://github.com/CDCgov/prime-reportstream/blob/6abe7c7867e3776eafa36fc802fb0cd4a63d4749/prime-router/src/main/kotlin/azure/HistoryFunctions.kt#L521; this is a good candidate to merge these things together to avoid desynchronization between these different places in the future.

I'm fine 'postponing' that to some later point but if that is the decided course, then I would like to see the GitHub Issue number of that postponed work here before resolving the conv.

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
```diff
@@ -0,0 +1,341 @@
...
+                // get the users within that OKTA group
+                var response = httpGet( url="${OKTA_GROUPS_API}/${grpId}/users", 
+                                    headers=mapOf( "Authorization" to "SSWS ${ssws}" ) );
+            
+                for( user in response.jsonArray )
+                    emails.add( (user as JSONObject).getJSONObject("profile").getString("email") );
+            }
+        }
+        catch( ex: Throwable ){
+            logger.warning( "Error in fetching emails" )
```
td-usds, line 291: To make our future lives easier, can you include some information in this log entry as to why the e-mails could not be fetched (while still not including any private information)?
The exception could be due to a variety of reasons and dependencies. It could be due to something Okta related, it could be due to something network related, but it could also be due to something related to a user not being able to be parsed as a JSON object. Having that information in the warning will be useful when debugging.

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
```diff
@@ -0,0 +1,341 @@
...
+
+        val mail : Mail = Mail()
+        mail.addPersonalization( p );
+        mail.setSubject(SUBJECT_EMAIL);
+        mail.setFrom( Email(FROM_EMAIL) );
+        mail.setTemplateId( template )
+
+        var sendgridId: String? = SecretHelper.getSecretService().fetchSecret("SENDGRID_ID")
+        
+        if( sendgridId !== null ){
```
td-usds, line 322: safety: make sure it isn't null nor empty instead of just null

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
```diff
@@ -0,0 +1,341 @@
...
+                request.setMethod(Method.POST);
+                request.setEndpoint("mail/send");
+                request.setBody(mail.build());
+                sg.api(request);
+            } catch (ex:IOException) {
+                logger.warning("Can't contact sendgrid")
+            }
+        }
+        else{
+            logger.info("Can't find SENDGRID_ID secret")
```
td-usds, line 336: It may be useful to output whether the `SENDGRID_ID` is `null` or empty (`""`).

td-usds (dismissed):

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
mry-usds: Good point - should be active for only DHPRIMEADMINS; changing

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
mry-usds: Agree - changing

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
mry-usds: Cut/Paste error - changing

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
mry-usds: Agree

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
mry-usds: OK

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
mry-usds: So, the timer function doesn't fire on exactly every 5 minutes, so this is the "catch" at assure that anytime within less than 5 minutes of the last timer execution is valid (i.e. timer trigger fires at 11:59 but the cron for the schedule is at noon, without this, they don't "line up" and the schedule is always reported as shouldFire = false)  This accounts for a difference in cron timers.

prime-router/build.gradle.kts
```diff
@@ -377,9 +377,11 @@ tasks.register("quickPackage") {
 
 repositories {
```
mry-usds, line 378: Added jitpack.io as repo

mry-usds (approved):

mry-usds (approved):

mry-usds (approved):

mry-usds (approved):

PR #1496, opened 2021-07-06
1471 Ignore errors that the blob already exists

### LINKED ISSUE #1471: Bug: rapid submissions in succession sporadically fail with 500 error
This problem was originally noted as:
`./prime test --run merge
`
 failing now sporadically with a 500 error, when run locally.  I think this may have something to do with the new metadata feature (see Exception in the comment below)?  It seems to occur most right after you restart the docker container?  Maybe some weird race condition with a transaction that is closing? 

See comment below for way to reproduce the problem.

Acceptance criteria:
1. Identify why this issue is happening.  Document the problem in this issue.
2. Fix the issue if it can be fixed.

FYI @MauriceReeves-usds

carlosfelix2:
This PR ignores an error that the Blob already exists.  This is not really an error for us, but can occur when concurrent calls are made to the API and those calls all try to create the Blob store at the same time.

## Changes
- ignore blob already created error

### Testing
- [x] Tested locally?
- [ ] Ran `quickTest all`?
- [ ] Added tests?

### Process

## Fixes
#1471

### DIFF
--- a/prime-router/docker-compose.yml
+++ b/prime-router/docker-compose.yml
@@ -72,7 +72,7 @@ services:
 
   settings:
     build: settings/.
-    command: "--wait=25 --check-last-modified prime_dev /settings/organizations-local.yml"
+    command: "--wait=20 --check-last-modified prime_dev /settings/organizations-local.yml"
     depends_on:
       - prime_dev
     volumes:
--- a/prime-router/docker-compose.yml
+++ b/prime-router/docker-compose.yml
@@ -72,7 +72,7 @@ services:
 
   settings:
     build: settings/.
-    command: "--wait=15 --check-last-modified prime_dev /settings/organizations-local.yml"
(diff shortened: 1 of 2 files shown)


prime-router/docker-compose.yml
```diff
@@ -72,7 +72,7 @@ services:
 
   settings:
     build: settings/.
-    command: "--wait=15 --check-last-modified prime_dev /settings/organizations-local.yml"
+    command: "--wait=20 --check-last-modified prime_dev /settings/organizations-local.yml"
```
jimduff-usds, line 75: Bleh.

jimduff-usds (approved):
Better would be the ability to do the exists() check and the create() as an atomic operation, but I presume no such thing exists in Azure.

carlosfelix2:
> Better would be the ability to do the exists() check and the create() as an atomic operation, but I presume no such thing exists in Azure.

It used to exist, but they removed it.

PR #1521, opened 2021-07-07
Td/dct docs

td-usds:
This PR adds some documentation regarding Docker Container Trust. I've attached the pdf version of the markdown to this PR as well.
[DCT.pdf](https://github.com/CDCgov/prime-reportstream/files/6778853/DCT.pdf)

Note that this is just documentation about DCT, it will be expanded later on with Azure-specifics.


ronaldheft-usds (approved):
Sounds straightforward for signing containers.

PR #1526, opened 2021-07-07
1390 - Refactor Bearer token authentication into reusable class

ahay-agile6:
Hey @jimduff-usds || @RickHawesUSDS , maybe @pablocaviglia as well,  this is a draft PR to see if I am on the right track for using the Okta authentication service in a reusable way.
I took the code from the SettingsFunctions class and placed it into its own class to be reusable.
The plan is to move it over to the HistoryFunctions next.

Please review this to see if I am on the right track with what you all were thinking.

I am definitely open to feedback!

Thank you :)

-----------

This PR refactors the token auth into a reusable class

## Changes
- Code about the bearer token is placed into a class to be used elsewhere
- settings database setup script send another header to set up the database without authentication **ONLY in a local environment**

### Testing
- [ ] Tested locally?
- [ ] Ran `quickTest all`?
- [ ] Added tests?

    - new endpoint A
    - new endpoint B

### Process

## Fixes
- #issue

## To Be Done
- #issue

### DIFF
--- a/prime-router/settings/put-local-settings.py
+++ b/prime-router/settings/put-local-settings.py
@@ -18,7 +18,7 @@ protocol = "http"
 org_file = args.org_file
 
 base_url = f"{protocol}://{host}:7071/api"
-headers = {"Authorization": "bearer xyz", "Content-Type": "application/json"}
+headers = {"Authorization": "bearer xyz", "Content-Type": "application/json", "SettingsSetup": "true"}
 
 
 def parse_file(file_name):
--- a/prime-router/settings/put-local-settings.py
+++ b/prime-router/settings/put-local-settings.py
@@ -18,7 +18,7 @@ protocol = "http"
 org_file = args.org_file
 
 base_url = f"{protocol}://{host}:7071/api"
-headers = {"Authorization": "bearer xyz", "Content-Type": "application/json", "SettingsSetup": "true"}
+headers = {"Authorization": "bearer xyz", "Content-Type": "application/json", "LocalNoAuth": "true"}
 
 
 def parse_file(file_name):
--- a/prime-router/src/main/kotlin/tokens/OktaAuthentication.kt
+++ b/prime-router/src/main/kotlin/tokens/OktaAuthentication.kt
@@ -45,10 +45,10 @@ class OktaAuthentication(private val minimumLevel: PrincipalLevel = PrincipalLev
         }
     }
 
-    fun handleRequest(
+    fun checkAccess(
         request: HttpRequestMessage<String?>,
         organizationName: String,
-        block: (claims: AuthenticatedClaims) -> HttpResponseMessage
+        block: (AuthenticatedClaims) -> HttpResponseMessage
     ): HttpResponseMessage {
         try {
             val accessToken = getAccessToken(request)
@@ -83,16 +83,16 @@ class OktaAuthentication(private val minimumLevel: PrincipalLevel = PrincipalLev
     /**
      * For endpoints that need to check if the organization is in the database
      */
-     fun checkOrganizationExists(context: ExecutionContext, userName: String, orgName: String?): Organization? {
+    fun checkOrganizationExists(context: ExecutionContext, userName: String, orgName: String?): Organization? {
         var organization: Organization? = null
         if (orgName != null) {
-                organization = WorkflowEngine().settings.findOrganization(orgName.replace('_', '-'))
-                if (organization != null) {
-                    return organization
-                } else {
-                    context.logger.info("User $userName failed auth: Organization $orgName is unknown to the system.")
-                }
+            organization = WorkflowEngine().settings.findOrganization(orgName.replace('_', '-'))
+            if (organization != null) {
+                return organization
+            } else {
+                context.logger.info("User $userName failed auth: Organization $orgName is unknown to the system.")
             }
+        }
         return organization
     }
 
--- a/prime-router/src/main/kotlin/tokens/OktaAuthentication.kt
+++ b/prime-router/src/main/kotlin/tokens/OktaAuthentication.kt
@@ -1,0 +1,74 @@
+package gov.cdc.prime.router.tokens
+
+import com.google.common.net.HttpHeaders
+import com.microsoft.azure.functions.HttpRequestMessage
+import com.microsoft.azure.functions.HttpResponseMessage
+import gov.cdc.prime.router.azure.*
+import org.apache.logging.log4j.kotlin.Logging
+
+class OktaAuthentication(private val minimumLevel: PrincipalLevel = PrincipalLevel.USER): Logging {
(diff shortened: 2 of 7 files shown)


prime-router/settings/put-local-settings.py
```diff
@@ -18,7 +18,7 @@
 org_file = args.org_file
 
 base_url = f"{protocol}://{host}:7071/api"
-headers = {"Authorization": "bearer xyz", "Content-Type": "application/json"}
+headers = {"Authorization": "bearer xyz", "Content-Type": "application/json", "SettingsSetup": "true"}
```
ahay-agile6, line 21: Sending this header makes it so the local environment can set up the Settings database without authentication, and we test authentication in the actual application where appropriate.
I check against the PRIME_ENVIRONMENT and this header in the Okta Authentication class.
Open to other suggestions :laughing:

prime-router/src/main/kotlin/tokens/OktaAuthentication.kt
```diff
@@ -0,0 +1,92 @@
...
+import com.microsoft.azure.functions.HttpResponseMessage
+import gov.cdc.prime.router.Organization
+import gov.cdc.prime.router.azure.*
+import org.apache.logging.log4j.kotlin.Logging
+
+class OktaAuthentication(private val minimumLevel: PrincipalLevel = PrincipalLevel.USER): Logging {
+    private val missingAuthorizationHeader = HttpUtilities.errorJson("Missing Authorization Header")
+    private val invalidClaim = HttpUtilities.errorJson("Invalid Authorization Header")
+
+    fun getAccessToken(request: HttpRequestMessage<String?>): String? {
```
RickHawesUSDS, line 15: Consider putting this function into HttpUtilitlies so other people can use it. @jimduff-usds for example.

RickHawesUSDS (approved):
I like it, because of the reuse of the existing work. 

Because this is security related, please get another approver

prime-router/src/main/kotlin/tokens/OktaAuthentication.kt
pablocaviglia-baires: agree on this, good idea to have it there, it will be useful

pablocaviglia: agree on this, good idea to have it there, it will be useful

jimduff-usds (commented):
I see you've moved this back to draft.   I think its fine to leave your code in the tokens/ dir.

I need to make some changes to my PR as well, then I'll re-review your work and we can merge it into mine, then into master.

prime-router/src/main/kotlin/tokens/OktaAuthentication.kt
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    companion object: Logging {
+        private var httpRequestMessage: HttpRequestMessage<String?>? = null
+
+        fun setRequest(request: HttpRequestMessage<String?>) {
+            httpRequestMessage = request
+        }
+
+        val claimVerifier: AuthenticationVerifier by lazy {
```
jimduff-usds, line 28: I think this will not do what you want -- since this is set only once, and since settings setup is usually the very first thing that runs, then all local calls to the endpoint after that will go to TestAuthenticationVerifier, even calls not from put-local-settings.py.

If you want to use a header parameter to "turn off" authentication locally, I suggest changing this to a function call, that gets the Verifier on-the-fly, on every endpoint call.   

Also, this ability to "turn off' auth occasionally seems like a general thing, not just for Settings Setup.  Maybe just give it a more general name?

I'm also worried about someone inadvertently messing up this code somehow, and this capability leaking into Prod.

prime-router/settings/put-local-settings.py
jimduff-usds: I get what you are doing here:   Currently, Rick's code (that you have now moved into OktaAuthentication.kt) turns off all Okta authentication if PRIME_ENVIRONMENT is 'local', and you want to instead default to having Okta auth "on" locally, unless this header is sent. 

A couple thoughts:  Not just for settings setup, but for almost all our endpoints Its super convenient (or maybe lazy) to not have to authenticate when local.  

Also, while I realize this feature sort of actually increases security, I'm worried this makes it just a bit easier to mistakenly mess up the code in OktaAuthentication and create a breach.

Are there other options?   For example, maybe if PRIME_ENVIRONMENT is 'local' it uses Okta, but if PRIME_ENVIRONMENT is 'local_noauth' it doesn't?   Or maybe a second env variable?  other ideas?

prime-router/src/main/kotlin/tokens/OktaAuthentication.kt
```diff
@@ -0,0 +1,92 @@
+package gov.cdc.prime.router.tokens
+
+import com.google.common.net.HttpHeaders
+import com.microsoft.azure.functions.ExecutionContext
+import com.microsoft.azure.functions.HttpRequestMessage
+import com.microsoft.azure.functions.HttpResponseMessage
+import gov.cdc.prime.router.Organization
+import gov.cdc.prime.router.azure.*
```
jimduff-usds, line 8: Our style is to not use .*
Can you change this? Thanks!

jimduff-usds (changes requested):
This is a huge step in the right direction!

See my comments on thinking about the SettingsSetup feature -- @ronaldheft-usds may have useful thoughts here too.

Also, a minor request on "*" imports.

I'd like to do another pass thru this myself, but I'm out of time at the moment - so I suspect I'll submit another review!)

prime-router/src/main/kotlin/tokens/OktaAuthentication.kt
jimduff-usds: @mauricereeves tells me that we plan to soon require even local dev usage to go thru Otka.  This would imply that we just get rid of the TestAuthenticationVerifier.

jimduff-usds (commented):
Aurora, I'd like to try running this again after the proposed change to the retrieval of the claimVerifier.    

In the meantime, there are a few change requests by Rick to my FHIR Auth code that need to be done by me, then I think we can merge your code into that one, then that into master, hopefully in the next few days.

prime-router/src/main/kotlin/tokens/OktaAuthentication.kt
ahay-agile6: done!

ahay-agile6:
Hey @jimduff-usds 
I addressed the change to the retrieval of the `claimVerifier` - even though I understand this may be going away. Let me know if that will work, or if I did it wrong  :stuck_out_tongue_winking_eye:  I really did it to learn and understand the code base
I also addressed the "*" import issue.
Does everything else look good? Good enough to merge into your branch, at least? Hah.

jimduff-usds (approved):

Issue #1236, opened 2021-06-07
Clear Dependabot Security Vulnerabilities


MauriceReeves-usds:
The list of security vulnerabilities we are being warned about by dependabot has increased. It is important we get these resolved as soon as possible: https://github.com/CDCgov/prime-data-hub/security/dependabot

PR #1540, opened 2021-07-08
Updates node packages to clear dependabot security vulnerabilities

MauriceReeves-usds:
This PR updates the devDependencies in the package lock file to clear four security vulnerabilities currently present in our ReportStream website. Given the fact that there was a transitive dependency I tested this by running the static site locally. If you want to do the same to verify functionality (and you should), you need to follow these steps:

- Open a terminal into the `frontend` folder
- Run `npx eleventy --config eleventy.config.js` to generate the files
- Run `npx eleventy --serve --config eleventy.config.js` to launch the site
- You will see two URLs presented to you, one labeled `local` and the other `ui`. The local is a copy of our static site running on a port like `localhost:8091`. Your port may be different. The `ui` link is to browsersync which lets you live reload the site and do some configuration
- If your port is not `8091` you must now log in to the Okta admin console and grant `localhost:<YOUR PORT>` access to Okta CORS and add your local address to the ReportStream (Test) application
- You will also need to update the start_func.sh script to use your port, and not the 8090 that is currently checked into code
- Once that's all done, spin up Docker and then log in to the application locally
- At this point you should see the site work, and results load locally from your Docker container

If you have any questions about the above, let me know and I will help you work through them.

## Changes
- Updates dependencies in `package-lock.json`

### Testing
- [x] Tested locally?
- [ ] Ran `quickTest all`?
- [ ] Added tests?

### Process

## Fixes
- #1236

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -1,7 +1,6020 @@
 {
+  "name": "frontend",
   "version": "0.0.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "version": "0.0.0",
+      "dependencies": {
+        "@okta/jwt-verifier": "^2.2.0",
+        "axios": "^0.21.1",
+        "chart.js": "^3.4",
+        "cypress": "^7.6.0",
+        "http-proxy-middleware": "^2.0.1",
+        "moment": "^2.29.1"
+      },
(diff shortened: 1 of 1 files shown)


drew-usds (approved):
11ty runs ok on my localhost after the updates.

Someone else should check out the start_func.sh/okta admin/docker part though, I don't have that set up on my local machine.

PR #1543, opened 2021-07-08
Updates docs to reflect the current reality

MauriceReeves-usds:
Some of the docs have not kept up with the times. They needed updated. They are now.


carlosfelix2 (approved):
Oh, man, I missed quite a few things in that getting started doc.  Sorry, and thanks for updating it.

PR #1545, opened 2021-07-08
1384 Added smoke test to gradle script

### LINKED ISSUE #1384: Create Smoke Test Task in Gradle and Add Coded Tests to It
### Overview
I think we should add a smokeTests task to gradle that will run a series of tests we've built into the project right on the Docker container. The intention is to do a full and complete end-to-end set of tests directly from gradle that will submit files to ReportStream running in Docker and then verify the results.

This is obviously similar to what `quick-test` and `./prime test` do now, but I think that we can expand on their checking and capabilities and run it from gradle.

I don't think that want to set this up as an automatic task that runs on every build because it will take a while and be expensive, but I think that having it in gradle will make it easier to run, and we can potentially add it as a GitHub task that has to run before a PR is merged.

carlosfelix2:
This PR adds the smoke test to the Gradle script.  I also refactored the CLI portion a bit to be able to create other tasks to call the CLI without having to copy the full CLI task code.  I tried once again to split the gradle script, so it is not a large file, but Gradle does not have the ability to do that.  :( 

To test:
Run `./gradlew testSmoke'

## Changes
- Added testSmoke task
- A bit of refactor to make future CLI tasks simpler to make.

### Testing
- [x] Tested locally?
- [ ] Ran `quickTest all`?
- [ ] Added tests?

### Process

## Fixes
#1384

### DIFF
--- a/prime-router/build.gradle.kts
+++ b/prime-router/build.gradle.kts
@@ -207,11 +207,15 @@ tasks.register<JavaExec>("primeCLI") {
     main = primeMainClass
     classpath = sourceSets["main"].runtimeClasspath
     // Default arguments is to display the help
-    args = listOf("-h")
     environment["POSTGRES_URL"] = dbUrl
     environment["POSTGRES_USER"] = dbUser
     environment["POSTGRES_PASSWORD"] = dbPassword
-    doFirst {
+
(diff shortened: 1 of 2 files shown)


jimduff-usds (approved):
Thanks!

Issue #1474, opened 2021-07-05
Further work on verbose mode; remove the "data went nowhere" for now.


jimduff-usds:
I kept messing with the language in the PR until I got to this:
```

  "destinations" : [ {
    "organization" : "Note: some data is not routing to any state/local jurisdiction",
    "organization_id" : "n/a",
    "service" : "n/a",
    "sending_at" : "n/a",
    "itemCount" : 10
  }
```

And realized that the real problem is I don't like that we're treating a notification/message like an organization.  So, I think I want to just get rid of it.  Sorry for sending you down this rabbit hole.

Let's do this instead:

- just get rid of the above destination for now.  if someone wants to know where their data failed to go, they have to use verbose mode.
- change `verbose=all` to `verbose=true`
- don't even show `"routing" : [ ],` at all, if verbose is not true.   It looks like nothing routed anywhere.
- test that  verbose=true plays nicely with option=SkipSend.  This will be a common use case, where you re-submit using both SkipSend and verbose, to safely figure out which data is not routing properly.   The routing info _with_ SkipSend should be completely identical to the routing printed _without_ SkipSend.
- add reportIndex as a field in the routing.   It should look like this example:

```
  "routing" : [ {
    "reportIndex" : 0,
    "trackingId" : "910498",
    "destinations" : [ "foo", "bar", "baz" ]
  }, {
   "reportIndex" : 1,
    "trackingId" : null,
    "destinations" : [ ]
  }, {
    "reportIndex" : 2
    "trackingId" : null,
    "destinations" : [ "az-phd.elr-prod", "pima-az-phd.elr" ]
  }, {

```
Do not key on trackingId.  Rather, key on the reportIndex, which is reliable.  Note the latter two examples show how to handle the case where there is no tracking id.

I suggest you remove PR #1231 and replace with a new PR.

PR #1549, opened 2021-07-09
Report routing in reports API end point response

tlosbo:
This PR addresses changes outlined in issue #1474 based on a previously closed PR. Code was removed and adjusted to accommodate the changes. 

## Changes
- added a new verbose parameter which shows the report routing when the value is true
- added an element to the Json response listing each item submitted in the report with the report index, tracking ID and list of receiving orgs and service destinations.

## Checklist
- [ ] submit a report with no parameters and ensure the response is same as before
- [ ] submit a report with `verbose=true` in the query string and verify/validate the routing element displays with this format:
```
"routing" : [ {
    "reportIndex" : 0,
    "trackingId" : "123456",
    "destinations" : [ "org1.service1", "org2.service2" ]
  }, {
    "reportIndex" : 1,
    "trackingId" : null,
    "destinations" : [ "hhsprotect.elr" ]
  },
  ...
]
```
- [ ] submit the same report as above with `verbose=true&option=SkipSend` and ensure the routing is exactly the same
- [ ] test all of the above on two different schemas, one with a trackingElement and one without

### Testing
- [x] Tested locally?
- [x] Ran `quickTest all`?
- [ ] Added tests?

    - new endpoint A
    - new endpoint B

### Process

## Fixes
- #1474 

## To Be Done
- #issue


prime-router/src/main/kotlin/azure/ReportFunction.kt
```diff
@@ -53,6 +55,13 @@ class ReportFunction {
         val defaults: Map<String, String> = emptyMap(),
         val routeTo: List<String> = emptyList(),
         val report: Report? = null,
+        val verbose: String = "",
+    )
+
+    data class ItemJurisdictionRouting(
+        val reportIndex: Int,
+        val trackingId: String?,
+        val destinations: Map<Organization.Jurisdiction, List<String>> = emptyMap(),
```
tlosbo, line 64: This could be changed back to a `List<String>` vs a `Map` keyed on Jurisdiction. Initially this map was for being able to find items that only routed to FEDERAL receivers. On the fence if we keep it this way or simplify it.

prime-router/src/main/kotlin/azure/ReportFunction.kt
```diff
@@ -418,4 +448,45 @@ class ReportFunction {
...
+            report.itemIndices.forEach { reportIndex ->
+                val trackingId = report.schema.trackingElement?. let {
+                    report.getString(reportIndex, report.schema.trackingElement)
+                }
+                val destinations = mutableMapOf<Organization.Jurisdiction, MutableList<String>>()
+                // find all outgoing reports based on the index and generate a mapped
+                // list of the receiving organizations keyed by the jurisdiction
+                actionHistory?. let { ah ->
+                    ah.itemLineages.filter { il ->
+                        il.parentIndex == reportIndex
```
jimduff-usds, line 476: Tom, as written is O(n squared) - you are iterating through the items in the report, and _for each one_ you are doing a filter of all the data in the itemLineages list, which also iterates thru a list!   This will be very slow for large reports.

At the same time, I think you can go ahead and remove the top level divvying by Jurisdiction type (eg FEDERAL, STATE, etc) if you like -- I think that would make it simpler to do this.

So, to do this in O(n) time, I think you can just walk the itemLineages list (only).    List<ItemJurisdictionRouting> should instead be a `Map<Integer, ItemLineageDetails>` where the Integer is the reportIndex, and ItemLineageDetails contains 
- the reportIndex
- the trackingId
- the jurisdiction
- and a destinations: List<String> where each String in destinations is a receiver fullName.
As you iterate thru itemLineages, you either create a new ItemLineageDetails if it doesn't exist, or if it already exists, you add a new receiver to the destinations list.

Be sure to sort the Map by the Integer key before printing the json.

A good test is this, where every other records go to both Pima County and AZ State:

`curl -X POST -H "client:simple_report"  -H "content-type:text/csv" --data-binary "@./src/test/csv_test_files/input/simplereport.csv"  "http://localhost:7071/api/reports"`

prime-router/src/main/kotlin/azure/ReportFunction.kt
jimduff-usds: Yes, per my other comment, I think you could get rid of the keying by Jurisdiction.

jimduff-usds (changes requested):
Asking for a refactor of createItemJurisdictionRouting per the comment there.  Thanks!

prime-router/src/main/kotlin/azure/ReportFunction.kt
tlosbo: This is a good catch. I had thought about writing a lookup map with the original PR. With the revelation that you can use report index vs trackingId I see how this is simplified now.

prime-router/src/main/kotlin/azure/ReportFunction.kt
```diff
@@ -395,6 +406,7 @@ class ReportFunction {
     ): String {
         val factory = JsonFactory()
         val outStream = ByteArrayOutputStream()
+        val itemRouting: List<ItemRouting> = createItemJurisdictionRouting(result, actionHistory)
```
pablocaviglia, line 409: I dont full grasp the logic inside the function **createItemJurisdictionRouting**, but wouldn't be possible to make the execution of this call to occur only if VERBOSE_TRUE = true (on line 423) ?
The logic inside the function is not easy to follow up, maybe my requeist is not a valid one, but I'm asking just to be sure.

prime-router/src/main/kotlin/azure/ReportFunction.kt
```diff
@@ -216,6 +224,10 @@ class ReportFunction {
                 )
             )
 
+        // extract the verbose param and default to empty if not present
+        val verboseParam = request.queryParameters.getOrDefault(VERBOSE_PARAMETER, "")
+        val verbose = if (verboseParam.equals(VERBOSE_TRUE, true)) verboseParam else ""
```
carlosfelix2, line 229: Why not make this verbose a Boolean?  There will be no need to know later what the value should be.

prime-router/src/main/kotlin/azure/ReportFunction.kt
```diff
@@ -30,7 +30,8 @@ private const val DEFAULT_PARAMETER = "default"
 private const val DEFAULT_SEPARATOR = ":"
 private const val ROUTE_TO_PARAMETER = "routeTo"
 private const val ROUTE_TO_SEPARATOR = ","
-
+private const val VERBOSE_PARAMETER = "verbose"
```
carlosfelix2, line 33: So, this is not truly a verbose flag, it is one to add routing to the response.  Should it be named differently?

carlosfelix2:
Provide information on how to test this PR.

prime-router/docs/releases/2021-07-22-release-notes.md
```diff
@@ -0,0 +1,28 @@
...
+
+- All Schemas are documented here:  [Link to detailed schema dictionaries](../schema_documentation)
+- The Hub API is documented here: [Hub OpenApi Spec](../openapi.yml)
+- [Click here for all release notes](../releases)
+
+## For this release
+
+### Changes to api/reports response
+
+This release adds a new `routing` element to the Json response if the query string contains a `verbose=true` parameter. This routing shows the destination(s) for each item in the report along with the report index and trackingId. The openapi.yml has been updated to reflect these changes with the structure of the routing element. An example of this routing element:
```
carlosfelix2, line 15: Perhaps add an example URL here for those that do not understand what a query is.

carlosfelix2 (approved):
Tested locally.

tlosbo (commented):
Changed to be a List<String> and dropped the Jurisdiction.

prime-router/src/main/kotlin/azure/ReportFunction.kt
```diff
@@ -418,4 +445,30 @@ class ReportFunction {
...
+     * @param actionHistory the instance generated while processing the report
+     * @return the report routing for each item
+     */
+    private fun createItemJurisdictionRouting(
+        validatedRequest: ValidatedRequest,
+        actionHistory: ActionHistory? = null,
+    ): List<ItemRouting> {
+        val routing = mutableMapOf<Int, ItemRouting>()
+        actionHistory?.let { ah ->
+            ah.itemLineages.forEach { il ->
```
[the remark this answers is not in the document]
jimduff-usds: Sounds right.  Thanks

jimduff-usds (approved):

PR #1564, opened 2021-07-12
Signed Container Images: enable trust policy in the registry and prepare for enforcement

td-usds:
This PR enables the Trust Policy (i.e. notary service) on our Azure Container Registry. This is step on towards making sure we run signed containers in our environment. (The second step is actual signing in the CI/CD pipeline - with validation - and the final step being enforcing it).

See https://github.com/CDCgov/prime-reportstream/blob/master/prime-router/docs/docker-content-trust.md for more information. 

NOTE: this should be validated in a developer resource group first, but due to logistical issues, I cannot do that (yet).

## Changes
- The main terraform code that manages the container registry (see `02-config/main.tf`) sets the value for `enable_content_trust` to `true` (note: the other highlighted changed lines are due to indentation)
- The terraform container_registry now accepts a boolean `enable_content_trust` variable that is set from the calling code

### Testing
- [ ] Tested locally?
- N/A Ran `quickTest all`?
- N/A Ran `./prime test` against local Docker ReportStream container?
- N/A Downloaded a file from `http://localhost:7071/api/download`?
- N/A Added tests?

- ~~Does this PR introduce new endpoints?~~
- ~~Does this PR include changes in authentication and/or authorization of existing endpoints?~~
- ~~Does this change introduce new dependencies that need vetting?~~
- Does this change require changes to our infrastructure? YES: it _is_ a change to our infrastructure
- ~~Does logging contain sensitive data?~~
- ~~Does this PR include or remove any sensitive information itself?~~

The change itself addresses and mitigates Spoofing, Tampering, and Repudiation. THe Information disclosure, Denial of Service, and Elevation of privilege threats are not relevant for this PR.

- ~~Have you ensured logging does not contain sensitive data?~~

### Process
- NO Are there licensing issues with any new dependencies introduced?
- NO Database changes are submitted as a separate PR?
- [X] DevOps team has been notified if PR requires ops support?

### DIFF
--- a/operations/app/src/environments/02-config/main.tf
+++ b/operations/app/src/environments/02-config/main.tf
@@ -39,9 +39,10 @@ module "key_vault" {
 }
 
 module "container_registry" {
-  source          = "../../modules/container_registry"
-  environment     = var.environment
-  resource_group  = var.resource_group
-  resource_prefix = var.resource_prefix
-  location        = var.location
+  source               = "../../modules/container_registry"
+  environment          = var.environment
+  resource_group       = var.resource_group
+  resource_prefix      = var.resource_prefix
+  location             = var.location
+  enable_content_trust = true
 }
--- a/operations/app/src/modules/container_registry/main.tf
+++ b/operations/app/src/modules/container_registry/main.tf
@@ -1,17 +1,21 @@
 resource "azurerm_container_registry" "container_registry" {
-  name = "${var.resource_prefix}containerregistry"
+  name                = "${var.resource_prefix}containerregistry"
   resource_group_name = var.resource_group
-  location = var.location
-  sku = "Premium"
-  admin_enabled = true
+  location            = var.location
+  sku                 = "Premium"
+  admin_enabled       = true
 
   network_rule_set {
     default_action = "Allow"
 
     virtual_network {
-      action = "Allow"
+      action    = "Allow"
       subnet_id = data.azurerm_subnet.public.id
     }
+  }
+
+  trust_policy = {
+    enabled = var.enable_content_trust
   }
 
   tags = {
--- a/operations/app/src/modules/container_registry/variables.tf
+++ b/operations/app/src/modules/container_registry/variables.tf
@@ -1,19 +1,24 @@
 variable "environment" {
-    type = string
-    description = "Target Environment"
+  type        = string
+  description = "Target Environment"
 }
 
 variable "resource_group" {
-    type = string
-    description = "Resource Group Name"
+  type        = string
+  description = "Resource Group Name"
 }
 
 variable "resource_prefix" {
-    type = string
-    description = "Resource Prefix"
+  type        = string
+  description = "Resource Prefix"
 }
 
 variable "location" {
-    type = string
-    description = "Container Registry Location"
+  type        = string
+  description = "Container Registry Location"
+}
+
+variable "enable_content_trust" {
+  type        = boolean
+  description = "Boolean value indicating enabledness of Content Trust"
 }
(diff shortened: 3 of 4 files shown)


operations/app/src/environments/02-config/main.tf
```diff
@@ -39,9 +39,10 @@ module "key_vault" {
...
-  environment     = var.environment
-  resource_group  = var.resource_group
-  resource_prefix = var.resource_prefix
-  location        = var.location
+  source               = "../../modules/container_registry"
+  environment          = var.environment
+  resource_group       = var.resource_group
+  resource_prefix      = var.resource_prefix
+  location             = var.location
+  enable_content_trust = true
```
ronaldheft-usds, line 47: Good call making this configurable.

ronaldheft-usds (commented):
Does the Container Registry need configuration to accept only signed code from specific certificates? The way I'm reading this now, it sounds like any valid signed container would work and not only containers signed with the specified CDC key.

td-usds:
> Does the Container Registry need configuration to accept only signed code from specific certificates? The way I'm reading this now, it sounds like any valid signed container would work and not only containers signed with the specified CDC key.

To the best of my understanding, this setting just _enables_ the Notary service on this registry. It does not control who can push up images, nor what gets pushed up, or which keys are used to sign things that are pushed up (if they are indeed signed).
With this setting on, you can _start_ taking advantage of Content Trust, but you don't have to: even with this on, you can still push up unsigned images.

The next step is Azure-dependent and requires that we grant those (Service) principal that will do signing, permissions to sign and push signed containers (https://docs.microsoft.com/en-us/azure/container-registry/container-registry-content-trust#grant-image-signing-permissions) into the registry.

To the best of my understanding at this point in time, those are the only levers of control we have on the "making signed images appear in the registry"-side. I don't think that at this point, there is a mechanism to limit which keys are allowed to sign images. As long as the key is valid (and we have one of those), and the principal doing the signing is allowed to push up signed images, then images can be signed and will appear in the registry.

In terms of validation of signatures, my understanding is that the onus is on the client to do that validation (setting `DOCKER_CONTENT_TRUST=1` when pulling enforces it and will fail on attempting to pull unsigned images or images for which the signature is -cryptographically- invalid). I think the best way to think of these signed images is more akin to this: the fact that the signed image appears in the repo is a flag that says "yes you can trust this because only a legit, validated user of this repo (which includes "knowing the repository key") could have pushed this signed image up; if they weren't legit, then they couldn't have pushed up the signed image because the notary wouldn't have allowed this".

operations/app/src/environments/02-config/main.tf
td-usds: > Does the Container Registry need configuration to accept only signed code from specific certificates? The way I'm reading this now, it sounds like any valid signed container would work and not only containers signed with the specified CDC key.

To the best of my understanding, this setting just enables the Notary service on this registry. It does not control who can push up images, nor what gets pushed up, or which keys are used to sign things that are pushed up (if they are indeed signed).
With this setting on, you can start taking advantage of Content Trust, but you don't have to: even with this on, you can still push up unsigned images.

The next step is Azure-dependent and requires that we grant those (Service) principal that will do signing, permissions to sign and push signed containers (https://docs.microsoft.com/en-us/azure/container-registry/container-registry-content-trust#grant-image-signing-permissions) into the registry.

To the best of my understanding at this point in time, those are the only levers of control we have on the "making signed images appear in the registry"-side. I don't think that at this point, there is a mechanism to limit which keys are allowed to sign images. As long as the key is valid (and we have one of those), and the principal doing the signing is allowed to push up signed images, then images can be signed and will appear in the registry.

In terms of validation of signatures, my understanding is that the onus is on the client to do that validation (setting DOCKER_CONTENT_TRUST=1 when pulling enforces it, and will fail on attempting to pull unsigned images or images for which the signature is -cryptographically- invalid). I think the best way to think of these signed images is more akin to this: the fact that the signed image appears in the repo is a flag that says "yes you can trust this because only a legit, validated user of this repo (which includes "knowing the repository key") could have pushed this signed image up; if they weren't legit, then they couldn't have pushed up the signed image because the notary wouldn't have allowed this".

ronaldheft-usds (approved):

PR #1576, opened 2021-07-12
Remove invalid/expired key from docker-compose

td-usds:
This PR removes an invalid and expired key from the development docker-compose file.

## Changes
- Comment out `SENDGRID_API_KEY` from `./prime-router/docker-compose.yml`; the value for this variable was invalid (not usable) anyway. We are leaving the environment variable in place (albeit commented out) to indicate that this is a value that can/should be set in running environments.

Note: there are some whitespace fixed included here as well (in accordance with editorconfig)


Bisonburger (approved):
lgtm

Issue #1416, opened 2021-06-26
Divide up the reports/ blobstore into new directories and subdirectories


jimduff-usds:
We'll stop using the reports/ directory, and we'll start using three new directories, `received/,` `internal/`, and `ready/`.  As follows:

1. The `receive` step will put the originally received report into `received/`, exactly byte for byte as the payload was received.  (Note:  If the payload was empty, I believe the code doesn't create an entry in the report file table, nor a empty blob in the Blob store. Please confirm this). So, the receive step should put exactly one file in the received/ folder per successful run.  Use a .csv or .hl7 suffix (or others in the future), as per the file type.   Do NOT store the 'internal' version of this file any more - there's no need for it.  (If you just do #1010, then you have done this step!)
2.  All other files created by the receive step should go into the internal/ folder, and should have an .internal.csv suffix.  (This is a change from current .internal, to make them easier to read when we pull them out for debugging)
3. The batch step creates files ready for download.  The batch step should put all its result files into the ready/ folder,in a subdirectory named after its full receiver name (not just the org).     For example, if the file is to go to receiver elr in organization vt-doh, create the directory ready/vt-doh.elr/

- In all cases, if any directory or subdirectory does not exist, create it.  (eg, the first time receive runs in prod, it will created the `received` directory)
- Test that the body_url in the report_file table reflects the new locations, and that it still "works" for files in the old reports/ location.
- Do not move or touch any files from the old reports/ location - that should continue to work.
- Test the download site carefully.   The new use of directories should be utterly hidden from the user.  Test that download of files in the old reports/ location also still works.
- For this ticket, don't change the filenames of internal/ or ready/*/ files.   I'll deal with that in a separate ticket.
- Test this in staging or in the datarobot staging, or get someone to test it explicitly in staging.
- I don't think any terraform changes are needed for this, but please confirm. (Eg, try greping for 'reports' in our .tf files)

Acceptance criteria:
1. Data received from a sender is stored in a new `received` folder in the blob store.  
2. Data received from a sender is stored with an appropriate extension based on the type of data (e.g. .csv or .hl7)
3. Data received from a sender is NOT stored as an internal file in the blob store.  See #1010
4. Intermediate steps that generate internal versions of the file need to the stored in a new `internal` folder in the blob store.
5. Internal files need to have the extension of .internal.csv, so they are easier to view/read.
[...]

Issue #1010, opened 2021-05-05
Fix storage of incoming ingested files


jimduff-usds:
When we ingest a file in the ReportFunction, we store a copy of it, in our blobstore.    The problem is that we store it in the wrong version.  The writing to the blob store of the internal version is done to write an action history.  The recommendation is to change the writing of the action history, so we write the raw data received from a sender and use that blob to write the action history.

We need to store the file exactly as it is sent to us, and instead we store it in our "INTERNAL" format.

This task is done when:
Instead of storing the INTERNAL format of the file, we store the file exactly as it is sent to us, byte for byte.  No reason to also store the INTERNAL formatted file.
Other features continue to work as before:  eg, the report is associated with the lineage, it shows up properly as a 'parent' in the report_lineage table, the 'blob_digest' is correct.

Acceptance criteria:
1. The raw received data is written to the blobstore with a proper file extension (e.g. CSV or HL7).
2. The internal version of the received data is no longer saved to the blob store.

jimduff-usds:
Hi Pathik, any progress on this?

pathikkothari:
I generally work on these tasks from Tues-Thurs unless it's urgent or have nothing else going on.  Not that far yet but should be able to complete it in a day or two.

pathikkothari:
Just an update......after making the changes...I was still seeing internal format being sent to the Azure Blob Storage (Azure Storage Explorer locally) and report file table in Postgres last week.  I believe I have made all the Internal format related changes so not sure where exactly it was coming from.  I will have to debug more.

jimduff-usds:
Pathik,

There are lots of steps that should and do create the .internal format.   _Only the very first creation of the received file needs to be modified_.    All others should remain as .internal.

Jim

jimduff-usds:
For example, if you submit a file with 10 rows and 5 go to NJ and 5 go to MD, then the receive step will create 3 files.

The 5 to NJ, and 5 to MD should both be .internal.    But we also want to store the 10-record original data, exactly byte for byte as it appeared in the payload.   That's the bug, right now we are storing an 'internal' representation of that data, instead of storing it exactly as it was sent to us.

jimduff-usds:
@tlosbo please check with Pathik, and take this issue over.  Thanks!

tlosbo:
@pathikkothari do you have a branch that I can view to see how far you got with this? I can take over from there.

pathikkothari:
[...]

PR #1584, opened 2021-07-13
1416 Add folders to organize files in Blob Store

carlosfelix2:
This PR adds two levels of folders for received, internal and send files as follows:
received/<org name>.<sender name>/file[.hl7|.csv]
internal/<org name>.<receiver name>/file.internal.csv
ready/<org name>.<receiver name>/file[.hl7|.csv]

Tested the UI changes by submitting data on both this branch and master's router to generate blobs with and without the folders, then ran the UI to download both files.

Test:
1. Use the `./gradlew testend2end` to send data and verify the test passes
2. Look in the blob store to verify the folders are generated as expected
3. Run the `./gradlew primeCLI --args="test --run compare"` and verify the test passes
4. Send the simplereport test data and login to http://localhost:7071/api/download and verify you can download the data files as expected.  E.g. `curl -X POST -H "client:simple_report"  -H "content-type:text/csv" --data-binary "@./src/test/csv_test_files/input/simplereport.csv"  "http://localhost:7071/api/reports"`

## Changes
- Added folders to the blob store
- Updated compare test files to pass due to changes in master
- Updated front end to remove blob folders from filename when present

### Testing
- [x] Tested locally?
- [x] Ran `quickTest all`?
- [ ] Added tests?

### Process

## Fixes
- #1010 
- #1416


frontend/src/js/prime-web-receiver-pre.js
```diff
@@ -184,7 +184,14 @@ function requestFile(reportId) {
 
     return window.jwt? axios.get(`${baseURL}/api/history/report/${reportId}`, apiConfig())
         .then(res => res.data)
-        .then(csv => download(csv.content, csv.filename, csv.mimetype)) : null;
+        .then(csv => {
```
td-usds, line 187: Does this mean that any call to `api/history/report/${reportId}` will return a path that cannot be downloaded just like that unless you 'fix' the path? If so, can we fix it server side instead to fix it for everyone? (unless I'm misunderstanding)

prime-router/src/test/kotlin/azure/ActionHistoryTests.kt
```diff
@@ -222,7 +208,7 @@ class ActionHistoryTests {
      * What I'd really like to do is confirm that two sql inserts were generated,
      * one to insert into ACTION and one to insert into REPORT_FILE.
      */
-//    @Test
+    @Test @Disabled
```
td-usds, line 211: Is this disabling intended? Why? Can you explain the reason for the disabling (maybe in the code with a `// TODO, #ISSUE_NUMBER` next to it for future re-enabling)

frontend/src/js/prime-web-receiver-pre.js
carlosfelix2: The API will return the blob name.  In the case of this code, the blob name will include the folder names as `ready%2Faz-phd.elr-prod%2Faz-covid-19-7edf57cf-a6d4-4a3f-98bb-300a2f9a0018-20210714141108.csv`.  I am worried about someone needing the blob name to do other operations, so not sure if removing the folders at the API level could hurt any future applications.

prime-router/src/test/kotlin/azure/ActionHistoryTests.kt
carlosfelix2: I added disabled because someone else commented out the @Test.  I cannot explain why it was done, but the proper way is to use @Disabled.

prime-router/src/main/kotlin/azure/ReportFunction.kt
```diff
@@ -97,7 +98,10 @@ class ReportFunction {
                     report = validatedRequest.report
                     routeReport(context, workflowEngine, validatedRequest, actionHistory)
                     val responseBody = createResponseBody(validatedRequest, actionHistory)
-                    workflowEngine.receiveReport(validatedRequest, actionHistory)
+                    workflowEngine.recordReceivedReport(
+                        report, request.body!!.toByteArray(), validatedRequest.sender!!,
```
MauriceReeves-usds, line 102: Rather than just use the not null assert operator here, I'd like you to check for null and then throw a meaningful error. I know that we verify that the body and sender are NOT null in validateRequest, however, using the not null assert just throws a `java.lang.NullPointerException` which is less clear and less readable than something that explains the error. This isn't a breaking review request, just something I feel is important. Any time you find yourself using `!!` in production code it should feel like code smell.

MauriceReeves-usds (commented):
Carlos, I'm fine with this going ahead, but there two concerns I have outlined if you want to tackle them first. Up to you.

prime-router/src/main/kotlin/azure/BlobAccess.kt
```diff
@@ -31,10 +31,41 @@ class BlobAccess(
...
+        reportName: String,
+        subfolderName: String? = null,
+        action: Event.EventAction = Event.EventAction.NONE
+    ): BlobInfo {
+        val subfolderNameChecked = if (subfolderName.isNullOrBlank()) "" else "$subfolderName/"
+        val blobName = when (action) {
+            Event.EventAction.RECEIVE -> "receive/$subfolderNameChecked$reportName"
+            Event.EventAction.SEND -> "ready/$subfolderNameChecked$reportName"
+            Event.EventAction.BATCH -> "batch/$subfolderNameChecked$reportName"
+            else -> error("Cannot determine folder to store blob.  Unsupported action $action for report $reportName")
```
MauriceReeves-usds, line 65: Carlos, this will only fix half the issue. We have receivers with the same file name format (Santa Clara County and California for example). If a file came in with results that would route to both, which is likely since Santa Clara County is in California, you'd still end up with a name collision when both files were written to these folders.

MauriceReeves-usds (commented):
Left one more comment that seemed to disappear. I'm not sure why.

frontend/src/js/prime-web-receiver-pre.js
td-usds: Ok, this make sense

prime-router/src/test/kotlin/azure/ActionHistoryTests.kt
```diff
@@ -222,7 +208,7 @@ class ActionHistoryTests {
...
     fun `test saveToDb with an externally received report`() {
```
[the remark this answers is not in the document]
carlosfelix2: I do not know.  Someone commented out the @Test.  That is the incorrect way to disable a test.  I readded @Test and added @Disable.  I do not know the history behind who or why this test was disabled.

prime-router/src/main/kotlin/azure/ReportFunction.kt
carlosfelix2: Makes sense.  It felt weird to check it again, but I'll add it.

prime-router/src/main/kotlin/azure/BlobAccess.kt
carlosfelix2: @mauricereeves I think this is resolved here.  Let's make sure.
If I understand correctly, you can have the same filename be sent to two different receivers.  Since the folder structure is now ready/<receiver>/filename then the collision will not happen.  Is this correct?  That subfolderName gets passed the sender or receiver name, so it is added as a folder level.

jimduff-usds: Yeah, I think we're ok as long as filenames are unique within a single receiver full name, since those are used as the subfolder names.

prime-router/src/main/kotlin/azure/BlobAccess.kt
```diff
@@ -72,19 +103,6 @@ class BlobAccess(
         return stream.toByteArray()
     }
 
-    /**
```
jimduff-usds, line 75: Copying is used for the HHSProtect.
I was leaving this code there, because its a better way to copy, but time for it to go, I guess.  So you know, I wasn't able to get this working, so abandoned it, and switched to the copyBlob below that pulls data into memory, then writes out again.   Less efficient, but it works.    We have an open ticket (I think) to fix this.

prime-router/src/main/kotlin/azure/ReportFunction.kt
```diff
@@ -96,8 +97,17 @@ class ReportFunction {
                     context.logger.info("Successfully reported: ${validatedRequest.report.id}.")
                     report = validatedRequest.report
                     routeReport(context, workflowEngine, validatedRequest, actionHistory)
+                    if (request.body != null && validatedRequest.sender != null) {
+                        workflowEngine.recordReceivedReport(
+                            report, request.body!!.toByteArray(), validatedRequest.sender!!,
+                            actionHistory, workflowEngine
+                        )
+                    } else error(
```
jimduff-usds, line 105: nice

jimduff-usds (approved):
lgtm - nice improvement - thanks!

PR #1628, opened 2021-07-16
frontend-react: adding containerization and abillity to run container

td-usds:
This PR adds a Dockerfile for the react-frontend and a mechanism to run the container locally (`run.container.sh`). This will make it easy for folks to be able to run this on their own box as well.


MauriceReeves-usds (approved):
@td-usds your review looks fine, but you say that this is going into one of @Bisonburger's branches, but the PR itself is set to merge into master. I don't see anything here that would be problematic going into Master, but I wanted to call that out in case it wasn't intentional.

td-usds:
> @td-usds your review looks fine, but you say that this is going into one of @Bisonburger's branches, but the PR itself is set to merge into master. I don't see anything here that would be problematic going into Master, but I wanted to call that out in case it wasn't intentional.

I think the old branch that this was slated to go into got merged and then GitHub (or 'base') automatically changed the PR to go into `master`, see:
`Base automatically changed from mry/frontend-react to master 3 days ago `

This is fine going into master from my POV; @mry-usds or @Bisonburger any concerns?

Bisonburger (approved):
lgtm

PR #1639, opened 2021-07-19
Fix quicktest to no longer use the build.sh script

td-usds:
This PR reverts the invocation of gradlew in quick-test.sh instead of using build.sh


jimduff-usds (approved):
lgtm

PR #1668, opened 2021-07-20
Terraform GitHub Action

ronaldheft-usds:
This PR lays the groundwork for GitHub Actions to manage Terraform. I have started with a relatively low-risk Terraform workflow that focuses on validating our Terraform configuration.

## Changes
- Adds a new Terraform workflow with two different jobs:
    - Validate Terraform YAML
        - Performs a `terraform validate` on every module
        - This is environment-agnostic and is validating syntax and resource usage
    - Generate Terraform Plan for Staging
        - Performs a `terraform plan` on every module against staging
        - This logs into Azure and connects to the VPN to perform the action
        - After generating the plan, the `.tfplan` and a plain text version of the plan are uploaded to an Azure Storage Account
            - An Azure Storage account was chosen instead of using a GitHub artifact since the plan may contain sensitive information
            - While GitHub Action artifacts require user authentication to download, you do not have to be a contributor to the repository to download build artifacts (this was tested and confirmed)
            - Using an Azure Storage Account is private and a safe location to store the plan, since it’s marked as private, requires a connection over the VPN, and an authorized Active Directory user
        - Once the plan is uploaded, a command is outputted that will download the plan
            -  This is to make it easy for the reviewer to download the plan on their location machine
- Adds two new targets to the operations `Makefile`:
    - `tf-validate`
        - Runs `terraform validate` within the container
    - `tf-plan`
        - Run `terraform plan-file` within the container
        - See below for more information on `plan-file` command
- Enhances the `tf` script to support new Terraform actions
    - Will not initialize the backend if the command does not require it
    - Adds a custom `plan-file` command that outputs the plan as a file and redirects output into a plain text file
- Grants the Terraform GitHub service principal access to read secrets
    - This was the only permission it was missing for the ability to generate plans
- Renames GitHub Actions for consistency
- Adds a new GitHub Action for building the tfcli container
- Adds a new GitHub Action for publishing the tfcli container on merge
  - This has been tested with this run: https://github.com/CDCgov/prime-reportstream/pull/1668/checks?check_run_id=3146748587
  - Package available here: https://github.com/CDCgov/prime-reportstream/pkgs/container/prime-reportstream_tfcli
- Changes the docker-compose.yml to use the published container, if available
- Changes the tfcli version references to be arguments instead of hard-coded in the Dockerfile

## Downside to This Approach

Since our Terraform configuration is broken into stages, any radical changes to preceding stages will likely cause later stages to fail.

For example, if we add a storage account in `03-persistent`, if we require that storage account in `04-app`, the storage account will not exist and running a `tf plan` on `04-app` will fail, thus triggering a fail of the GitHub Action. This is because these GitHub Actions are only generating the plans and not applying them.

I believe this is acceptable behavior, as we can work around this. If we submit the change in multiple PRs, i.e. submit a PR for the changes in `03-persistent` and then submit a PR for the changes in `04-app`, once we merge the PR for `03-persistent` the PR for `04-app` will succeed. This will push us to smaller PRs and also help us think through non-breaking changes when making changes across stages.

### Testing
- [x] Tested locally?
- [ ] Ran `quickTest all`?
- [ ] Added tests?

    - No
    - No
    - No
    - Yes, additional permissions are granted to the GitHub Action Azure service principal
- **Does logging contain sensitive data?**
    - Yes
    - Terraform plans may contain secret values or other sensitive information
    - This information has been excluded from the logs and is routed to a file, which is then stored in a private Azure Storage Account (the same one our Terraform state exists in)
    - A potential way information could leak is if there is a flaw in the way the output is routed to a file
        - I’m wondering if all Terraform output should be suppressed since log files are public
        - This would account for a scenario where a bug in Terraform exists or an unknown error path is trigger and sensitive information is logged
    - **A reviewer should focus in on any ways my approach may fail**
- **Does this PR include or remove any sensitive information itself?**
    - Yes
    - Outside of the logs, the `.tfplan` itself is sensitive
    - Like the logs, this is stored in a Storage Account and never exposed via GitHub artifacts

    - See above

### Process
- [x] Database changes are submitted as a separate PR?

## To Be Done
* Once merged, I would like to mark the Terraform workflow as required for merging a PR


.github/workflows/build_hub.yml
```diff
@@ -1,4 +1,4 @@
-name: Build router
+name: Router
```
ronaldheft-usds, line 1: Renamed so their consistency between Actions

.github/workflows/validate_terraform.yml
```diff
@@ -0,0 +1,103 @@
...
+              - 'operations/Dockerfile'
+              - 'operations/Makefile'
+              - '.github/workflows/validate_terraform.yml'
+      - id: fork_check
+        run: echo "::set-output name=is_fork::${{ github.event.pull_request.head.repo.full_name != github.repository }}"
+
+  validate_terraform_yaml:
+    name: Validate Terraform YAML
+    needs: pre_job
+    if: ${{ needs.pre_job.outputs.has_operations_change == 'true' && needs.pre_job.outputs.is_fork != 'true' }}
```
ronaldheft-usds, line 34: Only run if the terraform code has changed

.github/workflows/validate_terraform.yml
```diff
@@ -0,0 +1,103 @@
...
+          TLS_KEY: ${{ secrets.TLS_KEY }}
+        env:
+          CA_CRT: ${{ secrets.CA_CRT}}
+          USER_CRT: ${{ secrets.USER_CRT }}
+          USER_KEY: ${{ secrets.USER_KEY }}
+
+      - name: Generate a Plan for All Terraform Modules
+        working-directory: operations
+        env:
+          SERVICE_PRINCIPAL_CREDS: ${{ secrets.SERVICE_PRINCIPAL_CREDS }}
```
ronaldheft-usds, line 83: Since this is a GitHub secret, the output is never placed in the build log and will show `***` instead

.github/workflows/validate_terraform.yml
```diff
@@ -0,0 +1,103 @@
...
+          make TF_ENV=staging AZ_SERVICE_PRINCIPAL="$SERVICE_PRINCIPAL_CREDS" tf-plan
+
+      - name: Upload Terraform Plan to Storage Account
+        if: ${{ always() }} # We want to upload whatever part of the plan we have
+        uses: azure/CLI@v1
+        with:
+          inlineScript: |
+            az storage blob upload-batch --account-name pdhstagingterraform \
+              --destination terraformplan \
+              --destination-path "${{ github.run_id }}-${{ github.sha }}" \
```
ronaldheft-usds, line 94: The `run_id` is used first so the directories can be sorted in order.

operations/Makefile
```diff
@@ -50,9 +80,32 @@ tf-03-persistent: ## Preconfigures Terraform persistent environment module
...
 
 # AZURE
 
+# Internal target: set azure service principal values
+.az-configure-service-principal:
+ifdef AZ_SERVICE_PRINCIPAL
+ARM_CLIENT_ID := $(shell echo '$(AZ_SERVICE_PRINCIPAL)' | python3 -c "import sys, json; print(json.load(sys.stdin)['clientId'])")
+ARM_CLIENT_SECRET := $(shell echo '$(AZ_SERVICE_PRINCIPAL)' | python3 -c "import sys, json; print(json.load(sys.stdin)['clientSecret'])")
+ARM_SUBSCRIPTION_ID := $(shell echo '$(AZ_SERVICE_PRINCIPAL)' | python3 -c "import sys, json; print(json.load(sys.stdin)['subscriptionId'])")
+ARM_TENANT_ID := $(shell echo '$(AZ_SERVICE_PRINCIPAL)' | python3 -c "import sys, json; print(json.load(sys.stdin)['tenantId'])")
```
ronaldheft-usds, line 104: Python was chosen to parse the JSON since it's already installed on most machines. Using the dependency `jq` would be cleaner, but I didn't want to force a new dependency.

operations/app/src/modules/key_vault/main.tf
```diff
@@ -17,6 +17,7 @@ locals {
   ]
 
   frontdoor_object_id = "270e4d1a-12bd-4564-8a4b-c9de1bbdbe95"
+  terraform_object_id = "4d81288c-27a3-4df8-b776-c9da8e688bc7"
```
ronaldheft-usds, line 20: This is not a secret. This is just the UUID of the resource.

operations/app/src/environments/tf
```diff
@@ -165,6 +183,18 @@ if [[ "${ACTUAL_RESULTS?}" == "${EXPECTED_RESULTS?}" ]]; then
...
+        # Custom command we made for the purpose of CI/CD
+        EXPECTED_RESULTS="${EXPECTED_RESULTS?}0"
+        terraform plan \
+            --var-file "${TFVARS?}" \
+            >/dev/null -out=${PLAN_DIR?}/${ENVIRONMENT?}-${HERE?}.tfplan
+        terraform show -no-color \
+            ${PLAN_DIR?}/${ENVIRONMENT?}-${HERE?}.tfplan \
+            > ${PLAN_DIR?}/${ENVIRONMENT?}-${HERE?}.txt
+        ACTUAL_RESULTS="${ACTUAL_RESULTS?}${?}"
+        ;;
```
ronaldheft-usds, line 196: This is the block of code I am most concerned with from a security perspective. I want to ensure that under all scenarios, Terraform will never log sensitive information to the console.

.github/workflows/build_hub.yml
td-usds: Agreed

td-usds:
I know this is outside of what this PR touches but I'll ask for it anyway:
can you fix the example entry that does `@echo " make TF_ENV=\"test\" all"` since we don't have an `all` taget

operations/app/src/environments/tf
```diff
@@ -147,8 +165,8 @@ if [[ "${ACTUAL_RESULTS?}" == "${EXPECTED_RESULTS?}" ]]; then
         echo -e "${WHITE?}Executing terraform ${TFCMD?}${NORMAL?}..."
     fi
     case "${TFCMD?}" in
-    graph|output|state*)
-        # Graph & output are special in that they do not like arguments
+    graph|output|validate|state*)
```
td-usds, line 168: why the trailing `*`? I could now to `./tf --cmd statefoo --cmd test` and it would drop into this branch...
same for `import`, `plan-file` (but _not_ for `*)`...)

operations/app/src/environments/tf
```diff
@@ -165,6 +183,18 @@ if [[ "${ACTUAL_RESULTS?}" == "${EXPECTED_RESULTS?}" ]]; then
         ACTUAL_RESULTS="${ACTUAL_RESULTS?}${?}"
         ;;
 
+    plan-file*)
```
td-usds, line 186: why `plan-file` instead of `plan`? I would stick to the tf commands as close as possible; it is just a wrapper after all

operations/app/src/environments/tf
```diff
@@ -165,6 +183,18 @@ if [[ "${ACTUAL_RESULTS?}" == "${EXPECTED_RESULTS?}" ]]; then
...
+        # Custom command we made for the purpose of CI/CD
+        EXPECTED_RESULTS="${EXPECTED_RESULTS?}0"
+        terraform plan \
+            --var-file "${TFVARS?}" \
+            >/dev/null -out=${PLAN_DIR?}/${ENVIRONMENT?}-${HERE?}.tfplan
```
td-usds, line 191: I did not know that you could redirect stdout in the middle of the command (so TIL :) ); I would however prefer it at the end (and not chop up the command in the middle)

operations/app/src/environments/tf
td-usds: I do know that terraform will block 'sensitive' values but I don't know of the top of my head if there are edge cases where they might leak.
I'm with you

operations/app/src/modules/key_vault/main.tf
```diff
@@ -101,6 +102,15 @@ resource "azurerm_key_vault_access_policy" "frontdoor_access_policy" {
   certificate_permissions = [ "Get" ]
 }
 
+resource "azurerm_key_vault_access_policy" "terraform_access_policy" {
+    key_vault_id = azurerm_key_vault.application.id
+    tenant_id = data.azurerm_client_config.current.tenant_id
+    object_id = local.terraform_object_id
+
+    secret_permissions = [ "Get" ]
+    key_permissions = [ "Get" ]
```
td-usds, line 111: Why does terraform get `key_permissions = ["Get"]` on this vault but not on the app_config vault?

operations/Makefile
```diff
@@ -1,19 +1,26 @@
 SHELL := bash
 .SHELLFLAGS := -eu -o pipefail -c
 
+# Pretty colors
+COLOR_PURPLE = \033[1;35m
```
td-usds, line 5: a vote to change this to `\033[1;44m` (which will make this show up as bold white on an ... ahum ... Azure background
we can call it `COLOR_AZURE`

operations/Makefile
```diff
@@ -50,9 +80,32 @@ tf-03-persistent: ## Preconfigures Terraform persistent environment module
 tf-04-app: ## Preconfigures Terraform app environment module
 	${MAKE} tf-cli TF_STAGE="04-app"
 
+tf-validate: ## Validates the Terraform configuration across all modules
```
td-usds, line 83: This combined with the "don't use a backend" is great! Now anyone (even without Azure creds) can run validate super easily (otherwise it was ... less frictionless)

operations/app/src/environments/tf
```diff
@@ -17,6 +17,17 @@ VERBOSE=${VERBOSE:-0}
 # Where is this script located (and thus, where are the configuration files)?
 SCRIPT_DIR="$(dirname "$(realpath "${0}")")"
 CONFIGS_DIR="${SCRIPT_DIR?}/configurations"
+PLAN_DIR="${SCRIPT_DIR?}/../../plan"
+
+# Does the command require an Azure backend?
+case "${TFCMD?}" in
+validate*)
```
td-usds, line 24: `graph` should be here too
also; remove the trailing `*`

operations/app/src/environments/tf
```diff
@@ -17,6 +17,17 @@ VERBOSE=${VERBOSE:-0}
...
+    AZ_BACKEND_REQUIRED=0
+    ;;
+*)
```
td-usds, line 27: keep this `*`

td-usds:
I like this thoughtfulness:
"After generating the plan, the .tfplan and a plain text version of the plan are uploaded to an Azure Storage Account

    An Azure Storage account was chosen instead of using a GitHub artifact since the plan may contain sensitive information
    While GitHub Action artifacts require user authentication to download, you do not have to be a contributor to the repository to download build artifacts (this was tested and confirmed)
    Using an Azure Storage Account is private and a safe location to store the plan, since it’s marked as private, requires a connection over the VPN, and an authorized Active Directory user
"

We also already use the azure backend for our tfstate anyway, so this is fine

td-usds:
Since the make command involves running a container and this container is unlikely to change (cf. [`prime-router_builder`](https://github.com/orgs/CDCgov/packages/container/package/prime-router_builder)); should we also just publish this as a pre-build image into the packages?

operations/app/src/environments/tf
td-usds: Consider explicitly redirecting both stdout and stderr:
`1>/dev/null 2>/dev/null`
or maybe dump it in a file that is uploaded into Azure storage?

td-usds (changes requested):

.github/workflows/validate_terraform.yml
```diff
@@ -0,0 +1,103 @@
...
+      - master
+      - production
+
+jobs:
+  pre_job:
+    name: Pre Job
+    runs-on: ubuntu-latest
+    outputs:
+      has_operations_change: ${{ steps.skip_check.outputs.operations }}
+      is_fork: ${{ steps.fork_check.outputs.is_fork }}
```
ronaldheft-usds, line 15: Remove this check. We want the build to fail on a fork if anyone makes changes to any of these files. The Action will not normally run unless these files are modified, so this should not affect our normal application PR workflow. This will serve as a precaution and an extra check to ensure outside contributors are not making changes to the infrastructure.

.github/workflows/validate_terraform.yml
```diff
@@ -0,0 +1,103 @@
...
+    defaults:
+      run:
+        working-directory: operations
+    steps:
+      - name: Check Out Changes
+        uses: actions/checkout@v2
+
+      - name: Validate All Terraform Modules
+        run: |
+          make TF_ENV=staging tf-validate
```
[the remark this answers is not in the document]
ronaldheft-usds: Yes, we should.

.github/workflows/validate_terraform.yml
```diff
@@ -0,0 +1,103 @@
...
+          CA_CRT: ${{ secrets.CA_CRT}}
+          USER_CRT: ${{ secrets.USER_CRT }}
+          USER_KEY: ${{ secrets.USER_KEY }}
+
+      - name: Generate a Plan for All Terraform Modules
+        working-directory: operations
+        env:
+          SERVICE_PRINCIPAL_CREDS: ${{ secrets.SERVICE_PRINCIPAL_CREDS }}
+        run: |
+          make TF_ENV=staging AZ_SERVICE_PRINCIPAL="$SERVICE_PRINCIPAL_CREDS" tf-plan
```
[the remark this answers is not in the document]
ronaldheft-usds: Agreed.

.github/workflows/validate_terraform.yml
```diff
@@ -0,0 +1,103 @@
...
+      - name: Upload Terraform Plan to Storage Account
+        if: ${{ always() }} # We want to upload whatever part of the plan we have
+        uses: azure/CLI@v1
+        with:
+          inlineScript: |
+            az storage blob upload-batch --account-name pdhstagingterraform \
+              --destination terraformplan \
+              --destination-path "${{ github.run_id }}-${{ github.sha }}" \
+              --source operations/app/plan \
+              --pattern staging-*
```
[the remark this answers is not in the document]
ronaldheft-usds: I'll add single quotes here.

operations/Makefile
```diff
@@ -1,19 +1,26 @@
...
+COLOR_NONE = \033[0m
+
 TF_ENV = "dev" #VAR Your target environment (e.g. dev, test, staging, prod)
 TF_STAGE = ""  # not surfaced
+TF_CMD = ""  # not surfaced
 
 AZ_RESOURCE_PREFIX = "pdh$(TF_ENV)" ## VAR Azure environment resource prefix (e.g. pdhtest)
 AZ_RESOURCE_GROUP = "prime-data-hub-$(TF_ENV)" ## VAR Azure resource group (e.g. prime-data-hub-staging)
 
+AZ_SERVICE_PRINCIPAL =  #VAR JSON string containing service credential values provided by Azure. Be careful with escaping quotes when using this variable
```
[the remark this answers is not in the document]
ronaldheft-usds: Yes, I can make both of those changes.

operations/app/src/environments/tf
ronaldheft-usds: Agree, remove trailing `*` and add the `graph`.

operations/app/src/environments/tf
ronaldheft-usds: Will add a `1>.stdout` and `2>.stderr` as an output.

operations/app/src/environments/tf
ronaldheft-usds: Also capture the error as well.

.github/workflows/validate_terraform.yml
ronaldheft-usds: Done in next push.

.github/workflows/validate_terraform.yml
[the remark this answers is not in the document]
ronaldheft-usds: Fixed in next push

.github/workflows/validate_terraform.yml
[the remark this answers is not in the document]
ronaldheft-usds: Fixed in next push

operations/Makefile
ronaldheft-usds: Fixed in next push

operations/Makefile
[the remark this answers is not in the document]
ronaldheft-usds: Fixed in next push

operations/app/src/environments/tf
ronaldheft-usds: Fixed in next push

operations/app/src/environments/tf
ronaldheft-usds: Fixed in next push

operations/app/src/environments/tf
ronaldheft-usds: `plan-file` runs a number of commands in addition to `plan`

operations/app/src/environments/tf
ronaldheft-usds: Fixed in next push

operations/app/src/environments/tf
ronaldheft-usds: Fixed in next push

operations/app/src/modules/key_vault/main.tf
ronaldheft-usds: Key permission is used to reference the HSM key for the Storage Account and Postgres database

.github/workflows/build_docker.yaml
```diff
@@ -0,0 +1,40 @@
...
+        id: skip_check
+        with:
+          list-files: csv
+          filters: |
+            tfcli:
+              - 'operations/docker-compose.yml'
+              - 'operations/Dockerfile'
+              - '.github/workflows/build_docker.yml'
+
+  build_tfcli:
```
ronaldheft-usds, line 26: This job may be redundant, because technically Docker Compose will build the container if doesn't exist in the registry, but I'd rather have it explicitly defined as an Action for clarity.

.github/workflows/build_docker.yaml
```diff
@@ -0,0 +1,40 @@
+name: Docker
```
ronaldheft-usds, line 1: Figure this Action can have other jobs added on for other containers, like `prime-router_builder`.

.github/workflows/publish_docker.yaml
```diff
@@ -0,0 +1,58 @@
+name: Publish Container to GitHub
```
ronaldheft-usds, line 1: Figure this Action can have other jobs added on for other containers, like `prime-router_builder`.

.github/workflows/publish_docker.yaml
```diff
@@ -0,0 +1,58 @@
...
+    if: ${{ needs.pre_job.outputs.has_tfcli_change == 'true' }}
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: operations
+    env:
+      IMAGE_NAME: cdcgov/prime-reportstream_tfcli
+    permissions:
+      contents: read
+      packages: write
```
ronaldheft-usds, line 40: Magic! Settings these values ensures the `GITHUB_TOKEN` has the correct permissions.

.github/workflows/publish_docker.yaml
```diff
@@ -0,0 +1,58 @@
...
+    steps:
+      - name: Check Out Changes
+        uses: actions/checkout@v2
+
+      - name: Log In to the Container Registry
+        uses: docker/login-action@f054a8b539a109f9f41c372932f1ae047eff08c9
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
```
ronaldheft-usds, line 50: Recommended in https://docs.github.com/en/actions/guides/publishing-docker-images

.github/workflows/publish_docker.yaml
```diff
@@ -0,0 +1,58 @@
...
+
+      - name: Build Docker Terraform CLI
+        run: |
+          make build-tf
+
+      - name: Push to the Container Registry
+        run: |
+          docker push ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }} --all-tags
```
ronaldheft-usds, line 58: Using `--all-tags` so we don't have to parse out what the version tag is.

.github/workflows/validate_terraform.yml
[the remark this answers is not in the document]
ronaldheft-usds: Done! See the new GitHub Actions that go along with this.

operations/docker-compose.yml
```diff
@@ -3,8 +3,12 @@ version: "3.3"
 services:
   tf-cli:
     container_name: terraform_state
+    image: ghcr.io/cdcgov/prime-reportstream_tfcli:az2.18.0tf0.14.5
     build:
       context: .
+      args:
+        AZURE_CLI_VERSION: 2.18.0
+        TERRAFORM_VERSION: 0.14.5
```
ronaldheft-usds, line 11: Parameterized the build. Any ideas on how to ensure we only have to update the version numbers in one spot, so the images doesn't also have to be changed?

ronaldheft-usds:
All changes made! Great suggestions @td-usds.

ronaldheft-usds:
@jdorothy

.github/workflows/build_docker.yaml
td-usds: That is a brilliant idea, YES!

.github/workflows/build_docker.yaml
td-usds: agreed

.github/workflows/publish_docker.yaml
td-usds: see above, absolutely yes!

operations/docker-compose.yml
td-usds: I was hunting for this since I saw the change `ENV` -> `ARG` above :)
I don't know of a clean way (right now) to unify these - to make sure I understand, it's to avoid having to change the image and the args (both in these files), right?

td-usds (approved):

operations/docker-compose.yml
ronaldheft-usds: Correct, to avoid changing it in two places, eliminating the possibility that we mistag an image with the incorrect version.

Issue #1678, opened 2021-07-21
Files uploaded to SFTP receivers contain the blob folders in the filename


carlosfelix2:
The files being uploaded to the SFTP locations are of the form `ready%2Fignore.HL7%2Ffl-covid-19-38c3d116-e12c-4c19-bc9a-f6095750f42a-20210721151708.hl7`.  Those filenames cannot have the blob folders as part of the filename.

Acceptance criteria:
1. The filename of the data uploaded to SFTP does not contain the blob folders.
2. Identify what other code may be affected by this blob folder issue.

PR #1681, opened 2021-07-21
1678 Fixed SFTP filenames to not have blob folders

carlosfelix2:
This PR fixes the filenames uploaded via SFTP to not have the blob folders.  I searched all the code for any other places that the blob URL was being used besides SFTP and found the BlobTransport was affected too.

To test:
1. Build and run the API.
2. Run `./gradlew testend2end` and verify in the build/sftp folder that the filenames for the received data do not contain the blob folders.

## Changes
- Fixed SFTP so filenames do not have the blob folders
- Updated blob copy method to use proper filename.  This affects the BlobTransport.

### Testing
- [x] Tested locally?
- [x] Ran `quickTest all`?
- [x] Added tests?

### Process

## Fixes
- #1678

### DIFF
--- a/prime-router/src/main/kotlin/Report.kt
+++ b/prime-router/src/main/kotlin/Report.kt
@@ -1,5 +1,6 @@
 package gov.cdc.prime.router
 
+import gov.cdc.prime.router.azure.BlobAccess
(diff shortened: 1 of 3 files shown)


MauriceReeves-usds (approved):
lgtm

PR #1687, opened 2021-07-22
1686 Only publish test results when in our repo context

carlosfelix2:
This PR disables the publishing of test results for PRs that are not within the context of this repo (e.g. dependabot).  The issue is [described in this article](https://github.blog/changelog/2021-02-19-github-actions-workflows-triggered-by-dependabot-prs-will-run-with-read-only-permissions/) and it results in dependabot not being able to publish the test results for the build to the PR and breaking the build.  This change implements the [recommended check as described here](https://github.com/EnricoMi/publish-unit-test-result-action#support-fork-repositories-and-dependabot-branches) since the publishing is really an optional thing.

Test: 
1. Verify the build completes successfully and that the unit and integration tests are published.
Note that we will need to merge this change in to test this in a dependabot PR

## Changes
- Builds from dependabot and other users not running on the context of this repo do not publish test results

### Testing
- [ ] Tested locally?
- [ ] Ran `quickTest all`?
- [ ] Added tests?

### Process

### DIFF
--- a/.github/workflows/build_hub.yml
+++ b/.github/workflows/build_hub.yml
@@ -68,9 +68,14 @@ jobs:
 
       - name: Build Prime Router Package
         run: bash ./build.sh -- gradle clean package -x fatjar -Pshowtests
+        
       - name: Publish Unit Test Results
         uses: EnricoMi/publish-unit-test-result-action@v1
-        if: always()
+        # Per https://github.com/EnricoMi/publish-unit-test-result-action#support-fork-repositories-and-dependabot-branches
+        if: >
+          always() &&
+          github.event.sender.login != 'dependabot[bot]' &&
+          ( github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository )
         with:
           # This path is from the root of the repo as needed by the plugin
           files: prime-router/build/test-results/test/**/*.xml
@@ -80,7 +85,11 @@ jobs:
 
       - name: Publish Integration Test Results
         uses: EnricoMi/publish-unit-test-result-action@v1
-        if: always()
+        # Per https://github.com/EnricoMi/publish-unit-test-result-action#support-fork-repositories-and-dependabot-branches
+        if: >
+          always() &&
+          github.event.sender.login != 'dependabot[bot]' &&
+          ( github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository )
         with:
           # This path is from the root of the repo as needed by the plugin
           files: prime-router/build/test-results/testIntegration/**/*.xml
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -37,7 +37,11 @@ jobs:
 
       - name: Publish Unit Test Results
         uses: EnricoMi/publish-unit-test-result-action@v1
-        if: always()
+        # Per https://github.com/EnricoMi/publish-unit-test-result-action#support-fork-repositories-and-dependabot-branches
+        if: >
+          always() &&
+          github.event.sender.login != 'dependabot[bot]' &&
+          ( github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository )
         with:
           # This path is from the root of the repo as needed by the plugin
           files: prime-router/build/test-results/test/**/*.xml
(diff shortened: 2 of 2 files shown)


.github/workflows/build_hub.yml
```diff
@@ -68,9 +68,14 @@ jobs:
 
       - name: Build Prime Router Package
         run: bash ./build.sh -- gradle clean package -x fatjar -Pshowtests
+        
       - name: Publish Unit Test Results
         uses: EnricoMi/publish-unit-test-result-action@v1
-        if: always()
+        # Per https://github.com/EnricoMi/publish-unit-test-result-action#support-fork-repositories-and-dependabot-branches
+        if: >
+          always() &&
```
td-usds, line 76: remove the `always()` since it's used in `&&` and it always returns `true` anyway
see https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#always

Not quite sure why the referenced link instructs you to include the `always()` call (which it does)

.github/workflows/build_hub.yml
```diff
@@ -80,7 +85,11 @@ jobs:
 
       - name: Publish Integration Test Results
         uses: EnricoMi/publish-unit-test-result-action@v1
-        if: always()
+        # Per https://github.com/EnricoMi/publish-unit-test-result-action#support-fork-repositories-and-dependabot-branches
+        if: >
+          always() &&
```
td-usds, line 90: remove the `always()` since it's used in `&&` and it always returns `true` anyway
see https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#always

td-usds (changes requested):
For completeness, these changes should also be reflected in
- https://github.com/CDCgov/prime-reportstream/blob/07098b3a12c4833def2f6454f2c839c0f5142077/.github/workflows/release.yml#L38
- https://github.com/CDCgov/prime-reportstream/blob/07098b3a12c4833def2f6454f2c839c0f5142077/.github/workflows/release.yml#L48

td-usds:
My understanding is that _not_ having an `if:`, implies an `if: success()`; and wanting to always execute something thus means you have to override that implied condition by listing `if: always()`; maybe that is where this comes from?

ronaldheft-usds:
> My understanding is that _not_ having an `if:`, implies an `if: success()`; and wanting to always execute something thus means you have to override that implied condition by listing `if: always()`; maybe that is where this comes from in the linked-to url?
> 
> But if you have an `if:`, then that overrides the implied condition and you should be good...

This is correct on the default being `if: success()`. Shouldn't those actions always run even if there is a failure, so we can get the results of a failure?

td-usds:
> > My understanding is that _not_ having an `if:`, implies an `if: success()`; and wanting to always execute something thus means you have to override that implied condition by listing `if: always()`; maybe that is where this comes from in the linked-to url?
> > But if you have an `if:`, then that overrides the implied condition and you should be good...
> 
> This is correct on the default being `if: success()`. Shouldn't those actions always run even if there is a failure, so we can get the results of a failure?

They should, but the `always() &&` doesn't do anything for that as it evaluates to `true &&`; so we should be able to omit it

carlosfelix2:
> For completeness, these changes should also be reflected in
> 
> * https://github.com/CDCgov/prime-reportstream/blob/07098b3a12c4833def2f6454f2c839c0f5142077/.github/workflows/release.yml#L38
> * https://github.com/CDCgov/prime-reportstream/blob/07098b3a12c4833def2f6454f2c839c0f5142077/.github/workflows/release.yml#L48

I do not think this is needed as the master build will always be in kicked off by the repo itself.

.github/workflows/build_hub.yml
carlosfelix2: I agree the always is needed, but I do want to keep the code as references in the instructions.  The always does not hurt.

.github/workflows/build_hub.yml
carlosfelix2: I agree the always is needed, but I do want to keep the code as references in the instructions. The always does not hurt.

.github/workflows/build_hub.yml
td-usds: is _not_ needed, you mean?

carlosfelix2: In Slack combo we agreed that it was ok to leave the always.

carlosfelix2:
> For completeness, these changes should also be reflected in
> 
> * https://github.com/CDCgov/prime-reportstream/blob/07098b3a12c4833def2f6454f2c839c0f5142077/.github/workflows/release.yml#L38
> * https://github.com/CDCgov/prime-reportstream/blob/07098b3a12c4833def2f6454f2c839c0f5142077/.github/workflows/release.yml#L48

In Slack combo it was agree to add it to the release.yml just to make the code the same.

MauriceReeves-usds (approved):
Personally I'm ambivalent because I always rerun the build one more time as part of my workflow when checking dependabot bumps, so it failing the first time is a non-issue for me. Otherwise, looks good to me.

carlosfelix2:
> Personally I'm ambivalent because I always rerun the build one more time as part of my workflow when checking dependabot bumps, so it failing the first time is a non-issue for me. Otherwise, looks good to me.

It is actually failing all the time for me even after doing a rebase.

td-usds (approved):
Approved _provided_ you open a tracking bug (tagged with "Good first bug" or equivalent) for someone to address the `always()` and `github.event_name != 'pull_request' ||` parts since those are effectively no-ops in this context)

PR #1692, opened 2021-07-22
Enable CI for Forks

ronaldheft-usds:
This PR allows our CI GitHub Action to run for forks.

Originally we had disabled this because our CI build required the use of some secrets, but those limitations have since been removed.

### DIFF
--- a/.github/workflows/build_hub.yml
+++ b/.github/workflows/build_hub.yml
@@ -48,6 +48,16 @@ jobs:
         run: |
           echo >> $GITHUB_ENV DB_USER=${POSTGRES_USER}
           echo >> $GITHUB_ENV DB_PASSWORD=${POSTGRES_PASSWORD}
+          if [ "$GITHUB_BASE_REF" == "production" ]
+          then
+              echo "Building for the production environment."
+              echo >> $GITHUB_ENV ACR_REPO=pdhprodcontainerregistry.azurecr.io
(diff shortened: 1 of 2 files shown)


carlosfelix2 (approved):

.github/workflows/build_hub.yml
```diff
@@ -51,16 +48,6 @@ jobs:
...
-          if [ "$GITHUB_BASE_REF" == "production" ]
-          then
-              echo "Building for the production environment."
-              echo >> $GITHUB_ENV ACR_REPO=pdhprodcontainerregistry.azurecr.io
-              echo >> $GITHUB_ENV PREFIX=pdhprod
-          else
-              echo "Building for the test environment."
-              echo >> $GITHUB_ENV ACR_REPO=pdhstagingcontainerregistry.azurecr.io
-              echo >> $GITHUB_ENV PREFIX=pdhstaging
-          fi
```
ronaldheft-usds, line 63: There are old unused environment variables. Before we had our Azure service principal defined here as well because the CI job used to also do CD instead of being a separate workflow

td-usds (approved):

.github/workflows/build_hub.yml
ronaldheft-usds: This was wrong. They're still needed, but they're not secrets. Opted to keep them there for now and we can address in a future PR.

PR #1702, opened 2021-07-23
signing commits: proper indentation and adding export of GPG_TTY env var

td-usds:
This PR includes the instruction for folks on mac that they may have to `export GPG_TTY=$(tty)` in their `.bash_profile`
This PR also indents the code blocks in that file and fixes the numbered list sequence


carlosfelix2:
There is plenty of documentation online for this.  Why add this document instead of a link to an already existing doc?

td-usds:
> There is plenty of documentation online for this. Why add this document instead of a link to an already existing doc?

Because it was something that @jimduff-usds posted to the slack channel and seems like something that might be easily overlooked by others. Just trying to eliminate friction. We are already telling them how to set up signing anyway...

ronaldheft-usds (approved):

Issue #1704, opened 2021-07-23
Compare CSV tests need to handle new schema fields without failing


carlosfelix2:
The compare CSV test currently checks that all columns are the same in both actual and expected.  If a new field is added to the schema then the test fails as the new column in the actual data does not match the expected column in the expected data file since that file was created for the old schema.

Acceptance criteria:
1. Adding a new field to a schema does not break existing tests that compare CSV data.
2. A warning needs to be shown for any actual data that has not expected data.

PR #1705, opened 2021-07-23
1704 Compare CSV does not rely on expected data column order

carlosfelix2:
This PR changes the CSV comparison to not rely on the expected data column order.  This allows the test to still pass if new schema fields are added.  

Test:
1. Run the integration tests with `gradle testIntegration` and verify the test pass

## Changes
- CSV comparison does not rely on the expected data column order
- Changed one expected data to remove the site_of_care column, which was the last added and broke the tests.  Tests now passes with these code changes and the missing column.

### Testing
- [x] Tested locally?
- [ ] Ran `quickTest all`?
- [ ] Added tests?

### Process

## Fixes
- #1704


jimduff-usds (approved):
Adding this to our release.

PR #1706, opened 2021-07-23
Updated script to use ownership flag

wwilliams-Agile6:
This PR ...

## Changes
-
-

### Testing
- [ ] Tested locally?
- [ ] Ran `quickTest all`?
- [ ] Added tests?

    - new endpoint A
    - new endpoint B

### Process

## Fixes
- #issue

## To Be Done
- #issue

### DIFF
--- a/prime-router/cleanslate.sh
+++ b/prime-router/cleanslate.sh
@@ -14,6 +14,7 @@ KEEP_BUILD_ARTIFACTS=0
 KEEP_PRIME_CONTAINER_IMAGES=0
 KEEP_ALL=0
 PRUNE_VOLUMES=0
+RETAKE_OWNERSHIP=0
 
 function usage() {
   cat <<EOF
@@ -31,6 +32,8 @@ OPTIONS:
   --prune-volumes           Forces a docker volume prune -f after taking containers
                             down (disables --keep-vault, including when set via --keep-all)
   --instructions            Shows post-run instructions
+  --retake-ownership        Change ownership of directories that are potentially 
+                            shared with container instances
   --verbose                 Get "more" output
   --help|-h                 Shows this help
 
@@ -66,6 +69,10 @@ Examples:
 
   # Use this if you like chasing red herrings in debug-land
   $ ${0} --keep-images
+
+  # Use this if you like to change ownership of some directories that are
+  # potentially shared with container instances
+  $ ${0} --retake-ownership 
 
 EOF
 }
@@ -311,7 +318,10 @@ function populate_vault() {
 function cleanup() {
   info "> Cleaning up your environment..."
   docker_decompose
-  take_directory_ownership
+  if [[ ${RETAKE_OWNERSHIP?} != 0 ]]; then
+    info "Change directory ownership" 
+    take_directory_ownership
+  fi
   cleanup_build_artifacts
   reset_vault
   return 0
@@ -324,7 +334,10 @@ function initialize() {
   ensure_binaries
   activate_containers
   populate_vault
-  take_directory_ownership
+  if [[ ${RETAKE_OWNERSHIP?} != 0 ]]; then
+    info "Change directory ownership" 
+    take_directory_ownership
+  fi
   return 0
 }
 
@@ -369,6 +382,9 @@ while [[ -n ${1} ]]; do
     ;;
   "--verbose")
     VERBOSE=1
+    ;;
+  "--retake-ownership")
+    RETAKE_OWNERSHIP=1
     ;;
   *)
     usage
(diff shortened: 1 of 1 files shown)


prime-router/cleanslate.sh
```diff
@@ -31,6 +32,8 @@ OPTIONS:
   --prune-volumes           Forces a docker volume prune -f after taking containers
                             down (disables --keep-vault, including when set via --keep-all)
   --instructions            Shows post-run instructions
+  --retake-ownership        Change ownership of directories that are potentially
```
td-usds, line 35: I would call this `--take-ownership` not `retake` (and then fix down as well)

prime-router/cleanslate.sh
```diff
@@ -67,6 +70,10 @@ Examples:
   # Use this if you like chasing red herrings in debug-land
   $ ${0} --keep-images
 
+  # Use this if you like to change ownership of some directories that are
+  # potentially shared with container instances
+  $ ${0} --retake-ownership
```
td-usds, line 75: A good example here would be `${0} --keep-all --take-ownership` to _just_ take ownership

prime-router/cleanslate.sh
```diff
@@ -311,7 +318,10 @@ function populate_vault() {
 function cleanup() {
   info "> Cleaning up your environment..."
   docker_decompose
-  take_directory_ownership
+  if [[ ${RETAKE_OWNERSHIP?} != 0 ]]; then
+    info "Change directory ownership" 
+    take_directory_ownership
```
td-usds, line 323: I would do this in `take_directory_ownership` itself instead of the call-sites:

```bash

# Takes ownership of some directories that are potentially shared with container instances
# Parameter: TAKE_OWNERSHIP value
function take_directory_ownership() {
  if [[ ${1} != 0 ]]; then
    info "Taking ownership of directories (may require elevation)..."
    TARGETS=(
      "./build/"
      "./docs/"
      "./.gradle/"
      "./.vault/"
    )

    # If you have no sudo, then this is a no-op and we warn
    SUDO_CMD=$(which sudo | head -n 1)
    if [[ -z "${SUDO_CMD}" ]]; then
      warning "You do not apear to have 'sudo'; it is highly probable that incorrect file permissions will lead to sftp upload failures when you run tests"
    fi

    for d in ${TARGETS[*]}; do
      echo -ne "    - ${d?}..."
      if [[ -d "${d?}" ]]; then
        # No '?' in the variable reference, it's fine for it to not exist, this may just be a no-op
        # and we warned you about it just above
        ${SUDO_CMD} chown -R "$(id -u -n):$(id -g -n)" "${d?}"
        ${SUDO_CMD} chmod -R a+w "${d?}"
        echo "DONE"
      else
        echo "ABSENT(OK)"
      fi
    done
  else
    info "Leaving directories alone..."
  fi
}
```

and when you call it, you do:
`take_directory_ownership ${RETAKE_OWNERSHIP?}` (or whatever you end up calling the variable)

td-usds (changes requested):

prime-router/cleanslate.sh
wwilliams-Agile6: Change made

prime-router/cleanslate.sh
wwilliams-Agile6: Change made

prime-router/cleanslate.sh
wwilliams-Agile6: Done

prime-router/cleanslate.sh
```diff
@@ -92,33 +99,39 @@ function error() {
...
-    "./.vault/"
-  )
-
-  # If you have no sudo, then this is a no-op and we warn
-  SUDO_CMD=$(which sudo | head -n 1)
-  if [[ -z "${SUDO_CMD}" ]]; then
-    warning "You do not apear to have 'sudo'; it is highly probable that incorrect file permissions will lead to sftp upload failures when you run tests"
-  fi
+  
+  if [[ ${1} != 0 ]]; then
```
td-usds, line 105: cf second comment from original review: consider making this `if [[ ${1:-1} != 0 ]]; then` to make this function be callable without arguments as well

td-usds (changes requested):
One small change to make, then looks good

prime-router/cleanslate.sh
wwilliams-Agile6: Shouldn't it be defaulted as 0?

wwilliams-Agile6: Shouldn't it be defaulted to 0 and not 1?

prime-router/cleanslate.sh
wwilliams-Agile6: Shouldn't it be defaulted to 0 and not 1?

td-usds: no, given the name of the function, we _want_ it to do the action, _unless_ told otherwise (i.e. arg 1 == non-zero)

td-usds (approved):

PR #1735, opened 2021-07-26
added dual template function

mry-usds:
This PR adds dual template function,

## Changes
- by supplying a template set separated by : (i.e. d-1234:d-1235) this will identify a template to use for having reports in the past 24hrs (d-1234) and for not having any (d-1235).
-

### Testing
- [X] Tested locally?
- [ ] Ran `quickTest all`?
- [ ] Added tests?

    - new endpoint A
    - new endpoint B

### Process

## Fixes
- #issue

## To Be Done
- #issue

### DIFF
--- a/prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
+++ b/prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
@@ -331,11 +331,7 @@ class EmailScheduleEngine {
      * @param emails list of emails to send to
      * @param logger context logger
      */
-    private fun dispatchToSendGrid(
-            template: String,
-            emails: Iterable<String>,
-            logger: Logger
-    ){
+    private fun dispatchToSendGrid(template: String, emails: Iterable<String>, logger: Logger) {
         val p: Personalization = Personalization()
         emails.forEach { to -> p.addTo(Email(to)) }
         p.setSubject(SUBJECT_EMAIL)
@@ -369,4 +365,5 @@ class EmailScheduleEngine {
         } else {
             logger.info("Can't find SENDGRID_ID secret")
         }
+    }
 }
--- a/prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
+++ b/prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
@@ -28,13 +28,13 @@ import com.sendgrid.helpers.mail.objects.Email
 import com.sendgrid.helpers.mail.objects.Personalization
 import gov.cdc.prime.router.azure.db.enums.SettingType
 import gov.cdc.prime.router.secrets.SecretHelper
-import org.json.JSONObject
 import java.io.IOException
 import java.time.OffsetDateTime
 import java.time.ZonedDateTime
 import java.util.logging.Level
 import java.util.logging.Logger
 import khttp.get as httpGet
+import org.json.JSONObject
 
 const val SCHEDULE = "*/5 * * * *" // every 5 minutes
 const val OKTA_ISSUER = "https://hhs-prime.okta.com/oauth2/default"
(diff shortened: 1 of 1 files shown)


prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
```diff
@@ -136,16 +137,36 @@ class EmailScheduleEngine {
...
+                    org
+                ).size;     
+                val template = getTemplate(schedule.template, countOfRecords );     
                 val emails: Iterable<String> = if (schedule.emails.size > 0) schedule.emails else getEmails(org, logger)
-                logger.info("EmailEngineFunction:: processing organization $org within ${schedule.template}")
+                logger.info("EmailEngineFunction:: processing organization $org within ${template}")
                 emails.forEach { email ->
-                    logger.info("EmailEngineFunction:: sending email to $email")
-                    dispatchToSendGrid(schedule.template, listOf(email), logger)
+                    logger.info("EmailEngineFunction:: sending email template ${template} to $email")
```
td-usds, line 148: Would it make more sense to capture whether the dispatch to sendgrid is successful and then in a `finally` block say: "email to X send, here's the result code of the dispatch")? This may make it easier to debug in the future (which is why we log in the first place)

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
```diff
@@ -136,16 +137,36 @@ class EmailScheduleEngine {
...
+                    dispatchToSendGrid(template, listOf(email), logger)
                 }
             }
         }
     }
 
+    /**
+     * Get the template to use
```
td-usds, line 156: This would benefit from a description of what the template looks like to indicate why we are doing this split.

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
```diff
@@ -136,16 +137,36 @@ class EmailScheduleEngine {
...
+     * 
+     * @param template the template string from the database
+     * @param count of records found in the last 24hrs
+     * 
+     * @returns template name to use
+     */
+    private fun getTemplate( template: String, count: Int ): String {
+        var retValue = template;
+        if( template.split(":").size > 1 )
```
td-usds, line 165: question rather than a request for change: what happens when the template contains more than 1 `:` and thus `.size` return `(2,...[` ? Is that possible in the future? If so, I would prefer a(n) warning/error/critical logging or something that tells us about this.

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
```diff
@@ -30,6 +30,7 @@ import gov.cdc.prime.router.secrets.SecretHelper
 import org.json.JSONObject
 import java.io.IOException
 import java.time.ZonedDateTime
+import java.time.OffsetDateTime
```
td-usds, line 33: Prefer sorting imports

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
```diff
@@ -136,16 +137,36 @@ class EmailScheduleEngine {
...
+                ).size;     
+                val template = getTemplate(schedule.template, countOfRecords );     
                 val emails: Iterable<String> = if (schedule.emails.size > 0) schedule.emails else getEmails(org, logger)
-                logger.info("EmailEngineFunction:: processing organization $org within ${schedule.template}")
+                logger.info("EmailEngineFunction:: processing organization $org within ${template}")
                 emails.forEach { email ->
-                    logger.info("EmailEngineFunction:: sending email to $email")
-                    dispatchToSendGrid(schedule.template, listOf(email), logger)
+                    logger.info("EmailEngineFunction:: sending email template ${template} to $email")
+                    dispatchToSendGrid(template, listOf(email), logger)
```
td-usds, line 149: `dispatchToSendGrid` accepts any `Iterable<String>`, why the explicit conversion to a list?

cf. https://github.com/CDCgov/prime-reportstream/blob/b15e028ff7a9cfa8b2fa023a6c119251eda734a8/prime-router/src/main/kotlin/azure/EmailEngineFunction.kt#L327
```kotlin
    private fun dispatchToSendGrid(
        template: String,
        emails: Iterable<String>,
        logger: Logger
    )
```

td-usds (changes requested):
Needs to be run through ktlint, I think.

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
Bisonburger: Iterable is an interface - need a tangible instance

td-usds: ah, apologies, I thought `email` was already iterable, but it's just a `string`, and so you're turning the one string into a list of strings.

Bisonburger (approved):
lgtm

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
```diff
@@ -307,7 +336,7 @@ class EmailScheduleEngine {
         template: String,
         emails: Iterable<String>,
         logger: Logger
-    ) {
+    ) : Response {
```
td-usds, line 339: What's the point of this returning a result? The call sites do not do anything with the result yet (cf line 150`dispatchToSendGrid(template, listOf(email), logger)`)
I see we log the warning here, is that enough and can we "get away" with not returning the response? We are doing logging here already, this may be enough... (or maybe this returns a `true`/`false` indicating success or failure)

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
```diff
@@ -319,21 +348,30 @@ class EmailScheduleEngine {
...
                 request.setBody(mail.build())
-                sg.api(request)
+                response = sg.api(request)
             } catch (ex: IOException) {
                 logger.warning("Can't contact sendgrid")
             }
+            finally {
+                logger.info( "sending to $emails - result ${response.getStatusCode()}" )
+                if( response.getStatusCode() != 200 )
+                    logger.info( "error - ${response.getBody()}")
```
td-usds, line 369: Should this be a `logger.error` instead?

I am not familiar with SendGrid's API, and so this question is more for my information: are there cases where other `2xx` codes are returned indicating success or does it only return `200` on success. For instance (cf. https://en.wikipedia.org/wiki/Http_status_code#2xx_success) Would it make more sense to look for any status code inside the `[200, 299]` range (which I know is broader than what HTTP currently uses, but is semantically correct)

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
```diff
@@ -136,16 +138,43 @@ class EmailScheduleEngine {
...
+     * @param template the template string from the database (i.e d-12345678 or d-12345678:d-87654321)
+     * @param count of records found in the last 24hrs
+     * 
+     * @returns template name to use
+     */
+    private fun getTemplate( template: String, count: Int, logger: Logger ): String {
+        var retValue = template;
+        val split = template.split(":")
+
+        if( split.size > 2 )
```
td-usds, line 170: I like this! This is clean and neat!

td-usds (changes requested):

td-usds (approved):

Bisonburger (approved):

MauriceReeves-usds (approved):

PR #1758, opened 2021-07-29
tweaked capitalization of schema folder

jimduff-usds:
Changed capitalization of schema folder name.


jimduff-usds (commented):
I missed this during the review.

PR #1761, opened 2021-07-30
GitHub run_gitleaks Action Fix

### LINKED ISSUE #1760: run_gitleaks.yml GitHub action seems to run on the full history instead of only a subset
`run_gitleaks.yml` invokes `.environment/gitleaks/run-gitleaks.sh --since` with a specified hash. The hash does not appear to be respected and the scheduled run targets the full history

td-usds:
This PR changes how we run gitleaks on a schedule and fixes #1760 (linked issue for auto-close).
Scanning between commits is not reliable (yet) when there are multiple branches involved. When it's all in a single branch, then the tool can easily figure things out, but as soon as you have multiple, the tool ends up scanning an undeterministic subset of the entire history.

The fix for this is to scan since an LKG (Last Known Good) that is a date instead.

This PR introduces 2 new runmodes to the run-gitleaks script:
- `--since <datespec>`: scans commits that are created on or after this date
- `--depth <int>`: scans the last <int> commits from `HEAD` backwards

The `run_gitleaks.yml` GitHub action has been modified to use the `--since` version which now records an LKG date of the last time we had a leak. It currently uses `2021-06-01` as LKG date which at this contains roughly ~1500 commits:
```bash
$ .environment/gitleaks/run-gitleaks.sh --since '2021-06-01'; cat gitleaks.log 
Gitleaks> info: Scanning all commits since 2021-06-01.
time="2021-07-30T17:57:48Z" level=info msg="opening /repo\n"
time="2021-07-30T17:59:16Z" level=info msg="scan time: 1 minute 27 seconds 751 milliseconds 724 microseconds"
time="2021-07-30T17:59:16Z" level=info msg="commits scanned: 1457"
time="2021-07-30T17:59:16Z" level=info msg="No leaks found"
```

Lastly, this PR also whitelists some additional findings in the code that are false positives.

### DIFF
--- a/.environment/gitleaks/gitleaks-config.toml
+++ b/.environment/gitleaks/gitleaks-config.toml
@@ -175,6 +175,7 @@ title = "PRIME ReportStream Gitleaks Configuration"
             'val redoxKey = \"apiKey\"',                                                    # Environment variable name, not value
             'val secretEnvName = \"REDOX_SECRET\"',                                         # Environment variable name, not value
             'var keymap = require\(',
+            'var tokenSigningSecret = "UVD4QOJ3H295Zi9Ayl3ySuoXNKiE8WYuOsaXOZfug3dwTUVBC1ZIKRPpG5LEyZDZ"',         # Invalidated
             'xox[baprs]-([0-9a-zA-Z]{10,48})',
         ]
         paths = [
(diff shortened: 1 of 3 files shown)


jimduff-usds (approved):

ronaldheft-usds (approved):
Additional whitelist looks good and I agree with since being an acceptable alternative.

Question: This may be an extreme edge case, but what would happen if a branch that existed before the date would be merged in? If the branch had a secret in it, would the scanner catch it because it would be going over the merge commit?

PR #1792, opened 2021-08-04
Improved DB access docs

ronaldheft-usds:
This PR updates our DB access docs.

## Changes
- Example uses read-only connection
- Add to example a one-liner for setting and parsing the access token
- Adds notes about session expiration
- Fixes script syntax highlighting

### DIFF
--- a/prime-router/docs/azure_database_login.md
+++ b/prime-router/docs/azure_database_login.md
@@ -16,7 +16,7 @@ az account get-access-token --resource-type oss-rdbms
 
 You will receive a response that looks like:
 
-```shell
+```json
 {
   "accessToken": "token",
   "expiresOn": "2021-04-22 16:22:54.350957",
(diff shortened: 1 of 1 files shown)


prime-router/docs/azure_database_login.md
```diff
@@ -4,19 +4,19 @@ This is the process for logging into our database via Azure Active Directory.
...
 
-```aidl
+```shell
 az account get-access-token --resource-type oss-rdbms
 ```
 
 You will receive a response that looks like:
 
-```aidl
+```shell
```
td-usds, line 19: use `json` instead of `shell`

td-usds (approved):
One small thing: make the json verbatim code be `json` instead of `shell` but otherwise looks good

prime-router/docs/azure_database_login.md
td-usds, line 19:
```suggestion
```json
```

td-usds (approved):
see suggested change, otherwise looks good

td-usds (approved):

Issue #1420, opened 2021-06-28
HL7 ingest needs to verify CODE values


carlosfelix2:

jimduff-usds:
I think CSV has this same issue - is this just an HL7 issue, or wider?

san-nat:
To recreate this issue: 
a) Try an hl7 to csv test with the receiver (az-phd.elr-prod) & with an invalid specimen_source_site_code i.e. one that does not exist on prime-router/docs/schema_documentation/az-az-covid-19-hl7.md . 

b) tail the router logs
`docker logs -f  prime-router_prime_dev_1`

[az-elr.hl7.zip](https://github.com/CDCgov/prime-reportstream/files/6870892/az-elr.hl7.zip)

make a curl call
`curl -X POST -H "Content-Type:application/hl7-v2" -H "client: ignore.ignore-hl7" --data-binary @./abovefile.hl7 'http://localhost:7071/api/reports'`

Note that api reports endpoint succeeds while fails on translation. see the exception

 an java.lang.IllegalStateException: Internal Error: '718IG36000' cannot be formatted for 'Specimen_Site' ('specimen_source_site_code') in the router logs. see logs on https://gist.github.com/san-nat/0691061327947aa13c832bab16630073

Issue #1468, opened 2021-07-03
Write a proposal to simplify valuesets, tables, and $alt tables, so people writing schemas for Senders are not confused.


jimduff-usds:
We have at least 3 ways of specifying enumerated fields:
- valuesets
- tables
- alt tables.

Its all very confusing.   It also overlaps with our concept of mappers and default values.   Study all this before you start, and experiment with it too.  Talk to Rick, Mo, Jim.

Our customer for this is, for example, the Sender Automation team - people who are designing and building schemas on behalf of our Senders.    This ticket is for you to come up with a proposal to simplify, with the those "Sender representative" customers in mind, and present to Rick, Mo, Jim, etc as needed, as well as to folks who will be doing that kind of schema development work.

- Can we just have Table as our only way of doing this?
- Consider, for example, the many ways of specifying Yes, No, and Unknown/no answer.   
- A valueset is just a table with two columns.   Change this so that value sets are tables.
- Get rid of alt tables, and just allow adding new columns of possible values to valueset tables.
- Keep the need for flexibility ('pandemic ready') in mind.

(See also related ticket #230)

jimduff-usds:
See this file (in a PR as of this writing) for example of the craziness people are doing:    prime-router/metadata/valuesets/sender-automation.valuesets

Talk to Rick Hood and get his suggestions as a "user" of the valueeset feature.

PR #1794, opened 2021-08-05
1420 Verify Coded Element codes in HL7 ingest

carlosfelix2:
This PR adds the verification of coded element codes (e.g. hl70189 H for Hispanic).

To test:
1. Run `gradle test testIntegration` and verify all tests pass
2. Run `./prime test` and verify all tests pass
3. Run `./prime test --run compare` and verify all test pass

## Changes
- Validate CODE fields against existing codes and add warning if unable to validate
- Added CWE code to hl70125
- Added 119297000 code to specimen type.  This came from a real life example we have
- Fixed issue with CSV compare not handling new schema fields
- Updated test files to reflect new code

### Testing
- [x] Tested locally?
- [x] Ran `quickTest all`?
- [x] Added tests?

### Process

## Fixes
- #1420


prime-router/metadata/schemas/covid-19.schema
```diff
@@ -683,6 +699,22 @@ elements:
     documentation: |
       The patient's race. There is a common valueset defined for race values, but some states may choose to define different code/value combinations.
 
+  - name: patient_race_id
```
jimduff-usds, line 702: The other way to do this would be for hl7-ingest-covid-19 schema to override the `valueSet: hl7005` in `patient_race` with `valueSet: whatever`, then define a `whatever` Valueset.    Why not do it that way, instead of adding a new mapper( ...)?

This implementation trades off flexibility with readability.   Each row of data can potentially come from a different type system.  However, we'll no longer have a way to go into XYZ's schema and know what valueset XYZ is using for patient_race.

The design around this way of managing multiple valuesets was meant to be one of the major topics in #1468:   Let's say there are 50 different ways senders/receivers have to define the possible values for patient_race -- how do we manage that?  What tools do we have for translating between different value sets?

prime-router/src/main/kotlin/Mappers.kt
```diff
@@ -742,6 +742,60 @@ class NullMapper : Mapper {
...
+        return if (values.isEmpty() || values.size != 2) {
+            null
+        } else {
+            var value: String? = null
+            if (!element.valueSet.isNullOrBlank()) {
+                val system = values[1].value
+                val verifiedCode = element.toCode(values[0].value, metadata.findValueSet(system.lowercase()))
+                value = when {
+                    // Fix incorrect codes for hl70189
+                    "hl70189".equals(element.valueSet, true) && "hl70189".equals(system, true) -> {
```
jimduff-usds, line 769: Hmm.   I think I hit a very similar problem a few months back.  My hack solution was different from yours, as found in direct/direct-covid-19.schema:

```
  - name: patient_ethnicity
    documentation:  Internally, ReportStream uses hl70189 (H,N,U), but should use HHS values. (2135-2, 2186-5, UNK, ASKU). A mapping is done here, but\
 better is to switch all of RS to HHS standard.
    altValues:
      - code: H
        display: 2135-2
      - code: N
        display: 2186-5
      - code: U
        display: UNK
      - code: U
        display: ASKU
    csvFields: [{ name: patientEthnicity, format: $alt}]
```

jimduff-usds (changes requested):
Let's discuss where we're headed with this.   And maybe we should outline a plan for #1468 before we proceed with this.   See my other comment here.

prime-router/metadata/schemas/covid-19.schema
carlosfelix2: Removed these changes as discussed since it needs a lot more thought behind it.  Will leave for #1468

prime-router/src/main/kotlin/Mappers.kt
carlosfelix2: Removed these changes as discussed since it needs a lot more thought behind it.  Will leave for #1468

jimduff-usds (approved):
lgtm

PR #1824, opened 2021-08-10
gitleaks workflow: output the non-sensitive log after a gitleaks run

td-usds:
This PR adds outputting of the `gitleaks.log` that is generated by the gitleaks to the nightly `run_gitleaks.yaml` workflow. This log file does **NOT** contain the violations themselves, but it does contain metrics about the violations, thus telling us how _bad_ the run was:
- scan duration
- commits scanned
- violation count


ronaldheft-usds (approved):
This is a reasonable addition and safe to output.

PR #1825, opened 2021-08-10
gitleaks: fix 2021-08-10 run

td-usds:
This PR fixes the 2021-08-10 gitleaks run using a robust pattern that tripped for `aad_object_keyvault_admin` due to the presence of `key` in the name.


ronaldheft-usds (approved):
Thank you for improving this.

Issue #1828, opened 2021-08-10
Create property to set CLIA for out of State testing


brick-green-agile6:
California has asked for a single centralized CLIA number be sent for testing that originates outside the state. So we need a feature to set that CLIA number for the organization.

Acceptance criteria:
- Any single CLIA number can be set for a receiver
- The CLIA number set will only apply to results sent by facilities that are outside of the receiver jurisdiction
- The CLIA number will not change for results sent by facilities in the jurisdiction

Issue #1788, opened 2021-08-04
Further requests from CA


jimduff-usds:
I sent the attached file to CA-DPH, per their request, as an example of what we might be able to produce.   (See #1787)

(Jim Note:  HL7s run thru two programs at CA.   1) CCRS aka Optum, then 2) CalREDIE)

They ran it thru their system manually and sent us this email response:

---------------------

So one thing I can see that we need changed is the AOE questions which are currently showing as OBX segments.  We need those as NTEs sent after the OBX and before the SPM segment.  Here is an example format:
 
```
NTE|1|L|First test for condition of interest - UNKNOWN|
NTE|2|L|Employed in a healthcare setting - NO|
NTE|3|L|Has symptoms related to condition of interest - YES|
NTE|5|L|Admitted to intensive care unit for condition of interest - NO|
NTE|6|L|Resides in a congregate care setting - UNKNOWN|
 
```
Once I changed those and sent the HL7 in, it came right through for our Santa Clara build.

-  It would be nice to see “Simple Report” or “ReportStream” referenced somewhere.  I see OBX-17 is coming through as “10811877011290_DIT^^99ELR^^^^2.68^^10811877011290_DIT” so could we change it to “10811877011290_DIT^Simple Report^99ELR^^^^2.68^^10811877011290_DIT” if that works for you.  I tested it and it comes through okay into CalREDIE in the HL7.  It could also be added in the SFT segment. And if we want it to be visible to end users, we could add it as another NTE. 
 
Anyhow, happy to see this is moving along.  Please let us know once the temp changes you made below (I guess mainly the ISO reference) have been finalized, the AOEs moved to NTEs and the reference to Simple Report or ReportStream has been added and you have sent in more results. 
 
Thanks again for everyone’s work on this!
 
PS – just to see, I did send in a modified message using CLIA instead of ISO in ORC-2.4 and OBR-2.4 and it worked.  I checked our spec and it says it should handle up to 6 characters.  The default is ISO, but it seems like you can use CLIA after all.  And this was sent through CCRS, so it didn’t error there.  Here is the message if you want to review it.  I highlighted the changes (apart from patient demographics and the accession). 
 
```
[...]

PR #1840, opened 2021-08-12
Bg 1828 ca onboarding

brick-green-agile6:
This PR adds a property to the translation configuration to chance the CLIA number for testing completed outside the state of the receiver.

## Changes
- Property added to HL7 Configuration (cliaForOutOfStateTesting) accepting a string value
- Reviews the StateCode set in the organizations.yml for the receiver against the Performing Organization State (OBX-24.4) value. When those values are different, the replacement CLIA is set only in MSH-4.2.
- Set the property for California's primary feed and secondary feed

### Testing
- [X] Tested locally?
- [x] Ran `quickTest all`?
- [ ] Added tests?

## Fixes
- #1828 
- #1788

### DIFF
--- a/prime-router/src/main/kotlin/serializers/Hl7Serializer.kt
+++ b/prime-router/src/main/kotlin/serializers/Hl7Serializer.kt
@@ -12,8 +12,17 @@ import ca.uhn.hl7v2.parser.CanonicalModelClassFactory
 import ca.uhn.hl7v2.parser.EncodingNotSupportedException
 import ca.uhn.hl7v2.parser.ModelClassFactory
 import ca.uhn.hl7v2.util.Terser
-import gov.cdc.prime.router.*
-import gov.cdc.prime.router.cli.tests.CoolTest
+import gov.cdc.prime.router.Element
+import gov.cdc.prime.router.ElementAndValue
+import gov.cdc.prime.router.FileSettings
+import gov.cdc.prime.router.Hl7Configuration
+import gov.cdc.prime.router.Mapper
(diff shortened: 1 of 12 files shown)


prime-router/src/main/kotlin/serializers/Hl7Serializer.kt
```diff
@@ -565,6 +567,22 @@ class Hl7Serializer(val metadata: Metadata) : Logging {
             }
         }
 
+        // check for alt CLIA for out of state testing
+        if (!hl7Config?.cliaForOutOfStateTesting.isNullOrEmpty()) {
+            val testingStateField = "OBX-24-4"
+            val pathSpecTestingState = formPathSpec(testingStateField)
+            val testingState = terser.get(pathSpecTestingState)
+
+            val settings = FileSettings(FileSettings.defaultSettingsDirectory)
```
MauriceReeves-usds, line 576: This could potentially lead to performance issues and bugs. The biggest issue here is that in production we load the organization information from the database, and not from files. If you look at `WorkflowEngine` you'll see it does a check to see if the `FEATURE_FLAG_SETTINGS_ENABLED` is true, and if it is it loads from the DB instead. I'd like you to change the logic here and add `FileSettings` as a param we pass into the `HL7Serializer`, and then in `WorkflowEngine` pass it in like `metadata` is. Make it a nullable parameter, and you can mock it in in your unit tests.

MauriceReeves-usds (changes requested):

MauriceReeves-usds (approved):

PR #1843, opened 2021-08-12
Updated getting_started.md document with a few more details

carlosfelix2:
This PR updates the getting started doc for developers to add a few more details and tried to make some sections a bit simpler for new developers.  I also rearranged a few topics to try to make the document flow in order from building to running.  I did not go through the entire document, just a few sections. 

It is easier to read the file at this link: https://github.com/CDCgov/prime-reportstream/blob/0bf60a1c730bee9f765da074aa7bec7cda572ee4/prime-router/docs/getting_started.md

## Changes
- updates the getting started doc 
- Added new Gradle task to reset/reload the database

### Testing
- [x] Tested locally?
- [ ] Ran `quickTest all`?
- [ ] Added tests?

### Process

### DIFF
--- a/prime-router/docs/getting_started.md
+++ b/prime-router/docs/getting_started.md
@@ -377,9 +377,7 @@ dropdb --force ${PGDATABASE?}
 createdb --owner=${PGUSER?} ${PGDATABASE?}
 
 # migrate the local database by hand
-flyway "-user=${PGUSER?}" -password=changeIT! \
-    -url=jdbc:postgresql://${PGHOST?}:5432/${PGDATABASE?} \
-    -locations=filesystem:./src/main/resources/db/migration migrate
+./gradlew migrate
 ```
 
 # Additional tooling
--- a/prime-router/docs/getting_started.md
+++ b/prime-router/docs/getting_started.md
@@ -1,98 +1,107 @@
-% Developer Getting Started Guide
+# Developer Getting Started Guide
 
-This document will walk you through the setup instructions to get a functioning development environment from first clone to completion of your first run of the end-to-end tests.
+This document will walk you through the setup instructions to get a functioning development environment.
 
 # Table of contents
-
-* [Locally installed software prerequisites](#locally-installed-software-prerequisites)
-* [First build](#first-build)
-    * [Build dependencies](#build-dependencies)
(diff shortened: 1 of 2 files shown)


prime-router/docs/getting_started.md
```diff
@@ -386,22 +366,17 @@ cd ./prime-router
...
-
-# migrate the local database by hand
-flyway "-user=${PGUSER?}" -password=changeIT! \
-    -url=jdbc:postgresql://${PGHOST?}:5432/${PGDATABASE?} \
-    -locations=filesystem:./src/main/resources/db/migration migrate
-```
+## Resetting the Database
+1. Stop your ReportStream container if it is running.
+    ```bash
+    docker compose down
```
td-usds, line 372: I'd stick to using `docker-compose` as opposed to `docker compose`.

carlosfelix2: Hmmm... This is the message from docker-compose:
Docker Compose is now in the Docker CLI, try `docker compose`

cweagans: `docker compose` is v2 of compose (rewritten in Go), whereas `docker-compose` defaults to the python-based compose v1. Further testing may be needed to ensure that v2 works with our setup (see `docker-compose version` vs `docker compose version`).

jimduff-usds (approved):
Thanks!!

prime-router/docs/getting_started.md
carlosfelix2: I looked a bit more into this.  It looks like they just created a replacement to the tool regardless of the version of the files.  So, it really is the same command with just the Dash removed.  The one thing I could not find is where is this command supported and some text leads me to believe it may only work for Docker Desktop in Windows or Mac.  I'll switch this text to use the dash, but I am curious if Linux has this change too.
`The Docker CLI now supports the compose command, including most of the docker-compose features and flags, without the need for a separate tool. You can replace the dash (-) with a space when you use docker-compose to switch over to docker compose. You can also use them interchangeably, so that you are not locked-in with the new compose command and, if needed, you can still use docker-compose.`  - https://collabnix.com/docker-compose-now-shipped-with-docker-by-default/

Issue #1709, opened 2021-07-23
Create feature to provide replacement values for any HL7 field per organization


brick-green-agile6:
We need to be able to set replacement values for any HL7 field. There is currently an option to suppress fields which essentially sets the field to a blank string. This feature will expand that functionality to set the replacement to any value including blank.

Strategy:

- Add property to HL7Configuration to accept array of HL7 fields with replacement values
- loop through pairs to set default values in HL7 serializer

Acceptance criteria:

- empty or filled strings can be set as default values
- only fields with a value can be replaced. Blank values will be left blank.
- does not error for non existent HL7 field (ex. MSH-99999-1) in serializer only in test
- allows any number of fields to be set
- checks that replacement values are only set once per organization (don't want later replacements to overwrite earlier replacements)

Notes:
- Are there fields where replacements should not be allowed? (Test name, test result, patient name, etc)
     - Answer: This is currently only set by developers. No restriction needed. Restriction may be needed when/if the functionality is made available to receivers. 

Solution:
Use yaml dictionary then loop through map in hl7 serializer
```
replaceValue:
- OBX-23-1: ReplacementVal
- PID-14-5: ReplacementVal
```

brick-green-agile6:
@carlosfelix2 Can you take a look at this?

jlusds:
What's the status of this?

brick-green-agile6:
This is completed by PR #1740 . But it hasn't been merged in yet. Still need a review.

PR #1846, opened 2021-08-13
Bg bugfix replaceValue

brick-green-agile6:
This PR fixes the replaceValue translation property. Prior to this fix, only the first iteration of repeat OBX segments was effected. This patch will examines all the OBX segment repeats and replaces the values within.

## Changes
- Added nested for loop within the replaceValue loop in the createMessage function in hl7 serializer to loop through all OBX segments
- Edited unit tests to test loop

## Checklist
- [ ] an OBX value can be set to replace in an organization setting
- [ ] All iterations of that OBX value will be replaced

### Testing
- [X] Tested locally?
- [X] Ran `quickTest all`?
- [ ] Added tests?

## Fixes
- #1709


jorg3lopez (approved):
Tested locally.

Issue #1848, opened 2021-08-13
CA Onboarding - 20 char limit exceeded when special HL7 chars are used


brick-green-agile6:
## Description
Problem 1: The facility name "Shelters & Unhoused" which is 20 chars long gets changed to "Shelters \T\ Unhoused" which is 21 characters long, and so that breaks a max length rule. ("MSH-4.1 Data Type error, exceeded - Max Length is 20")

The fix presumably is to check the max length AFTER the value is transformed, not before.
This apparently occurred in all these places - make sure they are all fixed:

MSH-4.1 Data Type error, exceeded - Max Length is 20
OBR-2.2 is exceeding max length of 20
OBR-3.2 is exceeding max length of 20
ORC-2.2 is exceeding max length of 20
ORC-3.2 is exceeding max length of 20
PID-3.4.1 is more than 20 characters Max Length

## Acceptance Criteria
- [ ] the 20 character limit takes HL7 special characters into account

PR #1852, opened 2021-08-13
Bg 1848 20 char limit

brick-green-agile6:
This PR filters the usage of HD_TRUNCATION_LIMIT to account for HL7 special characters presetn in the incoming value. Special HL7 characters, when sent as part of a string input value, have to be escaped. That escaping adds additional characters to the string causing the 20 character limit to be exceeded. This update will first check if special characters are present then adjust the truncation limit accordingly..

## Changes
- added function setTruncationLimitWithEncoding() to HL7Serializer to find special characters in value then shorten the truncation limit
- set function in HL7 serializer

### Testing
- [X] Tested locally?
- [X] Ran `quickTest all`?
- [ ] Added tests?

## Fixes
- #1848


prime-router/src/main/kotlin/serializers/Hl7Serializer.kt
```diff
@@ -924,6 +924,16 @@ class Hl7Serializer(val metadata: Metadata) : Logging {
         terser.set("/PATIENT_RESULT/ORDER_OBSERVATION/OBSERVATION/OBX-23-7", "XX")
     }
 
+    private fun setTruncationLimitWithEncoding(value: String, truncationLimit: Int): Int {
```
carlosfelix2, line 927: Need kdoc

carlosfelix2: Add unit tests for this method.  What happens with 0, 1 and many escape characters (not sure if many are allowed).

prime-router/src/main/kotlin/serializers/Hl7Serializer.kt
```diff
@@ -924,6 +924,16 @@ class Hl7Serializer(val metadata: Metadata) : Logging {
...
+        val regex = "[^&~|]".toRegex()
+        val match = regex.containsMatchIn(value)
+        return if (match) {
+            truncationLimit.minus(2)
```
carlosfelix2, line 931: Can there be multiple escape characters?

carlosfelix2 (changes requested):

prime-router/src/main/kotlin/serializers/Hl7Serializer.kt
brick-green-agile6: Yep, there can be multiple and this wouldn't account for them all. Good catch! I'll adjust

prime-router/src/test/kotlin/serializers/Hl7SerializerTests.kt
```diff
@@ -635,6 +636,16 @@ NTE|1|L|This is a final comment|RE"""
         }
     }
 
+    @Test
+    fun `test setTruncationLimitWithEncoding`() {
+
+        val testValue = "Test & Value ~ Text ^ String"
```
carlosfelix2, line 642: When doing unit tests, you need to cover all scenarios.  In this case you are missing the no special characters scenario to cover the else in the if statement.

brick-green-agile6: got this fixed

carlosfelix2 (approved):

prime-router/src/test/kotlin/serializers/Hl7SerializerTests.kt
```diff
@@ -635,6 +636,19 @@ NTE|1|L|This is a final comment|RE"""
         }
     }
 
+    @Test
+    fun `test setTruncationLimitWithEncoding`() {
+
+        val testValueWithSpecialChars = "Test & Value ~ Text ^ String"
+        val testValueNoSpecialChars = "Test Value Text String"
+        val testLimit = 20
```
carlosfelix2, line 644: Don't feel you have to define all the numbers separately.  next time you could do just
        assertEquals(serializer.getTruncationLimitWithEncoding(testValueWithSpecialChars, testLimit), 14)

PR #1884, opened 2021-08-17
Update VPN doc with Windows DNS issue

carlosfelix2:
This PR adds info on how to fix the Windows DNS issue with OpenVPN

### DIFF
--- a/prime-router/docs/vpn.md
+++ b/prime-router/docs/vpn.md
@@ -44,7 +44,9 @@ $ kill -${SIGNAL} $(cat "/tmp/${USER}/openvpn.pid")
 ## Windows
 ### VPN DNS Resolution
 There is an issue where the DNS server for the VPN adapter is not used when resolving hostnames.  This is due to the interface metric of the VPN loosing to the metric of your normal network adapter.  Run the following command to look up the IP of a server in the Azure environment and test if this is an issue .  For example:
+
 `nslookup pdhstaging-pgsql.postgres.database.azure.com`
+
 should return a 10.x.x.x IP address which is in the range used by the VPN.  If you see an address outside of this 10.x.x.x range then continue with the instructions here to fix this issue.
 
 To fix this issue:
@@ -59,9 +61,6 @@ To fix this issue:
 
 Run the Windows command `netsh interface ip show config` to verify the InterfaceMetric value has change to 1 for the OpenVPN TAP-Windows interface.  For example:
 
-
-Reference: https://aws.amazon.com/premiumsupport/knowledge-center/client-vpn-fix-dns-query-forwarding/ - it says that changing this setting via the control panel does not work, but this has been confirmed to work.
-
 ```
 Configuration for interface "OpenVPN TAP-Windows6"
     DHCP enabled:                         Yes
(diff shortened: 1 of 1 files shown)


prime-router/docs/vpn.md
```diff
@@ -41,6 +41,38 @@ $ kill -${SIGNAL} $(cat "/tmp/${USER}/openvpn.pid")
 # On termination, openvpn will remove the PID file
 ```
 
+## Windows
+### VPN DNS Resolution
+There is an issue where the DNS server for the VPN adapter is not used when resolving hostnames.  This is due to the interface metric of the VPN loosing to the metric of your normal network adapter.  Run the following command to look up the IP of a server in the Azure environment and test if this is an issue .  For example:
+`nslookup pdhstaging-pgsql.postgres.database.azure.com`
```
carlosfelix2, line 47: Is there a better host I can use than this?

td-usds: They're going to be dependent on which environment you connect to; for prod, this would be `pdhprod-pgsql.postgres.database.azure.com`

prime-router/docs/vpn.md
```diff
@@ -41,6 +41,39 @@ $ kill -${SIGNAL} $(cat "/tmp/${USER}/openvpn.pid")
 # On termination, openvpn will remove the PID file
 ```
 
+## Windows
+### VPN DNS Resolution
+There is an issue where the DNS server for the VPN adapter is not used when resolving hostnames.  This is due to the interface metric of the VPN loosing to the metric of your normal network adapter.  Run the following command to look up the IP of a server in the Azure environment and test if this is an issue .  For example:
+
+`nslookup pdhstaging-pgsql.postgres.database.azure.com`
+
+should return a 10.x.x.x IP address which is in the range used by the VPN.  If you see an address outside of this 10.x.x.x range then continue with the instructions here to fix this issue.
```
td-usds, line 50: Consider using CIDR notation (i.e. `10.0.0.0/8` instead of `10.x.x.x`)
```suggestion
should return a 10.0.0.0/8 IP address which is in the range used by the VPN.  If you see an address outside of this 10.0.0.0/8 range then continue with the instructions here to fix this issue.
```

td-usds (changes requested):

prime-router/docs/vpn.md
ronaldheft-usds: Maybe?

```suggestion
`nslookup pdh<env>-pgsql.postgres.database.azure.com`
```

ronaldheft-usds (approved):

prime-router/docs/vpn.md
carlosfelix2: > Consider using CIDR notation (i.e. `10.0.0.0/8` instead of `10.x.x.x`)

I would like to use something that people that do not understand that notation can follow.  Something simple.  Any other ideas?

td-usds: I think it's fine to require folks to understand CIDR notation, especially given that this is a topic about networking.
If we were talking about `/20` or other uncommon masks then sure, but `/8`, `/16`, `/24` are pretty common and easily grokable.

carlosfelix2: Makes sense.  Will change.

td-usds (approved):

PR #1908, opened 2021-08-19
Run terraform fmt as githook and GH Action

ronaldheft-usds:
This PR brings some sanity to our Terraform formatting by using [terraform fmt](https://www.terraform.io/docs/cli/commands/fmt.html).

## Changes
- Adds a pre-commit githook that validates all Terraform files are formatted per `terraform fmt` specs before committing
  - The pre-commit githook provides details for fixing the files, including running the hook with the `--fix` option
- Runs a verification of Terraform formatting as part of GitHub Action as an added failsafe
- Formats all existing Terraform files
  - This will make for some merge conflicts, but it's a necessary evil
- Updates our `.editorconfig` to help the IDE make the same formatting choice at `terraform fmt`

### Testing
- [x] Tested locally?
- [ ] Ran `quickTest all`?
- [ ] Added tests?

    - new endpoint A
    - new endpoint B

### Process

## Fixes
- #issue

## To Be Done
- #issue

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -6,5 +6,5 @@ indent_style = space
 insert_final_newline = false
 tab_width = 4
 
-[{*.yaml,*.yml}]
+[{*.yaml,*.yml, *.tf}]
 indent_size = 2
--- a/.environment/terraform-fmt/run-terraform-fmt.sh
+++ b/.environment/terraform-fmt/run-terraform-fmt.sh
@@ -13,61 +13,35 @@ function note() {
 }
 
 REPO_ROOT=$(git rev-parse --show-toplevel)
-UNCOMMITTED_CHANGES=0
 TERRAFORM_NEEDS_FMT=0
-
-function git_stash() {
-    if [[ $(git diff-index --quiet HEAD --) ]]; then
-        warning "Stashing current changes so we verify formatting with committed files"
-        git stash
-    fi
-    RC=$?
-
-    return ${RC?}
-}
-
-function git_unstash() {
-    if [[ ${UNCOMMITTED_CHANGES?} != 0 ]]; then
-        warning "Restoring stashed changes"
-        git stash pop
-    fi
-    RC=$?
-
-    return ${RC?}
-}
+LOGFILE="terraform-fmt.log"
 
 function terraform_check_fmt() {
     note "Checking Terraform formatting."
-    make -f ${REPO_ROOT?}/operations/Makefile tf-cmd TF_CMD="terraform fmt -check -recursive /app/src"
-    RC=$?
-    TERRAFORM_NEEDS_FMT=${RC?}
-
-    return ${RC?}
+    make -C "${REPO_ROOT?}/operations" -f "${REPO_ROOT?}/operations/Makefile" tf-cmd TF_CMD="terraform fmt -check -recursive /app/src" > "${REPO_ROOT?}/${LOGFILE?}" 2>&1
+    TERRAFORM_NEEDS_FMT=$?
+    return ${TERRAFORM_NEEDS_FMT?}
 }
 
 function terraform_fmt() {
     if [[ ${TERRAFORM_NEEDS_FMT?} != 0 ]]; then
         warning "Formatting all Terraform files."
-        make -f ${REPO_ROOT?}/operations/Makefile tf-cmd TF_CMD="terraform fmt -recursive /app/src"
-        RC=$?
-
-        return ${RC?}
+        make -C "${REPO_ROOT?}/operations" -f "${REPO_ROOT?}/operations/Makefile" tf-cmd TF_CMD="terraform fmt -recursive /app/src" >> "${REPO_ROOT?}/${LOGFILE?}" 2>&1
+        return $?
     fi
-
     return 0
 }
-
-
-#git_stash
 
 terraform_check_fmt
 RC=$?
 
-#terraform_fmt
-#git_unstash
+terraform_fmt
 
 if [[ ${RC?} != 0 ]]; then
-    error "(return code=${RC?}) Your Terraform files are not formatted. \`terraform fmt\` has been run and the changes can be committed to comply with formatting requirements."
+    error "(return code=${RC?}) Your Terraform files are not formatted."
+    error "\`terraform fmt\` has been run and the changes can be committed to comply with formatting requirements."
+    error "Additional information can be found in the following files"
+    error "     - ${REPO_ROOT?}/${LOGFILE?}"
 fi
 
 exit ${RC?}
--- a/.environment/terraform-fmt/run-terraform-fmt.sh
+++ b/.environment/terraform-fmt/run-terraform-fmt.sh
@@ -1,0 +1,73 @@
+#!/usr/bin/env bash
+
+function error() {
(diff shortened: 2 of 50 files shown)


.environment/terraform-fmt/run-terraform-fmt.sh
```diff
@@ -0,0 +1,47 @@
...
+}
+
+terraform_check_fmt
+RC=$?
+
+terraform_fmt
+
+if [[ ${RC?} != 0 ]]; then
+    error "(return code=${RC?}) Your Terraform files are not formatted."
+    error "\`terraform fmt\` has been run and the changes can be committed to comply with formatting requirements."
```
td-usds, line 42: Any changes performed by `terraform fmt` are not going to be staged and so must be added to the user's staged changes:

```suggestion
    error "\`terraform fmt\` has been run and the changes can be committed to comply with formatting requirements."
    error "Please validate the output from terraform fmt and then git add it to your staged changes."
```

There's a subtle bug here: if I commit non-compliant code that triggers this, it will format my code thus making it compliant. But, the formatted - and thus compliant - code is not yet staged. However, when I try to recommit - without doing _any_thing else, the hook will now pass because it checks the content of the files listed in your staging as it is in your working directory.

So I think it may make more sense to NOT try to format for the user and tell them to do the formatting (and then any subsequently required staging/git adding) themselves.

My suggestion would thus actually be to NOT try to format for the user!

.environment/terraform-fmt/run-terraform-fmt.sh
```diff
@@ -0,0 +1,47 @@
...
+        make -C "${REPO_ROOT?}/operations" -f "${REPO_ROOT?}/operations/Makefile" tf-cmd TF_CMD="terraform fmt -recursive /app/src" >> "${REPO_ROOT?}/${LOGFILE?}" 2>&1
+        return $?
+    fi
+    return 0
+}
+
+terraform_check_fmt
+RC=$?
+
+terraform_fmt
```
td-usds, line 38: Would it make more sense to condition execution of `terraform_fmt` on the value of `${RC?}` (and then _not_ do the check inside `terraform_fmt`, thus making it callable from anywhere without having to set the `${TERRAFORM_NEEDS_FMT}=1`)  or is there another reason why `terraform_fmt` is the one doing the check?

```suggestion
if [[ ${RC?} != 0 ]]; then
    terraform_fmt
fi
```

.environment/terraform-fmt/run-terraform-fmt.sh
```diff
@@ -0,0 +1,47 @@
...
+
+function terraform_check_fmt() {
+    note "Checking Terraform formatting."
+    make -C "${REPO_ROOT?}/operations" -f "${REPO_ROOT?}/operations/Makefile" tf-cmd TF_CMD="terraform fmt -check -recursive /app/src" > "${REPO_ROOT?}/${LOGFILE?}" 2>&1
+    TERRAFORM_NEEDS_FMT=$?
+    return ${TERRAFORM_NEEDS_FMT?}
+}
+
+function terraform_fmt() {
+    if [[ ${TERRAFORM_NEEDS_FMT?} != 0 ]]; then
```
td-usds, line 27: See comment above to move the check outside of this function

operations/app/src/modules/front_door/variables.tf
```diff
@@ -1,29 +1,29 @@
...
 
 variable "location" {
-  type = string
+  type        = string
   description = "Function App Location"
 }
 
 variable "https_cert_names" {
-  type = list
+  type        = list(any)
```
td-usds, line 22: Is this supposed to be `any`? Should it be `object` or `string` instead?

operations/app/src/modules/sftp_container/main.tf
```diff
@@ -53,7 +53,7 @@ resource "azurerm_container_group" "sftp_container" {
     // Workaround. TF thinks this is a new resource after import
     ignore_changes = [
       network_profile_id,
-      container[0].volume[0]]
+    container[0].volume[0]]
```
td-usds, line 56: This is a weird formatting it did :/

operations/app/src/modules/key_vault/main.tf
```diff
@@ -1,234 +1,234 @@
...
-    secret_permissions      = [
-        "Get"]
-    certificate_permissions = [
-        "Get"]
+  key_vault_id = azurerm_key_vault.application.id
+  tenant_id    = data.azurerm_client_config.current.tenant_id
+  object_id    = local.frontdoor_object_id
+
+  secret_permissions = [
+  "Get"]
```
td-usds, line 85: What does the formatting do if you make this:
```hcl
secret_permissions = [
  "Get"
]
```
??

That might be a better way to do lists.

td-usds (changes requested):
There's a subtle bug here: if I try to commit non-compliant code that triggers this, it will format my code for me ('magic') thus making it compliant and then block the commit from going in.

This is cool, but...

The formatted - and thus compliant - code is not yet staged. However, when I try to recommit without doing _any_thing else, the hook will now pass because it checks the content of the files listed in your staging as it is in your working directory. This is because the files checked are now compliant, but that's not what'll be committed. What will be committed is what is staged. It's the same 'set of file names' but not 'the same content'!

So I think it may make more sense to NOT try to format for the user and tell them to do the formatting (and then any subsequently required staging/git adding) themselves.

My suggestion would thus actually be to NOT try to format for the user!

.environment/terraform-fmt/run-terraform-fmt.sh
ronaldheft-usds: Agreed. Updated the hook to not automatically format.

.environment/terraform-fmt/run-terraform-fmt.sh
ronaldheft-usds: Refactored the code so this is no longer an issue.

.environment/terraform-fmt/run-terraform-fmt.sh
ronaldheft-usds: Refactored the code so this is no longer an issue.

td-usds (approved):

.editorconfig
```diff
@@ -6,5 +6,5 @@ indent_style = space
 insert_final_newline = false
 tab_width = 4
 
-[{*.yaml,*.yml}]
+[{*.yaml,*.yml, *.tf}]
```
ronaldheft-usds, line 9: This matches what `terraform fmt` does.

operations/app/src/modules/front_door/variables.tf
ronaldheft-usds: Fixed

operations/app/src/modules/key_vault/main.tf
ronaldheft-usds: Agreed, I fixed up all of these instances throughout the formatted code and ensured the last item has a trailing `,`.

ronaldheft-usds:
@td-usds Ready for review again. I took your suggestions, cleaned up the auto-formatting, and added a `.editorconfig` adjustment to help IDEs match the formatting.

td-usds (approved):

Issue #1916, opened 2021-08-20
New settings can make it difficult/impossible to update settings


RickHawesUSDS:
**Describe the bug**
I was unable to change settings on August 10th. 

**Impact**
I was not able to make changes to production settings until the master branch was rolled to production. 

**To Reproduce**
1. Add a new setting to a receiver (or any other entity)
2. Merge to master
3. Test the new setting in staging <all good> with a `prime receiver set` command
4. Test the new setting in prod: <error>
5. Merge new setting to the production branch and deploy 
6. Test the new setting in prod: <all good>

**Expected behavior**
`prime receiver set` should always work in production and staging even when they are running different versions of the software. 

**Additional context**
<img width="633" alt="Screen Shot 2021-08-20 at 4 02 00 PM" src="https://user-images.githubusercontent.com/48692522/130301174-9349ba40-a52f-42e8-80fc-faf68ba60f14.png">

PR #1919, opened 2021-08-21
Fix in the settings API to handle unknown properties

RickHawesUSDS:
This PR fixes a bug where the settings features would not work in production. This situtation occurs when new settings properties are introduced, but not yet deployed to production. 

## Changes
- The json deserializer ignores unknown properties in the API and CLI

### Testing
- [x] Tested locally?
- [x] Ran `quickTest all`?
- [ ] Added tests?

    - new endpoint A
    - new endpoint B

### Process

## Fixes
- #1916

## To Be Done
- #issue

### DIFF
--- a/prime-router/src/main/kotlin/azure/SettingsFacade.kt
+++ b/prime-router/src/main/kotlin/azure/SettingsFacade.kt
@@ -39,7 +39,7 @@ class SettingsFacade(
         // Format OffsetDateTime as an ISO string
         mapper.registerModule(JavaTimeModule())
         mapper.disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS)
-        // Unknown properties in the json payload may be a new property defined in a new version of the CLI of the code
+        // Unknown properties in the payload may be caused by new versions of the CLI of the code
         mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
     }
 
--- a/prime-router/src/main/kotlin/azure/SettingsFacade.kt
+++ b/prime-router/src/main/kotlin/azure/SettingsFacade.kt
@@ -2,6 +2,7 @@ package gov.cdc.prime.router.azure
 
 import com.fasterxml.jackson.annotation.JsonCreator
 import com.fasterxml.jackson.annotation.JsonIgnore
+import com.fasterxml.jackson.databind.DeserializationFeature
 import com.fasterxml.jackson.databind.SerializationFeature
 import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule
 import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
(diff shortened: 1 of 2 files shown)


jimduff-usds (commented):
Because this will hide errors, I worry this could introduce as many problems as it solves. Could a submission from CLI version X+1 break production code in version X, by submitting data with missing or extra fields?     Another problem is that someone updates the .kt obj structures, but forgets to update the actual json data in Prod.   This solution might hide problems, but then they pop up hours or days later when that setting gets used?

Another solution might be to provide better / more complete error messaging about what's going on in these version mismatch situations. 

As a start in that direction, note that the three settings 'set' commands have a --verbose option that at least helps to expose the data being passed from the cli to the API.   Better would be to show the user the difference in the object versions - these errors can be quite perplexing when you first encounter them.

RickHawesUSDS:
@jimduff-usds 

> * Someone might submit extra new fields from a newer version of the CLI, into an older prod env, expecting those fields to take effect in production, and in fact they are silently ignored.   Is that a problem?

I think it is less of a problem than not being able to use the current CLI in production. 

> * Is this same deserialization config option used in the env when json is pulled in from the database?   If so, then another problem is that someone updates the .kt obj structures, but forgets to update the actual json data in Prod.   This solution might hide problems, but then they pop up hours or days later when that setting gets used.
> 

Today devs will only add fields to our structures because of this strict deserializer. For example, consider that our Hl7Configuration has two fields to control processing mode. Which should a dev use? One is deprecated and one is not, but you have to read the code comments to know. This wouldn't happen if the developer could remove fields. (I know that someone could write a migration to remove that field in the settings table json field, but this is not a trivial effort and not obvious to most people). I can't see that this is a better situation. 

> Another solution might be to provide better / more complete error messaging about what's going on in these version mismatch situations.
> 
Again, it doesn't work for the removing fields situation. 

> As a start in that direction, note that the three settings 'set' commands have a --verbose option that at least helps to expose the data being passed from the CLI to the API. Better would be to show the user the difference in the object versions - these errors can be quite perplexing when you first encounter them.
> 

> * It would be good to see a test wherein the CLI does a settings set with extra fields, missing fields.
Do you mean that `./prime test` would do a few settings?

prime-router/src/main/kotlin/azure/SettingsFacade.kt
```diff
@@ -38,6 +39,8 @@ class SettingsFacade(
         // Format OffsetDateTime as an ISO string
         mapper.registerModule(JavaTimeModule())
         mapper.disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS)
+        // Unknown properties in the payload may be caused by new versions of the CLI of the code
+        mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
```
clediggins-usds, line 43: Perhaps we should add a warning if we attempt to deserialize something with unknown properties.  Seems like the schema/data structures could get a bit out of hand if we never backfill / repair older records.

jimduff-usds (changes requested):
I agree with @clediggins-usds that this needs warnings or something like that - can we add those?

Using this PR, I submitted a Receiver setting with every single translation field mis-spelled, and it gleefully accepted it with no errors or warnings.  Then, when you pull it out, it has default values set for all those fields, which is confusing and very easy to not even notice.

Issue #1902, opened 2021-08-19
HL7 batch ingest thinks there is an empty HL7 message


carlosfelix2:
HL7 batch ingest thinks there is an empty HL7 message.  The warning "Cannot parse empty HL7 message" gets added to responses to any submitted batch message.

Acceptance criteria:
1. Submitting a correctly formatted batch payload shall not have a warning of "Cannot parse empty HL7 message".
2. An HL7 batch payload with no HL7 message shall have a warning of "Cannot parse empty HL7 message".

PR #1923, opened 2021-08-23
1902 Do not process blank strings generated from batch submission headers

carlosfelix2:
This PR makes sure the HL7 serializer is not generating a false warning about processing an empty message for batch payloads.  The problem was that the serializer would loop through the initial batch segments and generate a string with "\r\r" due to the appending of \r to the nextMessage variable.  The serializer would then generate a warning that it could not parse an empty message when it encountered the first true HL7 message of the batch message.  

## Changes
- Not parse a blank string.

### Testing
- [x] Tested locally?
- [ ] Ran `quickTest all`?
- [ ] Added tests?

### Process

## Fixes
- #1902


jimduff-usds (approved):
lgtm

Issue #1748, opened 2021-07-29
Add preferred language to the Simple Report schema


jimduff-usds:
On July 29, SimpleReport team gave us a request to handle preferred language.

Santa Clara requested this: patient preferred language.  Rick says there should be an HL7 standard for this.

@nullflux is technical Point of Contact; @aliciabeckett-gov is product owner on SimpleReport.

This ticket is done when patient preferred language is added as a column in the simple report schema, and it meets their needs.

See related ticket #1746

whytheplatypus:
Possible value set to use: http://hl7.org/fhir/R4/valueset-languages.html

aliciabeckett-gov:
This value set does not support all SimpleReport languages :*( For example, Amaric, Creole, and Farsi are all missing from the FHIR standard list. 

I ended up creating a custom list because NONE of the standards met the needs of CA's list of languages people actually speak. My recommendation is pass through the language as plain text and not a code, or ask CA how they do it in their other ELR feeds.

whytheplatypus:
Understood, so this is aiming to meet the needs of CA first and foremost?

It sounds like starting with a plain string is a reasonable first iteration.

whytheplatypus:
It looks like CA has a [`local_language`](https://github.com/CDCgov/prime-reportstream/blob/master/prime-router/metadata/schemas/CA/ca-scc-covid-19.schema#L70-L74) in their schema, but I'm assuming _this_ language field is something new?

For input I'm guessing it would be submitted in the CSV under something like `Preferred_language`?

aliciabeckett-gov:
@nullflux is the person to work with on adding preferred_language to the SimpleReport schema. We should be able to add that and Tribal Affiliation quickly, and already have those tickets in our backlog.

https://app.zenhub.com/workspaces/simplereport-better-data-team-605b43d51ea9f700119a48ee/issues/cdcgov/prime-simplereport/1799

PR #1935, opened 2021-08-23
Include preferred language in pdi schema

whytheplatypus:
This PR Adds the patients preferred language as a text field to the SimpleReport schema.

## Changes
- Creates a TEXT type entry in the simple report schema named `patient_preferred_language`. This will be populated from the csv header `Patient_preferred_language`

### Testing
- [ ] Tested locally?
- [ ] Ran `quickTest all`?

### Process

## Fixes
- #1748

### DIFF
--- a/prime-router/docs/schema_documentation/primedatainput-pdi-covid-19.md
+++ b/prime-router/docs/schema_documentation/primedatainput-pdi-covid-19.md
@@ -686,6 +686,20 @@ The patient's phone number with area code
 
 ---
 
+**Name**: Patient_preferred_language
+
+**Type**: TEXT
+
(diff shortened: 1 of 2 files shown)


MauriceReeves-usds (changes requested):
David, while this change looks very straightforward, there are several places where the primedatainput schema is used for testing, and could potentially error out because this field is not added to the expected comparison files. In particular, look in `testIntegration/resources` folder, you'll see a CSV file which outlines multiple tests to run and files to use for comparison, and in those cases, the integration test will now fail. Review the `readme` in that folder, and ping me if you have any questions.

PR #1940, opened 2021-08-23
Modified Smoke Test so it works in Staging

jimduff-usds:
This PR makes small tweaks to our standard smoke test so it works in Staging.

## Changes
- Removed watersauth test from the smoke test suite, since it currently requires a human in the loop to login.  The test is still available and can always be run manually, eg, `./prime --run watersauth`
- Redox tests no longer run in staging.   They always failed in staging, because we don't have a mock redox running there.  So I just removed them.  Redox tests still run locally.

### Testing
- [X] Tested locally?
- [X] Ran `quickTest all`?
- [ ] Added tests?

_no new risks._

    - new endpoint A
    - new endpoint B

### Process

### DIFF
--- a/prime-router/docs/releases/2021-08-24-release-notes.md
+++ b/prime-router/docs/releases/2021-08-24-release-notes.md
@@ -1,0 +1,17 @@
+#  ReportStream Release Notes
+
+*August 17, 2021*
+
+## General useful links:
+
+- All Schemas are documented here:  [Link to detailed schema dictionaries](../schema_documentation)
+- The Hub API is documented here: [Hub OpenApi Spec](../openapi.yml)
+- [Click here for all release notes](../releases)
+
+## For this release
+
+### Changes for Specific Senders
+
+- Data from Reddy Health is now flowing to HHSProtect
+- Patient_tribal_citizenship has been added to the Simple Report schema
+
--- a/prime-router/src/main/kotlin/cli/tests/Hl7Tests.kt
+++ b/prime-router/src/main/kotlin/cli/tests/Hl7Tests.kt
@@ -18,6 +18,7 @@ class Hl7Ingest : CoolTest() {
     override val status = TestStatus.SMOKE
 
     override fun run(environment: ReportStreamEnv, options: CoolTestOptions): Boolean {
+        initListOfGoodReceiversAndCounties(environment)
         var passed = true
         val sender = hl7Sender
         val receivers = allGoodReceivers
--- a/prime-router/src/main/kotlin/cli/tests/TestReportStream.kt
+++ b/prime-router/src/main/kotlin/cli/tests/TestReportStream.kt
@@ -453,21 +453,23 @@ abstract class CoolTest {
         val hl7Sender = settings.findSender("$orgName.$hl7SenderName")
             ?: error("Unable to find sender $hl7SenderName for organization ${org.name}")
 
-        val allGoodReceivers = settings.receivers.filter {
-            it.organizationName == orgName &&
-                !it.name.contains("FAIL") &&
-                !it.name.contains("BLOBSTORE") &&
-                !it.name.contains("QUALITY") &&
-                !it.name.contains("AS2") &&
-                !it.name.contains("OTC")
-        }
-        val allGoodCounties = allGoodReceivers.map { it.name }.joinToString(",")
+        val csvReceiver = settings.receivers.filter { it.organizationName == orgName && it.name == "CSV" }[0]
+        val hl7Receiver = settings.receivers.filter { it.organizationName == orgName && it.name == "HL7" }[0]
+        val hl7BatchReceiver = settings.receivers.filter { it.organizationName == orgName && it.name == "HL7_BATCH" }[0]
+        val redoxReceiver = settings.receivers.filter { it.organizationName == orgName && it.name == "REDOX" }[0]
+        val hl7NullReceiver = settings.receivers.filter { it.organizationName == orgName && it.name == "HL7_NULL" }[0]
 
-        val csvReceiver = allGoodReceivers.filter { it.name == "CSV" }[0]
-        val hl7Receiver = allGoodReceivers.filter { it.name == "HL7" }[0]
-        val hl7BatchReceiver = allGoodReceivers.filter { it.name == "HL7_BATCH" }[0]
-        val redoxReceiver = allGoodReceivers.filter { it.name == "REDOX" }[0]
-        val hl7NullReceiver = allGoodReceivers.filter { it.name == "HL7_NULL" }[0]
+        lateinit var allGoodReceivers:  List<Receiver>
+        lateinit var allGoodCounties: String
+
+        fun initListOfGoodReceiversAndCounties(env: ReportStreamEnv) {
+           allGoodReceivers = if (env == ReportStreamEnv.LOCAL)
+               listOf(csvReceiver, hl7Receiver, hl7BatchReceiver, redoxReceiver, hl7NullReceiver)
+            else
+                listOf(csvReceiver, hl7Receiver, hl7BatchReceiver, hl7NullReceiver)
+            allGoodCounties = allGoodReceivers.map { it.name }.joinToString(",")
+        }
+
         val blobstoreReceiver = settings.receivers.filter {
             it.organizationName == orgName && it.name == "BLOBSTORE"
         }[0]
@@ -516,10 +518,12 @@ abstract class CoolTest {
             val secsElapsed = OffsetDateTime.now().second % 60
             // Wait until the top of the next minute, and pluSecs more, for 'batch', and 'send' to finish.
             var waitSecs = 60 - secsElapsed + plusSecs
-            if (secsElapsed > (60 - plusSecs) || env != ReportStreamEnv.LOCAL) {
+            if (env != ReportStreamEnv.LOCAL) {
+                // We are in Test or Staging, which don't execute on the top of the minute. Hack:
+                waitSecs += 120
+            } else if (secsElapsed > (60 - plusSecs)) {
                 // Uh oh, we are close to the top of the minute *now*, so 'receive' might not finish in time.
-                // Or, we are in Test or Staging, which don't execute on the top of the minute.
-                waitSecs += 90
+                waitSecs += 20
             }
             echo("Waiting $waitSecs seconds for ReportStream to fully receive, batch, and send the data")
             for (i in 1..waitSecs) {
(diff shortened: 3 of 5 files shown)


prime-router/docs/releases/2021-08-24-release-notes.md
```diff
@@ -0,0 +1,17 @@
+#  ReportStream Release Notes
+
+*August 17, 2021*
```
jimduff-usds, line 3:
```suggestion
*August 24, 2021*
```

prime-router/src/main/kotlin/cli/tests/Hl7Tests.kt
```diff
@@ -18,6 +18,7 @@ class Hl7Ingest : CoolTest() {
     override val status = TestStatus.SMOKE
 
     override fun run(environment: ReportStreamEnv, options: CoolTestOptions): Boolean {
+        initListOfGoodReceiversAndCounties(environment)
```
jimduff-usds, line 21: List of receivers to test is now dependent on the Environment.  If not in staging, include REDOX.

prime-router/src/main/kotlin/cli/tests/TestReportStream.kt
```diff
@@ -453,21 +453,23 @@ abstract class CoolTest {
...
-                !it.name.contains("OTC")
-        }
-        val allGoodCounties = allGoodReceivers.map { it.name }.joinToString(",")
-
-        val csvReceiver = allGoodReceivers.filter { it.name == "CSV" }[0]
-        val hl7Receiver = allGoodReceivers.filter { it.name == "HL7" }[0]
-        val hl7BatchReceiver = allGoodReceivers.filter { it.name == "HL7_BATCH" }[0]
-        val redoxReceiver = allGoodReceivers.filter { it.name == "REDOX" }[0]
-        val hl7NullReceiver = allGoodReceivers.filter { it.name == "HL7_NULL" }[0]
+        val csvReceiver = settings.receivers.filter { it.organizationName == orgName && it.name == "CSV" }[0]
```
td-usds, line 456: Should this make sure that there is at least 1 result in the filtered sequence, since we're doing `[0]`? Maybe something like (make the message better though):
```kotlin
something.filter{it.name == "selector"}.elementAtOrElse(0) { index -> assert(false, "Element expected, but not found") }
```
It can assert or log, but _something_ that makes it blatantly clear from the output/log of "this is why this test failed" to make it easy to debug/remediate it.

Same for the lines below

prime-router/src/main/kotlin/cli/tests/TestReportStream.kt
```diff
@@ -453,21 +453,23 @@ abstract class CoolTest {
...
+        val hl7Receiver = settings.receivers.filter { it.organizationName == orgName && it.name == "HL7" }[0]
+        val hl7BatchReceiver = settings.receivers.filter { it.organizationName == orgName && it.name == "HL7_BATCH" }[0]
+        val redoxReceiver = settings.receivers.filter { it.organizationName == orgName && it.name == "REDOX" }[0]
+        val hl7NullReceiver = settings.receivers.filter { it.organizationName == orgName && it.name == "HL7_NULL" }[0]
+
+        lateinit var allGoodReceivers:  List<Receiver>
+        lateinit var allGoodCounties: String
+
+        fun initListOfGoodReceiversAndCounties(env: ReportStreamEnv) {
```
td-usds, line 465: Consider this instead:
```kotlin
fun initListOfGoodReceiversAndCounties(env: ReportStreamEnv) {
    allGoodReceivers = mutableListOf(csvReceiver, hl7Receiver, hl7BatchReceiver, hl7NullReceiver)
    if (env == ReportStreamEnv.LOCAL){
        allGoodReceivers.add(redoxReceiver)
    }

    allGoodCounties = allGoodReceivers.map { it.name }.joinToString(",")
}
```

Increased readability, clearer "these are the defaults and this is what changes in `LOCAL`"
Also, L466 will fail KTLint (indentation!)

prime-router/src/main/kotlin/cli/tests/TestReportStream.kt
```diff
@@ -453,21 +453,23 @@ abstract class CoolTest {
...
-        }
-        val allGoodCounties = allGoodReceivers.map { it.name }.joinToString(",")
-
-        val csvReceiver = allGoodReceivers.filter { it.name == "CSV" }[0]
-        val hl7Receiver = allGoodReceivers.filter { it.name == "HL7" }[0]
-        val hl7BatchReceiver = allGoodReceivers.filter { it.name == "HL7_BATCH" }[0]
-        val redoxReceiver = allGoodReceivers.filter { it.name == "REDOX" }[0]
-        val hl7NullReceiver = allGoodReceivers.filter { it.name == "HL7_NULL" }[0]
+        val csvReceiver = settings.receivers.filter { it.organizationName == orgName && it.name == "CSV" }[0]
+        val hl7Receiver = settings.receivers.filter { it.organizationName == orgName && it.name == "HL7" }[0]
```
td-usds, line 457: There's a lot of cases of `settings.receivers.filter { it.organizationName == orgName }` happening; this seems like something that can be cached or centralized or refactored better.

prime-router/src/main/kotlin/cli/tests/TestReportStream.kt
```diff
@@ -516,10 +518,12 @@ abstract class CoolTest {
             val secsElapsed = OffsetDateTime.now().second % 60
             // Wait until the top of the next minute, and pluSecs more, for 'batch', and 'send' to finish.
             var waitSecs = 60 - secsElapsed + plusSecs
-            if (secsElapsed > (60 - plusSecs) || env != ReportStreamEnv.LOCAL) {
+            if (env != ReportStreamEnv.LOCAL) {
+                // We are in Test or Staging, which don't execute on the top of the minute. Hack:
```
td-usds, line 522: Instead of adding `Hack:`, consider using `HACKHACK`; this ensures that when you grep for hacks (`HACKHACK`), lines that talk about "getting hacked:..." won't show up

It also means we can easily search for places where we have a 'hack' in place: `grep -Ri HACKHACK *.kt`

td-usds: And as an aside: what is the plan to get rid of this hack? It would make the tests run smoother, and faster, which I'm sure all developers would appreciate;

Can you file a GitHub Issue to address this hack?

prime-router/src/main/kotlin/cli/tests/TestReportStream.kt
```diff
@@ -640,6 +644,7 @@ class End2End : CoolTest() {
     override val status = TestStatus.SMOKE
 
     override fun run(environment: ReportStreamEnv, options: CoolTestOptions): Boolean {
+        initListOfGoodReceiversAndCounties(environment)
         var passed = true
```
td-usds, line 648: I know this is not part of this PR, but why do we assume tests pass? Should this not be `false` and set explicitly to true in the else branch that starts on L667?

prime-router/src/main/kotlin/cli/tests/TestReportStream.kt
```diff
@@ -841,6 +845,7 @@ class Strac : CoolTest() {
     override val status = TestStatus.SMOKE
 
     override fun run(environment: ReportStreamEnv, options: CoolTestOptions): Boolean {
+        initListOfGoodReceiversAndCounties(environment)
         ugly("Starting bigly strac Test: sending Strac data to all of these receivers: $allGoodCounties!")
         var passed = true
```
td-usds, line 850: See above about assuming success; this seems like a broader issue - can you file a GitHub issue to revisit this?

td-usds (changes requested):

prime-router/src/main/kotlin/cli/tests/TestReportStream.kt
carlosfelix2, line 465: Why have the function that needs to be called in all the test runs that need it?  I think you could just remove the function and keep all the code in the companion object.

carlosfelix2 (approved):

prime-router/src/main/kotlin/cli/tests/TestReportStream.kt
jimduff-usds: Yes, if this were not test code, I would definitely do this.

prime-router/src/main/kotlin/cli/tests/TestReportStream.kt
jimduff-usds: Changed.

prime-router/src/main/kotlin/cli/tests/TestReportStream.kt
jimduff-usds: Better yet, I have plans to get rid of these completely, and have these settings created on the fly.  This will enable us to do clever things like running multiple tests in parallel, which can't be done now, as they interfere with each other.  See #1942, for example.

prime-router/src/main/kotlin/cli/tests/TestReportStream.kt
jimduff-usds: Changed to 'kluge' ;) 
And good point - I forgot to ticket this.  See #1954

prime-router/src/main/kotlin/cli/tests/TestReportStream.kt
jimduff-usds: ok for now as test code.

prime-router/src/main/kotlin/cli/tests/TestReportStream.kt
jimduff-usds: Yeah, that's what I was thinking too, but because it depends on the environment value, its no longer really a static object.   There 's a kosher way to do this, which is to do this in an` init { }` clause in `CoolTest`, but that would have involved adding a constructor to pass in the Env arg and changing every single test's constructor call.  That was messier than what I ended up doing.

prime-router/src/main/kotlin/cli/tests/TestReportStream.kt
jimduff-usds: Since any one failure should fail the whole test, the most important logic is to make sure you don't overwrite an early  failure with a success later in the test.   Setting it to true once only, at the start, and then all other sets are to false, is a way to minimize (but not eliminate) the chance of a developer messing that up.

prime-router/src/main/kotlin/cli/tests/TestReportStream.kt
td-usds: This pleases me :)

td-usds (approved):
I can dig this

PR #1978, opened 2021-08-25
Fix for #1977: Only run terraform format check hook when there are terraform changes

### LINKED ISSUE #1977: terraform formatting git hook runs even when no terraform files were touched
Currently, the Terraform Formatting git hook runs at every commit, regardless of whether or not you touched terraform files. This is causing some issues:
- It makes commits slower
- It depends on `make` and not everyone has (nor should have) this installed. Those who do not have this installed can no longer commit.

Suggested solution:
- Terraform format (as invoked by the git hook, _not_ in the nightly run) should only run on top of files that are either modified or staged (I can live with the 'inconvenience' of that level of ambiguity).

td-usds:
This fixes #1977 by only running the terraform hook if `git status --porcelain` reports files with `.tf` extension. If no such files come back from that command, then the hook tells you that it will skip the check (and reports success). If such files _were_ changed then the script will report the return code from the (containerized) `terraform fmt` invocation as it did before.

This closes #1977

### DIFF
--- a/.environment/terraform-fmt/run-terraform-fmt.sh
+++ b/.environment/terraform-fmt/run-terraform-fmt.sh
@@ -38,13 +38,22 @@ LOGFILE="terraform-fmt.log"
 
 function terraform_fmt_check() {
     note "Checking Terraform formatting."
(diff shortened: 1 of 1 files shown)


carlosfelix2 (approved):
Tested locally with and without changes to terraform files

Issue #1960, opened 2021-08-24
Change CLI tests to run in parallel


jimduff-usds:
Change CLI tests to run in parallel to save time running all the smoke tests.

**Design notes:**
Replace all calls to `reportLineageCountQuery()` with calls to `itemLineageCountQuery()`.  They don't work quite the same, so you might need to rejigger things a bit.

- Please delete `reportLineageCountQuery()` as part of this.

Any test that uses reportLineageCountQuery could deliver wrong results when the tests are run in parallel.

Item lineages are cleanly immune to merging - individual items can only descend from a single originally submitted parent.
This is quite unlike reports as a whole (the REPORT_FILE table) which can contain data from many parents, so the `report_file.item_count` field data gets all messed up and confused when unwanted merging happens.

See also #1954 and related tickets.

Acceptance criteria:
1. CLI tests can run in parallel.
2. Output of CLI tests that run in parallel is shown separate from other tests (do not jumble the output to the screen)
3. Allow user to specify via command line argument to run the tests one at a time.
4. All tests run in parallel are success and do not interfere with other tests.

PR #1984, opened 2021-08-25
1960 Run CLI tests in parallel

carlosfelix2:
This PR adds the ability for the PRIME CLI to run tests in parallel.  Understand that this uses co-routines and not threads, so tests will run as Kotlin determines resources are available without incurring into a lot of overhead, so you will see a few tests run and as they finish more will run.

To test:
1. Run the smoke test by running `./prime test` or `gradle testSmoke` and verify the tests run in parallel.
2. Verify the merge test is successful.
3. Run the smoke test in sequence by running `./prime test --sequential` and verify the test runs in sequence.
4. Run the repeat waters test by running `./prime test --run=repeatwaters` and verify the test runs.

## Changes
- Changed the merge test to rely on report IDs, so the test can run with others at the same time
- Tweaked the repeatWaters test to allow for its multi-threading
- Changes the wait function to use delay, which suspends the running of the coroutine
- Hold onto any messages to be printed by a test until the test is done, that way the output is organized
- Added an option to run the tests in sequence which prints the test output as it happens

### Testing
- [ ] Tested locally?
- [ ] Ran `quickTest all`?
- [ ] Added tests?

### Process

## Fixes
- #1960


jimduff-usds (approved):
I ran it 3 times locally and twice on Staging.

Locally, it worked sometimes, then I got timeout errors sometimes, especially if RS containers were just started afresh.  I think some of the "waitABit" times are just a bit too slow for my Mac, maybe.  Solution is to implement replace the sleeps() with polling, which we have a ticket for already.

I only timed it on Staging - took 4 minutes.   :) :) :)

Issue #1095, opened 2021-05-17
HL7 serializer should reject any unsupported message types


carlosfelix2:
The HL7 serializer currently only supports ORU R01 messages.  Need to reject any other unsupported messages.  We also need to keep in mind that we will be supporting more types as we progress, so make this rejection flexible to accept other message types in the future.

Acceptance criteria:
1. ReportStream rejects received HL7 messages that are not of type ORU R01.

PR #2015, opened 2021-08-27
1095 HL7 Ingest reject non ORU R01 messages

carlosfelix2:
This PR adds a check to make sure we only process ORU R01 HL7 messages.  Any message that is not ORU R01 is ignored with a warning.  A warning instead of an error allows us to accept a mix of HL7 messages while we add support for other types.  Also added the report number where the message or error is for.

How does this work?  I am using the PreParser from HAPI to get the message structure as it simple parses a string with the HL7 delimiters.  Note that the message structure is a required field.  If I try to use the regular parser it can throw an exception as some message types are not supported by that parser.  Furthermore, the exception thrown by the regular parser will not show if it was some other type.

Test:
1. Send an HL7 batch with multiple ORU-R01 messages and at least one non ORU-R01.  Verify that verify you get a warning plus reportItemCount does not include the non ORU-R01 message.  Also verify the report number is shown in the warning.
2. Send an HL7 message without the MSH-9-1 field.  Verify you get an error due to the missing field.  MSH-9 fields are required in HL7.

## Changes
- Ignore non ORU-R01 messages
- Added report number to warnings and errors
- Updated Gradle wrapper to 7.2

### Testing
- [x] Tested locally?
- [ ] Ran `quickTest all`?
- [ ] Added tests?

### Process

## Fixes
- #1095


MauriceReeves-usds (commented):
lgtm, but I am going to want us to start accepting ACK/NACK soon, so we will need to revisit this in the very near future. :-D

MauriceReeves-usds (approved):

carlosfelix2:
> lgtm, but I am going to want us to start accepting ACK/NACK soon, so we will need to revisit this in the very near future. :-D

Yes.  I made such that we could process different type of messages from that when statement by just adding a new check.

Issue #1827, opened 2021-08-10
Identify performance bottlenecks when submitting data to the report API


carlosfelix2:
Identify performance bottlenecks when submitting data to the report API

Acceptance criteria:
1. Identify where the bottlenecks are for the reports API.

carlosfelix2:
Had a good one hour meeting with @Ron Heft and we looked at many different data points, but no smoking gun.  Ron will be looking at AppInsights a bit more to try to get more data, and we may need to play with settings and code on a separate instance to try to get to the bottom of this.  We will reconvene after we can look into AppInsights a bit more to continue the investigation and figure out next steps.

carlosfelix2:
[https://app.zenhub.com/files/304423150/70b3f74c-b51d-4dac-b887-51a59a68e6f5/download](https://app.zenhub.com/files/304423150/70b3f74c-b51d-4dac-b887-51a59a68e6f5/download)

PR #2046, opened 2021-08-31
1827 Turn on App Insights profiling

ronaldheft-usds:
This PR flips some Function App settings to enable App Insights profiling.

Reference:
* https://docs.microsoft.com/en-us/azure/azure-monitor/app/monitor-functions#how-to-enable-distributed-tracing-for-java-function-apps
* https://docs.microsoft.com/en-us/azure/azure-monitor/app/java-in-process-agent
* https://docs.microsoft.com/en-us/azure/azure-functions/functions-host-json#applicationinsights

Related to #1827

### DIFF
--- a/operations/app/src/modules/function_app/main.tf
+++ b/operations/app/src/modules/function_app/main.tf
@@ -35,7 +35,19 @@ locals {
 
     "WEBSITES_ENABLE_APP_SERVICE_STORAGE" = false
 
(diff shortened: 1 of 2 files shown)


carlosfelix2 (approved):
I reviewed the documentation and this looks good.  I see no issues on these changes also deploying to production based on the data we have collected so far and the low use of resources by our functions, although we can test this in staging.  This can provide good data for us to continually improve the software and can disable it in the future if we deem it to be resource intensive.

PR #2051, opened 2021-09-01
ReactJs add "content security policy" to main index.html page

TomNUSDS:
## Changes
- bump `@cdc/map` to "^2.6.2"
- Added `.env` files (similar to what ReportStream uses).
NEED TO VERIFY production `REACT_APP_BACKEND_URL`
- Add `Content-Security-Policy` to index.html
  attempt to configure it correctly to allow access to the things we need.
- Update some of the default template names to something that makes more sense
- Update reference to cdc image to be on the server, not remote (in oktaSigninConfig)
- Fix scss `u-text()` in `index.css` to be straight css.
- Fix react console warning about not including a key for list of states.

TODO: verify what the long and short product names are.

### Testing
- [x] Tested locally? 
> Tested locally (localhost debug)
> Build release and test that locally (still localhost). 
- [ ] Added tests?
> Interesting idea. Is there a test we can write to verify Content-Security-Policy loads content correctly AND fails correctly?
 
This add `Content-Security-Policy`, so it's very likely to break things that will require tweaking the spec.

```html
    <meta
      http-equiv="Content-Security-Policy"
      content="default-src 'self';
        script-src 'self'
          https://hss-prime-okta.com
          https://global.oktacdn.com
          https://www.google-analytics.com
          https://*.in.applicationinsights.azure.com;
        style-src 'self' 'unsafe-inline'
          https://global.oktacdn.com
          https://cdnjs.cloudflare.com;
        frame-src 'self' https://hhs-prime.okta.com;
        img-src 'self' https://hhs-prime.okta.com https://reportstream.cdc.gov/ data: ;
        connect-src 'self'
          https://www.google-analytics.com
          https://*.in.applicationinsights.azure.com
          https://hhs-prime.okta.com
          %REACT_APP_BACKEND_URL% ;"
    />
```

`https://cdnjs.cloudflare.com` may not be required, but it's only allowed to load styles, so fairly low risk.
`%REACT_APP_BACKEND_URL%` come from the .env and is `https://reportstream.cdc.gov/`

NOTE:
`INLINE_RUNTIME_CHUNK` in .env removes the use of unsafe `eval()` in react. It's enabled for dev and disabled for prod
See: SimpleReport 
https://github.com/CDCgov/prime-simplereport/blob/main/frontend/public/index.html 
https://github.com/CDCgov/prime-simplereport/blob/main/frontend/.env
https://github.com/CDCgov/prime-simplereport/blob/main/frontend/.env.development

- What are the potential security threats and mitigations?
> This change improves security. 

## To Be Done
- Deploy to staging and verify everything loads by looking at network requests. This has the potential to break the site.
- Verify `https://*.in.applicationinsights.azure.com` works for Azure Insights.
- Maybe disable `Content-Security-Policy` completely for dev? SimpleReport does this.
- Need to verify production's `REACT_APP_BACKEND_URL`


TomNUSDS:
Also, it looks like the `[Build Frontend React](https://github.com/CDCgov/prime-reportstream/runs/3480916987?check_suite_focus=true)` automated github action didn't run? Is this because the PR is in draft mode? It seems like it's a file filter issue?

Bisonburger (approved):
lgtm

kevinhaube (approved):
This looks good, thanks Tom!

PR #2057, opened 2021-09-01
Adds an alias for the QuickVue Rapids test kit

MauriceReeves-usds:
This PR adds an alias for the QuickVue Rapids test kit that doesn't currently exist in the LIVD table but SimpleReport is using as a test type.


jimduff-usds (commented):
I don't get it - is this valid device or not a valid device?

If its a valid device, why isn't it in the LIVD table?
if its not valid, why is SimpleReport sending it?


Also, there's a more recent LIVD table - why aren't we hacking that one?   LIVD-SARS-CoV-2-2021-08-11.csv

MauriceReeves-usds:
@jimduff-usds It's a valid device, I just think that SR is using the wrong name for it. 

I will be patching the new one as well, but that's not been merged in yet.

I need to get this in post haste because this test is being used in a lot of places and more states are starting to raise this as an issue.

jimduff-usds:
> @jimduff-usds It's a valid device, I just think that SR is using the wrong name for it.
> 
> I will be patching the new one as well, but that's not been merged in yet.
> 
> I need to get this in post haste because this test is being used in a lot of places and more states are starting to raise this as an issue.

Still not getting it - if SR is using the wrong name, then their data is invalid and we should not be sending it to states.   Its bad data.

jimduff-usds (approved):
Better would be to add a qualityFilter, since in this case SimpleReport is sending us an erroneous value.

PR #2058, opened 2021-09-01
Added rapids to the new LIVD table too

MauriceReeves-usds:
This PR adds the QuickVue Rapids test to the new LIVD table too


jorg3lopez (approved):
Looks good.

PR #2074, opened 2021-09-01
Quick fix to get staging working with `Content-Security-Policy`

TomNUSDS:
A quick fix to get staging working again (and to see if there are other URLs that break once this is fixed).

Always allows access to https://staging.prime.reportstream.cdc.gov/api even in production. 

This needs to be fixed in a better way, but having it be runtime based versus configuration based is problematic.

see `getBaseUrl()`

related to: #2051

### DIFF
--- a/frontend-react/public/index.html
+++ b/frontend-react/public/index.html
@@ -23,7 +23,9 @@
           https://www.google-analytics.com
           https://*.in.applicationinsights.azure.com
           https://hhs-prime.okta.com
-          %REACT_APP_BACKEND_URL% ;"
+          https://staging.prime.reportstream.cdc.gov/api
+          %REACT_APP_BACKEND_URL%
+          ;"
     />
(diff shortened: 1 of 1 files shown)


ahay-agile6 (approved):

TomNUSDS:
@mauricereeves Sure enough #2051 didn't work on staging. Can you merge this so it can get on staging? 

This is a bit of a hack to hard code the staging api server to the list. Naturally, this isn't ideal, but it is better than what we had before (No Content Security Policy directive) and allowing access from production->staging is better than the other way around.

TomNUSDS:
@mauricereeves Sure enough #2051 didn't work on staging. Can you merge this so it can get on staging? 

This is a bit of a hack to hard code the staging api server to the list. Naturally, this isn't ideal, but it is better than what we had before (No Content Security Policy directive) and allowing access from production->staging is better than the other way around.

Issue #1973, opened 2021-08-25
Stop calling a missing transport an error


jimduff-usds:
In the `action` table, if we don't send data to a particular state, we call that a `send_error `right now.

Its probably not an error - we have set up receivers for all 56 states, even though only about 20 or so are actually 'live' with us.   In fact, some states are live with us, but get their data via the download website, in which case we also may not have a transport set up for them- again, not an error.

Here's what the message looks like in the action table right now:
```
action_name: send_error
action_result: Not sending f1c1de41-5873-4b50-8768-ceb5ea557a61 to ny-phd.elr: No transports defined
```

### This ticket is DONE when:

- there is a new action_name, 'send_info", available in postgres
- The `No transports defined `action uses send_info instead of send_error.
- sftp and redox Retries (which say " Will retry sending report...") use send_info instead of send_error. 
- However, its very important that the final failure (which says "All retries failed") is a send_error.
- All other send_error cases have not changed - they are still using send_error.
- (Bonus points if you think of a better name than `send_info`)

No need to change the action_result message.

Note:  this is a database sql change, which is done via Flyway.  Check out earlier .sql updates to this field - its annoying - you have to change the type of the column (its used in two tables), then change the enum type, then change the type of the column back.  Be careful.

After your change is deployed to Staging, please submit some data that causes it to occur, and doublecheck the Metabae "Last 7 days" dashboard in Staging- there may be queries that need to be modified to know about this new action name.

anshulkumar-usds:
@jimduff-usds this is assigned to current sprint but no assignee. Is anyone working on this?

PR #2110, opened 2021-09-07
Adding send_info action type

whytheplatypus:
This PR ...

## Changes
- Migrate `task_action` type to include a `send_warning` value
- Use `send_warning` for action history in the case where a receiver has no transports defined.

### Testing
- [x] Tested locally?
- [x] Ran `quickTest all`?
- [x] Added tests?

## Fixes
- #1973 

Open review questions:
- Any better name suggestions than `send_info`?
- Is the migration kosher?

---

To explore this locally, change the ignore.CSV transport to have a bogus `host`. Update the `maxRetries = 2` in SendFunction.kt and run end to end tests. You may also have to `@Ignore` a couple of unit tests to pass `gradlew package`. This will populate the action table with a series of `send_warning` and then `send_error` pairs.

### DIFF
--- a/prime-router/src/main/kotlin/transport/SftpTransport.kt
+++ b/prime-router/src/main/kotlin/transport/SftpTransport.kt
@@ -86,7 +86,7 @@ class SftpTransport : ITransport, Logging {
                     "$sftpTransportType (orgService = ${header.receiver?.fullName ?: "null"})" +
                     ", Exception: ${t.localizedMessage}"
             context.logger.warning(msg)
-            actionHistory.setActionType(TaskAction.send_error)
+            actionHistory.setActionType(TaskAction.send_warning)
             actionHistory.trackActionResult(msg)
             RetryToken.allItems
         }
--- a/prime-router/src/main/kotlin/azure/SendFunction.kt
+++ b/prime-router/src/main/kotlin/azure/SendFunction.kt
@@ -67,7 +67,7 @@ class SendFunction(private val workflowEngine: WorkflowEngine = WorkflowEngine()
                 val serviceName = receiver.fullName
                 val nextRetryItems = mutableListOf<String>()
                 if (receiver.transport == null) {
-                    actionHistory.setActionType(TaskAction.send_error)
+                    actionHistory.setActionType(TaskAction.send_warning)
                     actionHistory.trackActionResult("Not sending $inputReportId to $serviceName: No transports defined")
                 } else {
                     val retryItems = retryToken?.items
@@ -130,6 +130,7 @@ class SendFunction(private val workflowEngine: WorkflowEngine = WorkflowEngine()
             if (nextRetryCount >= maxRetryCount) {
                 // Stop retrying and just put the task into an error state
                 val msg = "All retries failed.  Send Error report for: $reportId to $serviceName"
+                actionHistory.setActionType(TaskAction.send_error)
                 actionHistory.trackActionResult(msg)
                 context.logger.info(msg)
                 ReportEvent(Event.EventAction.SEND_ERROR, reportId)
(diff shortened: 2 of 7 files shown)


whytheplatypus:
I'm getting some failures in `prime test`. The same failures happen against `master` which makes me think it's unrelated to this change. Confirmation would be much appreciated.

prime-router/src/main/resources/db/migration/V17__add_send_info_to_action_type.sql
```diff
@@ -0,0 +1,21 @@
...
+  Add new actions 'resend' and 'rebatch', and fix the current task table to use it.
+ */
+ALTER TABLE task        ALTER COLUMN next_action TYPE VARCHAR(255);
+ALTER TABLE report_file ALTER COLUMN next_action TYPE VARCHAR(255);
+ALTER TABLE action      ALTER COLUMN action_name TYPE VARCHAR(255);
+DROP TYPE IF EXISTS task_action;
+CREATE TYPE task_action AS ENUM ('receive', 'translate', 'batch', 'send', 'download', 'wipe', 'batch_error', 'send_error', 'wipe_error', 'none', 'resend', 'rebatch', 'token_auth', 'token_error', 'send_info');
+ALTER TABLE task        ALTER COLUMN next_action TYPE task_action USING next_action::task_action;
+ALTER TABLE report_file ALTER COLUMN next_action TYPE task_action USING next_action::task_action;
+ALTER TABLE action      ALTER COLUMN action_name TYPE task_action USING action_name::task_action;
```
ronaldheft-usds, line 21: If I understand this migration correctly, it sets the column to a `VARCHAR` so it doesn't error with dropping `task_action` column. After the column is dropped, the ENUM values are updated and then it sets them back.

ronaldheft-usds (changes requested):
The migration looks good. This is a situation where we probably don't gain much splitting the database migration into a separate PR, but the process would be as follows:

* Submit a separate PR with just the migration
* Get the database PR approved, merged, and deployed into production
* Get the code PR approved and merge after the database changes are in production

prime-router/src/main/resources/db/migration/V17__add_send_info_to_action_type.sql
whytheplatypus: That's my understanding as well, the code here was stolen from a previous migration.

whytheplatypus:
@ronaldheft-usds are you looking for this PR to be split up or is it ok to leave as is in this case?

ronaldheft-usds (approved):
You're good to proceed as is.

whytheplatypus:
@jimduff-usds would you mind taking a look to sanity check the changes in the different transport types?

jimduff-usds (changes requested):
@whytheplatypus , I think as written this won't give the desired behavior for retries.  (which is, that each retry is labelled as a send_warning, then the final one is still a send_error)

Due to the way the code is structured, I suggest you just overwrite the send_warning actionName with a send_error actionName right at the moment in the code that figures out if you've run out of retry chances.   eg, add something like this:
```
                // overwrite existing send_warning or send action name, with send_error:
                actionHistory.action.actionName = TaskAction.send_error
```

Here's what it would look like in the code, around line 133 in SendFunction.kt, but I have totally not tested this:

```
                val msg = "All retries failed.  Send Error report for: $reportId to $serviceName"
                // overwrite existing send_warning or send action name, with send_error:
                actionHistory.action.actionName = TaskAction.send_error
                actionHistory.trackActionResult(msg)
                context.logger.info(msg)
                ReportEvent(Event.EventAction.SEND_ERROR, reportId)
```

prime-router/src/main/kotlin/transport/SftpTransport.kt
```diff
@@ -86,7 +86,7 @@ class SftpTransport : ITransport, Logging {
                     "$sftpTransportType (orgService = ${header.receiver?.fullName ?: "null"})" +
                     ", Exception: ${t.localizedMessage}"
             context.logger.warning(msg)
-            actionHistory.setActionType(TaskAction.send_error)
+            actionHistory.setActionType(TaskAction.send_warning)
```
whytheplatypus, line 89: I think these are somewhat redundant now, but I've left the change for clarity.

jimduff-usds: Thanks, my exact thought.

jimduff-usds (approved):
Thanks - this and the work Cle did are the clinchers that will allow us to start notifying on errors.

And - the unit tests were already written, you just had to add a line or two to them.  How pleasing (and rare) is that?

Issue #13, opened 2020-10-21
Prototype Router works in Mirth


RickHawesUSDS:
**Story**
As a demonstrator, I would like to show that the prime router can work in a Mirth Connect channel to show that the data router can be used by standard healthcare integration engines. 

**Assumptions**
- Mirth Connect 3.9 running in a Docker container
- Java 11
- Prototype type of volume and uptime

**Acceptance Criteria**
- A channel that accepts in a CSV file and routes it

RickHawesUSDS:
Closing as no longer valid

PR #2118, opened 2021-09-08
Proposal #13 - Simplify value sets and tables in schema configuration

carlosfelix2:
This PR contains the proposal to simplify the use of value sets and tables in schema configuration.  This proposal is fairly long due to all the background information to understand all the current variations of using value sets and tables.

To review, it is better to read the proposal in view mode by clicking this link: https://github.com/CDCgov/prime-reportstream/blob/01c2604b3c5346b1c96cb90c725a562ca1ebec18/prime-router/docs/proposals/0013-simplify-tables-and-value-sets.md

### DIFF
--- a/prime-router/docs/proposals/0013-simplify-tables-and-value-sets.md
+++ b/prime-router/docs/proposals/0013-simplify-tables-and-value-sets.md
@@ -1,0 +1,246 @@
+# Simplification of Value Sets and Tables Proposal
+
+## Context and Motivation
+New data translation features have been added over time to ReportStream to support onboarding senders and receivers, but some of these features have evolved to have similar functionality that may be confusing to those onboarding new entities. For this proposal, we are examining the value sets and tables translation features and their configuration.  
+
+## Goals
+Simplify the use of value sets and tables in schemas to make them simpler to configure and understand.
+
+## Background
+Value sets and tables are used to translate data values from different sources. The following subsections provide some background on the current functionality of value sets and tables.
+
+### Value Sets
+Value sets are defined in a value set file (see metadata/valuesets folder), and are identified in a schema field by the CODE field type and a valueSet parameter that specifies the name of the value set to use. A simple way to see value sets is as a key/value pair. Value sets are defined in YAML and each file can have one or more value sets defined as a YAML array. Values are comprised of the following fields:
+ - code
+ - display
+ - system (Optional)
+
+Example value set:
+```yaml
+- name: hl70136
+  system: HL7
+  values:
+    - code: Y
+      display: Yes
+    - code: N
+      display: No
+    - code: UNK
+      display: Unknown
+      system: NULLFL
+```
+
+Example schema configuration:
+```yaml
+- name: employed_in_healthcare
+  type: CODE
+  valueSet: hl70078
+```
+
+There are a few variations on how value sets are used in the schema and in code:
+
+|Variation|Serialization (data from a sender)|Deserialization (data to a receiver)
+|---|---|---|
+|Code ($code) (default)|Data is translated from a display to a coded value based on the value set|Data is translated from a display to a coded value based on the value set|
+|Display ($display)|Data is translated from a display to a coded value based on the value set|Data is translated from a coded value to a display value based on the value set|
+|Caret  ($code^$display^$system) - for HL7 related value, e.g. SNOMED|NOT SUPPORTED|Supported in a schema, but only used by REDOX serializer code - A combination of normalizedValue^display^systemCode|
+|System ($system)|NOT SUPPORTED|UNUSED - Translates a code to a system name that supports the code (e.g. NULLFL, hl70136)|
+
+Note that variations can only be configured in a schema for CSV data fields, so non-default variations can only be configured to work on CSV and internal CSV data. Here is an example of a schema configuration using the display token variation:
+```yaml
+- name: specimen_source_site_code
+  type: CODE
+  valueSet: covid-19/specimen_source_site
+  csvFields: [{name: Specimen_source_site, format: $display}]
+```
+  
+###Alternate Value Sets
+Alternate display values work the same way as value sets of the display variation, but are a custom set of values defined in a schema instead of a value set file as a means to have a unique value set for a specific receiver or sender.
+
+|Variation|Serialization (data from a sender)|Deserialization (data to a receiver)|
+|---|---|---|
+|Alternate Values|Data is translated from a display to a coded value based on the alternate values specified in the schema|Data is translated from a coded value to a display value|
+
+Example alternate values:
+```yaml
+name: patient_ethnicity
+altValues:
+- code: H
+  display: 2135-2
+- code: N
+  display: 2186-5
+- code: U
+  display: UNK
+  csvFields: [{ name: patientEthnicity, format: $alt}]
+```
+
+### Tables
+Tables are used to lookup values defined in a CSV file containing columns and rows of data.  The table lookup feature uses one or two fields and indices to try to match a row of data from the table and return the value from the specified column if found. Table fields are defined in a schema using the TABLE field type, table name parameter and tableColumn parameter. Table fields have different meanings depending on its configuration as folllows.  Of note, is that the fake data generator grabs data from a table for each of the table fields.
+A table field that uses NO mapper is simply a definition of a field that could be looked up by the specified tableColumn name. This field name is then used in a table lookup for a different field with the table mapper using the specified tableColumn as an index.
+A table field with a lookup mapper with one or two arguments will lookup a row in the table using the specified field values (indices) passed as arguments to the mapper.  When using two indices, both field values must be found in the same table row (an AND logical operator) for a match to occur.  If the row is found then the value from the column specified in the tableColumn parameter is returned.
+
+Example table index field:
+```yaml
+- name: patient_county
+  type: TABLE_OR_BLANK
+  table: fips-county
+  tableColumn: County
+```
+
+Example table lookup:
+```yaml
+- name: ordering_provider_county_code
+  type: TABLE
+  pii: false
+  table: fips-county
+  tableColumn: FIPS
+  mapper: lookup(ordering_provider_state, ordering_provider_county)
+```
+
+Example table data:
+```csv
+FIPS,County,State
+01001,Autauga,AL
+01003,Baldwin,AL
+```
+
+### LIVD Lookup
+The LIVD lookup works similarly to the table lookup, but the LIVD lookup mapper matching logic is more complicated.  The LIV lookup looks for a row in a table that matches ANY (a logical OR operation) of device ID, equipment model ID, test kit name ID or equipment model name (in that specific order). Furthermore, there are some specific rules on how to handle received compounded device IDs to match the values in the table.
+
+Example LIVD lookup mapper use:
+```yaml
+- name: equipment_model_id_type
+  type: TABLE
+  table: LIVD-SARS-CoV-2-2021-04-28
+  tableColumn: Equipment UID Type
+  mapper: livdLookup()
+```
+
+## Detailed proposal
+### Remove the Use of the Table and Code Field Types
+It is proposed to remove the TABLE and CODE field types and use the more primitive field types like text, number, email, etc. for fields to be clearer as to what type a field really is. Table fields already require the table and tableColumn parameters, and code fields require the valueSet parameter, so those fields can be identified by the presence of those required parameters.
+
+Example table lookup with primitive field type:
+```yaml
+- name: ordering_provider_county_code
+  type: NUMBER
+  table: fips-county
+  tableColumn: FIPS
+  mapper: lookup(ordering_provider_state, ordering_provider_county)
+```
+
+### Remove the Use of Fields to Define Index Columns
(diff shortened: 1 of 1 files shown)


MikeC-A6:
@carlosfelix2 - in the section "use tables for valuesets," does your proposal accommodate being able to use `else`  statements?  Such as if the result doesn't match any of these values, display a default value?

carlosfelix2:
> @carlosfelix2 - in the section "use tables for valuesets," does your proposal accommodate being able to use `else` statements? Such as if the result doesn't match any of these values, display a default value?

That scenario will be handled by the default value already present in the schema.

prime-router/docs/proposals/0013-simplify-tables-and-value-sets.md
```diff
@@ -0,0 +1,246 @@
...
+
+### Remove the Use of a Mapper for Table Lookups
+It is proposed to remove the use of the lookup mapper when looking up data in a table. The mapper will no longer be needed once the above proposals are implemented.
+
+Example proposed table lookup:
+```yaml
+- name: ordering_provider_county_code
+  type: NUMBER
+  pii: false
+  table: fips-county
```
whytheplatypus, line 159: I liked the point of all "all the knowledge of the table is in one place", but what this is doing is providing all the information for the mapper. So it should be made explicit. This would also make different mappers (e.g. `livdLookup`) easy to implement, use, and read.

e.g.

```
- name: ordering_provider_county_code
  type: NUMBER
  pii: false
  mapper:
    - lookup
       table: fips-county
       tableColumn: FIPS
      etc et.c
```

whytheplatypus: it could also be an array of mappers I suppose, leaving room for a pipeline type flow

jimduff-usds (approved):
@carlosfelix2 I like this proposal!
FYI @whytheplatypus @clediggins-usds  @mauricereeves 

This comment is on the related topic of how to implement this in a database, and focusses on the use cases we discussed to support the Sender Tools team, with the suggestion that **we create a database-based lookup to support their immediate needs**, then subsequently use it to support Valuesets, and maybe Tables.

Here goes:

# Database representation of Valuesets, Tables, etc

Any valueset, table, default lookup, or weird ways in which senders/receivers represent values, can all be represented with a generic table:

```
create table lookup (
 table_name string,                 -- eg, hl70126, or zip-code-data
 internal_value string,             -- like the old $code
 external_value string,             -- like the old $display, $alt, etc
 external_value_style string,       -- eg, column_name in a table
);
```

(And the actual table will have other important fields, like a primary key, timestamp, version number, is_deleted, etc, etc, which I'm ignoring for simplicity)

## New mapper

Then this new mapper takes an external_value in an incoming CSV column, and converts it to internal_value:

`convertToInternal(table_name, external_value_style, column_with_the_input_data)
`
## Examples

### ValueSet lookup example:

```
You might have data like this:
(
 "hl70136",
 "Y",
 "Si",
 "hl70136 en espanol",
),
(
 "hl70136",
 "N",
 "No",
 "hl70136 en espanol",
),

```
#### Now suppose a sender sends us Si/No in this field:
```
- patient_died_en_espanol
  type: TEXT

```
#### Then we can calculate this internal field based on it:

```
- name: patient_died
  mapper: convertToInternal("hl70136", "hl70136 en espagnol", "patient_died_en_espanol")

```

#### Future: require mappers always run

And if we require that mappers always run (in the future Gage Theory proposal)
Then we don't need 'patient_died_en_espanol" and this simplifies to:

```
- name: patient_died
  mapper: convertToInternal("hl70136", "hl70136 en espagnol", "patient_died")
```

Or even (if Gage allows us a _tiny bit_ of hidden magic):

```
- name: patient_died
  mapper: convertToInternal("hl70136", "hl70136 en espagnol")
```

### NPI Lookup Example (use case from Sender Automation/Tools teams)

I'm proposing we do this use case first, to support those teams.  Here's an example of how we'd use this table to support their _default NPI_ use case

```
("sender_id to NPI Lookup",
 "1245319599",
 "Reddy Family Clinic"
 "Sender ID",
)
```

#### Use the value found in sender_id to lookup the NPI to use for that clinic

```
- name: ordering_provider_id
  mapper: convertToInternal("sender_id to NPI Lookup", "Sender ID" , "sender_id")

```
### Table lookups

Tables are fundamentally different, as they might not have any unique column we can use as the internal_value, and they are used in different ways, as Carlos has noted.   

Let's concentrate on the Sender Tools use cases first.

MikeC-A6:
@jimduff-usds - just making sure this was captured in your comment re: NPI Lookup Example (use case from Sender Automation/Tools teams).  A default NPI will be unique to a facility for a given sender.  Example - GUC sender will have a default NPI for each of its 3 facilities if the NPI is blank.  We will know which facility based on the facilty_name field

MauriceReeves-usds:
@carlosfelix2 @clediggins-usds @whytheplatypus @jimduff-usds I'm all for this proposal. The only thing that concerns me is cases where we could end up needing to send non-standard data (or alt values) to a receiver. In the simplified model in the proposal, there are multiple different ways UNK could be converted to Unknown, but there's not discussion in the proposal for how that maps for a specific receiver, so it doesn't seem possible to be consistent about which alt value I will send to a receiver without having some additional information in the table to tie an alt value to sender and/or a receiver.

For example, we have a non-profit in WA state that wants to receive deidentified data from K-12 schools in WA so they can do trends and analysis on best practices in getting schools re-opened, and WA's PHD has given us approval to work with them. We are going to be sending them a CSV and we many need to map valuesets to things they can understand (i.e. not LOINC or SNOMED codes). Right now I couldn't guarantee them if I select a value for UNK that they'll get the same value every time.

So how do we achieve consistent mapping in all directions? It feels to me like part of every call to a mapper we need to pass who the intended recipient is. We may also want to pass in who the sender is and use that as well, though I could see an argument being made to make matching on many different values on ingest makes sense. It would give our system more flexibility.

Also, we should consider what the CRUD cycle will look like. Do we tombstone older values that we aren't using anymore? Is there an audit trail for changes so we can chase down potential bugs. For example, imagine someone tries to put in all the possible versions of "no" they can think of, and it causes problems for senders and receivers. Will we be able to see who added those values?

I'm fine moving forward with the case to get the sender tools team moving on their work because it is high priority, but I would like to see if we can give some thought to what I've asked above.

carlosfelix2:
> @carlosfelix2 I like this proposal!
> FYI @whytheplatypus @clediggins-usds @mauricereeves
> 
> This comment is on the related topic of how to implement this in a database, and focusses on the use cases we discussed to support the Sender Tools team, with the suggestion that **we create a database-based lookup to support their immediate needs**, then subsequently use it to support Valuesets, and maybe Tables.
> 
> Here goes:
> 
> # Database representation of Valuesets, Tables, etc
> Any valueset, table, default lookup, or weird ways in which senders/receivers represent values, can all be represented with a generic table:
> 
> ```
> create table lookup (
>  table_name string,                 -- eg, hl70126, or zip-code-data
>  internal_value string,             -- like the old $code
>  external_value string,             -- like the old $display, $alt, etc
>  external_value_style string,       -- eg, column_name in a table
> );
> ```
> 
> (And the actual table will have other important fields, like a primary key, timestamp, version number, is_deleted, etc, etc, which I'm ignoring for simplicity)
> 
> ## New mapper
> Then this new mapper takes an external_value in an incoming CSV column, and converts it to internal_value:
> 
> `convertToInternal(table_name, external_value_style, column_with_the_input_data) `
> 
> ## Examples
> ### ValueSet lookup example:
> ```
> You might have data like this:
> (
>  "hl70136",
>  "Y",
>  "Si",
>  "hl70136 en espanol",
> ),
> (
>  "hl70136",
>  "N",
>  "No",
>  "hl70136 en espanol",
> ),
> ```
> 
> #### Now suppose a sender sends us Si/No in this field:
> ```
> - patient_died_en_espanol
>   type: TEXT
> ```
> 
> #### Then we can calculate this internal field based on it:
> ```
> - name: patient_died
>   mapper: convertToInternal("hl70136", "hl70136 en espagnol", "patient_died_en_espanol")
> ```
> 
> #### Future: require mappers always run
> And if we require that mappers always run (in the future Gage Theory proposal)
> Then we don't need 'patient_died_en_espanol" and this simplifies to:
> 
> ```
> - name: patient_died
>   mapper: convertToInternal("hl70136", "hl70136 en espagnol", "patient_died")
> ```
> 
> Or even (if Gage allows us a _tiny bit_ of hidden magic):
> 
> ```
> - name: patient_died
>   mapper: convertToInternal("hl70136", "hl70136 en espagnol")
> ```
> 
> ### NPI Lookup Example (use case from Sender Automation/Tools teams)
> I'm proposing we do this use case first, to support those teams. Here's an example of how we'd use this table to support their _default NPI_ use case
> 
> ```
> ("sender_id to NPI Lookup",
>  "1245319599",
>  "Reddy Family Clinic"
>  "Sender ID",
> )
> ```
> 
> #### Use the value found in sender_id to lookup the NPI to use for that clinic
> ```
> - name: ordering_provider_id
>   mapper: convertToInternal("sender_id to NPI Lookup", "Sender ID" , "sender_id")
> ```
> 
> ### Table lookups
> Tables are fundamentally different, as they might not have any unique column we can use as the internal_value, and they are used in different ways, as Carlos has noted.
> 
> Let's concentrate on the Sender Tools use cases first.

Great ideas.  I do want to keep the schema simple, so I prefer not to have another mapper when we can use the Table lookup definition to achieve the same.  I want to get rid of mappers completely for tables.

Here is my train of thought for steps to help the senders automation team:
1. Proposal on how to move the tables to the database.  The reason this is first is that it should not take long, plus it takes time to review and approve.  Also, having the table in the database is not a blocker.  Once we implement this, updating the tables will be much quicker if we want to, but a process needs to be in place.  I do want to make the database structure generic, so we do not have to change any DB schema when a new table is added.
2. Implement step 1 of my proposal - remove the TABLE type at least.  This is so we can do step 2.  This should be quick to implement.
3. Implement step 2 of my proposal - define table indices as part of the field lookup definition.  Further clean up to not have to use the same table column names for different tables.  This should be quick to implement.
4. Implement a new parameter to define if the raw field data (coming from the provided data) will be overwritten or not by a mapper or table lookup.  This will serve to keep an NPI value if it was provided.
5. Implement tables in database.  This will require tools and a process to manage the tables.  Note we will need to detect changes in the DB for tables.  This will allow for quick changes to the tables.

This should solve the case of NPI.

jimduff-usds:
> @jimduff-usds - just making sure this was captured in your comment re: NPI Lookup Example (use case from Sender Automation/Tools teams). A default NPI will be unique to a facility for a given sender. Example - GUC sender will have a default NPI for each of its 3 facilities if the NPI is blank. We will know which facility based on the facilty_name field

@MikeC-A6 Thank-you - I had forgotten about this complexity!   I've added a proposed solution to this (I added a new section "More Complex NPI Lookup" above).   Not ideal, but would be relatively fast to implement.

carlosfelix2:
> @carlosfelix2 @clediggins-usds @whytheplatypus @jimduff-usds I'm all for this proposal. The only thing that concerns me is cases where we could end up needing to send non-standard data (or alt values) to a receiver. In the simplified model in the proposal, there are multiple different ways UNK could be converted to Unknown, but there's not discussion in the proposal for how that maps for a specific receiver, so it doesn't seem possible to be consistent about which alt value I will send to a receiver without having some additional information in the table to tie an alt value to sender and/or a receiver.
> 
> For example, we have a non-profit in WA state that wants to receive deidentified data from K-12 schools in WA so they can do trends and analysis on best practices in getting schools re-opened, and WA's PHD has given us approval to work with them. We are going to be sending them a CSV and we many need to map valuesets to things they can understand (i.e. not LOINC or SNOMED codes). Right now I couldn't guarantee them if I select a value for UNK that they'll get the same value every time.
> 
> So how do we achieve consistent mapping in all directions? It feels to me like part of every call to a mapper we need to pass who the intended recipient is. We may also want to pass in who the sender is and use that as well, though I could see an argument being made to make matching on many different values on ingest makes sense. It would give our system more flexibility.
> 
> Also, we should consider what the CRUD cycle will look like. Do we tombstone older values that we aren't using anymore? Is there an audit trail for changes so we can chase down potential bugs. For example, imagine someone tries to put in all the possible versions of "no" they can think of, and it causes problems for senders and receivers. Will we be able to see who added those values?
> 
> I'm fine moving forward with the case to get the sender tools team moving on their work because it is high priority, but I would like to see if we can give some thought to what I've asked above.

Good thoughts.  I don't think there is any rush on getting of alt values here, so that could be something that stays until we decide a better way.  I like using the sender name, but I wonder if that makes things more complicated.  Need to think about this.  I say hold on replacing alt values for now.

jimduff-usds:
###  What to do with Mappers

@carlosfelix2 , I think @whytheplatypus is planning to propose, rather than get-rid-of-mappers, is that we make-everything-a-mapper - that is, all data transforms use mappers, including table lookups.   He does plan to propose a better syntax and some other enhancements to mappers too.

His philosophy is "no magic", that is, make it as easy as possible to see what's going on, and I agree with that.
It also lets us move much more quickly with the above plan to support Sender Tools - we can just use the existing mapper() structure, knowing that we plan to migrate it to our new Gage Theory plan in the future.

### Receiver-specific values

 @mauricereeves, regarding your question about how to handle sending data to a specific receiver - the structure I proposed above would handle that this way:

### Using the database table for outgoing translations to receiver-specific weirdness:

("hl70136",   -- The Y,N,UNK valueset
 "UNK",
 "Mysterious and Enigmatic",
 "Deidentified for WA"
)

In the outgoing receiver schema definition, use the value found in sender_id to lookup the NPI to use for that clinic

- name: patient_died
  mapper: convertToExternal("hl70136", "Deidentified for WA" , "patient_died")

Note that I'm assuming mappers always run, to make this work.  (Currently mappers only run if a field is empty)

carlosfelix2:
This simplification will be combined on a future proposal that covers more of the RS features.

Issue #2036, opened 2021-08-31
Test Results ELR Primary & Secondary Tabbed List


kevinhaube:
On the React site, there is currently only a single list of test results whereas the 11ty site had this list split into ELR Primary and ELR Secondary lists with a tabbed interface to switch between the two.

Acceptance Criteria:
- ELR Primary and ELR Secondary lists present
- Proper data displayed in each list
- Can switch lists by clicking on the desired list's tab

PR #2121, opened 2021-09-08
Adding tabbed list feature for web-receiver Daily Data list

kevinhaube:
This PR adds a tab feature to the Daily Data list for Receivers. It allows Receivers to view data sent by an organization on its own exclusive list rather than all on a single list.

## Changes
- Introduced the `ReportController` class to house functions needed to alter `ReportResource` objects
- Removed `useResource()` hook from `<ReportLink>` in favor of passing the report in via props
- `<TableReports>` was updated to handle getting the list of reports from the API in my last PR (#2108) and now also parses the result and uses `getListOfSenders(report: ReportResource[])` to get an array of unique senders to pass to `<TableButtonGroup>` for labeling
- Added `useState()` hooks to `<TableReports>` and `<TableButtonGroup` with a callback function to sync the two as the buttons are clicked. This controls which sender's reports are displayed in the table.
- Filtering `reports` in `<TableReports>` based on active button (state)
- Only show `<TableButtonGroup>` when `senders` is greater than 1
- UPDATE: Adds `sendingOrg` property to API return values (still need senders to apply property when sending data)
- UPDATE: Handles refactoring and prop typing for `Daily.tsx` into single-component files (thanks @TomNUSDS for suggesting prop typing!)

### Testing
- [X] Tested locally?
- [X] Ran `quickTest all`?
- [ ] Added tests?

No
No
No
No
No
No

### Process

## Fixes
- #2036 
- #2128 (partial)


frontend-react/src/pages/Daily.tsx
```diff
@@ -5,6 +5,9 @@ import OrganizationResource from "../resources/OrganizationResource";
 import { useOktaAuth } from "@okta/okta-react";
 import { groupToOrg } from "../webreceiver-utils";
 import download from "downloadjs";
+import { getListOfSenders } from "../controllers/ReportController";
+import { ButtonGroup, Button } from "@trussworks/react-uswds"
+import { useState } from "react";
 
 const TableData = (props) => {
```
TomNUSDS, line 12: Not part of this PR, but it would be nice to get a type other than `*any` here for the props at some point. 
(e.g. below, is `props.report.content` required or could report be `null`?)

frontend-react/src/pages/Daily.tsx
```diff
@@ -33,46 +36,55 @@ const TableData = (props) => {
...
             ))}
         </tbody>
     );
 };
 
-const ReportLink = ({ reportId }) => {
-    let report = useResource(ReportResource.detail(), { reportId });
-
-    console.log(report);
+const ReportLink = (props) => {
```
TomNUSDS, line 47: Again, a nice to have in the future might be to define types for props.

TomNUSDS (approved):
There's one possible thing to fix and some general feedback, but it LGTM.

I can't really test locally without figuring out how to populate my server with data.

frontend-react/src/pages/Daily.tsx
kevinhaube: I wouldn't be against refactoring our `src/pages` files a bit; we also have some duplicated components like `<ReportLink>` which exists in `Daily.tsx` and `Details.tsx`. During that process, I'll add Prop types to each component.

kevinhaube:
@TomNUSDS I've refactored `Daily.tsx` and included prop typing for each component. Thank you for the suggestion! Much better for documentation and for any forthcoming front-end people :+1:

frontend-react/src/pages/Daily.tsx
TomNUSDS: Sounds like a great plan.

PR #2136, opened 2021-09-10
master -> production

MauriceReeves-usds:
master -> production


jimduff-usds (approved):
I'm curious about the change to line 250 in HistoryFunction.kt
It looks like you are returning both sending_org and receiving_org, however be aware that no rows in REPORT_FILE have both of those set at the same time.  One or the other is always null.

'Not both a receiver and a sender be.'
 -- Polonius

Issue #2175, opened 2021-09-13
Add New Auth Type for DE


MauriceReeves-usds:
As part of the testing for the DE connection I discovered that they use a form of SSH multi-factor auth. You have to provide a public key first (which has no password in this case) and then you need to log in with user name and password. I need to add a new auth type that allows for both types for partners that need it.

MauriceReeves-usds:
fysa: @Adrian-Brewster @jimduff-usds @RickHawesUSDS

PR #2179, opened 2021-09-13
Adds user password option for the two key file auths

MauriceReeves-usds:
This PR adds the ability for our PEM and PPK file auth to also accept a user password in addition to the file password. This is used by DE as part of what is called "partial authentication," where you need to authenticate with both a key file AND a password. Authenticating with either individually is not supported. Our existing code base did not support this.

## Changes
- Adds a user pass property for both the PEM and PPK file types and logic to chain the auth 

### Testing
- [x] Tested locally?
- [x] Ran `quickTest all`?
- [ ] Added tests?

### Process

## Fixes
- #issue

## To Be Done
- #2175


ronaldheft-usds (changes requested):
Can you test the method `Credential.fromJSON()` with a PPK JSON string that lacks the `pass` property? Deserialization may fail with the property missing.

MauriceReeves-usds:
@ronaldheft-usds will do.

PR #2182, opened 2021-09-14
Gitleaks Violation Suppression

MauriceReeves-usds:
This PR is for a gitleaks suppression in a unit test file for the DE User PPK changes.

## Changes
- Adds a unit test to check both variations of the UserPpk and UserPem JSON serialization
- Adds gitleaks suppression after getting flagged

### Testing
- [x] Tested locally?
- [x] Ran `quickTest all`?
- [x] Added tests?

### Process

## Fixes
- #issue

## To Be Done
- #issue


.environment/gitleaks/gitleaks-config.toml
```diff
@@ -32,6 +32,7 @@ title = "PRIME ReportStream Gitleaks Configuration"
         'prime-router/frontend/src/assets/pdf',
         'prime-router/build.gradle.kts',
         'frontend-react/src/components/ReportStreamHeader.tsx',
+        'prime-router/src/test/kotlin/credentials/CredentialTests',
```
td-usds, line 35: Can you sort these entries?

td-usds (changes requested):

.environment/gitleaks/gitleaks-config.toml
MauriceReeves-usds: Sorted

prime-router/src/test/kotlin/credentials/CredentialTests.kt
```diff
@@ -2,10 +2,23 @@ package gov.cdc.prime.router.credentials
 
 import assertk.assertThat
 import assertk.assertions.isEqualTo
+import assertk.assertions.isNull
 import assertk.assertions.isTrue
 import kotlin.test.Test
 
 internal class CredentialTests {
+    private val ppkCredentialWithPass = """
```
td-usds, line 10: Dumb suggestion, but make the values semi-constants - maybe something that is generated at run time (Uuid?) so that they change at every run - that you then re-reference in the test to check 'was this value picked up'.
values such as '`key`' or '`keyPass`' or '`pass`' are a bit too generic.

td-usds (approved):
One small suggested change, outside of that, looks good

prime-router/src/test/kotlin/credentials/CredentialTests.kt
MauriceReeves-usds: yea, that makes sense

td-usds (approved):

td-usds:
I resolved the false positive codesmell as Won't Fix; that particular code smell really shouldn't apply to Unit Tests and in fact, unit tests should be encouraged to be named as descriptively as possible

Issue #2185, opened 2021-09-14
BUG:  Make TASK table FK to REPORT_FILE table


jimduff-usds:
See #2180 for the initial occurrence of this bug and PR for the matching fix.

This is to do a better fix for that issue:

Note the root problem in #2180 is that the TASK table doesn't Foreign key over to the REPORT_FILE table. If it did, then no TASK could ever exist without a proper entry in the REPORT_FILE table, and this bug would never occur.
However, this is hard to fix because currently reportFunction.kt inserts into the TASK table first. And that insert is part of a nice transaction that aborts and removes the blob upon failure. The solution is probably to move that entire transaction to the end of ReportFunction (where the actionhistory inserts happen now), but that'll require a LOT of careful checking and testing.

JGreenFearless:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Pipeline-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/2185?planning-poker) with ZenHub @carlosfelix2 @jh765 @jimduff-usds @smythe5 @TurkeyFried @whytheplatypus

Issue #2180, opened 2021-09-13
Bug:  Batch queue contains Requests to process reports that failed to make it into the system.


jimduff-usds:
I believe bug #2178 is causing this after effect bug:

In today's incident, there were somehow two reports that made it into the batch queue, but never made it into the REPORT_FILE table, so when Batch wakes up to run it can’t find them.

I think I can see the problem - we’re swallowing an exception on ReportFunciton.kt line 186, then writing to the queue after it, in ReportFunction,kt line 194.

jimduff-usds:
This was very hard to recreate, but here's how:

First, its very useful to watch the TASK table during this.  This query is useful:
`select * from task T where T.created_at > now() - interval '15 minutes';`

Then, create fake data going to AL.   Then also change the timing of al-phd to 15 minutes, so you have time to do this test.

Then modify the code to force an exception prior to workflowengine.recordAction in ReportFunction.kt line 194, so that the recordAction never takes place, and do a curl submit of a file so that that exception gets thrown, and the REPORT_FILE entry never happens.   _Note that even though no entry goes into the Batch queue, the entry in the TASK table is there, lurking._

Then change the code back so that no exception is thrown, recompile, restart, and then do the exact same curl submit again.  This causes a message to go onto the BatchFunction queue.   You need to do all this quickly enough so that both the first TASK entry and the second TASK entry get the same `next_action_at` value. 

When Batch function wakes up, it'll attempt to run both the failed and the successful report, but the failed one won't be in the database, causing the entire function to fail.   The missing report poisons the entire batch.

prime_dev_1     | [2021-09-14T13:00:05.752Z] Batch function exception for event: receiver&BATCH&al-phd.elr&2021-09-14T09:00:00-04:00
prime_dev_1     | [2021-09-14T13:00:05.752Z] Result: Batch function exception for event: receiver&BATCH&al-phd.elr&2021-09-14T09:00:00-04:00
prime_dev_1     | [2021-09-14T13:00:05.752Z] Exception: Could not find dda5857a-5f16-4e59-991f-02ff0af32d85 in REPORT_FILE
prime_dev_1     | [2021-09-14T13:00:05.752Z] Stack: java.lang.IllegalStateException: Could not find dda5857a-5f16-4e59-991f-02ff0af32d85 in REPORT_FILE
[...]

Issue #2178, opened 2021-09-13
Bug: On certain exceptions, code is attempting to insert non-json into a json column.


jimduff-usds:
Bug:  error messages are trying to be inserted into the Action table action_response column, and its failing because they are not proper json.

Exception while executing function: Functions.reports Result: Failure
Exception: PSQLException: ERROR: invalid input syntax for type json
  Detail: The input string ended unexpectedly.
  Where: JSON data, line 1: 
Stack: java.lang.reflect.InvocationTargetException
	at jdk.internal.reflect.GeneratedMethodAccessor62.invoke(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source)
	at java.base/java.lang.reflect.Method.invoke(Unknown Source)
	at com.microsoft.azure.functions.worker.broker.JavaMethodInvokeInfo.invoke(JavaMethodInvokeInfo.java:22)
	at com.microsoft.azure.functions.worker.broker.EnhancedJavaMethodExecutorImpl.execute(EnhancedJavaMethodExecutorImpl.java:55)
	at com.microsoft.azure.functions.worker.broker.JavaFunctionBroker.invokeMethod(JavaFunctionBroker.java:57)
	at com.microsoft.azure.functions.worker.handler.InvocationRequestHandler.execute(InvocationRequestHandler.java:33)
	at com.microsoft.azure.functions.worker.handler.InvocationRequestHandler.execute(InvocationRequestHandler.java:10)
	at com.microsoft.azure.functions.worker.handler.MessageHandler.handle(MessageHandler.java:45)
	at com.microsoft.azure.functions.worker.JavaWorkerClient$StreamingMessagePeer.lambda$onNext$0(JavaWorkerClient.java:92)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Unknown Source)
	at java.base/java.util.concurrent.FutureTask.run(Unknown Source)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
	at java.base/java.lang.Thread.run(Unknown Source)
[...]

PR #2186, opened 2021-09-14
Fixed two exceptions that occurred in Prod yesterday:  #2178, #2180

jimduff-usds:
This PR fixes
- bug where an empty string ("") json was causing the action inserts to fail.
- related bug where failures in action inserts causes later batch function fails.

See #2178 and #2180 for more details.

### Testing
- [X] Tested locally?
- [X] Ran `quickTest all`?
- [ ] Added tests?

No new known security issues.

## Fixes
- #2178, #2180

## To Be Done
- Root problem is the the report_id in TASK table does not foreign key to the REPORT_FILE table, but the fix for that is complex.   Ticketed as #2185
-


prime-router/settings/organizations.yml
```diff
@@ -1157,7 +1157,7 @@
         credentialName: DEFAULT-SFTP
       timing:
         operation: MERGE
-        numberPerDay: 1440
+        numberPerDay: 360   # Useful when you need a bit more of: This thing all things devours: Birds, beasts, trees, flowers
```
jimduff-usds, line 1160: Alabama is now a nice test state to generate fake data for when you need tests that don't run every minute.

prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -199,7 +199,8 @@ class ActionHistory {
      */
     fun trackActionResponse(response: HttpResponseMessage, verboseResponse: String) {
         action.httpStatus = response.status.value()
-        action.actionResponse = JSONB.valueOf(verboseResponse)
+        if (!verboseResponse.isNullOrBlank())
+            action.actionResponse = JSONB.valueOf(verboseResponse)
```
jimduff-usds, line 203: An empty string is not valid json.   this fixes a bug #2178 that occurred in production - there are valid paths thru the code that do not generate a non-empty json action_response field.

prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -645,7 +646,10 @@ class ActionHistory {
                     msg = "reportFiles is null"
                 } else {
                     if (tasks.size != reportFiles.size) {
-                        msg = "Different result count: Got ${tasks.size} headers but ${reportFiles.size} reportFiles"
+                        msg = "Different report_file count: Got ${tasks.size} TASKS," +
```
jimduff-usds, line 649: A report that is missing from the REPORT_FILE table is an unrecoverable bug.   However currently such bugs cause an entire batch to fail, even if one report is missing.   To prevent this badness, I've changed to code to skip around these errors, but this at least helps to log the details of what occurred.

prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -688,14 +692,6 @@ class ActionHistory {
                         msg += "header.itemCount = ${task.itemCount}, " +
                             "but reportFile.itemCount= ${reportFile.itemCount}, "
                     }
-                    // Not checking this.   Because Task updates, the nextAction change to 'none' normally.
```
jimduff-usds, line 691: Old useless logging.   Removed because its not an error.

prime-router/src/main/kotlin/azure/WorkflowEngine.kt
```diff
@@ -308,23 +308,35 @@ class WorkflowEngine(
...
             val reportFiles = ids
-                .map { db.fetchReportFile(it, org = null, txn = txn) }
+                .mapNotNull {
+                    try {
+                        db.fetchReportFile(it, org = null, txn = txn)
+                    } catch (e: Exception) {
+                        println(e.printStackTrace())
+                        // sanityCheckReport below will log the problem in better detail.
+                        // note that we are logging but ignoring this error, so that it doesn't poison the entire batch
+                        null // id not found. Can occur if errors in ReportFunction fail to write to REPORT_FILE
```
jimduff-usds, line 318: This is the key fix to #2180.   Skip over reports that are in the TASK table but missing from the REPORT_FILE table.

prime-router/src/main/kotlin/azure/WorkflowEngine.kt
```diff
@@ -308,23 +308,35 @@ class WorkflowEngine(
...
+                if (reportFiles[it.reportId] != null) {
+                    createHeader(it, reportFiles[it.reportId]!!, null, organization, receiver)
+                } else {
+                    null
+                }
             }
 
             updateBlock(headers, txn)
 
-            headers.forEach {
```
jimduff-usds, line 324: Note the change here to iterate thru tasks rather than headers.  This causes the original list of TASK to all get marked as complete, including any TASK with missing REPORT_IDs.   Why?  Because missing report_id is an unrecoverable error.  Solution would be to resubmit the entire report.

prime-router/src/main/kotlin/azure/WorkflowEngine.kt
clediggins-usds: How could a task for a report end up in the database without there being a report in the REPORT_FILE table? Is there a case to resolve that issue?

carlosfelix2:
@jimduff-usds How do we test the changes?  How would I break it in the old code?

prime-router/src/main/kotlin/azure/WorkflowEngine.kt
```diff
@@ -308,23 +308,35 @@ class WorkflowEngine(
             )
             val ids = tasks.map { it.reportId }
             val reportFiles = ids
-                .map { db.fetchReportFile(it, org = null, txn = txn) }
+                .mapNotNull {
+                    try {
+                        db.fetchReportFile(it, org = null, txn = txn)
+                    } catch (e: Exception) {
```
carlosfelix2, line 314: So, if the row is not found then db.fetchReportFile returns null (see DatabaseAccess 192).  You will only get an exception if there is more than one row or something went wrong in the DB query.  Not sure this catch does what you want.

prime-router/src/main/kotlin/azure/WorkflowEngine.kt
```diff
@@ -308,23 +308,35 @@ class WorkflowEngine(
...
+                        println(e.printStackTrace())
+                        // sanityCheckReport below will log the problem in better detail.
```
carlosfelix2, line 316: This comment confused me.  It sounds like there is a sanity check below, but it only says null.

prime-router/src/main/kotlin/azure/WorkflowEngine.kt
carlosfelix2: This would be a great comment to have in the code

carlosfelix2 (approved):
Approved, but see my comments for WorkflowEngine

prime-router/src/main/kotlin/azure/WorkflowEngine.kt
carlosfelix2: Catch DataAccessException instead.

prime-router/src/main/kotlin/azure/WorkflowEngine.kt
jimduff-usds: done

prime-router/src/main/kotlin/azure/WorkflowEngine.kt
jimduff-usds: done

prime-router/src/main/kotlin/azure/WorkflowEngine.kt
jimduff-usds: The fetchReportFile function calls error() which throws an exception when the report is not found.

prime-router/src/main/kotlin/azure/WorkflowEngine.kt
jimduff-usds: Yes, see #2185

PR #1965, opened 2021-08-24
Jorgel/githooks/ktlint

jorg3lopez:
This PR adds ktlint format checker as part of the githooks. The ktlint format checker checks for code violations.  If the check finds code violations, the githook will not pass.

### Checks
- ./gradlew test passed
- ./gradlew testEnd2End passed
- All tasks from build passed
- ./quick-test all passed

### Testing
- [ X] Tested locally?
- [ ] Ran `quickTest all`?
- [ ] Ran `./prime test` against local Docker ReportStream container?
- [ ] Downloaded a file from `http://localhost:7071/api/download`?
- [ ] Added tests?

## Fixes
- #1963 
- #1745

PR #2220, opened 2021-09-16
Running more in CI

whytheplatypus:
This PR ...

Adds linting and other checks to the CI checks and moves some checks into parallel jobs.

Test Steps:
1. Check jobs run in this PR e.g. https://github.com/CDCgov/prime-reportstream/pull/2220/checks?check_run_id=3624351115

## Changes
- Adds klint linting check to CI checks
- Moves docs check and linting into parallel jobs (results in longest check running in roughly 11min instead of 16min, and failing much faster)
- Adds Smoke tests

## To Be Done
- [ ] update github settings to require more CI jobs to pass

### DIFF
--- a/.github/workflows/build_hub.yml
+++ b/.github/workflows/build_hub.yml
@@ -9,6 +9,11 @@ on:
     branches:
       - master
 
+env:
+  # These are for CI and not credentials of any system
+  DB_USER: prime
+  DB_PASSWORD: changeIT!
+
 jobs:
   pre_job:
     name: Pre Job
@@ -32,10 +37,6 @@ jobs:
     needs: pre_job
     if: ${{ needs.pre_job.outputs.has_router_change == 'true' }}
     runs-on: ubuntu-latest
-    env:
-      # These are for CI and not credentials of any system
-      POSTGRES_USER: prime
-      POSTGRES_PASSWORD: changeIT!
     defaults:
       run:
         working-directory: prime-router
@@ -43,21 +44,6 @@ jobs:
     steps:
       - name: "Check out changes"
         uses: actions/checkout@v2
-
-      - name: Set Environment Variables
-        run: |
-          echo >> $GITHUB_ENV DB_USER=${POSTGRES_USER}
-          echo >> $GITHUB_ENV DB_PASSWORD=${POSTGRES_PASSWORD}
-          if [ "$GITHUB_BASE_REF" == "production" ]
-          then
-              echo "Building for the production environment."
-              echo >> $GITHUB_ENV ACR_REPO=pdhprodcontainerregistry.azurecr.io
-              echo >> $GITHUB_ENV PREFIX=pdhprod
-          else
-              echo "Building for the test environment."
-              echo >> $GITHUB_ENV ACR_REPO=pdhstagingcontainerregistry.azurecr.io
-              echo >> $GITHUB_ENV PREFIX=pdhstaging
-          fi
 
       # Appears not to be needed on GitHub (but needed when running act [https://github.com/nektos/act] locally)
       # - name: Install docker-compose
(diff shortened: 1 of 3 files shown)


.github/workflows/build_hub.yml
```diff
@@ -27,15 +27,15 @@ jobs:
               - 'prime-router/**'
               - '.github/workflows/build_hub.yml'
 
-  build_router:
+  unit_tests:
```
carlosfelix2, line 30: Why not keep the job and name the same?  Why is the next line different?

.github/workflows/build_hub.yml
```diff
@@ -141,5 +155,43 @@ jobs:
...
+    name: Build docker image
+    needs:
+      - unit_tests
+      - linting
+      - docs
+    if: ${{ needs.pre_job.outputs.has_router_change == 'true' }}
+    runs-on: ubuntu-latest
+    env:
+      # These are for CI and not credentials of any system
+      POSTGRES_USER: prime
```
carlosfelix2, line 168: You are duplicating the setting of these values.  I say define a variable early on and reuse it.

.github/workflows/build_hub.yml
```diff
@@ -91,6 +76,35 @@ jobs:
...
+    defaults:
+      run:
+        working-directory: prime-router
+
+    steps:
+      - name: "Check out changes"
+        uses: actions/checkout@v2
+
+      - name: Run Linting Check
+        run: bash ./build.sh -- gradle ktlintCheck
```
carlosfelix2, line 94: I have no such task in my branch.  Where is this coming from?

carlosfelix2 (changes requested):

.github/workflows/build_hub.yml
whytheplatypus: it's setup [here](https://github.com/CDCgov/prime-reportstream/blob/master/prime-router/build.gradle.kts#L33) and the including the plugin [automatically adds the tasks](https://github.com/JLLeitschuh/ktlint-gradle)

carlosfelix2:
Change the title of the PR as this is not an experiment.

ronaldheft-usds:
FYI, @jorg3lopez is always working on running ktlint as a pre-commit hook: https://github.com/CDCgov/prime-reportstream/pull/1965

.github/workflows/build_hub.yml
```diff
@@ -141,5 +155,43 @@ jobs:
         run: |
           false
 
+  build_docker:
+    name: Build docker image
+    needs:
+      - unit_tests
+      - linting
+      - docs
```
ronaldheft-usds, line 163: I'm confused with the addition of all these lines in the job (158-195). These don't seem to have anything to do with linting?

.github/workflows/build_hub.yml
```diff
@@ -27,15 +27,15 @@ jobs:
...
     name: Build Router
     needs: pre_job
     if: ${{ needs.pre_job.outputs.has_router_change == 'true' }}
     runs-on: ubuntu-latest
     env:
       # These are for CI and not credentials of any system
-      POSTGRES_USER: prime
-      POSTGRES_PASSWORD: changeIT!
+      DB_USER: prime
+      DB_PASSWORD: changeIT!
```
ronaldheft-usds, line 38: Why was this changed?

ronaldheft-usds (changes requested):

.github/workflows/build_hub.yml
whytheplatypus: I also split some of the jobs out to allow them to run in parallel. This tells github actions to require these three jobs to pass before starting this one.

btw, is the reason for this job just to check that docker builds?

.github/workflows/build_hub.yml
whytheplatypus: It allows the removal of the GITHUB_ENV call later.

whytheplatypus: eventually we could use github environments to nicely store some of these settings and reduce the conditionals.

.github/workflows/build_hub.yml
ronaldheft-usds: Ah, I see it now.

Correct, this job verifies the Docker container builds along with all unit tests.

.github/workflows/build_hub.yml
```diff
@@ -141,5 +156,37 @@ jobs:
...
+          then
+              echo "Building for the production environment."
+              echo >> $GITHUB_ENV ACR_REPO=pdhprodcontainerregistry.azurecr.io
+              echo >> $GITHUB_ENV PREFIX=pdhprod
+          else
+              echo "Building for the test environment."
+              echo >> $GITHUB_ENV ACR_REPO=pdhstagingcontainerregistry.azurecr.io
+              echo >> $GITHUB_ENV PREFIX=pdhstaging
+          fi
+
```
ronaldheft-usds, line 187: We can actually drop these lines if you if you update line 192 to be something like `CDCgov/reportstream`. The actual image tag name only matters in the release job, so it's kind of pointless for this extra logic in the CI job.

ronaldheft-usds (approved):
Looking great, with just one suggested change. After you merge this, can you ping me on Slack? I want to updated our required passing CI jobs to include the new parallelized docs job and the linting job.

.github/workflows/build_hub.yml
whytheplatypus: ok makes total sense, that's why I was wondering if it was just checking the build.

whytheplatypus:
@carlosfelix2 ok i think I got everything! but double check me

jimduff-usds (approved):

carlosfelix2 (approved):

PR #2256, opened 2021-09-20
fix: DNS records cannot be registered against multiple private endpoints

ronaldheft-usds:
This PR fixes multiple private endpoints overwriting DNS records within the same resource group.

Azure [only allows one private DNS zone per resource group](https://dev.to/kaiwalter/using-azure-private-links-and-private-dns-zones-with-globally-distributed-resources-4ce3), so when using multiple private endpoints in the same resource group, only one can register a DNS record against a private DNS zone. Allowing them all to registered a DNS record creates unpredictable results.

### DIFF
--- a/operations/app/src/modules/common/private_endpoint/main.tf
+++ b/operations/app/src/modules/common/private_endpoint/main.tf
@@ -59,20 +59,11 @@ resource "azurerm_private_endpoint" "endpoint" {
 
   # Automatically register the private endpoint in the private DNS zones
   dynamic "private_dns_zone_group" {
-    for_each = [0] # DNS needed for all zones. For each block retained to keep existing resource paths.
+    for_each = var.create_dns_record ? [0] : []
     content {
       name                 = local.endpoint_name
       private_dns_zone_ids = [for dns_zone in data.azurerm_private_dns_zone.private_dns_cname : dns_zone.id]
     }
-  }
-
-  lifecycle {
-    ignore_changes = [
-      # Added the subnet id to new private endpoints to avoid collisions, but we don't want to rename existing endpoints
-      name,
-      private_service_connection[0].name,
-      private_dns_zone_group[0].name,
-    ]
   }
 }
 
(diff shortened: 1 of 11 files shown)


operations/app/src/modules/common/private_endpoint/main.tf
```diff
@@ -59,21 +59,12 @@ resource "azurerm_private_endpoint" "endpoint" {
...
   }
-
-  lifecycle {
-    ignore_changes = [
-      # Added the subnet id to new private endpoints to avoid collisions, but we don't want to rename existing endpoints
-      name,
-      private_service_connection[0].name,
-      private_dns_zone_group[0].name,
-    ]
-  }
```
ronaldheft-usds, line 76: Due to DNS records being overwritten, our existing DNS records are no longer guaranteed to be correct, so we want to allow the private endpoints to be recreated.

operations/app/src/modules/common/private_endpoint/main.tf
```diff
@@ -59,21 +59,12 @@ resource "azurerm_private_endpoint" "endpoint" {
 
   # Automatically register the private endpoint in the private DNS zones
   dynamic "private_dns_zone_group" {
-    for_each = [0] # DNS needed for all zones. For each block retained to keep existing resource paths.
+    for_each = var.create_dns_record ? [0] : []
```
ronaldheft-usds, line 62: This conditional allow us to disable DNS association per private endpoint.

operations/app/src/modules/storage/candidate_slot.tf
```diff
@@ -68,6 +70,40 @@ module "storageaccount_candidate_queue_private_endpoint" {
...
+
+module "storage_candidate_queue_private_endpoint" {
+  source             = "../common/private_endpoint"
+  resource_id        = azurerm_storage_account.storage_account_candidate.id
+  name               = azurerm_storage_account.storage_account_candidate.name
+  type               = "storage_account_queue"
+  resource_group     = var.resource_group
+  location           = var.location
+  endpoint_subnet_id = data.azurerm_subnet.endpoint_subnet.id
+  create_dns_record  = false
```
ronaldheft-usds, line 106: These private endpoints were actually missing from the original Palo Alto network PR.

PR #2319, opened 2021-09-22
Consolidate error and warning messages for the CSV uploader

sean-usds:
## Context
When there are error and/or warning during CSV uploading, each one is listed in the UI. This can become verbose and unfriendly to the end user. The solution is to consolidate these messages as shown in [these mocks](https://www.figma.com/file/ctpYEoOBWGkUnk9cmXzVp2/ReportStream-Submitter-Front-End?node-id=980%3A3216)

Tech specs proposal detailed here: https://app.zenhub.com/workspaces/prime-data-hub-5ff4833beb3e08001a4cacae/issues/cdcgov/prime-reportstream/1966

## Changes
- Introduces a `ResponseMessage` interface that `ResultDetail` uses instead of a `String` for the message
- Creates data classes for each error type that implement `ResponseMessage`
- Creates a generic `ResponseMsgType` all HL7 serialization warnings/errors since these do not need consolidation
- Adds new fields `consolidatedErrors` and `consolidatedWarnings` to the waters api response that contain the consolidated messages

## Test Steps:
1. Create an upload csv with known problems (for example, using the test-waters-data.csv, make the patientAge column blank)
2. Upload csv in the csv uploader
3. Verify that the errors/warnings are conolidated

### Before
image

### After
image

### DIFF
--- a/frontend-react/src/pages/Upload.tsx
+++ b/frontend-react/src/pages/Upload.tsx
@@ -23,8 +23,8 @@ export const Upload = () => {
     const {authState} = useOktaAuth();
     const [isSubmitting, setIsSubmitting] = useState(false);
     const [file, setFile] = useState(null);
-    const [warnings, setWarnings] = useState([]);
-    const [errors, setErrors] = useState([]);
+    const [consolidatedWarnings, setConsolidatedWarnings] = useState([]);
+    const [consolidatedErrors, setConsolidatedErrors] = useState([]);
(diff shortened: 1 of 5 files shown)


ahay-agile6 (approved):
this looks really nice. thanks for taking this on, @sean-usds !
image

jimduff-usds (commented):
See #2455 for follow on work.

PR #2340, opened 2021-09-23
perf: Indices for getReports

ronaldheft-usds:
We can probably improve the performance of creation: https://www.cybertec-postgresql.com/en/postgresql-parallel-create-index-for-better-performance/

### DIFF
--- a/prime-router/src/main/resources/db/migration/V18__add_get_reports_indicies.sql
+++ b/prime-router/src/main/resources/db/migration/V18__add_get_reports_indicies.sql
@@ -1,0 +1,44 @@
+/*
+ * The Flyway tool applies this migration to create the database.
+ *
+ * Follow this style guide https://about.gitlab.com/handbook/business-ops/data-team/platform/sql-style-guide/
+ * use VARCHAR(63) for names in organization and schema
+ *
+ * Copy a version of this comment into the next migration
+ *
+ */
+
+/*
+ * PROBLEM QUERY:
+ *
+ * select
+ * a.testing_lab_clia,
+ * a.testing_lab_name,
+ * count(a.covid_results_metadata_id) as COUNT_RECORDS,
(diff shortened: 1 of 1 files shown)


prime-router/src/main/resources/db/migration/V18__add_get_reports_indicies.sql
```diff
@@ -0,0 +1,44 @@
...
+ * a.testing_lab_city, a.testing_lab_state
+ * from covid_result_metadata a
+ * join report_ancestors(cast('fc218c2a-e7ef-4226-adfc-940058005a94' as uuid))
+ *     on report_ancestors.report_ancestors = a.report_id
+ * group by a.testing_lab_name,
+ * a.testing_lab_clia,
+ * a.testing_lab_city,
+ * a.testing_lab_state
+*/
+
```
jimduff-usds, line 28: @ronaldheft-usds , I just noticed something:   the existing FK index on covid_result_metdata on report_id is marked "INVALID" in Staging.    It is not marked that way in my local Postgres, nor in production.

I found this reference on invalid indexes:
https://medium.com/carwow-product-engineering/problems-with-concurrent-postgres-indexes-and-how-to-solve-them-c57f7656c852

I suggest we fix that before doing anything else.

Weirdly, there are already indexes on report_lineage's parent_report_id and child_report_id in STAGING but not on my localhost or in PROD:
   "report_lineage_child_id" btree (child_report_id)
    "report_lineage_parent_id" btree (parent_report_id)
Maybe y'all add this manually as a test?

I'm not able to re-create any slowness with this query, either in Staging or locally, using a "sent" report_id, but maybe it works in staging because of those new indexes?   

We need to think carefully about adding all these indexes, especially the last one, as they can potentially slow down the ReportFunction inserts, even if they are set too CONCURRENTLY.

PR #2342, opened 2021-09-24
Removes the call to get the facilities from getReports

MauriceReeves-usds:
This PR removes the call to get the facilities from `getReports`


kevinhaube (approved):
lgtm!

PR #2345, opened 2021-09-24
chore: Add alerts to Azure metrics

ronaldheft-usds:
This PR adds the following PagerDuty alerts ([inherited from SimpleReport’s alerts](https://github.com/CDCgov/prime-simplereport/blob/main/ops/services/alerts/app_service_metrics/main.tf)):

* App Service / Function App
    * CPU >= 70%
    * Memory >= 85%
    * Single API response duration >= 10 seconds
    * HTTP Server 2xx Errors (where successful request == false)
    * HTTP Server 4xx Errors (excluding 401s) >= 10
    * HTTP Server 5xx Errors >= 10
    * App Insights Anomalies - Unusual rise in failed requests or dependencies
* PostgreSQL Database
    * CPU >= 70%
    * Memory >= 85%
    * Single database query duration >= 10 seconds
    * 10+ database queries duration >= 1.25 seconds in last 5 minutes


operations/app/src/modules/application_insights/alert_db.tf
```diff
@@ -0,0 +1,114 @@
...
+  resource_group_name = var.resource_group
+  severity            = 2
+  frequency           = 5
+  time_window         = 5
+
+  data_source_id = azurerm_application_insights.app_insights.id
+
+  query = <<-QUERY
+dependencies
+| where timestamp >= ago(5m) and name has "SQL:" and duration >= 10000
```
RickHawesUSDS, line 75: Worried that this trigger in normal operation. Suggest 60 seconds.

operations/app/src/modules/application_insights/alert_db.tf
```diff
@@ -0,0 +1,114 @@
+# Severity 0 - Critical
+# Severity 1 - Error
+# Severity 2 - Warning
+# Severity 3 - Informational
+# Severity 4 - Verbose
+
+locals {
+  database_cpu_threshold = 70
+  database_mem_threshold = 85
+}
```
RickHawesUSDS, line 10: Seems like a good level.

operations/app/src/modules/application_insights/alert_db.tf
jimduff-usds: Yeah, 10 seems too low.   60 might be too high for Azure function queries, which seem to get killed at 30 seconds.

jimduff-usds (approved):
So we might get a ton of false alarms but we won't know until we try it, then we can fine tune.
Bring it on.

operations/app/src/modules/application_insights/alert_db.tf
ronaldheft-usds: My thinking was it should be under 30 seconds, due to Azure functions getting killed at the 30 second mark.

The severity of the alert is also set a warn, so PagerDuty will send an email but not page.

Issue #1974, opened 2021-08-25
Track whether a state is "live" and activated with ReportStream, in the Settings table


jimduff-usds:
Whether a state is active/live with ReportStream is somewhat independent of whether they have a fully set up transport in the Settings table.   A state might be "live", but using the download site, so with no transport.   Or a state might not yet be live, but have working transports ready to go. 

This makes it hard to flag what's an error and what's not an error, in our metabase dashboards.   If we had solid/correct values for which states are supposed to be getting data, then we could much more easily flag errors, maybe even with PagerDuty.

Also, we could generate the map directly and automatically from database data.

So, this ticket is to
- add a new field,   `is_customer_activated`, to both `Sender.kt `and `Receiver.kt`, fed by data in the settings table.   Values are true or false.
- Note that a particular organization might have one receiver activated, and another one not.
- if you have a better name, let me know.   I added the word 'customer' to show that this is not a _technical_ status of our data feed, but more a human/legal status.
- default is false, not activated.
- you must backfill this value in organizations.yml!
- you must also backfill it for all receivers and senders in Staging and Production   The values for what's activated or not may differ between Staging and Production.   Mo has written code that does these sweeping backfills of json structured data.  See `src/main/resources/db/migration/V10__Update_settings_for_name_templates.sql`
- Be careful with flyway.  Once you deploy, unless you've written a rollback .sql, it'll be hard to go back if its messed up - you may have to remove the table in your database and start over.
- when we roll this to Prod, we should be sure to pull down the full set of settings prior to deployment, in case we mess up the data.
- note also that the act of "going live" will no longer just be an announcement in Slack - it'll have to be accompanied with a settings change.   The need for a proper settings UI increases daily.

@MauriceReeves-usds , @jlusds, @anshulkumar-usds  FYI.

jimduff-usds:
Based on @clediggins-usds suggestion, we can make this an `activationStatus` enumeration rather than boolean.

I'm also wondering if it should be a database column rather than in the json.  That would make it much more accessible to simple queries.

Note that our design is that each Organization can have zero or more Senders and zero or more Receivers.   Currently (and somewhat confusingly, since there's also a parent-child relationship) each gets its own separate row in the Settings table.

So now the proposal here is that every Organization, Sender and Receiver gets an activationStatus, as a new column in the Settings table.

[...]

PR #2351, opened 2021-09-24
#1974 - add customerStatus to sender and receiver

clediggins-usds:
This PR ...

Test Steps:
This change simple adds a customerStatus value to each sender and receiver json blog in the database, and to the Sender and Receiver classes. Run src/scripts/db_one_off/1974_customer_status.sql against your local postgres database, then select a receiver and verify the jsonb object contains a customerStatus value.  Do the same for sender.

Build and run tests locally to ensure the classes are getting their data properly, and populating the customerStatus value.

## Changes
- Add customerStatus to Receiver.kt
- Add customerStatus to Sender.kt
- Update organizations.yml to provide initial values for customerStatus for all senders and receivers
- Create one-off script to populate existing sender and receiver values in the database

### Testing
- [x ] Tested locally?
- [ x] Ran `./prime test` or `./gradlew testSmoke` against local Docker ReportStream container?
- [? ] Downloaded a file from `http://localhost:7071/api/download`?
- [ -] Added tests?

### Process
- [N ] Are there licensing issues with any new dependencies introduced?
- [ Y] Includes a summary of what a code reviewer should test/verify?
- [ ?] Updated the release notes?
- [ N] Database changes are submitted as a separate PR?
- [ ?] DevOps team has been notified if PR requires ops support?

## Fixes
- 1974

## To Be Done
- #issue
- N/A

    - No
  - No
  - No 
  - No
  - No
  - No

  - Logging did not change as part of this PR
  - None required


prime-router/src/main/kotlin/Receiver.kt
```diff
@@ -26,6 +27,7 @@ open class Receiver(
     val name: String,
     val organizationName: String,
     val topic: String,
+    val customerStatus: String,
```
jimduff-usds, line 30: Let's contstrain the values for this using an enum type eg,

enum class CustomerStatus { inactive, testing, active }

Put it somewhere it can be shared between Sender.kt and Receiver.kt.  Maybe put it in SettingsProvider.kt?

prime-router/src/main/kotlin/Sender.kt
```diff
@@ -6,14 +6,15 @@ import gov.cdc.prime.router.tokens.JwkSet
...
- * to the data hub (minus the credentials used by that agent, of course). It
+ * the data hub (minus the credentials used by that agent, of course). It
  * contains information about the specific topic and schema that the sender uses.
  */
 open class Sender(
     val name: String,
     val organizationName: String,
     val format: Format,
     val topic: String,
+    val customerStatus: String,
```
jimduff-usds, line 17: Same comment - everywhere this is used, lets use an enum instead of a string.

prime-router/src/main/kotlin/cli/ProcessDataCommands.kt
```diff
@@ -483,6 +483,7 @@ class ProcessData(
                             "emptyReceiver",
                             "emptyOrganization",
                             "covid-19",
+                            "unknown",
```
jimduff-usds, line 486: Instead of "unknown", probably can just use the enum's `inactive` value.

prime-router/src/test/kotlin/FileSettingsTests.kt
```diff
@@ -91,7 +94,7 @@ class FileSettingsTests {
                 countyName = null,
                 senders = listOf(),
                 receivers = listOf(
-                    Receiver("elr", "single", "topic", "schema")
+                    Receiver("elr", "single", "topic", "status", "schema")
```
jimduff-usds, line 97: use of the enum values, instead of "status"

prime-router/settings/organizations-prod.yml
```diff
@@ -9,6 +9,7 @@
     - name: default
```
jimduff-usds, line 9: Sorry we didn't catch this sooner - we are actually thinking of deleting this and probably organizations-test.yml too.   See conversation here:  https://usds.slack.com/archives/C01HFHPNJEL/p1632227581289000

jimduff-usds (changes requested):
@clediggins-usds  - this was a lot more changes than I was thinking, since there are so many settings yamlitos floating around in .yml files, plus in the code, plus dummy Receiver and Sender settings all over the place -  **Thanks for all this work!**

Per my comments on the files, I do think it would be valuable to have an enum class for the customerStatus value, so that we have a tight control over the vocabulary.  This will be a bit of work for you to make that change everywhere, but I think will be worth it over the long term.

(Also, unfortunately, considering you did a lot of work to update these, we are considering just deleting organizations-prod.yml (and -test.yml too) from GitHub, since the db is the real source of truth for settings in test and prod.)

prime-router/src/main/kotlin/Receiver.kt
clediggins-usds: Updated

prime-router/src/main/kotlin/Sender.kt
clediggins-usds: Upated

prime-router/src/test/kotlin/FileSettingsTests.kt
clediggins-usds: Updated

prime-router/settings/organizations-prod.yml
clediggins-usds: Should I remove the changes from this, or leave them because it doesn't matter?

prime-router/src/main/kotlin/cli/ProcessDataCommands.kt
clediggins-usds: Updated

prime-router/settings/organizations-prod.yml
jimduff-usds: seems fine to leave them

prime-router/src/scripts/db_one_off/1974_customer_status.sql
```diff
@@ -0,0 +1,28 @@
...
+-- set all customerStatus to default 'inactive'
+update setting
+set values = jsonb_set(values, '{customerStatus}', '"inactive"', true)
+where type in ('SENDER','RECEIVER');
+
+-- set active senders and receivers to 'active'
+update setting
+set values = jsonb_set(values, '{customerStatus}', '"active"', true)
+where type in ('SENDER','RECEIVER')
+  and  values ->> 'organizationName' in ('simple_report','waters','safehealth','cue','inbios','imagemover','anavasidx','strac',
```
jimduff-usds, line 16: Feel free to make minor last minute additions to this without having to do another PR.

jimduff-usds (approved):
Awesome - thank you!

PR #2370, opened 2021-09-27
[Draft] Enable type checking and stricter linting and yarn

TomNUSDS:
Test Steps:
1. Run should work.
2. Build should work.
3. Build should deploy to staging.

## Changes
Hey. I have a massive PR I'm working on that "kind of fixes" a ton of things.

**Typescript**
Typescript had `noImplicitAny: false` which basically bypasses a bunch of type checking.

I flipped this back to the default, fixed most of the type checking errors. The places I couldn't get to work (most noticeably `AuthorizedRoute.ts`, `Upload.tsx`), I added  `// @ts-nocheck // TODO:` at the top of these files as a reminder.

I'll open a bug to fix these.

**Linting**
The package already had the `eslint-config-airbnb-base` linter syntax package installed, but wasn't using it. So, I modified the build using it and literally no file passed.

I don't have a strong opinion about "which style guide is best", but I do believe there should be SOME style defined and followed.
My solution was to fill the package.json's `eslintConfig.rules` section with approx 29 rules turned off or to "warn". This allows things to mostly work but still warns us they need to be cleaned up.
It also required the package's start to have the `DISABLE_ESLINT_PLUGIN` flag `"start": "DISABLE_ESLINT_PLUGIN=true react-scripts start"` 

Clearly this isn't ideal, but I'm not sure of a better approach to incrementally fix files. And the longer we wait, the more cowboy code will be written that does not follow any real style guide. I'm open to other suggestions.

I suspect that just running the linter-auto-fix will solve most of the issues, but again, that will modify every single file, so maybe best done on it's own PR.

**Yarn**
I'd like to add `"npmClient": "yarn"` to the package.json and switch to using yarn to do package management. SimpleReport uses it. This means `yarn.lock` replaces `npm.lock`

**Build**
I've removed the "static html crawling build" which makes me a bit nervous about deploying to staging. The build dir output seems to be reasonable, but I'm not 100% certain how to test end-to-end.

**CORS**
Speaking of testing, cors for localhost still doesn't work. I spent a lot more time on this than I'd care to admit, but it was over the weekend, so denting my forehead on the table is a fine programmer hobby, right?

The ONLY way I could get it to work was to completely turn off security in an instance of chrome.
`"/Applications/Google Chrome.app/Contents/MacOS/Google Chrome" --user-data-dir=chrome-dev-data --disable-web-security`

The only remaining idea I wasn't able to test, was to run `https:` on localhost and test that. But someone (Ron?) will need to had https://localhost:3000 and https://localhost:7071 to the allowed Okta domains (I assume it already has these localhost ports but just not HTTPS).

I've also remove the `Content-Security-Policy` meta from the index.hmtl page until we get this resolved. It just makes everything more complicated and honestly, the policy should be enforce via the http header anyways.

**USWDS**
Was going to add support for USWDS build support, but honestly, there's so many other changes, I'm splitting this out into a different PR.

**Jest testing**
Didn't touch this, but like strict linting, the longer we wait, the harder it is.

### Testing
- [x] Tested locally?

### Process
- [ ] Create bugs to do follow-up work (fix typescript and linter errors that weren't addressed).

## Fixes


sean-usds (changes requested):
just marking this as "request changes" so we don't some how accidentally try to merge this since this is being broken into multiple PRs

PR #2383, opened 2021-09-28
ACK IN BLACK!!!

MauriceReeves-usds:
This PR is for the ACK proposal. The old PR got wrecked by a bad rebase, but now we are ACK IN BLACK.


clediggins-usds:
How often do we get ack/nack responses from our recipients?
Have we had any senders asking for ack/nack (other than HCA, who asked about that today)?
Would this be opt-in and we would still return json responses to senders?
We would likely need to formalize a process to manually input an ack/nack from a recipient who responds via email/phone/other.
If a recipient agreed to send responses, how do we handle when a transmitted message is not responded to at all - is that an error?  Trigger a followup?

Seems like a good infrastructure thing to implement for both senders and receivers.

carlosfelix2 (approved):

Issue #2161, opened 2021-09-10
Part 2: Talk to the UI team about data format


jimduff-usds:
#2138 part 2: You have worked with the UI team to understand and document the desired data structure and format the UI needs this data.

Issue #2160, opened 2021-09-10
API design Part 1: openapi Spec for API


jimduff-usds:
#2138 part 1.  An openapi (eg, swagger) doc is written for a Sender API that meets these requirements:

- return a list of basic info on all receive actions and associated reports (for admin role access only) in various data ranges (default last 30 days).
- request list of basic info on receive actions and associated reports with my org only in various date ranges (default last 30 days)
- request basic info on a single receive action, and its received report. Special emphasis on the action_response json.

Issue #2138, opened 2021-09-10
Design a REST Sender API endpoint supporting  list of reports, error/warning json


jimduff-usds:
### Done when:
1. #2160 An openapi (eg, swagger) doc is written for a Sender API that meets these requirements:
- return a list of basic info on all _receive_ actions and associated reports (for admin role access only) in various data ranges (default last 30 days).  
- request list of basic info on _receive_ actions and associated reports _with my org only_ in various date ranges (default last 30 days)
- request basic info on a single receive action, and its received report.  Special emphasis on the action_response json.

2. #2161 You have worked with the UI team to understand and document the desired data structure and format the UI needs this data.

3.  #2162 Generate tickets for follow-on work to do the actual coding.   Think carefully about tests needed.   Remember that, when the time comes, @ronaldheft-usds must review the code since it involves auth features.

REMOVED FROM TICKET FOR NOW:  4. Work with @ronaldheft-usds to figure out how we need to set up the roles for this in Okta.  We currently add roles for receiver organizations.  Will this be any different for sender organizations?
REMOVED FROM TICKET FOR NOW:  5.  Work with @anshulkumar-usds on the process whereby a Sender user is given Okta creds.  (Should be very similar to current process for giving Receiver user creds in Okta)REMOVED FROM TICKET FOR NOW:  2. You have Studied and understand the Authentication code as used in the Settings APIs and in HistoryFunction.kt.  Note:  For this coding, we'll be re-using the auth code as designed for the Settings API, but with the organization claim included, as is done in HistoryFunction.kt.  (@ahay-agile6 had done work to combine these, but I had to remove it due to too many folks coding in HistoryFunction.kt at the same time)  

Note:  All the queries here involve joins on the ACTION and REPORT_FILE tables.

Note: we already have a documented openapi spec for the reports/ endpoint.

Note:  The current HistoryFunction.kt operates as a _Receiver_ API, that is, it provides data useful to Receivers.   This task is to design a new _Sender_ API.  Yes, there's lots of overlap - both use the same underlying database REPORT_FILE table, for example.  My intention is to eventually replace HistoryFunction.kt, and expand this new API to work for both Senders and Receivers.

PR #2394, opened 2021-09-29
Document New Sender/Submission History API

derekkinsman:
This PR ...

Documents several new API end points and filter options for the Submission History.

Test Steps:

## Changes
- Adds four new endpoints: `/submissions`, `/submissions/multorg`, `/submissions/{reportid}`, `/submissions/{organizationName}`
- Adds twelve new parameters: `offset`, `limit`, `dateStart`, `dateEnd`, `year`, `month`, `organizations`, `recipients`, `sender`, `warnings`, `errors`, `receipts`.

## Fixes
- #2138
- #2160 
- #2161 

## To Be Done

    - new endpoint A
    - new endpoint B


prime-router/docs/openapi.yml
```diff
@@ -9,6 +9,380 @@ info:
   version: 0.2.0-oas3
```
TomNUSDS, line 9: Don't forget to bump the version number.

prime-router/docs/openapi.yml
```diff
@@ -9,6 +9,380 @@ info:
...
+  # ?sender=ARRAY
+  # ?warnings=BOOLEAN
+  # ?errors=BOOLEAN
+  # ?receipts=BOOLEAN
+  #
+  #
+  #
+  #
+  # 30 days (based on where request orignates)
+  # pagination?
```
TomNUSDS, line 38: Think we can do pagination using `offset` and `limit` (aka pagesize) in the same call?

prime-router/docs/openapi.yml
```diff
@@ -9,6 +9,380 @@ info:
...
+  # ?dateEnd=STRING (YYYY-MM-DD | default: Today)
+  # ?year=INT (YYYY | default: ALL)
+  # ?month=INT (MM | default: ALL | requires ?year)
+  # ?organizations=INT (default: ALL)
+  # ?recipients=ARRAY
+  # ?sender=ARRAY
+  # ?warnings=BOOLEAN
+  # ?errors=BOOLEAN
+  # ?receipts=BOOLEAN
+  #
```
TomNUSDS, line 33: Another useful feature for client the client UI is to have something like a `?getlength=true` that just returns the `count()`
(e.g. It's used when building the pagination UI to show the total number of pages.)

prime-router/docs/openapi.yml
derekkinsman: How would you want to handle that? Like `/?pagesize=N`? Or would you want to set the page size as a setting and then do `/endpoint/1/`, /endpoint/2/`, /endpoint/{PAGENUMBER}/`

TomNUSDS: Good question. I'm not sure what is considered standard, so I don't have a strong opinion on the exact syntax. Jim Duff's opinion is probably valuable here.

I would like to double-check that the client can pass **both** `offset` and `pagesize` in the same call. 

When doing pagination, the client UI typically looks like this:
image

When the user clicks on Page 5, it would call the API passing `offset=5` **and** `pagesize=25`. The server then does the math of 25*5 to get the actual row start offset, then does a query limit of 25 to only return the data that's going to be displayed.

derekkinsman: Ah gotcha. Yeah so as is you could do `/endpoint?offset=N&limit=25&anotherQuery=N` we could rename it /endpoint?page=N&pagesize=N`. That way `page * pagesize = offset` is really what you're doing. `offset` generally indicates how many items you skip. But yeah, definitely can chain queries.

TomNUSDS: Good point about offset (as an absolute) vs page (which is dependant on the pagesize).

I think Java Spring uses "page" and "size" [see this article - but have an ad blocker](https://www.baeldung.com/rest-api-pagination-in-spring). 
But the client can also do the math and give an absolute "offset" and "limit". 

Defining _which_ approach is being used is probably the important part. 

Interested to hear what others think.

prime-router/docs/openapi.yml
```diff
@@ -9,6 +9,380 @@ info:
...
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Report'
+        '400':
+          description: Bad Request
+        '401':
+          description: Unauthorized
+
+  /submissions/{reportID}:
```
jimduff-usds, line 256: Note that failed submissions do not have a reportId, but they would have an actionId.  

A query like this could support both failed and successful submissions:
`/submissions/{actionID}   (or maybe /submissions/action/{actionID} or something like that)`
because, given the actionID, you can find the one submitted report associated with it.

However, our current json response only returns the reportId, not the actionId.   So we also need to support this:

`/submissions/{reportID} or maybe /submissions/report/{reportID} or something like that`

... and because each report FKs to a single Action, this can find the action as well.

All this is a long-winded way of saying, we should support queries both on actionIDs and reportIDs.  In both cases they return the json response string.

jimduff-usds (commented):
I think there's enough here to start....

derekkinsman:
@TomNUSDS @jimduff-usds updated based on your feedback. Going to start putting code together for this.

Initial API code will return all the rows, handle authentication for a non-admin client/customer, and have queries needed for pagination. Additional parameter queries (date ranges, warnings, errors, &c) will come later.

prime-router/docs/openapi.yml
```diff
@@ -9,6 +9,399 @@ info:
   version: 0.2.0-oas3
 paths:
 
+  # The sender end-point handles receive actions.
+  #
+  # END-POINTS
+  # /submissions
+  # /submissions/multorg
```
RickHawesUSDS, line 16: Typo `multorg`?

prime-router/docs/openapi.yml
```diff
@@ -9,6 +9,399 @@ info:
...
+  # 
+  #
+  /submissions:
+    get:
+      description: Return a list of all basic info on all receive actions and associated reports. Must have admin access.
+      security:
+        - OAuth2: [ system_admin ]
+        - ApiKeyAuth: []
+      parameters:
+        - in: query
```
RickHawesUSDS, line 62: In general, offset pagination doesn't work well for APIs for logs and other data that are changing. I suggest that you change this to a cursor-based pagination scheme. There are many articles explaining the reasoning, but this StackOverflow post explains this a little bit. https://stackoverflow.com/questions/18314687/how-to-implement-cursors-for-pagination-in-an-api. For examples of cursor-based APIs consider the Slack or Okta APIs.

prime-router/docs/openapi.yml
```diff
@@ -9,6 +9,399 @@ info:
...
+              schema:
+                $ref: '#/components/schemas/Report'
+        '400':
+          description: Bad Request
+        '401':
+          description: Unauthorized
+
+
+  /submissions/multiorg:
+    get:
```
RickHawesUSDS, line 164: I'm not convinced that `multiorg` is needed if it is the same as the `submissions` org. From a resource name perspective, it seems redundant.

RickHawesUSDS (changes requested):
Let me know if my comments make sense.

prime-router/docs/openapi.yml
TomNUSDS: Agreed that cursors are a more robust approach for data that changes frequently and can be very large (there's a reason DBs use them).

My primary focus is on end-user usability that these features expose since I'm on the client side. Cursor based client UI lends itself to things like cloud log UIs and "infinite scrolling" lists. Often, to find anything useful with this approach, search and/or filter features are required from the start (because the user cannot jump to "page N"). It's typically less novice-user-friendly.

(Speaking of Search vs Filter: are there any expectation that deep searches might become a requirement versus just filtering on the metadata?)

We can drop my suggestion of using pagination with offsets and page sizes and having a call to get the total length.

We should plan the React client's features accordingly. Filtering by date will probably be required much sooner. I can discuss our UX options with @drew-usds

PR #2443, opened 2021-09-30
Fix blank page when there's a network error the loading the table

TomNUSDS:
- Network errors include okta permissions issues and cors errors.
- I keep hitting (and having to fix locally) typescript errors in Upload.tsx


kevinhaube (approved):
:+1: lgtm

Issue #2270, opened 2021-09-21
Allow a mapper mapper to  run even when field data is present


carlosfelix2:
As a schema developer, I want to be able to specify if a mapper is to always run even when a field has data. Current mappers only run if a filed has no data. This change will allow further manipulation of a raw data value without having to create a new field to hold a converted value. A disadvantage of this method is that the original raw value will not be saved in ReportStream. Also note that by default the mappers will not run if data is present (current design).

Design considerations:
You must note that mappers are used in various places, like the serializers, deserializers and translation, so it is recommended to refactor the code with this logic to a central location (e.g. Element.kt).

Acceptance criteria:
1. ReportStream shall run a configured mapper even when data is present in a field if the always to run flag in the schema is set.
2. ReportStream shall run a configured mapper on a blank field only if the always to run flag in the schema is unset.
3. The always to run flag defaults to unset if not specified.

carlosfelix2:
As a schema developer, I want to be able to specify if a mapper is to always run even when a field has data.  Current mappers only run if a filed has no data.  This change will allow further manipulation of a raw data value without having to create a new field to hold a converted value.  A disadvantage of this method is that the original raw value will not be saved in ReportStream.  Also note that by default the mappers will not run if data is present (current design).

Design considerations:
You must note that mappers are used in various places, like the serializers, deserializers and translation, so it is recommended to refactor the code with this logic to a central location (e.g. Element.kt).

Acceptance criteria:
1. ReportStream shall run a configured mapper even when data is present in a field if the always to run flag in the schema is set.
2. ReportStream shall run a configured mapper on a blank field only if the always to run flag in the schema is unset.
3. The always to run flag defaults to unset if not specified.

whytheplatypus:
This makes sense and can be broken down further. The problem is that there should be things that can run to handle a `default` value, meaning the field is blank. And there are things that can run to modify an existing value. Rather than add a flag maybe making those use cases distinct will make it easy to deal with.

```yml
- default: <function names / args, or straight value>
- mappers:
    - name: mapper1
      args: 
        arg1: hello
        arg2: world
```
or go further and say the field can have either a `default` or a `default_function` or something like that.

Allowing it to be very clear that a `default` is used only if a field is blank, while a `mapper` (probably should be renamed to transform) runs only of this field is present

PR #2444, opened 2021-09-30
2270 Allow mappers to run regardless of field value

carlosfelix2:
This PR adds a configuration parameter to the schemas to allow for a mapper to run regardless if the field has a value or not.  Previously mappers could only run if no value was present.  These changes also migrate the logic to decide when to use mappers and default values to the Element class as that logic was present in three different places.  Note the Translator logic to use a mapper is different as it looks for input values in both input and output reports.

When using the new property, the YAML will look as follows:
```
- name: abnormal_flag
    type: CODE
    valueSet: hl70078
    natFlatFileField: Abnormal_flag
    hhsGuidanceField:
    hl7Field: OBX-8
    mapper: obx8(test_result)
    mapperOverrideValues: true
```

Test Steps:
1. Run the unit and integration (`gradle test testintegration`) tests and verify that there are no errors.

## Changes
- Added mapperAlwaysRun parameter to schema config
- Moved logic on when to use mappers and default to Element class.
- Made some minor cleanups

### Testing
- [x] Tested locally?
- [x] Added tests?

### Process

## Fixes
- #2270


carlosfelix2:
Does this change require an addition to the release notes?

prime-router/src/main/kotlin/Element.kt
```diff
@@ -55,6 +55,7 @@ data class Element(
     val maxLength: Int? = null, // used to truncate outgoing formatted String fields.  null == no length limit.
     val default: String? = null,
     val mapper: String? = null,
+    val mapperAlwaysRun: Boolean = false,
```
whytheplatypus, line 58: The use case for this is to override an existing value, the name can make that clear. Something that makes it clear that the mapper will `override` any existing value

whytheplatypus (changes requested):

prime-router/src/main/kotlin/Element.kt
carlosfelix2: How about mapperOverridesValue?

whytheplatypus (approved):
I like the cleanup, feels like a good step in the right direction.

prime-router/src/main/kotlin/Element.kt
```diff
@@ -759,6 +766,72 @@ data class Element(
         return codeValue?.code
     }
 
+    /**
+     * Determines if an element needs to use a mapper given the [elementValue].
+     * @return true if a mapper needs to be run
+     */
+    fun useMapper(elementValue: String): Boolean {
+        val overrideValue = mapperOverridesValue != null && mapperOverridesValue
+        return mapperRef != null && (overrideValue || elementValue.isBlank())
```
jimduff-usds, line 775: it would be weird if the mapperOverrides flag is true but there is no mapper.

Moving to the "mappers always run always" philosophy would help clean this up, but this is a step in that direction.

prime-router/src/main/kotlin/Element.kt
```diff
@@ -759,6 +766,72 @@ data class Element(
...
+                }
+            }
+            // Only overwrite an existing value if the mapper returns a string
+            val value = mapperRef?.apply(this, args, valuesForMapper)
+            if (!value.isNullOrBlank()) {
+                retVal = value
+            }
+        }
+
+        // Finally, add a default value or empty string to elements that still have a null value.
```
jimduff-usds, line 821: Interesting.   The default applies _after_ the mapper.   I would have thought the default would apply _before_ the mapper, then the mapper might do further transforms on it (or on the external/incoming value if there is one).

But I haven't thought about this in detail.

prime-router/src/main/kotlin/Report.kt
```diff
@@ -604,25 +605,29 @@ class Report : Logging {
         }
     }
 
+    /**
+     * The second pass runs any mappers as needed for the given [toElement] and the [pass1Columns] data from the first
+     * pass.  Mappers will use data from either schema as inputs.
+     * @param mapping the mapping
+     * @return the mapped column
+     */
```
jimduff-usds, line 613: In the future @whytheplatypus Theory world, can we get rid of this first pass / second pass glop, and replace with an ordered explicit list of mappers?  I hope so...

prime-router/src/main/kotlin/serializers/Hl7Serializer.kt
```diff
@@ -233,7 +233,7 @@ class Hl7Serializer(
         val errors = mutableListOf<String>()
         val warnings = mutableListOf<String>()
         // key of the map is the column header, list is the values in the column
-        val mappedRows: MutableMap<String, MutableSet<String>> = mutableMapOf()
+        val mappedRows: MutableMap<String, String> = mutableMapOf()
```
jimduff-usds, line 236: It seems wrong to me that you have to make changes to both Hl7Serializer and CsvSerializer.

Again, in the heavenly @whytheplatypus Theory world, it seems like we'd transform to FHIR as early as computerly possible, then do all transforms in the FHIR world.   Deserializers and Serializers would be blissfully unaware of such things as mappers/transforms.    True?

jimduff-usds (approved):
Nice.

I have one minor additional suggestion:  I wish there was just an example of an actual chink of yaml in the ticket or in the PR, just showing the new mapperOverridesValue as its meant to be used.    As is, I had to reverse engineer the code in my head to what the yaml looks like --- and then I realized its quite simple.   So, an example yaml would be great.

prime-router/src/main/kotlin/serializers/Hl7Serializer.kt
carlosfelix2: Yeah, the problem is how the two serializers run through the same process with different logic.

PR #2461, opened 2021-09-30
Use utf-8 chars

sean-usds:
The hyphen characters are not coming through correctly in the rolled up error messages for the csv uploader:
image

This problem only happens on staging, I am not able to repro locally. Switching to using the unicodes for these characters may solve the problem. I'll have to see what it looks like in staging.


MauriceReeves-usds (approved):

kevinhaube (approved):
:candy: :jack_o_lantern: Lgtm!

PR #2465, opened 2021-10-01
Rename V18 migration to V21 to ensure it runs

whytheplatypus:
This PR ...

Renames the V18 migration to V21 to ensure it runs on existing environments.


jimduff-usds (approved):
I've always wanted to do a valid/needed PR with exactly 1 byte changed.   But with zero bytes changed?! - an impressive feat indeed!!

Issue #2252, opened 2021-09-20
Do a "diff" between SR schema and Direct schema to find data needed by states


jimduff-usds:
Do a virtual “diff” between the SimpleReport schema and ‘direct’ schema, and then figure out minimal additional fields we should add to ‘direct’ to mimic the delightful quality that SR delivers.  Review with Jim and a few folks from the Ops side of things (Brick, Adrian)

Then make those changes, communicate to customers, and push it out.

anshulkumar-usds:
@jimduff-usds @jbiskie isn't this pending deployment?

PR #2471, opened 2021-10-01
Update direct schema to use reporting_facility_name

jbiskie:
This PR modifies the direct schema to use reporting_facility_name instead of reporting_facility.  The outcome is that MSH-4 is now populating for the states where it was previously blank.

Test Steps:
1. Create a set of fake records for a direct sender.  >=56 covers all states and territories. `./prime data --input-fake 56 --input-schema direct/imagemover-covid-19 --output ./fake-imagemover-56rec.csv`
2. From master, run records through local RS ` ./prime data --input fake-imagemover-56rec.csv --input-schema direct/imagemover-covid-19 --output-dir junk --route`
3. Switch to this branch and run records through again
4. Compare outputs

Test results:
1. For states where MSH-4 is not defaulted, it is now populating with reporting_facility_name as desired.
2. For states where MSH-4 is defaulted, MSH-4 is still populating correctly.  Additional note (and win): NH, NJ, and OH previously populated MSH-4.1 (name) and MSH-4.2 (CLIA #) only.  After this change, MSH-4.3 ("CLIA") is also populating, which mirrors what SimpleReport does.  NJ brought this to our attention last week when it was not populating.
3. The value in reporting_facility_name is also populating PID-3.6.1, SPM-2.1.2, and SPM-2.2.2
4. **CSV receivers are unaffected, except for NM's CSV.  I have a question out in the receivers Slack channel to verify if that's actually used anymore.**

## Changes
- Changed the direct schema to use reporting_facility_name instead of reporting_facility.
- Changed the HHSProtect receiver schema to match.

### Testing
- [x] Tested locally?
- [ ] Added tests?

## Fixes
- #2252 Do a "diff" between SR schema and Direct schema to find data needed by states (partial fix, more will be coming.)

## To Be Done
- Continue with additional fixes related to #2252


carlosfelix2:
@jbiskie Minor: Change your test to use a folder in build like build/junk and add a step to create the folder.

carlosfelix2 (approved):

Issue #2462, opened 2021-09-30
Org Switcher Missing from React App


sean-usds:

kevinhaube:
The org switcher is showing and visibly working within React now (:tada:), but the implementation is a bit more tricky than that. The calls to our API require an `organization` header. To share this organization choice from the component over to the `ReportsTable` and `FacilitiesTable` would require insane prop piping, OR ... the `useContext()` hook. :sunglasses: The coding of this is going to introduce something in the JSX hierarchy of views that looks like this: `<GlobalStateContext.Provider>` <- this is a wrapper; anything in it can use the context. I chose this instead of a full Redux package because our need for global state is minimal and does not require a full application state management system.

PR #2472, opened 2021-10-01
React Org Dropdown

kevinhaube:
This PR introduces `GlobalContextProvider` and the org switcher for admins.

Test Steps:
1. pull `kevinhaube/add-org-dropdown`
2. `npm install` and `npm run`
3. Press F12 to open your dev tools
4. Log in
5. In dev tools, navigate to Application > Local Storage > localhost:3000 and ensure your `global-context-org` value is your normal default org
6. (Requires React Dev Tools extension) In dev tools, navigate to Components and find `GlobalContextProvider`
7. Ensure `state` shows the correct value for organization
8. Switch orgs. The page should reload. Try this on various pages!

## Changes
- `GlobalContextProvider` now wraps the pages and provides a global state
- `GlobalContext` values are persisted in `localStorage`
- Updates around the app to utilize `GlobalContext` org value

### Testing
- [X] Tested locally?
- [ ] Added tests?

### Process

## Fixes
- #2462


frontend-react/src/App.tsx
```diff
@@ -65,25 +67,27 @@ const App = () => {
...
+                            <Route path="/login" render={() => (<Login config={oktaSignInConfig} />)} />
+                            <Route path="/login/callback" component={LoginCallback} />
+                            <AuthorizedRoute path='/daily-data' authorize={PERMISSIONS.RECEIVER} component={Daily} />
+                            <AuthorizedRoute path='/upload' authorize={PERMISSIONS.SENDER} component={Upload} />
+                            <SecureRoute path="/report-details" component={Details} />
 
-                        {/* Handles any undefined route */}
-                        <Route render={() => (<ErrorPage code={CODES.NOT_FOUND_404} />)} />
+                            {/* Handles any undefined route */}
+                            <Route render={() => (<ErrorPage code={CODES.NOT_FOUND_404} />)} />
```
sean-usds, line 86: nit: white spaces

frontend-react/src/components/header/HIWDropdown.tsx
```diff
@@ -0,0 +1,66 @@
+import { Link, Menu, NavDropDownButton } from "@trussworks/react-uswds";
+import { useEffect, useState } from "react";
+
+const HIWDropdown = () => {
```
sean-usds, line 4: nit: preference for descriptive components like `HowItWorksDropdown` since HIW might not be a clear acronym

frontend-react/src/components/header/OrgDropdown.tsx
```diff
@@ -0,0 +1,42 @@
...
+    const { updateOrganization } = useGlobalContext();
+
+    let orgs = useResource(OrganizationResource.list(), {
+        sortBy: undefined,
+    }).sort((a, b) => a.name.localeCompare(b.name));
+
+    let setValue = (e: any) => {
+        setOrg(e)
+        updateOrganization(e)
+        window.location.reload()
```
sean-usds, line 18: instead of reloading, thoughts on sending users to the home page? if we reload and their on a report page, they'll probably get and error/crash. open to other ideas!

frontend-react/src/components/header/ReportStreamHeader.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+        if (permissionCheck(PERMISSIONS.SENDER, authState)) {
+            itemsMenu.splice(1, 0,
+                <Link href="/upload"
+                    key="upload"
+                    data-attribute="hidden"
+                    hidden={true}
+                    className="usa-nav__link">
+                    <span>Upload</span>
+                </Link>
```
sean-usds, line 43: These need to be NavLinks now! if you rebase and re-copy-and-paste you should be good

frontend-react/src/App.tsx
```diff
@@ -24,15 +24,20 @@ import { permissionCheck, reportReceiver } from "./webreceiver-utils";
...
-
     const history = useHistory();
-    const customAuthHandler = () => {
+
+    const customAuthHandler = (): void => {
         history.push("/login");
     };
+
+    const handleIdle = (): void => {
+        if (OKTA_AUTH.authStateManager._authState.isAuthenticated) OKTA_AUTH.signOut();
```
sean-usds, line 39: we'll want to clear the global context here just to be safe

sean-usds (changes requested):
Overall the approach lgtm! Could small changes requested 🤏

frontend-react/src/App.tsx
kevinhaube: Good catch! Totally forgot we logout more than one place. :joy:

frontend-react/src/components/header/ReportStreamHeader.tsx
kevinhaube: Will do!

frontend-react/src/components/header/OrgDropdown.tsx
kevinhaube: I wasn't getting any crashes locally, is there something about the staging/prod env that might prompt an error?

frontend-react/src/utils/UserUtils.ts
```diff
@@ -0,0 +1,20 @@
+import { OktaAuth } from "@okta/okta-auth-js";
+import { GLOBAL_STORAGE_KEYS } from "../components/GlobalContextProvider"
```
TomNUSDS, line 2: I like directly including the keys and configs instead of passing them around!

Style suggestion: include trailing `;` for statements for consistency.

(In fact, we need to fix the `login.txs` so it directly includes the config data instead of loading it externally and passing it into the component.)

frontend-react/src/utils/UserUtils.ts
```diff
@@ -0,0 +1,20 @@
...
+import { oktaAuthConfig } from "../oktaConfig";
+
+const OKTA_AUTH = new OktaAuth(oktaAuthConfig);
+
+function clearGlobalContext() {
+    for (let key in localStorage) {
+        if (key.includes(GLOBAL_STORAGE_KEYS.GLOBAL_BASE)) localStorage.removeItem(key)
```
TomNUSDS, line 9: We should probably always use `{ }` for `if` blocks even if they are one line. I think when we enforce a style guide, this will be an error.

TomNUSDS (approved):
LGTM. 

Looks like you've already started fixing stuff from Sean's review, so I pulled them out (hopefully there's not too much overlap).

Mostly, I'm just just have linter nitpicks that will eventually be done by a stricter linter.

frontend-react/src/components/header/OrgDropdown.tsx
kevinhaube: OHHHH I get what you're saying; if it's on `/report-details/...` it'll error out. I think catching that and bumping them back to `/daily` would be a better experience than all the way home, but we'll have to build out that check for any detail-related pages on the sender side as well (i.e. bumping them back to `/upload` or wherever the base of their experience lies).

frontend-react/src/utils/UserUtils.ts
kevinhaube: I did have to change it to receive an `OktaAuth` param because what the line `const OKTA_AUTH = new OktaAuth(oktaAuthConfig);` did was just create a new auth on top of our current, thus, we had two and logged out of the one. However, getting our config from any component really only takes a quick `useOktaAuth()` hook call, so it's not _so_ bad. I'd still love to find a way to refactor this so we could just call `logout()`.

frontend-react/src/utils/UserUtils.ts
kevinhaube: Fixed in new commit! :+1:

frontend-react/src/components/header/OrgDropdown.tsx
sean-usds: This works! It might start to get gnarly if/when there are a lot of pages we need to make exceptions for. Can cross that bridge when we get to it though

sean-usds (approved):
thanks for making the changes! 🎸

PR #2476, opened 2021-10-01
Turn on noImplicitAny in typescript config

sean-usds:
- Set `noImplicitAny: true`
- Fix types where solution is simple/obvoius
- Add `// @ts-nocheck // TODO: fix types in this file` to top of the file where further changes are needed


kevinhaube (approved):
:fallen_leaf: lgtm!

PR #2508, opened 2021-10-04
Use NavLink for report details

sean-usds:
Switch the link to /report-details to instead be handled within the single page app so that a page request is not required.


kevinhaube (approved):
:money_mouth_face: lgtm!

PR #2512, opened 2021-10-04
Palo Alto: Enable for Lower Environments

ronaldheft-usds:
This PR ...

Test Steps:

## Changes
-

### Testing
- [ ] Tested locally?
- [ ] Added tests?

### Process

## Fixes

## To Be Done
- #issue

    - new endpoint A
    - new endpoint B

### DIFF
--- a/operations/app/src/modules/database/main.tf
+++ b/operations/app/src/modules/database/main.tf
@@ -51,21 +51,21 @@ module "postgres_private_endpoint" {
   create_dns_record  = true
 }
 
-//module "postgres_server_private_endpoint" {
-//  source             = "../common/private_endpoint"
-//  resource_id        = azurerm_postgresql_server.postgres_server.id
-//  name               = azurerm_postgresql_server.postgres_server.name
-//  type               = "postgres_server"
-//  resource_group     = var.resource_group
-//  location           = var.location
-//  endpoint_subnet_id = data.azurerm_subnet.endpoint_subnet_east.id
-//  create_dns_record  = false
-//
-//  depends_on = [
-//    # Prevent unexpected order-of-operations by placing a hard dependency against the current private endpoint
-//    module.postgres_private_endpoint
-//  ]
-//}
+module "postgres_server_private_endpoint" {
+  source             = "../common/private_endpoint"
+  resource_id        = azurerm_postgresql_server.postgres_server.id
+  name               = azurerm_postgresql_server.postgres_server.name
+  type               = "postgres_server"
+  resource_group     = var.resource_group
+  location           = var.location
+  endpoint_subnet_id = data.azurerm_subnet.endpoint_subnet_east.id
+  create_dns_record  = false
+
+  depends_on = [
+    # Prevent unexpected order-of-operations by placing a hard dependency against the current private endpoint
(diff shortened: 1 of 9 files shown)


.github/workflows/build_docker.yaml
```diff
@@ -22,9 +23,11 @@ jobs:
               - 'operations/docker-compose.yml'
               - 'operations/Dockerfile'
               - '.github/workflows/build_docker.yml'
+            dnsmasq:
+              - 'operations/dnsmasq/**'
 
   build_tfcli:
-    name: Build Terraform CLI
+    name: Build dnsmasq
```
cwinters-usds, line 30: I think this should remain "Terraform CLI"

RickHawesUSDS (commented):
This file probably should be deleted, since it doesn't do anything because the settings in `test` are in the DB. Makes me ask does there need to be a settings update to point all sftp transports to `sftp.prime.local`

operations/dnsmasq/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+# Install dnsmasq
+RUN apk --no-cache add dnsmasq
+
+# Expose the DNS port
+EXPOSE 53/udp
+
+# Place the environment's config in the config directory
+ARG AZ_ENV
+COPY config /tmp/config
+RUN cp -r /tmp/config/${AZ_ENV}/. /etc/dnsmasq.d/
```
cwinters-usds, line 12: should we delete `/tmp/config` after doing this copy? just to make sure configurations from other environments aren't on here (even in a weird place)

ronaldheft-usds: I like it! Added!

FYI, this method of baking the configuration into the container, I dislike. Doing so I am choosing implementation speed over maintainability.

I think the optimal solution would involve an environment-agnostic Docker container with the configuration provided by a bind mount. Just doing so involves a few implementation details that need be thought through, that I didn't want to tackle at this time.

Namely, storage accounts (which would be used for the bind mount) are created after our core network infrastructure. Accessing the storage accounts needs the VPN, which then needs this DNS server. I think to get to bind mounts we'd need to create a separate Terraform module.

ronaldheft-usds: True, but this container is used only for our VPN connection. DNS resolution on Azure is mess. This document explains how it works in various situations: https://docs.microsoft.com/en-us/azure/virtual-network/virtual-networks-name-resolution-for-vms-and-role-instances

TLDR;

The Azure DNS server `168.63.129.16` will return our private DNS records when resolved from within the VNET. However, when resolved outside the VNET, for example, through the VPN Gateway, the Azure DNS server is unable to identify our DNS queries as from the VNET, so the public query records are used.

To resolve that, Microsoft recommends setting up a DNS forwarder in your VNET, which we are doing via dnsmasq. By setting up a forwarded, we have a VNET IP we can point the VPN clients at and Azure's VPN server sees the requests coming from our VNET and returns the private IP responses.

Prior to this PR, we were good with just forwarding functionality, but now we need to have some control over the DNS results due to multiple VNETs and only one of them being accessible via the VPN (we can't access the CDC network via our VPN).

PR #2386, opened 2021-09-28
Start using `yarn` instead of `npm`

TomNUSDS:
Switch the package management over to yarn.

Issue #2011, opened 2021-08-27
Refactor GitHub Actions to share code


ronaldheft-usds:
GitHub Actions released the ability to re-use YAML in actions with a feature called composition: https://github.blog/changelog/2021-08-25-github-actions-reduce-duplication-with-action-composition/

Our release action could greatly benefit from this, as each environment copies and pastes the exact same code. We may be able to share GitHub Action code among other actions as well.

TomNUSDS:
Another resource: https://docs.github.com/en/actions/learn-github-actions/reusing-workflows

ronaldheft-usds:
@bjohnson-fearless When you have findings regarding this, please post them here so we can discuss as a group. Thanks!

PR #2513, opened 2021-10-04
change `npm` to `yarn` for build_frontend_react's package management

TomNUSDS:
We still want to lock npm at version 14, but use the yarn package manager.

See SimpleReport's frontend github action as a reference:
https://github.com/CDCgov/prime-simplereport/blob/main/.github/actions/build-frontend/action.yml

Related change: https://github.com/CDCgov/prime-reportstream/pull/2386


TomNUSDS:
@sean-usds @kevinhaube FYI

sean-usds (approved):
lgtm! 🧶

TomNUSDS:
OK. I updated the `release.yml` section for the react app even though I don't think it's being used for the release yet.

@ronaldheft-usds Recently, Github actions added support for calling other actions by reference to reduce duplication.
https://github.blog/changelog/2021-08-25-github-actions-reduce-duplication-with-action-composition/

kevinhaube (approved):
Nice :+1:

ronaldheft-usds (approved):
Looks great!

We are actually deploying the react app to staging, so good call in updating the `release.yml` workflow.

Thanks for also pointing out the recently released ability to de-duplicate code in GitHub Actions. Our release workflow definitely needs it! Tracking it as https://github.com/CDCgov/prime-reportstream/issues/2011

PR #2515, opened 2021-10-04
Get uswds trussworks scss build working

TomNUSDS:
In `package.json`
- Reworked the build system to use `craco` (similar to what SimpleReport uses).
- Removed unused gulp dependencies 
- Removed unused `node-sass` (to avoid confusion since it's no longer used)
- `uswds-gulp` became `uswds` in package
- `react-scripts build` -> `craco build`

The layout
```
|-- ./App.tsx
|-- ./global.scss
|-- ./index.tsx
`-- ./styles
    |-- ./styles/_print.scss
    |-- ./styles/_reportstream.scss
    |-- ./styles/_uswds_overrides.scss
    `-- ./styles/_uswds_settings.scss
```

| File               | Description                                                                            |
|--------------------|----------------------------------------------------------------------------------------|
| global.scss        | Simply includes other resources in the correct order. ONLY referenced from `index.tsx` |
| _print.scss        | Styles related to printing pages to pdf.                                               |
| _reportstream.scss | Styles for Report Stream specific components/elements                                  |
| _uswds_overrides.scss | This is processed after USWDS libraries and Trusswork's React library. It can be use to override things from those libraries if they can't be controlled by a theme setting |
| _uswds_settings.scss | USWDS theme specific variables |

Note: `theme/uswds-theme-general` and the other "USWDS theme include" files are loaded directly from USWDS distribution. This is done for forward compatibility. _When_ USWDS modifies these files, merging and maintaining them will be painful. Any variable that are changed should be copied out of these files and put into `_uswds_overrides.scss`

This approach also makes it trivial to see what's we've changed from the USWDS defaults.

TODO:
- Need to add details about the system to a README or something.


TomNUSDS:
@drew-usds Can you check out this branch and see if the styles seem like they are working for you?

drew-usds (approved):
👍 niiiice. Yarn worked without a hitch. I added a commit that tested out using uswds variables in other stylesheets ✔️ , and tested out changing `_uswds_settings.scss` ✔️. I also removed some duplicate styles for html, head, body, etc that were in both `_reportstream.css` and `_uswds_overrides.scss` since I was already poking around.

TomNUSDS:
@drew-usds Merged into main. You'll get an error if you try to update this branch.

PR #2516, opened 2021-10-04
Use a NavLink where possible

sean-usds:
- NavLink for going back to /daily-data from report page
- NavLink for going to /about
- NavLinks in the /how-it-works/data-download-guide page

This will keep all navigation within the single page app and prevent unnecessary page requests


TomNUSDS (approved):
LGTM. And using yarn!

PR #2529, opened 2021-10-05
Update global.scss

TomNUSDS:
USWDS has a variable to put the theme compiler into verbose mode. I had it set to `true` while doing scss work, but it can be `false`.


drew-usds (approved):
👍

PR #2558, opened 2021-10-07
Change idle timeout from 30 minutes (default) to 60 seconds

ronaldheft-gov:
https://github.com/brettwooldridge/HikariCP#gear-configuration-knobs-baby

### DIFF
--- a/prime-router/src/main/kotlin/azure/DatabaseAccess.kt
+++ b/prime-router/src/main/kotlin/azure/DatabaseAccess.kt
@@ -768,6 +768,8 @@ class DatabaseAccess(private val create: DSLContext) : Logging {
             // connection are closed
             // after about 10 minutes
             config.maxLifetime = 180000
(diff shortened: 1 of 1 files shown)


prime-router/src/main/kotlin/azure/DatabaseAccess.kt
```diff
@@ -768,6 +768,8 @@ class DatabaseAccess(private val create: DSLContext) : Logging {
             // connection are closed
             // after about 10 minutes
             config.maxLifetime = 180000
+            // Release connections after 60 seconds of idle
+            config.idleTimeout = 60000
```
whytheplatypus, line 772: The docs have a very confusing line:

> Idle connections will not be retired once the pool reaches `minimumIdle` connections.

Does this mean that once we've got 2 idle connections we'll just keep accumulating connections??

Issue #2572, opened 2021-10-11
Enable receiver-transform errors to be properly surfaced back to the Sender; get rid of internal data format.


jimduff-usds:
### Background

Currently the Batch step does the final transformation of data from our in-memory format into the format desired by the Receiver.    However, these transforms can create errors and warnings  (See #2354 for example of this).   The question is:  what to do with these errors and warnings.    The solution for #2354 is less than ideal - a message logged, labelled as an "Exception" so that our notification tools can pick it up.

Note:  Prior to the fix for #2354, the solution was even worse:  throw a real Exception, which would prevent the data from going to the state, just because of one invalid value.

### How to do this better?

A better solution would be to expose the error/warning to the sender.   In order to do this, the simplest approach would be to put all code that can possible produce data related errors and warnings into a single microservice.

### Design Implications

However, in order to do that, that service (I'll call it ReportFunction for now) must create data in the final format desired by Receivers, instead of our internal /intermediate format.

However, the internal / intermediate format is useful, because it allows for us to change versions of schemas with less likelihood of errors.

So, before we do any of the above changes, we first need to enable versioning of schemas, and when we do that, we might as well put them in the database where they belong.  That line of work is ticketed in #2571, and needs to be completed before this work can be started.

### Work Items

The work here is to
- modify the Batch function so that it no longer parses or "understands" the data and therefore can never generate data-related errors any more.
- modify the ReportFunction so that it creates data in the final receiver's format, rather than the internal format.   This allows us to collect errors in one place, and they should be put in the standard json response back to the Sender.
- Test/QA
- A good way to test is to look at #2354, and recreate that error condition as documented in a comment there.    Those $alt table unknown-value errors should now appear in the Json response.

Issue #2571, opened 2021-10-11
Write a proposal to enable modifying and versioning of schemas without having to do PRs


jimduff-usds:
### Background and Use Cases

Currently schemas are stores in files, which makes them hard to modify rapidly.   This story is to put schemas into the database, so they can be modified without needing a PR, and so they can be versioned.

As part of this, we need to allow multiple versions of schemas in use in ReportStream at the same time.  This is very much needed for up to 24 hours during a schema upgrade, when data may be queued for Batch in an old schema format, while newer data uses the new schema format.

[Future: Perhaps eventually versions of schemas could also be tied to Receivers, so that one receiver could be sending an old version of the data while another Receiver can be tested using the newer schema.]

So, there are two main goals here:  version schemas, and get them into the database.  You could, if you like, split these into two separate mini-projects.

### Design Considerations

The Settings classes and/or the new TableLookup feature can act as model architecture and code.   In particular, these features are API-based, they do versioning of the data, and they provide a "diff" tool, and they are designed to (someday) support a decent UI on top of the API.

There are various ways versioning could be implemented.   One is to mimic the settings table.   Another perhaps simpler way is to build the version number into the "path" or "name" of the schema, a feature that already exists, eg `direct/direct-covid-19`.

In the new design, we might have 
`direct/direct-covid-19/1` and `direct/direct-covid-19/2`.
I kinda like this way, because its simple and readable, and easily allows for any number of versions to be "active" at one time.

### What to put in the Proposal

This ticket is done when a proposal is written to meet the above use cases, and stories have been generated for the actual work.  The proposal should meet these requirements:
- proposal must describe the API, with okta security
- proposal must describe ./prime CLI commands to do CRUD on schemas, to activate/deactivate schema/versions and to do some sort of "diff" activity on schemas.
- schemas must be versioned and can be active or inactive
- [Future: it should be impossible to deactivate a schema if any WIP is associated with that schema]
- check with @whytheplatypus that the plan is compatible with the FHIR engine plans.
- have the concept of a 'default' version, which is just the latest version of a schema.  When a receiver looks up a schema, if no version is specified, it must always get the default schema.   this will give you backward compatibility with current system, which is a requirement.

### Thoughts on the Tasks

At the moment, I'm seeing these follow-on work items that you'll need to ticket:
- write this proposal
- implement the API
- end2end test suites that exercise the API
- implement the CLI
- deployment plan, including a script to do initial load of existing schemas upon deployment
[...]

PR #2573, opened 2021-10-11
Fix #2354, Exception in alt value set code causing Batch steps to fail

### LINKED ISSUE #2354: equipment_model_name not found in Pima device_type alt valueset lookup causes Exception
The Pima Schema has a field not found in other schemas, called device_type.    device_type has an alt valueset that maps from equipment_model_name to alternate test labelling strings that are useful to them.   However, as new tests have come on the market, Pima is getting equipment_model_names not found in the table, which is throwing an exception.

(NOTE: THIS INITIAL FIX HAS BEEN DONE:  Add the equipment_model_names that are causing errors to the valueset in the AZ/pima schema.  There are two of these so far -- see the detailed exceptions attached as comments below.)

THIS IS THE WORK THAT NEEDS TO BE DONE:  This should not throw an exception.  If the field is not required, (which it isn't) it should zero-out the value if the input equipment_model_name is not found.

This is a good example of why we need the table/valuesets in the database, that @carlosfelix2 is working on.

FYI @MauriceReeves-usds 

A/C
Include a unit test

jimduff-usds:
Currently the Batch step does the final transformation of data from our in-memory format into the format desired by the Receiver. However, these transforms can create errors and warnings, as we discovered with ticket #2354.  In this bug, the code throws an Exception, which kills the Batch immediately , and prevents all associated data from going to the state, just because of one invalid value.

So, obviously the Exception is Bad, but then the question is: what to do with these errors and warnings?   The solution here for #2354 is an improvement -a message logged, labelled as an "Exception" so that our notification tools can pick it up.  However, then the value is turned to an empty string, and processing continues.

Since this is a stopgap solution, I've documented a path to a proper fix.   See #2571 and #2572 .

## Test Steps:
If you go to the bug ticket and scroll way down, you'll see the Steps to Reproduce.   If you do those, then the test passes if, instead of an Exception, you now get a nasty error message like this in the logs:

## Changes
- In addition to the above fix, there's some tweaks to the "parallel" test, and a last minute tweak to our customerActive script (already run in prod, so this is just documenting after the fact)

### Testing
- [X ] Tested locally?
- [ ] Added tests?

## Fixes
- #2354 

## To Be Done
- #2571 
- #2572

### DIFF
--- a/prime-router/src/main/kotlin/Element.kt
+++ b/prime-router/src/main/kotlin/Element.kt
@@ -170,8 +170,9 @@ data class Element(
 
     val isCodeType get() = this.type == Type.CODE
 
-    val isOptional get() = this.cardinality == null ||
-        this.cardinality == ZERO_OR_ONE || canBeBlank
+    val isOptional
+        get() = this.cardinality == null ||
+            this.cardinality == ZERO_OR_ONE || canBeBlank
 
     val canBeBlank
         get() = type == Type.TEXT_OR_BLANK ||
@@ -237,10 +238,12 @@ data class Element(
     /**
      * A formatted string is the Element's normalized value formatted using the format string passed in
      * The format string's value is specific to the type of the element.
+     * The schema is passed in for logging purposes only.
      */
     fun toFormatted(
-        normalizedValue: String,
-        format: String? = null
+        normalizedValue: kotlin.String,
+        format: kotlin.String? = null,
+        schema: Schema? = null,
     ): String {
         if (normalizedValue.isEmpty()) return ""
         val formattedValue = when (type) {
@@ -268,7 +271,18 @@ data class Element(
                     // TODO Revisit: there may be times that normalizedValue is not an altValue
                     altDisplayToken ->
                         toAltDisplay(normalizedValue)
-                            ?: error("Schema Error: '$normalizedValue' is not in altValues set for $fieldMapping")
+                            ?: "".also {
+                                // Warn in the logs rather than actually throwing an exception.
+                                // This occurs in the Batch step, so no way to error back to the sender.
+                                println(
+                                    "Exception: Outgoing receiver schema ERROR:" +
+                                        " '$normalizedValue' is not in altValues set for $fieldMapping." +
+                                        " Replacing with empty-string in generated data for element $name" +
+                                        " and continuing to process." +
+                                        " Consider fixing by adding $normalizedValue to the " +
+                                        " alt valueset in schema ${schema?.name ?: ""}"
+                                )
+                            }
                     codeToken ->
                         toCode(normalizedValue)
                             ?: error(
--- a/prime-router/src/main/kotlin/cli/tests/TestReportStream.kt
+++ b/prime-router/src/main/kotlin/cli/tests/TestReportStream.kt
@@ -1146,29 +1146,33 @@ class Parallel : CoolTest() {
         options: CoolTestOptions
     ): Boolean {
         var passed = true
-        val elapsedMillis = measureTimeMillis {
+        var totalMillisAllSubmissions: Long = 0
+        val elapsedMillisTotal = measureTimeMillis {
             val threads = mutableListOf<Thread>()
             echo("Parallel Test: Starting $numThreads threads, each submitting $numRounds times")
             for (threadNum in 1..numThreads) {
                 val th = thread {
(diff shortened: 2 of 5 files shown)


prime-router/src/main/kotlin/Element.kt
```diff
@@ -268,7 +271,18 @@ data class Element(
...
+                                // This occurs in the Batch step, so no way to error back to the sender.
+                                println(
+                                    "Exception: Outgoing receiver schema ERROR:" +
+                                        " '$normalizedValue' is not in altValues set for $fieldMapping." +
+                                        " Replacing with empty-string in generated data for element $name" +
+                                        " and continuing to process." +
+                                        " Consider fixing by adding $normalizedValue to the " +
+                                        " alt valueset in schema ${schema?.name ?: ""}"
+                                )
+                            }
```
jimduff-usds, line 285: Above is the main change in this PR:   replace the error() exception with logging.

prime-router/src/main/kotlin/cli/tests/TestReportStream.kt
```diff
@@ -1146,29 +1146,33 @@ class Parallel : CoolTest() {
...
-                                ReportFunction.Options.SkipSend
-                            )
-                        if (responseCode != HttpURLConnection.HTTP_CREATED) {
-                            echo(json)
-                            passed = bad("$threadNum: ***Parallel Test FAILED***:  response code $responseCode")
-                        } else {
-                            val reportId = getReportIdFromResponse(json)
-                            if (reportId == null) {
-                                passed = bad("$threadNum: ***Parallel Test FAILED***:  No reportId.")
+                        val elapsedMillisOneSubmission = measureTimeMillis {
```
jimduff-usds, line 1156: Just wrapped this `./prime test --run parallel` test to print out more useful info on how fast it ran.

prime-router/src/main/kotlin/Element.kt
```diff
@@ -237,10 +238,12 @@ data class Element(
     /**
      * A formatted string is the Element's normalized value formatted using the format string passed in
      * The format string's value is specific to the type of the element.
+     * The schema is passed in for logging purposes only.
```
carlosfelix2, line 241: Change this comment to 
The [schema] is passed ...
This will generate the proper kdoc

prime-router/src/main/kotlin/Element.kt
```diff
@@ -268,7 +271,18 @@ data class Element(
                     // TODO Revisit: there may be times that normalizedValue is not an altValue
                     altDisplayToken ->
                         toAltDisplay(normalizedValue)
-                            ?: error("Schema Error: '$normalizedValue' is not in altValues set for $fieldMapping")
+                            ?: "".also {
+                                // Warn in the logs rather than actually throwing an exception.
+                                // This occurs in the Batch step, so no way to error back to the sender.
+                                println(
+                                    "Exception: Outgoing receiver schema ERROR:" +
```
carlosfelix2, line 278: Why not create an exception then log it with the stack trace?  May be useful.

prime-router/src/main/kotlin/cli/tests/TestReportStream.kt
```diff
@@ -1146,29 +1146,33 @@ class Parallel : CoolTest() {
         options: CoolTestOptions
     ): Boolean {
         var passed = true
-        val elapsedMillis = measureTimeMillis {
+        var totalMillisAllSubmissions: Long = 0
```
carlosfelix2, line 1149: You should add running this as part of your tests in the PR

prime-router/src/main/kotlin/serializers/CsvSerializer.kt
```diff
@@ -212,7 +212,7 @@ class CsvSerializer(val metadata: Metadata) {
                             element.csvFields.map { field ->
                                 val value = report.getString(row, element.name)
                                     ?: error("Internal Error: table is missing ${element.fieldMapping} column")
-                                element.toFormatted(value, field.format)
+                                element.toFormatted(value, field.format, report.schema)
```
whytheplatypus, line 215: Consider moving the error handling logic here as a try catch, exposing it earlier in the flow. This would also let you keep the `toFormated` function signature the same.

whytheplatypus (changes requested):

Issue #2579, opened 2021-10-12
Design the new lookup tables API to manage lookup tables in the database


carlosfelix2:
Design the new lookup tables API to manage lookup tables in the database

Acceptance criteria:
1. Document the design of the lookup table API

PR #2592, opened 2021-10-12
2579 Design lookup table API using OpenAPI spec

carlosfelix2:
This PR adds the design for the lookup table API.

**Background**
We wish to allow secure access to manipulate the lookup tables that are stored in the database.  Authenticated users will be able to manage the lookup tables and create new versions to quickly respond to data requirements.  This API will serve for a possible future UI.  This API will also be used by the PRIME CLI management commands.

**Goals**
An API that is useful for a future management UI page.
An API that is useful for on-boarding states by the ReportStream team.
An API which doesn't require specific code knowledge.
An API which allows the separation of operations by sensitivity
An API which is implementable in an Azure Function.
Require Okta authentication to use all endpoints.

**API**
A set of REST functions to manage the lookup tables as follows:
**OUTDATED - SEE OPEN API DOC**

_Fetch the list lookup tables_
HTTP Method: GET
Return: JSON array with list of table information
HTTP Status Code: 200 for successful fetch, 400 if there was an error
api/lookuptables - fetch a list of active tables
api/lookuptables?showAll=true - fetch all table including inactive

_Fetch the contents of a lookup table_
HTTP Method: GET
Return: JSON array with list of table rows
HTTP Status Code: 200 for successful fetch, 404 if the table name was not found, 400 if there was an error
api/lookuptables/<table-name> - Fetch the latest version of a table
api/lookuptables/<table-name>/<version-number> - Fetch the given version of a table

_Create a new version of a lookup table_
HTTP Method: POST
Payload: JSON array with list of table rows
Return: JSON object with new table version information
HTTP Status Code: 201 for successful creation, 400 if there was an error
api/lookuptables/<table-name>

_Activate a specific version of a table_
HTTP Method: POST
Payload: None
Return: JSON object with activated table version information
HTTP Status Code: 200 for successful activation, 404 if the table was not found, 400 if there was an error
api/lookuptables/<table-name>/<version>

## Changes
- Added lookup table API OpenAPI spec

### Testing
- [ ] Tested locally?
- [ ] Added tests?

### Process

## Fixes
- #2579

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -7,13 +7,14 @@ Test Steps:
 
 ## Changes
 - *Include a comprehensive list of changes in this PR*
--
+- *(For web UI changes) Include screenshots/video of changes*
 
 ## Checklist
 
 ### Testing
 - [ ] Tested locally?
 - [ ] Ran `./prime test` or `./gradlew testSmoke` against local Docker ReportStream container?
+- [ ] (For Changes to /frontend-react/...) Ran `npm run lint:write`? 
 - [ ] Added tests?
 
 ### Process
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,6 +6,7 @@ on:
       - master
       - production
       - dev-rheft
+      - ronheft/test-investigation
 
 defaults:
   run:
@@ -222,7 +223,7 @@ jobs:
     needs:
       - build_release
       - build_frontend_release
-    if: github.ref == 'refs/heads/master'
+    if: github.ref == 'refs/heads/ronheft/test-investigation'
     environment: test
     concurrency: test
     runs-on: ubuntu-latest
--- a/frontend-react/README.md
+++ b/frontend-react/README.md
@@ -1,0 +1,14 @@
+# React App Front End
+## Getting Started
+```bash
+cd ./frontend-react
+yarn install
+npm start
+```
+
+## Lint
+### Auto change
+```bash
+cd ./frontend-react
+npm run lint:write
+```
--- a/frontend-react/package.json
+++ b/frontend-react/package.json
@@ -8,7 +8,7 @@
         "@craco/craco": "^6.3.0",
         "@fortawesome/fontawesome-svg-core": "^1.2.36",
         "@fortawesome/free-solid-svg-icons": "^5.15.4",
-        "@fortawesome/react-fontawesome": "^0.1.15",
+        "@fortawesome/react-fontawesome": "^0.1.16",
         "@okta/okta-auth-js": "^5.2.3",
         "@okta/okta-react": "^6.1.0",
         "@okta/okta-signin-widget": "^5.9.0",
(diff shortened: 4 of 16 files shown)


prime-router/docs/api/lookup-tables-openapi.yaml
```diff
@@ -0,0 +1,162 @@
...
+              schema:
+                $ref: '#/components/schemas/ErrorMsg'
+        '400':
+          description: if there was an error fetching the table data
+          content:
+            text/plain:
+              schema:
+                $ref: '#/components/schemas/ErrorMsg'
+    post:
+      description: Activate a specific version of a table
```
jimduff-usds, line 106: How does deactivation of a table version occur?
Is there a rule that only one version of a table can be active at a time?
If so, does activating a version also deactivate all other versions?

carlosfelix2: yes, only one version at a time is allowed, and it is enforced via a database constraint.  In the code on the other PR you can see how I deactivate and activate within the same transaction.

RickHawesUSDS: What is the content for this post?

prime-router/docs/api/lookup-tables-openapi.yaml
```diff
@@ -0,0 +1,164 @@
...
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorMsg'
+        '400':
+          description: if there was an error fetching the table data
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorMsg'
+    post:
```
RickHawesUSDS, line 105: I suggest that document that post requires PRIME admin rights, while get is open.

clediggins-usds (commented):
Is the intent that every minor change to any setting will result in a new version of that settings table?  For example, a receiver's status changes from 'inactive' to 'testing' and then to 'active' - is that 3 different versions?  If not, we may want a way to make minor updates/spelling fixes at `PATCH /api/lookuptables/{tableName}/{tableVersion}`.

prime-router/docs/api/lookup-tables-openapi.yaml
RickHawesUSDS: This is really a REST-style comment. Reading REST posts should have a representation of the state change that you want to effect. In other words, you should have a content part that represents the `isActive` state.

prime-router/docs/api/lookup-tables-openapi.yaml
```diff
@@ -0,0 +1,174 @@
...
+              schema:
+                $ref: '#/components/schemas/TableVersionInfo'
+        '400':
+          description: if there was an error fetching the list of tables
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorMsg'
+
+  /api/lookuptables/{tableName}:
```
whytheplatypus, line 37: Given the behavior of `/api/lookuptables` and `/api/lookuptables/{tableName}/{tableVersion}` I'd expect this to return a list of versions for this specific table. Is the intent for this to return the `latest` version of the table or the `active` version of a table? Endpoints that are aliases for each might be needed.

carlosfelix2: We only want to return the active table data.  I can see how we can benefit from a better endpoint here.

carlosfelix2: I added /content.  Take a look again

PR #2595, opened 2021-10-12
Azure Scaling Improvements

ronaldheft-gov:
This PR improves our scaling within our Azure environment:

* Moves our App Service instances [from P2v2 to P3v2](https://azure.microsoft.com/en-us/pricing/details/app-service/linux/)
  * This is the server resources per instance
  * This increases CPU from `2` vCPU to `4` vCPU
  * This increases memory from `7` GB to `14` GB
* Enables autoscaling in all environments
  * This was previously only enabled in production
* Changes the default minimum App Service instances from `1` to `2`
  * This will ensure we always have a redundant App Service instance
* Changes our auto-scaling metrics to match our PagerDuty alerts
  * Instances will scale out by `1` when CPU reaches `70%` on average over the last `5` minutes
  * Instances will scale out by `1` when memory reaches `85%` on average over the last `5` minutes
* Tweeks the cooldown on a scale-out to wait `1` minute before an additional scaling operation can occur instead of `5` minutes
  * This will allow our scaling to be more responsive when receiving a large amount of load


RickHawesUSDS (approved):

jimduff-usds (approved):
Whole-heartedly approve.  My only request is let's please do a run of that "parallel" test before and after so we have throughput metrics on the improvement.

Issue #2613, opened 2021-10-14
[React client] "/login" vs "/login/"


TomNUSDS:
From the build:

----

✅  crawled 15 out of 18 (/login/)
✅  crawled 16 out of 18 (/login)

----

There are `<NavLink>`s to both  `"/login"` and `"/login/"`

Investigate and make it consistent. (I _think_ it should be `"/login"`)

PR #2615, opened 2021-10-14
Remove trailing `/` from paths

TomNUSDS:
This is causing the static html crawler crawl pages twice: once as a document with no trailing `/` and again as a directory with a tailing `/`

Fixes: https://github.com/CDCgov/prime-reportstream/issues/2613

Note: this drops the crawled pages from 18 down to 14.

I noticed this issue while debugging why the crawl was crashing. Figured I should just fix it instead of putting it off. 

@sean-usds @kevinhaube Any ideas on how we can make sure we don't create duplicate <Link> paths in the future? (e.g. `"/how-it-works/elr-checklist/"` vs `"/how-it-works/elr-checklist"`)

### DIFF
--- a/prime-router/docs/schema_documentation/strac-strac-covid-19.md
+++ b/prime-router/docs/schema_documentation/strac-strac-covid-19.md
@@ -743,6 +743,28 @@ The date which the specimen was collected. The default format is yyyyMMddHHmmssz
 
 ---
 
+**Name**: specimen_id
+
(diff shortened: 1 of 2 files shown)


kevinhaube:
@TomNUSDS Is it time we have a `PATH` enum? Much more consistency across the board calling something like `PATH.HOME` or `PATH.ELR_CHECKLIST`

PR #2633, opened 2021-10-15
Production Load Simulator, first pass

jimduff-usds:
This PR creates Simulator.kt, on top of the existing in-code integration test suite, to generate runs that simulate production.

The simulations are like those virtual "courses" you have to run when using a exercise bike or an exercise stepper.

Test Steps:
1. Run `./prime test  --run simulator --sequential --env staging --key <SECRET>`

### Testing
- [X] Tested locally?
- [X] Added tests?

## To Be Done
- The simulation code needs to be moved outside of being a child of CoolTest, so that it can be used as static functions by other classes.  However, this requires a lot of changes to TestReportStream.kt, and I kinda wanted to wait until @oslynn is done with his work, so as to not create painful merge conflicts.
- Once that's done, all the other load tests can be vastly simplified to simply be calls to the Simulator, with a variety of simulations.
- There is a small library of Simulations in the code; this could be greatly expanded and/or completely pulled out of the code into a config file that is read by the Simulator.


carlosfelix2 (changes requested):

prime-router/src/main/kotlin/cli/tests/Simulator.kt
```diff
@@ -0,0 +1,342 @@
...
+            bad("$result")
+        }
+        return result
+    }
+
+    /**
+     * A library of useful [Simulation]
+     */
+    companion object {
+        val primeThePump = Simulation(
```
carlosfelix2, line 155: These all should have kdoc describing what they are.

prime-router/src/main/kotlin/cli/tests/Simulator.kt
```diff
@@ -0,0 +1,342 @@
...
+            "IG",
+            "EVERY_5_MINS",
+            "ignore.EVERY_5_MINS",
+            0,
+            simpleRepSender,
+            true
+        )
+    }
+
+    fun setup(environment: ReportStreamEnv, options: CoolTestOptions) {
```
carlosfelix2, line 266: Add kdoc

prime-router/src/main/kotlin/cli/tests/Simulator.kt
```diff
@@ -0,0 +1,342 @@
...
+        ugly("Starting $name test.")
+        val result = runOneSimulation(primeThePump, environment, options)
+        if (!result.passed) {
+            bad("$name test FAILED to primeThePump prior to actual test.")
+            error("Can't continue to test")
+        }
+        echo("Ready for the real test:")
+    }
+
+    fun teardown(results: List<SimulatorResult>, entireTestMillis: Long): Boolean {
```
carlosfelix2, line 276: kdoc

carlosfelix2 (approved):
@jimduff-usds Made some more comments about kdoc, but approved so you do not have to wait for me.

Issue #2152, opened 2021-09-10
Create API Endpoint for Querying LIVD Values


MauriceReeves-usds:
## Background
We need to create an API endpoint that SimpleReport can access that will give them the ability to query and see if we have a test by the name they're trying to use. 

## Design
The API endpoint should return an array of JSON objects with the following format:
```json
{
    "manufacturer": "",
    "model": "",
    "vendorAnalyteName": "",
    "vendorSpecimenDescription": [
        ""
    ],
    "vendorResultDescription": [
        ""
    ],
    "testPerformedLoincCode": "",
    "testPerformedLoincLongName": "",
    "testOrderedLoincCode": "",
    "testOrderedLoincLongName": "",
    "vendorComment": "",
    "vendorAnalyteCode": "",
    "vendorReferenceId": "",
    "testKitNameId": "",
    "equipmentUid": "",
    "equipmentUidType": "",
    "component": "",
    "property": "",
    "time": "",
    "system": "",
    "scale": "",
    "method": "",
    "publicationVersionId": "",
    "loincVersionId": ""
}
```
These names match the headers of the `LOINC Mapping` tab in the LIVD Excel spreadsheet.

The endpoint should live at `/api/metadata/livd`.

If we want to filter, it would look like:
`/api/metadata/livd?model=BinaxNOW`

## Acceptance Criteria
* The API would accept GET requests, and would return a JSON array that reflects the data in the LIVD table if something matches against the field specified. The search should be case-sensitive and by default do a `contains`.
* If no filter is specified, the JSON array returns **ALL** the current LIVD objects with no filtering
* If the filter does not match any values, the API endpoint returns an empty array
* The endpoint should accept multiple filters: `/api/metadata/livd?model=BinaxNOW&testkitNameIdType=EUA`

PR #2646, opened 2021-10-18
Adds endpoint to query LIVD data in the app

MauriceReeves-usds:
## Overview
This PR adds an API endpoint that lets us query the current LIVD table for information about tests.

## Additional Details
I have created this endpoint so we can provide a way for senders to determine what types of tests we do support. We have run into multiple issues with senders sending us records with names we don't support, which means we end up sending results to PHA's without LOINC codes. This is one step towards supporting our different partners and giving them the tools to manage the data themselves.

## Changes
- Created a copy of the most recent LIVD table as `LIVD-SARS-CoV-2-Latest.csv` that should ALWAYS be the most recent copy of the LIVD table. **This is a process change that we should follow moving forward until we integrate the LIVD table into the DB**
- Updated the `LookupTable` class to implement `Iterable` so we get all the magic extensions methods that come with that
- Updated the `LookupTable` class to add a `DataRows` property that lets us get the rows for the table without a header
- Added the endpoint for querying

### Testing
- [x] Tested locally?
- [ ] Added tests?

### Process

## Fixes
- #2152


prime-router/src/main/kotlin/azure/MetaDataFunction.kt
```diff
@@ -0,0 +1,94 @@
...
+
+    /**
+     * gets the LIVD table filtered by the values passed in
+     */
+    private fun getLivdTable(filters: Map<String, String>): List<LivdData>? {
+        logger.info("Getting the LIVD table for the metadata api")
+        val livdTable = workflowEngine
+            .metadata
+            .findLookupTable("LIVD-SARS-CoV-2-latest") ?: return null
+        logger.info("Pulling rows")
```
cdigg, line 81: Do we want to do any validation / look for jagged CSV / verify the column count of at least the header row?  Ideally the file would always be properly formatted but since this is no longer directly tied to it there may be problems in the future if a file is badly formatted.

PR #2650, opened 2021-10-18
Create 2020-10-19 Release Notes

sean-usds:
Create 2020-10-19 Release Notes

### DIFF
--- a/prime-router/docs/releases/2021-10-19-release-notes.md
+++ b/prime-router/docs/releases/2021-10-19-release-notes.md
@@ -1,0 +1,22 @@
+#  ReportStream Release Notes
+
+*October 18, 2021*
+
+## General useful links:
+
+- All Schemas are documented here:  [Link to detailed schema dictionaries](../schema_documentation)
+- The Hub API is documented here: [Hub OpenApi Spec](../openapi.yml)
+- [Click here for all release notes](../releases)
+
+## For this release
+
+### More instructive error and warning messages in API response
+
+Previously, when there were problems with health data the error or warning message would simply state where the problem occurs. 
+These changes update the messages so also include instructions for correcting the problems. For example:
+
+Previous message:
+> Invalid phone number format in row 3.
+
+Updated message:
+> Invalid phone number format in row 3. Reformat to a 10-digit phone number (e.g. (555) - 555-5555).


prime-router/docs/releases/2021-10-19-release-notes.md
```diff
@@ -0,0 +1,22 @@
...
+### More instructive error and warning messages in API response
+
+Previously, when there were problems with health data the error or warning message would simply state where the problem occurs. 
+These changes update the messages so also include instructions for correcting the problems. For example:
+
+Previous message:
+> Invalid phone number format in row 3.
+
+Updated message:
+> Invalid phone number format in row 3. Reformat to a 10-digit phone number (e.g. (555) - 555-5555).
```
jimduff-usds, line 22: An issue that I have brought up with the Data Automation team is that we can't be encouraging repeated resubmissions, as most states are not set up to handle that. Better is to fix the format for _future_ data, rather than fix and resubmit current data.   I'm concerned that the language "Reformat to ..." encourages novice senders to keep uploading into production until they get it right.

Here's my original post on this topic, from a chat with Drew and Lori on August 26, 2021:

> Some more thinking regarding my comment in the demo-prep meeting, on the awesome new uploader tool:
> Even one repeat of a data upload is concerning - I suspect many states will treat it as new data or otherwise get confused.  If this occurs rarely that’s probably OK, but not if its common.
>
> An even bigger concern would be that a sender keeps uploading the same data over and over, while they debug it.   That would be unacceptable - this is a production system, in a crisis situation, and the data is going to states who are already overwhelmed.
>
> One way to mitigate this is to require uploading senders to go thru an onboarding process, using a version of the uploader that is harmless.    This would be super easy - the same UI, but behind the scenes add our magic ‘SkipSend’ argument that prevents the data from forwarding.  Once they have proved they can upload data that we decide is of acceptable quality, we let their data flow downstream.  That is, we control the ‘SkipSend’ switch, and, even after they are accepted, give them the option to utilize it as well, when they want to experiment.
>
> If a sender has future quality issues, we might turn them back to ‘SkipSend’ mode until they are able to address it.

anshulkumar-usds (approved):
Looks good to me.

prime-router/docs/releases/2021-10-19-release-notes.md
loripusey: We updated the header message for the Warnings so that it now says this for these scenarios to encourage people to NOT re-submit if the file was accepted.

- When the file was accepted by has warnings   
image
- When the file was rejected and has warnings
    - We removed the "Your file has been accepted with warnings"

Here is the ticket where we made this change:
https://app.zenhub.com/workspaces/prime-data-hub-5ff4833beb3e08001a4cacae/issues/cdcgov/prime-reportstream/2501

PR #2808, opened 2021-10-27
Jim/new senders

jimduff-usds:
This PR adds configs for several new senders.

Test Steps:
1. Generate fake data for the abbott schema and confirm it goes thru the pipeline to hhsprotect.

## Changes
- 

### Testing
- [X] Tested locally?
- [ ] Added tests?

## To Be Done
- Add yamlitos for staging for the new senders
- check whether we need to generate state or county patient and/or facility information based on zip codes.
- if we do this, we'll need to include the routeTo parameter in the URL, or as a sender configuration field.
- fix bug wherein fake data generate generates duplicate fields for direct schema.


prime-router/src/main/kotlin/azure/HttpUtilities.kt
```diff
@@ -17,7 +17,7 @@ import java.time.ZoneOffset
 import java.time.format.DateTimeFormatter
 
 enum class ReportStreamEnv(val urlPrefix: String) {
-    TEST("https://pdhtest-functionapp.azurewebsites.net"),
+    TEST("https://test.prime.cdc.gov"),
```
carlosfelix2, line 20: Update the Environments.kt file as well or use the Environments.kt info.  This enum will be replaced/removed soon with Environment.kt

carlosfelix2 (commented):
One comment, but have not tested it

jbiskie (approved):

Issue #2225, opened 2021-09-16
return qualityFilter per-receiver results back to the sender


jimduff-usds:
As a Sender I need to know if data I've submitted fails a Receiver's QualityFilter, so I can make adjustments.

This ticket is done when the response json back to the sender includes, for Receivers, information on why one or more rows failed to go that receiver.

These qualityFilter actions are currently being logged at the level of the entire-receiver report (see JurisdictionalFilters.kt::logFiltering and logAllEliminated.   So the hard part of this work, figuring out the info of what got filtered and why, has already been done for you.   In the examples below, the strings printed in the json response are pretty much the same as what's currently getting logged.

So this story is to provide that same data currently being logged, back to the sender in the json response.

#### Some requirements:
- only provide feedback if one or more rows was eliminated by the qualityFilter, exactly the same as how the current logging functions (listed above) work.
- never show qualityFilters info in the json response for any receiver with the string 'secondary' in it's name.
    - These are cases where the quality filter behavior has been reversed `reverseTheFilter`. If we were to show qualityFilter errors, it would look like all the "good" data was filtered out.
- currently receivers are only listed in the json response if they are receiving one or more rows.   However, with this change, we'll begin to report on receivers that got zero rows.    So in these cases you'll need to add receivers to the list, prior to generation of the json response object.

Here's an example.  let's say a sender sent a file with 17 rows all destined for California, but _all 17 were eliminated_ by the quality filter.  **Currently this is what will show up in their json response**:

```
  "destinations": [
    {
      "service": "elr",
      "itemCount": 17,
      "sending_at": "2021-09-16T13:25-04:00",
      "organization": "HHSProtect",
      "organization_id": "hhsprotect"
    },
  ],

```
**But after you make this change, this is what will appear:**

```
  "destinations": [
    {
      "service": "elr",
      "itemCount": 17,
      "sending_at": "2021-09-16T13:25-04:00",
      "organization": "HHSProtect",
      "organization_id": "hhsprotect"
    },
    {
      "service": "elr",
      "itemCount": 0,
      "organization": "California Department of Public Health",
      "organization_id": "ca-dph"
      "qualityFiltering":  [ "For ca-dph.elr, qualityFilter isValidCLIA(testing_lab_clia) reduced the Items from 17 to 0.  All rows were filtered" ]
    }
  ],

```

[...]

PR #2819, opened 2021-10-28
#2225 - Include qualityFilter information in response

whytheplatypus:
This PR ...
- [x] includes destinations filtered out based on quality in response.
- [x] lists the items that have been filtered for each destination in response.

**Test Steps:**
With SmokeTest:
1. run `./gradlew testSmoke`
2. Look at results for Quality Filter tests, they will have entries for `filteredReportRows`

By Hand:
1. Submit files that should have rows or whole entries filtered out.
2. See that response contains the correct `filteredReportRows`
3. See that the `secondary` does _not_ have `filteredReportRows`

[ReportStream HL7 Filtered Rows.postman_collection.json.zip](https://github.com/CDCgov/prime-reportstream/files/7443761/ReportStream.HL7.Filtered.Rows.postman_collection.json.zip)

**For Review:**
- Clarity
- Separation of concerns
- Unexpected consequences

**Fixes:**
#2225

**Example response:**
```json
{
    "id": "d9cf2c2e-b138-49fb-a0e6-7bd00a4c638c",
    "timestamp": "2021-10-29T17:06:08.857916Z",
    "topic": "covid-19",
    "reportItemCount": 3,
    "destinations": [
        {
            "organization": "California Department of Public Health",
            "organization_id": "ca-dph",
            "service": "elr",
            "filteredReportRows": [
                "For ca-dph.elr, qualityFilter hasValidDataFor, [message_id, equipment_model_name, specimen_type, test_result, patient_last_name, patient_first_name, patient_dob] filtered out Rows 1 reducing the Item count from 3 to 2."
            ],
            "sending_at": "2021-10-29T13:07-04:00",
            "itemCount": 2
        },
        {
            "organization": "California Department of Public Health",
            "organization_id": "ca-dph",
            "service": "elr-secondary",
            "sending_at": "2021-10-29T13:07-04:00",
            "itemCount": 1
        }
    ],
    "destinationCount": 2,
    "warningCount": 3,
    "errorCount": 0,
    "errors": [],
    "warnings": [
        {
            "scope": "ITEM",
            "id": "",
            "details": "Report 1: Timestamp for ORC-15 - order_test_date should provide more precision. Should be formatted as YYYYMMDDHHMM[SS[.S[S[S[S]+/-ZZZZ"
        },
        {
            "scope": "ITEM",
            "id": "",
            "details": "Report 2: Timestamp for ORC-15 - order_test_date should provide more precision. Should be formatted as YYYYMMDDHHMM[SS[.S[S[S[S]+/-ZZZZ"
        },
        {
            "scope": "ITEM",
            "id": "",
            "details": "Report 3: Timestamp for ORC-15 - order_test_date should provide more precision. Should be formatted as YYYYMMDDHHMM[SS[.S[S[S[S]+/-ZZZZ"
        }
    ],
    "consolidatedErrors": [],
    "consolidatedWarnings": [
        {
            "message": "Report 3: Timestamp for ORC-15 - order_test_date should provide more precision. Should be formatted as YYYYMMDDHHMM[SS[.S[S[S[S]+/-ZZZZ",
            "rows": ""
        },
        {
            "message": "Report 2: Timestamp for ORC-15 - order_test_date should provide more precision. Should be formatted as YYYYMMDDHHMM[SS[.S[S[S[S]+/-ZZZZ",
            "rows": ""
        },
        {
            "message": "Report 1: Timestamp for ORC-15 - order_test_date should provide more precision. Should be formatted as YYYYMMDDHHMM[SS[.S[S[S[S]+/-ZZZZ",
            "rows": ""
        }
    ]
}
```

```json
{
    "id": "c07a22fe-30ae-4366-8e88-470873313712",
    "timestamp": "2021-10-29T17:06:35.530668Z",
    "topic": "covid-19",
    "reportItemCount": 1,
    "destinations": [
        {
            "organization": "California Department of Public Health",
            "organization_id": "ca-dph",
            "service": "elr",
            "filteredReportRows": [
                "For ca-dph.elr, qualityFilter hasValidDataFor, [message_id, equipment_model_name, specimen_type, test_result, patient_last_name, patient_first_name, patient_dob] filtered out Rows 0 reducing the Item count from 1 to 0."
            ],
            "sending_at": "never - all items filtered out",
            "itemCount": 0
        },
        {
            "organization": "California Department of Public Health",
            "organization_id": "ca-dph",
            "service": "elr-secondary",
            "sending_at": "2021-10-29T13:07-04:00",
            "itemCount": 1
        }
    ],
    "destinationCount": 2,
    "warningCount": 1,
    "errorCount": 0,
    "errors": [],
    "warnings": [
        {
            "scope": "ITEM",
            "id": "",
            "details": "Report 1: Timestamp for ORC-15 - order_test_date should provide more precision. Should be formatted as YYYYMMDDHHMM[SS[.S[S[S[S]+/-ZZZZ"
        }
    ],
    "consolidatedErrors": [],
    "consolidatedWarnings": [
        {
            "message": "Report 1: Timestamp for ORC-15 - order_test_date should provide more precision. Should be formatted as YYYYMMDDHHMM[SS[.S[S[S[S]+/-ZZZZ",
            "rows": ""
        }
    ]
}
```


prime-router/src/main/kotlin/Translator.kt
```diff
@@ -65,10 +59,7 @@ class Translator(private val metadata: Metadata, private val settings: SettingsP
         }
     }
 
-    /**
-     * This does both the filtering by jurisdiction, by qualityFilter, and also the translation.
-     */
-    private fun translateByReceiver(input: Report, receiver: Receiver, defaultValues: DefaultValues): Report {
+    private fun filterByJurisdiction(input: Report, receiver: Receiver): Report {
```
carlosfelix2, line 62: Add kdoc

prime-router/src/main/kotlin/Report.kt
```diff
@@ -338,6 +359,19 @@ class Report : Logging {
...
+            if (doDetailedFilterLogging && filterFnSelection.size() < table.rowCount()) {
+                val before = Selection.withRange(0, table.rowCount())
+                filteredRows.add(
+                    FilterResult(
+                        receiver.fullName,
+                        filterFn.name,
+                        fnArgs,
+                        table.rowCount(),
+                        filterFnSelection.size(),
+                        before.andNot(filterFnSelection).toArray()
```
MauriceReeves-usds, line 371: If I'm reading this correctly, this will just show me which row numbers were filtered out, but no additional information about them. Does this always correctly map to the index that we can refer back to in the `ItemLineage` table. If I wanted to go back and figure out which rows were washed out, would it be easy to figure that out with this? Should we include the `messageId` field?

Thinking from our side of the work short-term, senders will want to know why records didn't go to a state, and I want to be able to find the details. Long-term I'd like to be able to put this into a dashboard so it's simple for support staff to quickly see and examine what did and did not go where.

If this ties cleanly and consistently to the `parentIndex` field in `ItemLineage`, then we should be good, but I almost want a table that just shows us all the filters that fired for a report.

jimduff-usds (changes requested):
I created fake data then deleted the patient_last_name by hand, then submitted it.
```
./prime data --input-fake 1 --input-schema primedatainput/pdi-covid-19 --output ./junk/fake.csv
<CHANGE THE FILE MANUALLY>
curl -X POST -H "client:simple_report"  -H "content-type:text/csv"  --data-binary "@./junk/fake.csv"  "http://localhost:7071/api/reports"
```
The error message I got back was missing the field name

```
 "errors" : [ {
    "scope" : "ITEM",
    "id" : "106062",
    "details" : "Blank value for element "
  } ],
```

prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -71,6 +72,10 @@ class ActionHistory {
      */
     val reportsOut = mutableMapOf<ReportId, ReportFile>()
 
+    val filteredOutReports = mutableMapOf<ReportId, ReportFile>()
```
carlosfelix2, line 75: add kdoc

prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -258,6 +263,21 @@ class ActionHistory {
             error("For report ${report.id}:  Externally submitted reports should never have item lineagee.")
     }
 
+    fun trackFilteredReport(
```
carlosfelix2, line 266: add kdoc

prime-router/src/test/kotlin/CsvFileTests.kt
```diff
@@ -84,13 +84,14 @@ class CsvFileTests {
...
     }
 
-    private fun translateReport(inputReport: Report, baseName: String) {
-        val outputReports = Translator(metadata, settings).translateByReceiver(inputReport)
-        assertThat(outputReports).hasSize(2)
-
+    private fun translateReport(inputReport: Report, baseName: String, expected: List<String>) {
+        val outputReports = Translator(metadata, settings).filterAndTranslateByReceiver(inputReport)
+        assertThat(outputReports).hasSize(expected.size)
+        // listOf("AZ-test-receiver-", "federal-test-receiver-")
```
carlosfelix2, line 90: Remove this dead line of code.

prime-router/src/main/kotlin/Report.kt
```diff
@@ -39,6 +39,21 @@ const val REPORT_MAX_ITEMS = 10000
 const val REPORT_MAX_ITEM_COLUMNS = 2000
 const val REPORT_MAX_ERRORS = 100
 
+data class FilterResult(
```
carlosfelix2, line 42: Add kdoc

prime-router/src/main/kotlin/Report.kt
```diff
@@ -276,6 +295,7 @@ class Report : Logging {
             bodyFormat ?: this.bodyFormat,
         )
         copy.itemLineages = createOneToOneItemLineages(this, copy)
+        copy.filteredItems.addAll(this.filteredItems)
```
carlosfelix2, line 298: Maybe add more to the function description as you are adding more functionality

prime-router/src/main/kotlin/Translator.kt
```diff
@@ -46,8 +38,10 @@ class Translator(private val metadata: Metadata, private val settings: SettingsP
                 (limitReceiversTo.isEmpty() || limitReceiversTo.contains(receiver.fullName))
         }.mapNotNull { receiver ->
             try {
-                val mappedReport = translateByReceiver(input, receiver, defaultValues)
-                if (mappedReport.itemCount == 0) return@mapNotNull null
+                val jurisFilteredReport = filterByJurisdiction(input, receiver)
+                if (jurisFilteredReport.isEmpty()) return@mapNotNull null
+                val mappedReport = translateByReceiver(jurisFilteredReport, receiver, defaultValues)
+                // if (mappedReport.isEmpty() && mappedReport.filteredItems.isEmpty()) return@mapNotNull null
```
carlosfelix2, line 44: Remove dead code

prime-router/src/main/kotlin/Translator.kt
```diff
@@ -97,23 +93,20 @@ class Translator(private val metadata: Metadata, private val settings: SettingsP
...
             logger.warn(
                 "Data quality problem in report ${input.id}, receiver ${receiver.fullName}: " +
-                    "There were ${jurisFilteredReport.itemCount} rows prior to qualityFilter, and " +
+                    "There were ${input.itemCount} rows prior to qualityFilter, and " +
                     "${qualityFilteredReport.itemCount} rows after qualityFilter."
             )
         }
 
-        // Always succeed in translating an empty report after filtering (even if the mapping process would fail)
-        if (qualityFilteredReport.isEmpty()) return buildEmptyReport(receiver, input)
```
carlosfelix2, line 115: Is this logic still needed?

prime-router/src/test/kotlin/CsvFileTests.kt
```diff
@@ -60,7 +60,7 @@ class CsvFileTests {
         }
```
MauriceReeves-usds, line 60: I only see a "happy path" test and a "no header" test in here. We should have a few tests that invoke filters and check the results as well. Am I missing them?

MauriceReeves-usds (changes requested):
@whytheplatypus I have three small requests, and then two questions. Otherwise, looks good.

prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -574,20 +619,34 @@ class ActionHistory {
...
+            }
+        }
+
         jsonGen.writeStringField(
             "sending_at",
             when {
                 reportOptions == ReportFunction.Options.SkipSend -> {
                     "never - skipSend specified"
                 }
+                countToPrint == 0 -> {
```
carlosfelix2, line 647: Is there a doc that will need to be changed due to these and other response changes?  I don't think I have seen one.

prime-router/src/main/kotlin/azure/ReportFunction.kt
```diff
@@ -460,6 +460,13 @@ class ReportFunction : Logging {
                     validatedRequest.routeTo,
                     validatedRequest.warnings,
                 )
+                .filter { (report, receiver) ->
+                    if (report.itemCount > 0) {
+                        return@filter true
+                    }
```
carlosfelix2, line 466: I would change it to an if/else, so you do not have to force the returns.

carlosfelix2 (changes requested):
Some minor things. I did not test at this time since there are other requested changes.

prime-router/src/main/kotlin/Report.kt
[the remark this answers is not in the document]
whytheplatypus: @MauriceReeves-usds I'll add KDoc as well but do you have any suggestions to clarify the code itself? Maybe being more specific and calling it a `QualityFilterResult`and possibly modifying some of the attributes to show what it's communicating?

prime-router/src/main/kotlin/Report.kt
whytheplatypus: It does match with the 0 based index of the `row` being filtered. So it'll say it filtered rows `1,2,5` etc. Moving to a struct rather than a string leaves room for the more machine readable response in follow up work.

I think adding the `messageId` should be doable from what I'm seeing. Is that both guaranteed to be filled out and enough of a value add to do now?

prime-router/src/test/kotlin/CsvFileTests.kt
whytheplatypus: There are some SmokeTests that I found that exercise the filters, however they also only flow through the `cli` so while they provide some confidence I think some follow up tickets for solidifying testing would be great.

prime-router/src/main/kotlin/Translator.kt
whytheplatypus: I was thinking no because it's equivalent (since it's `isEmpty`) but now I'm pretty sure it is.. will add the early return back.

prime-router/src/main/kotlin/azure/ReportFunction.kt
whytheplatypus: I prefer early returns as a style, the code is more delete-able, but agreed it's funky with the named return

prime-router/src/main/kotlin/azure/ActionHistory.kt
whytheplatypus: looks like at least the [openapi.yml](https://github.com/CDCgov/prime-reportstream/blob/master/prime-router/docs/openapi.yml#L537-L554) needs updating

prime-router/src/main/kotlin/azure/ReportFunction.kt
whytheplatypus: This made me reconsider the pattern completely, I think I'm happier with what came out it's very clear what it's doing and why.

whytheplatypus:
@jimduff-usds it looks like that field is required to not be blank by the scheme? I tried with dob and it worked as expected.

prime-router/src/main/kotlin/Report.kt
jimduff-usds: No, there's no guarantee that we get a messageId.   So it would be an 'include it if you have it' feature.

jimduff-usds:
> @jimduff-usds it looks like that field is required to not be blank by the scheme? I tried with dob and it worked as expected.

@whytheplatypus   To clarify, the problem is that the message "Blank value for element" doesn't tell you which field is the problem.

prime-router/src/main/kotlin/Report.kt
jimduff-usds: The 'consolidated warnings' is also printing out "row" numbers, eg

`  "consolidatedWarnings" : [ {
    "message" : "Invalid code: '3.981E+12' is not a display value in altValues set for 'healthcareEmployeeType' ('healthcare_employee_type').",
    "rows" : "Rows: 7"
  }

We need to make sure you and they are using the same row numbers!

whytheplatypus: There's a [bug here](https://github.com/CDCgov/prime-reportstream/blob/master/prime-router/src/main/kotlin/azure/ReportFunction.kt#L607-L608) where it assumes the type is CSV. How about normalizing it so that we report 1 offset rows. And document that the csv header is not included.

carlosfelix2 (approved):

prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -416,7 +446,9 @@ class ActionHistory {
             insertReportFile(it, txn)
         }
         reportsOut.values.forEach {
-            insertReportFile(it, txn)
+            if (it.itemCount > 0) {
```
clediggins-usds, line 449: What happened to a reportOut value with 0 itemCount before?
Does this change represent a change in functionality - we would potentially have fewer reports tracked.
How could something end up in reportOut.values that we don't actually want to track?

prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -458,7 +490,7 @@ class ActionHistory {
         var parentReports2 = mutableSetOf<ReportId>()
         parentReports2.addAll(reportsReceived.keys)
         parentReports2.addAll(reportsIn.keys)
-        val childReports2 = reportsOut.keys
+        val childReports2 = reportsOut.filterValues { it.itemCount > 0 }.keys
```
clediggins-usds, line 493: How do we end up with something in reportsOut that is not in fact a child report?

prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -574,20 +599,34 @@ class ActionHistory {
...
+        reportOptions: ReportFunction.Options,
+        reportId: ReportId? = null
     ) {
         jsonGen.writeStartObject()
         // jsonGen.writeStringField("id", reportFile.reportId.toString())   // TMI?
         jsonGen.writeStringField("organization", organization.description)
         jsonGen.writeStringField("organization_id", orgReceiver.organizationName)
         jsonGen.writeStringField("service", orgReceiver.name)
 
+        reportId?.let {
```
clediggins-usds, line 611: What is the use case in which a reportId is null?
Would that be if we filtered out EVERYTHING?  in that case, wouldn't we want to write this element?

clediggins-usds (commented):
A few comments added - let me know if you want to chat. I have only a tenuous understanding of what 'itemCount' actually means in most of this.

prime-router/docs/releases/2021-11-09-release-notes.md
```diff
@@ -0,0 +1,115 @@
...
+    "warningCount": 0,
+    "errorCount": 0,
+    "errors": [],
+    "warnings": [],
+    "consolidatedErrors": [],
+    "consolidatedWarnings": []
+}
+```
+
+Multiple Items filtered out
```
jimduff-usds, line 55:
```suggestion
Some but not all items filtered out:

In this example one of the items passed CA DPH's filter criteria, but two items failed.

A note on the item numbering:  ReportStream can accept data in CSV and HL7 batch formats.

For a CSV submission, an 'item' is just a row of data in the CSV, and item 1 is the first row of actual data _after_ the header row, item 2 is the second row of actual data, and so on.

For an HL7 batch submission an 'item' is a single HL7 message in the batch, and item 1 is the first message in the batch, item 2 is the second message, and so on.


```

jimduff-usds (approved):
Added text to the release notes.

prime-router/src/main/kotlin/azure/ActionHistory.kt
whytheplatypus: let me take another look at this, this change may not be needed anymore.

Before this reportOut would never contain a report with a 0 itemCount, they were filtered out before this. So they essentially vanished

MauriceReeves-usds (approved):

prime-router/src/main/kotlin/Translator.kt
```diff
@@ -46,8 +38,9 @@ class Translator(private val metadata: Metadata, private val settings: SettingsP
                 (limitReceiversTo.isEmpty() || limitReceiversTo.contains(receiver.fullName))
         }.mapNotNull { receiver ->
             try {
-                val mappedReport = translateByReceiver(input, receiver, defaultValues)
-                if (mappedReport.itemCount == 0) return@mapNotNull null
+                val jurisFilteredReport = filterByJurisdiction(input, receiver)
+                if (jurisFilteredReport.isEmpty()) return@mapNotNull null
```
whytheplatypus, line 42: These steps were split apart (filterByJursdiction, then filter by quality + translate) to allow a report that was empty due to quality filtering to be returned.. while keeping reports filtered based on jurisdiction out.

prime-router/src/main/kotlin/Translator.kt
```diff
@@ -97,22 +97,21 @@ class Translator(private val metadata: Metadata, private val settings: SettingsP
...
                 "Data quality problem in report ${input.id}, receiver ${receiver.fullName}: " +
-                    "There were ${jurisFilteredReport.itemCount} rows prior to qualityFilter, and " +
+                    "There were ${input.itemCount} rows prior to qualityFilter, and " +
                     "${qualityFilteredReport.itemCount} rows after qualityFilter."
             )
         }
 
-        // Always succeed in translating an empty report after filtering (even if the mapping process would fail)
-        if (qualityFilteredReport.isEmpty()) return buildEmptyReport(receiver, input)
+        if (qualityFilteredReport.isEmpty()) return qualityFilteredReport
```
whytheplatypus, line 114: need to return the actual report for records even if it's empty.

prime-router/src/main/kotlin/Translator.kt
```diff
@@ -139,13 +138,9 @@ class Translator(private val metadata: Metadata, private val settings: SettingsP
...
             transformed = transformed.deidentify()
-        return transformed.copy(destination = receiver, bodyFormat = receiver.format)
-    }
-
-    fun buildEmptyReport(receiver: Receiver, from: Report): Report {
-        val toSchema = metadata.findSchema(receiver.schemaName)
-            ?: error("${receiver.schemaName} schema is missing from catalog")
-        return Report(toSchema, emptyList(), listOf(ReportSource(from.id, "mapping")), metadata = metadata)
+        var copy = transformed.copy(destination = receiver, bodyFormat = receiver.format)
+        copy.filteredItems.addAll(qualityFilteredReport.filteredItems)
```
whytheplatypus, line 142: might not need this anymore

whytheplatypus: This is required because `transformed` / `toReport` is a result of a call of `applyMapping` on `qualityFilteredReport`. which doesn't and maybe shouldn't copy the filteredItems

PR #2836, opened 2021-10-29
Revert the Palo Alto in production

ronaldheft-usds:


ronaldheft-usds:
This has run already last night, but we should get the state of `master` to match the current deployed state.

cwinters-usds (approved):
Can't get much smaller

PR #2838, opened 2021-10-31
Remove dependency on `react-snap`

### LINKED ISSUE #2514: Remove `react-snap` from package's postbuild or replace with Gatsby
Is it safe to remove "postbuild": "react-snap" from the package.json's scripts?

This is the action that uses `puppeteer` to crawls the site to generate static pages, but it sounds like we're now using purely react SPA routes?

These are the two lines to be removed:
```json
    "scripts": {
        ...
        "postbuild": "react-snap"
    },
    ... 
        "devDependencies": {
        ...
        "react-snap": "^1.23.0",
```

But we really need to understand how the current server is handling the various urls for a SPA (e.g. a user bookmarks and goes to `http://localhost:3000/how-it-works/getting-started`).

If we really need a static html site (for the public facing side), we probably want to switch to something like Gatsby.

TomNUSDS:
react-snap is used to generate static html as a `postbuild` step using Puppeteer to crawl the site.
Ran `yarn remove react-snap` to update yarn.lock

Fixes #2514

### DIFF
--- a/frontend-react/package.json
+++ b/frontend-react/package.json
@@ -61,8 +61,7 @@
         "lint:prettier": "prettier -c 'src/**/*.{js,ts,jsx,tsx}'",
         "lint:prettier:write": "prettier --write 'src/**/*.{js,ts,jsx,tsx}'",
         "cypress-open": "./node_modules/.bin/cypress open",
-        "cypress-run": "./node_modules/.bin/cypress run",
-        "postbuild": "react-snap"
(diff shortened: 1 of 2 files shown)


kevinhaube (approved):
lgtm!

TomNUSDS:
@sean-usds I'd like to get your eyes on this real quick before a merge... just for awareness.

PR #2858, opened 2021-11-01
Turn Palo Alto on for production

cwinters-usds:
This PR flips the switch to enable the Palo Alto in production. (A retry from https://github.com/CDCgov/prime-reportstream/pull/2828)

## Deployment Checklist

- [ ] Ensure changes to CDC DNS are applied (Richard Teasley should notify us)
- [ ] Notify senders of downtime - Anshul
- [ ] Disable the release GitHub Action - Chris
- [ ] Merge this PR - Chris
- [ ] Deploy Terraform to production - Chris
    - [ ] Deploy `tf-01-network` to production
    - [ ] Deploy `tf-02-config` to production
    - [ ] Deploy `tf-03-persistent` to production
    - [ ] Deploy `tf-04-app` to production
- [ ] Verify we have App Insights logs - Jim
- [ ] Verify we have no blocked SFTP connection to senders - Jim
- [ ] Inspect Palo Alto logs to see if any unexpected traffic is blocked - Timothy
- [ ] Perform post-deployment validations - Jim
- [ ] Enable the release GitHub Action - Chris
- [ ] Notify receivers they can remove our old IP address from their firewall - Ming/Adrian

### DIFF
--- a/operations/app/src/environments/configurations/prod.tfvars
+++ b/operations/app/src/environments/configurations/prod.tfvars
@@ -11,4 +11,4 @@ https_cert_names = [
 okta_redirect_url         = "https://prime.cdc.gov/download"
 aad_object_keyvault_admin = "5c6a951e-a4c2-4890-b62c-0ed8179501bb" # CT-PRIMEDATAHUBPRD-DMZ-AZ-Contributor
 
-use_cdc_managed_vnet = false
+use_cdc_managed_vnet = true
--- a/operations/app/src/modules/function_app/main.tf
+++ b/operations/app/src/modules/function_app/main.tf
@@ -22,8 +22,10 @@ locals {
     # Route storage account access through the VNET
     "WEBSITE_CONTENTOVERVNET" = 1
 
-    # Use the VNET DNS server (so we receive private endpoint URLs)
-    "WEBSITE_DNS_SERVER" = "168.63.129.16"
+    # Use the CDC DNS for everything; they have mappings for all our internal
+    # resources, so if we add a new resource we'll have to contact them (see
+    # prime-router/docs/dns.md)
+    "WEBSITE_DNS_SERVER" = "172.17.0.135"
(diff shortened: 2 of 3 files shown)


ronaldheft-usds:
@cwinters-usds Thanks for updating the checklist! These steps are no longer needed:

> - [ ] Update the dnsmasq mapping file for production IP addresses - Chris
>   - [ ] Submit a PR with the changes
>   - [ ] Apply tf-01-network to force the dnsmasq server to update

cwinters-usds:
@ronaldheft-usds thanks much for the pointer, fixed

ronaldheft-usds (changes requested):
I think we also need a step to change the DNS for the function app to `172.17.0.135`. It's the `WEBSITE_DNS_SERVER` environment variable in the Terraform. That can be set ahead of time for this PR.

cwinters-usds:
@ronaldheft-usds good point -- I forgot to add that after bringing it up in our standup!

Is there any reason not to make this change [in the terraform here](https://github.com/CDCgov/prime-reportstream/blob/master/operations/app/src/modules/function_app/main.tf#L26)? It'd get applied in the 'Deploy Terraform to Production' step and the function app would be restarted as a result.

ronaldheft-usds:
> @ronaldheft-usds good point -- I forgot to add that after bringing it up in our standup!
> 
> Is there any reason not to make this change [in the terraform here](https://github.com/CDCgov/prime-reportstream/blob/master/operations/app/src/modules/function_app/main.tf#L26)? It'd get applied in the 'Deploy Terraform to Production' step and the function app would be restarted as a result.

Nope! That's the perfect spot! I would add it to this PR, so when we merge, that change will be included.

cwinters-usds:
@ronaldheft-usds added that config item, also a short doc about DNS -- hopefully it makes sense.

PR #2860, opened 2021-11-01
Updates remote check timing to something less aggressive than every five minutes

MauriceReeves-usds:
This PR updates the timing in the local settings file to once a day instead of every five minutes.


whytheplatypus (approved):
lgtm. We could also remove the verbose logging in [testTransport](https://github.com/CDCgov/prime-reportstream/blob/master/prime-router/src/main/kotlin/azure/CheckFunction.kt#L230-L231)  if it's not useful.

carlosfelix2 (approved):

Issue #2750, opened 2021-10-25
Move the new lean&hungry receive to its own Azure Function, to better support Maintenance Windows


jimduff-usds:
A major use case in @MauriceReeves-usds 's original proposal #2730 was to support Maintenance Windows.

That is, you could leave the lean&hungry receive running while shutting down or otherwise doing maintenance on the rest of the pipeline.

I believe our current design for lean&hungry is that it would at least need the Postgres database - that seems tricky or bad to change, so let's leave it for the moment.

However, we could move lean&hungry to a different azure function to make it independent; maybe have it deploy as part of a different GitHub action.  This would be a step towards Mo's original idea.

clediggins-usds:
This needs more evaluation - it may not need to be done at all and not gain us anything.

Issue #2628, opened 2021-10-15
asynchronous report receiving / processing


whytheplatypus:
**Problem**
Accepting a `report` from a `sender` and processing that `report` happen in one request/response flow. Because the intensive processing must happen immediately and in the same process as the request/response flow it limits the systems ability to handle a large number of concurrent requests.

**Proposed Solution**
The "Lean&Hungry" Reports endpoint / pipeline. This will accept a submission from a sender, ensuring they are authenticated and the submission is _reasonable_ (this is any fast validation that makes sense to do within the request/response flow). The endpoint will then queue the submission along with all necessary information for processing (A report ID for example). Another new AzureFunction will pick work up off that queue and process the submission, adding errors to the ActionHistory, the rest of the pipeline will then proceed as normal.

This allows many requests to be handled at once, at the sacrifice of a rich and informative result returned immediately.

**Alternatives**
- Increased horizontal scaling of infrastructure (this is going to be done as well.)

**Additional context**
To provide utility to a sender the `submission` API also needs to exist to provide transparency into the status of a submission.

PR #2869, opened 2021-11-01
Async Endpoint and message processing

clediggins-usds:
Test Steps:
1. send a message just like you normally would, including the header 'processing=async'
2. run any prime cli test as usual with the option --async
3. run end2end tests. It ignore the --async flag and instead forces both sync and async test process

## Changes
- Separates out /reports and /waters, using common pathing within ReportFunction
- Adds an option to process messages asynchronously
- Adds a new queue function, to handle 'Process' queue messages
- adds tests for asynch functionality
- adds prime CLI option --async to include the async flag in messages
- adds test to check a covidResultMeatadata row was added

### Testing
- [X ] Tested locally?
- [ X] Ran `./prime test` or `./gradlew testSmoke` against local Docker ReportStream container?
- [ X] Added tests?

### Process
- [- ] Updated the release notes?
- [N] Database changes are submitted as a separate PR?
- [-] DevOps team has been notified if PR requires ops support?

## Fixes
- #2628

## To Be Done
- #issue

    - No
    - Abstracts logic out of /reports and /waters, leaves auth as it is
    - No
     - No
     - No
     - No

Followup / Cleanup / Out of Scope ticket creation:
#2982 
#2750


prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -713,4 +732,157 @@ class ActionHistory {
             }
         }
     }
+
+    fun createResponseBody(
```
carlosfelix2, line 736: You need Kdoc for all your classes, functions and attributes.

carlosfelix2 (changes requested):

prime-router/src/main/kotlin/azure/ActionHistory.kt
clediggins-usds: Can do - this was just moving this function from ReportFunction.kt to ActionHistory (did not have kdoc in ReportFunction).  I'll review again and add missing kdocs.

prime-router/src/main/kotlin/azure/Event.kt
```diff
@@ -75,23 +82,90 @@ abstract class Event(val eventAction: EventAction, val at: OffsetDateTime?) {
...
+    val reportId: UUID,
+    val options: Options,
+    val defaults: Map<String, String>,
+    val routeTo: List<String>,
+    at: OffsetDateTime? = null,
+    val retryToken: RetryToken? = null
+) : Event(eventAction, at) {
+    override fun toQueueMessage(): String {
+        // turn the defaults and route to into strings that can go on the process queue
+        val routeToQueueParam = routeTo.let { it.joinToString(",") }
```
MauriceReeves-usds, line 143: If all you're doing is joining the `routeTo` to string, you don't need to do the `let`.

prime-router/src/main/kotlin/azure/Event.kt
```diff
@@ -75,23 +82,90 @@ abstract class Event(val eventAction: EventAction, val at: OffsetDateTime?) {
...
+        val defaultsQueueParam = defaults.let { it.map { pair -> "${pair.key}:${pair.value}" }.joinToString(",") }
```
MauriceReeves-usds, line 144: Ditto here about the let. Because defaults is not null, you don't need to wrap in a let.

MauriceReeves-usds (dismissed):

prime-router/src/main/kotlin/Report.kt
```diff
@@ -72,7 +87,11 @@ class Report : Logging {
     /**
      * the UUID for the report
      */
-    val id: ReportId
+    // TODO: Made this var instead of val so we can update the report ID after creation in the async process
+    //  functionality. There is a way to do it as a passed in variable, but the way we create reports via parsing
+    //  contentBody does not lend itself to that way of doing it. Once we are on FHIR internal spec, this could be
+    //  changed back to val - CD
+    var id: ReportId
```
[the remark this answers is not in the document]
clediggins-usds: This was a way to minimize changes to existing structure for the async endpoint - working on changing this to save the generated internal CSV and load that back instead of regenerating the Report from the original received file.

prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -713,4 +732,157 @@ class ActionHistory {
...
+                        isListing = true
+                    } else if (isListing) {
+                        sb.append(" to " + rows[i - 1].toString() + ", " + row.toString())
+                        isListing = false
+                    } else {
+                        sb.append(", " + row.toString())
+                        isListing = false
+                    }
+                    if (i == rows.lastIndex && isListing) {
+                        sb.append(" to " + rows[rows.lastIndex].toString())
```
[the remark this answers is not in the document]
clediggins-usds: Updated

prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -713,4 +732,157 @@ class ActionHistory {
...
+                }
+                it.writeEndArray()
+            }
+            writeDetailsArray("errors", errors)
+            writeDetailsArray("warnings", warnings)
+
+            fun createRowsDescription(rows: MutableList<Int>?): String {
+                // Consolidate row ranges, e.g. 1,2,3,5,7,8,9 -> 1-3,5,7-9
+                if (rows == null || rows.isEmpty()) return ""
+                rows.sort() // should already be sorted, just in case
```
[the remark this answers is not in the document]
clediggins-usds: Updated

prime-router/src/main/kotlin/azure/Event.kt
clediggins-usds: Updated

prime-router/src/main/kotlin/azure/Event.kt
clediggins-usds: Updated

prime-router/src/main/kotlin/azure/ActionHistory.kt
clediggins-usds: Updated

prime-router/src/main/kotlin/Report.kt
```diff
@@ -39,6 +40,20 @@ const val REPORT_MAX_ITEMS = 10000
 const val REPORT_MAX_ITEM_COLUMNS = 2000
 const val REPORT_MAX_ERRORS = 100
 
+// constants used for parsing and processing a report message
+const val ROUTE_TO_SEPARATOR = ","
+const val DEFAULT_SEPARATOR = ":"
+
+// options are used to process and route the report
+enum class Options {
```
carlosfelix2, line 48: Add kdoc maybe even explain where options come from or how they are used

prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -42,6 +46,12 @@ import java.time.OffsetDateTime
  *
  */
 class ActionHistory {
+    data class ItemRouting(
```
carlosfelix2, line 49: Add kdoc

prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -203,6 +213,14 @@ class ActionHistory {
             action.actionResponse = JSONB.valueOf(verboseResponse)
     }
 
+    /**
+     * Set the action response. This is primarily used by backend functions that do not have an associated HttpRequest
+     * @param jsonResponse the generated stringified json representing the response
+     */
+    fun trackActionResponse(jsonResponse: String) {
+        action.actionResponse = JSONB.valueOf(jsonResponse)
```
carlosfelix2, line 221: This code is a duplicate of line 213.  Can you go back and call this function where this code is used?

prime-router/src/main/kotlin/Report.kt
carlosfelix2: It almost feels this enum should not be in this file as it is not used or contained here.  Maybe a more suitable file?

prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -713,4 +732,166 @@ class ActionHistory {
...
+
+    /**
+     * Creates a string that will be used to populate the action_response column of an action
+     * after that action has been completed
+     * @param options Message options passed in
+     * @param warnings Store of warnings generated during the pipeline
+     * @param errors Store of errors generated during the pipeline
+     * @param verbose If true, all item routing details will be included in the response
+     * @param actionHistory The ongoing action history instance being used
+     * @param report The report this response body is for
```
carlosfelix2, line 744: Add @return to the kdoc

prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -713,4 +732,166 @@ class ActionHistory {
...
+            it.useDefaultPrettyPrinter()
+            it.writeStartObject()
+            if (report != null) {
+                it.writeStringField("id", report.id.toString())
+                it.writeStringField("timestamp", DateTimeFormatter.ISO_INSTANT.format(Instant.now()))
+                it.writeStringField("topic", report.schema.topic)
+                it.writeNumberField("reportItemCount", report.itemCount)
+                // TODO: need to get the correct path and set the ENDPOINT_BASE correctly, blocked by waiting for
+                //  the historyApi to be ready.
+                // it.writeStringField("location", "[base path]/${report.id}")
```
carlosfelix2, line 766: Is this dead code or part of the comment above?  If dead code then remove

prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -713,4 +732,166 @@ class ActionHistory {
...
     }
+
+    /**
+     * Creates a string that will be used to populate the action_response column of an action
+     * after that action has been completed
+     * @param options Message options passed in
+     * @param warnings Store of warnings generated during the pipeline
+     * @param errors Store of errors generated during the pipeline
+     * @param verbose If true, all item routing details will be included in the response
+     * @param actionHistory The ongoing action history instance being used
```
carlosfelix2, line 743: What happens when actionHistory is null?

prime-router/src/main/kotlin/azure/ActionHistory.kt
carlosfelix2: What happens when report is null?

prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -713,4 +732,166 @@ class ActionHistory {
...
+                    it.writeEndArray()
+                    it.writeEndObject()
+                }
+                it.writeEndArray()
+            }
+
+            it.writeNumberField("warningCount", warnings.size)
+            it.writeNumberField("errorCount", errors.size)
+
+            fun writeDetailsArray(field: String, array: List<ResultDetail>) {
```
carlosfelix2, line 789: kdoc

prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -713,4 +732,166 @@ class ActionHistory {
...
+                    it.writeStringField("id", error.id)
+                    it.writeStringField("details", error.responseMessage.detailMsg())
+                    it.writeEndObject()
+                }
+                it.writeEndArray()
+            }
+            writeDetailsArray("errors", errors)
+            writeDetailsArray("warnings", warnings)
+
+            fun createRowsDescription(rows: List<Int>?): String {
```
carlosfelix2, line 803: kdoc

prime-router/src/main/kotlin/azure/ActionHistory.kt
carlosfelix2: I will suggest that you move all this functions that are inside this other function and write unit tests for them.

prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -713,4 +732,166 @@ class ActionHistory {
...
+                it.writeEndArray()
+            }
+            writeConsolidatedArray("consolidatedErrors", errors)
+            writeConsolidatedArray("consolidatedWarnings", warnings)
+        }
+        return outStream.toString()
+    }
+
+    /**
+     * Creates a list of [ItemRouting] instances with the report index
```
carlosfelix2, line 857: Do not use the [] in the [ItemRouting] as that is not an argument to this function.

prime-router/src/main/kotlin/azure/Authentication.kt
```diff
@@ -17,12 +18,26 @@ enum class PrincipalLevel {
     USER
 }
 
+data class AuthenticationResult(
```
carlosfelix2, line 21: kdoc

prime-router/src/main/kotlin/azure/Authentication.kt
```diff
@@ -17,12 +18,26 @@ enum class PrincipalLevel {
...
+    fun checkAccess(request: HttpRequestMessage<String?>, sender: String): AuthenticationResult
+}
+
 data class AuthenticatedClaims(
     val userName: String,
     val principalLevel: PrincipalLevel,
     val organizationName: String?
 )
 
+data class AuthenticatedClaimsResult(
```
carlosfelix2, line 36: kdoc

prime-router/src/main/kotlin/azure/Event.kt
```diff
@@ -75,23 +82,90 @@ abstract class Event(val eventAction: EventAction, val at: OffsetDateTime?) {
     companion object {
         fun parseQueueMessage(event: String): Event {
             val parts = event.split(messageDelimiter)
-            if (parts.size < 3 || parts.size > 4) error("Internal Error: bad event format")
+            if (parts.size < 3 || parts.size > 7) error("Internal Error: bad event format")
```
carlosfelix2, line 85: Can you explain why the 7?  Maybe even add a comment here.  From the code just below, maybe this error can be incorporated into the when based on type?  I know this would be an improvement on old code.  Maybe consider it.

prime-router/src/main/kotlin/azure/Event.kt
```diff
@@ -75,23 +82,90 @@ abstract class Event(val eventAction: EventAction, val at: OffsetDateTime?) {
...
+                    val after = parts.getOrNull(3)?.let { OffsetDateTime.parse(it) }
                     val reportId = UUID.fromString(parts[2])
                     ReportEvent(action, reportId, after)
                 }
                 ReceiverEvent.eventType -> {
+                    val after = parts.getOrNull(3)?.let { OffsetDateTime.parse(it) }
                     ReceiverEvent(action, parts[2], after)
                 }
+                ProcessEvent.eventType -> {
+                    val after = parts.getOrNull(6)?.let { OffsetDateTime.parse(it) }
```
carlosfelix2, line 98: Should you check for exceptions on the parse here?  This code also is a duplicate as above.  maybe factor this out of the when.

prime-router/src/main/kotlin/azure/Event.kt
```diff
@@ -75,23 +82,90 @@ abstract class Event(val eventAction: EventAction, val at: OffsetDateTime?) {
...
 }
 
+/**
+ * An event that goes on and is read off of the 'process' queue.
+ * @param reportId The report ID to be processed
+ * @param options Options passed in on the initial call
+ * @param defaults Defaults passed in on the initial call
+ * @param routeTo Route to overrides passed in on the initial call
+ * @param at Allows future scheduling of the queue message
+ * @param retryToken For retrying a process queue message
```
carlosfelix2, line 130: Add @return

prime-router/src/main/kotlin/azure/HttpUtilities.kt
```diff
@@ -259,16 +262,23 @@ class HttpUtilities {
             if (key == null && environment == ReportStreamEnv.TEST) error("key is required for Test environment")
             if (key != null)
                 headers.add("x-functions-key" to key)
-            val url = environment.urlPrefix + oldApi + if (option != null) "?option=$option" else ""
-            return postHttp(url, bytes, headers)
+
+            // if we have this value and it is true, add the 'processing=async' query param
```
carlosfelix2, line 266: What is "this value"?

prime-router/src/main/kotlin/azure/HttpUtilities.kt
```diff
@@ -311,4 +321,17 @@ class HttpUtilities {
             }
         }
     }
+
+    class UrlBuilder(var baseUrl: String) {
```
carlosfelix2, line 325: Why not use Java's URL, URI, URLBuilder, URIBuilder, etc.?

prime-router/src/main/kotlin/azure/ProcessFunction.kt
```diff
@@ -0,0 +1,40 @@
+package gov.cdc.prime.router.azure
+
+import com.microsoft.azure.functions.ExecutionContext
+import com.microsoft.azure.functions.annotation.FunctionName
+import com.microsoft.azure.functions.annotation.QueueTrigger
+import com.microsoft.azure.functions.annotation.StorageAccount
+import java.util.logging.Level
+
+const val process = "process"
```
carlosfelix2, line 9: I don't like this too much.  This const has two meanings: the name of the function and the name of the queue.  I would say that if you have to use constants then make two of them.

carlosfelix2 (changes requested):

prime-router/src/main/kotlin/azure/ActionHistory.kt
clediggins-usds: Part of the comment above - once we know the endpoint for a file via the history API it will need to be uncommented and populated correctly.

prime-router/src/main/kotlin/azure/ActionHistory.kt
clediggins-usds: This is an overload of the function call on 213 - we don't have a HttpResponseMessage in this case.

prime-router/src/main/kotlin/azure/ActionHistory.kt
clediggins-usds: return added
When report is null it doesn't include the report-specific data (id, timestamp, topic, reportItemCount)

prime-router/src/main/kotlin/azure/Event.kt
clediggins-usds: For processEvent there are 6 parts to the message, not 3.
There should be no way for exceptions since the message is serialized by ProcessEvent.toQueueMessage

prime-router/src/main/kotlin/azure/Event.kt
clediggins-usds: Updated

prime-router/src/main/kotlin/azure/ProcessFunction.kt
clediggins-usds: Lets discuss - could easily change here, right now it follows the pattern used in both BatchFunction and SendFunction

prime-router/src/main/kotlin/azure/HttpUtilities.kt
clediggins-usds: Because I forgot about them and my IDE was not cooperating. Updated.

prime-router/src/main/kotlin/azure/ReportFunction.kt
```diff
@@ -103,103 +109,140 @@ class ReportFunction : Logging {
...
+            )
+            val response = HttpUtilities.httpResponse(
+                request,
+                responseBody,
+                HttpStatus.BAD_REQUEST
+            )
+            actionHistory.trackActionResult(response)
+            actionHistory.trackActionResponse(response, responseBody)
+            workflowEngine.recordAction(actionHistory)
+            return response
```
carlosfelix2, line 194: Try to limit the number of returns you do in a function.  Aim for just one if you can, but many tends to decrease readability.  Consider using an else here.

carlosfelix2 (changes requested):

prime-router/src/main/resources/db/migration/V24__add_processed_at.sql
```diff
@@ -0,0 +1,16 @@
+/*
```
carlosfelix2, line 1: Why two files?  Combine into one.

prime-router/src/main/kotlin/tokens/OktaAuthentication.kt
```diff
@@ -48,40 +51,53 @@ class OktaAuthentication(private val minimumLevel: PrincipalLevel = PrincipalLev
...
+        if (claims != null) {
             return block(claims)
-        } catch (ex: Exception) {
-            if (ex.message != null)
-                logger.error(ex.message!!, ex)
-            else
-                logger.error(ex)
-            return HttpUtilities.internalErrorResponse(request)
         }
+        return HttpUtilities.unauthorizedResponse(request, status ?: "")
```
carlosfelix2, line 100: The string being passed back needs to be json.  use errormsg from Httputils.  Also, is there something better than empty string if there is no status?

carlosfelix2:
You have quite a bit of new functions that you need to be writing unit tests against.

prime-router/src/main/kotlin/azure/WorkflowEngine.kt
```diff
@@ -288,6 +304,161 @@ class WorkflowEngine(
...
+            val reportFile = db.fetchReportFile(id, org = null, txn)
+            val header = createHeader(task, reportFile, null, null, null)
+            val currentAction = Event.EventAction.parseQueueMessage(task.nextAction.literal)
+
+            // get sender record
+            val sender = settings.findSender(reportFile.sendingOrg + "." + reportFile.sendingOrgClient)
+
+            //  create report
+            val report = createReport(
+                sender!!,
```
carlosfelix2, line 412: Lots of !! here.  Shouldn't you be checking this variables before using this?

prime-router/src/main/kotlin/azure/WorkflowEngine.kt
```diff
@@ -288,6 +304,161 @@ class WorkflowEngine(
...
+
+    fun handleProcessEvent(
+        messageEvent: ProcessEvent,
+        context: ExecutionContext,
+        actionHistory: ActionHistory
+    ) {
+        val errors: MutableList<ResultDetail> = mutableListOf()
+        val warnings: MutableList<ResultDetail> = mutableListOf()
+
+        db.transact { txn ->
```
carlosfelix2, line 400: Change this transaction so it only covers the first two fetches, then use another for the create, that way you are not having all this other non-database code inside the transaction block.  Transactions are only really useful for writes, so you do not gain anything from having all three calls in one transaction.

carlosfelix2 (changes requested):

whytheplatypus (changes requested):

prime-router/src/main/kotlin/azure/ActionHistory.kt
clediggins-usds: Updated

prime-router/src/main/kotlin/Report.kt
clediggins-usds: Added comment explaining why it is here, and referencing #2926

prime-router/src/main/kotlin/azure/ActionHistory.kt
clediggins-usds: Good call. Updated.

prime-router/src/main/kotlin/azure/HttpUtilities.kt
clediggins-usds: Updated

prime-router/src/main/kotlin/azure/ProcessFunction.kt
clediggins-usds: Updated to split into two variable. This no longer matches our BatchFunction or SendFunction structure but is more clear.

prime-router/src/main/kotlin/azure/ReportFunction.kt
```diff
@@ -12,64 +11,51 @@ import com.microsoft.azure.functions.annotation.FunctionName
...
 class ReportFunction : Logging {
-    enum class Options {
-        None,
-        ValidatePayload,
-        CheckConnections,
-        SkipSend,
-        SkipInvalidItems,
-        SendImmediately,
+
+    enum class ProcessingType {
```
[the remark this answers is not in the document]
clediggins-usds: The const is the name of the query string parameter. Updated to PROCESSING_TYPE_PARAMETER

prime-router/src/main/kotlin/azure/ReportFunction.kt
```diff
@@ -12,64 +11,51 @@ import com.microsoft.azure.functions.annotation.FunctionName
...
 
 private const val CLIENT_PARAMETER = "client"
 private const val OPTION_PARAMETER = "option"
 private const val DEFAULT_PARAMETER = "default"
-private const val DEFAULT_SEPARATOR = ":"
 private const val ROUTE_TO_PARAMETER = "routeTo"
-private const val ROUTE_TO_SEPARATOR = ","
 private const val VERBOSE_PARAMETER = "verbose"
 private const val VERBOSE_TRUE = "true"
+private const val PROCESSING_TYPE = "processing"
```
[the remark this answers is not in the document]
clediggins-usds: Updated

prime-router/src/main/kotlin/azure/WorkflowEngine.kt
```diff
@@ -543,4 +715,65 @@ class WorkflowEngine(
...
+                        schemaName = sender.schemaName,
+                        input = ByteArrayInputStream(content.toByteArray()),
+                        sources = listOf(ClientSource(organization = sender.organizationName, client = sender.name)),
+                        defaultValues = defaults
+                    )
+                    errors += readResult.errors
+                    warnings += readResult.warnings
+                    readResult.report
+                } catch (e: Exception) {
+                    errors.add(
```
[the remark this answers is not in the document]
clediggins-usds: This changes the way that we are currently handling errors while creating a report and would need additional testing. I'd prefer to put this change in a future 'code cleanup' case.

prime-router/src/main/kotlin/azure/WorkflowEngine.kt
```diff
@@ -543,4 +715,65 @@ class WorkflowEngine(
...
+                    val readResult = this.hl7Serializer.readExternal(
+                        schemaName = sender.schemaName,
+                        input = ByteArrayInputStream(content.toByteArray()),
+                        ClientSource(organization = sender.organizationName, client = sender.name)
+                    )
+                    errors += readResult.errors
+                    warnings += readResult.warnings
+                    readResult.report
+                } catch (e: Exception) {
+                    errors.add(
```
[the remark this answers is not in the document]
clediggins-usds: This changes the way that we are currently handling errors while creating a report and would need additional testing. I'd prefer to put this change in a future 'code cleanup' case.

prime-router/src/main/kotlin/azure/ReportFunction.kt
clediggins-usds: In this case, 'else' makes sense. Updated.

prime-router/src/main/resources/db/migration/V24__add_processed_at.sql
clediggins-usds: Flyway doesn't like it. Tried numerous ways, still no go. Talked with Ron, he agreed flyway doesn't like it.

prime-router/src/main/kotlin/Report.kt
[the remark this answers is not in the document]
clediggins-usds: Updated to use .copy instead, not sure if it is better - now it creates an extra copy of the in-memory report which we then use to copy to a new one and set the existing ID. This will all change anyway when we switch to internal FHIR format

prime-router/src/main/kotlin/tokens/OktaAuthentication.kt
clediggins-usds: @whytheplatypus can you address this?  I'm not certain of the scope of changes to auth.

prime-router/src/main/kotlin/azure/ActionHistory.kt
clediggins-usds: Ticketed: #2945

prime-router/src/main/kotlin/azure/Authentication.kt
```diff
@@ -17,12 +18,26 @@ enum class PrincipalLevel {
     USER
 }
 
+data class AuthenticationResult(
+    val authenticated: Boolean,
+    val status: String?
+)
+
+interface Authenticator {
```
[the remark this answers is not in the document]
clediggins-usds: @whytheplatypus Probably a good idea to reduce scope of each individual PR.

prime-router/src/main/kotlin/azure/Event.kt
```diff
@@ -75,23 +82,91 @@ abstract class Event(val eventAction: EventAction, val at: OffsetDateTime?) {
...
+        val defaultClause = if (defaultsQueueParam.isEmpty()) "" else "$messageDelimiter$defaultsQueueParam"
+        val routeToClause = if (routeToQueueParam.isEmpty()) "" else "$messageDelimiter$routeToQueueParam"
+
+        // generate the process queue message
+        return "$eventType$messageDelimiter$eventAction$messageDelimiter$reportId$messageDelimiter$options" +
+            "$defaultClause$routeToClause$afterClause"
+    }
+
+    override fun equals(other: Any?): Boolean {
+        return other is ReportEvent &&
```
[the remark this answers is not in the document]
clediggins-usds: Yep

prime-router/src/main/kotlin/azure/Event.kt
clediggins-usds: Pulled our splitting and verification into separate private function, included parts length checking on a per-event-type basis.

prime-router/src/main/kotlin/azure/WorkflowEngine.kt
```diff
@@ -288,6 +304,161 @@ class WorkflowEngine(
...
+        actionHistory: ActionHistory
+    ) {
+        val errors: MutableList<ResultDetail> = mutableListOf()
+        val warnings: MutableList<ResultDetail> = mutableListOf()
+
+        db.transact { txn ->
+            val task = db.fetchAndLockTask(messageEvent.reportId, txn)
+            val id = task.reportId
+            val reportFile = db.fetchReportFile(id, org = null, txn)
+            val header = createHeader(task, reportFile, null, null, null)
```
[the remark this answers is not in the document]
clediggins-usds: Good point. Updated.

jimduff-usds (changes requested):
During the **sync end2end smoke test**, I saw this in the logs:

`prime_dev_1     | [2021-11-08T19:26:52.725Z] Skipping write to metadata table because reportId ab395408-29ca-43a6-a08b-4b079a93901c does not exist in report_file.
`
And sure enough, a bit later I see these logs where its creating the report_file row.

```
prime_dev_1     | [2021-11-08T19:26:55.563Z] Saved to ACTION: receive, id=15
prime_dev_1     | [2021-11-08T19:26:55.588Z] Saved to REPORT_FILE: ab395408-29ca-43a6-a08b-4b079a93901c (ignore.ignore-simple-report --> action receive)

```

The solution is to put the code that inserts into covid_result_metadata into the ActionHistory code, and have it happen after the create of the parent/incoming/received report.    This is a much better place for this code.

Be careful:  currently we ONLY make entries in covid_result_metadata for the initially received report, not for any later reports.    I believe this is erroneous, we should do it for every step, however that needs to be done carefully because its a ton of data, may cause slowness, and just needs to be done carefully as a separate task.  So for now, just insert metadata for the received file only.

Further note:   the test for the sync end2end passed, despite this error.

-------

I get an exception when I run ./prime test --sequential --run end2end, in the 'async' version of the test:

(note the use of the word sequential is very confusing and overloaded now....here, it means to run each smoke test sequentially, not in parallel)

The CLI code seems to submit and get back a successful response, but when it goes to query the database for the first time, its getting a NullPointerException.

```
Running end2end asynchronously -- with query param
Created datafile ./build/csv_test_files/pdi-covid-19-c6e1b3be-0239-4e8a-a62f-5e8f89bb1488-20211108142858.csv
Posting of async report succeeded with response code 201
{
  "id" : "4a6fd2ff-5f16-4774-acc6-184dde67e40d",
  "timestamp" : "2021-11-08T19:29:05.620842Z",
  "topic" : "covid-19",
  "reportItemCount" : 25,
  "destinations" : [ ],
  "destinationCount" : 0,
  "warningCount" : 0,
  "errorCount" : 0,
  "errors" : [ ],
  "warnings" : [ ],
  "consolidatedErrors" : [ ],
  "consolidatedWarnings" : [ ]
}
Id of submitted report: 4a6fd2ff-5f16-4774-acc6-184dde67e40d
'topic' is in response and correctly set to 'covid-19'
No errors detected.
Polling for ReportStream process results.  (Max poll time 180 seconds)
....................
Exception: java.lang.NullPointerException, null: gov.cdc.prime.router.cli.tests.CoolTest$Companion.processActionResultQuery(TestReportStream.kt:786)
gov.cdc.prime.router.cli.tests.CoolTest$queryForProcessResults$1.invoke(TestReportStream.kt:477)
gov.cdc.prime.router.cli.tests.CoolTest$queryForProcessResults$1.invoke(TestReportStream.kt:476)
gov.cdc.prime.router.azure.DatabaseAccess.transact$lambda-0(DatabaseAccess.kt:74)
org.jooq.impl.DefaultDSLContext.lambda$transaction$5(DefaultDSLContext.java:611)
org.jooq.impl.DefaultDSLContext.lambda$transactionResult0$3(DefaultDSLContext.java:549)
org.jooq.impl.Tools$4$1.block(Tools.java:5243)
java.base/java.util.concurrent.ForkJoinPool.managedBlock(ForkJoinPool.java:3128)
org.jooq.impl.Tools$4.get(Tools.java:5240)
org.jooq.impl.DefaultDSLContext.transactionResult0(DefaultDSLContext.java:597)
org.jooq.impl.DefaultDSLContext.transactionResult(DefaultDSLContext.java:521)
org.jooq.impl.DefaultDSLContext.transaction(DefaultDSLContext.java:610)
gov.cdc.prime.router.azure.DatabaseAccess.transact(DatabaseAccess.kt:74)
gov.cdc.prime.router.cli.tests.CoolTest.queryForProcessResults(TestReportStream.kt:476)
gov.cdc.prime.router.cli.tests.CoolTest.pollForProcessResult(TestReportStream.kt:452)
gov.cdc.prime.router.cli.tests.CoolTest$pollForProcessResult$1.invokeSuspend(TestReportStream.kt)
```

prime-router/src/main/kotlin/azure/WorkflowEngine.kt
clediggins-usds: There is very low chance of a null sender - this has already been pulled as part of initial receive. Updated to check anyway just in case something odd happens.

prime-router/src/main/kotlin/Report.kt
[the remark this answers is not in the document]
clediggins-usds: After discussion with team members I am going to leave it as it is, and not use .copy.  It is tech debt, added numerous comments to indicate this.
Overall it reduces our performance improvements to need to re-create the report in memory again just to copy it to get the id. The structure of the code precludes passing in the existing ID in the first creation since our serializers are both parsing AND creating the report.

prime-router/src/main/kotlin/azure/WorkflowEngine.kt
clediggins-usds: We can't use separate transactions here due to the way we are locking tasks - the fetch is running .forUpdate() so that the row is locked until the transaction commits, preventing another instance of the function from grabbing the same task.

jimduff-usds (approved):

prime-router/src/main/kotlin/azure/ReportFunction.kt
```diff
@@ -103,103 +109,140 @@ class ReportFunction : Logging {
...
+            validatedRequest.content,
+            validatedRequest.defaults,
+            errors,
+            warnings
+        )
+
+        // if we are processing a message asynchronously, the report's next action will be 'process'
+        if (isAsync && report != null) {
+            report.nextAction = TaskAction.process
+        }
```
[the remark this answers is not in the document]
clediggins-usds: Unsure what you mean, and code has moved a bit since this comment was made.

carlosfelix2 (approved):

prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -713,4 +756,194 @@ class ActionHistory {
...
+     * Saves covid result metadata to the database
+     */
+    private fun insertCovidMetadata(
+        txn: Configuration
+    ) {
+        covidMetaDataRecords.forEach {
+            context?.logger?.info("Writing deidentified report data to the DB")
+            // wrap the insert into an exception handler
+            try {
+                DSL.using(txn).newRecord(COVID_RESULT_METADATA, it).store()
```
whytheplatypus, line 928: This behavior used to use the `saveTestData` function from `DatabaseAccess.kt`. It looks like that has different behavior, is what's here equivalent or better?

prime-router/src/main/kotlin/azure/Event.kt
whytheplatypus: I think there's a problem here. As an example the case where a `default` is not passed but an `at` is results in what should be the `after` being in slot 5.

prime-router/src/main/kotlin/azure/ReportFunction.kt
```diff
@@ -103,103 +109,140 @@ class ReportFunction : Logging {
...
+        // determine if we should be following the sync or async workflow
+        val isAsync = processingType(request) == ProcessingType.Async
+        val errors: MutableList<ResultDetail> = mutableListOf()
+        val warnings: MutableList<ResultDetail> = mutableListOf()
+        // The following is identical to waters (for arch reasons)
+        val validatedRequest = validateRequest(workflowEngine, request)
+        warnings += validatedRequest.warnings
+        handleValidation(validatedRequest, request, actionHistory, workflowEngine)?.let {
+            return it
+        }
```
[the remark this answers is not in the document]
whytheplatypus: Should be handled during #2926

prime-router/src/main/kotlin/azure/ReportFunction.kt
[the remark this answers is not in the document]
whytheplatypus: We should be able to move this into the other `if` that checks for `isAsync` to make it easier to figure out what the differences are between the flows.

prime-router/src/main/kotlin/azure/ReportFunction.kt
```diff
@@ -85,7 +71,27 @@ class ReportFunction : Logging {
...
+        val sender = workflowEngine.settings.findSender(senderName)
+            ?: return HttpUtilities.bad(request, "'$CLIENT_PARAMETER:$senderName': unknown sender")
+        val actionHistory = ActionHistory(TaskAction.receive, context)
+        actionHistory.trackActionParams(request)
+
+        try {
+            return processRequest(request, sender, context, workflowEngine, actionHistory)
+        } catch (ex: Exception) {
+            if (ex.message != null)
+                logger.error(ex.message!!, ex)
```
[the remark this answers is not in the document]
whytheplatypus: Good candidate for a tech-debt ticket

prime-router/src/main/kotlin/azure/ReportFunction.kt
```diff
@@ -103,152 +114,159 @@ class ReportFunction : Logging {
...
+            errors,
+            warnings
+        )
+
+        // if we are processing a message asynchronously, the report's next action will be 'process'
+        if (isAsync && report != null) {
+            report.nextAction = TaskAction.process
+        }
+
+        var response: HttpResponseMessage
```
whytheplatypus, line 202: Good candidate for a tech-debt ticket. There's a more fluent builder pattern for responses that would probably help normalize this. Either way we should choose a common response building pattern for the system

whytheplatypus (changes requested):

prime-router/src/main/kotlin/azure/Event.kt
clediggins-usds: Updated.

prime-router/src/main/kotlin/azure/ReportFunction.kt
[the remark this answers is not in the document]
clediggins-usds: Updated

prime-router/src/main/kotlin/azure/ActionHistory.kt
clediggins-usds: Updated

whytheplatypus (approved):

prime-router/src/main/kotlin/azure/ReportFunction.kt
whytheplatypus: #3041

prime-router/src/main/kotlin/azure/ReportFunction.kt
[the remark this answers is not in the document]
whytheplatypus: #3042

PR #2924, opened 2021-11-04
Fix ts erros for ErrorPage

TomNUSDS:
- Fix ts errors and tighten up types in general.
- Add unit tests ErrorPage.

### DIFF
--- a/frontend-react/src/pages/details/Details.tsx
+++ b/frontend-react/src/pages/details/Details.tsx
@@ -11,7 +11,7 @@ import ReportDetails from "./ReportDetails";
 import FacilitiesTable from "./FacilitiesTable";
 
 function useQuery(): { readonly [key: string]: string } {
-    let query = window.location.search.slice(1);
+    const query = window.location.search.slice(1);
     const queryMap = {};
     Object.assign(
         queryMap,
@@ -24,9 +24,9 @@ function useQuery(): { readonly [key: string]: string } {
 }
 
 const DetailsContent = () => {
-    let queryMap = useQuery();
-    let reportId = queryMap ? ["reportId"] : "";
-    let report = useResource(ReportResource.detail(), { reportId: reportId });
+    const queryMap = useQuery();
+    const reportId = queryMap?.["reportId"] || "";
+    const report = useResource(ReportResource.detail(), { reportId: reportId });
 
     return (
         <>
--- a/frontend-react/src/pages/details/Details.tsx
+++ b/frontend-react/src/pages/details/Details.tsx
@@ -1,4 +1,3 @@
-// @ts-nocheck // TODO: fix types in this file
 import { NetworkErrorBoundary, useResource } from "rest-hooks";
 import { Suspense } from "react";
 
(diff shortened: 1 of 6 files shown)


frontend-react/src/pages/details/Details.tsx
```diff
@@ -26,7 +25,7 @@ function useQuery() {
 
 const DetailsContent = () => {
     let queryMap = useQuery();
-    let reportId = queryMap["reportId"];
+    let reportId = queryMap ? ["reportId"] : "";
```
kevinhaube, line 28: Wait, does this work the same way as `queryMap ? queryMap["reportId"] : "" `??

kevinhaube (approved):
Thank you for doing this! Love how clean things look with TS :)

frontend-react/src/pages/details/Details.tsx
kevinhaube: I also believe I've used optional chaining like `queryMap?.["reportId"] : ""`. I love finding new little shortcuts like these!

TomNUSDS: Good catch! Yeah, this seems like a mistake. 
Seems like: `const reportId = queryMap?.["reportId"] || "";`  would make more sense.

frontend-react/src/pages/error/ErrorPage.tsx
```diff
@@ -1,50 +1,52 @@
...
    e.g. <ErrorPage code={CODES.NOT_FOUND_404} />
  */
 export enum CODES {
     UNSUPPORTED_BROWSER = "unsupported-browser",
     NOT_FOUND_404 = "not-found",
+    UNKNOWN = "unknown-error",
+}
+
+interface ErrorPageProps {
+    code?: CODES.UNSUPPORTED_BROWSER | CODES.NOT_FOUND_404 | CODES.UNKNOWN;
```
TomNUSDS, line 19: @kevinhaube This limits the possible values to a subset of values from CODES... but I wonder if I shouldn't just use the enum as a type?  `code?: CODES` ? (LOL.. code reviewing my own stuff.)

frontend-react/src/pages/error/ErrorPage.tsx
```diff
@@ -1,50 +1,52 @@
...
+    error?: string;
+    errorInfo?: React.ErrorInfo;
+    type?: "page" | "message";
 }
 
-function ErrorPageWrapper({ children }: JSX.Element) {
+const ErrorPageWrapper = (props: React.PropsWithChildren<ErrorPageProps>) => {
```
TomNUSDS, line 25: `React.PropsWithChildren` is a nice way to do do this. The `props` then has `props.children` automatically but also can contain OTHER props, too.

frontend-react/src/pages/details/Details.tsx
TomNUSDS: Fixed.

frontend-react/src/pages/error/ErrorPage.tsx
TomNUSDS: Fixed

kevinhaube: I saw this and thought "man, it sucks we'd have to update both this and the enum. I had no idea you could use an enum as a type, though! My brain thought about whether we should make a type, but ultimately we would've just been changing where we updated the list of values from here to that type. The enum typing is great!

PR #2927, opened 2021-11-04
Remove use of azure/CLI action, run commands directly

cwinters-usds:
* Something (not sure what) changed recently in the `azure/login`action
  where the credentials are not passed properly to the `azure/CLI`
  action

Test Steps:
1. Run the `release` and `validate_terraform` actions

## Changes
- Remove use of `azure/CLI` action


ronaldheft-usds (approved):
Reviewed via screen share as well. Great catch that additional spots in the release pipeline needed the change!

Issue #2714, opened 2021-10-21
Replace the use of the SettingsCommands environment with the new Environment class


carlosfelix2:
The SettingsCommands has some environment related code that other commands are using.  An Environment class was created with this functionality, but the existing commands were not updated so to not create a large PR for that initial work.  The issue is to update the rest of the commands to use this new Environment class, so other commands are not referencing SettingsCommands directly.

Acceptance criteria:
1. Refactor the CLI commands to use the new Environment class.
2. Delete the unused code from SettingsCommands.

PR #2928, opened 2021-11-04
2714 Refactored environment settings to be in a common place

carlosfelix2:
This PR refactors the environment settings to be in a common place.

Test Steps:
1. Run the unit tests, integration tests and smoke test

## Changes
- use the Environment class instead of directly getting the environment variable.
- Consolidated all environment data in one place

### Testing
- [x] Tested locally?
- [ ] Added tests?

### Process

## Fixes
- #2714

    - new endpoint A
    - new endpoint B


whytheplatypus (changes requested):
The `Environment.kt` is used almost exclusively by the cli, do you think we should expand it to be the place to store expected environment variables and their sane defaults?

carlosfelix2:
> The `Environment.kt` is used almost exclusively by the cli, do you think we should expand it to be the place to store expected environment variables and their sane defaults?

We could, but it would depend what values and what environment variables.  Some are very specific to the environment, E.g. the blob store connection string which I do not think should be defaulted in code.  You can bring that up at the engineering sync. This PR was meant just to clean up what was already there, so I suggest anything new we want should be a new issue.

whytheplatypus (approved):

prime-router/src/main/kotlin/azure/HttpUtilities.kt
```diff
@@ -16,13 +17,6 @@ import java.time.OffsetDateTime
 import java.time.ZoneOffset
 import java.time.format.DateTimeFormatter
 
-enum class ReportStreamEnv(val urlPrefix: String) {
-    TEST("https://test.prime.cdc.gov"),
-    LOCAL("http://" + (System.getenv("PRIME_RS_API_ENDPOINT_HOST") ?: "localhost") + ":7071"),
-    STAGING("https://staging.prime.cdc.gov"),
-    PROD("not implemented"),
```
clediggins-usds, line 23: Is this change in functionality intended?  Currently, if tests were run in prod they would have no effect because of this value. Now the test would use the actual PROD url. Do we want to be able to run tests against real prod endpoints?

carlosfelix2: There was really no check in the code for "not implemented".  It would have just crashed due to not having a proper URL.  I changed it on purpose to allow for the tests, but if we do not want to test in prod a new check somewhere should be done against Environment.PROD.

clediggins-usds: Wouldn't crashing be better than unintentionally running tests that could insert data into prod?

clediggins-usds: This is more of a business rule question: would we want to run these tests in prod as a legitimate use case ever?  Looks like we have not wanted that in the past (since it would have crashed), not sure if we want that in the future. This change was the only functional change to the PR as far as I could tell.

carlosfelix2: It was decided to not allow tests to run in prod, so the environment option for the CLI does not allow prod.  This is mainly because the ignore org does not exist in prod.

clediggins-usds (commented):
Looks good

Issue #2658, opened 2021-10-19
A new user cannot sign into ReportStream 


RickHawesUSDS:
**Describe the bug**
A new user cannot sign into ReportStream when the user is not provisioned with ReportStream(test) application. 

**Impact**
All users must be given access to ReportStream test to login into ReportStream prod. 

**To Reproduce**

Start with a group like DHaz_phd which only has the ReportStream (prod) app and doesn't have the ReportStream (test) app assigned to it. Open two browser tabs, one for the admin and one for the user. 

In Okta:
1. In Okta logged in as an admin
2. add a new user (associated with the group)

In a user browser
2. User does activation process
3. User browses to reportstream.cdc.gov 
4. User signin in
=> Error: User not associated with app

Back in Okta:
1. add the ReportStream (test) app to the user

Back in the user browser
1. User does a sign-in
=> Successful login

**Expected behavior**
Should be able to sign-in with a group that only has the ReportStream app assigned to it. 

**Screenshots**

**Logs**
If applicable, please attach logs to help describe your problem.

**Desktop (please complete the following information):**

**Additional context**
This happened on the non-React site. 

There is likely a configuration that is missing in the ReportStream prod app. 

## TO DO
- This ticket is for the front-end change

sean-usds:
@RickHawesUSDS @MauriceReeves-usds We had a conversation with the Okta reps yesterday and based on their recommendations here is a solution:

1. (Sean) Change all Okta groups to be linked to both "ReportStream" and "ReportStream (test)". This can be done via the Okta web interface. There are only a few dozen groups so it shouldn't take too long.
2. (@kevinhaube) Change the client code to check for Okta clientId based on env var so that it uses "ReportStream" in production and "ReportStream (test)" for staging and testing

Once these changes are in a new user will only have to be added to an Okta group and they will automatically have access to both staging and production. 

Let me know if there are any objections to this solution!

kevinhaube:
@RickHawesUSDS @MauriceReeves-usds I drafted a PR for you to review alongside the release plan

PR #2932, opened 2021-11-04
Okta ClinetID fixes

kevinhaube:
This PR extracts the hard-coded value of `clientId` and pulls the proper ID based on its environment. The 11ty site has instruction to pull the from the env but default to the ReportStream production id. @sean-usds If this env var exists in prod we should make sure we coordinate with devops to change it to the proper id.

Test Steps:
1. pull `kevinhaube/2658-okta-clientid-env`
2. `yarn` and `yarn start`
3. New tab -> navigate to Okta admin panel -> People -> you
4. Remove yourself from the `DHPrimeAdmin` and any state groups (keep a list of these to add yourself back to)
5. Add yourself to the `_TEST` group that only has the ReportStream application (not test) assigned
6. Back to RS tab -> Login
7. You _should_ see an error about not having access to that application because it is pulling the dev/staging key, a.k.a. `ReportStream (test)` client id.
8. Add yourself back to your old groups and ensure you have access to `ReportStream (test)`.
9. Clear all local Okta data
10. Re-login with success (hopefully!)

## Changes
- `oktaConfig.ts` now uses env vars for `clientId`
- Okta config for 11ty now uses env var for `clientId` with the production ID as a default to replace the old dev/staging id

### Testing
- [X] Tested locally?
- [ ] Added tests?

### Process

## Fixes
- #2658


frontend/src/js/prime-web-receiver-pre.js
```diff
@@ -147,7 +147,7 @@ function idleTimer() {
  *  logged in, directs to /daily-data page
  */
 function login() {
-    const OKTA_clientId = '0oa6fm8j4G1xfrthd4h6';
+    const OKTA_clientId = process.env.OKTA_clientId || '0oa6kt4j3tOFz5SH84h6';
```
TomNUSDS, line 150: If you wanted, you could default the ID OKTA_clientId `''` and then runtime check if it's not set correctly. If it's missing, you could throw an error or `console.error()` that it's not set in the env file.

My thought experiment:  The oka id changes for the `test` env at some point in the future (using a test okta account would allow for safer experimentation)... but if it's miss configured in the test .env, then the code defaults to the _production_ id in code (which is less safe). 

So, it seems better it fails with a clear message.

kevinhaube: This is for the 11ty site, which is only in production as I understand it, hence the hard-coded production string fallback. Shockingly, there isn't a .env file present in that directory, but there are a few `process.env` calls being made. I'm guessing it's something we didn't check into the repo. Regardless though, this is merely patchwork for the soon-to-be-deprecated site.

TomNUSDS: Ohhh. Thanks for the clarification!

frontend-react/src/oktaConfig.ts
```diff
@@ -2,7 +2,7 @@ import { OktaAuthOptions } from "@okta/okta-auth-js";
 
 const oktaAuthConfig: OktaAuthOptions = {
     issuer: `https://hhs-prime.okta.com/oauth2/default`,
-    clientId: "0oa6fm8j4G1xfrthd4h6",
+    clientId: process.env.REACT_APP_OKTA_CLIENTID as string,
```
TomNUSDS, line 5: nice

frontend-react/src/oktaConfig.ts
```diff
@@ -13,15 +13,15 @@ const oktaAuthConfig: OktaAuthOptions = {
 };
 
 const oktaSignInConfig = {
-    logo: "https://reportstream.cdc.gov/assets/img/cdc-logo.svg",
+    logo: "/assets/cdc-logo.svg",
```
TomNUSDS, line 16: Good catch since CSP may block if used from staging.
Are we sure we don't need a full path? e.g.
```
redirectUri: `${window.location.origin}/assets/cdc-logo.svg`,
```

TomNUSDS (approved):
LGTM. Maybe verify a full path isn't needed for the image?

frontend-react/src/oktaConfig.ts
kevinhaube: Pull the branch and give it a look, but it worked locally for me with the relative path :)

sean-usds (approved):
:+1:

PR #2941, opened 2021-11-05
Move the download-artifact step to just before the upload one

cwinters-usds:
* This is to see if any intervening step is mucking with the current working directory, or is cleaning up files/directories behind the scenes.


jeremy-page (approved):
👍

Issue #2952, opened 2021-11-05
Unknown `text/html` urls return 404 instead of being routed to react's root index.html page.


TomNUSDS:
**Describe the bug**
Single Page Apps (SPAs) don't issue http requests to load "pages", but they do change the url path in the Address Bar to make it look like the pages are being changed. 

This path is "abstract" and doesn't lead to an actual page on the server. So, browser refreshing the page and bookmarks won't work correctly.

This is a common issue with SPA setups.

**Impact**
Deep links and browser refreshes result in 404 errors from the web server.

**To Reproduce**
1. Go to https://staging.reportstream.cdc.gov/
2. Use react to "navigate" to the About in the navbar (React's SPA means no http requests happen, but notice the address bar's url changes).
3. Once on https://staging.reportstream.cdc.gov/about (and it loads successfully), hit refresh.
**Result**: webserver returns 404.

**Expected behavior/Additional context**
The old "crawl" basically put a copy of the index.html and the javascript files into each possible url path. 
This could just be a file system alias or a http server mapping. The redirect should include all `text/html` media type requests, but NOT other types (e.g. `image/jpeg`, `text/css`, etc).

TomNUSDS:
Suggestion from [here](https://medium.com/medialesson/best-way-to-host-a-single-page-application-spa-in-microsoft-azure-3e70cbd075c3) (in the _ Azure App Service_ section).

```xml
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <system.webServer>
    <rewrite>
      <rules>
        <rule name="SPA" stopProcessing="true">
          <match url="^(?!.*(.js|.css|.png|.jpg|.ico|.svg)).*$" />
          <conditions logicalGrouping="MatchAll">
          </conditions>
          <action type="Rewrite" url="/"  appendQueryString="true" />
        </rule>
      </rules>
    </rewrite>
  </system.webServer>
</configuration>
```

TomNUSDS:
There's an alternate solution. Switch the React app from using url paths to using fragment identifier.

> So we simply changed the routing in the app itself to use a fragment identifier. So instead of expecting `http://hostname.com/path/to/content`, the app now expects `http://hostname.com/#/path/to/content`. That way, index.html is always being served by default, without URL rewriting and without using the ‘Override backend path’ setting. This may negatively affect indexation by search engines ...

I'm going to do more research down this path.

cwinters-usds:
[...]

PR #2958, opened 2021-11-06
Trying experiment to handle missing paths for a SPA

TomNUSDS:
see https://github.com/CDCgov/prime-reportstream/issues/2952

Fix is to simply duplicate the index.html page (which hosts the SPA) to 404.html. Any 404 result should just load this page.


cwinters-usds (approved):
let's give it a shot!

PR #2990, opened 2021-11-09
Test fix of CSP issue 

TomNUSDS:
After deploying a CSP, we're seeing the warning:
> `Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive: "script-src 'self'`

`INLINE_RUNTIME_CHUNK=false` is being set in all the `.env` but I'm not 100% positive it's being used but the build.

Force it in the build line just to make sure and test in staging.


sean-usds (approved):
let's give it a go!

PR #2992, opened 2021-11-09
Modify the production flow to build/release React site

cwinters-usds:
This PR modifies the release github flow to build and deploy the React site, removing the same for the 11ty site. The actual process is already working in staging.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -69,33 +69,8 @@ jobs:
           path: prime-router/build
           retention-days: 7
 
-  build_frontend_release:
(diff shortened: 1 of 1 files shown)


TomNUSDS (approved):
LGTM

sean-usds (approved):
`deploy_release_test` is still using the old static site instead of React but otherwise lgtm!

cwinters-usds:
@sean-usds I noticed that too but I'm not sure what we're going to do with the `test` environment. Probably best to update it to mirrors the others though, will do.

PR #3002, opened 2021-11-10
closes #1885 readme address terraform azurerm import issue

### LINKED ISSUE #1885: tf import CLI command could use some rework
# The Problem

1. `cd operations`
2. `make TF_ENV=dev tf-01-network` (environment and module do not matter)
3. `az login`
4. `tf import <resource> <id>`

The last command fails.

# Background

What it should be doing is using the `tf` script and essentially issuing this command:

```
terraform import \
  --var-file <var-file> \
  -var="terraform_caller_ip_address=${PUBLIC_IP_CIDR?}" \
  <resource> \
  <id>
```

However, somewhere the format is getting messed up and not issuing the script call as above. The logic for this command is here: https://github.com/CDCgov/prime-reportstream/blob/master/operations/app/src/environments/tf#L247

## Original Context

I think I get it, is it because we do: `./tf -e ${ENV_NAME} -c "import a b"`? And thus, it's `"import a b"` (note the quotes around the whole value) that we need to key off of, and not `"import"`.
That seams odd... could we not instead use `${*}` for the `${TFCMD_AFTER}` and take advantage of the `--` breaker on line 90 so that you do: `./tf -e "${ENV_NAME?}" -c import -- resource_type resource_name`?

So this would then turn into a
```bash
    import)
        # Some commands require arguments in a different location
        EXPECTED_RESULTS="${EXPECTED_RESULTS?}0"
        TFCMD_BEFORE=$(echo "${TFCMD?}" | cut -d' ' -f 1)
        terraform ${TFCMD_BEFORE?} \
            --var-file "${TFVARS?}" \
            ${*}
        ACTUAL_RESULTS="${ACTUAL_RESULTS?}${?}"
        ;;
```

_Originally posted by @td-usds in https://github.com/CDCgov/prime-reportstream/pull/1868#r690594220_

JosiahSiegel:
This PR updates the `operations/` README to clarify instances where terraform import may fail for azurerm resources and how to avoid failure.

## Changes
1. Modified `operations/README.md` to include a **Known issues** section with the following:

      ### Known issues
      
      A few `terraform import` commands for azurerm have known to fail unless the resource path is quoted.
      Since `./tf.sh -c ""` expects input within double quotes, use single quotes instead: `./tf -e ${ENV_NAME} -c "import a 'b'"` (note `'b'`).
      
      Shown in [Terraform docs](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/monitor_diagnostic_setting#import) to address the reported [issue](https://github.com/hashicorp/terraform-provider-azurerm/issues/9347).

## Fixes
- #1885


cwinters-usds (approved):
Perfect, thank you! :raised_hands:

Issue #2581, opened 2021-10-12
Unable to run multiple smoke test runs in Github Actions


carlosfelix2:
We get failures when trying to run ReportStream as docker containers and run the smoke tests against it.  This problem requires research, but some initial research shows that running multiple smoke test runs at the same time may be creating a clash between the docker containers as the hostname used on all the builds is the same.  We see errors like:
```
[2021-10-08T16:29:05.004Z] 16:29:04.988 [pool-2-thread-13][31] ERROR gov.cdc.prime.router.tokens.OktaAuthentication - Cursor returned more than one result
[2021-10-08T16:29:05.004Z] org.jooq.exception.TooManyRowsException: Cursor returned more than one result
[2021-10-08T16:29:05.004Z] 	at org.jooq_3.15.1.POSTGRES.debug(Unknown Source) ~[?:?]
[2021-10-08T16:29:05.004Z] 	at org.jooq.impl.Tools.fetchOne(Tools.java:2325) ~[jooq-3.15.1.jar:?]
[2021-10-08T16:29:05.004Z] 	at org.jooq.impl.ResultQueryTrait.fetchOne(ResultQueryTrait.java:506) ~[jooq-3.15.1.jar:?]
[2021-10-08T16:29:05.005Z] 	at gov.cdc.prime.router.azure.DatabaseAccess.fetchSetting(DatabaseAccess.kt:309) ~[prime-router-0.1-SNAPSHOT.jar:?]
[2021-10-08T16:29:05.005Z] 	at gov.cdc.prime.router.azure.SettingsFacade$putSetting$1.invoke(SettingsFacade.kt:161) ~[prime-router-0.1-SNAPSHOT.jar:?]
[2021-10-08T16:29:05.005Z] 	at gov.cdc.prime.router.azure.SettingsFacade$putSetting$1.invoke(SettingsFacade.kt:158) ~[prime-router-0.1-SNAPSHOT.jar:?]
[2021-10-08T16:29:05.005Z] 	at gov.cdc.prime.router.azure.DatabaseAccess.transactReturning$lambda-1(DatabaseAccess.kt:77) ~[prime-router-0.1-SNAPSHOT.jar:?]
```
This last error seems due to having duplicate settings in a database.  There are posts talking about clashes when running multiple builds with docker components and trying to connect to them using a hostname that may be common.  The settings uploader connects to the API to upload settings, and the API connects to the database using a hostname as well.

Again, this issue requires research.

PR #3004, opened 2021-11-10
Stable Smoke Tests in Github Actions

whytheplatypus:
This PR stabilizes the smoke tests in our CI.

Before this the `volume` used by the postgres docker service could be manipulated by other steps or running jobs that caused smoke test failures. This removes the volume for CI runs.

Test Steps:
1. See that the Build Router check has passed, and the the Smoke Test step within that has passed.
2. Feel free to rerun the workflow to see it pass

Another method:
1. choose multiple passing commits from this branch
2. click rerun all jobs on for the Build Router workflow
3. see that they all pass and don't influence eachother
e.g.: 
- https://github.com/CDCgov/prime-reportstream/actions/runs/1472235578
- https://github.com/CDCgov/prime-reportstream/actions/runs/1467101999
- https://github.com/CDCgov/prime-reportstream/actions/runs/1463281590

Any other suggestions to assert that it won't cause the same trouble as before are greatly appreciated.

## Changes
- Use java setup action with caching
- Use gradle wrapper check to assert it's a valid jar
- Explicitly set up a local test environment for the CI run
- Add a postgres docker-compose with no persistent volume (I think this is what caused the earlier behavior)

## Fixes
- #2581 

## To Be Done
- https://github.com/CDCgov/prime-reportstream/pull/3098
- https://github.com/CDCgov/prime-reportstream/issues/3104


MauriceReeves-usds (approved):

.github/workflows/build_hub.yml
```diff
@@ -78,40 +77,43 @@ jobs:
...
-      - name: Dump docker logs on failure
-        if: failure()
-        uses: jwalton/gh-docker-logs@v2.0.2
-        with:
-          images: 'prime-router_settings,prime-router_prime_dev'
+      - name: Load settings
+        run: docker-compose -f docker-compose.yml up settings
 
+      - name: Correct sftp perms
+        run: docker-compose exec -T sftp chmod 777 /home/foo/upload
```
carlosfelix2, line 100: Interesting.  You had to do this for Github?

prime-router/config/detekt/detekt.yml
```diff
@@ -0,0 +1,796 @@
+build:
```
carlosfelix2, line 1: You are mixing work here.  This issue is for the smoke tests.  You should move this detekt work to another issue, so we can play with it and figure out if we want it.

prime-router/docker-compose.yml
```diff
@@ -140,20 +140,17 @@ services:
       - prime-router_build
 
   web_receiver:
-    image: nginx:latest
-    ports:
-      - "8090:80"
+    image: node:14-alpine
```
carlosfelix2, line 143: What are this changes here?

prime-router/docker-compose.postgres.yml
```diff
@@ -0,0 +1,16 @@
+# Build environment
```
carlosfelix2, line 1: What is this?  How is this different from docker-compose.build.yml and why have the same config?

carlosfelix2 (changes requested):

.github/workflows/build_hub.yml
whytheplatypus: yeah, the perms aren't translating when being mounted in the docker container.

prime-router/docker-compose.yml
whytheplatypus: to make the frontend-react changes work. This was breaking with the removal of the `frontend` folder

prime-router/docker-compose.postgres.yml
whytheplatypus: this doesn't use a persistent volume, which I'm theorizing was at least a part of the problem.

prime-router/docker-compose.yml
JosiahSiegel: @whytheplatypus  you beat me to it :) [my-commit](https://github.com/CDCgov/prime-reportstream/pull/3062/commits/77f9a61a0e1eaeb3dd543d74ca82898fcafa26ad)

Will your changes fix the issue with running .cleanslate.sh? If so, I can delete my PR.

whytheplatypus: @JosiahSiegel oh awesome! lets use yours so it's not tied to the rest of this CI work.

JosiahSiegel: Sure. Once your commit is merged into master, i'll fetch your changes and test on my branch. Up to you if you want to back out any commits.

cwinters-usds (approved):
LGTM - I like standardizing on `gradlew` since it seems that's what everyone is using; thanks much for tracking this down and experimenting!

carlosfelix2 (approved):

PR #3062, opened 2021-11-15
fixes #3035 update dev env to use react fe

### LINKED ISSUE #3035: Update `cleanslate.sh` to use `frontend-react/`
**Describe the bug**
`cleanslate.sh` fails after the removal of `frontend/` (replaced by `frontend-react/`).

**Impact**
Local development testing is impacted.

**To Reproduce**
1. cd `prime-router`
2. Run `./cleanslate.sh`
3. Wait for failure when vault is brought up with `docker-compose up --detach vault`

**Expected behavior**
`./cleanslate.sh` should complete without error, especially if running with `--prune-volumes`.

**Screenshots**
image

JosiahSiegel:
This PR fixes `cleanslate.sh` by adding `frontend-react/` dockerfile.

Test Steps:
1. Before this PR, running `./cleanslate.sh` fails with the removal of the old `frontend/` directory.
2. After this PR, `./cleanslate.sh` should succeed.

## Changes
- Added `Dockerfile_dev` to `frontend-react/`
- Changed `docker-compose.yml` to use `frontend-react/` as the **web_receiver** service.

### Testing
- [x] Tested locally?
- [ ] Added tests?

## Fixes
- #3035


JosiahSiegel:
Potentially already fixed with PR: https://github.com/CDCgov/prime-reportstream/pull/3004
Waiting for confirmation.

JosiahSiegel:
This PR will proceed after #3004 is merged

frontend-react/Dockerfile_dev
```diff
@@ -0,0 +1,20 @@
+FROM node:14-alpine
```
cwinters-usds, line 1: This builds and runs ok for me locally, but I'm curious why you didn't just replace the `Dockerfile` in this directory? 

Also, naming convention-wise it looks like we use in `prime-router` a `.` vs a `_` in the filename, so if we keep this let's call it `Dockerfile.dev`.

frontend-react/package.json
```diff
@@ -123,6 +123,7 @@
         "eslint": "^7.26.0",
         "eslint-config-airbnb": "^18.2.1",
         "eslint-config-prettier": "^8.3.0",
+        "eslint-config-react-app": "^6.0.0",
```
cwinters-usds, line 126: was this added by you? or just the result or running yarn?

cwinters-usds (approved):
See comments regarding file naming; but if the front-end folks think this is good then :+1: Thank you!

frontend-react/Dockerfile_dev
JosiahSiegel: The old `frontend/` directory used a `Dockerfile_dev` for ./cleanslate and docker-compose. I'm not aware of the impact vs the native Dockerfile, so I named with `_dev` to match what it was before. If there is no fear with replacing `Dockerfile` and NOT use `_dev`, then sure I can replace it

frontend-react/package.json
JosiahSiegel: It was a result of running yarn, but I included it due to the new dependencies with the new fe dockerfile

JosiahSiegel: It was a result of running yarn, but I included it due to the new dependencies with the new fe dockerfile

carlosfelix2:
I am thinking we should not run the front end as part of the backend build.  The front end is even going on separate deployments.

carlosfelix2 (approved):

PR #3066, opened 2021-11-15
[draft] Fix typescript errors in Login.tsx

TomNUSDS:
- Passing config around as an opaque type was problematic. Just load the config data directly in the only file that needs to access it.
- Removed `ts-nocheck` from top of file.
- Tested locally but watch for login issues when deployed to staging.


TomNUSDS:
Going to just merge these changes into an existing PR.

PR #3074, opened 2021-11-16
Removed the entire response body from the data stored in action history

jimduff-usds:
This PR changes the download action history so it no longer stores the verbose response.

Test Steps:
1. Do a download, then confirm that just the brief http status string (eg, "OK") appears in the action.action_response field in the database.

### Testing
- [ ] Tested locally?
- [ ] Added tests?


cwinters-usds (approved):
Pretty straightforward, glad this is an easy fix. :+1:

Issue #2447, opened 2021-09-30
California SimpleReport Centralized CLIA 


Adrian-Brewster:
California is requesting that we send all SimpleReport data through a centralized CLIA. 

This would include putting the centralized CLIA in MSH 4.2 and moving the sending facility name/CLIA to OBX. 

we will do this by creating a map in the translation configuration that lets us specify: sender1 : CLIA1 , sender2 : CLIA2 etc.

Adrian-Brewster:
Made deployment last Friday.  Need to update settings.

brick-green-agile6:
verified that MSH-4.2 was sent with CDPH000085 in file to CA. Still need CA to verify on their end.

PR #3080, opened 2021-11-16
Bg 2447 setting clia for sender FIX

brick-green-agile6:
This PR fixes the cliaForSender functionality by adding a call to the database. That call is required to collect the report ancestor and query the results metadata table to select the sender ID. Included in the PR is a change to refactor the cliaForSender code out of the createMessage function and into a dedicated function. Additionally, the unit test has been edited to test that new function explicitly. 

Test Steps:
1. pull CSV file from SimpleReport out of blobstorage
2. edit file to set patient state to CA
3. Set cliaForSender on california with "image_mover.default" and new CLIA (ex CDPH000075)
4. start docker (docker-compose up)
5. curl CSV file into docker
6. check that MSH-4-2 has been set to the new CLIA

## Changes
- moved cliaForSender to separate function
- added jooq query to retreive data from covid result metadata
- changed unit test to explicitly test cliaForSender function

### Testing
- [X] Tested locally?
- [X] unit test 'test cliaForSender'

## Fixes
- #2447


RickHawesUSDS (changes requested):
This feels like a violation of the separation of concerns. You shouldn't need to query the DB within a serializer. That is the work to give the serializer all the data it needs to make should be done beforehand.

prime-router/src/main/kotlin/serializers/Hl7Serializer.kt
```diff
@@ -25,6 +25,7 @@ import gov.cdc.prime.router.Schema
 import gov.cdc.prime.router.SettingsProvider
 import gov.cdc.prime.router.Source
 import gov.cdc.prime.router.ValueSet
+import gov.cdc.prime.router.azure.WorkflowEngine
```
RickHawesUSDS, line 28: This feels like a violation of the separation of concerns. You shouldn't need to query the DB within a serializer. That is the work to give the serializer all the data it needs to make should be done beforehand.

RickHawesUSDS (changes requested):
I suggest that instead of doing all the work in the serializer you adjust the pipeline to give you senderID as part of the report the serializer gets. This makes it available to other serializer.  See me if this doesn't make sense.

MauriceReeves-usds (approved):

Issue #2803, opened 2021-10-27
Primary Health Onboarding


anshulkumar-usds:
Onboarding operations for Primary Health

Adrian-Brewster:
Moved to prod last week. No issues.

PR #3090, opened 2021-11-16
Primary Health to production

jbiskie:
This PR completes the go live steps for Primary Health.  It sets processing_mode_code to "P" and adds SNOMED value 82334004 "Indeterminate" to the covid-19/test_result value set.  

Test Steps:
1. Send Primary data file with 82334004 as value for test_result.  Message properly routes and produces no warnings for test_result.

## Changes
- Add code/display to covid-19/test_result value set in covid-19.valuesets
- Add value to 'when' list in OBX-8 mapper
- Change Primary schema processing_mode_code from "T" to "P"
- Moves Primary schema location from /schemas/direct/ to /schemas/hl7
- Updates all the schema documentation markdown files (due to new result value)

### Testing
- [x] Tested locally?

## Fixes
- #2803

### DIFF
--- a/prime-router/src/main/kotlin/Mappers.kt
+++ b/prime-router/src/main/kotlin/Mappers.kt
@@ -520,6 +520,7 @@ class Obx8Mapper : Mapper {
                 "840535000" -> "A" // Antibody to severe acute respiratory syndrome coronavirus 2 (substance)
                 "840534001" -> "A" // Severe acute respiratory syndrome coronavirus 2 vaccination (procedure)
                 "373121007" -> "N" // Test not done
+                "82334004" -> "N" // Indeterminate
                 else -> null
             }
         }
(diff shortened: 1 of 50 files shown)


prime-router/src/main/kotlin/Mappers.kt
```diff
@@ -520,6 +520,7 @@ class Obx8Mapper : Mapper {
                 "840535000" -> "A" // Antibody to severe acute respiratory syndrome coronavirus 2 (substance)
                 "840534001" -> "A" // Severe acute respiratory syndrome coronavirus 2 vaccination (procedure)
                 "373121007" -> "N" // Test not done
+                "82334004" -> "N" // Indeterminate
```
jimduff-usds, line 523: This obx8 mapper seems like it should be in a value set.   Or maybe its a good first candidate for @carlosfelix2 's database table lookup feature.

jimduff-usds (approved):
Nice.  Good to see Primary going live.

MauriceReeves-usds (approved):

PR #3118, opened 2021-11-18
Revert move to gh actions path filter

whytheplatypus:
Before this a required check would not run when changes weren't made to router files. GitHub Actions isn't fully ready for mono-repos: https://github.community/t/status-check-with-path-filter/18049

This PR moves back to using https://github.com/dorny/paths-filter

Test Steps:
1. See that the check still runs
2. after this is merged the router check should pass but no heavy weight tasks should run if no router files were changed.

## Changes
- re-implement https://github.com/dorny/paths-filter

### Testing
- [ ] Tested locally?
- [ ] Added tests?

### DIFF
--- a/.github/workflows/build_hub.yml
+++ b/.github/workflows/build_hub.yml
@@ -5,15 +5,9 @@ on:
     branches:
       - master
       - production
-    paths:
-      - 'prime-router/**'
-      - '.github/workflows/build_hub.yml'
   push:
     branches:
       - master
-    paths:
-      - 'prime-router/**'
-      - '.github/workflows/build_hub.yml'
 
 env:
   # These are for CI and not credentials of any system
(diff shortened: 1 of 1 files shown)


cwinters-usds (approved):
Bummer this doesn't work, thanks for including the discussion link. :+1:

.github/workflows/build_hub.yml
```diff
@@ -5,26 +5,38 @@ on:
...
 env:
   # These are for CI and not credentials of any system
   DB_USER: prime
   DB_PASSWORD: changeIT!
 
 jobs:
+  pre_job:
+    name: Pre Job
+    runs-on: ubuntu-latest
+    outputs:
```
carlosfelix2, line 21: Strange that the step title in the build is dorny/paths etc.  I can't see a place here to change that.  Oh well.

carlosfelix2 (approved):

PR #3124, opened 2021-11-19
Fix check for toomanycols smoke test

whytheplatypus:
Before this a change to the json error response didn't match the smoke test.

Test Steps:
1. see that the toomanycols smoke test now passes

## Changes
- update the check in the toomanycols test to look at the `message` key in the errors list.


jimduff-usds (approved):
Thanks!

sean-usds:
thanks for the fix here! it's on me that it made it through

PR #3665, opened 2022-01-02
BackstopTime set per receiver, based on next_action_at

jimduff-usds:
This PR enables a fairly robust retry mechanism for BatchFunction, by using a backstopTime that is set per-receiver, based on that receiver's frequency of batching.

Test Steps:
1. Run smokes.  Pay attention to the logs showing the `backstopTime`
2. I tested by hacking in a thrown Exception into BatchFunction, then removing it.

### Testing
- [X] Tested locally?
- [X] Added tests?

## Fixes
- This fixes the problem that frequent batchings that fail in unrecoverable ways would give us days of grief.

### DIFF
--- a/prime-router/src/main/kotlin/azure/BatchDeciderFunction.kt
+++ b/prime-router/src/main/kotlin/azure/BatchDeciderFunction.kt
@@ -5,6 +5,7 @@ import com.microsoft.azure.functions.annotation.FunctionName
 import com.microsoft.azure.functions.annotation.StorageAccount
 import com.microsoft.azure.functions.annotation.TimerTrigger
 import org.apache.logging.log4j.kotlin.Logging
+import java.time.OffsetDateTime
 import kotlin.math.ceil
 import kotlin.math.roundToInt
 
@@ -42,8 +43,17 @@ class BatchDeciderFunction : Logging {
                         // TODO: to support sending empty batches, we will need to check if the receiver wants that
                         //  and add the queue message regardless. Currently functionality only adds a message if
                         //  there is at least one batch to run
+
+                        // Calculate how far to look back based on how often this receiver batches.
+                        val backstopTime = OffsetDateTime.now().minusMinutes(
+                            WorkflowEngine.getBatchLookbackMins(
+                                rec.timing?.numberPerDay ?: 1, NUM_BATCH_RETRIES
+                            )
+                        )
                         // get the number of messages outstanding for this receiver
-                        val recordsToBatch = workflowEngine.db.fetchNumberOutstandingBatchRecords(rec.fullName, txn)
+                        val recordsToBatch = workflowEngine.db.fetchNumReportsNeedingBatch(
+                            rec.fullName, backstopTime, txn
+                        )
                         val queueMessages = ceil((recordsToBatch.toDouble() / rec.timing!!.maxReportCount.toDouble()))
                             .roundToInt()
                         logger.info(
--- a/prime-router/src/main/kotlin/azure/BatchFunction.kt
+++ b/prime-router/src/main/kotlin/azure/BatchFunction.kt
@@ -7,10 +7,16 @@ import com.microsoft.azure.functions.annotation.StorageAccount
 import gov.cdc.prime.router.Receiver
 import gov.cdc.prime.router.Report
 import org.apache.logging.log4j.kotlin.logger
+import java.time.OffsetDateTime
 import java.util.logging.Level
 
 const val batch = "batch"
 const val defaultBatchSize = 100
+
+/**
+ * Min number of times to retry a failed batching operation.
+ */
+const val NUM_BATCH_RETRIES = 2
 
 /**
  * Batch will find all the reports waiting to with a next "batch" action for a receiver name.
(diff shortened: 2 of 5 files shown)


prime-router/src/main/kotlin/azure/BatchDeciderFunction.kt
```diff
@@ -42,8 +43,17 @@ class BatchDeciderFunction : Logging {
                         // TODO: to support sending empty batches, we will need to check if the receiver wants that
                         //  and add the queue message regardless. Currently functionality only adds a message if
                         //  there is at least one batch to run
+
+                        // Calculate how far to look back based on how often this receiver batches.
```
jimduff-usds, line 47: This is the main change - both here and in BatchFunction, calculate the backstop time on a per-receiver basis, rather than a global constant.   This lets us wait days for batches that only run once a day, and minutes for batches that run every minute, and minimizes the "death spiral" problem without having to invent yet another retry mechanism.

prime-router/src/main/kotlin/azure/BatchFunction.kt
```diff
@@ -38,10 +44,12 @@ class BatchFunction {
             val maxBatchSize = receiver.timing?.maxReportCount ?: defaultBatchSize
             val actionHistory = ActionHistory(event.eventAction.toTaskAction(), context)
             actionHistory.trackActionParams(message)
-            context.logger.info(
-                "BatchFunction for $message" +
-                    " will look for tasks created after $backstopTime, for receiver ${event.receiverName}"
+            backstopTime = OffsetDateTime.now().minusMinutes(
+                WorkflowEngine.getBatchLookbackMins(
```
jimduff-usds, line 48: I separated out the calculation of the number of "lookback minutes" because its much easier to write complete unit tests for it, than a function that returns an exact time.

prime-router/src/main/kotlin/azure/DatabaseAccess.kt
```diff
@@ -93,9 +93,19 @@ class DatabaseAccess(private val create: DSLContext) : Logging {
...
+     * If the [at] time is not null, do an exact match against next_action_at. TODO remove this option - not used.
+     * If the [at] time is null, limit query to when next_action_at is after [backstopTime].
+     *
+     * BatchFunction's job in life is operating on a timer, so it only operates on Tasks with nonnull next_action_at.
+     *
+     * Note that this uses TASK.next_action_at as the backstopTime comparison, rather than TASK.created_at.
+     * This allows us to manually recover from failures by simply setting a TASK's next_action_at in the future.
+     * This also avoids "losing" batch tasks if a receiver's [Receiver.Timing.numberPerDay] changes while there is
+     * work-in-progress waiting to be batch.  For example, we might change the setting from once a day to once an hour.
+     */
```
jimduff-usds, line 107: The backstop time is compared to the next_action_at time, not the created_at time.   I thought about this a lot -- the two good reasons for doing it this way are in the comment here - easier to do manual recover/retry, and it elegantly avoids the problem of "losing" batches if someone suddenly changes a receiver's timing while there's WIP.

prime-router/src/main/kotlin/azure/DatabaseAccess.kt
```diff
@@ -124,6 +133,23 @@ class DatabaseAccess(private val create: DSLContext) : Logging {
...
+        receiverFullName: String,
+        backstopTime: OffsetDateTime,
+        txn: DataAccessTransaction
+    ): Int {
+        return DSL.using(txn)
+            .select(TASK.asterisk())
+            .from(TASK)
+            .where(TASK.NEXT_ACTION.eq(TaskAction.batch))
+            .and(TASK.RECEIVER_NAME.eq(receiverFullName))
+            .and(TASK.NEXT_ACTION_AT.greaterOrEqual(backstopTime))
```
jimduff-usds, line 149: Comparing backstopTime to next_action_at here as well.

prime-router/src/main/kotlin/azure/DatabaseAccess.kt
```diff
@@ -808,15 +818,6 @@ class DatabaseAccess(private val create: DSLContext) : Logging {
             )
         }
 
-        /**
-         * Max number of hours to wait for a Batch task to succeed.
-         * We only run batch on tasks created after the backstop time.
-         */
-        const val BACKSTOP_HOURS = 26L
```
jimduff-usds, line 815: Yuck - my initial pass at this was a global fixed backstop time.   The new per-receiver way is better

prime-router/src/main/kotlin/azure/WorkflowEngine.kt
```diff
@@ -850,6 +849,31 @@ class WorkflowEngine(
                 FileSettings("$baseDir/settings", orgExt = ext)
             }
         }
+
+        /**
+         * Always find tasks at least this old.  This covers for  extended downtime due to a crash,
+         * as well as for 25 hour days etc.
+         */
+        const val BATCH_LOOKBACK_PADDING_MINS: Long = 180 // 3 hours
+
```
jimduff-usds, line 858: This padding is important.   The backstopTime is always at least this big.     In cases where there is an unrecoverable failure inside BatchFunction, we'll get exceptions for 3 hours for that receiver, until it moves on, which is a minor pain.   OTOH, we want the system to be able to go down for considerable times and self-heal upon restart.  Hence the 3 hours.  Note this is purposely NOT a function of the receiver batching frequency.

Note if the system went down for longer, we can easily and safely manually force batch TASKs to run by simply setting their next_action_at to now().

clediggins-usds (approved):
Looks good. I did not run tests locally, Jim has done extensive testing on this.

PR #4606, opened 2022-03-01
Two-column state layout

TomNUSDS:
- Make the long list of states more condensed. This is part of the #4004 but it's taking a while to get Iowa onboarded, so I'm breaking it out.
- Added definition of ELR by using the full term the first time it's mentioned. "Electronic Laboratory Reporting (ELR) connection"

Desktop
image

Mobile
image


drew-usds:
Closing, @TomNUSDS merged these in with #4004

PR #4965, opened 2022-03-30
Fix EditableCompare bug introduced by #4936

TomNUSDS:
Without end-to-end unit tests, some of these features are fragile. I broke the ability to edit content with #4936

I'm opening an issue (#4966) to figure out some way to create unit tests for this component. It might require cypress.


frontend-react/src/components/EditableCompare.tsx
```diff
@@ -149,10 +152,8 @@ export const EditableCompare = forwardRef(
         );
```
TomNUSDS, line 153: Below is what the code was before #4936

kevinhaube (approved):

PR #5230, opened 2022-04-21
Save org list to csv

TomNUSDS:
Added button to end of Org list to save result as a .csv file. Uses the list filter.

Also, `meta` is required, so change interface to reflect that.

### DIFF
--- a/frontend-react/src/components/Admin/OrgsTable.tsx
+++ b/frontend-react/src/components/Admin/OrgsTable.tsx
@@ -177,7 +177,7 @@ export function OrgsTable() {
                                     size="small"
                                     className="padding-1 usa-button--outline"
                                 >
-                                    Save List to CSV File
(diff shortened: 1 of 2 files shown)


sethdarragile6 (approved):
Nice feature! Works locally. It's a little hidden, being at the very bottom like that, tho. Does it look weird to have to between the filter box and the New Org button?

TomNUSDS:
@sethdarragile6 
> Nice feature! Works locally. It's a little hidden, being at the very bottom like that, tho. Does it look weird to have to between the filter box and the New Org button?

Oh, this is a good idea. I'm a bit worried this area might fill up with commands over time, but I guess that we can address that when it becomes a real problem.

TomNUSDS:
@sethdarragile6 
Here's what moving the button looks like:
image

PR #5263, opened 2022-04-26
Upgrade react-scripts to resolve alert

bgantick:
This PR updates `react-scripts` to resolve a dependency alert. Tested locally (patch version bump, no breaking changes).

### DIFF
--- a/frontend-react/package.json
+++ b/frontend-react/package.json
@@ -186,6 +186,5 @@
         "minimist": "^1.2.6",
         "webpack": "^5.71.0"
     },
-    "resolutions": {
-    }
+    "resolutions": {}
 }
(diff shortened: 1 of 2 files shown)


kevinhaube (approved):

frontend-react/yarn.lock
```diff
@@ -7898,9 +8163,9 @@ react-router@5.2.1:
     tiny-warning "^1.0.0"
 
 react-scripts@^5.0.0:
```
TomNUSDS, line 8165: I expected this to be `^5.0.1`

frontend-react/package.json
```diff
@@ -186,6 +186,5 @@
         "minimist": "^1.2.6",
         "webpack": "^5.71.0"
     },
-    "resolutions": {
-    }
+    "resolutions": {}
```
TomNUSDS, line 189: `"react-scripts": "^5.0.0",` should be `"react-scripts": "^5.0.1",` in the package.json file.

The way you've just changed the lock file means if anyone deletes and regenerates it from the package.json, then the version bumps will be lost.

TomNUSDS (commented):
I have questions.

frontend-react/package.json
bgantick: `^x.x.x` syntax installs latest minor/patch version. If someone were to delete/regenerate the lock file they would get the latest `5` version, which is `5.0.1` at the moment.

Issue #5371, opened 2022-05-03
Unable to get authorized by submission history API


carlosfelix2:
**Describe the bug**
Unable to get authorized into the submission history API in staging.

**Impact**
Users or UIs will not be able to use the submission history API. 

**To Reproduce**
1. Login to the staging Okta.
2. Do a get request to a submission history API endpoint (e.g. `/waters/org/hca/submissions`)
3. Verify you get a 401 response.
4. Do a get request to a the lookup table API that also uses the Okta login (e.g. `/lookuptables/list`)
3. Verify you get a good response.

**Expected behavior**
The submission history API does not return a 401 when we are properly logged in to Okta.

**Screenshots**
n/a

**Logs**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

PR #5378, opened 2022-05-03
Fixed issue in history test and similar issue in UI

jimduff-usds:
This PR fixes the issue found in #5371 .   On staging, the history test was attempting to use an Okta token to connect using the Server2Server auth.  The solution is to add a header that tells the system its using Okta auth.

In addition, several frontend-react API calls to the back end were missing the authentication-type: okta header too.  So, I added it there.

### Backend Pipeline Test Steps:

Since this change is only in the test, you can actually run the test against staging.  After doing a build with this branch, do:
```
./prime login --env staging
./prime test --run history --sequential --env staging --key <SECRETKEY>

```
### Frontend test steps

Confirm that the both the Submissions and the Submissions details pages work locally.

## Changes
- added the authentication-type header in the cli test, and in the frontend code.

### Testing
- [X] Tested locally?
- [ ] Added tests?

### Process

## Fixes
- #5371 

## To Be Done
- n/a

### DIFF
--- a/frontend-react/src/resources/AuthResource.ts
+++ b/frontend-react/src/resources/AuthResource.ts
@@ -20,6 +20,7 @@ export default class AuthResource extends Resource {
                 ...init.headers,
                 Authorization: `Bearer ${accessToken}`,
                 Organization: organization || "",
+                "authentication-type": "okta",
             },
         };
     };
--- a/prime-router/src/main/kotlin/cli/tests/HistoryApiTest.kt
+++ b/prime-router/src/main/kotlin/cli/tests/HistoryApiTest.kt
@@ -183,7 +183,7 @@ class HistoryApiTest : CoolTest() {
             HistoryApiTestCase(
                 "simple history API happy path test",
                 "${environment.url}/api/waters/org/$historyTestOrgName/submissions",
-                emptyMap(),
+                mapOf("authentication-type" to "okta"),
                 listOf("pagesize" to options.submits),
                 bearer,
                 HttpStatus.OK,
@@ -194,7 +194,7 @@ class HistoryApiTest : CoolTest() {
             HistoryApiTestCase(
                 "no such organization",
                 "${environment.url}/api/waters/org/gobblegobble/submissions",
-                emptyMap(),
+                mapOf("authentication-type" to "okta"),
                 listOf("pagesize" to options.submits),
                 bearer,
                 HttpStatus.UNAUTHORIZED,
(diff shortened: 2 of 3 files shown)


frontend-react/src/resources/AuthResource.ts
```diff
@@ -20,6 +20,7 @@ export default class AuthResource extends Resource {
                 ...init.headers,
                 Authorization: `Bearer ${accessToken}`,
                 Organization: organization || "",
+                "authentication-type": "okta",
```
jimduff-usds, line 23: UI now specifies OktaAuth for submissions queries.  (Previously submissions queries only supported okta, so the additional header flag was not needed.  Now submissions supports two different auth techniques)

prime-router/src/main/kotlin/azure/SubmissionFunction.kt
```diff
@@ -100,9 +100,7 @@ class SubmissionFunction(
 
             // Confirm the org name in the path is a sender in the system.
             val sender = workflowEngine.settings.findSender(organization) // err if no default sender in settings in org
-                ?: return HttpUtilities.unauthorizedResponse(
-                    request, HttpUtilities.errorJson("Authentication Failed: $organization: unknown sender")
-                )
+                ?: return HttpUtilities.notFoundResponse(request, "$organization: unknown ReportStream sender")
```
jimduff-usds, line 103: submissions API returns a 404 Not Found when the sender is not found.

prime-router/src/main/kotlin/cli/tests/HistoryApiTest.kt
```diff
@@ -183,7 +183,7 @@ class HistoryApiTest : CoolTest() {
             HistoryApiTestCase(
                 "simple history API happy path test",
                 "${environment.url}/api/waters/org/$historyTestOrgName/submissions",
-                emptyMap(),
+                mapOf("authentication-type" to "okta"),
```
jimduff-usds, line 186: This is what was causing the History test to fail.

prime-router/src/main/kotlin/cli/tests/WatersAuthTests.kt
```diff
@@ -189,6 +191,7 @@ class OktaAuthTests : CoolTest() {
         val historyApiTest = HistoryApiTest()
         historyApiTest.outputToConsole = this.outputToConsole
         passed = passed and historyApiTest.runHistoryTestCases(testCases)
+        this.outputMsgs.addAll(historyApiTest.outputMsgs)
```
jimduff-usds, line 194: Fixed an issue in several places where the tests weren't printing output properly when run without the --sequential option.

kevinhaube (approved):
Thanks for adding the header to our frontend calls!

bgantick (approved):

PR #5416, opened 2022-05-06
[SPIKE] Slow loading times for Submissions API calls

### LINKED ISSUE #4927: [SPIKE] Slow loading times for Submission Details API call
## Problem statement
Users experience long loading times when loading the details of a submission through the SubmissionDetails page

Ex:

Requests going to https://prime.cdc.gov/api/history/${ORG}/submissions?pagesize=10&cursor=&sort=DESC are taking over 30 seconds. This may be related to the recent "pagination" changes?

Investigate and make recommendations.

## Acceptance criteria
- [ ] When loading the SubmissionsDetails, the user should not have to wait longer than any other RS API call

## To do
- [ ] Investigate issue and share findings with team.

kevinhaube:
This PR contains research and findings related to API performance on `/submissions` and `/submissions/:id`. To view the flight recorder data, open `prime-router -> docs -> profiler-reports -> jar_2022_05_06....jfr` in IntelliJ IDEA.

Fixes #4927 

## Environment Performance

Due to local and staging data not being representative of what is in production, I wanted to gut check that all three environments were performing relatively the same when it came to API call performance. In my original tests, staging and local ran at about 2 to 3 seconds for both the list and detail Submission calls. Prod was noticeably slower at as low as 6, and as high as 11, seconds for each.

## JVM Profiling

After running the app through a flight recorder, I was able to see that the bulk of CPU usage for each API call is coming from a call named `gov.cdc.prime.router.Metadata.loadDatabaseLookupTables$prime_router()`. This call is part of the WorkflowEngine that's passed in the constructor of the Azure Function.

Screen Shot 2022-05-06 at 12 23 27 PM

> Note, this is focused on the sub-stack, so the percentages of CPU usage are relative to this call, not the app as a whole. You can view the whole call stack via the `.jfr` file provided in this branch.

My assumption is that a new WorkflowEngine is created every time a function is called, since Azure Function endpoints trigger from methods inside this class, and the Metadata object has to be instantiated as a part of that. That could be what's slowing down this  particular API.

The same call is present on other calls in the app, as well. For example, we see it here in a call to get an organization, taking a whopping 5.6% of the total 6.6% of CPU usage for this call:

Screen Shot 2022-05-06 at 12 27 17 PM

## The `WorkflowEngine`

The WorkflowEngine is a top-level object that contains all the helpers and accessors to power the workflow. It handles db access, serializers, blob access, queueing, and more. Metadata is the notable offender for our speed issues, but as it pertains to our endpoints in question, the only necessary parts of this are the db access and settings.  

From what I can see, beyond environmental inconsistency, there's a heavy load placed on the `prime-router` Azure Functions app by loading the full `WorkflowEngine` when only parts of it are necessary. I see a `Builder` that allows us to custom build a WorkflowEngine, but the issue is, the `build` method on this will plug in defaults if values aren't provided, **and** the WorkflowEngine object constructs with defaults. So either way, we seem stuck with instantiating the Metadata object. 

## Recommendation

We should work with `pipeline` on a solution for constructible workflow engines that can be configured for a specific service rather than one general purpose constructor. Fetching services like `getOneOrganization` or getting a list of submissions for a sender shouldn't need access to Metadata in particular, but in best practice, each endpoint or service should be assessed and given only what workflow engine parts it needs.


TomNUSDS:
I think this write-up is super interesting/useful. 

Not sure if the debug dump should go into source control. Maybe better is a github wiki document containing how you did these tests and the recommendations.

kevinhaube:
@TomNUSDS The intention is to just provide it to anyone wanting to see the results and poke around when viewing this branch, I'm going to close the PR once we review in eng sync

kevinhaube:
Closing PR since there's nothing to merge. This was discussed as a part of the RS Eng Sync on May 9th, 2022, and in the sprint review for Experience Team the same day.

Issue #5512, opened 2022-05-18
Make FilterManager configurable on instantiation


kevinhaube:
## Problem statement

Right now, to set any non-default values in our `FilterManager`, we have to use `useEffect` with an empty deps array, which is frowned upon by our linter and also unnecessary mounting behavior. We should be able to pass a `Partial<FilterManager>` to provide defaults.

## To-do

- [ ] Instantiate FilterManager with partial object
- [ ] Pipe those into the filter hooks
- [ ] Filter hooks need to be able to take default values as params and set them

kevinhaube:
#4859 will add in defaulting PageSize

PR #5513, opened 2022-05-18
Adds localSort capabilities to SortSettings

### LINKED ISSUE #5461: Extend SortSettings interfaces and useSortOrder hook for Value Sets admin UI
Current interface
```typescript
type SortOrder = "ASC" | "DESC";
interface SortSettings {
    column: string;
    order: SortOrder;
}
```

This will need to be extended in a way that it can account for it's current use, 

- [x] providing the `SortOrder` value to API calls that are sorted on `createdAt` (see: `/submissions`)

AND a new use, 

- [ ] sorting the value sets on the front-end as a user clicks the table header

This isn't explicitly "dependent" upon the `Table` work, but is a large enough task that
it should be handled separately.

kevinhaube:
This PR extends: 
- `SortSettings` to include a flag for local sorting 
- `ColumnConfig` to include a flag for local sorting
- `Table` to parse these changes and accurately sort table content

Test Steps:
1. This can be tested using the `/daily-data` table in local, sorting on Sent column

## Changes
Screen Shot 2022-05-18 at 12 50 51 PM
Screen Shot 2022-05-18 at 12 50 58 PM

### Testing
- [X] Tested locally?
- [ ] Added tests?

### Process

## Linked Issues
- Fixes #5461 

## To be done
- #5512 (Related to a comment left)

    - new endpoint A
    - new endpoint B

### DIFF
--- a/frontend-react/src/hooks/filters/UseSortOrder.ts
+++ b/frontend-react/src/hooks/filters/UseSortOrder.ts
@@ -3,6 +3,7 @@ import { Dispatch, useReducer } from "react";
 enum SortSettingsActionType {
     CHANGE_COL = "change-column",
     SWAP_ORDER = "swap-order",
+    APPLY_LOCAL_SORT = "apply-local-sort",
     RESET = "reset",
 }
 
@@ -15,13 +16,14 @@ interface SortSettingsAction {
 interface SortSettings {
     column: string;
     order: SortOrder;
+    locally: boolean;
 }
 interface SortFilter {
     settings: SortSettings;
     update: Dispatch<SortSettingsAction>;
 }
 
-const sortSettingsReducer = (
+export const sortSettingsReducer = (
     state: SortSettings,
     action: SortSettingsAction
 ): SortSettings => {
@@ -29,18 +31,24 @@ const sortSettingsReducer = (
     switch (type) {
         case SortSettingsActionType.CHANGE_COL:
             return {
+                ...state,
                 column: payload?.column || state.column,
-                order: state.order,
             };
         case SortSettingsActionType.SWAP_ORDER:
             return {
-                column: state.column,
+                ...state,
                 order: state.order === "ASC" ? "DESC" : "ASC",
+            };
+        case SortSettingsActionType.APPLY_LOCAL_SORT:
+            return {
+                ...state,
+                locally: payload?.locally || false,
             };
         case SortSettingsActionType.RESET: // Also able to manually update settings
             return {
                 column: payload?.column || "",
                 order: payload?.order || "DESC",
+                locally: payload?.locally || false,
             };
         default:
             return state;
@@ -51,6 +59,7 @@ const useSortOrder = (): SortFilter => {
     const [settings, dispatchSettings] = useReducer(sortSettingsReducer, {
         column: "",
         order: "DESC",
+        locally: false,
     });
 
     return {
--- a/frontend-react/src/components/Table/Table.tsx
+++ b/frontend-react/src/components/Table/Table.tsx
@@ -1,20 +1,23 @@
 /* Makes row objects string-indexed */
 import {
     Button,
+    IconArrowDownward,
+    IconArrowUpward,
     IconNavigateBefore,
     IconNavigateNext,
-    IconArrowUpward,
-    IconArrowDownward,
 } from "@trussworks/react-uswds";
 import { NavLink } from "react-router-dom";
-import React from "react";
+import React, { useMemo } from "react";
 
 import {
     CursorActionType,
     CursorManager,
 } from "../../hooks/filters/UseCursorManager";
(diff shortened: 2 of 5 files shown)


frontend-react/src/components/Table/Table.test.tsx
```diff
@@ -2,8 +2,61 @@ import { fireEvent, screen } from "@testing-library/react";
...
+            linkBasePath: key.includes("link") ? "/base/" : undefined,
+            valueMap: key.includes("map") ? sampleMapper : undefined,
+            transform: key.includes("transform") ? transformFunc : undefined,
+            sortable: key.includes("sort") || key.includes("Sort"),
+            localSort: key.includes("local"),
+        } as ColumnConfig;
+    });
+};
+
+const getTestConfig = (rowCount: number): TableConfig => {
```
kevinhaube, line 50: I'm hoping these helpers will let us ween off of `TestTable` so we can ditch it altogether.

frontend-react/src/components/Table/Table.test.tsx
```diff
@@ -28,57 +81,104 @@ const clickFilterButton = () => {
...
     test("Transform columns use transformed value", () => {
-        const transformedElements = screen.getAllByText("transformed");
-        expect(transformedElements).toHaveLength(2);
+        renderWithRouter(<SimpleTable />);
+        expect(screen.getByText("Transformed Value")).toBeInTheDocument();
     });
 });
 
+describe("Sorting integration", () => {
+    test("(Locally) Sorting swaps on header click", () => {
```
kevinhaube, line 130: When trying to assert the content order change, I ran into some issues with the test renderer; clearly it _works_ in browser, but the test renderer wasn't change memoized rows. Help would be much appreciated!

frontend-react/src/hooks/filters/UseSortOrder.ts
```diff
@@ -15,32 +16,39 @@ interface SortSettingsAction {
...
     order: SortOrder;
+    locally: boolean;
 }
 interface SortFilter {
     settings: SortSettings;
     update: Dispatch<SortSettingsAction>;
 }
 
-const sortSettingsReducer = (
+export const sortSettingsReducer = (
```
kevinhaube, line 26: @mreifman Everything SHOULD still be in tact to receive and update sortSettings.order from this filter hook but mind doing a double check for me?

mreifman: This doesn't change any of the behavior around the `order` value exposed by the hook, so no concerns here.

> Also worth considering how we handle the freezing of backend sort when sorting locally. We'd want the data coming back in a predictable way, no? Might be as easy as checking "if locally then DESC else sortSettings.order"

Put another way, we want to avoid superfluous API calls, which would be triggered by changes to `sortSettings.order`. What about tracking the local sort order separately in the hook state to avoid any confusion?

kevinhaube: Ooo, I like the idea. Might create a bit of complexity in the arrow display logic but I need to go in and fix some of that anyway. Worth the effort!

frontend-react/src/components/Table/Table.test.tsx
```diff
@@ -2,8 +2,69 @@ import { fireEvent, screen } from "@testing-library/react";
...
+            columnHeader: `${key[0].toUpperCase()}${key
+                .slice(1)
+                .toLowerCase()}`,
+            feature: key.includes("link")
+                ? {
+                      link: true,
+                      linkBasePath: "/base/",
+                  }
+                : {
+                      // TODO: Add actionable
```
sethdarragile6, line 47: do we need this?

frontend-react/src/hooks/filters/UseSortOrder.ts
```diff
@@ -3,6 +3,8 @@ import { Dispatch, useReducer } from "react";
 enum SortSettingsActionType {
     CHANGE_COL = "change-column",
     SWAP_ORDER = "swap-order",
```
sethdarragile6, line 5: @kevinhaube what exactly do we mean by "local" here, generally?

sethdarragile6 (approved):
@kevinhaube fantastic work! I just had one general question about what you mean by "locally" (versus "globally"?). But I pulled it down and it works adverbly (I mean swimmingly). One little nice-to-have for me would be to be able to have the cursor change when you mouse over the arrow. it would just give a little extra guidance to the user that it is meant to be clicked. But that's the worst thing i saw, so that's pretty good! Thanks for this!

frontend-react/src/hooks/filters/UseSortOrder.ts
kevinhaube: The state to manage local sorting (sorting of the API response array on the front-end) is decoupled from the state that manages API sort, so the actions had to be decoupled as well. Anything updating the front-end sorting is labeled local

Issue #5443, opened 2022-05-10
Valueset editor: UI for success/error on valueset details page


bgantick:
## Problem statement

As a team member of O&O team, I want to be able to view, add and update value sets via a user interface so that the change is easy to make and these settings changes don't have to wait on deployment.

## Acceptance criteria

- [ ] User is able to view a message indicating success if the update or addition of a new value is successful
- [ ] User is displayed an error message if the update or addition fails and the reason behind the failure.

## What you need to know
Component/state handling for success/error on the valueset details page:
![Screen Shot 2022-05-10 at 11.11.05 AM.png](https://images.zenhubusercontent.com/5a6b4b784b5806bc2bc46da6/5c571885-4b97-41ed-9976-84cd771c4108)

![Screen Shot 2022-05-10 at 11.11.26 AM.png](https://images.zenhubusercontent.com/5a6b4b784b5806bc2bc46da6/0960d316-e130-49d0-8f6f-c354ada4e61d)

Design file: https://www.figma.com/file/fKl2M96ylkxYezwjVVUSIX/Internal-Admin-%26-Tools?node-id=0%3A1

Issue #5442, opened 2022-05-10
Valueset editor: UI for valueset details page


bgantick:
## Problem statement

As a team member of O&O team, I want to be able to view, add and update value sets via a user interface so that the change is easy to make and these settings changes don't have to wait on deployment.

## Acceptance criteria

- [ ] User is able to view a list of existing values for a selected value set
- [ ] User is able to update an existing value
- [ ] User is able to add an new value without creating a duplicate value for the same system and version
- [ ] Version field only allows number values or periods.

## What you need to know
Page/components for Valueset details page:
![Screen Shot 2022-05-10 at 11.07.25 AM.png](https://images.zenhubusercontent.com/5a6b4b784b5806bc2bc46da6/4928588f-bdba-439d-8377-0208fcadb08b)

Design file: https://www.figma.com/file/fKl2M96ylkxYezwjVVUSIX/Internal-Admin-%26-Tools?node-id=0%3A1

## Work breakdown

- [x] display with dummy data
- [x] api call and data display
- [x] table functionality for row editing
- [x] api call for edit
- [x] table functionality for adding row
- [x] api call for adding item
- [x] UI cleanup
- [ ] expand unit tests on Valueset Index

kevinhaube:
Table changes needed are ticketed in #5460
Sort changes needed are ticketed in #5461

Issue #5441, opened 2022-05-10
Valueset editor: UI for index page


bgantick:
## Problem statement

As a team member of O&O team, I want to be able to view, add and update value sets via a user interface so that the change is easy to make and these settings changes don't have to wait on deployment.

## Acceptance criteria

- [ ] User is able to view a list of existing value sets in the layout defined by the wireframe
- [ ] User is able to take action to view the list of values/aliases in one value set 

## What you need to know
- Page/components for main "Value Sets" page:
![Screen Shot 2022-05-10 at 11.04.36 AM.png](https://images.zenhubusercontent.com/5a6b4b784b5806bc2bc46da6/9b52ec4d-8114-4641-8487-9a4d29a93143)

Design file: https://www.figma.com/file/fKl2M96ylkxYezwjVVUSIX/Internal-Admin-%26-Tools?node-id=0%3A1

## To do

- [ ]

kevinhaube:
Table changes needed are ticketed in #5460
Sort changes needed are ticketed in #5461

PR #5759, opened 2022-06-09
Experience/5441 valueset UI index page

sethdarragile6:
This PR implements the first draft of the Valuesets Editor UI - Index page! Big props to @ahay-agile6 and @doug-s-nava who mobbed this with me.

Test Steps:
1. Login to RS
2. Navigate manually to /admin/value-sets in the URL (no navigation exists yet, since this is a work in progress)
3. Observe the list of valuesets

## Changes
- Changed the template/placeholder content of ValueSetsIndex.tsx and ValueSetsIndex.test.tsx to populate/test for actual content
- Added a hook (UseLookupTable.ts) and a network layer (LookupTableAPI.ts) with unit tests and a mock server (LookupTableMockServer.ts)
- Screenshot:
image
NOTE: Some artifacts such as the legend (above the header row) and the Add item button will _not_ ultimately be on this page, but will be moved to the Details page when it is developed in #5442 . Also, the Edit buttons will eventually navigate to the Details page- right now they do nothing, which is expected.

### Testing
- [ x ] Tested locally?
- [ x ] (For Changes to /frontend-react/...) Ran `npm run lint:write`? 
- [ x ] Added tests?

## Linked Issues
- Addressed #5441 

## To Be Done
- #5442 #5443


frontend-react/src/hooks/UseLookupTable.test.ts
```diff
@@ -0,0 +1,63 @@
...
+import useLookupTable, {
+    generateUseLookupTable,
+    GetLatestData,
+    GetLatestVersion,
+    getSenderAutomationData,
+} from "./UseLookupTable";
+
+jest.mock("@okta/okta-react", () => ({
+    useOktaAuth: () => {
+        const authState = {
```
kevinhaube, line 15: Check with @doug-s-nava , or take a peek at some of the other Okta mocks; we have a new way to mock this out and shouldn't continue this pattern.

frontend-react/src/hooks/UseLookupTable.ts
```diff
@@ -0,0 +1,100 @@
...
+
+import {
+    lookupTableApi,
+    LookupTable,
+    ValueSet,
+    LookupTables,
+} from "../network/api/LookupTableApi";
+
+export const generateUseLookupTable =
+    <T extends unknown>(tableName: LookupTables) =>
```
kevinhaube, line 12: Is this `unknown` something we can abstract from all possible `T`'s? If not, and if we think it's necessary, a pattern I've seen in the past for type safety of unrelated types is union type checking
```typescript
type ValidExtensions = TypeA | InterfaceOne | TypeC
....<T extends ValidExtensions>
```

frontend-react/src/pages/admin/value-set-editor/ValueSetsIndex.tsx
```diff
@@ -31,41 +26,36 @@ const Legend = ({ items }: { items: LegendItem[] }) => {
...
-        {
-            value: "Patient last name",
-            header: "patient_last_name",
-            type: "Text",
-        },
-    ]);
-    /* We'd pass our config and our API response in this */
+    const valueSetArray = useValueSetsTable();
+
+    const [, setValueSet] = useState<ValueSet[]>([]);
```
kevinhaube, line 54: Is this an artifact of the linter? Can't imagine we'd have a use case for setting a state variable without ever using it 😆

kevinhaube (changes requested):
Just wanna make sure we mock Okta properly before merging, the other stuff are just some observations 👍🏼

frontend-react/src/hooks/UseLookupTable.test.ts
```diff
@@ -0,0 +1,63 @@
...
+import { LookupTables, ValueSet } from "../network/api/LookupTableApi";
+
+import useLookupTable, {
+    generateUseLookupTable,
+    GetLatestData,
+    GetLatestVersion,
+    getSenderAutomationData,
+} from "./UseLookupTable";
+
+jest.mock("@okta/okta-react", () => ({
```
doug-s-nava, line 13: I don't think this mock is doing anything. I removed it and tests still pass.

frontend-react/src/hooks/UseLookupTable.test.ts
```diff
@@ -0,0 +1,63 @@
...
+    },
+}));
+
+describe("test all hooks and methods", () => {
+    beforeAll(() => lookupTableServer.listen());
+    afterEach(() => lookupTableServer.resetHandlers());
+    afterAll(() => lookupTableServer.close());
+
+    test("test GetLatestVersion", async () => {
+        let version = await GetLatestVersion(LookupTables.VALUE_SET);
```
doug-s-nava, line 28: the `let`s in these 3 tests can all be `const`s

frontend-react/src/hooks/UseLookupTable.test.ts
```diff
@@ -0,0 +1,63 @@
...
+        return { authState: authState };
+    },
+}));
+
+describe("test all hooks and methods", () => {
+    beforeAll(() => lookupTableServer.listen());
+    afterEach(() => lookupTableServer.resetHandlers());
+    afterAll(() => lookupTableServer.close());
+
+    test("test GetLatestVersion", async () => {
```
doug-s-nava, line 27: can these test descriptions be a little more specific? maybe `GetLatestVersion returns expected version` etc?

frontend-react/src/hooks/UseLookupTable.test.ts
```diff
@@ -0,0 +1,63 @@
...
+        let version = await GetLatestVersion(LookupTables.VALUE_SET);
+        expect(version).toEqual(2);
```
doug-s-nava, line 29: can we assert a few more things either in this test or subsequent ones? I'd love to see the filter and sort  functionality tested. Maybe a bridge too far, but would also love to see an assertion that axios is being called using the result of `lookupTableApi.getTableList()` and that error states are being handled properly.

frontend-react/src/hooks/UseLookupTable.ts
```diff
@@ -0,0 +1,100 @@
...
+    let response;
+    try {
+        response = await axios(lookupTableApi.getTableList()).then(
+            (response) => response.data
+        );
+
+        let filteredBody: LookupTable[] = response.filter(
+            (tv: LookupTable) => tv.tableName === tableName && tv.isActive
+        );
+        if (filteredBody.length === 0) {
```
doug-s-nava, line 29: I'm a little confused about the purpose of this piece of logic. It seems like in the end we have a list of versions where tableNames match, regardless of whether they are active, with no indication of whether active or inactive versions are being used. Can we just skip the `isActive` check?

frontend-react/src/hooks/UseLookupTable.test.ts
```diff
@@ -0,0 +1,63 @@
...
+            LookupTables.VALUE_SET
+        )) as ValueSet[];
+        expect(data[0].name).toEqual(LookupTables.VALUE_SET);
+    });
+
+    test("test getSenderAutomationData", async () => {
+        let data = await getSenderAutomationData<ValueSet>(
+            LookupTables.VALUE_SET
+        );
+        expect(data.length).toEqual(3);
```
doug-s-nava, line 44: similar to `GetLatestVersion`, can we test the data mapping here? Maybe we can also check that `GetLatestData` is being called with the version returned from the call to `GetLatestVersion`?

frontend-react/src/pages/admin/value-set-editor/ValueSetsIndex.test.tsx
```diff
@@ -13,28 +13,30 @@ describe("ValueSetsIndex (to be updated once page is constructed)", () => {
...
-        expect(editButtons.length).toEqual(rows.length - 1);
-
-        // activate editing mode for first row
-        fireEvent.click(editButtons[0]);
-
-        // assert input element is rendered in edit mode
-        const input = screen.getAllByRole("textbox");
-        expect(input.length).toEqual(1); // Only 1 editable item per row right now
-    });
+    // NOTE: the following test will be applicable to the ValueSets details page
```
doug-s-nava, line 25: 👍

doug-s-nava (commented):
looking good! I've got a few thoughts, but most of my suggestions here are non-blocking. I also see one test failing that should be fixed up.

frontend-react/src/hooks/UseLookupTable.test.ts
sethdarragile6: @kevinhaube as it turns out, I didn't even need this mock! just a copy/paste artifact.

frontend-react/src/hooks/UseLookupTable.ts
sethdarragile6: @kevinhaube turns out we don't need this either! Removed.

frontend-react/src/pages/admin/value-set-editor/ValueSetsIndex.tsx
sethdarragile6: @kevinhaube the linter guided us there, but the real reason is that it is not actually being used at the moment (no updates on this page for MVP). I left that there because the Details page will need this to clone from

frontend-react/src/hooks/UseLookupTable.ts
```diff
@@ -0,0 +1,100 @@
...
+const useLookupTable = <T extends unknown>(
+    tableName: LookupTables
+): ValueSet[] => {
+    const [valueSetArray, setValueSetArray] = useState<ValueSet[]>([]);
+
+    useEffect(() => {
+        getSenderAutomationData<T>(tableName).then((results) => {
+            setValueSetArray(results);
+        });
+        // eslint-disable-next-line react-hooks/exhaustive-deps
```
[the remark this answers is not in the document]
sethdarragile6: @doug-s-nava spot on. for some reason we could not get around this before, but now it seems to be behaving. Fixed!

frontend-react/src/hooks/UseLookupTable.ts
```diff
@@ -0,0 +1,100 @@
...
+    LookupTables,
+} from "../network/api/LookupTableApi";
+
+export const generateUseLookupTable =
+    <T extends unknown>(tableName: LookupTables) =>
+    () => {
+        return useLookupTable<T>(tableName);
+    };
+
+export async function GetLatestVersion(
```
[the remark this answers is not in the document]
sethdarragile6: @doug-s-nava agreed. done.

frontend-react/src/hooks/UseLookupTable.ts
```diff
@@ -0,0 +1,100 @@
...
+            );
+        }
+        const table: LookupTable = filteredBody.sort(
+            (a: LookupTable, b: LookupTable) =>
+                b["tableVersion"] - a["tableVersion"]
+        )[0];
+
+        return table?.tableVersion;
+    } catch (error) {
+        return -1;
```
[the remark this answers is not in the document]
sethdarragile6: @doug-s-nava added some better error handling. Still returning -1 because I have to return _something_. But the experience here tests out as: an error dialog pops up (toast) and no data displays on the page

frontend-react/src/hooks/UseLookupTable.test.ts
sethdarragile6: @doug-s-nava resolved.

frontend-react/src/hooks/UseLookupTable.ts
sethdarragile6: @doug-s-nava I'd like to (and almost did yesterday), but we're trying to address a little fuzziness in regards to how the API handles active/inactive state. We were told that when creating new versions of a lookup table (via updating the table) that the new version would be by default _inactive_, so that OnO could control exactly when the table version goes "live" (via a separate CLI+API call). But the way it seems to play out in reality is: when you create a new version, that new version is automatically activated. Perhaps someone decided along the way that that extra call wasn't needed after all etc? Unknown.

Regardless, the ability to inactivate tables exists (though it wasn't accounted for in the designs) and is probably only used in emergency rollback scenarios. So what we're trying to do here is prioritize grabbing the latest _active_ version, and in the (hopefully) rare case that there _are_ no active versions- at least display the latest _inactive_ version and visually indicate that on the page somehow (TBD)

Now that I've written all that, however, I'm wondering if it's actually better (in light of all this ambiguity) to really just restrict everything to **active only** for the time being. Why don't we add this to our To do list for the details page work (#5442) in the interests of moving this along...

frontend-react/src/hooks/UseLookupTable.test.ts
sethdarragile6: @doug-s-nava let's add these to the To Do list for #5442

frontend-react/src/hooks/UseLookupTable.test.ts
sethdarragile6: let's add these to the To Do list for #5442 @doug-s-nava

frontend-react/src/hooks/UseLookupTable.ts
```diff
@@ -0,0 +1,100 @@
...
+
+export const getSenderAutomationData = async <T extends unknown>(
+    tableName: LookupTables
+): Promise<any[]> => {
+    const version: number = await GetLatestVersion(tableName);
+    if (version === undefined) {
+        console.error("DANGER! no version was found");
+        return [];
+    }
+    const data: T | any[] = await GetLatestData<T[]>(version, tableName);
```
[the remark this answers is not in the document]
sethdarragile6: @doug-s-nava for some reason it gets sticky trying to change that. Maybe we can look into that together.

frontend-react/src/hooks/UseLookupTable.ts
```diff
@@ -0,0 +1,100 @@
...
+                b["tableVersion"] - a["tableVersion"]
+        )[0];
+
+        return table?.tableVersion;
+    } catch (error) {
+        return -1;
+    }
+}
+
+export async function GetLatestData<T extends unknown>(
```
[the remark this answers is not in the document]
sethdarragile6: @doug-s-nava resolved

frontend-react/src/pages/admin/value-set-editor/ValueSetsIndex.tsx
```diff
@@ -31,41 +26,36 @@ const Legend = ({ items }: { items: LegendItem[] }) => {
...
-        dataAttr: "header",
-        columnHeader: "Column header",
+        dataAttr: "system",
+        columnHeader: "System",
     },
     {
-        dataAttr: "type",
-        columnHeader: "Value type",
-        editable: true,
+        dataAttr: "createdBy",
```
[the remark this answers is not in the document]
sethdarragile6: @doug-s-nava not in the designs, but we are using them in the display now

frontend-react/src/hooks/UseLookupTable.test.ts
sethdarragile6: @doug-s-nava yup. resolved

frontend-react/src/hooks/UseLookupTable.ts
doug-s-nava: I am fine with moving this to a follow up ticket! can you add a todo in the code, or update the ticket so we don't forget to revisit?

this is definitely a little more complicated than I appreciated. It might be that there is a slightly more elegant way to do what you're proposing. Or maybe there isn't!

frontend-react/src/hooks/UseLookupTable.ts
```diff
@@ -0,0 +1,106 @@
...
+    try {
+        return await axios
+            .get<T>(endpointHeader.url, endpointHeader)
+            .then((response) => response.data);
+    } catch (e: any) {
+        console.trace(e);
+        showError(
+            `An error occurred while retrieving the latest data: ${e.toString()}`
+        );
+        return new Promise<T[] | T>(() => Promise.resolve());
```
doug-s-nava, line 65: I don't think you'd have a problem just returning `[]` here. Async functions should be smart enough to cast that into a promise.

frontend-react/src/hooks/UseLookupTable.ts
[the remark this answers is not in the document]
doug-s-nava: looks like you got this working!

doug-s-nava (approved):
nice! I'm good to approve, but can you make sure to add todos in the code or open a ticket / modify a ticket for the follow up work on testing and the active table question?

✅ ✅ ✅ ✅ ✅ ✅

ahay-agile6 (approved):

frontend-react/src/hooks/UseLookupTable.ts
sethdarragile6: @doug-s-nava ha! 🤦‍♂️ you're right, and I tried everything _but_ that

kevinhaube (approved):

PR #5882, opened 2022-06-22
Fix/remove dev null for error output from deploy backend/action.yml

jeremy-page:
Tiny PR that changes the backed deploy GHA to write errors out instead of sending them to /dev/null. Regular, non-error output _will_ be discarded


JosiahSiegel (approved):
Looks good. The only other area we may need to confirm output is not blocked is here: https://github.com/CDCgov/prime-reportstream/blob/5dd2b754ef5963b0fe599770307fc1dfcdec5631/operations/Makefile#L123

PR #5922, opened 2022-06-24
dns builds for demo env

JosiahSiegel:
This PR adds dns docker builds for demo environments.

## Changes
- Adds demo environments to dns docker build env array


cwinters-usds (approved):
should work, let's give it a shot

PR #5945, opened 2022-06-27
Exit with a useful message when the Terraform executable is not found

jeremy-page:
Small change that updates the `run-terraform-fmt.sh` git hook script to provide a more accurate error message when the user does not have Terraform installed.


ahay-agile6 (dismissed):
lgtm! :laughing:

JosiahSiegel (changes requested):
For developers that do not rebase, they will not encounter a terraform issue. I would like to support machines that may be used to make a minor change and not need Terraform installed. Instead of forcing a terraform install for EVERY machine, the error message can be updated to be displayed in the CLI instead of the log file. The current error message is "Terraform is not installed", but that is written to the log file. You can instead make the error more obvious.

jeremy-page:
With the change the script checks for the executable before it does anything else and fails saying that Terraform needs to be installed.

PR #5951, opened 2022-06-28
Experience/5304 deprecate meta

doug-s-nava:
The pipeline team is in the process of deprecating a `meta` object that exists on `organization/ settings` entities returned by the API, moving any data from the meta object on the entity itself, one level up. In order to safely deprecate this object fully, the frontend needs to refactor any pieces of code that rely on it to instead look one level up at the `createdBy`, `createdAt` and `version` data that is now located there. This change refactors the necessary frontend code to deal with this change.

Also addresses some metadata related bugs:
* removes `createdAt` field from csvs generated on the organization settings page, as this data is not available from the API call for the organizations list
* adds `county` as a searchable field for organizations
* cleans up some code and messaging in version mismatch warnings

## Test Steps:

1. start servers on this branch with `yarn run start:localdev` from `frontend-react` and `./gradlew run` from `prime-router`

#### sender
1. visit http://localhost:3000/admin/orgsettings/org/abbott
3. click edit
4. change CSV to HL7
5. click preview
6. _VERIFY_: the preview loads with no errors
7. click save
8. _VERIFY_: organization is saved with no errors
9. _VERIFY_: meta for sender appears correctly on org page after save, see screenshot below

#### receiver
1. visit http://localhost:3000/admin/orgsettings/org/ak-phd
3. click edit
4. fill in description field
5. click preview
6. _VERIFY_: the preview loads with no errors
7. click save
8. _VERIFY_: organization is saved with no errors
9. _VERIFY_: meta for sender appears correctly on org page after save, see screenshot below

### Receiver with metadata
Screen Shot 2022-06-28 at 5 13 29 PM

### Sender with metadata

Screen Shot 2022-06-28 at 5 12 30 PM

## Changes

### Testing
- [ ] Tested locally?
- [ ] Added tests?

### Process

## Linked Issues
- Fixes #issue

## To Be Done
- #issue

    - new endpoint A
    - new endpoint B

### DIFF
--- a/frontend-react/src/components/Admin/DisplayMeta.tsx
+++ b/frontend-react/src/components/Admin/DisplayMeta.tsx
@@ -1,24 +1,38 @@
 import { useEffect, useState } from "react";
 
-import { MetaData } from "../../resources/OrgSettingsBaseResource";
+import { MetaTaggedResource } from "../../resources/OrgSettingsBaseResource";
 import { formatDate } from "../../utils/misc";
 
-type Props = {
-    metaObj?: MetaData | undefined;
+type DisplayMetaProps = {
+    metaObj?: MetaTaggedResource;
 };
 
-export const DisplayMeta = (props: Props) => {
-    const [metaData, setMetaData] = useState<MetaData>();
+export const DisplayMeta = ({ metaObj }: DisplayMetaProps) => {
+    const [metaData, setMetaData] = useState<MetaTaggedResource>();
 
     useEffect(() => {
-        if (!props.metaObj) {
+        if (!metaObj) {
             return;
         }
-        setMetaData(props.metaObj);
-    }, [props]);
+        setMetaData(metaObj);
+    }, [metaObj]);
 
-    return metaData ? (
-        <>{`v${metaData.version} [${formatDate(metaData.createdAt)}] 
-        ${metaData.createdBy}`}</>
-    ) : null;
+    // if there is no object with data to display
+    // we can return early. If the metadata object is present but
+    // version, createdAt, and/or createdBy are missing
+    // that case is handled below
+    if (!metaData) {
+        return null;
+    }
+
+    const { version, createdAt, createdBy } = metaData;
+
+    // handle cases where individual metadata are not available
+    const versionDisplay = version || version === 0 ? `v${version} ` : "";
+    const createdAtDisplay = createdAt
+        ? `[${formatDate(metaData.createdAt)}] `
+        : "";
+    const createdByDisplay = createdBy ?? "";
+
+    return <>{`${versionDisplay}${createdAtDisplay}${createdByDisplay}`}</>;
 };
--- a/frontend-react/src/components/Admin/OrgsTable.tsx
+++ b/frontend-react/src/components/Admin/OrgsTable.tsx
@@ -62,7 +62,6 @@ export function OrgsTable() {
     };
 
     const saveListToCSVFile = () => {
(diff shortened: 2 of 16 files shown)


frontend-react/src/components/Admin/DisplayMeta.tsx
```diff
@@ -1,24 +1,34 @@
...
-        setMetaData(props.metaObj);
-    }, [props]);
+        setMetaData(metaObj);
+    }, [metaObj]);
 
-    return metaData ? (
-        <>{`v${metaData.version} [${formatDate(metaData.createdAt)}] 
-        ${metaData.createdBy}`}</>
-    ) : null;
+    if (!metaData) {
```
kevinhaube, line 20: Nitpick, but can this live down with the other return? Components don't benefit from an early return so all the code below runs anyway. This made me think we were conditionally performing what was underneath it at first

frontend-react/src/components/Admin/OrgsTable.tsx
```diff
@@ -67,13 +66,19 @@ export function OrgsTable() {
                         eachOrg.jurisdiction,
                         eachOrg.stateCode,
                         eachOrg.countyName,
-                        new Date(eachOrg.meta.createdAt).toDateString(),
+                        // eachOrg.createdAt
```
kevinhaube, line 69: Left in for later use or forgot to remove?

kevinhaube (approved):
Thanks for handling this!

frontend-react/src/components/Admin/OrgsTable.tsx
doug-s-nava: I left it in for later, but happy to remove. Not sure how likely we are to ever get or need this

frontend-react/src/components/Admin/DisplayMeta.tsx
doug-s-nava: yeah, could take this out and default metaData to an empty object
```suggestion
    const { version, createdAt, createdBy } = {}  = metaData;
```
or better, yet default the state `metaData` to an empty object from the start.

However, I'm confused about the idea of the early return not working. I did a quick smoke test on this idea, and it looks like the function is exiting on line 21 when `metaData` comes through undefined. Can you clarify what you mean here? feel like I may be missing something important...

kevinhaube: There's a sneaky little one-liner at the _veeeery_ bottom of the page that explains it: https://reactjs.org/docs/conditional-rendering.html#preventing-component-from-rendering

> Returning `null from a component’s `render` method does not affect the firing of the component’s lifecycle methods. For instance `componentDidUpdate` will still be called.

It's tricky because this uses class-based component verbiage, but essentially when a functional component is called, it's stepping through a setup lifecycle and I believe it runs through without returning during setup to initiate the hooks, and then renders based on the default state from hooks. Also, [here](https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level), there's a great explanation lower on this page about why hooks _can't_ be conditionally rendered and why this lifecycle is not something you can return early from.

kevinhaube: Retracing what's happening in this, per the lifecycle, I think it's in this order:
- Lifecycle runs setup
- UseState will run
- UseEffect will run
- `metaData` is destructured
- `versionDisplay` is set
- ...
- Component mounted and ready to render
- Renders
- Exits at `null` if `!metaData`

PR #5986, opened 2022-07-01
Add new `adm/resend` api endpoint

TomNUSDS:
- New admin endpoint `adm/resend` that changes the authorization from Azure function to Okta.
- Leave old endpoint in place until the new API is verified.
- Refactor to avoid duplicate code.


JGreenFearless:
@TomNUSDS  Is this something that requires work from the Pipeline Team?

TomNUSDS:
@JGreenFearless It would be great to get a review on it. I posted a few times in the pipeline channel.

doug-s-nava (commented):
code makes sense to me, but not sure how to validate that it works

TomNUSDS:
@doug-s-nava 
Yeah, it always works on localhost. It needs to be deployed to staging to verify it (which is how the bug got introduced before).

The testing would be to verify the current `curl` way of doing resend still works on staging AND use the admin website and verify it works there too. Once this change is up on staging, I'll be doing this verification.

doug-s-nava (approved):
since there's no way to validate before merging and deploying to staging, I'm happy to approve

PR #6020, opened 2022-07-05
add index for item-scoped error message for invalid message type

clediggins-usds:
This PR ensures the correct error message is returned for incorrect HL7 message types when submitted to the full ELR pipeline.

Test Steps:
1. Submit an HL7 message with a type other than OR_R01 to the full ELR pipeline
2. Verify the error message returned specifies that the type submitted is not supported.

## Changes
- Update item logger to include index

### Testing
- [x] Tested locally?

### DIFF
--- a/prime-router/src/main/kotlin/SubmissionReceiver.kt
+++ b/prime-router/src/main/kotlin/SubmissionReceiver.kt
@@ -295,6 +295,10 @@ class ELRReceiver : SubmissionReceiver {
         )
     }
 
+    /**
+     * Checks that a [message] is of the supported type(s), and uses the [actionLogs] to add an error
+     * message for item with index [itemIndex] if it is not.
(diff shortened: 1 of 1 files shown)


prime-router/src/main/kotlin/SubmissionReceiver.kt
```diff
@@ -295,7 +295,7 @@ class ELRReceiver : SubmissionReceiver {
         )
     }
 
-    internal fun checkValidMessageType(message: Message, actionLogs: ActionLogger) {
+    internal fun checkValidMessageType(message: Message, actionLogs: ActionLogger, itemIndex: Int) {
```
carlosfelix2, line 298: You may want to add kdoc

carlosfelix2 (approved):

PR #6157, opened 2022-07-18
Create client SFTP

### LINKED ISSUE #6135: Create SCP server to receive monkeypox data
Create via IaC an SFTP server for each environment

Who will handle credentials?, Stephen Smythe will do customer communication

Four labs sending monkeypox data via reportstream, however labs have trouble so we want an alternate method

Script to copy contents to a dedicated blob storage, then remove, run at least daily, target directory should be ~128g

Mo or Jim Duff for assistance/questions

JosiahSiegel:
This PR creates client accessible SFTP (via SSH certs) for all environments, but only enables for test/staging.

## Linked Issues
- Fixes #6135


jeremy-page (approved):
awesome, looks like we don't have to support a full VM

PR #6203, opened 2022-07-21
Adds fix for the comment not carrying over on inbound HL7 messages

MauriceReeves-usds:
This PR fixes an issue with the NTE segments not carrying over when ingesting HL7. It walks the body of an HL7 message and collates all of the different NTE segment values into a single message that is then put into an NTE segment associated with the first OBX segment in the output HL7 document. The content is trimmed to the length of the `FTE` data type in HL7, which 65,536 characters.

Test Steps:
1. Run `gradle test testIntegration`
2. Run `./prime test`
3. Generate a test file that contains multiple comment sections and for target state of MN or TN
4. Submit the file to your local Docker container for a monkeypox sender
5. Examine the output in your local `build/sftp` folder
6. Confirm it has collated the NTE comments into a single NTE segment associated with the first OBX segment

## Changes
- Schema update to remove duplicated NTE-3 output field
- Method to pull all the NTE segments into a single comment
- Unit tests!

### Testing
- [x] Tested locally?
- [x] Added tests?


brick-green (approved):
works for output to STLT and dcipher

brick-green:
Test HL7 message for Aegis: [monkeypox-e0386067-6d1e-4d6a-b210-c9a4f5862f29-20220720190541.txt](https://github.com/CDCgov/prime-reportstream/files/9185216/monkeypox-e0386067-6d1e-4d6a-b210-c9a4f5862f29-20220720190541.txt)

Note on testing. There are only monkeypox receivers for MN or NY (not TN).

jimduff-usds (approved):
My question is:   I might be misreading, but we seem to be assuming that Notes are 'test_correction_reason's?    I had thought that Notes could serve any purpose, not just reasons for correcting a test.

prime-router/src/main/kotlin/serializers/Hl7Serializer.kt
```diff
@@ -2119,6 +2124,42 @@ class Hl7Serializer(
...
+            val oruR01: ORU_R01 = (message as? ORU_R01) ?: return ""
+            val sb = StringBuilder()
+
+            // walk all the patient results, and the notes that are listed in there, appending to the stringbuilder
+            oruR01.patienT_RESULTAll.forEach { patientResult ->
+                patientResult.patient.nteAll.forEach { patientNote ->
+                    sb.append(patientNote.comment.joinToString(" ") { it.value })
+                    sb.append(" ")
+                }
+                // inside each patient record is the order observation, which also has notes
```
carlosfelix2, line 2145: These notes are sometimes very specific to observations. I am not sure if bundling them up like this would work to generate HL7 back.  For example:
1. Each HL7 message can have patient notes.  Those are for the patient regardless of any observations or orders.
2. In an HL7 you can have multiple orders and each order can have notes. Do you always get only one order (one OBR segment)?
3. Within those orders, you can have multiple observations and the notes will belong to one of those observations.  Do you always get only one OBX segment?

If you get only one of all those segments then you may be fine, but you will not be able to position the notes in the correct location of the HL7 if you bundle them all.

MauriceReeves-usds: @carlosfelix2 I know that they are location specific, but as you also know, our current model for structuring messages doesn't really support tying messages to a specific segment, at least not without some rework, and I don't think that the work is justified given how close you all are with FHIR. At this point, this is a "close enough" solution that ensures the notes carry through.

PR #6232, opened 2022-07-26
Cle/5742 fhir convert

clediggins-usds:
This PR adds in the queues and processors for infrastructure to make FHIR turn back into HL7. More work will be needed  (and more unit tests) as we flesh this out.

Test Steps:
1. run all tests (automatic)
2. Send an HL7 message as a sender with topic Full ELR and verify that it turns into FHIR and then back into HL7 by looking at the files in the Blobstore. All reports, actions, tasks, and lineages will be present in the database.
3. After the HL7 message goes though (this may take a few minutes, since it is going through numerous queues and azure functions) run the following SQL. You should see actions for 'receiver', 'process', 'route', and 'translate'.
```

select a.action_id, a.action_name, t.next_action, t.body_format, rf.body_url
from task t
inner join report_file rf on rf.report_id = t.report_id
inner join action a on a.action_id = rf.action_id
order by a.action_id desc
```

## Changes
- change name of convert queue
- split fhir engine into 3 distinct classes that perform different fhir engine activities
- add 'route'  task action
- remove auto-queue of batch, route, translate events
- add ROUTED_AT to task table
- make each component of the fhir pipeline its own class / queue
- updated tests

## Linked Issues
- https://app.zenhub.com/workspaces/pipeline-6182b02547c1130010f459db/issues/cdcgov/prime-reportstream/5742

### Testing
- [x] Tested locally?
- [x] Added tests?


carlosfelix2:
Add a link to the issue in your PR description to make it easier to access it.  There is a section in the PR template you can add it in.

carlosfelix2:
1. The scope of the changes in this PR are more than what the issue scoped and the AC does not cover features you added, so you need more test steps to cover all the changes you made.  Include steps on checking the DB data you inserted to track all the actions and to make sure the message gets from one function to the other. The unit tests are fine, but you need black box tests for all the pieces here.
2. Because you added those features as part of this PR then I think we need to make a list of the features you added and make issues to track what was done.  I am not sure which or how many as we had not gotten to some of those parts.
3. In the future split work to make the PRs smaller.

carlosfelix2 (changes requested):

prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -401,7 +404,10 @@ class ActionHistory : Logging {
         trackItemLineages(report)
 
         // batch queue messages are added by the batchDecider, not ActionHistory
-        if (event.eventAction != Event.EventAction.BATCH)
```
TurkeyFried, line 404: doc here should be updated to explain ROUTE and TRANSLATE being redirected

prime-router/src/main/kotlin/fhirengine/engine/FHIRTranslator.kt
```diff
@@ -0,0 +1,144 @@
...
+
+            // todo: iterate over each receiver, translating on a per-receiver basis - for phase 1, hard coded to CO
+            val receivers = listOf<String>("co-pdh")
+
+            receivers.forEach {
+                // todo: get schema for receiver - for Phase 1 this is solely going to convert to HL7 and not do any
+                //  receiver-specific transforms
+
+                // todo: do translation, get hl7 message
+                var schema = ConfigSchema("FHIR_HL7", hl7Type = "ORU_R01", hl7Version = "2.5.1")
```
carlosfelix2, line 65: Just FYI: This will change to be loaded from files.

clediggins-usds: Agreed, my understanding is we are not ready to do that yet.

prime-router/src/main/kotlin/azure/ActionHistory.kt
clediggins-usds: Updated.

prime-router/src/main/kotlin/fhirengine/azure/FHIRRouteFunction.kt
```diff
@@ -0,0 +1,59 @@
...
+     */
+    @FunctionName("route-fhir")
+    @StorageAccount("AzureWebJobsStorage")
+    fun process(
+        @QueueTrigger(name = "message", queueName = elrRoutingQueueName)
+        message: String,
+        // Number of times this message has been dequeued
+        @BindingName("DequeueCount") dequeueCount: Int = 1,
+    ) {
+        logger.debug("Routing message: $message for the $dequeueCount time")
```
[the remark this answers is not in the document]
clediggins-usds: I think this is about as un-duplicate as we can get without making it more obfuscated / complicated - this is just the azure function code that then calls the underlying actual logic code.

TurkeyFried (approved):
Tests pass, code looks good, test coverage is pretty good too.

prime-router/src/main/kotlin/fhirengine/azure/FHIRRouteFunction.kt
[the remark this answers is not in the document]
clediggins-usds: Talked with Carlos, refactored to one azure FhirFunction file

carlosfelix2 (changes requested):
Sent an HCA sample and got this error in the logs:
`ERROR gov.cdc.prime.router.fhirengine.engine.FHIRConverter - java.lang.IllegalArgumentException: Unable to find HL7 messages in provided data.`

carlosfelix2 (approved):

PR #6249, opened 2022-07-27
Monkeypox schema updates

jbiskie:
This PR updates the monkeypox schemas

1. Sends a hash of the patient_id to DCIPHER
2. Fixes a hardcoded value sample_type_other_text for DCIPHER

Test Steps:
1. Submit a test monkeypox result to RS. Confirm DCIPHER Patient_id is a hash and changes per patient_id (PID-3); confirm sample_type_other_text is blank.

### Testing
- [x] Tested locally?
- [ ] Added tests?


prime-router/metadata/valuesets/sender-automation.valuesets
```diff
@@ -74,7 +74,7 @@
       display: 1002-5
     - code: 2054-5
       display: Black or African American
-    - code: 2054-5
+    - code: 2058-6
```
jimduff-usds, line 77: 2058-6 is not from the hl7005 valueset, so this is mixing apples and oranges a bit.

2058-6 appears to be part of a much larger and fine-grained race valueset defined here:  https://phinvads.cdc.gov/vads/ViewCodeSystem.action?id=2.16.840.1.113883.6.238#
This is a superset of HL7005.

Furthermore, this sender-automation/race seems to have a special use:   to map all kinds of non-standard fields into a relatively small set of hl7005 race fields.  It appears to be used to map from the "display" value into the "code" value, and never the reverse, since some of the codes are repeated over and over.

All current usages of this valueset go from display -> code, for example:
    csvFields: [{ name: Race, format: $display}]

Normally, this would imply that you do the same - when the lab sends "2058-6" we map it to "2054-5".  So do this instead:

```
- code: 2054-5
  display: 2058-6
```

However, I think that "reverse" mapping doesn't work with HL7?   Leastways, I can't find any example in any schema file that does that. 

The "right" answer would be to load the entire gigantic valueset https://phinvads.cdc.gov/vads/ViewCodeSystem.action?id=2.16.840.1.113883.6.238# into our system as a proper new valueset.  Sigh.

jbiskie: Agree and we need to take a broader look at this + consider which valuesets DCIPHER is going to want us to send. I've removed the race and ethnicity changes for now - we'll tackle those once we have more details from DCIPHER.

jimduff-usds (approved):
lgtm

Issue #6104, opened 2022-07-13
FHIR to HL7 library converts date and date/time formats to HL7 formats


carlosfelix2:
### User Story:
As a developer, I want the HL7 to FHIR conversion library to convert date and date/time FHIR formats to HL7 formats, so I can generate HL7 data.

### Description/Use Case
The format presented by FHIR is different than the one used for HL7.  Need to convert the data as needed by the HAPI HL7 v2 library.

### Risks/Impacts/Considerations
* 

### Dev Notes:
* You can detect that a FHIR type is of type Instant or Date and perform the conversion automatically (no need for something in the configuration).

### Acceptance Criteria
* The FHIR to HL7 v2 conversion library can store a FHIR date into and HL7 v2 date field.
* The FHIR to HL7 v2 conversion library can store a FHIR Instant (date/time) into and HL7 v2 date/time field.

PR #6256, opened 2022-07-28
5746 Generate HL7 SFT segment from FHIR

carlosfelix2:
This PR generates the SFT segment from FHIR data.  It also fixes some bugs found in the process of generating the config.

Test Steps:
1. Check out the branch for this PR.
2. Verify the tests pass by running the command `./gradlew package` and verifying the build is successful.
3. Run the following command to convert a FHIR bundle to HL7.  
`./prime fhirdata --input-file "src/testIntegration/resources/datatests/HL7_to_FHIR/sample_co_1_20220518-0001.fhir" -s metadata/hl7_mapping/ORU_R01/ORU_R01-base.yml`
4. Verify an HL7 message is generated from the conversion and includes the SFT segment.
5. Look at the source HL7 used at `src/testIntegration/resources/datatests/HL7_to_FHIR/sample_co_1_20220518-0001.hl7` and verify the data for the SFT segment is identical with the following exceptions:
  a. SFT-6 requires #6104, but once that issue is merged the date will appear

## Changes
- Added initial set of configs for the FHIR to HL7 conversion
- Had to remove the FHIR path validation at the schema level as some paths will use custom constants that are not available at that time and hence the validation will fail.  Validation will now happen at conversion time instead of schema load time.
- Fixed bug where the top level schema constants were not loading and made it so the arguments of functions will require a custom context to eliminate the possibility of a coding mistake
- Fixed get value not stopping at the first valid value
- Added try/catch around evaluates.  I was getting an index out of bound exception when trying to fetch an extension that did not exist.
- Added support for prefix constants.  This is the same feature that the FHIR Path Engine has for the ext constant.  If you use a constant in FHIR path it can replace a prefix.  E.g., `%'rsext-suffix'` will be converted to `<rsextvalue>suffix`.  
- Fixed an issue where the constant resolver is called at all times and incorrect error out.  Now it returns null instead of an exception

### Testing
- [x] Tested locally?
- [x] Added tests?

### Process

## Linked Issues
- #5746


TurkeyFried (approved):
Testing result:
`SFT|MEDITECH, Inc.^L^^^^MEDITECH&1.3.6.1.4.1.24310&ISO^XX^^^MEDITECH|5.67|Laboratory Applicati|20210810`

This is consistent with the sample file's value:
`SFT|MEDITECH, Inc.^L^^^^MEDITECH&1.3.6.1.4.1.24310&ISO^XX^^^MEDITECH|5.67|Laboratory Applicati|20210810||20210810`

The only difference is the one expected according to the test steps.

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirPathUtils.kt
```diff
@@ -34,59 +34,86 @@ object FhirPathUtils : Logging {
...
-        expressionNode: ExpressionNode
+        expression: String
     ): List<Base> {
-        val resources = defaultPathEngine.evaluate(appContext, focusResource, bundle, bundle, expressionNode)
-        return resources.map {
-            if (it.hasType("Reference") && (it as Reference).resource != null) {
-                it.resource as Base
-            } else it
+        val retVal = try {
+            val resources = defaultPathEngine.evaluate(appContext, focusResource, bundle, bundle, parsePath(expression))
```
clediggins-usds, line 48: Why are we calling this `defaultPathEngine`?  There is no way to provide an alternate path engine - is that for future phases?

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirPathUtils.kt
```diff
@@ -34,59 +34,86 @@ object FhirPathUtils : Logging {
...
     ): Boolean {
-        val value = defaultPathEngine.evaluate(appContext, focusResource, bundle, bundle, expressionNode)
-        return if (value.size == 1 && value[0].isBooleanPrimitive) (value[0] as BooleanType).value
-        else {
-            throw SchemaException("Condition did not evaluate to a boolean type")
+        val retVal = try {
+            val value = defaultPathEngine.evaluate(appContext, focusResource, bundle, bundle, parsePath(expression))
+            if (value.size == 1 && value[0].isBooleanPrimitive) (value[0] as BooleanType).value
+            else {
+                throw SchemaException("Condition did not evaluate to a boolean type")
```
clediggins-usds, line 85: Should this be a `SchemaException`?  Seems like in this case the schema could be fine, but the provided bundle is the thing with the problem. BundleException?  ResourceException?

clediggins-usds (approved):

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirPathUtils.kt
carlosfelix2: It can be called just path engine.  I just grabbed the name Gage has, but I think you are right and there will not be any other type of engine

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirPathUtils.kt
carlosfelix2: Not in this case.  If the condition did not evaluate to a boolean it was because is was not made to be of type boolean.  E.g. someone did `Bundle.id` instead of `Bundle.id.exists()`

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/ConstantResolver.kt
```diff
@@ -76,20 +76,47 @@ object ConstantSubstitutor {
...
-    override fun resolveConstant(appContext: Any?, name: String?, beforeContext: Boolean): Base {
+    override fun resolveConstant(appContext: Any?, name: String?, beforeContext: Boolean): Base? {
         // Name is always passed in from the FHIR path engine
         require(!name.isNullOrBlank())
 
         return when {
             appContext == null || appContext !is CustomContext ->
                 throw PathEngineException("No context available to resolve constant '$name'")
+
+            // Support prefix constant replacement like for URLs similar to %ext in FHIRPathEngine
```
PatriciaPerozoUSDS, line 87: Might be worth adding the example of using this for extensions. Something like:
```// for example: resource: '%resource.source.extension(%`rsext-source-software-vendor-org`).value' ```

**Review by PatriciaPerozoUSDS — APPROVED** [review]
Slightly meta commentary - After reading through your mapping yml files, I'm wondering if we do want to throw an error in evaluateString if it can't evaluate to anything. It would be a schema error for not checking whether the data type exists first/is being parsed incorrectly.

**Comment by carlosfelix2** [comment]
> Slightly meta commentary - After reading through your mapping yml files, I'm wondering if we do want to throw an error in evaluateString if it can't evaluate to anything. It would be a schema error for not checking whether the data type exists first/is being parsed incorrectly.

Now that I think about it, if we evaluate to nothing then it should just be no value as data may or may not have been provided for a specific field.  Everything is really optional in these formats.

File: prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/ConstantResolver.kt (hunk 3 above)
Comment (carlosfelix2) [reply prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/ConstantResolver.kt:87]: Will add.

File: prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirPathUtils.kt [hunk 4]
```diff
@@ -61,73 +53,99 @@ object FhirPathUtils : Logging {
...
     ): String {
-        val evaluated = defaultPathEngine.evaluate(appContext, focusResource, bundle, bundle, expressionNode)
+        val evaluated = defaultPathEngine.evaluate(appContext, focusResource, bundle, bundle, parsePath(expression))
         return when {
             // If we couldn't evaluate the path we should return an empty string
             evaluated.isEmpty() -> ""
 
             evaluated.size > 1 -> {
-                val msg = "Could not evaluate multiple results: $evaluated for path: $expressionNode"
+                val msg = "Could convert to string multiple FHIR path results for path: $expression"
```
PatriciaPerozoUSDS, line 139: This is a clearer message thank you

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirPathUtils.kt
```diff
@@ -137,15 +155,8 @@ object FhirPathUtils : Logging {
...
-                try {
-                    defaultPathEngine.convertToString(evaluated[0])
-                } catch (e: Exception) {
-                    val msg = "Could not parse ${evaluated[0]} to string."
-                    logger.error(msg, e)
-                    throw HL7ConversionException(msg, e)
-                }
-            }
+            // Use the string representation of the value for any other types.
+            else -> evaluated[0].toString()
```
PatriciaPerozoUSDS, line 159: Just using toString threw errors for me in one of the random one off tests I did. I think its worth using `defaultPathEngine.convertToString` instead

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirPathUtils.kt
```diff
@@ -154,36 +165,83 @@ object FhirPathUtils : Logging {
...
+                hl7DateTime.setOffset(hour * 100 + min)
+            }
+            return hl7DateTime
+        }
 
-        return offsetDateTime.format(dateTimeFormatter)
+        val hl7DateTime = DTM(null)
+
+        return when (dateTime.precision) {
+            // HL7 Date time does not support only year, or year and month
```
PatriciaPerozoUSDS, line 186: I'm confused by this comment the HL7 DTM says it supports just year by itself in the spec - https://hl7-definition.caristix.com/v2/HL7v2.6/DataTypes/DTM

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirPathUtils.kt
carlosfelix2: Changed.  I see that it actually calls `primitiveValue`

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirPathUtils.kt
carlosfelix2: Old comment.  Needs to be removed

PatriciaPerozoUSDS (approved):
Fully re-approve the second round of changes

PR #6265, opened 2022-07-29
Automatically convert FHIR Dates and Times to HL7 format

PatriciaPerozoUSDS:
This PR ...
When translating from FHIR back to HL7, automatically convert dates and times to the right format so that we don't have to add per field transforms in the schema. This converts FHIR Dates, Times, DateTimes, and Instants into their associated HL7 DTM or TM.

Relevant Specs:
- FHIR Datatypes: https://build.fhir.org/datatypes.html#primitive
- HL7 DTM: https://hl7-definition.caristix.com/v2/HL7v2.6/DataTypes/DTM
- HL7 TM: https://hl7-definition.caristix.com/v2/HL7v2.8/DataTypes/TM

Test Steps:
1.  Run `./prime fhirdata --input-file "src/testIntegration/resources/datatests/HL7_to_FHIR/sample_co_1_20220518-0001.fhir" -s "src/test/resources/fhirengine/translation/hl7/schema/schema-read-test-01/ORU_R01.yml"`
2. You should see a properly translated Timestamp in the MSH segment `MSH|^~\&|||||20280808112805.0000-0400||ORU^R01^ORU_R01|Bundle/df81fb05-e2aa-4538-8552-b8554bdea307||2.5.1`
3. You can also run the added unit tests to see all the different format types

## Changes
- Updates the evaluateString FhirPathUtils Function to use newly added converter methods
- Added unit tests for each of the methods introduced
- Added a timestamp to the FHIR test schema so that it can be tested using the primeCLI command

### Testing
- [X] Tested locally?
- [X] Added tests?

### Process

## Linked Issues
- Fixes #issue

## To Be Done
- #issue

    - new endpoint A
    - new endpoint B


prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirPathUtils.kt
```diff
@@ -87,6 +95,40 @@ object FhirPathUtils : Logging {
         bundle: Bundle,
         expressionNode: ExpressionNode
     ): String {
-        return defaultPathEngine.evaluateToString(appContext, focusResource, bundle, bundle, expressionNode)
+        val evaluated = defaultPathEngine.evaluate(appContext, focusResource, bundle, bundle, expressionNode)
+        if (evaluated.size == 0) return ""
```
carlosfelix2, line 99: Use INtelliJ's problem tab to see minor warnings about these statements (assuming you have ktlint loaded).  Change this to evaluated.isEmpty

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirPathUtils.kt
```diff
@@ -87,6 +95,40 @@ object FhirPathUtils : Logging {
...
+            throw IllegalStateException()
+        }
+        return when (val primitive = evaluated[0]) {
+            is InstantType, is DateTimeType -> convertDateTimeToHL7(primitive)
+            is DateType -> convertDateToHL7(primitive)
+            is TimeType -> convertTimeToHL7(primitive)
+            else -> defaultPathEngine.convertToString(primitive)
+        }
+    }
+    fun convertDateTimeToHL7(dateTime: Base): String {
```
carlosfelix2, line 111: Add kdoc for all these functions

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirPathUtils.kt
```diff
@@ -87,6 +95,40 @@ object FhirPathUtils : Logging {
         bundle: Bundle,
         expressionNode: ExpressionNode
     ): String {
-        return defaultPathEngine.evaluateToString(appContext, focusResource, bundle, bundle, expressionNode)
+        val evaluated = defaultPathEngine.evaluate(appContext, focusResource, bundle, bundle, expressionNode)
+        if (evaluated.size == 0) return ""
+        if (evaluated.size > 1 || !evaluated[0].isPrimitive) {
```
carlosfelix2, line 100: Chain this into an else if or a when to include the when below too.

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirPathUtils.kt
carlosfelix2, line 99: Having multiple returns in a function is not good practice.  Try to refactor this one to have one return.

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirPathUtils.kt
```diff
@@ -87,6 +95,40 @@ object FhirPathUtils : Logging {
...
+            // Should only be evaluating primitives here
+            throw IllegalStateException()
```
carlosfelix2, line 102: When throwing an exception always add a message.  On top of that, do we need to log anything or is it logged somewhere else?

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirPathUtils.kt
```diff
@@ -87,6 +95,40 @@ object FhirPathUtils : Logging {
...
+        }
+        return when (val primitive = evaluated[0]) {
```
carlosfelix2, line 104: You can simplify the code for this function into one when

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirPathUtils.kt
```diff
@@ -87,6 +95,40 @@ object FhirPathUtils : Logging {
...
+            is InstantType, is DateTimeType -> convertDateTimeToHL7(primitive)
+            is DateType -> convertDateToHL7(primitive)
+            is TimeType -> convertTimeToHL7(primitive)
+            else -> defaultPathEngine.convertToString(primitive)
+        }
+    }
+    fun convertDateTimeToHL7(dateTime: Base): String {
+        val pattern = "YYYYMMddHHmmss.SSSSxxxx"
+        val formatter: DateTimeFormatter = DateTimeFormatter.ofPattern(pattern)
```
carlosfelix2, line 113: Predefine these formatters in a companion object.  This will save time and increase performance.  If there is very little code left on each you may want to consider if it is worth including any code left on these functions to the one above.

prime-router/src/test/kotlin/fhirengine/translation/hl7/utils/FhirPathUtilsTests.kt
```diff
@@ -88,4 +102,92 @@ class FhirPathUtilsTests {
...
+        val observation = Observation()
+        observation.effective = DateTimeType("2015-04-11T12:22:01-04:00")
+        bundle.addEntry().resource = observation
+
+        // Test timestamp of java Date
+        var path = FhirPathUtils.parsePath("Bundle.timestamp")
+        var result = FhirPathUtils.evaluateString(null, bundle, bundle, path!!)
+        assertThat(result).isNotEmpty()
+        assertThat(result).isInstanceOf(String::class.java)
+        assertThat(regex.matches(result)).isTrue()
```
carlosfelix2, line 120: I am not sure the regex check is necessary as you are hard coding the pattern in another file.  This may make the maintenance of the pattern a bit more complicated as it now resides in two places.  Perhaps use the terser to make sure it accepts the format?  On top of that, you are only checking the pattern and not the contents (is the time or day in the right place?).  Maybe just test the string returned has the correct value (e.g. 20150411...)

carlosfelix2 (changes requested):

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirPathUtils.kt
PatriciaPerozoUSDS: Interesting, I've always been taught that returning early when conditions to continue have not been met is a good thing

clediggins-usds: Short circuiting logic that doesn't need to be run is a good thing, but having multiple returns is not since it makes it more difficult to debug and can result in spaghetti code. Having an if statement that short circuits/skips logic and goes directly to a single return is a good way to handle this.

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirPathUtils.kt
PatriciaPerozoUSDS: Can we talk about the compilation benefit of the companion object? I'm not sure I understand what you're asking for here.

prime-router/src/test/kotlin/fhirengine/translation/hl7/utils/FhirPathUtilsTests.kt
```diff
@@ -88,4 +102,92 @@ class FhirPathUtilsTests {
...
+
+        val observation = Observation()
+        observation.effective = DateTimeType("2015-04-11T12:22:01-04:00")
+        bundle.addEntry().resource = observation
+
+        // Test timestamp of java Date
+        var path = FhirPathUtils.parsePath("Bundle.timestamp")
+        var result = FhirPathUtils.evaluateString(null, bundle, bundle, path!!)
+        assertThat(result).isNotEmpty()
+        assertThat(result).isInstanceOf(String::class.java)
```
carlosfelix2, line 119: You do not have to test this.  This is controlled by the return type of the function.

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirPathUtils.kt
carlosfelix2: When you have the code creating a DateTimeFormatter like here, an object of that type will be created every single time this method is called.  The object will be the same and do the same every time, so you could refactor that out for reused.  You can add to the class this:
```
companion object {
   private val dateTimeFormatter = DateTimeFormatter.ofPattern("YYYYMMddHHmmss.SSSSxxxx")
}
```
this will create this static object once for the life of the JVM and you can reuse it anywhere.  This is a minor optimization, but one that can create this formatter as a constant for your class and define the pattern at a higher level in your class (which is good for these type of constants).

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirPathUtils.kt
```diff
@@ -87,6 +95,40 @@ object FhirPathUtils : Logging {
...
+        val zonedDateTime: ZonedDateTime = ZonedDateTime.parse(dateTime.dateTimeValue().valueAsString)
+
+        return zonedDateTime.format(formatter)
+    }
+
+    fun convertTimeToHL7(timeType: TimeType): String {
+        val pattern = "HHmmss.SSSS"
+        val formatter: DateTimeFormatter = DateTimeFormatter.ofPattern(pattern)
+        val localTime: LocalTime = LocalTime.parse(timeType.valueAsString)
```
carlosfelix2, line 122: Can you use the time itself, instead of converting it first to a string?  E.g., date times can generate a Date object, which then you can use to generated the formatted string.

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirPathUtils.kt
PatriciaPerozoUSDS: Gotcha - I think whats confusing is that the FHIRPathUtils is already an object and we can't create companion objects inside of other objects. Is it worth switching FHIRPathUtils to a class to allow for this optimization?

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirPathUtils.kt
PatriciaPerozoUSDS: From what I can tell, the only way to go from a FHIR DateType, TimeType, or DateTimeType object to its equivalent kotlin object is to get the string value first. I haven't found a way to directly format the FHIR objects

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirPathUtils.kt
PatriciaPerozoUSDS: I tried to refactor this to both be explicit that an empty evaluated list would return an empty string, only have one return, and better chain the if statements.

To be honest I don't like what I came up with so I'm open to more feedback.

carlosfelix2 (approved):

PR #6079, opened 2022-07-12
Experience/validation UI design

doug-s-nava:
Changes for https://app.zenhub.com/workspaces/experience-607d9d5e68b95200150fec37/issues/cdcgov/prime-reportstream/6089

This includes, vaguely, bringing the UI in line with the designs from this Figma file https://www.figma.com/file/b2Zd1xzmp4cdwFO8O7edcI/CSV-Uploader-%26-Validation?node-id=2288%3A1157

Specifically:

- [x] positioning and element sizing
- [x] text updates 
- [x] removing unused elements (timestamps etc) that were carried over from upload form but will not be used in validation

Open questions: 
- [x] alert styling / state vs uswds
- [x] upload another (where to drop), what button does
- [x] reupload functionality, similar to above
- [x] cancel functionality
- [x] links to schema documentation
- [x] text color vs uswds

## Test Steps:

1. start servers on this branch with `yarn run start:localdev` from `frontend-react` and `./gradlew run` from `prime-router`
2. visit http://localhost:3000/admin/settings and log in as necessary
3. set up your user up as a sender that is set up to send csv files, ex. `all-in-one-health-ca`
4. activate the `validation-service` feature flag via the `admin -> feature flags` navigation
5. click `validate` in the nav bar
6. _VERIFY_: page looks like `initial` design from figma
7.  upload a valid csv file
6. _VERIFY_: page looks like `success - validated - csv` design from figma (with some modifications?)
7. upload an invalid csv file
8. _VERIFY_: page looks like `errors / warnings / quality filters - csv` design from figma (with some modifications?)
9. switch to a sender that is set up for hl7 like `hcintegrations`
10. navigate back to validations
7.  upload a valid hl7 file
6. _VERIFY_: page looks like `success - validated - hl7` design from figma (with some modifications?)
7. upload an invalid csv file
8. _VERIFY_: page looks like `errors / warnings / quality filters - hl7` design from figma (with some modifications?)

### Testing
- [ ] Tested locally?
- [ ] Ran `./prime test` or `./gradlew testSmoke` against local Docker ReportStream container?
- [ ] (For Changes to /frontend-react/...) Ran `npm run lint:write`? 
- [ ] Added tests?

### Process
- [ ] Are there licensing issues with any new dependencies introduced?
- [ ] Includes a summary of what a code reviewer should test/verify?
- [ ] Updated the release notes?
- [ ] Database changes are submitted as a separate PR?
- [ ] DevOps team has been notified if PR requires ops support?

[...]

PR #6266, opened 2022-07-29
Experience/6169 file handler component tests

doug-s-nava:
Adds tests for File Handler (validator / uploader) components

A few other things that have been changed to support testing / make things a little smoother:

* in the file change handler, `event.currentTarget -> event.target`. Testing library overwrite currentTarget on any mock event you use, so in order to properly test behavior, had to make this switch.
* renamed `UseOrganizationResouce -> UseOrganizationResource`. Just a typo, but it confused me for long enough when trying to find the file that I felt the need to change it
* removed <span> tag on Cancel button. This wasn't adding anything and made it harder to target the button in testing

Test Steps:
1. _VERIFY_: tests pass in ci
2. _BONUS_: run tests from https://github.com/CDCgov/prime-reportstream/pull/6079

## Changes

### Testing
- [ ] Tested locally?
- [ ] Added tests?

### Process

## Linked Issues
- Fixes #issue

## To Be Done
- #issue

    - new endpoint A
    - new endpoint B


frontend-react/src/components/FileHandlers/FileHandlerMessaging.tsx
```diff
@@ -203,7 +214,9 @@ export const FileWarningsDisplay = ({
                 </thead>
                 <tbody>
                     {warnings.map((w, i) => {
-                        return <ErrorRow error={w} index={i} />;
+                        return (
+                            <ErrorRow error={w} index={i} key={`warning{i}`} />
```
kevinhaube, line 218: `warning{i}` -> `warning${i}`?

frontend-react/src/components/FileHandlers/FileHandlerMessaging.tsx
```diff
@@ -159,7 +161,13 @@ export const FileErrorDisplay = ({
                         </thead>
                         <tbody>
                             {errors.map((e, i) => {
-                                return <ErrorRow error={e} index={i} />;
+                                return (
+                                    <ErrorRow
+                                        error={e}
+                                        index={i}
+                                        key={`error{i}`}
```
kevinhaube, line 168: Another spot to add the missing `$` before your curly braces

frontend-react/src/components/FileHandlers/FileHandlerMessaging.tsx
```diff
@@ -31,11 +31,10 @@ export const FileSuccessDisplay = ({
     const { destinations, timestamp, reportId } = extendedMetadata;
     const destinationsDisplay =
         destinations || "There are no known recipients at this time.";
-
     return (
         <>
             <StaticAlert type={"success"} heading={heading} message={message} />
-            <div>
+            <div data-testid="sucess-display">
```
kevinhaube, line 37: We should shy away from `data-testid` -- [see this](https://youtu.be/mSYJ5Ehbo3U)

> This library encourages your applications to be more accessible and allows you to get your tests closer to using your components the way a user will, which allows your tests to give you more confidence that your application will work when a real user uses it.

kevinhaube (changes requested):
There's very few cases where `data-testid` makes sense to use over the other available options. It seems like you're able to query what you're looking for by text instead.

frontend-react/src/components/FileHandlers/FileHandlerMessaging.tsx
doug-s-nava: 🤦

frontend-react/src/components/FileHandlers/FileHandlerMessaging.tsx
doug-s-nava: looks like I wasn't targeting this anywhere, probably added it and realized later that I didn't need it but forgot to remove. Thanks for the catch. One interesting point in this conversation is that the trussworks components we use are littered with `test-id`s: https://github.com/trussworks/react-uswds/blob/a4501ddc24c12bbe3cf1279edbcf650f9dccd412/src/components/forms/FileInput/FileInput.tsx#L153

Related testing library best practices question: since it seems like in some cases (such as `<Label>`) where there is no role assignment, your only real option is to find elements by text, which is sort of awkward when what you want to assert is that a certain element has a certain text. I addressed that just targeting an element by the expected text, (as I did here https://github.com/CDCgov/prime-reportstream/pull/6266/files#diff-af78dbee2b26ac35a63755e5665a4afb277e6a04ea9fbc35c246ad1f744954a7R33) but that seems a little awkward and backwards. Am I doing that right, or is there another way?

doug-s-nava: also, thanks for the inspirational (?) video. are you familiar with [Brendan and Cory](https://www.youtube.com/watch?v=IXai3D4FqXY )?

kevinhaube: 😂 Wrong link, my bad! Meant to link to the RTL home page section where that quote was from.

YES, I'm not a fan of Trussworks for that. We had something of an internal debate at Fearless once about testing methodology -- literal (by the code) vs practical (like an end user interacting with it). Both serve well, and clearly there's a need for both. I like the Label example, with it not having a defined role. I think to better understand which makes more sense in your linked example is whether the `findByText` returns the element you need to test. If not, there might be another way to query, or it might warrant a test-id! There's times you get this weird `couldn't find element by text -- maybe it's split into different elements` error from RTL. There's workarounds for that kinda thing, but that's where I've used the test-id in the past.

kevinhaube (approved):
LGTM!

PR #6268, opened 2022-07-29
Dashboard Fixes

TomNUSDS:
Now that I have the server running in staging, the page is dealing with a ton more data. This PR fixes some bugs that become obvious with more data.

- Reduce number of days show by default from 7 -> 3. 7 Days of data is **_A LOT_** of data.
- Remove unused `Spinner` object from container page.
- Clean up some data boundary issues with selecting date range and converting from locatime to iso time.
- use `moment(d).startOf("day")` and `moment(d).endOf("day")` since we have moment available and correct implementation is non-trivial.
- Remove unused `successForDay` variable. I originally thought each day may have a different colour, but not required.
- Add notice to page about data being in LOCAL TIME.


TomNUSDS:
I'd like to get this reviewed and the fixes in before the the next deploy if possible.

frontend-react/src/components/Admin/AdminReceiverDashboard.tsx
```diff
@@ -517,7 +501,6 @@ function MainRender(props: {
         }
         perDayElements = [];
         // </editor-fold>
-        keyOffset++;
```
doug-s-nava, line 520: [tldr; consider renaming a few variables for clarity] removing this makes sense, I think (?), but trying to figure out why is making me more confused than I was before. I think the confusion comes down to what `key` and `currentEntry` represent, as that is somewhat obscured by the DataDictionary type. Each `key` is essentially a `receiverTimeSlotCheckId` or a `recieverCheckId` and each entry is a `timeSlotCheckResult`. Maybe it's just me, but I had forgotten this, and it really seemed like each key represented a receiver, so this was very confusing.

frontend-react/src/components/Admin/AdminReceiverDashboard.tsx
```diff
@@ -646,6 +631,11 @@ export function AdminReceiverDashboard() {
                 Each slot is a 2hr time slice. Colored slots had a check run.
                 Clicking on a slot shows details.
             </section>
+            <SiteAlert variant="info">
+                {
+                    "Times show are in YOUR LOCAL timezone. Be aware that receivers and servers may be in different zones."
```
doug-s-nava, line 636:
```suggestion
                    "Times shown are in YOUR LOCAL timezone. Be aware that receivers and servers may be in different zones."
```

doug-s-nava (changes requested):
Strange that as I look at this again, I'm noticing new things! Still think it's very good work, but a couple things to think about as far as possibly helpful refactors, and one main troubling thing:

I'm noticing that I can't tell the color difference on failure and success boxes or slice backgrounds (color blindness). Can you update to use the USWDS suggested colors? https://designsystem.digital.gov/utilities/color/

frontend-react/src/components/Admin/AdminReceiverDashboard.tsx
```diff
@@ -355,16 +344,13 @@ function MainRender(props: {
...
 
     // we use double cursors (move through time and through entries)
     let keyOffset = 0;
     // readability
     const perReceiverElements: JSX.Element[] = [];
     let perDayElements: JSX.Element[] = [];
     let sliceElements: JSX.Element[] = [];
-
     // loop over all receivers (each is its own row)
     while (keyOffset < keys.length) {
```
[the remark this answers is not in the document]
TomNUSDS: I actually have plans to optimize this once it's functionally correct. But a comment and issue will keep me honest.

frontend-react/src/components/Admin/AdminReceiverDashboard.tsx
TomNUSDS: Good feedback. Naming is hard and I always love feedback on better naming.

The bug was that offset inc was causing first time slot to be skipped.

image

TomNUSDS: Part of the confusion around the key, keyOffset and currentEntry: I think code could be refactored to remove the key. Originally, I was assuming I could use the key to make sorting and lookups faster, but I think it just makes everything more complex.

Can I keep the name for the PR and see if I convert the key/value into just an ordered array of values, then I can use the index instead. So `data[key[keyOffset]]` -> `data[offset]`

doug-s-nava: yeah, that's fine

TomNUSDS: https://github.com/CDCgov/prime-reportstream/issues/6288

frontend-react/src/components/Admin/AdminReceiverDashboard.tsx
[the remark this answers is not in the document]
TomNUSDS: See #6288

TomNUSDS:
@doug-s-nava I think I addressed all the issues. (Note to anyone following along ticket, we resolved the 508 color issue interactively and changed the colors to use USWDS).

doug-s-nava (approved):
✅ ✅ ✅ ✅

Issue #5700, opened 2022-06-07
Look up patient state from patient zip code


jbiskie:
**Is your feature request related to a problem? Please describe.**
The RADx spec allows senders to provide a patient zip code, but no state. Currently, we require patient state to be entered for proper message routing.

**Describe the solution you'd like**
We need to be able to populate patient state from patient zip code, in order to meet the RADx data spec.

- Change HL7 ingest schema to allow patient state to be populated from a look up of the patient zip code.
- Do not change if there is something present in the patient state.

**Additional context**
Some example messages are located in keybase: prime_dev_ops/sender_info/Intrivo

Adrian-Brewster:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Onboarding--Operations-6166edbd409257001e09f1b2/issues/CDCgov/prime-reportstream/5700?planning-poker) with ZenHub @brick-green @greene858 @jbiskie @jorg3lopez @MauriceReeves-usds @oslynn @TomNUSDS

Adrian-Brewster:
Blocker for moving Intrivo forward. @MauriceReeves-usds to look into smarty streets

Adrian-Brewster:
@MauriceReeves-usds looking into smartystreets this sprint.

TomNUSDS:
If it's _JUST_ zip->state, that's a pretty small table since the zip 3-digit prefix determines the state.

See: https://en.wikipedia.org/wiki/List_of_ZIP_Code_prefixes

The layout is also in prefix _ranges_. E.g.  California is 900-961, so the table is likely ~60 rows (AP and discontinued are in there)

sarahnw:
At this point in time we need to unblock API senders using RADX MARS so we're moving ahead without SmartyStreets. We will revisit a feature for SmartyStreets in the near future.

brick-green:
Code completed. Need to write the PR next.

brick-green:
of course things got tricky at the end. One of the smoke tests is failing but I can't quite figure out why. Gonna pair with some other engs tomorrow to get it sorted out.

brick-green:
Lookup mapper was failing and has now been fixed. Based on our discussion in slack, I had to make patient state an optional field. Writing up PR now

brick-green:
Merging to master. Here are the schemas that will be affected: 

- direct-careevolution-covid-19
- direct-hca-covid-19
- hl7-detect-covid-19
[...]

PR #6283, opened 2022-08-01
Create zipCodeToState Mapper for RADx

brick-green:
This PR adds a new mapper, ZipCodeToState, that utilizes the zip-code-data lookup table to add the patient state based on the zip code. Currently, the application will overwrite any incoming value based on the lookup results. To avoid overwriting a non-blank state value, this adds an additional field to the hl7-ingest-covid-19 schema, patient_state_zip_lookup. Additionally, this PR fixes a bug with the lookup mapper. The lookup mapper assumed that fields used in the mapper all used the same lookup table. That is an unsafe assumption. The PR fixes it so that the lookup columns are set in the parameters of the mapper removing the dependency.

Test Steps:
1. build and run all three test files through the application
     - `./prime data --input /PATH/TO/TEST/FILE.txt --input-schema HL7/hl7-ingest-covid-19-training  --output-dir /PATH/TO/OUTPUT/DIR/ --route`
2. Check output HL7
     - intrivo-test-missing-zip.txt: patient state in the output should be the same as input
     - intrivo-test-with-state.txt: patient state in the output should be same as input
     - intrivo-test-missing-state.txt: patient state in output should be different from input

Test Files
- [intrivo-test-missing-zip.txt](https://github.com/CDCgov/prime-reportstream/files/9237768/intrivo-test-missing-zip.txt)
- [intrivo-test-with-state.txt](https://github.com/CDCgov/prime-reportstream/files/9237769/intrivo-test-with-state.txt)
- [intrivo-test-missing-state.txt](https://github.com/CDCgov/prime-reportstream/files/9237770/intrivo-test-missing-state.txt)

## Changes
- added zipCodeToState mapper
- add column parameters for lookup mapper
- add unit test for zipCodeToState

### Testing
- [X] Tested locally?

## Linked Issues
- Fixes #issue

## To Be Done
- #5700


prime-router/src/main/kotlin/metadata/Mappers.kt
```diff
@@ -468,16 +469,20 @@ class LookupMapper : Mapper {
...
-                values.forEach {
-                    val indexColumn = it.element.tableColumn
-                        ?: error("Schema Error: no tableColumn for element ${it.element.name}")
-                    tableFilter.equalsIgnoreCase(indexColumn, it.value)
+                values.forEachIndexed { index, elementAndValue ->
+                    // retrieve column name to use for lookup.
+                    // Specified in sender settings lookup mapper parameters
+                    val indexColumn = args[index.times(2).plus(1)].split(":")[1]
+                    if (indexColumn.isEmpty())
+                        "Schema Error: no tableColumn for element ${elementAndValue.element.name}"
```
TomNUSDS, line 484: Think this should be `error("Schema Error: no tableColumn for element ${elementAndValue.element.name}"`?

prime-router/src/main/kotlin/metadata/Mappers.kt
```diff
@@ -932,6 +937,37 @@ class ZipCodeToCountyMapper : Mapper {
...
+/**
+ * [ZipCodeToStateMapper] runs a lookup using zip code and returns the single associated state. Null is returned if
+ * no zip code is provided, the zip code is not found, or if multiple records are found in the lookup.
+ */
+
+class ZipCodeToStateMapper : Mapper {
+    override val name = "zipCodeToState"
+
+    override fun valueNames(element: Element, args: List<String>): List<String> {
+        return args
```
TomNUSDS, line 949: My kotlin isn't super strong. What does this call do?

prime-router/src/test/kotlin/metadata/MapperTests.kt
```diff
@@ -685,6 +685,60 @@ class MapperTests {
...
     }
 
+    @Test
+    fun `test zip code to state mapper`() {
+        val mapper = ZipCodeToStateMapper()
+        val csv = """
+            zipcode,state_abbr
+            32303,FL
+            42223,TN
+            42223,KY
```
TomNUSDS, line 695: Maybe add some 5-4 zip codes (e.g. `32303-1234`) since code should handle them?

TomNUSDS (changes requested):
Looks pretty great. I did have one thing that might be a bug with needing to return error().

prime-router/metadata/schemas/NM/nm-covid-19-csv.schema
```diff
@@ -170,7 +170,7 @@ elements:
     type: TABLE
     table: LIVD-SARS-CoV-2
     tableColumn: Component
-    mapper: lookup(equipment_model_name)
+    mapper: lookup(equipment_model_name, $Column:Model)
```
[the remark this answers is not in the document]
brick-green: There was a bug in the lookup mapper that I found while working on the state mapper. This is part of the fix for that.

prime-router/metadata/schemas/HL7/hl7-ingest-covid-19-training.schema
```diff
@@ -17,4 +17,4 @@ elements:
       - code: T
         display: T
       - code: T
-        display: P
+        display: P
```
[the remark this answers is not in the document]
brick-green: I think this was incidental. I originally added a field to this file then decided it fit better elsewhere. Must have deleted the line feed that was there.

prime-router/src/main/kotlin/metadata/Mappers.kt
```diff
@@ -468,16 +469,20 @@ class LookupMapper : Mapper {
...
                 val lookupTable = element.tableRef
                     ?: error("Schema Error: could not find table ${element.table}")
                 val tableFilter = lookupTable.FilterBuilder()
-                values.forEach {
-                    val indexColumn = it.element.tableColumn
-                        ?: error("Schema Error: no tableColumn for element ${it.element.name}")
-                    tableFilter.equalsIgnoreCase(indexColumn, it.value)
+                values.forEachIndexed { index, elementAndValue ->
+                    // retrieve column name to use for lookup.
+                    // Specified in sender settings lookup mapper parameters
```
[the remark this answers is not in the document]
brick-green: I like that idea. I'll add to the comment

prime-router/src/test/kotlin/metadata/MapperTests.kt
brick-green: good calll here too. Will do

prime-router/src/main/kotlin/metadata/Mappers.kt
brick-green: That provides the list of fields used by the mapper. It has to be overwritten by each mapper. Some don't need to do any processing (like this one) but other's do(like the lookup mapper).

TomNUSDS (approved):
Approved. Thanks for the fixing this state matching issue!

Anything that needs to be communicated to others before it goes live?

PR #6291, opened 2022-08-02
no storing report file when response is 400

### LINKED ISSUE #6114: BUG: Consistent blob storage behavior on 400 errors
### User Story:
As ReportStream, I want to have consistent behavior on when we store report blobs, so I can easily track down issues.

### Description/Use Case
See spike #6080
We should not store report_file database records or files in the blob store when the result of a submission is a 400. Right now if a CSV is submitted and generates a 400 response a blob will be stored and a report_file record created, but when a HL7 generates a 400 response it will not be. Neither should be storing a blob or report_file record In addition, when duplicate detection is turned on and a duplicate is detected, do not store a report_file

* Let experience know about this change, in case it impacts them.

### Dev Notes:
* In effect, if there are an errors in the actionLog before we get to the 'save to db' portion, do not store a report_file

### Acceptance Criteria
* When a duplicate is detected, do not store report_file or anything in the blob
* When a CSV is submitted and it is an invalid file, do not store report_file or anything in the blob
* When a HL7 is submitted and it is an invalid file, do not store report_file or anything in the blob

clediggins-usds:
This PR gives consistent behavior relating to 400 responses when a bad submission is sent in. report_file records will no longer be saved if the response is a 400

Test Steps:
1. Run unit tests
2. Verify no errors

1. Submit an erroneous hl7 or csv file
2. verify no record was saved in report_file, but action_log is still saved

## Changes
- Move 'record received report' to after where errors are thrown during ingestion
- Update storing of action_log values to remove a reference to reportId if the report is not being stored
- Update tests to reflect the consistent logic

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- Fixes #6114


prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -595,6 +595,19 @@ class ActionHistory : Logging {
         insertItemLineages(itemLineages, txn)
 
         actionLogs.forEach {
+            // if we have an action log that is for a report that is not being recorded, remove the link to the report.
+            // this is a valid use case when the client submission is incorrect - a report is created in memory but
+            // is not stored in the database or blob store, but we will want the action log
+            if (it.reportId != null &&
```
TurkeyFried, line 601: good doc!

TurkeyFried (approved):
Both unit and manual tests pass. Code is very clear.

TurkeyFried:
I do wanna note that any body I submit returns a 201, which doesn't sound right but is outside the scope of this ticket. Using an empty body returns a 400 so I was able to test using that.

PR #6302, opened 2022-08-03
Adds a patient age mapper for outbound calculation of age for a patient

MauriceReeves-usds:
This PR adds a mapper for the calculation of patient age. It's currently only added to the DCIPHER schema, but we could use it elsewhere if we wanted to.

The rules around the calculation are pretty straightforward:

1. If a patient age is in the data, pass that through without any checking or cleaning. If it exists but it's garbage we still pass through garbage
2. If the patient age does not exist, but the DOB exists and the specimen collection date time exists...
  a. If the patient DOB is less than or equal to the specimen collection date, get the years between
  b. If the patient DOB is after the specimen collection date, then return null

### Test Steps:
1. Run the unit tests
2. Run a test file through the local system, outbound for DCIPHER
3. Verify that patient age is calculated for a patient with a DOB and specimen collection date time

## Changes
- Created a new mapper for the patient age element
- Moved the `toYears` function into the DateUtilities class
- Updated the DCIPHER schema to use the new mapper
- Unit tests

### Testing
- [x] Tested locally?
- [x] Added tests?


prime-router/src/main/kotlin/Report.kt
```diff
@@ -964,13 +963,6 @@ class Report : Logging {
      *  @return age - result of patient's age.
      */
     private fun getAge(patient_age: String?, patient_dob: String?, specimenCollectionDate: OffsetDateTime?): String? {
```
jimduff-usds, line 965: Why does Report.kt have functions related to patient age?  Sigh.

jimduff-usds (approved):
I didn't run it but your testing looks thorough.

PR #6315, opened 2022-08-04
Dashboard optimizations

### LINKED ISSUE #6288: Improve performance of Receiver Dashboard
- [ ] The key/value can be refactored to be `values[]` since they can be resorted and order preserved. `data[key[keyOffset]]` will become just `data[offset]`.
- [ ] Memorize the output of each row's components. The content of these don't change if the filters are changed. The filter should probably just change a className on each row to hide/show it, so the html isn't even heavily modified.
- [ ] Design issue: If the user wants to pick _a week_ from a year ago, they first much pick a starting date from a year ago, then the ending date. Problem is that picking a new starting day a year ago requests A YEARS worth of data! 
Possible Solutions:
    - Don't refresh on date just one date selection, but add a "Reload button".
    - Sanity check that no more than 2 weeks worth of data and block it with a warning.
- [ ] Limit number of days in range that's allowed. (e.g. 7 days).

TomNUSDS:
Fixes #6288 
 - Improve performance by using useMemo and not re-rendering individual rows during a filter.
 - Convert dictionary to simple array to simplify logic.
 - Limit max day range to 10 to prevent requesting too much data.
 - Date picker is "atomic" both start and end picked before update.
 - Handle if there are more than one result per-slot.
 
 ToDo:
 - deBounce key inputs in the filter field. Will use React v18's `useDeferredValue()` once we've upgraded from v17.


frontend-react/src/components/Admin/AdminReceiverDashboard.tsx
```diff
@@ -529,63 +678,165 @@ function ModalInfoRender(props: { subData: AdmConnStatusDataType[] }) {
...
-            <Grid row className={"modal-info-row"}>
-                <Grid className={"modal-info-label "}>Receiver:</Grid>
-                <Grid className={"modal-info-value"}>
-                    {dataItem.receiverName} (id: {dataItem.receiverId})
-                </Grid>
-            </Grid>
-
-            <Grid row className={"modal-info-row"}>
-                <Grid className={"modal-info-label"}>Result:</Grid>
+            {props.subData.map((dataItem) => (
```
doug-s-nava, line 683: [?] are these changes just nesting the existing content under the map? Is there a way to test this behavoir for multiple data points

frontend-react/src/components/Admin/AdminReceiverDashboard.tsx
```diff
@@ -290,79 +302,66 @@ class TimeSlots implements IterateTimeSlots {
...
+// PreRenderedRowComponents breaks out row status and org+reciever name into props
+// so parent can more quickly filter at a higher level without changing the whole DOM
+function renderAllReceiverRows(props: {
+    data: AdmConnStatusDataType[];
+    startDay: Date;
+    endDay: Date;
+    filterErrorText: string;
+    onClick: (dataItems: AdmConnStatusDataType[]) => void;
+}): JSX.Element[] {
+    console.log("renderAllReceiverRows");
```
doug-s-nava: can get rid of this

frontend-react/src/components/Admin/AdminReceiverDashboard.tsx
```diff
@@ -457,56 +500,165 @@ function MainRender(props: {
...
+            hideRowBasedOnRecv ||
+            hideRowBasedOnErrorFilter;
+
+        if (!hideRow) {
+            resultArray.push(renderedRow);
+        }
+    }
+
+    if (resultArray.length === 0) {
+        return <>No rows matching time range and any filters</>;
```
doug-s-nava, line 608: [nit] makes more sense to me for this to be a vanilla JS fn that returns data, and then you just wrap that in jsx in the render. I know that would mean refactoring along the lines I mentioned elsewhere - doing the filter before dealing with JSX, so feel free to ignore

frontend-react/src/components/Admin/AdminReceiverDashboard.tsx
```diff
@@ -457,56 +500,165 @@ function MainRender(props: {
...
+ */
+function FilterRenderedRows(props: {
+    renderedRows: JSX.Element[];
+    filterRowStatus: SuccessRate;
+    filterRowReceiver: string;
+    filterErrorText: string;
+    onClick: (dataItem: AdmConnStatusDataType[]) => void;
+}) {
+    const filterRowStatus = props.filterRowStatus.toUpperCase(); // enum is uppercased
+    const filterRowReceiver = props.filterRowReceiver.toLowerCase();
```
doug-s-nava, line 570: [nit] you have control over the values you're filtering against here, is there a reason not to just make the data look like you want it to look in the first place, rather than having to trim and change case here?

frontend-react/src/components/Admin/AdminReceiverDashboard.tsx
```diff
@@ -457,56 +500,165 @@ function MainRender(props: {
...
+        const titleClassName =
+            SUCCESS_RATE_CLASSNAME_MAP[successForRow.currentState];
+
+        // if we have error text, then check if we filtered out all the slices
+        const allSlicesFiltered = filterErrorText.length
+            ? visibleSliceCount === 0
+            : false;
+
+        // cheat confession: using `data-` props allow us to stick properties
+        // on components without type checking requiring it be a formal prop.
```
doug-s-nava, line 522: yeah but there are some nice things about using props. Why not make this a component? I have a sneaking suspicion that when things get split out into actual components you see better re-render performance, but that could just be superstition

frontend-react/src/components/Admin/AdminReceiverDashboard.tsx
```diff
@@ -371,9 +370,17 @@ function MainRender(props: {
             for (let [timeSlotStart, timeSlotEnd] of timeSlots) {
                 const successForSlice = new SuccessRateTracker();
 
-                // loop over keys that are in this range. Build aggregates for success, needed for layout
+                let filteredSlice =
```
doug-s-nava, line 372: don't really understand this variable name. seems more like `errorFilterMatchedSlice`?

frontend-react/src/components/Admin/AdminReceiverDashboard.tsx
```diff
@@ -401,31 +425,50 @@ function MainRender(props: {
                 const sliceClassName =
                     SUCCESS_RATE_CLASSNAME_MAP[successForSlice.currentState];
 
-                // we can use the currentkey in the data dict as a unique key for this cell
+                const sliceFilterClassName =
+                    MATCHING_FILTER_CLASSNAME_MAP[filteredSlice];
+
+                // it's possible to match more than on data entry per slice
```
doug-s-nava, line 432: 👍

frontend-react/src/components/Admin/AdminReceiverDashboard.tsx
```diff
@@ -401,31 +425,50 @@ function MainRender(props: {
...
                                 : (evt) => {
-                                      // get saved offset from "data-keyoffset" attribute on this element
-                                      const dataKeyOffset = parseInt(
-                                          evt.currentTarget?.dataset[
-                                              "keyoffset"
-                                          ] || "-1"
+                                      // get saved offset from "data-offset" attribute on this element
+                                      const target = evt.currentTarget;
+                                      const sliceStart = parseInt(
+                                          target?.dataset["offset"] || "-1"
```
doug-s-nava, line 454: any reason not to do?
```suggestion
                                          target?.dataset.offset || "-1"
```

frontend-react/src/components/Admin/AdminReceiverDashboard.tsx
```diff
@@ -401,31 +425,50 @@ function MainRender(props: {
...
+                                          // should never happen, but safety
+                                          sliceEnd =
+                                              sliceEnd > sliceStart
+                                                  ? sliceEnd
+                                                  : sliceStart;
+                                          const subsetData = props.data.slice(
+                                              sliceStart,
+                                              sliceEnd
+                                          );
+                                          props.onClick(subsetData);
```
doug-s-nava, line 473: a note that this handler is generally going to be used to open the modal and pass data into it. Kind of hard to trace that flow through

doug-s-nava (approved):
Looking better all the time. I don't think any of this is blocking and don't want to hold things up while I'm out.

a few notes on the ui to take or leave:

* might be nice to default the end date in the date picker to today. Seeing a blank date there looks broken to me, and makes it so you always have to enter two dates
* would be good to try and keep the modal render on the screen and scroll the modal rather than the page

frontend-react/src/components/Admin/AdminReceiverDashboard.tsx
```diff
@@ -529,63 +678,165 @@ function ModalInfoRender(props: { subData: AdmConnStatusDataType[] }) {
...
+ *  - We want start AND end picked before the expensive fetch.
+ *  - Picker fields are LARGE and take up a bunch of space.
+ */
+function DateRangePickingAtomic(props: {
+    defaultStartDate: string;
+    defaultEndDate: string;
+    onChange: (props: { startDate: string; endDate: string }) => void;
+}) {
+    const [startDate, setStartDate] = useState<string>(props.defaultStartDate);
+    const [endDate, setEndDate] = useState<string>(props.defaultEndDate); // may be null because it's optional
```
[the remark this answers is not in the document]
TomNUSDS: I'm going to fix the end date to today (as recommended above), so this will go away. The dates here are ISO strings. I should probably rename the vars so it's super obvious.

frontend-react/src/components/Admin/AdminReceiverDashboard.tsx
```diff
@@ -529,63 +678,165 @@ function ModalInfoRender(props: { subData: AdmConnStatusDataType[] }) {
...
+    }, [startDate, endDate]);
+
+    const formatDateFromString = (d: string) => {
+        if (d === "") return "now";
+        return new Date(d).toLocaleDateString();
+    };
+
+    return (
+        <>
+            <span className={""}>
```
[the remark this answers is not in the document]
TomNUSDS: Oversight. I thought I might apply some style here. If I'm going to make the Expanding DateRange Picker a separate component, then I'll probably have a className prop for this.

I'll clean it up either way.

frontend-react/src/components/Admin/AdminReceiverDashboard.tsx
```diff
@@ -529,63 +678,165 @@ function ModalInfoRender(props: { subData: AdmConnStatusDataType[] }) {
...
+                    onClick={() => {
+                        modalRef.current?.toggleModal(undefined, false);
+                        props.onChange({ startDate, endDate });
+                    }}
+                >
+                    Update
+                </Button>
+                {isDateRangeOk() ? null : (
+                    <span className={"rs-admindash-warn-font"}>
+                        Dates are too far apart (too many days)
```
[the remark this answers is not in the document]
TomNUSDS: It's right above, but repeating it here isn't horrible UI.

frontend-react/src/components/Admin/AdminReceiverDashboard.tsx
TomNUSDS: Yeah. Each "slice of day" could contain multiple results. Right now it's every two hours so it's pretty much 1:1... but if you change the slices to every 6 hrs (so there are 4x-slices-per-day), the there would be multiple results for that slice.

In that situation, the slice becomes "candy cane" striped. 

I'll add another comment at that loop.

frontend-react/src/components/Admin/AdminReceiverDashboard.tsx
```diff
@@ -290,79 +302,66 @@ class TimeSlots implements IterateTimeSlots {
...
-        const key = `${item.organizationName}|${item.receiverName}\t${item.connectionCheckStartedAt}`;
-        data[key] = item;
-        if (lastkey > key) {
-            needsSorting = true;
+const sortStatusData = (
+    dataIn: AdmConnStatusDataType[]
+): AdmConnStatusDataType[] => {
+    const strcmp = (a: string, b: string) => (a < b ? -1 : a > b ? 1 : 0);
+
+    const sortCompare = (
```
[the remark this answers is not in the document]
TomNUSDS: Yeah, I tend to tightly scope EVERYTHING even when I don't need to do so or it makes it harder to read. I'll move it out.  It will make it easier to unit test, too.

frontend-react/src/components/Admin/AdminReceiverDashboard.tsx
```diff
@@ -750,13 +984,15 @@ export function AdminReceiverDashboard() {
...
             <MainRender
                 data={data}
-                filterRowStatus={filterRowSuccessState}
                 datesRange={[
                     new Date(startDate),
                     endDate
                         ? new Date(endOfDayIso(new Date(endDate)))
                         : new Date(endOfDayIso(new Date())),
                 ]}
+                filterRowStatus={filterRowSuccessState}
```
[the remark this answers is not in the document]
TomNUSDS: Yeah, pushing the fetch down into sub-components is a great idea. We should do it in more places. I hate how all the page UI vanishes when there's a refresh.

I put it at the top because it makes testing components a bit easier because of rest-hook mocks.

I'll see how big of a refactor it is. If need be, I can always do another quick refactor to fetch deeper down. Might be nice to break up these big changes anyways.

frontend-react/src/components/Admin/AdminReceiverDashboard.tsx
TomNUSDS: Excellent point. I think I made it a component initially as I refactored just move stuff around and never noticed it was really just plain js. Now that you mentioned it, it's going to bother me, so I'll fix.

frontend-react/src/components/Admin/AdminReceiverDashboard.tsx
TomNUSDS: There was, but I don't remember why. Clearly past me should have left a commented, I'll go try and fix and see what issue I hit.

frontend-react/src/components/Admin/AdminReceiverDashboard.tsx
TomNUSDS: I didn't make them slice component because of fatigue. I also figured by sticking them in data html props, react wouldn't feel the need to get involved with them and it would improve performance.

But now I'm curious.

frontend-react/src/components/Admin/AdminReceiverDashboard.tsx
TomNUSDS: I like that name better! Fixed.

frontend-react/src/components/Admin/AdminReceiverDashboard.tsx
TomNUSDS: Just consistency with the other dataset variables all being quoted since they have "-" names. If I make these props on a Slice component, these go away.

frontend-react/src/components/Admin/AdminReceiverDashboard.tsx
TomNUSDS: I agree that confusing since it's drilled down. I wasn't sure of a good solution since the modal UI lives at the top-most level. 

In a non-react-nested-component systems, we might have some messaging pipeline and dialog would be a listener?

Any suggestions on how to improve it?

frontend-react/src/components/Admin/AdminReceiverDashboard.tsx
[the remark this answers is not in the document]
TomNUSDS: This was easier and cleaner than expected. Thanks for the suggestion. Unit tests now mock the data at the rest-hook level.

Issue #6263, opened 2022-07-29
`AuthenticationStrategy.authenticate()` returns wrong Okta organization name


kevinhaube:
> `Experience` label is present so we can list this as a blocker for another ticket on our ZenHub board

**Describe the bug**

When authenticating with okta, the AuthenticationStrategy returns the wrong parsed Okta organization name.
Input: DHxx_yyy
Output: xx_yyy
Expected: xx-yyy

**Impact**

We cannot implement the `/waters/delivery` endpoints for STLT features.

**To Reproduce**

1. Login as a user with a STLT organization group, not as a PrimeAdmin.   Per the example below, I get the failure with a user associated with DHmd_phd.
1. Use Postman (or similar) to call `/api/waters/org/{org.service}/deliveries` (ex. org.service = md-phd.elr)
2. You should get a 401 Unauthorized. Check logs, should be something about mismatched xx-yyy != xx_yyy (for this exact call it's line 77 in `ReportFileFunction.kt`
3. Use Postman (or similar) to call `/api/waters/org/{org.service}/deliveries` again, but with an underscored org name (ex. org.service = md_phd.elr)
4. You should get a 404 Not Found. Check logs, should be something about xx_yyy not being an organization. This failure is on line 73 of the same file.

**Expected behavior**

`xx-yyy` should be returned as the `organizationNameClaim` from the `AuthenticationStrategy` when given an Okta group with the format `DHxx_yyy`

**Screenshots**

This is the code block in question -- see above for line numbers.
![Screen Shot 2022-07-25 at 10.44.02 AM.png](https://images.zenhubusercontent.com/611a907ec3e013d0e33e58a7/0ccc972d-1390-408c-8573-6daf856451bf)

jimduff-usds:
I'll grab this and take it on.

jimduff-usds:
After studying the code, I decided that its likely we can simply change all the _ groups in okta to match the - used in RS Settings.
Note that simple_report has a "_" in both Okta and in RS Settings, however, at the moment the front-end code does not change DHSender settings such as DHSender_simple_report.

So I tested this.   In Staging, I have tested simply changing the group name in okta, eg, from DHmd_phd to DHmd-phd.   This appears to work.

I tested three users:
1. a prime admin
2. a user only with DHmd-phd group.
3. a user only with DHSender_simple_report group

I tested them both before and after making the change from DHmd_phd to DHmd-phd.

In each case, I tested three calls:
[...]

PR #6316, opened 2022-08-05
Enable HistoryFunction to accept both dashes and underscores

jimduff-usds:
Many of our DHxxxxx group names in Okta are wrong, because they don't match the organization names in settings.   The problem is that the group names in Okta mostly use `_` where the settings mostly use `-`.  This has a been a continuing source of confusion, messy fixes, and bugs.

I've been experimenting with simply changing all the group names in Okta to be correct, to exactly match the settings names.  (eg, change from DHmd_phd to DHmd-phd)

This lovely fix appears to work (yay!), with one exception:  the old deprecated HistoryFunctions.kt has strange code in it that depends on the group name being wrong, that is, having a `_`.  (eg, DHmd_phd)  

This PR fixes that code, so that HistoryFunctions.kt will succeed even with good correct group names (eg, DHmd-phd) and also still work with wrong names (eg DHmd_phd).    Since both old and new will both work, this will enable us to fix the erroneous names in production, without having to carefully time those Okta changes with this PR release.    

Note that HistoryFunctions.kt is deprecated code, so this fix will no longer be needed once the full Deliveries API is available. Double Yay!

## Test Steps:
1.  Run both reportstream and the frontend locally.
2. Via the front-end, login with an OktaPreview user that only has permissions to group DHmd-phd. (which used to be incorrectly called DHmd_phd, until I changed it)
3. Confirm that the download page works.
4. Change the group name back to the old bad DHmd_phd
5. Logout, clear browser cache, log back in, and confirm the download page still works.

## Changes
- replaced the auth check in HistoryFunctions.kt that depended on `_` to one that matches user's claims with requested claim, ignoring underscores and dashes.

### Testing
- [X] Tested locally?
- [X] Added tests?

### Process

## Linked Issues
- #6263

## To Be Done
- After this goes live in Prod, need to go into Okta in Prod and fix all the darn group names!

- This PR includes changes in authentication and/or authorization of existing endpoints?  **Yes**.

**Answer**:   The authentication code in HistoryFunctions.kt has not been touched.   However, the authorization check (_Does the user belong to an Okta group that matches the organization for which they are requesting access?_) has been replaced with new code.   The danger would be that I have inadvertently introduced a bug in the logic that might somehow allow improper access.    

This has been mitigated by placing the auth check in a separate function and writing thorough unit tests for that function, and by making as few other changes to the logic in HistoryFunctions.kt as possible. 

- Have you ensured logging does not contain sensitive data?  **Answer: Yes**
- Have you received any additional approvals needed for this change?  **Answer: Yes**

### DIFF
--- a/prime-router/src/main/kotlin/azure/HistoryFunctions.kt
+++ b/prime-router/src/main/kotlin/azure/HistoryFunctions.kt
@@ -455,6 +455,8 @@ open class BaseHistoryFunction : Logging {
         /**
          * @return true if the [oktaOrgs] list included a PrimeAdmin role,
          * or if [requestedOrgName] is one of the organizations in [oktaOrgs].
+         * In all other cases, the user is not authorized, and this returns false.
+         *
          * The comparison treats dashes and underscores as identical.
          *
          * This is needed, temporarily, as a step to support fixing group names in Okta to exactly match
@@ -464,9 +466,9 @@ open class BaseHistoryFunction : Logging {
             if (requestedOrgName.isBlank()) return false
             oktaOrgs.forEach {
                 when {
-                    it == null -> { } // do nothing ; skip
+                    it == null -> { } // do nothing ; skip.
                     it.contains("DHPrimeAdmins") -> return true
-                    it.replace("_", "-") == "DH${requestedOrgName.replace('_', '-')}"
+                    it.replace('_', '-') == "DH${requestedOrgName.replace('_', '-')}"
                     -> return true
                 }
             }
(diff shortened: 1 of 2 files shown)


kevinhaube (approved):
Tested locally and approved! 👍🏼 `DHmd-phd` from Okta successfully verified on the `reports` call and retrieved the list without errors. Thanks, Jim!

PatriciaPerozoUSDS (approved):
Reviewed for security - I think this change makes sense and is clearly written unit tests.

Issue #6285, opened 2022-08-02
Swap out validation endpoint


bgantick:
## Problem statement

`/file-handler/validate` needs to be wired up to the new validation endpoint so data that's passed through the validator isn't saved/processed.

## What you need to know

New validation endpoint: https://github.com/CDCgov/prime-reportstream/pull/6262

## Acceptance criteria
- [ ] Validation page submits to the new validation endpoint instead of the waters endpoint
- [ ] I'm able to upload a CSV file for validation and get back errors and warnings without saving anything to the database
- [ ] I'm able to upload an HL7 file for validation and get back errors and warnings without saving anything to the database

Issue #5713, opened 2022-06-08
Validation Service: Create validation endpoint


RakshindaAslam:
## Problem statement

We need a new API endpoint (called `validate`) for senders to validate a potential incoming CSV or HL7 file, without actually sending the file or storing the file in the database. It should return any and all errors and warnings that a normal submission would in its JSON response- all without logging or otherwise storing _this_ information, either.

## Recommended approach

Start by cloning the `/api/waters` endpoint, then factor out any codepaths that result in a save to the database.

## For possible future iterations:
- add quality filters to the response coming back
- return more detailed information about the file uploaded file (HL7 version, etc)
- store error metadata

## What you need to know
Research work: #5548

## Acceptance criteria

- [ ] Able to submit HL7 or CSV to new `validate` endpoint
- [ ] Endpoint returns warnings and errors (the same as the current endpoint), without saving data to the DB
- [ ] Unit and integration tests as need

HeatherRienks:
@sethdarragile6 & @bgantick spoke with Carlos & Cle on this ticket on 6/29/22.

bgantick:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Experience-607d9d5e68b95200150fec37/issues/CDCgov/prime-reportstream/5713?planning-poker) with ZenHub @ahay-agile6 @sethdarragile6 @kevinhaube

sethdarragile6:
### internal fun processRequest(a love story)

this is glorified pseudocode describing (most of) the flow of the `processRequest` method, with the main purpose of identifying places along that codepath that will need to be factored out of our new `/api/validate` endpoint- for the following reasons:

```
  * READS FROM DB (something that won't exist) 
  ** WRITES TO Blob Storage
 *** WRITES TO DB
**** SENDS FILES
```
BEGIN processRequest:
```
- // gathers options + sync/async
- actionHistory.trackActionSenderInfo() // begins building ActionHistory object
- validateRequest() // begins building ActionLogger, which has errors and warnings
  - // if error thrown, calls actionHistory.trackLogs() which just adds error(s) to ActionHistory.actionLogs
- validateAndMoveToProcessing() // 2 different implementations based on if sender is CovidSender or not
  - if CovidSender:
    - workflowEngine.parseCovidReport() // calls either csvSerializer or hl7Serializer
	- workflowEngine.recordReceivedReport()
[...]

PR #6350, opened 2022-08-09
Experience/6341: return Valid status if warnings but no errors are submitted

### LINKED ISSUE #6341: Validate endpoint returns 400 when only warnings are returned
**Describe the bug**
When submitting a file to the `/api/validate` endpoint that only returns warnings (but no errors), a 400 Bad Request status is returned. When submitting that _same_ file to the `/api/waters` endpoint, a 200 OK or 201 Created status is returned. Since the `validate` endpoint is supposed to effectively behave exactly like the `waters` endpoint (minus the logging, db-writing and sending), this results in a confusing inconsistency.

**Impact**
Impact is minimal for the time being, since the only consumer of the endpoint is the Validate page of the web app, which is currently hidden behind a feature flag. However, even that page currently works around the inconsistency, and is not adversely affected. The impact will be greater when/if the service becomes consumable by external consumers who come to depend on that return status.

**To Reproduce**
1. Log into RS as prime admin
2. Make sure the "validation-service" feature flag is set
3. Set your org and choose "Validate" from the "Upload & Validate" menu
4. Open up your browser's web tools and go to the Network tab
5. Upload your sample HL7 or CSV file that only throws warnings (not errors)
6. Observe that the "validate" network call returns a 400, when it should be 200

Alternatively, contrast and compare API calls to `validate` and `waters` in PostMan or curl

**Expected behavior**
A 200 http status should be returned when only warnings are returned

**Screenshots**

**Logs**
If applicable, please attach logs to help describe your problem.

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

sethdarragile6:
This PR returns Valid status (http 201) if warnings but no errors are submitted to the `/api/validate` endpoint. This makes the endpoint's behavior more consistent with the actual file submission endpoint  (`/api/waters`).

Test Steps:
1. Log into RS as prime admin
2. Make sure the "validation-service" feature flag is set
3. Set your org and choose "Validate" from the "Upload & Validate" menu
4. Open up your browser's web tools and go to the Network tab
5. Upload your sample HL7 or CSV file that only throws warnings (not errors)
6. Observe that the "validate" network call now returns a 201 in the Network tab, indicating a Valid status

### Testing
- [x] Tested locally?
- [ ] Added tests?

## Linked Issues
- Fixes #6341 
- Relates to #5713 and #6285

### DIFF
--- a/prime-router/src/main/kotlin/azure/ValidateFunction.kt
+++ b/prime-router/src/main/kotlin/azure/ValidateFunction.kt
@@ -108,26 +108,21 @@ class ValidateFunction(
                     sender.allowDuplicates
                 }
 
-                // Only process the report if we are not checking for connection or validation.
-                if (options != Options.CheckConnections && options != Options.ValidatePayload) {
-                    val receiver = ValidationReceiver(workflowEngine, actionHistory)
+                val receiver = ValidationReceiver(workflowEngine, actionHistory)
+                receiver.validateAndMoveToProcessing(
+                    sender,
+                    validatedRequest.content,
+                    validatedRequest.defaults,
+                    options,
+                    validatedRequest.routeTo,
+                    false,
+                    allowDuplicates,
+                    rawBody,
+                    payloadName
+                )
 
-                    // send report on its way, either via the COVID pipeline or the full ELR pipeline
-                    receiver.validateAndMoveToProcessing(
-                        sender,
-                        validatedRequest.content,
-                        validatedRequest.defaults,
-                        options,
-                        validatedRequest.routeTo,
-                        false,
-                        allowDuplicates,
-                        rawBody,
-                        payloadName
-                    )
-
-                    // return OK status, report validation was successful
-                    HttpStatus.OK
-                } else HttpStatus.OK
+                // return OK status, report validation was successful
+                HttpStatus.OK
             } catch (e: ActionError) {
                 actionHistory.trackLogs(e.details)
                 HttpStatus.BAD_REQUEST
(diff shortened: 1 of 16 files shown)


ahay-agile6 (approved):

prime-router/src/main/kotlin/azure/ValidateFunction.kt
```diff
@@ -125,8 +125,8 @@ class ValidateFunction(
                         payloadName
                     )
 
-                    // return CREATED status, report submission was successful
-                    HttpStatus.CREATED
+                    // return OK status, report validation was successful
+                    HttpStatus.OK
                 } else HttpStatus.OK
```
clediggins-usds, line 130: Why is there an if/else here? both result in HttpStatus.OK. Is the if/else still even valid?  Do the options CheckConnections and ValidatePayload even make sense?  I'd say no, since this *is* the validation endpoint.

Probably just get rid of the if/else entirely. Run the validateAndMoveToProcessing for validation, and then return HttpStatus.OK if it isn't kicked out with an exception by the validateAndMoveToProcessing function

clediggins-usds (changes requested):

prime-router/src/main/kotlin/azure/ValidateFunction.kt
sethdarragile6: @clediggins-usds good call, I've removed it.

clediggins-usds (approved):

PR #6372, opened 2022-08-11
Jmp/6337/consolidate sonar on frontend

jeremy-page:
Before this change we had a workflow that triggered every PR which would built and test the front end code and reporting the results to Sonarcloud. 
After merging this change we will no longer have a specific workflow for reporting test results, instead we will use the results already created from the build-frontend action which means we will be building the front end one less time in the CI\CD pipeline.


.github/actions/build-frontend/action.yml
```diff
@@ -46,10 +46,24 @@ runs:
...
+          -Dsonar.coverage.exclusions=frontend-react/**/__mocks__/**,frontend-react/**/mocks/**,frontend-react/**/*.test.*
+          -Dsonar.sources=frontend-react/src
+          -Dsonar.projectKey=CDCgov_prime-data-hub
+          -Dsonar.organization=cdcgov
+
     - name: Tar frontend files
       shell: bash
       working-directory: frontend-react
-      run: tar -czvf static-website-react.tar.gz build
+      run: tar -czf static-website-react.tar.gz build
```
snesm, line 66: I know the 'v' makes the output really long, but it could have debug or forensic value.

doug-s-nava: I haven't seen any errors on the tar step before so probably fine to drop the v

doug-s-nava: I haven't seen any errors on the tar step before so probably fine to drop the v

.github/actions/build-frontend/action.yml
```diff
@@ -46,10 +46,24 @@ runs:
...
 
+    - name: SonarCloud Scan
+      uses: sonarsource/sonarcloud-github-action@de2e56b42aa84d0b1c5b622644ac17e505c9a049
+      env:
+        GITHUB_TOKEN: ${{ inputs.GITHUB_TOKEN }}
+        SONAR_TOKEN: ${{ inputs.SONAR_TOKEN }}
+      with:
+        args: >
+          -Dsonar.cpd.exclusions=frontend-react/**/*.test.*
+          -Dsonar.javascript.lcov.reportPaths=frontend-react/coverage/lcov.info
```
doug-s-nava, line 57: my only concern here is that for reasons that I am not sure of, we seem to be not runnign tests when building for prod. In that case there won't be a coverage report for sonarcloud to read here. We could either:
* run tests in prod, which seems like it makes sense to me
* not run sonarCloud in prod, which if there is logic around not running tests in prod, it may make sense not to run sonarCloud either

.github/actions/build-frontend/action.yml
jeremy-page: It's only to archive the results - it will still fail if there is a problem (and then we can debug it)

.github/actions/build-frontend/action.yml
jeremy-page: Ah, you are right we have a conditional in the action. I will add the yarn test:ci to the production step. Good catch!

jeremy-page: Hmm, this change will mean it reports both on master & production though - so the net will be that it saves one build (after creating the deployment branch) and will stop running for feature branches.

jeremy-page:
Sorry forgot to push my last changes.

lucasdze (approved):

Issue #6386, opened 2022-08-12
Receiver Dashboard url doesn't include date range


TomNUSDS:
## Problem Statement
Navigating back and forth with the Dashboard loses its state (date range and search filters). Solution is to put state into router's path.

1. Go to Reciever Status Dashboard /admin/send-dash
2. Change filter settings at top of dashboard (e.g. enter Receiver Name 'az-phd' or Success Type: "Failed") This reduces the table list to a smaller set of interesting results.
3. Click on Org name or Receiver Setting name in the right column to navigate to the Settings for that item.
4. Click back to return to the dashboard page.
Result: the filter changes in step 2 are lost and the page is full of uninteresting results. The filters must be re-entered to find items again.

Expected: Nice if navigating back would preserve the filter settings. In react this can be done by putting the fields into the url path.

NOTE: If more information/clarity needed, contact @stephenkao 

## Acceptance Criteria
- [ ] Navigating back from Receiver Status Dashboard does not reset the date range and search filters
- [ ] Navigating directly to the page with the parameters will automatically populate the search filters

RakshindaAslam:
@TomNUSDS - can you give more context on this? Which page is this referring to?

TomNUSDS:
Fair. I updated the issue to include steps.

kevinhaube:
@CnoteNava For Sprint 58 consideration!

kevinhaube:
Updating this requires more work than we thought. All functionality within the Receiver Status Dashboard is proprietary, as are its UX and engineering patterns, so it not only subverts the expectation of reliance on the `FilterManager` hook, but further complicates the implementation of standards by using a modal-based UX pattern instead of on-page date-pickers. State-to-url-to-state logic gets very complex when attempting to use url query params with its current design.

This prompts a larger discussion, though. The admin tools were created mostly in the O&O silo rather than as a part of a concerted, research-supported collaborative effort between O&O and Experience. All Admin features and their pages should be assessed for their use of shared resources and engineering patterns, the outcome being tickets for necessary refactor work. 

TL;DR
>The main reasons to support the refactor of those tools are:
> - The code (mostly) does not use shared resources so it’s essentially its own monolith of functionality inside our app
> - Code does not follow any established patterns (i.e. re-spinning its own UX/UI for filtering by date) so it’s harder to dissect its monolithic functionality and implement established utilities like a FilterManager

chris-kuryak:
This should be broken out into a bigger architecture fix. 
[...]

PR #6387, opened 2022-08-12
Add deep link to settings for org

TomNUSDS:
Simply add `<Link` to Receiver Settings to the Dashboard to make navigating easier. 

image

Note: date range in Dashboard is lost when the navigation happens. Created https://github.com/CDCgov/prime-reportstream/issues/6386 to track


frontend-react/src/components/Admin/AdminReceiverDashboard.tsx
```diff
@@ -536,9 +539,9 @@ function renderAllReceiverRows(props: {
             >
                 <Grid className={`title-column ${titleClassName}`}>
                     <div className={"title-text"}>
-                        {orgName}
+                        <Link to={linkOrgSettings}>{orgName}</Link>
```
kevinhaube, line 542: Looks like the use of `Link` is causing some build error...perhaps a test needs to be wrapped in `renderWithRouter`?

TomNUSDS: Thanks. Somehow I didn't spot that issue. Fixed it and the checks are rerunning.

Issue #5909, opened 2022-06-23
Valueset Editor: Refactor lookup table backend structure


sethdarragile6:
## Problem statement

For MVP of Valueset Editor, it was decided to store all valueset detail data in one large table called `sender_automation_value_set_row`. At the time it seemed that, in light of some of the limitations of the lookuptable api (links below) we would need to make an embarrassing number of calls to the API if we had to store each valueset in their own table. While those limitations still hold, and there are a series of stories now written to address some of that, it became apparent while doing the MVP work that storing each valueset in their own table would _not_ lead to an increased number of calls in and of itself, since each set of valueset data would only be pulled once per page load.

Fortunately, our MVP frontend code was built generically enough that moving to this model should _not_ be that difficult. All valuesets share now and will continue to share the same basic structure (same fields) and still can relate to their parent table `sender_automation_value_set` by their name fields. Moving to this model will also solve our awkward "Last Update" issue [link to story here TBD], since we'll now be able to pull that information directly from the API rather than having to store it as custom fields on the related `sender_automation_value_set` record. And besides just being better aligned with industry-standard data normalization practices, it will also lead to less redundant data being stored in the database _aggregate_. In other words, instead of having to re-insert the 1600+ row `sender_automation_value_set_row` table into the database with each little update to _any_ valueset, we can now re-insert generally much smaller tables into the database with each little update- but _only_ to the table corresponding to the valueset we're editing (the `patient_preferred_language` will still be a problem, there). 

Basic workflow:

Split the existing sender_automation_value_set_row CSV into 11 different CSVs corresponding to each valueset. We can modify the ConvertValuesetsYamlToCSV.kt CLI tool to do this easily from the source document (sender-automation.valuesets). Then, load them per usual via the lookuptable CLI. 

IF, however, Valuesets UI has already gone live and is in use, a _new_ migration tool will have to be created. Instead of the `sender_automation_value_set` file acting as the source of truth, it will instead be whatever the latest version of the `sender_automation_value_set_row` table. You can get the latest content of that table in CSV form via the Lookuptable CLI, for example:

`$ ./prime login --env prod`
`$ ./prime lookuptables list --env prod  # to determine the latest version, which plugs into the next command`
`$ ./prime lookuptables get --name sender_automation_value_set_row --env prod --output-file output.csv --version 1`

## What you need to know

[...]

PR #6389, opened 2022-08-12
Experience/5909: valueset backend refactor

sethdarragile6:
This PR refactors the Valuesets backend from one large table containing all value sets (`sender_automation_value_set_row`) to a table per value set (i.e. `gender`, `test_results`, `yesno`, etc). Minor adjustments to the associated mapper, `LookupSenderAutomationValuesetsMapper` and tests were also made.

Test Steps:
1. Restart local containers locally, so that each value set CSV is automatically loaded into the database
2. In a Sender's schema file such as metadata/schemas/non-standard/all-in-one-health-ca-covid-19.schema, update an element that is using a valueSet: sender-automation/<type> by changing the type to TABLE, add the table property and set it to <type>, add the tableColumn property and set it to code, and add the mapper property and set it to lookupSenderAutomationValuesets([element name],<type>), and removing the valueSet property
3. Here is an example of an all-in-one-health-ca-covid-19.schema, element being converted to a sender automation value set mapper element:
Starting at line 86
Before:
```
-     name: patient_gender
      cardinality: ONE
      type: CODE
      valueSet: sender-automation/gender
      documentation: Translate multiple inbound Gender values to RS values
      csvFields: [{ name: Patient Sex, format: $display}]
      default: U
```
After:
```
-     name: patient_gender
      cardinality: ONE
      type: TABLE
      table: sender_automation_value_set_row
      tableColumn: code
      documentation: Translate multiple inbound Gender values to RS values
      csvFields: [{ name: Patient Sex, format: $display}]
      default: U
      mapperOverridesValue: true
      mapper: lookupSenderAutomationValuesetsMapper(patient_gender, gender)
```

### Testing
- [x] Tested locally?
- [ ] Added tests?

## Linked Issues
- Relates to #5909


TurkeyFried:
Just linking the ticket here for easy access: https://github.com/CDCgov/prime-reportstream/issues/5909

prime-router/src/main/kotlin/metadata/Mappers.kt
```diff
@@ -634,10 +635,10 @@ class LookupSenderAutomationValuesets : Mapper {
         }
 
         val tableFilter = lookupTable.FilterBuilder()
-        val valueSetName = args[0]
+        val valueSetName = args[1]
```
TurkeyFried, line 638: Would be good to leave a comment explaining why we don't start with 0 here.

TurkeyFried (approved):
I was able to build and run tests both with and without the mapping config changes. Just left a comment regarding one of the code changes

PR #6397, opened 2022-08-15
Experience/refactor session storage

doug-s-nava:
This PR solves two main issues:
* when spoofing an organization as an admin user, your spoofed organization settings are lost when refreshing the page
* the `organization` request header which is required by some ReportStream API endpoints (not sure which ones) is sometimes out of sync with a users actual organization

And also solves these less apparent issues in the background:
* removes some duplicative functionality around setting membership state on login
* removes some redundant dependencies on session storage, and multiple state storage locations.
* Improves centrality of SessionContext data, and simplifies interface for interacting with it
* establishes flow whereby the SessionContext based state is the source of truth for the application and session storage is used as a way to persist this state for use on application refreshes

There is still a good amount of work to be done on fully accomplishing some of these background issues, and there will be at least 2 follow up PRs to this work to further improve how this code works and make it cleaner.

### Fixes

1. start servers on master
2. visit http://localhost:3000/admin/settings logging in as necessary
3. click `set` somewhere to spoof an organization
4. refresh the page
5. _VERIFY_: based on the display in the top right of the screen, your spoofed org preference has not persisted
6. repeat step 3 and choose an org with some submissions in your local env
7. open your dev tools network tab
8. click `submissions`
8. _VERIFY_: the call out to the waters endpoint (something like http://localhost:3000/api/waters/org/all-in-one-health-ca/submissions?pagesize=11&cursor=3000-01-01T00:00:00.000Z&endcursor=2000-01-01T00:00:00.000Z&sort=DESC&showfailed=false) contains an `organization` header of `ignore`
9. repeat steps 1 - 4 on this branch
10. _VERIFY_: your spoofed organization is maintained on refresh
11. repeat steps 6 - 9 on this branch
8. _VERIFY_: the call out to the waters endpoint (something like http://localhost:3000/api/waters/org/all-in-one-health-ca/submissions?pagesize=11&cursor=3000-01-01T00:00:00.000Z&endcursor=2000-01-01T00:00:00.000Z&sort=DESC&showfailed=false) contains an `organization` header matching your spoofed organization

### Regression

Please dogfood the app to your heart's content to make sure that nothing has been broken

## Screenshots

Call with incorrect organization header 
Screen Shot 2022-08-15 at 5 16 52 PM

### Testing
- [ ] Tested locally? (needs more dogfooding!)
- [x] Added tests?

## To Be Done

- [ ] delete some unused stuff
- [ ] refactor session context value shape
- [ ] remove dependency on getStoredOrg and getStoredOktaToken ?? (this may need to wait until the fetch refactor)

- Does this PR include changes in authentication and/or authorization of existing endpoints? **YES**

- Have you received any additional approvals needed for this change? **NO**

### Threat Mitigation

If there is a bug present in this refactor it is possible that an access token or organization membership state could be maintained after a user has lost authentication / authorization, thus opening us up to either **I** or **E**. 

Another possible but could result in a case where a user somehow loses authentication / authorization when they should not, resulting in **D**.

Sufficient review and dogfooding should be done to ensure that there are no such bugs present in this work.


frontend-react/src/contexts/SessionStorageTools.ts
```diff
@@ -2,47 +2,79 @@
...
+    }
+};
+
+// temporary solution.
+// TODO: replace all occurrances of this with reads from SessionContext
 export function getStoredOktaToken(): string | undefined {
-    return (
-        sessionStorage.getItem(GLOBAL_STORAGE_KEYS.OKTA_ACCESS_TOKEN) ||
-        undefined
+    const tokenJsonString = sessionStorage.getItem(
```
doug-s-nava, line 35: rather than manage storing the token ourselves, rely on okta's storage of the token, and just process it to our specifications as necessary

frontend-react/src/contexts/SessionStorageTools.ts
```diff
@@ -2,47 +2,79 @@
  * right, but by no means is it a context. Feel free to move
  * it if you can think of a better spot. */
 
-import { AxiosRequestHeaders } from "axios";
-
-import { updateApiSessions } from "../network/Apis";
-
-const headersFromStoredSession = (): AxiosRequestHeaders => ({
```
doug-s-nava, line 9: this has moved to Api.ts

frontend-react/src/contexts/SessionContext.tsx
```diff
@@ -36,15 +31,13 @@ const SessionProvider = ({
...
-    const store = useSessionStorage();
-    const memberships = useOktaMemberships(authState?.accessToken);
 
+    const membershipController = useOktaMemberships(authState);
     return (
         <SessionContext.Provider
             value={{
                 oktaToken: authState?.accessToken,
-                memberships,
-                store: store,
```
doug-s-nava, line 47: UseSessionStorage has been deprecated in this change and will be removed, but I didn't do that in this PR for the sake of keeping the diff down. The next PR will be a cleanup PR that will include that removal

frontend-react/src/hooks/UseOktaMemberships.ts
```diff
@@ -119,49 +131,111 @@ export const membershipsFromToken = (token: AccessToken): MembershipState => {
...
         case MembershipActionType.ADMIN_OVERRIDE:
-            return {
+            const newState = {
                 ...state,
                 active: {
                     ...state.active,
                     ...(payload as MembershipSettings),
                 },
             };
+            storeOrganizationOverride(JSON.stringify(payload));
```
doug-s-nava, line 171: key pt 1 for fixing the refresh bug

frontend-react/src/hooks/UseOktaMemberships.ts
```diff
@@ -119,49 +131,111 @@ export const membershipsFromToken = (token: AccessToken): MembershipState => {
...
-                active:
-                    state.memberships?.get(payload as string) || state.active,
-            };
-        case MembershipActionType.UPDATE:
-            return membershipsFromToken(payload as AccessToken);
+        case MembershipActionType.SET_MEMBERSHIPS_FROM_TOKEN:
+            const parsedMemberships = membershipsFromToken(
+                payload as AccessToken
+            );
+            return calculateMembershipsWithOverride(
```
doug-s-nava, line 160: key pt 2 for fixing the refresh bug

frontend-react/src/utils/OrganizationUtils.ts
```diff
@@ -90,7 +88,7 @@ const getRSOrgs = (
     return filterGroups().map((group) => parseOrgName(group));
 };
 
-function parseOrgs(orgs: Array<string>): Array<Partial<SessionStore>> {
+function parseOrgs(orgs: Array<string>): any[] {
```
kevinhaube, line 91: Is there any specificity that can be brought to this type definition? Does it even need to exist now that we have `useOktaMemberships`?

frontend-react/src/contexts/SessionStorageTools.ts
```diff
@@ -2,47 +2,79 @@
...
+const fetchJsonFromSession = (storageKey: string) => {
+    const storedString = sessionStorage.getItem(storageKey);
+    if (!storedString) {
+        return {};
+    }
+    try {
+        const sessionJson = JSON.parse(storedString);
+        return sessionJson;
+    } catch {
+        return {};
```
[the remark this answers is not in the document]
doug-s-nava: I think the question here is whether errors in parsing session storage in these cases are actually useful errors for a user. In my mind, the result of an error for the user would be the same as the result of empty storage, which is that nothing gets read from storage, so state is just reinitialized.

Another thing to consider is that barring a case where a user, plugin, or something external is messing with the session storage, what's getting in here is what we're putting in here, so if we read out something that throws an error, that's on us for storing something that's throwing an error. So, there is definitely a conversation to be had about, in cases where we are storing JSON strings, that we validate the strings before we store them. Do you think we should build that in here?

Or maybe I'm missing a case where this error is something that should be exposed to users. Can you think of an example?

frontend-react/src/hooks/UseOktaMemberships.ts
```diff
@@ -96,10 +103,15 @@ export const makeMembershipMapFromToken = (
 };
 
 const defaultState: MembershipState = {
+    // note that active will be set to {} rather than undefined in most real world cases on initialization
```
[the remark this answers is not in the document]
doug-s-nava: I went back and forth on this a bit but where I ended up was:

* we're storing and reading a json string from session storage for the organization override, so returning a non-json value there would be breaking the expectation that we're getting JSON
* if we do return a falsey value from `getOrganizationOverride` we would need to throw some logic into `calculateMembershipsWithOverride` to account for that possibility, which complicates the function a bit
* there's no inherent value in having an undefined value for `active` at this point since we are, in all the cases in the code I've found, referring to a property of `active`, and thus sort of expecting that it's an object already

Definitely willing to put in that slightly more complex logic to end up with undefined if you think it's worth it.

frontend-react/src/hooks/UseOktaMemberships.ts
doug-s-nava: I'm not sure I follow exactly how the teardown would work, would love to talk more about that. 

But in any case the idea is that for non-admin users, their state is maintained on refresh through the normal process: on refresh, the okta hook will recall their user state from token storage -> the useOktaMemberships hook will pick up that state from the token -> reducer updates the app state. That's the idea - or have you seen this not working properly?

frontend-react/src/contexts/SessionStorageTools.ts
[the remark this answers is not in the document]
kevinhaube: I may be thinking of this in a different context (this was the first file in the diff) but my main thought was debug-ability of our inner functions. I haven't gotten to the other comment, but this is in the same vein as patterns like empty objects vs undefined, and why we use what when, and where.

frontend-react/src/hooks/UseOktaMemberships.ts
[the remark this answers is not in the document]
kevinhaube: I guess it's a philosophical question of whether this backs us into difficulties with React/TS. Falsey values within rigidly defined interfaces are a big help in TypeScript React when data is constantly updating, and this seems like it'd create an _extra_ layer of complexity.

frontend-react/src/hooks/UseOktaMemberships.ts
kevinhaube:
```typescript
useEffect(() => {
    // storage empty? Do nothing
    // storage not empty? Read in and remove storage item so it cannot be altered
    return () => { //stash in storage here }
}, [])
```

Something like this pattern. That `return` runs on teardown of a component, so this can stash it on teardown, and read it in on first render, then remove it from the storage so it won't be altered outside of our process.

kevinhaube: And as far as consistency, while we aren't using persisted values for users yet, it's partially _because_ of the limits of our provider prior to this PR. It's one of those "why back ourselves into a corner if we don't have to" philosophy questions again.

frontend-react/src/hooks/UseOktaMemberships.ts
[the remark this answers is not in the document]
doug-s-nava: Yeah, I think that's a fair point. I will refactor this to set `active` as undefined rather than `{}` if there is no override or active membership in state.

frontend-react/src/hooks/UseOktaMemberships.ts
doug-s-nava: [for posterity after offline discussion] I will make an attempt at a solution that sets session values on teardown only

doug-s-nava: so, turns out that a teardown based solution where session storage is cleared on each startup would not work without a full refactor of the many places in the app that are still reliant on `getStoredOrg` returning a valid current org (with override) during the app lifetime. This is something that we should definitely revisit after react-query has been introduced, and all of that parts of the app dependent on getStoredOrg are factored out

frontend-react/src/contexts/SessionStorageTools.ts
```diff
@@ -2,47 +2,79 @@
...
     );
+    if (!tokenJsonString) {
+        return "";
+    }
+    try {
+        const tokenJson = JSON.parse(tokenJsonString);
+        return tokenJson?.accessToken?.accessToken;
+    } catch (e) {
+        console.error("Error retrieving access token", e);
+        return "";
```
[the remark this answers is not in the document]
doug-s-nava: looking at this okta token piece separately from the issue of reading our own session values, it is a slightly different case. We may want to give a user good information about this error. However, the real error that the user cares about in the case where we can't retrieve an okta token is that they can't log in / have been logged out / can't access what they want to, and they will probably want to know why. At this point we would only know that we weren't able to parse the token that okta put in session storage. Most likely if there is an error here, the app will error out somewhere farther downstream with a message that might be more meaningful to the user. I think this deserves some thought when we have time to dig into our okta behavior (are we correctly handling automatic token refreshes, etc.), but not sure there's much to be done at this point

kevinhaube (approved):
Looks good! This was a fun one 🎉

PR #6411, opened 2022-08-17
CRON sanity checks settings before checking

### LINKED ISSUE #6367: receiver_connection_check_results table full of data that will never succeed
**Background**
The receiver_connection_check_results table is used by `CheckFunction.scheduleRun()` and checks that a receiver is working by connecting to it on a periodic schedule (every two hours currently). It then adds a row to the `receiver_connection_check_results` table.

This table is used by the PrimeAdmin Dashboard page [/adm/send-dash](https://staging.reportstream.cdc.gov/admin/send-dash).

**Issue**
There are lots of receivers that NEVER succeed (about 50%). These are never expected to succeed for various reasons. 

This results in:
 - Worse signal:noise when looking at results.
 - Taking up more significantly more storage since the failures including a long java stack trace.
 - Slows down the Dashboard page because querying for a timespan required more data and a much bigger DOM.
 - Limits max number of days displayed in the Dashboard page. The query for a span of days gets bigger than 9000 rows and gets clipped by api.
 
 **Thoughts**
 - Solution should work in staging and production.

TomNUSDS:
Fixes #6367 Reduce the number of Receiver Checks that always fail in the Receiver Check Dashboard. These come from the CRON job that runs every 2hrs.

Checks:
- the customerStatus is Active.
- there is a transport defined in the Setting.
- the transport is of a type that the checker can handle sftp right now

Todo:
- On Staging, no real checks should run, but simulated checks could fill in the database so the Dashboard page has something to display.

How to test:
 1. Before getting this change, change the frequency of the CRON from once at 8am to every hour in `prime-router/local.settings.json` e.g. (`  "REMOTE_CONNECTION_CHECK_SCHEDULE": "0 0 * * * *"`)
 2. Let run for a few hours to see ALL the settings are running using the Dashboard page.
 3. Get and run these changes. 
 Observe: only Receiver Settings that are run are Active and using SFTP transport.
 
image
Only some checks that were running (left most slices) are still running after change applied (right most slices).


TomNUSDS:
@MauriceReeves-usds Bump for review if you get time.

MauriceReeves-usds (approved):
Looks good to me. We should probably expand out checks for active/inactive, but this is a good start.

TomNUSDS:
@MauriceReeves-usds Ah, do you mean checking the ORGANIZATION level settings? I can create a ticket for it so we don't forget.

MauriceReeves-usds:
@TomNUSDS yes we should probably check that as well.

PR #6415, opened 2022-08-17
Clean-up and separation of IA concerns

kevinhaube:
The diff on this PR looks terrifying **BUT** it isn't so bad. Here's a breakdown of the changes for this cleanup:

## Changes
- Moves the content-related variables out of files in the `pages` directory (i.e. `ResourceCardGrid.tsx`) and into the `content` directory (i.e. `content/resources/index.tsx`).  They import and bundle content into `ContentDirectory[]` and provide the `DirectoryTools` as a secondary export. These are consumed by the components in the `pages` directory.
- Extracts the `IACardGridTemplate` from common uses between Support and Resources into a single reusable component
- Extracts `IAMetaAndRouter` from common uses between Support and Resources
- Removes `faq.md`, `contact.md`, and `ReleaseNotes.tsx` as all three were unused -- FAQ and Contact still use legacy `tsx` files and Release Notes uses the newer `.md` file
- Restructured legacy IA components into legacy folders and provided indexes for cleaner imports. **They are ready to be ripped out as soon as Markdown replaces each.**
- As you add new MD files, you'll need to add them to their respective `markdown-adapters.tsx` file under `pages/[page name]` (see `pages/resources -> markdown-adapters.tsx` for a sample)
- Removes old snapshot from tests

## What this means for Content folks
IF something needs to be altered in an `ElementDirectory`'s source file (i.e. a `tsx` file), they can be found in `pages -> [page name] -> legacy-page-content`.

Screen Shot 2022-08-17 at 3 55 15 PM

IF something needs to be altered in a `MarkdownDirectory`'s source file, they within the `content -> [page name]` folders.

Screen Shot 2022-08-17 at 3 08 59 PM

Ideally, nothing _new_ is added as an ElementDirectory as that's reverting back to reliance on markup like HTML/JSX. As things are migrated off of markup onto the new markdown system, we can remove the files from `legacy-page-content` folders. 

Lastly, **NO changes should be made to the page files within `/pages` unless it is to swap the template of the page**. Layout changes are applied via `IA...`-prefixed template components, like `<IACardGridTemplate>`, and all except legacy content can be changed via the `content` folder.

### Testing
- [X] Tested locally?
- [ ] Added tests?

### Process

## Linked Issues
- Fixes #issue

## To Be Done
- #issue

    - new endpoint A
    - new endpoint B


frontend-react/src/content/product/index.tsx
```diff
@@ -0,0 +1,48 @@
+/* Data that drives breadcrumb creation and slug appending */
```
doug-s-nava, line 1: [nit] if the idea with these index files is that they are creating the scaffolding that the .tsx files in the /pages directory will use to render the pages, but not actually containing any render logic themselves, then for the sake of clarity (and nothing else because functionally who cares), it seems like these should be .ts files

doug-s-nava (commented):
only one small nit, but I think I'm missing how this change allows for creating sections as described in the card. is this just pre-req work and the work of allowing for sections within the pages is coming next?

frontend-react/src/content/product/index.tsx
kevinhaube: This one could probably by a ts file for now. Take a look in the one for Resources, though, where it uses a wrapper function that takes a component in JSX form. 🤔 I'll see if I can remove the need for that function there and maybe use it somewhere in the template instead but that was the original need for the tsx extension. 

If I can't, I'll have to leave it as is but that does kinda blend into the layout logic imo

kevinhaube:
@doug-s-nava Yes, sorry, this is only prelim work! When extracting the template I figured I'd clean up from our hackathon on that stuff so I had a clean workbench.

frontend-react/src/content/product/index.tsx
kevinhaube: I ended up altering the function taking JSX to take either `() => JSX.Element` OR `JSX.Element` so either can be passed in and rendered with crumbs. That removed the need for any JSX syntax in these files. I couldn't, however, make the wrapper a part of the rendering process instead of this instantiation process.

frontend-react/src/components/Content/PageGenerationTools.tsx
```diff
@@ -2,13 +2,14 @@
...
+type ElementFunction = () => JSX.Element;
 export const contentContainer = (
-    content: JSX.Element,
+    content: JSX.Element | ElementFunction,
     crumbs: CrumbConfig[]
 ) => {
     const wrappedElement = (
         <div className="grid-container rs-documentation usa-prose">
-            {content}
+            {content instanceof Function ? content() : content}
```
doug-s-nava, line 12: is the hope that once we've moved off of the legacy system that we can remove this complexity? if so, can we leave a comment to that effect?

doug-s-nava (approved):
one more small nit, but looks good to me

✅ ✅ ✅ ✅

frontend-react/src/components/Content/PageGenerationTools.tsx
kevinhaube: Hmm... 🤔 I actually may have set us up to _only_ need the `content()` bit anyway. Fancy that!

PR #6429, opened 2022-08-18
bugfixes from 5745

clediggins-usds:
This PR fixes an issue in primeCLI that prevents processing of hl7 files with 5 delimiter characters, and updates messageSender config to fix a bug there.

Test Steps:
1. run fhirdata comment on a source hl7 with 5 delimiters, see it doesn't break

## Changes
- replace 5 delimiters with 4 on hl7
- update messageSender hl7 source


PatriciaPerozoUSDS (approved):
Tested locally - thank you for pulling out this bug fix!

PR #6452, opened 2022-08-22
Fix to issue that submission details is returning tons of repeated data

### LINKED ISSUE #6454: Bug:   Submissions API returns tons of repeated destinations over and over
**Describe the bug**
@kevinhaube  writes:

I’m finding some awkward behavior with our submissions detail pages that use the destination attribute from the response given by /waters/report/{reportId}/history — It seems like maybe something is going wrong in the pipeline somewhere that creates well over 100 destinations each time, often repeats of the same 3-5 actual destinations, nearly or completely indistinguishable from one another in all cases.
To recreate:
Log in to prod and mock a sender (I used elitemedlabs-co for this)
Navigate to Submissions
If you’re following along, the report I’m looking at is a07be4e1-035d-47cb-b7cf-97c2ea6c5c6c — otherwise, find a submission with a manageable amount of records and click the ID to view details
When the details page loads, your network response should have the destinations array collapsed. Expanding it will reveal all the data that’s rendered to that page. This is where you’ll find the duplicates.

**Impact**
The Submissions API returns terrible results.

**To Reproduce**
See above.  

**Expected behavior**
Each destination is represted once only!

jimduff-usds:
This PR fixes an issue that the Submission Details query was returning tons of extraneous repeats of reports.   See ticket #6454.

#### How to test localhost:

Generate a bunch of data going to several receivers, into a file called ./junk/mix.csv:

`./prime data --input-fake 30 --input-schema primedatainput/pdi-covid-19 --output ./junk/mix.csv --target-states MD,HI,FL
`

Send that data to your local reportstream and get back a report id:

```
curl -X POST -H "client:simple_report" -H "Content-Type: text/csv"  --data-binary "@./junk/mix.csv" "http://localhost:7071/api/reports"
```

Response:
```
{
  "id" : "7cee139b-bbaa-4b03-abb3-2a4e11ba339c",
 <etc>
}  
```

Local API query to see the Submission details:

`curl "http://localhost:7071/api/waters/report/7cee139b-bbaa-4b03-abb3-2a4e11ba339c/history"
`
If you see reasonable destination data, it works.  If you see a ton of repeats, it failed.  You can even submit the data prior to changing the code to the new branch, then do the history curl to see the bad result, then update the code to this branch, then do the history curl again, to see the good result.

#### How to test in Staging

I have sent a test submission 01a6f9ff-0fd8-465f-9ab5-c38dd34f139a to Staging, that exhibits the problem.    You can try running these steps before and after merging this PR onto Staging.  Before = bad/screwed-up data.   After = hopefully fixed.

Run these three commands:

```
./prime login --env staging
export TOK=$(cat ~/.prime/accesstoken.json | grep token | cut -d"," -f1 | cut -d":" -f2 | tr -d \")
curl -H "authentication-type: okta" -H "authorization: bearer $TOK" "https://staging.prime.cdc.gov/api/waters/report/01a6f9ff-0fd8-465f-9ab5-c38dd34f139a/history"

```
## Changes
- The problem was caused by the fact that a receive Action has multiple reports attached to it.  The query where clause used to constrain to just the report with a non-null sending_org, but that constraint was removed from code; I'm not sure why.  Its now back in.

### Testing
- [X] Tested locally?
- [X] Added tests?

### Process
- [N] Updated the release notes?

## Fixes 
#6454 

## To Be Done
- I did notice that if the useBatchHeaders is false for a receiver, the bug still surfaces.  CA has useBatchHeaders as false in localhost, so if you generate this data, you'll still see the issue:
./prime data --input-fake 40 --input-schema primedatainput/pdi-covid-19 --output ./junk/mix.csv --target-states 
MD,CA,FL,VT

I believe there are currently no receivers in Prod that use that setting, so this is low priority, I'm hoping.


prime-router/src/main/kotlin/history/azure/DatabaseSubmissionsAccess.kt
```diff
@@ -46,19 +46,24 @@ class DatabaseSubmissionsAccess(
...
         klass: Class<T>
     ): T? {
+        if (orgName == null) error("Submission query must be constrained by a sender's organization name")
         return db.transactReturning { txn ->
             DSL.using(txn)
                 .select(detailedSelect())
                 .from(ACTION)
                 .where(
                     ACTION.ACTION_ID.eq(actionId)
+                        .and(ACTION.SENDING_ORG.eq(orgName))
```
jimduff-usds, line 66: This is the change, here.

All the other changes in this PR are just to get the orgName to this query, so we can use it here.

kevinhaube (commented):
When testing, I sent as `simple_report` and loading the details page from the receiver's Daily Data list resulted in 

```
Exception: NullPointerException: action.sendingOrg must not be null
```

which seems related to the change you called out in this PR. That said, the items _did_ send as `simple_report` and could be viewed in their Submissions list. When loading the submission details from the sender's end, I got _more_ consistent behavior than before, but still multiples of each: a pattern of 3 destinations, only having items in the `sentReports` array on the first of the 3 objects.

Screen Shot 2022-08-22 at 3 20 27 PM

jimduff-usds:
@kevinhaube thanks for catching those issues!    I think the NullPointer is fixed.

Regarding the second issue, I think there were multiple issues with the multiple destinations and I’ve only fixed one of them.    The queries are very complicated; it'll take me a while to dissect the issue, so we might want to go with this fix for now, then I'll do further work.

kevinhaube (approved):
The `NullPointerException` is fixed! Noted, regarding complexity of the issue -- I figured the positive change (i.e. less duplicates) was still a step forward! 👍🏼 Thanks for taking this on

PR #6476, opened 2022-08-23
refactor: move value sets gets to react-query

doug-s-nava:
this is the first PR for creation of an MVP of react query based usage in the app.

this change covers:

* installation of react-query, creation of QueryClient Provider in app
* replacement of existing value sets data fetching hooks with react-query based hooks
* react query test tooling
* tests around new hooks and some additions and updates to tests for value sets index, which consumes them

## Test Steps:

Regression testing for value sets:

1. start servers on this branch
2. visit http://localhost:3000/admin/value-sets, logging in as necessary
3. _VERIFY_: value sets page loads as normal, see screenshot
4. click on `residence_type` value set
5. _VERIFY_: value sets detail page loads as normal, see screenshot
6. update or create a new a value set row, and save
5. _VERIFY_: your change is reflected on the page
6. refresh
5. _VERIFY_: value sets detail page loads correctly and includes your change

## Screenshots

Screen Shot 2022-08-23 at 5 22 45 PM

Screen Shot 2022-08-23 at 5 20 19 PM

### Testing
- [x] Tested locally?
- [x] Added tests?


frontend-react/src/App.tsx
```diff
@@ -112,166 +115,184 @@ const App = () => {
...
-                        fallbackComponent={() => <ErrorPage type="page" />}
-                    >
-                        <DAPHeader
-                            env={process.env.REACT_APP_ENV?.toString()}
-                        />
-                        <GovBanner aria-label="Official government website" />
-                        <SenderModeBanner />
-                        <ReportStreamHeader />
-                        {/* Changed from main to div to fix weird padding issue at the top
+                <QueryClientProvider client={queryClient}>
```
doug-s-nava, line 126: all of the below changes are just whitespace changes as a result of adding the QueryClientProvider into the heirarchy.

Note: it might be nice to split all of our providers out into an app wrapper component so we don't get this huge diff every time we update this file...

frontend-react/src/__mocks__/LookupTableMockServer.ts
```diff
@@ -23,12 +27,14 @@ const lookupTables: LookupTable[] = [1, 2, 3].map((i) => ({
     tableSha256Checksum: "checksum",
 })) as LookupTable[];
 
-const lookupTableData: ValueSet[] = [1, 2, 3].map((i) => ({
+const lookupTableData: ApiValueSet[] = [1, 2, 3].map((_i) => ({
     name: "sender_automation_value_set",
-    createdBy: `test${i}@example.com`,
-    createdAt: "now",
+    created_by: "",
```
doug-s-nava, line 32: this is an inconsequential change, but better mirrors the data that is actually coming back from the API

frontend-react/src/utils/CustomRenderUtils.tsx
```diff
@@ -50,6 +51,25 @@ const SessionWrapper =
         );
     };
 
+export const QueryWrapper =
+    (client: QueryClient = new QueryClient()) =>
+    ({ children }: PropsWithChildren<{}>) =>
+        <QueryClientProvider client={client}>{children}</QueryClientProvider>;
+
+const AppWrapper =
```
doug-s-nava, line 59: realizing now that this wrapper i created is unused. But maybe we should be using it everywhere by default?

frontend-react/src/hooks/UseValueSets.ts
```diff
@@ -0,0 +1,152 @@
...
+Have not tested, but I feel like to get the full value of react-query's refetching behavior we may want to go with the last option.
+When react-query refetches it will just re-run the function we passed to it, this would not take into account changes in variable dependencies
+unless we bake those into function itself?
+
+Note that I ran into type errors when trying to use higher order functions to avoid the inlining
+*/
+
+// will return either a value set or value set row depeding on
+// value set table name
+// we may not need the generic here, react-query probably handles that. But it couldn't hurt?
```
kevinhaube, line 37: Agreed, I think we're good to just type it at the `useQuery<T>()` level since we aren't touching the data before then

frontend-react/src/hooks/UseValueSets.ts
```diff
@@ -0,0 +1,152 @@
...
+const getLatestData = <T>(tableName: LookupTables, version: number) => {
+    const endpointHeader = lookupTableApi.getTableData<T>(version, tableName);
+
+    return axios
+        .get<T>(endpointHeader.url, endpointHeader)
+        .then((response) => response.data);
+};
+
+// not going to type this yet, as this function will likely go away when we bring in fetch by provider
```
kevinhaube, line 46: > ...when we bring in fetch by provider

Can you elaborate? I'm blanking on "fetch by provider"

frontend-react/src/hooks/UseValueSets.ts
```diff
@@ -0,0 +1,152 @@
...
+    valueSetArray: Array<T>;
+    error: any;
+} => {
+    let versionData: Partial<TableAttributes> | null;
+    let error;
+    let valueSetArray;
+
+    // get all lookup tables
+    const { error: tableError, data: tableData } = useQuery<LookupTable[]>(
+        ["lookupTables"],
```
kevinhaube, line 106: I'm a little confused on what the key does. Are the integral enough that we should be enumerating them?

frontend-react/src/hooks/UseValueSets.ts
```diff
@@ -0,0 +1,152 @@
...
+            versionData = findTableByName(tableData, dataTableName);
+        } catch (e) {
+            error = e;
+            versionData = null;
+        }
+    }
+
+    // unclear how to resolve the issue of defining a function to account for the absence of
+    // version when the `enabled` param should take care of not running if that's not present...
+    // get value for lookup table version
```
kevinhaube, line 129: It doesn't seem like we'd need the extra safety 🤷🏼‍♂️ If I understand, we're not calling the first query if a supplied version is given...which then could possibly be `null` resulting in the second query not calling? Would that result in us getting back `{ error: undefined, valueSetArray: undefined }`? If so, the return type should be updated to match that possibility.

frontend-react/src/hooks/UseValueSets.ts
```diff
@@ -0,0 +1,152 @@
...
+unless we bake those into function itself?
+
+Note that I ran into type errors when trying to use higher order functions to avoid the inlining
+*/
+
+// will return either a value set or value set row depeding on
+// value set table name
+// we may not need the generic here, react-query probably handles that. But it couldn't hurt?
+const getLatestData = <T>(tableName: LookupTables, version: number) => {
+    const endpointHeader = lookupTableApi.getTableData<T>(version, tableName);
```
kevinhaube, line 39: This is nice, it seems like we have a _lot_ of flexibility in how we set up API tooling around react-query since everything is written outside of `react` now :D

frontend-react/src/hooks/UseValueSets.test.ts
```diff
@@ -0,0 +1,111 @@
...
+            () => useValueSetsTable(LookupTables.VALUE_SET_ROW),
+            { wrapper: QueryWrapper() }
+        );
+        await waitFor(() => !!result.current.error);
+        expect(result.current.error.message).toEqual(
+            `Table 'sender_automation_value_set_row' was not found!`
+        );
+    });
+
+    test("returns error when table list call errors", async () => {
```
kevinhaube, line 62: Because of how long the console errors from react-query are, can we do a mock for `console.error` in this file?

frontend-react/src/hooks/UseValueSets.test.ts
```diff
@@ -0,0 +1,111 @@
...
+        expect(name).toEqual("sender_automation_value_set");
+        expect(createdAt).toEqual("now");
+        expect(createdBy).toEqual("test@example.com");
+        expect(system).toEqual("LOCAL");
+    });
+
+    test("returns expected values when using supplied table version", async () => {
+        const { result, waitFor } = renderHook(
+            () => useValueSetsTable<ValueSet>(LookupTables.VALUE_SET, 3),
+            { wrapper: QueryWrapper() }
```
kevinhaube, line 40: Is there a comparable setup to the [custom renderer pattern](https://testing-library.com/docs/react-testing-library/setup/#custom-render) we can use to abstract this wrapper setup?

frontend-react/src/utils/CustomRenderUtils.tsx
kevinhaube: It looks like you're using this in `ValueSetsIndex` tests via `renderWithFullAppContext`?

kevinhaube (commented):
Love that we're moving fetch logic outside of `react`. Looks like we're going to have a lot more flexibility now!

Leaving the approval off for now because there's some open discussions and thing I wanna understand first.

frontend-react/src/hooks/UseValueSets.ts
doug-s-nava: yeah, that is some confusing short hand there. I'm planning in PR 3 (or 4, technically, with the app wrapper thing in the mix now), to introduce a provider that will `provide` access to a universal fetch function. Mainly meant to mention that it's not worth spending too much time worrying about this function as it will go away soon.

frontend-react/src/hooks/UseValueSets.ts
doug-s-nava: down the line, definitely. I thought about that a bit this morning, and I think that this is something we can handle with our approach to resource classes. Without getting into the business of defining endpoints the way rest-hooks does, we can put some configuration into the resource class to keep track of the query keys for later reference.

as far as what they're used for, it's cache related stuff for the most part. One thing I mentioned in the comment [here](https://github.com/CDCgov/prime-reportstream/pull/6476/files#diff-82cdafd58a9ecaad3fbce3add39b460b71607726c282ed734647b7434a1386aaR25), though, is that you can also pass values in the keys for things like query params that, in addition to defining the cache storage location, also can be used within the fetch function. This is a decent resource on that, the official docs hint at it but don't really show it in action: 

https://stackoverflow.com/a/68111112/5500298

We may want to dig into that pattern later, but it is a little more complex and with what I've done so far it seems to work fine without it

frontend-react/src/hooks/UseValueSets.test.ts
doug-s-nava: yeah. Maybe it's worth bringing in that tech debt ticket at this point?

frontend-react/src/utils/CustomRenderUtils.tsx
doug-s-nava: o yeah, tru

frontend-react/src/pages/admin/value-set-editor/ValueSetsIndex.test.tsx
```diff
@@ -1,18 +1,68 @@
...
+        name: "engelbert anyname",
+        system: "a very different system",
+        createdAt: "Wednesday",
+        createdBy: "me",
+    },
+];
+
+let mockUseValueSetsTable = jest.fn();
+
+jest.mock("../../../hooks/UseValueSets", () => {
```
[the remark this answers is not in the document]
doug-s-nava: 🤔 your way is probably better! This is the pattern I came up with on previous projects and it always worked for me there. For whatever reason it works less well on this project, as for some reason if you pass in a mock fn to the mocked import definition, changes you make to that mock (like calling `mockImplementation` within a given test) don't get picked up, so I end up just recreating the entire function in each test.

I probably didn't use spyOn due to the fact that in other test libraries, spies and mocks are different, and spies generally don't allow you to change behavior. Looks like in jest spyOn just returns a mock, which, ok!

If it's cool with you, maybe we can leave this as is for now, and going forward we can make a point to use spyOn instead? Maybe make a cleanup ticket that we'll never get around to?

frontend-react/src/hooks/UseValueSets.test.ts
kevinhaube: Which ticket in particular? Couldn't find it but I know we've talked about maybe wanting to use the errors somehow in tests

frontend-react/src/hooks/UseValueSets.ts
kevinhaube: Ahh, yeah I was having trouble deciphering the pattern in that post but I se now...it uses the array with the key first, and trailing with params. Interesting that we have that option. I'm with you on it being more complex. Wonder what, if any, benefit that has or if it was built in strictly for having an option.

frontend-react/src/hooks/UseValueSets.ts
kevinhaube: Oh, that's super cool! Good to know.

frontend-react/src/hooks/UseValueSets.test.ts
doug-s-nava: oh, yeah, it might be worth setting that up to use the AppWrapper by default. We'd still need some of the individual wrappers for when we want custom behavior (like turning off retries https://github.com/CDCgov/prime-reportstream/pull/6476/files#diff-a5b4cf7595f3c0e3d62866d65b3d69e271f15379db38680101b637af99d3fa35R73), but that could potentially be helpful.

frontend-react/src/pages/admin/value-set-editor/ValueSetsIndex.test.tsx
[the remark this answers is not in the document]
kevinhaube: All good by me! I was curious! I didn't realize other frameworks treated spies differently -- it baffles me why a spy in jest _also_ overrides the function with a mock. I guess that's how jest is set up to track whether it's called or not.

frontend-react/src/hooks/UseValueSets.ts
doug-s-nava: a few things to mention here:
* my first pass at this function included early returns. This got complex in part because with the hooks rules, we need to call all useQuery hooks in all invocations. The better pattern seemed to be what I have here where we define the return variables (error and data, or in this case `valueSetArray`) up top, and then mess with them during the course of the function, making sure that by the end of the function they are defined the way we want them before returning.
* as such, I don't think there's a possibility of returning `undefined` values here. The function will always reach the end, and at the end we always define the values one way or another
* the complexity I was getting at in this comment is around needing to pass that `Promise.reject` function to handle the case of `versionData?.version` not being defined, even though we're explicitly telling useQuery not to run the function if it's not defined. There may be a better way to do this, but I couldn't find one.

kevinhaube: Ahhh, okay, I get the flow a little better now. So the _first_ call up top is the conditionally-run call based on if there's version data or not already supplied. If not, it gets it for us. But the call below, where you have `Primse.reject` _always_ runs. Error uses the right error, valueSetArray creates the array and defaults to empty.

I'll add a suggestion to try out but no biggie if that doesn't work

frontend-react/src/hooks/UseValueSets.ts
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    // unclear how to resolve the issue of defining a function to account for the absence of
+    // version when the `enabled` param should take care of not running if that's not present...
+    // get value for lookup table version
+
+    // not entirely accurate typing. What is sent back by the api is actually ApiValueSet[] rather than ValueSet[]
+    // does not seem entirely worth it to add the complexity needed to account for that on the frontend, better
+    // to make the API conform better to the frontend's expectations. TODO: look at this when refactoring the API
+    const { error: valueSetError, data: valueSetData } = useQuery<Array<T>>(
```
kevinhaube, line 134:
```suggestion
    const { error: valueSetError, data: valueSetData } = useQuery<Array<T>>(
        ["lookupTable", versionData?.version, dataTableName],
        () => getLatestData<Array<T>>(dataTableName, versionData!.version!!),
        { enabled: !!versionData?.version }
    );
```
This didn't throw any issues for me locally. The bang-chain `!.` asserts `versionData` has `version`, and then the trailing double-bang`!!` asserts that `version` is not undefined.

kevinhaube (approved):
LGTM! Suggested commit is non-blocking, just a way to clean up some of the logic for now

frontend-react/src/hooks/UseValueSets.ts
doug-s-nava: as our friends across the pond would say: `brilliant`

frontend-react/src/hooks/UseValueSets.test.ts
doug-s-nava: this is the ticket https://app.zenhub.com/workspaces/experience-607d9d5e68b95200150fec37/issues/cdcgov/prime-reportstream/6289. If I'm adding `--silent` here I guess that would close the ticket. I'm a fan of that approach - if you need logs for debugging you can just remove the flag. I could do something like this though
```
beforeAll(() => {
  jest.spyOn(console, 'log').mockImplementation(() => {});
  jest.spyOn(console, 'error').mockImplementation(() => {});
  jest.spyOn(console, 'warn').mockImplementation(() => {});
});
```

PR #6498, opened 2022-08-24
New receivers for Nevada

### LINKED ISSUE #6463: Create a new feed for NV while moving to EpiTrax
@MauriceReeves-usds recommends we set up a separate feed for EpiTrax. We will want to set up 'deidentify: true' for their feed until we get the all-clear to move to production which should happen by Sept 9th at the latest.

brick-green:
This PR creates new receivers for Nevada. Nevada is changing vendors. The new vendor has requested test and production receivers.

Test Steps:
1. set new settings: `./prime multiple-settings set --input ./settings/organizations.yml`
2. create fake simple report file for NV: `./prime data --input-fake 2 --input-schema primedatainput/pdi-covid-19 --output <PATH TO FILE>.csv --target-states NV`
3. Open file and set processing mode code to `T` for one result. The other result should have a value of `P`
4. Run files through and ensure one result goes to prod receiver and other result goes to test receiver

## Changes
- added elr-test reciever
- added scripts to /staging and /prod folders

### Testing
- [X] Tested locally?

## Linked Issues
- Fixes #6463


GilmoreA6 (commented):
Hey Brick, question about why the settings in the Prod and Staging folders have a lot more to them than the settings in the  organizations.yml file?

brick-green:
@GilmoreA6 There's two things going on there. First, the files in /prod and /staging include all the default values. We don't typically include those in the organizations file just to keep the file brief. But if you run `./prime multiple-settings set --input ./settings/organizations.yml` then run `./prime multiple-settings get --filter nv-phd` you can see that all those default fields are added.

Second, I removed the exisiting receiver in the organization file but I left it in the /staging and /prod files. I think we will need to do parallel reporting so we still need the existing receiver. But I didn't see the value in leaving it locally then coming back later to do another PR to remove it. Figured I could just knock it out now.

GilmoreA6 (approved):
Thanks Brick!

GilmoreA6 (approved):
Approve addition to default any MSH-4-1 to CDC PRIME and any MSH-4-2 to 11D2030855. And add MSH-5 (receiving application) and MSH-6 (receiving facility) value defaults of "NVDOH' to previous changes.

PR #6510, opened 2022-08-25
5754 Generate HL7 ORC segment from FHIR

carlosfelix2:
This PR generates the ORC segment from FHIR data. Note the ORC segment is optional, and it is generated only when needed.

Test Steps:
1. Check out the branch for this PR.
2. Run the following command to convert a FHIR bundle to HL7.  
`./prime fhirdata --input-file "src/testIntegration/resources/datatests/HL7_to_FHIR/sample_co_1_20220518-0001.fhir" -s metadata/hl7_mapping/ORU_R01/ORU_R01-base.yml`
3. Verify an HL7 message is generated from the conversion and includes the ORC segments.
4. Look at the source HL7 used at `src/testIntegration/resources/datatests/HL7_to_FHIR/sample_co_1_20220518-0001.hl7` and verify the data for the ORC segments are identical.  There are exceptions:
   a. ISO is not added when there is no universal ID specified
   b. A person's suffix is added in the suffix field instead of the professional suffix or degree field per FHIR's spec location.
   c. Date times have added precision, but there is an issue open for that

## Changes
- Added config for the FHIR to HL7 conversion of the ORC segment
- Modified values for address type

### Testing
- [x] Tested locally?
- [ ] Added tests?

### Process

## Linked Issues
- #5754


prime-router/metadata/hl7_mapping/ORU_R01/base/order.yml
```diff
@@ -0,0 +1,98 @@
...
+
+  - name: ordering-organization
+    resource: '%resource.basedOn.resolve().extension(%`rsext-entering-organization`)'
+    condition: '%resource.exists()'
+    value: [ '%resource.value' ]
+    hl7Spec: [ '%{hl7Order}-17' ]
+
+  - name: ordering-facility-name
+    resource: '%resource.basedOn.resolve().requester.resolve().organization.resolve()'
+    schema: xon-organization
```
PatriciaPerozoUSDS, line 78: ohh smart to add the organization

PatriciaPerozoUSDS (changes requested):
When I ran this locally to test, the TQ1 segments that existed on master disappeared. It seems like adding the conditional order schema above it, removes the TQ1 segment. I'm not sure why this is happening to be honest.

Everything else looks good I've rebased onto this branch already and refactored to use a bunch of these improvements!

carlosfelix2:
Good catch!  Fixed.
> When I ran this locally to test, the TQ1 segments that existed on master disappeared. It seems like adding the conditional order schema above it, removes the TQ1 segment. I'm not sure why this is happening to be honest.
> 
> Everything else looks good I've rebased onto this branch already and refactored to use a bunch of these improvements!

PatriciaPerozoUSDS (approved):
Thanks for fixing the missing TQ1 issue!

PR #6512, opened 2022-08-25
Cle/update azure functions version

clediggins-usds:
Copy of depbot that we already merged to test, updates azure functions to 2.x


MauriceReeves-usds (approved):
Everything looked good in my local testing.

PR #6517, opened 2022-08-25
Implements table name into LookupTableApi calls

### LINKED ISSUE #5947: Valueset Editor: Refactor UI for refactored lookup table backend structure
## Problem statement

Once the backend structure for Valuesets UI has been refactored (#5909), we will need to change the UI to read/write from the new tables. Modify the Valuesets Editor UI Details page to load the appropriate table based on the existing url routing, i.e. instead of using :value-set to _filter_ on `sender_automation_value_set_row`, it now simply replaces "sender_automation_value_set_row" in the API calls to retrieve and update.

## What you need to know

- [Valueset Editor Backend Architecture spike](https://cdc-my.sharepoint.com/:w:/g/personal/srz7_cdc_gov/EcT4ehC-ya1HjWqswoF64ksBKooh0wmuth75DY-lCWKeWg?e=2VbYEF)
- Lookup Table API Enhancement stories: #5706 #5707 #5708

## Acceptance criteria

- [ ] Modify UI to pull from valueset-specific tables
- [ ] Unit tests

## To do

- [ ] ...

kevinhaube:
This PR following #5909 this PR will swap out the enumerated value of `SENDER_AUTOMATION_TABLE_ROW` with the true table name in all API calls to the lookuptable api.

Test Steps:
1. Log in as a PrimeAdmin
2. Ensure the Value Sets list loads -> select one and go to its detail page
3. Ensure the details page loads -> try to edit one and save
4. Ensure the save and activation calls in the network tab conclude with 200s
5. Check all network calls to this api (`content`, `activate`, and a call titled the same as your table should appear in your network tab upon save) use the new format of URL:

NEW
```
http://localhost:3000/api/lookuptables/gender/{version}/content
# where "gender" is the name of any lookup table
```
OLD
```
http://localhost:3000/api/lookuptables/sender_automation_table_row/{version}/content
# where "sender_automation_table_row" was an enumerated value
```

Known Issues:

When saving a new value, creating a new version, then swapping the value _back_ to its previous value, the following error occurs. Seems related to API functionality, not front-end calling of the API:
Screen Shot 2022-08-25 at 3 15 06 PM

## Changes
- Hook layer of LookupTableApi -> piping the true table name through instead of replacing with enumerated value

### Testing
- [X] Tested locally?
- [ ] Added tests?

### Process

## Linked Issues
- Fixes #5947 

## To Be Done
- #issue

    - new endpoint A
    - new endpoint B


frontend-react/src/hooks/UseLookupTable.ts
```diff
@@ -123,21 +123,21 @@ export const getSenderAutomationDataRows = async <T>(
 };
 
 const useLookupTable = <T>(
-    tableName: LookupTables,
-    dataSetName: string | null = null,
+    dataSetName: string | LookupTables = LookupTables.VALUE_SET,
```
kevinhaube, line 126: This looks to be the only use of this enum anymore (that and the comparison on line 134 below).
@sethdarragile6 @doug-s-nava Waited to get input from y'all: is this enum something we'd intended to disappear after this change? If so I'll take it out and we can use a single string default for `dataSetName` here and on line 134 in the comparison.

frontend-react/src/hooks/UseLookupTable.ts
```diff
@@ -146,19 +146,14 @@ const useLookupTable = <T>(
...
             });
-    }, [dataSetName, tableName, version]);
+    }, [dataSetName, version]);
 
     return { valueSetArray, error };
 };
 
-export const useValueSetsTable = () =>
-    useLookupTable<ValueSet>(LookupTables.VALUE_SET);
+export const useValueSetsTable = () => useLookupTable<ValueSet>();
```
kevinhaube, line 154: Opted to provide a default in `useLookupTable` params with refactor

frontend-react/src/hooks/UseLookupTable.ts
kevinhaube: It is also still used in tests it appears

doug-s-nava: yeah, the enum would still be in use. What do we gain by eliminating it? Not exactly clear on the goal of this PR, beyond the simplification of not needing to default to null (which will go away)

doug-s-nava (changes requested):
can you clarify what's going on in this PR? I think I may not being seeing the purpose

kevinhaube:
The PR swaps out the old API URL format for the new API URL format in a response to Seth's backend refactor:
```
new:
http://localhost:3000/api/lookuptables/{tableName}/{version}/content

old:
http://localhost:3000/api/lookuptables/sender_automation_table_row/{version}/content
```

frontend-react/src/hooks/UseLookupTable.ts
kevinhaube: The original params required both a `LookupTables` enum value AND the data set name -- then, its only use for `dataSetName` was to measure if it was null or not, which it was by default, meaning that it'd pass along the `tableName` value to the network calls. By passing setting this to default to the enum value, we're saying "if they don't provide `dataSetName` then this is a call for `getSenderAutomationData`. If they DO provide one (i.e. it isn't the default) then it's for `getSenderAutomationDataRows`

doug-s-nava:
My confusion was around the fact that these changes support a change in the API logic behind the `http://localhost:3000/api/lookuptables/{tableName}/{version}/content` endpoint. Previously we sent all requests to `http://localhost:3000/api/lookuptables/sender_automation_table_row/{version}/content` and that response was filtered on the frontend here https://github.com/CDCgov/prime-reportstream/blob/14ece8856c01d7e979bfa5f72c484e2aa21bb983/frontend-react/src/pages/admin/value-set-editor/ValueSetsDetail.tsx#L158 . That filtering, I gather is now happening on the backend

frontend-react/src/hooks/UseLookupTable.ts
doug-s-nava: I see now why this enum may no longer be useful. I think you can remove it and just hardcode the value of  LookupTables.VALUE_SET where necessary

frontend-react/src/pages/admin/value-set-editor/ValueSetsDetail.tsx
```diff
@@ -105,9 +105,8 @@ const saveData = async (
         throw new Error("A null row was encountered in saveData");
     }
 
-    const endpointHeaderUpdate = lookupTableApi.saveTableData<ValueSetRow[]>(
-        LookupTables.VALUE_SET_ROW
-    );
+    const endpointHeaderUpdate =
```
doug-s-nava, line 108: based on what I'm seeing I think we can clean up this file a bit more. 

Can we:

* remove the dependency on sending up the entire table on save? see https://github.com/CDCgov/prime-reportstream/blob/14ece8856c01d7e979bfa5f72c484e2aa21bb983/frontend-react/src/pages/admin/value-set-editor/ValueSetsDetail.tsx#L113
* remove the need to filter all data based on the name of the data set we want to display? see https://github.com/CDCgov/prime-reportstream/blob/14ece8856c01d7e979bfa5f72c484e2aa21bb983/frontend-react/src/pages/admin/value-set-editor/ValueSetsDetail.tsx#L158

doug-s-nava (changes requested):
has the API change to support this gone out yet? if not, it seems like we should hold off

frontend-react/src/hooks/UseLookupTable.ts
sethdarragile6: Just my $.02 here. It might be better form to just change the values in the LookupTables enum to match what we expect on the backend, rather than eliminating it. Just thinking of the fact that that table name is coming from the URL, which one can manually change to whatever crazy thing you can dream up. And if we're not already doing it, it could be a quick win to validate that against the enum before it's even passed on to the API call.

sethdarragile6:
> My confusion was around the fact that these changes support a change in the API logic behind the `http://localhost:3000/api/lookuptables/{tableName}/{version}/content` endpoint. Previously we sent all requests to `http://localhost:3000/api/lookuptables/sender_automation_table_row/{version}/content` and that response was filtered on the frontend here
> 
> https://github.com/CDCgov/prime-reportstream/blob/14ece8856c01d7e979bfa5f72c484e2aa21bb983/frontend-react/src/pages/admin/value-set-editor/ValueSetsDetail.tsx#L158
> 
> . That filtering, I gather is now happening on the backend, based on the string passed in the {tableName} path param

No filtering was implemented in the API. Essentially the only thing that changed on the backend is that the sender_automation_table_row table was split out into a table per valueset. Once the UI is hooked up to these tables instead, sender_automation_table_row should be removed from the backend entirely (I think I still need to put in a ticket for that)

frontend-react/src/pages/admin/value-set-editor/ValueSetsDetail.tsx
sethdarragile6: For the first comment: we will _still_ need to save/send up the entire table. Unfortunately, that is still a limitation of the lookuptable api. The nice thing about that now is that the tables we're saving are muuuuch smaller than sender_automation_table_row. Except that patient_preferred_langage one, it's still a beast, but whatcha gonna do.

For the second comment: absolutely! no need to filter anymore.

frontend-react/src/hooks/UseLookupTable.ts
kevinhaube: @sethdarragile6 Changes like that sound like something that'll be overwritten by Doug's react-query refactor of this part of the app

frontend-react/src/pages/admin/value-set-editor/ValueSetsDetail.tsx
doug-s-nava: @sethdarragile6 So the POST request to update table data will go to `{api_root}/{tableName}` ex. `http://localhost:3000/api/lookuptables/patient_preferred_language` instead of `{api_root}/sender_automation_value_set_row` but we will still send up ALL rows? That doesn't really make sense to me - not sure why the endpoint would change if the API is still expecting all rows

sethdarragile6: @doug-s-nava maybe we should zoom about this, but yes- we are still sending up ALL rows in whatever particular table we're editing (`patient_preferred_language`, in this case). The only real change in the POST endpoint request should be swapping out the tablename `sender_automation_value_set` for the tablename `patient_preferred_language` (or `gender` or `yesno` etc). 

As far as I know, the POST endpoint still expects that all rows for the specified table are being sent. If you were to only send the particular item you edited, for example, you would wind up updating the table so that it had ONLY that row in it. Hilarious! (not so much)

doug-s-nava: ok, just wanted to make sure. I'm assuming that the back end is then doing some sort of filter based on the path segment (ex. `patient_preferred_language`)? Just trying to be clear on why that endpoint segment is changing if we still need to send up the entire table

kevinhaube:
@doug-s-nava Do the suggested changes make sense if this is being overridden by your work with `react-query`?

kevinhaube:
Yes, the changes have been merged. They were a prereq for this work to get done: https://github.com/CDCgov/prime-reportstream/pull/6389

doug-s-nava (approved):
✅ ✅ ✅ ✅

PR #6572, opened 2022-08-29
Improve Lastmile tracking

TomNUSDS:
## Problem: 
Last Mile page showing send failure allows a Prime Admin to manually requeue for a resend; however, it a didn't have any indication if a resend has happened or not. This lead to confusion over if it needed to be done.

### server-side:
- New API endpoint to return Actions that are of type `resend` (these are already added when a resend happens). `api/adm/getresend` w/ parameter `days_to_show` that returns matching rows. All columns are returned. Only allows Prime Admin access. Similar to existing `  /adm/getsendfailures`
- Updated `resend` Action to include the username of who ran it.
- Updated OpenAPI doc `admin.yml` to include new call.
 
### client-side:
- Added rest-hook for `api/adm/getresend` endpoint.
- Refactored UI to:
  - Refresh button
  - support showing if an item has been resent
  - dialog displaying info shows failure and all resend attempts (there may be more than one match per item)


frontend-react/src/pages/admin/AdminLastMileFailures.tsx
```diff
@@ -20,15 +16,9 @@ export function AdminLastMileFailures() {
                     <Suspense fallback={<Spinner />} />
                 </h3>
             </section>
-            <NetworkErrorBoundary
```
TomNUSDS, line 23: Moved down into table so whole page doesn't refresh

frontend-react/src/pages/resources/legacy-page-content/CsvSchemaDocumentation.tsx
```diff
@@ -270,10 +270,11 @@ export const CsvSchemaDocumentation = () => {
                                         {section.title}
                                     </h3>
 
-                                    {section.items?.map((item) => {
+                                    {section.items?.map((item, index) => {
                                         return (
                                             <CsvSchemaDocumentationItem
                                                 item={item}
+                                                key={`CsvSchemaDocumentationItem-${index}`}
```
TomNUSDS, line 277: Noticed this key-missing error in the console while debugging, figured I'd fix it.

kevinhaube: Oh! Good to know, I'll remove this from one of my branches -- I'd plugged in a fix earlier today.

frontend-react/src/components/Admin/AdminLastMileFailuresTable.tsx
```diff
@@ -9,24 +9,251 @@ import {
...
+        });
+
+    return (
+        <Table
+            key="lastmiletable"
+            aria-label="List of failed sends"
+            striped
+            fullWidth
+        >
+            <thead>
```
kevinhaube, line 235: Can this component be refactored with the RS `Table` component? It seems like a lot of this JSX can get cleaned up.

frontend-react/src/components/Admin/AdminLastMileFailuresTable.tsx
```diff
@@ -9,24 +9,251 @@ import {
...
+                        </Button>
+                        <span
+                            className={"rs-resendmarker"}
+                            title={"Resends attempted."}
+                        >
+                            {resends.length ? " ⚠️ " : null}
+                        </span>
+                    </td>
+                    <td>
+                        <Link
```
kevinhaube, line 199: This is possible through the `LinkableColumn` feature. 

Shape: https://github.com/CDCgov/prime-reportstream/blob/master/frontend-react/src/components/Table/Table.tsx#L29-L33

Code: https://github.com/CDCgov/prime-reportstream/blob/58599e355ad2a0b357005190c185a8b8497f0b81/frontend-react/src/components/Table/ColumnData.tsx#L77-L92

Example: https://github.com/CDCgov/prime-reportstream/blob/58599e355ad2a0b357005190c185a8b8497f0b81/frontend-react/src/pages/daily/Table/ReportsTable.tsx#L105-L108

frontend-react/src/components/Admin/AdminLastMileFailuresTable.tsx
```diff
@@ -9,24 +9,251 @@ import {
...
+                            title={"Jump to Settings"}
+                            to={linkRecvSettings}
+                            key={`recv_link_${eachRow.pk()}`}
+                            className={"font-mono-xs"}
+                        >
+                            {eachRow.receiver}
+                        </Link>
+                    </td>
+                    <td>
+                        <Button
```
kevinhaube, line 209: This can be achieved with the ActionableColumn feature.

Shape: https://github.com/CDCgov/prime-reportstream/blob/58599e355ad2a0b357005190c185a8b8497f0b81/frontend-react/src/components/Table/Table.tsx#L24-L27

Code: https://github.com/CDCgov/prime-reportstream/blob/58599e355ad2a0b357005190c185a8b8497f0b81/frontend-react/src/components/Table/ColumnData.tsx#L94-L114

kevinhaube (commented):
Nothing seems broken, but I'd like to file a non-blocking request to refactor this to use our `Table` component. It handles a _lot_ of the JSX for you, and it handles the use cases you have here, it seems.

frontend-react/src/pages/resources/legacy-page-content/CsvSchemaDocumentation.tsx
TomNUSDS: We could just make it a different PR and just merge it quickly? It's simple/save.

kevinhaube: Throw me on for review :)

frontend-react/src/components/Admin/AdminLastMileFailuresTable.tsx
TomNUSDS: Humm.. how did we end up with a component named `<Table>` since there's _already_ one in Trussworks? It is super confusing. https://trussworks.github.io/react-uswds/?path=/story/components-table--striped

frontend-react/src/components/Admin/AdminLastMileFailuresTable.tsx
TomNUSDS: > Nothing seems broken, but I'd like to file a non-blocking request to refactor this to use our `Table` component. It handles a _lot_ of the JSX for you, and it handles the use cases you have here, it seems.

I'm not against refactoring to use RS's Table vs Trusswork's Table; however, that's a major refactor that is atomic and could be done as a separate PR. I suspect to make the page work with RS's Table, some features may need to be added to that component.

I'll create an issue to do the refactor. Maybe it's something we can pair code on?

TomNUSDS:
Created a techdept Issue #6636

kevinhaube (approved):
LGTM!

PR #6597, opened 2022-08-31
Adoption of React Router v6

### LINKED ISSUE #6694: AuthElement redirects to "/login" on app refresh
**Describe the bug**
When on an authorized page (behind `AuthElement`), you can refresh the browser and the app will direct you back to the root url (assumed because auth has not re-processed yet).

**Impact**
Annoyance for engineers.
Unsure how often core users refresh, but some impact there.

**To Reproduce**
1. Go to any RS page that requires login
2. Use F5 or your browsers refresh button to refresh the page

**Expected behavior**
The same page/path should be in the window, no redirect should occur if you are logged in.

kevinhaube:
This PR updates our dependency version of `react-router` from major version 5 to major version 6. Additionally, the following refactors were made to either solve deprecation errors:

- Documentation added for all new patterns
- App Router extracted from App.tsx
- Deprecation of AuthRoute, now AuthElement pattern
- Updates the IA framework routing
- Deprecates path matching for useParams pattern
- Adds useValidParams to check params aren't undefined and throw if they are
- Deprecates useHistory for useNavigate hook
- Updates tests for all broken by router update
- Adds tests for new elements and functions

Test Steps:
1. Pull latest
2. Do not log in -- check the public site for routing errors.
3. Log in -- check Daily Data, Submissions, and Admin pages for routing errors or uncaught errors

### Testing
- [X] Tested locally?
- [X] Added tests?

### Process

## Linked Issues
- Fixes #5004 
- Fixes #6694

    - new endpoint A
    - new endpoint B


frontend-react/src/components/Admin/EditSenderSettings.tsx
```diff
@@ -253,39 +254,31 @@ const EditSenderSettingsForm: React.FC<EditSenderSettingsFormProps> = ({
...
-    /* useParams now returns possible undefined. This will error up to a boundary
-     * if the url param is undefined */
-    if (
-        orgName === undefined ||
-        senderName === undefined ||
-        action === undefined
-    )
-        throw Error("Expected orgName & settingType from path");
+    const { orgname, sendername, action } =
+        useValidParams<EditSenderSettingsProps>([
```
kevinhaube, line 264: This PR fully implements a pattern seen in PR 1 with undefined params checks. `useValidParams` will take a string array of keys to check incoming params values are not undefined.

frontend-react/src/components/Admin/EditSenderSettings.tsx
```diff
@@ -253,39 +254,31 @@ const EditSenderSettingsForm: React.FC<EditSenderSettingsFormProps> = ({
 };
 
 export type EditSenderSettingsProps = {
-    orgName: string;
-    senderName: string;
+    orgname: string;
+    sendername: string;
```
kevinhaube, line 258: Changed back because they're specific to the URL string. We can clean this up in a subsequent PR if camel case is mega important!

frontend-react/src/components/header/SignInOrUser.tsx
```diff
@@ -35,7 +35,7 @@ const SignInOrUser = () => {
             </a>
         </div>
     ) : (
-        <NavLink to="/daily-data">
+        <NavLink to="/login">
```
kevinhaube, line 38: Not sure how this happened but it seemed isolated to my branches 🤷🏼 fixing it.

frontend-react/src/pages/Login.tsx
```diff
@@ -15,7 +15,12 @@ export const Login = () => {
 
     const navigate = useNavigate();
     useEffect(() => {
-        if (authState && authState.isAuthenticated) navigate("/");
+        if (authState && authState.isAuthenticated) {
+            // TODO (router-refactor): Dynamically navigate users
```
kevinhaube, line 19: Will address this on shadow branch final pass before PR opens.

frontend-react/src/components/AuthElement.tsx
```diff
@@ -34,10 +34,23 @@ export const AuthElement = ({
...
+            navigate("/login");
+            return;
+        } // Not logged in, needs to log in.
+        if (requiredUserType && !authorizeMemberType()) {
+            navigate("/");
+            return;
+        } // Not authorized as current member type
+        if (requiredFeatureFlag && !CheckFeatureFlag(requiredFeatureFlag)) {
+            if (activeMembership.memberType === MemberType.PRIME_ADMIN) {
+                navigate("/admin/features");
```
kevinhaube, line 47: My favorite line in this PR:  as logged-in admins, if we try to navigate to a flagged route without the flag, it'll take us to the UI to add the flag. 
We could extend some functionality that conditionally checks if we pass a query like `?flagToAdd=[missing-flag]` and maybe use that to auto-populate the text field.

TomNUSDS (commented):
This PR does a lot of great stuff. As I mentioned in chat, maybe in the future doing the renaming/refactoring (that doesn't change functionality) should be broken out into a separate PR. It's a pain to do, but it makes reviewing easier.

That said, I have a couple of questions/comments before Approval.

frontend-react/src/hooks/UseValidParams.ts
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * @param expectedParams {string[]} A list of string keys to check incoming params
+ * */
+export const useValidParams = <T>(expectedParams: string[]): T => {
+    // Get all params as object
+    const params = useParams();
+    useEffect(() => {
+        // String index the object
+        const stringIndexedParams = params as StringIndexed;
+        expectedParams.forEach((param) => {
```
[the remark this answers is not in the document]
kevinhaube: This has to be an effect. Anything containing logic that interrupts rendering, _even_ if conditionally called, **all** gets called on the first render of a component unless inside a hook. Not wrapping this in `useEffect` made every page with the hook fail on the first render.

frontend-react/docs/rs-auth-element.md
```diff
@@ -0,0 +1,29 @@
...
+### Example use
+
+```typescript jsx
+import {AuthElement} from "./AuthElement";
+import {MemberType} from "./UseOktaMemberships";
+
+const Page = () => <div>I'm a page</div>
+export const PageWithAuth = () => (
+    <AuthElement
+        element={Page}
```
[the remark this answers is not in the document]
kevinhaube: Updated it. 

I also looked into the pattern a bit. React router's change doc made a good point here...
> In the section about upgrading to v5.1, we promised that we'd discuss the advantages of using regular elements instead of components (or element types) for rendering. Let's take a quick break from upgrading and talk about that now.
>
>For starters, we see React itself taking the lead here with the <Suspense fallback={<Spinner />}> API. The fallback prop takes a React element, not a component. This lets you easily pass whatever props you want to your <Spinner> from the component that renders it.

https://reactrouter.com/en/v6.3.0/upgrading/v5#advantages-of-route-element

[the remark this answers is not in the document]
kevinhaube: Seems like their use of `compnoent` is what I equate to a `ComponentAsFunction` type (aka `() => JSX.Element`). Whereas a `react element` to them is just the rendered JSX.Element

frontend-react/src/components/Admin/EditSenderSettings.tsx
kevinhaube: The type is just piped right to `useParams` but that doesn't handle validating whether or not the param came through. The goal is to _not_ render when url params are missing, so just calling use params doesn't suffice, you need to throw if one is missing that's required for render. Because paths are **runtime** variables, we can't expect TS to handle their type assertions. 

Those prop types were leftover, actually, from before. I had removed them, but felt that may step on some toes, so I added them back in according to their definitions in `master`.

kevinhaube: I think an ideal footprint and use would be

```typescript
type UrlParams = { one: string, two: string }
const params = useValidParams<UrlParams>(["one"])
params.one // Required by hook so no need for optional chain
params?.two // Still possibly undefined, requires optional chain
```

and to rename the parameter of `useValidParams` to `requiredParams` to indicate that the reason for the array since in these it just matches the destructured response...that's definitely a weird way to call it.

kevinhaube: The above approach offers the ability to put our filter params in URLs now, with MUCH easier assertion, and that'll enable users to share links to filtered submissions or daily-data tables. That's where this idea was headed when I decided to hook-ify it.

frontend-react/docs/rs-auth-element.md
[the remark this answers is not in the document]
TomNUSDS: OK. This is a nice deep dive into understanding why. Thanks!

frontend-react/src/components/Admin/EditSenderSettings.tsx
TomNUSDS: I'm still missing something. The call doesn't understand types and only checks for undefined. So if it's required, just make it required in the typedef. If it's optional, then the behavior when it's missing is custom based on the component. (e.g. Components with date ranges, the optional end date missing just means "use the current date".)

What happens when a required component parameter comes from a cgi parameter and it's malformed? Should there just be a default param used instead?

This might require a quick conversation to explain the real world use case? It would be nice to include @doug-s-nava and others so we can all understand the correct use case.

kevinhaube: Oh, I didn't clarify. UseParams returns everything as possibly undefined, independent of whatever type you pass in. 

Equal to Partial<T>. 

That's why the array is needed for checking required keys.

kevinhaube: Here's a quick screen grab of what that would look like.
Screen Shot 2022-09-09 at 1 49 06 PM

kevinhaube: Updated the comment and parameter name to better describe its existence

frontend-react/src/pages/Login.tsx
doug-s-nava: can we remove the comment, or is there still something to do here?

frontend-react/src/pages/Upload.tsx
```diff
@@ -53,3 +57,7 @@ export const Upload = () => {
         </div>
     );
 };
+
+export const UploadWithAuth = () => (
+    <AuthElement element={Upload} requiredUserType={MemberType.SENDER} />
```
doug-s-nava, line 62: [thought for future change] since this page isn't functional anymore, does it still need to be behind auth?

frontend-react/src/pages/admin/value-set-editor/ValueSetsIndex.tsx
```diff
@@ -19,9 +19,11 @@ import {
     LookupTable,
     LookupTables,
     ValueSet,
-} from "../../../config/endpoints/lookupTables";
+} from "../../../config/endpoints/lookupTables"; // TODO: move to config
+import { MemberType } from "../../../hooks/UseOktaMemberships";
+import { AuthElement } from "../../../components/AuthElement";
 
-const PAGE_TITLE = process.env.REACT_APP_TITLE; // TODO: move to config
```
doug-s-nava, line 24: I'd still like to move all of our process.env related globals to a config file

doug-s-nava (commented):
a few things to think about. note that I copied in some of my comments from the previous PR since I wasn't sure what had happened with those

in terms of smoke testing, things look pretty good except
* value sets detail pages don't seem to load
* refreshing on the daily data page returns you to home

frontend-react/src/components/Content/Templates/IASideNavTemplate.tsx
```diff
@@ -0,0 +1,40 @@
...
+            <div className="grid-row grid-gap">
+                <section className="tablet:grid-col-4 margin-bottom-6">
+                    <GeneratedSideNav directories={directories} />
+                </section>
+                <section className="tablet:grid-col-8 usa-prose rs-documentation">
+                    <Routes>
+                        <Route
+                            path={"/"}
+                            element={
+                                <Navigate to={"/product/overview"} replace />
```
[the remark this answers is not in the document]
kevinhaube: Side navigation templates include their own router because of how side navigation component logic works.

frontend-react/src/pages/admin/value-set-editor/ValueSetsIndex.tsx
kevinhaube: Sounds like a good tech-debt item, but out of scope here

frontend-react/src/pages/Login.tsx
doug-s-nava: bumping this comment - should we make a new card for re-implementing this?

frontend-react/src/pages/admin/value-set-editor/ValueSetsIndex.tsx
doug-s-nava: sure - can you add the comment back then?

doug-s-nava (commented):
a couple outstanding nits, and I'm still seeing the two bugs I mentioned in the last review. Any thoughts on those?

kevinhaube:
@doug-s-nava The first bug is probably your end. The page loads fine for me. The second bug is ticketed and included in the PR. #6694

frontend-react/src/pages/admin/value-set-editor/ValueSetsIndex.tsx
kevinhaube: Didn't even see it had moved. It was next to an import in my branch. 🤷🏼

frontend-react/src/pages/Login.tsx
kevinhaube: This was added to hopefully prevent someone who's logged in from landing on `/login` but breaks the okta redirect handler, so I removed it.

doug-s-nava:
Tracked down the value sets detail bug to here - remove this line and things should work https://github.com/CDCgov/prime-reportstream/blob/25db9a7e1201e8d483c382fa5fa37de1ff43694a/frontend-react/src/pages/admin/value-set-editor/ValueSetsIndex.tsx#L48

kevinhaube:
Yeah, think that was a bug I'd fixed somewhere along the way and then through merging the multiple PRs to one it got overwritten by a PR that I _didn't_ fix it on

frontend-react/src/components/AuthElement.tsx
```diff
@@ -0,0 +1,58 @@
...
+    const authorizeMemberType = useMemo(() => {
+        if (memberType === MemberType.PRIME_ADMIN) return true; // Admin authentication always
+        return requiredUserType instanceof Array
+            ? requiredUserType.includes(memberType)
+            : requiredUserType === memberType;
+    }, [requiredUserType, memberType]);
+    // All the checks before returning the route
+    useEffect(() => {
+        if (!oktaToken?.accessToken || !activeMembership) {
+            navigate("/login");
```
doug-s-nava, line 38: I think the refresh bug we're seeing is due to token and membership being briefly undefined when the app starts up. Not sure what the best way to handle that might be. We could be really dumb and track the initial load here or within useOktaMemberships, and wait until the second render to do anything? Ok I see why you want to tackle this in a separate ticket. Maybe worth just commenting here until we figure it out?

frontend-react/src/components/Content/Templates/IASideNavTemplate.tsx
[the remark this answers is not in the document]
doug-s-nava: yeah, I'm mainly interested in getting rid of the hardcoded redirect here, and it seems like replacing this with an IARouter would provide us with the ability to make that dynamic. If you'd like to tackle that later, can we make sure that we card it so that it doesn't surprise us when we want to use this template for something other than a product page?

doug-s-nava (commented):
Looking good

I'm a little wary of approving this with the regression on refresh but if everyone is ok with tackling this as a follow up I'm good to approve, as the rest of the remaining comments are just nits https://github.com/CDCgov/prime-reportstream/issues/6694 @bgantick @TomNUSDS any thoughts?

frontend-react/src/components/Content/Templates/IASideNavTemplate.tsx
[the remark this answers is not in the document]
kevinhaube: This is a pre-index IA template, and the only one we needed for the MVP. When we did updates a few sprints ago, we had an added requirement of index pages. This template doesn't fit that model, so the IARouter that was built for that specific need doesn't fit into the way this template renders. An index is its own entire element, so for this template to adapt the new IARouter, rather than state switching to show new content within a single element, we'd have to have a unique element to render for _each section_, included with its own uniquely rendered sidenav. If we want to refactor it outside of the scope of adopting react-router we can ticket it.

frontend-react/src/components/AuthElement.tsx
kevinhaube: > We could be really dumb and track the initial load

So far as I've tried, it's a non-solution for this behavior. Because our application doesn't require users to refresh for any expected behavior, I think it's safe if this is a bug left for follow-up work.

doug-s-nava (approved):
As long as we're planning to take care of https://github.com/CDCgov/prime-reportstream/issues/6694 as a fast follow on this (maybe aim to get it in before the Thursday deployment?), then I'm good to sign off on this.

Good job with a tough task here!

✅ ✅ ✅ ✅ ✅ ✅

frontend-react/src/components/AuthElement.tsx
```diff
@@ -22,7 +22,7 @@ export const AuthElement = ({
     const { oktaToken, activeMembership } = useSessionContext();
     // A way to check if this is a logged-in user refreshing the app
     const tokenAvailable = useMemo(
-        () => getStoredOktaToken() === undefined || getStoredOktaToken() === "",
+        () => getStoredOktaToken() !== undefined && getStoredOktaToken() !== "",
```
doug-s-nava, line 25: why isn't null included here?

doug-s-nava (commented):
one quick nit

frontend-react/src/components/AuthElement.tsx
kevinhaube: I don't believe it can return `null`. The actual sessionStorage API gives undefined if no key/value pair is matched, and we have a layer atop that and it sometimes, for some reason not all the time (noted by it's possibly undefined return value) makes it an empty string inside of the `getStoredOktaToken` function, so either response of a token missing should be covered in this. 
Is there a benefit of using `null` vs `undefined` for checks like these? It seems most JS APIs and libraries lean into undefined, but I grew up with java so null makes sense...is `undefined === null`?

doug-s-nava: good question, because javascript is very whack on this point. apologies if I'm saying stuff you already know:
```
undefined !== null
undefined == null

undefined !== ""
undefined == ""
```
`undefined`, `null` and `""` (and `NaN` and `0` and `false` for that matter, but no need to drag them into this) are all FALSEY, which means they are all `!<value> ===  false, !!<value> === true`

In a case like this, based on what you're saying, I think we're safe to say 
```suggestion
        () => !!getStoredOktaToken(),
```

sethdarragile6:
@kevinhaube I pulled down the branch, reup'd and rebuilt, did a `yarn install` then a `yarn start:localdev`, logged in successfully.

First unusual thing is that the same non-logged in homepage content is still there instead of the expected Org Admin page
Second unusual thing is that when you click on the org picker (top right), it seems to flash the content of the Org Admin page, but is replaced immediately by the non-logged in homepage content.
In a nutshell- can't get to the Org Admin page. Even if I type directly in the URL bar "localhost:3000/admin/settings" it immediately redirects back to "localhost:3000"

image

sethdarragile6:
> @kevinhaube I pulled down the branch, reup'd and rebuilt, did a `yarn install` then a `yarn start:localdev`, logged in successfully.
> 
> First unusual thing is that the same non-logged in homepage content is still there instead of the expected Org Admin page Second unusual thing is that when you click on the org picker (top right), it seems to flash the content of the Org Admin page, but is replaced immediately by the non-logged in homepage content. In a nutshell- can't get to the Org Admin page. Even if I type directly in the URL bar "localhost:3000/admin/settings" it immediately redirects back to "localhost:3000"
> 
> image

This was related to my okta configuration- not a code issue.

sethdarragile6 (approved):
I pulled this down and clicked around a lot- all looks good!

TomNUSDS (approved):
Approved! :rubber_stamp:

PR #6647, opened 2022-09-06
Experience/shadow master react query

doug-s-nava:
This PR gathers all previously approved changes for the MVP of the new react-query based approach to data fetching, as implemented on the value sets pages.

## Test Steps:
value sets regression testing. see https://github.com/CDCgov/prime-reportstream/pull/6476

## Changes
For more information on individual changes, see:

* https://github.com/CDCgov/prime-reportstream/pull/6616
* https://github.com/CDCgov/prime-reportstream/pull/6552
* https://github.com/CDCgov/prime-reportstream/pull/6477
* https://github.com/CDCgov/prime-reportstream/pull/6476

### Testing
- [x] Tested locally?
- [ ] Ran `./prime test` or `./gradlew testSmoke` against local Docker ReportStream container?
- [x] (For Changes to /frontend-react/...) Ran `npm run lint:write`? 
- [x] Added tests?

### Process
- [ ] Are there licensing issues with any new dependencies introduced?
- [x] Includes a summary of what a code reviewer should test/verify?
- [ ] Updated the release notes?
- [ ] Database changes are submitted as a separate PR?
- [ ] DevOps team has been notified if PR requires ops support?

## Linked Issues
- Fixes https://app.zenhub.com/workspaces/experience-607d9d5e68b95200150fec37/issues/cdcgov/prime-reportstream/6392

## To Be Done
- Search the zenhub board for `fetch refactor` https://app.zenhub.com/workspaces/experience-607d9d5e68b95200150fec37/board

## Specific Security-related subjects a reviewer should pay specific attention to
- Does this PR introduce new endpoints?
    - new endpoint A
    - new endpoint B
- Does this PR include changes in authentication and/or authorization of existing endpoints?
- Does this change introduce new dependencies that need vetting?
- Does this change require changes to our infrastructure?
- Does logging contain sensitive data?
- Does this PR include or remove any sensitive information itself?

If you answered '_yes_' to any of the questions above, conduct a detailed Review that addresses at least:

- What are the potential security threats and mitigations? Please list the _STRIDE_ threats and how they are mitigated
    - **S**poofing (faking authenticity)
        - Threat _T_, which could be achieved by _A_, is mitigated by _M_
    - **T**ampering (influence or sabotage the integrity of information, data, or system)
    - **R**epudiation (the ability to dispute the origin or originator of an action)
    - **I**nformation disclosure (data made available to entities who should not have it)
    - **D**enial of service (make a resource unavailable)
[...]

PR #6648, opened 2022-09-06
docs: new request patterns [experience-6392]

doug-s-nava:
updates and adds documentation to cover the changes made in https://github.com/CDCgov/prime-reportstream/pull/6647

## Test Steps:
1. review documentation for completeness, grokability, etc.


frontend-react/docs/data-fetching-patterns.md
```diff
@@ -0,0 +1,312 @@
...
+A simplified example taken from [ValueSetsIndex.tsx](https://github.com/CDCgov/prime-reportstream/blob/58599e355ad2a0b357005190c185a8b8497f0b81/frontend-react/src/pages/admin/value-set-editor/ValueSetsIndex.tsx)
+
+```typescript
+import { useValueSetsTable } from "../../../hooks/UseValueSets";
+const ValueSetsTable = () => {
+    const { valueSetArray, isLoading, error } = useValueSetsTable<ValueSet[]>(
+        LookupTables.VALUE_SET
+    );
+    if (isLoading) {
+    	// may use suspense, but if not, you can render a loading state here
```
kevinhaube, line 40: This seems like it encourages the pattern we chose not to implement for loading and error state. Maybe move the example use of `isLoading` and `error` into a separate section?

frontend-react/docs/rs-react-testing-network-calls.md
```diff
@@ -38,77 +39,116 @@ const handlers = [
...
+    }),
+];
+
+const failureHandlers = [
+    rest.get(tableListUrl, (req, res, ctx) => {
+        return res(ctx.status(404));
+    }),
+];
+
+export const successServer = setupServer(...successHandlers);
```
kevinhaube, line 83: You're able to conditionally return within the handler functions. Since each of these requires its own startup/teardown when testing, should we keep them combined as much as possible?

kevinhaube (changes requested):

frontend-react/docs/data-fetching-patterns.md
doug-s-nava: I left the comments here as a note that these may not be the particular patterns for handling loading and error states, but more a demonstration of what the hook may return. I feel like we should revisit this once we set out the pattern for loading and error handling, and maybe write up a full separate document on that if it's warranted

frontend-react/docs/rs-react-testing-network-calls.md
doug-s-nava: Definitely. This documentation is present in the existing doc. https://github.com/CDCgov/prime-reportstream/pull/6648/files#diff-0b4dc3618e3953b315ccddbf2b2536d54903bde6c8869a4d9a4b68d0fda228d6R147 It doesn't seem like all that common of a pattern, but if you do want to test different behaviors from exactly the same request within a given test file, it's an interesting approach. I covered the distinction between this approach and the approach of conditional handling within the callback here https://github.com/CDCgov/prime-reportstream/pull/6648/files#diff-0b4dc3618e3953b315ccddbf2b2536d54903bde6c8869a4d9a4b68d0fda228d6R89 but I could do a separate example if you think that's useful. I generally erred on the side of reducing churn on this document while making sure it was up to date.

kevinhaube (approved):

PR #6653, opened 2022-09-07
Generate SPM segments from FHIR

PatriciaPerozoUSDS:
This PR ...
- Adds the generation of the SPM segment
- Adds a "coded with exceptions" CWE HL7 mapping helper.

Test Steps:
1. Run the FHIR -> HL7 conversion command ```./gradlew primeCLI --args='fhirdata --input-file "src/testIntegration/resources/datatests/HL7_to_FHIR/sample_co_1_20220518-0001.fhir" -s metadata/hl7_mapping/ORU_R01/ORU_R01-base.yml'```
2. You will notice these differences:
- SPM.2.2.4 Doesn't include ISO . This follows the previous decision not to include ISO if the universal id is not included
- SPM.14 only includes RA, This is noted in the mapping and is due to the fact that HCA is supposed to be sending us text but instead is sending a CE
- SPM.26. In the sample file some numbers are ints while others are doubles. The output of the generated HL7 is always a double.

## Changes
- Adds the generation of the SPM segment
- Adds a "coded with exceptions" CWE HL7 mapping helper.
- Adds a None concept to the CodingSystemMapper so throw an exception if there is no mapping.

### Testing
- [X] Tested locally?
- [ ] Added tests?

## Linked Issues
- #5755


prime-router/metadata/hl7_mapping/ORU_R01/base/order-observation.yml
```diff
@@ -2,10 +2,6 @@ name: ORU-R01-order-observation
 constants:
   hl7Order: /PATIENT_RESULT/ORDER_OBSERVATION(%{orderIndex})
 elements:
-  - name: request-set-ID
```
PatriciaPerozoUSDS, line 5: This was also duplicated in the observation-request schema. It seems more appropriate to stay there.

prime-router/metadata/hl7_mapping/ORU_R01/base/coded-with-exceptions.yml
```diff
@@ -0,0 +1,18 @@
+name: ORU-R01-coded-with-exceptions
+elements:
+  - name: coded-element
+    constants:
+      ceFieldPath: '%{cweFieldPath}'
+    schema: coded-element
+
+  - name: coding-system-version-id
+    resource: '%resource.extension(%`rsext-alternate-identifier`).value.coding.version'
+    condition: '%resource.exists()'
```
carlosfelix2, line 10: Remove this condition.  It is already checked by value

prime-router/metadata/hl7_mapping/ORU_R01/base/coded-with-exceptions.yml
```diff
@@ -0,0 +1,18 @@
...
+    value: [ '%resource']
+    hl7Spec: [ '%{cweFieldPath}-7' ]
+
+  - name: coding-system-version-id
+    resource: '%resource.extension(%`rsext-alternate-identifier`).value.coding.extension(%`rsext-alternate-version`)'
+    condition: '%resource.exists()'
```
carlosfelix2, line 16: Remove this condition. It is already checked by value

carlosfelix2 (changes requested):
Per my test, you are missing SPM-14-2 and SPM-14-3

carlosfelix2:
> Per my test, you are missing SPM-14-2 and SPM-14-3

NM.  You are right.  The type is incorrect from the original data.

carlosfelix2 (dismissed):

PR #6655, opened 2022-09-07
Lastmile fix layout

TomNUSDS:
Fixing some layout data issues that showed up when using staging data.

Still need to track down why hl7 Receiver column is empty, but that will be a different PR.


kevinhaube (approved):
👍🏼

PR #6679, opened 2022-09-09
Allow FHIR to HL7 library schema constants to support FHIR Path expressions

carlosfelix2:
This PR adds the ability to the FHIR to HL7 library schema constants to support FHIR Path expressions.  This change works similarly to how the constants were working before.  

Constants for hl7Spec (strings) are resolved when the element that uses it is being processed.  Any constants used by the constants are also resolved at that time.

For FHIR Path, this new change allows you to specify a FHIR Path expression in the constant, but that expression remains a string until the time the element is processed.  At that time, the string within the constant is resolved, then the result is used in the FHIR Path expression for the element's property.  Just like with string constants, you have to be careful what other constants you use as they will be resolve when the FHIR Path expression for an element property is processed.  E.g., if you use `%resource` make sure that value is correct at the element using the constant.  For this reason is best not to use %resource at this level.

Test Steps:
1. Build this branch.
2. Run `./prime fhirdata --input-file src/testIntegration/resources/datatests/HL7_to_FHIR/sample_co_1_20220518-0001.fhir -s metadata/hl7_mapping/ORU_R01/ORU_R01-base.yml`
3. Verify that the HL7 file was generated and the data looks good specially for the NTE segments which data is not hard coded

## Changes
- Allow FHIR to HL7 library schema constants to support FHIR Path expressions

### Testing
- [x] Tested locally?
- [x] Added tests?

### Process

## Linked Issues
- None


prime-router/metadata/hl7_mapping/ORU_R01/base/note.yml
```diff
@@ -5,33 +5,38 @@ elements:
     hl7Spec: [ '%{hl7NotePath}/NTE(%{noteIndex})-1' ]
 
   - name: note-source
-    value: [ '''L''' ]
+    value:
```
carlosfelix2, line 8: Note: These are the updates needed to get the NTE segments not be hard coded.

clediggins-usds (approved):
Works locally

PR #6769, opened 2022-09-19
Add support for Settings Revision History

### LINKED ISSUE #5187: [Feature] Setting Revision History
## Problem Statement:
As a member of the O&O team, I want to be able to view, compare and clone previous version of Sender, Receiver and full settings for an organization to troubleshoot issues.

## Acceptance Criteria:
- User is able to view any version of settings from history
- User is able to compare the current version with any previous version of settings
- User is able to clone a previous version and create a new version for any setting.

## What you need to know:
Background: All past Setting for an organization are only marked deleted (tombstoned). These older revisions contain a version number in the meta. A nice safety-net feature is to allow seeing past versions and comparing to recent to help track down issues.

API:
- [ ] Given an Org name and Setting type, return a full list of all past revisions.
`settings/revision/organizations/{organizationName}/{settingSelector}`

Frontend:
- [ ] Show History as a list.
- [ ] Use DiffEditor to show differences across two revisions AND across different named settings.
- [ ] Add mode to DiffEditor to NOT allow editing of text.

## Modifications to existing Settings page

Add link to every settings type. This will lead to a new Settings Revision History page. 

Note: This link is ALWAYS displayed even when a list might be empty (like Sender Settings) because there might be settings that were deleted.

## New Settings Revision History page
New page will contain two copies of the same lists side-by-side. The list will be a 2-level hierarchy with the top level being the setting name and the inner level being the revision. It will use [USWDS's accordion view](https://trussworks.github.io/react-uswds/?path=/docs/components-accordion--borderless).

Clicking on any item will fill in that side of the existing Text Difference Compare component and that will highlight the differences.

No editing will be supported. We might have a Back button somewhere at the top of the page.

The new page's url will include the org name and settings type so it can be linkable.

TomNUSDS:
Client and server changes to support displaying and comparing changes to Settings.

Fixes #5187

Test Steps:
1. See issue above for steps on how to test UI.
2. Unit tests are in place.

## Changes
- Adds a new API to get the settings history
- Adds a new client page to display this data. (See #5187 for screen shots)

### Testing
- [x] Tested locally?
- [x] Added tests?

### Process
- [x] Are there licensing issues with any new dependencies introduced? No
- [x] Database changes are submitted as a separate PR? `n/a`
- [x] DevOps team has been notified if PR requires ops support? `n/a`

## Linked Issues
- Fixes #5187

- [x] Does this PR introduce new endpoints?
    - new endpoint `/waters/org/{organizationName}/senders/revs/{settingType}`
- [ ] Does this PR include changes in authentication and/or authorization of existing endpoints?
- [ ] Does this change introduce new dependencies that need vetting?
- [ ] Does this change require changes to our infrastructure?
- [ ] Does this PR include or remove any sensitive information itself?


frontend-react/src/network/api/Organizations/SettingRevisions.ts
```diff
@@ -0,0 +1,34 @@
+import { API } from "../NewApi";
```
TomNUSDS, line 1: @doug-s-nava This was my start at doing the API the new way. But you'll see I added `OrgSettingRevisionsResorce.ts` to get the rest of the code working first.

doug-s-nava: ahah! NewApi is one of the old ways. Take a look at https://github.com/CDCgov/prime-reportstream/blob/44b7a1e0d21fa8e905e814887d4b249326e18ef5/frontend-react/src/pages/admin/value-set-editor/ValueSetsDetail.tsx#L214 for an example of the new way, or docs here https://github.com/CDCgov/prime-reportstream/blob/master/frontend-react/docs/data-fetching-patterns.md. Happy to pair whenever there's a good moment

prime-router/src/main/kotlin/azure/DatabaseAccess.kt
```diff
@@ -642,6 +642,95 @@ class DatabaseAccess(private val create: DSLContext) : Logging {
             .into(Setting::class.java)
     }
 
+    /** data returned by fetchSettingRevisionHistory. Only used to shape json response. **/
```
carlosfelix2, line 645: Use proper kdoc here with properties and description

prime-router/src/main/kotlin/azure/SettingsFacade.kt
```diff
@@ -156,6 +156,19 @@ class SettingsFacade(
         }
     }
 
+    /**
+     * Queries for the settings history for an org based on the type (RECEIVER, SENDER, ORG)
+     * The whole history (incl inactive and deleted) returned across all names for a given type.
+     * @param organizationName Restrict query to this org
+     * @param settingsType Type of setting. column in db and used to select
```
carlosfelix2, line 163: Missing @ return

prime-router/src/main/kotlin/azure/SettingsFunctions.kt
```diff
@@ -217,6 +218,46 @@ class UpdateReceiver(
...
+ *
+ * Type returned depends on the request settingSelector parameter.
+ * ALL named settings are return and the caller must group accordingly.
+ * Return ALL names solves the problem where knowing a deleted name become impossible
+ *
+ * From the OpenAPI view, this is just 3 different api calls
+ *   `settings/revision/organizations/{organizationName}/sender`
+ *   `settings/revision/organizations/{organizationName}/receiver`
+ *   `settings/revision/organizations/{organizationName}/organization`
+ */
```
carlosfelix2, line 234: Missing kdoc properties

prime-router/src/main/kotlin/azure/SettingsFunctions.kt
```diff
@@ -250,6 +291,25 @@ open class BaseFunction(
         }
     }
 
+    /**
+     * Returns all revisions of a settings (version) for an org/settingName
+     * Handles Auth
+     * @param request Incoming http request
+     * @param organizationName Org name to auth again and use for query
+     * @param settingType SettingType
+     * @HttpResponseMessage Resulting json or HTTP error response
```
carlosfelix2, line 300: Missing @ return.

prime-router/src/main/kotlin/azure/DatabaseAccess.kt
```diff
@@ -642,6 +642,95 @@ class DatabaseAccess(private val create: DSLContext) : Logging {
...
+        val selectCols = DSL.using(txn).select(
+            settings.SETTING_ID.`as`("id"),
+            settings.NAME,
+            settings.VERSION,
+            settings.CREATED_AT,
+            settings.CREATED_BY,
+            settings.IS_ACTIVE,
+            settings.IS_DELETED,
+            settings.VALUES.cast(String::class.java).`as`("settingJson")
+        ) // see kdoc
```
carlosfelix2, line 699: This comment did not make sense to me.  I did not know what to look for in the kdoc.  Add the comments you want in the code.

prime-router/src/main/kotlin/azure/DatabaseAccess.kt
```diff
@@ -642,6 +642,95 @@ class DatabaseAccess(private val create: DSLContext) : Logging {
...
+
+        when (settingType) {
+            SettingType.ORGANIZATION ->
+                return selectCols
+                    .from(settings)
+                    .where(
+                        settings.TYPE.eq(SettingType.ORGANIZATION)
+                            .and(settings.TYPE.eq(SettingType.ORGANIZATION))
```
carlosfelix2, line 707: Why is settingType here twice?

prime-router/src/main/kotlin/azure/SettingsFacade.kt
```diff
@@ -156,6 +156,19 @@ class SettingsFacade(
...
+     */
+    fun findSettingHistoryAsJson(organizationName: String, settingsType: SettingType): String {
+        val settings = db.transactReturning { txn ->
+            db.fetchSettingRevisionHistory(organizationName, settingsType, txn)
+        }
+        return if (settings.isNotEmpty()) mapper.writeValueAsString(settings) else "[]"
```
carlosfelix2, line 169: Instead of "[]" use a Json library to write an empty array

prime-router/src/main/kotlin/azure/DatabaseAccess.kt
```diff
@@ -642,6 +642,95 @@ class DatabaseAccess(private val create: DSLContext) : Logging {
             .into(Setting::class.java)
     }
 
+    /** data returned by fetchSettingRevisionHistory. Only used to shape json response. **/
+    data class SettingsHistoryData(
```
carlosfelix2, line 646: This data class looks out of place here.  Find where the other classes are and put them in the same area and level.

prime-router/src/main/kotlin/azure/DatabaseAccess.kt
```diff
@@ -642,6 +642,95 @@ class DatabaseAccess(private val create: DSLContext) : Logging {
...
+
+            // Sending/receiver needs join with the latest organization setting record
+            // in order to restrict the organization
+            SettingType.SENDER, SettingType.RECEIVER ->
+                return selectCols
+                    .from(settings)
+                    .join(org)
+                    .on(settings.ORGANIZATION_ID.eq(org.SETTING_ID))
+                    .where(
+                        settings.TYPE.eq(settingType)
```
carlosfelix2, line 722: I think type is used twice here too

prime-router/src/main/kotlin/azure/DatabaseAccess.kt
```diff
@@ -642,6 +642,95 @@ class DatabaseAccess(private val create: DSLContext) : Logging {
...
+                            .and(org.IS_ACTIVE.isTrue)
+                            .and(org.TYPE.eq(SettingType.ORGANIZATION))
+                            .and(org.ORGANIZATION_ID.isNull)
+                            .and(org.NAME.eq(organizationName))
+                    )
+                    .limit(MAX_RECORDS_TO_RETURN)
+                    .fetch()
+                    .into(SettingsHistoryData::class.java)
+        }
```
carlosfelix2, line 731: Check your linter in IntelliJ.  It should be complaining you are missing an else for the when.  You need to add it.

prime-router/src/main/kotlin/azure/DatabaseAccess.kt
```diff
@@ -642,6 +642,95 @@ class DatabaseAccess(private val create: DSLContext) : Logging {
...
+            SettingType.SENDER, SettingType.RECEIVER ->
+                return selectCols
+                    .from(settings)
+                    .join(org)
+                    .on(settings.ORGANIZATION_ID.eq(org.SETTING_ID))
+                    .where(
+                        settings.TYPE.eq(settingType)
+                            .and(org.IS_ACTIVE.isTrue)
+                            .and(org.TYPE.eq(SettingType.ORGANIZATION))
+                            .and(org.ORGANIZATION_ID.isNull)
```
carlosfelix2, line 725: Why do you check is organization is null?  Wouldn't the type check give you what you need?  Seems redundant.

prime-router/src/main/kotlin/azure/SettingsFacade.kt
TomNUSDS: Seems heavyweight since it just generates "[]", but will do.

prime-router/src/main/kotlin/azure/DatabaseAccess.kt
TomNUSDS: Will do.

Note: This location was intentional but maybe it's not a standard java style. If I could have scoped it to just the function doing the json serialization I would, but Kotlin/Java doesn't seem to support defining data classes within function scopes. It should definitely not be in the global scope since it's ONLY use is to serialize the result of the query.

prime-router/src/main/kotlin/azure/DatabaseAccess.kt
TomNUSDS: It's used in similar queries, so I assumed it was required. (search file for `ORGANIZATION_ID.isNull`). I will test out removing it.

prime-router/src/main/kotlin/azure/DatabaseAccess.kt
TomNUSDS: It's an enumerator of `SettingType` and all cases are listed in the `when`, so no `else` should be required? I think I'm running the latest kotlin `222-1.7.10-release-334-IJ3739.54`?

prime-router/src/main/kotlin/azure/DatabaseAccess.kt
TomNUSDS: OK. I've looked around and cannot figure out where the best place for this is. All the other types in the DatabaseAccess file seem to come out of pojo's auto generating types from schemas.

I see a similar style used with `AuthClaims` where the data class is defined right next to the function that uses it.

I'm not asserting some strong style opinion, I literally just don't know the best place for it so I can resolve this issue.

Any suggestions?

prime-router/src/main/kotlin/azure/DatabaseAccess.kt
TomNUSDS: Good catch. This comment was meant for the line above it to explain why the query was casting the JSONB to a string.

prime-router/src/main/kotlin/azure/DatabaseAccess.kt
TomNUSDS: Seems like a mistake. Good catch. Fixed.

prime-router/src/main/kotlin/azure/DatabaseAccess.kt
TomNUSDS: This one I think is fine? It's a self join. `org.TYPE` vs `settings.TYPE`

prime-router/src/main/kotlin/azure/SettingsFacade.kt
TomNUSDS: Just doing a `return mapper.writeValueAsString(settings)` does the right thing and returns `[ ]` and is easier to read so 👍

prime-router/src/main/kotlin/azure/DatabaseAccess.kt
TomNUSDS: Good call. I grabbed the query it was generating and tested removing that condition and it wasn't required, so I removed it.

frontend-react/src/AppRouter.tsx
```diff
@@ -85,6 +86,10 @@ export const AppRouter = () => {
                 path={"/admin/value-sets"}
                 element={<ValueSetsIndexWithAuth />}
             />
+            <Route
+                path="/admin/revisionhistory/org/:org/settingtype/:settingType"
```
doug-s-nava, line 90: [nit] we've already got a route for `/admin/orgnewsetting/org/:orgname/settingtype/:settingtype"` - resistance to using `:orgname` as our segment variable here for consistency?

frontend-react/src/components/StaticCompare.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+import { Diff, SES_TYPE } from "../utils/diff";
+import { splitOn } from "../utils/misc";
+
+interface StaticCompareProps {
+    rightText: string;
+    leftText: string;
+}
+
+const insertHighlight = (
```
doug-s-nava, line 13: can we get a jsdoc here? i know it's not exported, but it's somewhat complex and the arg names are not super illuminating as to what's being processed

frontend-react/src/components/StaticCompare.tsx
```diff
@@ -0,0 +1,143 @@
...
+    s1: string,
+    offset: number,
+    length: number
+): string => {
+    if (s1 === "" || length === 0) {
+        return "";
+    }
+    // we want to insert a <span></span> around text.
+    const three_parts = splitOn(s1, offset, offset + length);
```
doug-s-nava, line 22: [nit] `threeParts` - imo snake case has no place in js variable naming outside of all caps constants

frontend-react/src/components/StaticCompare.tsx
```diff
@@ -0,0 +1,143 @@
...
+        },
+        [leftHighlightHtml, rightHighlightHtml]
+    );
+
+    useEffect(() => {
+        if (props.leftText?.length > 0 && props.rightText?.length > 0) {
+            // initialization only
+            refreshHighlights(props.leftText, props.rightText);
+        }
+    }, [props, refreshHighlights]);
```
doug-s-nava, line 85: would be more accurate to depend on `props.leftText?.length ` and same for right - otherwise this is running whenever any props change. upon writing this, this is a nit, since the length will change inherently. still, I favor this, since dependency checking on objects is always a bit of a black box

frontend-react/src/components/StaticCompare.tsx
```diff
@@ -0,0 +1,143 @@
+import { ReactElement, useCallback, useEffect, useState } from "react";
+import DOMPurify from "dompurify";
+import { ScrollSync, ScrollSyncPane } from "react-scroll-sync";
+
+import { Diff, SES_TYPE } from "../utils/diff";
```
doug-s-nava, line 5: out of scope since this is not a new thing, but why is `Diff` in caps here?

frontend-react/src/components/StaticCompare.tsx
```diff
@@ -0,0 +1,143 @@
...
+            differ.compose();
+            const sesses = differ.getses();
+
+            // because we're modifying text, it will change offsets UNLESS we go backwards.
+            // the later items in the patches array are sequential
+            let patchedLeftStr = rightText;
+            let patchedRightStr = leftText;
+
+            for (let ii = sesses.length - 1; ii >= 0; --ii) {
+                const eachses = sesses[ii];
```
doug-s-nava, line 51: `eachSes`

frontend-react/src/components/StaticCompare.tsx
```diff
@@ -0,0 +1,143 @@
...
+            const differ = Diff(rightText, leftText);
+            differ.compose();
+            const sesses = differ.getses();
+
+            // because we're modifying text, it will change offsets UNLESS we go backwards.
+            // the later items in the patches array are sequential
+            let patchedLeftStr = rightText;
+            let patchedRightStr = leftText;
+
+            for (let ii = sesses.length - 1; ii >= 0; --ii) {
```
doug-s-nava, line 50: what does `ii` stand for? I am removing the task of fully understanding the differ, but ii to me would denote a sub index to an already existing `i` which I don't see. Can this be more clear?

frontend-react/src/components/StaticCompare.tsx
```diff
@@ -0,0 +1,143 @@
...
+                {/*Text has two components. The text on top and the highlight under it.*/}
+                {/*scrolling is synced so they line up*/}
+
+                <div className="rs-editable-stacked-container">
+                    <ScrollSyncPane>
+                        <div
+                            className="rs-static-compare-base rs-editable-compare-static"
+                            data-testid={"left-compare-text"}
+                            contentEditable={false}
+                            dangerouslySetInnerHTML={{
```
doug-s-nava, line 99: why can't we set this as a chlid?

frontend-react/src/hooks/UseCreateFetch.ts
```diff
@@ -40,7 +40,12 @@ function createTypeWrapperForAuthorizedFetch(
         // this system assumes that we want to be making authenticated
         // requests whenever possible
         if (!headers.authorization || headers.authorization.length < 8) {
-            console.warn("Unauthenticated request to ", options.url);
+            console.warn(
```
doug-s-nava, line 43: fine to keep this change if it's still helpful to you, but maybe format the whole thing as a template string rather than mixing template string with argument based log calls. Or you know, pick one implementation and go with it, either way

frontend-react/src/pages/admin/AdminRevHistory.tsx
```diff
@@ -0,0 +1,273 @@
...
+    useSettingRevisionEndpointsQuery,
+} from "../../network/api/Organizations/SettingRevisions";
+import Spinner from "../../components/Spinner";
+import { AuthElement } from "../../components/AuthElement";
+import { MemberType } from "../../hooks/UseOktaMemberships";
+import { jsonSortReplacer } from "../../utils/JsonSortReplacer";
+import { formatDate } from "../../utils/misc";
+import { StaticCompare } from "../../components/StaticCompare";
+
+/** Accordion components need data in specific format **/
```
doug-s-nava, line 20: can you link to trusswork docs?

frontend-react/src/pages/admin/AdminRevHistory.tsx
```diff
@@ -0,0 +1,273 @@
...
+        data: SettingRevision
+    ) => void;
+    data: SettingRevision[];
+}): AccordionItemProps[] => {
+    const results: AccordionItemProps[] = [];
+    if (props.data.length === 0) {
+        return [];
+    }
+
+    // should come back sorted by name and version from server. Sort by name, then by version
```
doug-s-nava, line 36: if they come back sorted why are we sorting again?

frontend-react/src/pages/admin/AdminRevHistory.tsx
```diff
@@ -0,0 +1,273 @@
...
+    props.data.sort((a, b) =>
+        a.name === b.name ? a.version - b.version : a.name.localeCompare(b.name)
+    );
+
+    // group the data to make generating the prop content logic easier to follow
+    const grouped: { [key: string]: SettingRevision[] } = {};
+    props.data.forEach((item) => {
```
doug-s-nava, line 43: would prefer a reducer implementation

frontend-react/src/pages/admin/AdminRevHistory.tsx
```diff
@@ -0,0 +1,273 @@
...
+        const name = item.name;
+        if (!grouped[name]) {
+            // key to array map. create initial entry for this group if needed
+            grouped[name] = [];
+        }
+        grouped[name].push(item);
+    });
+
+    // turn each group into a html list and add to the content of the accordion
+    for (let [key, settings] of Object.entries(grouped)) {
```
doug-s-nava, line 53: can this be a reducer too? personal bias here, but feel like `for of` loops without a compelling reason are an antipattern

frontend-react/src/pages/admin/AdminRevHistory.tsx
```diff
@@ -0,0 +1,273 @@
...
+    }
+    return results;
+};
+
+/** this extends SettingRevisionsParams so it can be passed to down without having to be recomposed, these include
+ * the org and settingType cgi params */
+interface MainComponentProps extends SettingRevisionParams {
+    leftSelectedListItem: string;
+    rightSelectedListItem: string;
+    onClickHandler: (
```
doug-s-nava, line 96: this type is also used above, can you split it into its own type?

frontend-react/src/pages/admin/AdminRevHistory.tsx
```diff
@@ -0,0 +1,273 @@
...
+/** main page, not exported here because it should only be loaded via AdminRevHistoryWithAuth() **/
+const AdminRevHistory = () => {
+    const { org, settingType } = useParams<SettingRevisionParams>(); // props past to page via the route/url path args
+    const [leftJson, setLeftJson] = useState("");
+    const [rightJson, setRightJson] = useState("");
+    // used to highlight which item is selected.
+    const [leftSelectedListItem, setLeftSelectedListItem] = useState("");
+    const [rightSelectedListItem, setRightSelectedListItem] = useState("");
+    const [leftItem, setLeftItem] = useState<SettingRevision | null>(null);
+    const [rightItem, setRightItem] = useState<SettingRevision | null>(null);
```
doug-s-nava, line 160: great candidate for a useReducer...

frontend-react/src/pages/admin/AdminRevHistory.tsx
```diff
@@ -0,0 +1,273 @@
...
+
+    const onClickHandler = useCallback(
+        (key: string, itemClickedKey: string, data: SettingRevision) => {
+            const normalizeJson = (jsonStr: string): string =>
```
doug-s-nava, line 164: remove this from render / use callback, maybe into util file

frontend-react/src/pages/admin/AdminRevHistory.tsx
```diff
@@ -0,0 +1,273 @@
...
+                    setRightItem(data);
+                    break;
+            }
+        },
+        []
+    );
+
+    return (
+        <>
+            <Helmet>
```
doug-s-nava, line 186: use <BasicHelmet>

frontend-react/src/pages/admin/AdminRevHistory.tsx
```diff
@@ -0,0 +1,273 @@
...
+                                            {String(rightItem.isActive)}
+                                        </>
+                                    )}
+                                </Grid>
+                            </Grid>
+                        </Grid>
+                    </Grid>
+                </GridContainer>
+            </section>
+            <HipaaNotice />
```
doug-s-nava, line 254: ???

doug-s-nava (commented):
sorry if this is rough / gruff - delivered under time pressure. Do you have test steps? haven't loaded it up yet

frontend-react/src/AppRouter.tsx
TomNUSDS: I switched it because you said you didn't like allsmallcaps. I am neutral on naming, so happy to switch it bac,.

frontend-react/src/components/StaticCompare.tsx
TomNUSDS: Will do. My next PR (which will be ready later today) reworks this whole highlighting approach to better deal with structured json versus unstructured text.

frontend-react/src/components/StaticCompare.tsx
TomNUSDS: Good call.

Is there a big difference between `props.leftText?.length` and `props.leftText`?

frontend-react/src/components/StaticCompare.tsx
TomNUSDS: This was a library I ported to typescript. It may be historical. I still need to contribute back the port to the source project.

We _really_ need our linter to enforce cap conventions.

frontend-react/src/pages/admin/AdminRevHistory.tsx
TomNUSDS: You're one step ahead of me. The next PR does this, so I'm going to mark this resolved.

frontend-react/src/pages/admin/AdminRevHistory.tsx
TomNUSDS: Great catch. This PR is taking so long, things are changing underneath it.

frontend-react/src/pages/admin/AdminRevHistory.tsx
TomNUSDS: `<HipaaNotice />` seems to be on every page. I'm not sure how to fix this comment

frontend-react/src/pages/admin/AdminRevHistory.tsx
TomNUSDS: Agreed. But can I do that in another PR so to get this one merged it?

frontend-react/src/AppRouter.tsx
TomNUSDS: OK. Switching it has a ripple effect because it's also used in the API call. So, I'm asking to punt on this item. 

I just hated the mixture of `settingType` with `orgname` everywhere. I figured it was better to be locally consistent and eventually fix `orgname` to be `orgName` or `org`

frontend-react/src/pages/admin/AdminRevHistory.tsx
doug-s-nava: Yup!

frontend-react/src/components/StaticCompare.tsx
doug-s-nava: lol one can dream

frontend-react/src/AppRouter.tsx
doug-s-nava: yep, forgot about that earlier convo, we can punt

frontend-react/src/components/StaticCompare.tsx
TomNUSDS: I like that name better. The names came the ported code.

frontend-react/src/pages/admin/AdminRevHistory.tsx
doug-s-nava: interesting, never noticed that (as evidenced by the fact that any pages I've built don't have it). Wonder what the actual requirement is. If it really needs to be everywhere should we add this info to the footer? All of that is out of scope for now

frontend-react/src/components/StaticCompare.tsx
TomNUSDS: `ii` vs `i` is a C code style thing that's stuck with me I have to be careful to not use. I can change it.

frontend-react/src/components/StaticCompare.tsx
TomNUSDS: Done. FYI all this code is going away in my next PR.

frontend-react/src/components/StaticCompare.tsx
TomNUSDS: Getting the two divs to line up perfectly and then stay in sync while scrolling (via `ScrollSync`) is tricky. This was the only way I could get it to work.

frontend-react/src/hooks/UseCreateFetch.ts
TomNUSDS: Oh, this is very much intentional.

I'm leveraging `console`'s ability to format objects but making each a separate parameter to the call verses trying to stringify it all in javascript.

The debugger also allows dev to dig arbitrarily deep into objects:

image

image

frontend-react/src/pages/admin/AdminRevHistory.tsx
TomNUSDS: I tend to never trust server API results blindly (even if I control the client and server). Burned too many times with something seemingly sorted correctly, but then someone else comes along and removes the sort to speed up the query (or it happened to be sorted correctly but it's a natural sort and someone restores the DB and the order gets weird). 

It's not a lot of items, so I figured why not.

frontend-react/src/pages/admin/AdminRevHistory.tsx
TomNUSDS: Fair. I'll rewrite.

frontend-react/src/pages/admin/AdminRevHistory.tsx
TomNUSDS: `for (let [key, value] of Object.entries() {}` loops are an anti-pattern? I'd love to read more so I can stay up on the latest trends like this.

Personally, I tend to flip between array function (map, reduce, filter) operators and `for of` loops depending on complexity (or size) of work being done inside of them.

"reducers" can be fairly complex with their 4 params (e.g. what does previous mean on the first iteration... what's even being returned?)

There's also general confusing around how "reducers" work in React (`useReducer`) vs Native Js (`array.reduce`).

frontend-react/src/pages/admin/AdminRevHistory.tsx
TomNUSDS: Wait `array.reduce` or `useReducer`? There's non-constant logic above it so I assume you mean some array level operation?

doug-s-nava: yeah, I'm just talking about Array.reduce to avoid an unnecessary closure. feel free to tell me to bug off

frontend-react/src/pages/admin/AdminRevHistory.tsx
doug-s-nava: it's a personal bias, but it's backed up by some other folks who have thought about these things: https://stackoverflow.com/questions/49420891/why-should-foreach-be-preferred-over-regular-iterators

beyond the point made here about mutability and side effects there's also

* confusion between `for of` and `for in`
* the fact that the iterating item value in the `for of` loop is a new instance of a variable, even though the loop syntax seems like it would work otherwise

i tend to only use `for of` if I'm dealing with a non-array iterable, but again, that's just me. Without a style guide I'm just a guy with an opinion

frontend-react/src/pages/admin/AdminRevHistory.tsx
doug-s-nava: sound reasoning. the comment just threw me off

frontend-react/src/hooks/UseCreateFetch.ts
doug-s-nava: got it, cool.

frontend-react/src/pages/admin/AdminRevHistory.tsx
TomNUSDS: But I _like_ your opinion, I'm just trying to understand/learn.

frontend-react/src/components/StaticCompare.tsx
doug-s-nava: the difference would just be around the hypothetical situation where `leftText` changed but the length didn't. If you're only dependent on the length, and the value changes without changing the length, your callback won't run

frontend-react/src/components/StaticCompare.tsx
doug-s-nava: ooo I never wrote enough C to pick up on that. my blindspot is showing

frontend-react/src/pages/admin/AdminRevHistory.tsx
TomNUSDS: OK. This is a _common_ problem. I found a stock reducer that can be used to d a groupBy that was in typescript and added it to misc (with tests).

doug-s-nava (approved):
✅ ✅ ✅ ✅ ✅ 

a couple things on the design / ux side (is there a plan to run this by design for a second iteration?)

- the history link appears even for areas where there are no revisions
- the flags display beneath the differ is a little confusing, may want to think about a better way to display that info

frontend-react/src/pages/admin/AdminRevHistory.tsx
```diff
@@ -0,0 +1,264 @@
...
+} from "../../network/api/Organizations/SettingRevisions";
+import Spinner from "../../components/Spinner";
+import { AuthElement } from "../../components/AuthElement";
+import { MemberType } from "../../hooks/UseOktaMemberships";
+import { jsonSortReplacer } from "../../utils/JsonSortReplacer";
+import { formatDate, groupBy } from "../../utils/misc";
+import { StaticCompare } from "../../components/StaticCompare";
+import { BasicHelmet } from "../../components/header/BasicHelmet";
+
+type clickHandler = (
```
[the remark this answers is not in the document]
TomNUSDS: Naming is hard! I like `AccordionClickHandler` and will make that change.

TomNUSDS:
> a couple things on the design / ux side (is there a plan to run this by design for a second iteration?)
> 
> * the history link appears even for areas where there are no revisions
> * the flags display beneath the differ is a little confusing, may want to think about a better way to display that info

💯 Agreed.

Right now our users are only internal. Before we expose these pages to customers, we'd need a full UX round of cleanup and reorganization.

I think I put this in the comments, but History shows up even without any settings for a section because there _might_ be items that were deleted.

The flags look like an afterthought because they were. I noticed while writing unit tests that the `IsDeleted` looked wrong on some items, so decided to surfacing it. Wasn't sure the best place for it, so I just stuck it at the bottom for now.

PR #6795, opened 2022-09-21
Intrivo - ORC-22-4 Blank

### LINKED ISSUE #6780: Intrivo post-product issue: Error : Blank value for element ORC-22-4 (ordering_facility_state)
Intrivo is having some messages fail to send through based on what they expect is a blank state, zip code provided.
They're receiving this error "Error : Blank value for element ORC-22-4 (ordering_facility_state)"

I think this is because we handled this via patient state and not ordering facility state in ticket 5700. 

A/C
- confirm that the RADX MARS spec either requires or does not require ORC-22-4
- confirm if RS requires ORC-22-4 or if optional
- if it does not require and we should fill in based on zip code, create a new ticket to implement same logic from 5700 for ORC-22-4

jorg3lopez:
This PR addresses the issue where OTC data came without ORC-22-4 (ordering_facility_state) and this was causing our api filters to exclude these OTC messages, coming from intrivo. The solution was to use the zip code lookup mapper to display the state.

Test Steps:
1. Create fake data for intrivo
2. Blank out ORC-22-4
3. Fire up docker
4. reloadtables and reloadsettings
5. Send to api and verify generated files have the state in ORC-22-4
6. Repeat with fake data having ORC-22-4 populated

## Changes
- added new entry to hl7-ingest-covid-19.schema: `-name: ordering_facility_zip_lookup` and `-name: ordering_facility_state`

### Testing
- [x] Tested locally?
- [ ] Added tests?

## Linked Issues
- Fixes #6780

### DIFF
--- a/prime-router/metadata/schemas/HL7/hl7-ingest-covid-19.schema
+++ b/prime-router/metadata/schemas/HL7/hl7-ingest-covid-19.schema
@@ -80,3 +80,18 @@ elements:
     natFlatFileField: Patient_state
     hhsGuidanceField:
     hl7Field: PID-11-4
+
+  - name: ordering_facility_zip_code
+    type: POSTAL_CODE
+    natFlatFileField: Ordering_facility_zip_code
+    hhsGuidanceField:
+    hl7Field: ORC-22-5
+    documentation: The zip code of the facility which the test was ordered from
+
+  - name: ordering_facility_state
+    mapper: use(ordering_facility_state, ordering_facility_state_zip_lookup)
+    cardinality: ZERO_OR_ONE
+    natFlatFileField: Ordering_facility_state
+    hhsGuidanceField:
+    hl7Field: ORC-22-4
+    documentation: The state of the facility which the test was ordered from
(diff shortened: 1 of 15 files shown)


prime-router/metadata/schemas/HL7/hl7-ingest-covid-19.schema
```diff
@@ -80,3 +80,18 @@ elements:
     natFlatFileField: Patient_state
     hhsGuidanceField:
     hl7Field: PID-11-4
+
+  - name: ordering_facility_zip_code
```
jbiskie, line 84: I don't think this would work correctly?

The ordering_facility_state uses the `mapper: use(ordering_facility_state, ordering_facility_state_zip_lookup)` and `ordering_facility_state_zip_lookup` isn't defined. I think that needs to be where you have `ordering_facility_zip_code` defined.

jorg3lopez: Good catch. Some how using the paste and cotrol z did something. Thanks!

jbiskie (approved):

PR #6815, opened 2022-09-23
Update org file for  CA

jorg3lopez:
This PR updates one of the receivers for CA in the local org file, to match what is in prod.

Test Steps:
1. generate fake data
2. fire up docker
3. reloadsettings
4. send fake data to api
5. save the generated file
6. pull settings from prod env
7. copy and paste the prod env receiver settings to replace the settings for the branch. Do not replace the transport, leave that alone.
8. reloadsettings
9. send fake data to api
10. compare generated file with the other one, make sure they are the same

## Changes
- updated the prod feed for CA in the local org file

### DIFF
--- a/prime-router/settings/organizations.yml
+++ b/prime-router/settings/organizations.yml
@@ -1641,31 +1641,107 @@
         initialTime: 00:00
         timeZone: EASTERN
 
-- name: ca-dph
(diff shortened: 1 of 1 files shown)


brick-green (approved):
thanks for putting this in!

PR #6811, opened 2022-09-23
fix: do not make useQuery requests with uninitialized okta [experience-unticketed]

doug-s-nava:
Currently there exists the possibility of making a request using the function output by `useAuthorizedFetch` before the js okta library has fully initialized. This means that requests can go out without an auth headers if made quickly enough after page load.

This change fixes this bug by:
* adding an `initialized` flag to tell the `useAuthorizedFetch` internals whether okta is ready to be used yet
* based on the value of this flag, telling useQuery whether or not it should be `enabled` (if an `enabled: false` option is passed to a `useQuery` invocation the hook will not run)
* outputting from `useAuthorizedFetch` an `rsUseQuery` function that is basically just `useQuery` with the `initialized / enabled` state management mentioned above folded into it
* using `rsUseQuery` instead of `useQuery` in existing hooks

## Test Steps:

Value sets regression testing, see https://github.com/CDCgov/prime-reportstream/pull/6476

### Bonus:

Make sure Tom's PR works with this commit cherry picked, and `rsUseQuery` thrown in in the correct place https://github.com/CDCgov/prime-reportstream/pull/6769

## Changes
- *Include a comprehensive list of changes in this PR*
- *(For web UI changes) Include screenshots/video of changes*

### Testing
- [ ] Tested locally?
- [ ] Ran `./prime test` or `./gradlew testSmoke` against local Docker ReportStream container?
- [ ] (For Changes to /frontend-react/...) Ran `npm run lint:write`? 
- [ ] Added tests?

### Process
- [ ] Are there licensing issues with any new dependencies introduced?
- [ ] Includes a summary of what a code reviewer should test/verify?
- [ ] Updated the release notes?
- [ ] Database changes are submitted as a separate PR?
- [ ] DevOps team has been notified if PR requires ops support?

## Linked Issues
- Fixes #issue

## To Be Done
*Create GitHub issues to track the work remaining, if any*
- #issue

## Specific Security-related subjects a reviewer should pay specific attention to
- Does this PR introduce new endpoints?
    - new endpoint A
    - new endpoint B
- Does this PR include changes in authentication and/or authorization of existing endpoints?
- Does this change introduce new dependencies that need vetting?
- Does this change require changes to our infrastructure?
- Does logging contain sensitive data?
- Does this PR include or remove any sensitive information itself?

If you answered '_yes_' to any of the questions above, conduct a detailed Review that addresses at least:

- What are the potential security threats and mitigations? Please list the _STRIDE_ threats and how they are mitigated
[...]

PR #6719, opened 2022-09-13
Experience/cypress login 6692

doug-s-nava:
## Changes

* fixes a bug in the existing functioning cypress test to get it passing
* completes system for programatically logging in to ReportStream via Cypress to allow for testing of parts of the site behind auth wall
* adds a very basic test to validate that login system works

## Notes

* this will only work locally for now, a further card(s) should be created to handle getting this ready to run in CI
   * will we set up a container for this?
   * how will we source secrets?
* this has been tested with the `Cypress Nava` test user that I created in okta preview.
   * this user has been assigned to the Prime Admins group. If this is a security concern, we should discuss
   * if we want to test behavior specific to sender or receiver users, we will need to create new okta test users for those cases, and supply creds for each user to the test environment
* as you can see from test steps below, for local testing commands must be run with a bunch of env vars supplied. So far I've just been doing this when running the commands, but we could build (at least some of) these variables into .env files

## User / Secret Creation

1. create a new user in okta preview. The email address does not need to be real, but will be useful in case you need to reset a password or something.
2. SAVE THE PASSWORD YOU CHOOSE FOR LATER
3. assign user to report stream application
4. assign any groups that the user should have (DHSender, DHmd, DHPrimeAdmin, etc.)
5. log in to report stream (staging, or local) with the new user credentials
6. when asked to set up MFA, choose Google Authenticator
7. when asked to scan a QR code, click the link to use a secret instead
8. SAVE THIS SECRET FOR LATER
9. follow the directions to set up Google Authenticator on your device
10. you're all set, you can now run tests with this user, using the credentials and secret you've noted

## Test Steps:
1. obtain secrets from Doug
2. run `REACT_APP_OKTA_USERNAME=report_stream_tester@navapbc.com REACT_APP_OKTA_PASSWORD=<pw> REACT_APP_OKTA_SECRET=<secret> REACT_APP_OKTA_CLIENTID=0oa2fs6vp3W5MTzjh1d7 REACT_APP_OKTA_URL=https://hhs-prime.oktapreview.com yarn e2e`
3. _VERIFY_: tests pass

### Testing
- [ ] Tested locally?
- [ ] Ran `./prime test` or `./gradlew testSmoke` against local Docker ReportStream container?
- [ ] (For Changes to /frontend-react/...) Ran `npm run lint:write`? 
- [ ] Added tests?

### Process
- [ ] Are there licensing issues with any new dependencies introduced?
- [ ] Includes a summary of what a code reviewer should test/verify?
[...]

PR #6826, opened 2022-09-26
Experience/membership hook init

doug-s-nava:
in order to account for lag between:
* okta hook initializing (it returns null on first render even when credentials exist)
* useOktaMemberships hook initializing (it returns null on first render even when an active membership exists)

We have expanded the idea of an `initialized` user state to wait until the membership hook has had a chance to render and return data (we had already made this change to account for the okta hook's initialization in a previous PR https://github.com/CDCgov/prime-reportstream/pull/6811). This change is necessary because cypress runs fast enough that the lag between the okta hook being initialized and the useOktaMemberships hook being initialized is significant.

## Test Steps:
1. cherry pick this commit into https://github.com/CDCgov/prime-reportstream/pull/6719
2. run tests from that PR
3. _VERIFY_: it works
4. also does this fix problems on Tom's branch?

## Changes

### Testing
- [ ] Tested locally?
- [ ] Added tests?

### Process

## Linked Issues
- Fixes #issue

## To Be Done
- #issue

    - new endpoint A
    - new endpoint B

### DIFF
--- a/frontend-react/src/hooks/UseOktaMemberships.ts
+++ b/frontend-react/src/hooks/UseOktaMemberships.ts
@@ -1,5 +1,6 @@
 import React, { useEffect, useReducer, useMemo } from "react";
 import { AccessToken, AuthState } from "@okta/okta-auth-js";
+import omit from "lodash.omit";
 
 import { getOktaGroups, parseOrgName } from "../utils/OrganizationUtils";
 import {
@@ -21,6 +22,7 @@ export enum MembershipActionType {
     ADMIN_OVERRIDE = "override",
     RESET = "reset",
     SET_MEMBERSHIPS_FROM_TOKEN = "setMemberships",
+    INITIALIZE = "initialize",
 }
 
 export interface MembershipSettings {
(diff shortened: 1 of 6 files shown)


frontend-react/src/hooks/UseOktaMemberships.ts
```diff
@@ -169,20 +177,24 @@ const calculateNewState = (
...
 // try to read from existing stored state
+// since this is only called on first render, initialized will always be `false`
+// even though we are (needlessly) storing and receving initialized values from store
 export const getInitialState = () => {
     const storedState = getSessionMembershipState();
     const storedStateWithOverride = calculateMembershipsWithOverride(
         storedState || {}
     );
-    return { ...defaultState, ...storedStateWithOverride };
+    return { ...defaultState, ...storedStateWithOverride, initialized: false };
```
kevinhaube, line 197: This could be problematic. I'd advocate for `initialized: false` to be moved into `defaultState` so storedStateWithOverride's initialized value is present on override.

kevinhaube (changes requested):
Just one thing I think might be a semantic bug, but we know it works for Cypress tests so yay!

frontend-react/src/hooks/UseOktaMemberships.ts
doug-s-nava: yeah - as discussed this is to make sure that the `false` from the defaultState doesn't get overwritten by whatever is stored in state. We always want to make sure we initialize with a `false` value here, and let the hook do the work on second render

kevinhaube (approved):
✅ Good work!

PR #6858, opened 2022-09-28
Adds a hard check for admin auth in memberships map

kevinhaube:
This PR removes the need for Admins to closely manage their Okta groups. Previously, if you okta groups claim contained anything other than PrimeAdmins, it would mess up admin auth depending on the order of claims. Now, instead, we hard check where hard admin requirements are required by looking at the `memberships` map and seeing if you have an admin membership anywhere, not just in your `activeMembership` slot.

Test Steps:
1. Go onto OktaPreview and change your user's Groups to include a receiver (DHxx_phd) and DHPrimeAdmins (in that order)
2. Login to your local on this branch
3. Try to navigate to an admin-only page like `/admin/settings`

You should be redirected back home.

## Changes
- AuthElement now hard checks for admin permissions
- SessionContext now provides a memoized hard check for admin permissions

### Testing
- [X] Tested locally?
- [X] Added tests?

### Process

## Linked Issues
- Fixes #issue

## To Be Done
- #issue

    - new endpoint A
    - new endpoint B

### DIFF
--- a/frontend-react/src/components/AuthElement.test.tsx
+++ b/frontend-react/src/components/AuthElement.test.tsx
@@ -1,6 +1,6 @@
 import { render, screen } from "@testing-library/react";
 
-import { MemberType } from "../hooks/UseOktaMemberships";
+import { MembershipSettings, MemberType } from "../hooks/UseOktaMemberships";
 import { mockSessionContext } from "../contexts/__mocks__/SessionContext";
 import { mockFeatureFlagContext } from "../contexts/__mocks__/FeatureFlagContext";
 import { FeatureFlagName } from "../pages/misc/FeatureFlags";
@@ -196,5 +196,34 @@ describe("AuthElement unit tests", () => {
         );
         const spinner = await screen.findByTestId("rs-spinner");
         expect(spinner).toBeInTheDocument();
+    });
+    test("Permits admins whose active membership is not DHPrimeAdmins", () => {
+        mockSessionContext.mockReturnValueOnce({
+            oktaToken: {
+                accessToken: "TOKEN",
+            },
+            memberships: new Map<string, MembershipSettings>().set(
+                "DHPrimeAdmins",
+                {
+                    memberType: MemberType.PRIME_ADMIN,
+                    parsedName: "PrimeAdmins",
+                }
+            ),
+            activeMembership: {
+                memberType: MemberType.NON_STAND,
+                parsedName: "IHaveNoGroupsSadFace",
+            },
+            dispatch: () => {},
+            initialized: true,
+            adminHardCheck: true,
+        });
+        render(
+            <AuthElement
+                element={<TestElement />}
+                requiredUserType={MemberType.PRIME_ADMIN}
+            />
+        );
+        expect(screen.getByText("Test Passed")).toBeInTheDocument();
+        expect(mockUseNavigate).not.toHaveBeenCalled();
     });
 });
--- a/frontend-react/src/components/AuthElement.test.tsx
+++ b/frontend-react/src/components/AuthElement.test.tsx
@@ -197,6 +197,11 @@ describe("AuthElement unit tests", () => {
         const spinner = await screen.findByTestId("rs-spinner");
         expect(spinner).toBeInTheDocument();
     });
+    /* This can happen if a user is a member of multiple Okta orgs, and the one set for them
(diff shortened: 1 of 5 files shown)


frontend-react/src/components/AuthElement.test.tsx
```diff
@@ -197,4 +197,33 @@ describe("AuthElement unit tests", () => {
...
+                    parsedName: "PrimeAdmins",
+                }
+            ),
+            activeMembership: {
+                memberType: MemberType.NON_STAND,
+                parsedName: "IHaveNoGroupsSadFace",
+            },
+            dispatch: () => {},
+            initialized: true,
+            adminHardCheck: true,
```
penny-lischer, line 218: Do we have a test when adminHardCheck = false?

kevinhaube: As far as the AuthElement is concerned, it's false in any of the tests above without the `adminHardCheck` value present, but there was something about this that still didn't settle right with me.

I mulled this over a bit, and thought "well, it's a null case that _should_ never happen", so what I did was went and made some SessionContext tests that handle just the logic of that memoized `adminHardCheck` value against 3 cases. Check those out on the updated diff and let me know if the address what the concern was here! :)

frontend-react/src/contexts/SessionContext.test.tsx
```diff
@@ -1,47 +1,69 @@
...
+import { screen } from "@testing-library/react";
 
 import { MemberType } from "../hooks/UseOktaMemberships";
+import { mockUseOktaMemberships } from "../hooks/__mocks__/UseOktaMemberships";
+import { renderWithSession } from "../utils/CustomRenderUtils";
 
-import { SessionContext } from "./SessionContext";
+import { useSessionContext } from "./SessionContext";
 
-const TestComponent = () => {
```
kevinhaube, line 8: What the old version of this component did is A) my fault and B) very annoying. It's also covered in useOktaMemberships tests. The provider's only logic is the admin hard check. The diff will show the updated component with a comment as to why it even needs to be there. If anyone knows a better way to test context state without a component, let me know!

frontend-react/src/components/AuthElement.tsx
```diff
@@ -33,11 +34,12 @@ export const AuthElement = ({
     );
     // Dynamically authorize user from single or multiple allowed user types
     const authorizeMemberType = useMemo(() => {
```
doug-s-nava, line 36: out of scope of this change, so ignore if you want, but since this is just a memoized value, could it be `memberTypeAuthorized`?

frontend-react/src/components/AuthElement.tsx
```diff
@@ -33,11 +34,12 @@ export const AuthElement = ({
...
-        if (memberType === MemberType.PRIME_ADMIN) return true; // Admin authentication always
+        if (adminHardCheck || memberType === MemberType.PRIME_ADMIN)
```
doug-s-nava, line 37: do we need the second half of this? wouldn't `memberType === MemberType.PRIME_ADMIN` in any situation where adminHardCheck returned true?

frontend-react/src/contexts/SessionContext.test.tsx
doug-s-nava: yeah, the context is really only being used to create the provider, which only does anything when interacting with a dom of some sort, so you kind of have to render something.

i do think you could do this sort of test against the provider, rather than the hook, though. I think this is fine, but maybe something to look at more and try to standardize in testing conventions soon.

doug-s-nava (approved):
✅ ✅ ✅ ✅ 

this works!!

out of scope here, but I think part of the reason this bug was so concerning is that no reason was given about the redirect. Can we open a ticket to show some sort of toast when un-authorized users are redirected back to home?

frontend-react/src/components/AuthElement.tsx
kevinhaube: Yeah, we might be good to take that out for AuthElement -- I kept looking at it today like "Why _is_ that there?"

frontend-react/src/contexts/SessionContext.test.tsx
kevinhaube: What do you mean "you could do this sort of test against the provider, rather than the hook"?

PR #6862, opened 2022-09-28
Json differ support

### LINKED ISSUE #6804: Text Differ for side-by-side compare doesn't function well for json
The component that shows differences between to text areas is great for text, but doesn't work so well for json.

example
image

The single character `n` in the middle of the diff block matches on both sides and creates a confusing compare for json (it matches the `n` in `name`).

TomNUSDS:
Tried to find/use off-the-self json compare feature, but wasn't able to find a reliable one. 
 

This change uses nested new pieces:
 - DiffEditor (and StaticDiff) use to intelligently highlight differences on structured json (as opposed to text like it currently does).
 - A JsonDifferMarkup that injects `<mark></marks>` around json differences based on the JsonDiffer
 - JsonDiffer that support a structured data differences between two json objects based on JsonMapSource.
 - A JsonMapSource that breaks down json and produces a map of offsets into key and values

Other details:
 - LOTS Unit tests for each layer.
 - Since there are more individual files related to doing comparisons, moved them all into a common directory `/src/utils/DiffCompare`

This fixes #6804 
There are two UI locations that are affected: The Static Diff (when using Setting History) and the Edit Diff (when saving Settings).

Compare results of simple text Json vs structure text comparison. 

Text:
image

Json
image

How to test:
https://youtu.be/BKEzoijleK8


frontend-react/src/components/Admin/AdminFormEdit.tsx
```diff
@@ -86,10 +87,20 @@ export const TextAreaComponent = (params: {
...
+                            showError(
+                                `JSon data generated an error "${errorMsg}"`
+                            );
+
+                            // now select the problem area inside the TextArea
+                            const start = Math.max(offset - 4, 0); // don't let go negative
+                            const end = Math.min(offset + 4, text.length); // don't let go past len
+                            inputRef?.current?.focus();
+                            inputRef?.current?.setSelectionRange(start, end);
                         }
```
TomNUSDS, line 104: The above section was moved from misc. It needs deep access to the inputRef, so it makes more sense to have it in the component itself. I did this because other location needed a generic `checkJson()` feature without specifics.

frontend-react/src/components/StaticCompare.text.tsx
```diff
@@ -0,0 +1,33 @@
+import { render, screen } from "@testing-library/react";
```
doug-s-nava, line 1: this filename should be `.test.tsx` otherwise jest won't pick it up

frontend-react/src/components/StaticCompare.tsx
```diff
@@ -1,150 +1,91 @@
...
-                            className="rs-static-compare-base rs-editable-compare-background"
-                            data-testid={"right-compare-highlight"}
-                            dangerouslySetInnerHTML={{
-                                __html: `${DOMPurify.sanitize(
-                                    rightHighlightHtml
-                                )}<br/>`,
-                            }}
-                        />
-                    </ScrollSyncPane>
+        <>
```
doug-s-nava, line 48: is this change just related to wrapping in a fragment in order to add the checkbox?

frontend-react/src/components/StaticCompare.tsx
```diff
@@ -1,150 +1,91 @@
...
+                                    ),
+                                }}
+                            />
+                        </ScrollSyncPane>
+                    </div>
                 </div>
+            </ScrollSync>
+            <div className={"padding-bottom-2"}>
+                <Checkbox
+                    id={"scrollsync"}
```
doug-s-nava, line 81: would recommend a more descriptive name here - `scroll-sync-checkbox`, `scroll-sync-toggle`

frontend-react/src/utils/DiffCompare/AbstractDiffer.ts
```diff
@@ -0,0 +1,4 @@
+export type DifferMarkupResult = {
```
doug-s-nava, line 1: smallest nit possible - if this is all that's going in this file you could maybe stash it in diff.ts

frontend-react/src/utils/DiffCompare/JsonDiffer.ts
```diff
@@ -0,0 +1,327 @@
...
+/**
+ * Loop over array of parentPaths and if childPath has any parentPaths as the
+ * parent return true.
+ * Not a great algo because it doesn't stop as soon as there's a match, but ok for
+ * small amounts of data.
+ */
+const isNotInAnyPath = (childPath: string, parentPaths: string[]) =>
+    parentPaths.filter((p) => isInPath(childPath, p)).length === 0;
+
+const pathDept = (path: string) => path.split("/").length;
```
doug-s-nava: `pathDepth`?

frontend-react/src/utils/DiffCompare/JsonDiffer.ts
```diff
@@ -0,0 +1,327 @@
...
+
+/**
+ * Technically, if a nested node changes in json, then all parent nodes
```
doug-s-nava, line 54: 👍

frontend-react/src/utils/DiffCompare/JsonDiffer.ts
```diff
@@ -0,0 +1,327 @@
...
+ * @param pathArray
+ */
+const extractLeafNodes = (pathArray: string[]): string[] => {
+    // we want to remove parent paths from changed elements.
+    if (pathArray.length <= 1) {
+        // empty or single item, nothing to do.
+        return pathArray;
+    }
+
+    // iterate is since leaf nodes will be last.
```
doug-s-nava: not sure what this comment means - can you clarify or remove?

frontend-react/src/utils/DiffCompare/JsonDiffer.ts
```diff
@@ -0,0 +1,327 @@
...
+    return pathArray;
+};
+
+type Marker = {
+    start: number;
+    end: number;
+};
+
+/**
+ * Given a list of keys into the json map, produce and array of start/end markers
```
doug-s-nava, line 127: `and` -> `an`

frontend-react/src/utils/DiffCompare/JsonDiffer.ts
```diff
@@ -0,0 +1,327 @@
...
+                end: jsonMap.pointers[each].valueEnd.pos,
+            },
+        ],
+        []
+    );
+};
+
+/**
+ * Given a list of keys into the json map, produce and array of start/end markers.
+ * This is the *start of the key* to the end of the value
```
doug-s-nava: unclear based on these comments what the difference is in terms of use cases for these two functions. maybe an example?

frontend-react/src/utils/DiffCompare/JsonDiffer.ts
```diff
@@ -0,0 +1,328 @@
...
+    // this is looking for diffs between the two lists.
+    let addedLeftKeys = leftKeys.filter(
+        (key) => key.length && !rightKeys.includes(key)
+    );
+    let addedRightKeys = rightKeys.filter(
+        (key) => key.length && !leftKeys.includes(key)
+    );
+
+    // now we want intersection (aka NOT changed and see if the values have changed).
+    let intersection = leftKeys.filter((key) => rightKeys.includes(key));
```
doug-s-nava, line 231: `commonKeys` ? `intersectingKeys`?

frontend-react/src/utils/DiffCompare/JsonDiffer.ts
```diff
@@ -0,0 +1,328 @@
...
+
+    // for readability improvements only, pull out start/end values
+    const getStartEnd = (key: string, data: SourceMapResult) => {
```
doug-s-nava, line 234: can these utility functions be split out of this parent function?

frontend-react/src/utils/DiffCompare/JsonDiffer.ts
```diff
@@ -0,0 +1,328 @@
...
+
+/**
+ * Given two json find the difference, then return a normalized json string and one with <mark>s
+ * added for the differences.
+ *
+ * @param leftJson Valid Json object
+ * @param rightJson Valid Json object
+ */
+export const jsonDifferMarkup = (
+    leftJson: any,
```
doug-s-nava, line 267: i think `unknown` is generally preferred to `any` unless that causes problems

frontend-react/src/utils/DiffCompare/JsonDiffer.ts
```diff
@@ -0,0 +1,328 @@
...
+    rightJson: any
+): DifferMarkupResult => {
+    // left and right should be json objects, but there's really no way to typescript enforce it.
```
doug-s-nava, line 270: will this error somewhere if strings are passed instead of pojos?

doug-s-nava (commented):
only made it to jsonDifferMarkup, will try and look again if I have time later this week

devils advocate - could we just use this and output whatever this spits out? much easier to maintain if not quite as flashy https://github.com/andreyvit/json-diff

frontend-react/src/components/StaticCompare.text.tsx
TomNUSDS: Oh, typo!!! Great catch.

frontend-react/src/components/StaticCompare.tsx
TomNUSDS: Yes. I was trying to do a demo and couldn't line up the left and right sides and it was frustrating, so this is an experiment to see if disabling the scroll sync is useful.

frontend-react/src/components/StaticCompare.tsx
TomNUSDS: Oh, much better names! Done.

frontend-react/src/utils/DiffCompare/AbstractDiffer.ts
TomNUSDS: I see your point. My logic for breaking it out was that the text differ and the json differ should have the same interface so they can be easily swapped. I should add a comment saying as much.

TomNUSDS: Ah, I should make a whole interface and have them both implement it. I'll make this change.

frontend-react/src/utils/DiffCompare/JsonDiffer.ts
TomNUSDS: They were originally just inlined. I split them out _just_ to improve readability.
`getStartEnd()` is only called just on the next line which is `getValue()`. 
And `getValue()` is only called on the next line `intersection.filter()`.

frontend-react/src/utils/DiffCompare/JsonDiffer.ts
TomNUSDS: Good to know! I just converted this parameter to a json string to keep it consistent with textDiffer. Going to mark this "resolved".

TomNUSDS: Ended up using elsewhere so thanks!

frontend-react/src/utils/DiffCompare/JsonDiffer.ts
TomNUSDS: This is a good question. A string with quotes doesn't fail because it can be passed into JSON.parse(`"This doesn't fail"`). But the function doesn't do anything. I added a test to make sure this is the case.

frontend-react/src/utils/DiffCompare/JsonDiffer.ts
TomNUSDS: Added some visual aids (ascii art!)

   { "key": "value"}
     ▲            ▲
     └────────────┘
     
vs

  { "key": "value"}
           ▲     ▲
           └─────┘

frontend-react/src/components/Admin/AdminFormEdit.tsx
```diff
@@ -86,10 +87,20 @@ export const TextAreaComponent = (params: {
...
-                        const result = checkTextAreaJson(text, key, inputRef);
-                        if (result !== false) {
-                            // checkTextAreaJson made sure the following call won't throw.
-                            params.savefunc(result);
+                        const { valid, offset, errorMsg } = checkJson(text);
+                        if (valid) {
+                            // checkJson made sure the following JSON.parse won't throw.
+                            params.savefunc(JSON.parse(text));
+                        } else {
+                            showError(
```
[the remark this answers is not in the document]
TomNUSDS: Yeah, I've converted the code to highlight the background in red when there's a mistake, so this bit of "select to show error" can go away.

frontend-react/src/utils/DiffCompare/AbstractDiffer.ts
TomNUSDS: OK. I tried to change it to be an interface, but it cascaded into a bunch of weirdness since the function really does want json objects.

Look it over and see if it's good enough for now?

frontend-react/src/components/Admin/AdminFormEdit.tsx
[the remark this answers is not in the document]
TomNUSDS: Here's an example of the behavior: https://youtu.be/BrxolvksPLg

frontend-react/src/utils/DiffCompare/JsonDiffer.ts
```diff
@@ -0,0 +1,330 @@
...
+): DifferMarkupResult => {
+    const leftMap = jsonSourceMap(leftJson, 2);
+    const rightMap = jsonSourceMap(rightJson, 2);
+    const diffs = jsonDiffer(leftMap, rightMap);
+
+    // so `{level1: { level2: { level3: "value"} } }` vs `{level1: { level2: { level3MOD: "value"} } }`
+    // The keys `/level/level2/level3` (left) and  `/level1/level2/level3MOD` (right) are different,
+    // BUT so is the PARENT VALUE `/level1/level2`
+    // so we go back and remove changedKeys from the addedKeys
+    let leftChanged = diffs.changedKeys;
```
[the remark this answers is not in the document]
TomNUSDS: I did a bit of normalization around the left/right repeated logic until at first, but I undid it because I felt it made following the logic harder (I think I got it down to 5 actual lines per side).

TomNUSDS:
> a couple more nits and one big question around dependency management
> 
> I think my main concern on this work is whether it's really necessary (😬 )
> 
> Would O&O be ok with a simpler out of the box solution for JSON diffing? I am worried about the complexity of this system and what the cost is going to be to maintain it

I attempted to use an out-of-box package JSON differ and couldn't find one that had any significant number of users and was actively maintained.

I suspect this bit of code should be spun out into its own independent package and then pulled back into this project (because I think there's a need out there), but that's more of a time sync. I might take that up as a weekend project and use my non-usds git account.

Hopefully, this code will be like the text differ I did 6 months ago and not be touched once it's working.

frontend-react/src/utils/misc.test.ts
```diff
@@ -27,6 +28,24 @@ test("splitOn test", () => {
     expect(JSON.stringify(r4)).toBe(`["","fooBAr",""]`);
 });
 
+test("checkTextAreaJson", () => {
```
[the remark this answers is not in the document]
doug-s-nava: bump

doug-s-nava (approved):
✅ ✅ ✅ ✅ (but fix the test name please)

frontend-react/src/utils/misc.test.ts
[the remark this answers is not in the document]
TomNUSDS: Opps missed this one. Changed to "verify checking json for errors"

PR #6877, opened 2022-09-30
Experience/6311 display QF messages

### LINKED ISSUE #6311: UI - Integrate Quality Filter response into Validation Service
## Problem statement

As a user of Validation Service for HL7 data, I would like to see the response from quality filters when selecting a jurisdiction, without transmitting the data forward. Please refer to #6312 for design work.

## Acceptance criteria

- [x] The validation service displays results from quality filters on the UI based on the selected jurisdiction for HL7 test data.

## To do

penny-lischer:
This PR ...

Test Steps:

## Changes
- Updates to the file Validator  - addition of quality filter messages and UI changes
Screen Shot 2022-09-29 at 8 45 31 PM

### Testing
- [X] Tested locally?
- [X] Added tests

## Linked Issues
- Fixes #6311


frontend-react/src/styles/_reportstream.scss
```diff
@@ -110,6 +110,11 @@
   border-color: #dfe1e2;
 }
 
+.record-count {
```
kevinhaube, line 113: Is there a way we can handle this via the uswds sass utility classes instead of our own CSS? https://designsystem.digital.gov/utilities/font-size-and-family/

frontend-react/src/components/FileHandlers/FileHandlerMessaging.test.tsx
```diff
@@ -182,3 +180,205 @@ describe("FileWarningsDisplay", () => {
         expect(firstCells[3]).toHaveTextContent("first_id");
     });
 });
+
+describe("FileQualityFilterDisplay", () => {
+    test("renders expected content", async () => {
+        const destinations: Destination[] = [
```
kevinhaube, line 186: This brings up a good question about test data and verbosity. Some places we generate fake data, others we have hard-coded stuff in variables like this. Maybe CC @doug-s-nava on this question:  Can we cut this data down to 1-2 destinations?

And maybe more of a follow-up for Doug:  test data generation, yay or nay? I know we're erring on the side of more verbose testing for clarity; does data generation hurt that clarity we're going for?

kevinhaube (changes requested):
I think you should be okay to use the USWDS utility classes like `font-sans-sm` and `text-normal` rather than adding custom CSS

I also left a question in here for Doug related to test data, but that's non-blocking imo. 👍🏼 

Functionally, everything works and looks good! I went in, mutilated a file's contents, and got some pretty good messaging about it on my UI.

doug-s-nava (commented):
in my testing it seems like a file that does not meet quality filters, but has no errors, will still come up as a successful validation, listed above the quality filter results.


since a validation attempt that returns some quality filter results means that at least some of the messages contained in the file were filtered out for at least some jurisdictions I feel like we should either:

* not show a success message
* show a success message with a specific notice around how many messages passed validation and how many did not

@a-chappelear - does that make sense to you?

frontend-react/src/components/StaticAlert.tsx
```diff
@@ -17,13 +17,16 @@ export const StaticAlert = ({
     const alertClasses = classNames({
         "usa-alert": true,
         "usa-alert--success": type === "success",
+        "usa-alert--success usa-alert--slim": type === "success slim",
```
doug-s-nava, line 20: [nit] `usa-alert--slim` could be handled independently

```suggestion
        "usa-alert--slim": type.indexOf("slim") > -1,
```

a-chappelear:
Yeh, wondered how this may work in practice. You have passed "structural" tests, but not the additional filters - I would suggest not showing any success message if you technically are not passing validation still due to the quality filters. And if that's true then -  we may need to adjust the "warning" message at the top to only say  "We recommend addressing warnings to enhance clarity."  (cutting out the "you passed" message)

doug-s-nava (commented):
one more small nit

frontend-react/src/components/FileHandlers/FileHandlerMessaging.test.tsx
penny-lischer: Spoke with Doug and will clean up test data

kevinhaube (approved):
Approving to remove `changes requested` re-review requirement, but it looks like you're still going to be blocked because of some strange commits showing up in your diff here that triggered the need for DevOps review.

frontend-react/src/components/FileHandlers/FileHandler.test.tsx
```diff
@@ -348,14 +348,9 @@ describe("FileHandler", () => {
             // testing creation of error messaging for upload
             // for now, assuming that if this works, it will work for validation as well
             const message = await screen.findByText(
-                "The following warnings were returned while processing your file. Your file has passed validation, but these warning areas can be addressed to enhance clarity."
+                "We recommend addressing warnings to enhance clarity."
```
doug-s-nava, line 351: Andrea and Florence can weigh in before this goes to a wide release, but I would keep the first sentence `The following warnings were returned while processing your file. We recommend addressing warnings to enhance clarity.`

doug-s-nava (approved):
one minor nit on verbiage, but otherwise LGTM, nice job!

✅ ✅ ✅ ✅

frontend-react/src/components/FileHandlers/FileHandler.test.tsx
a-chappelear: That works, thanks @doug-s-nava

penny-lischer: Will update

Issue #6582, opened 2022-08-30
FHIR Engine uses STLT specific schema to send data to CO


carlosfelix2:
### User Story:
As a receiver, I want the data sent to me to follow my specific schema as configured in my receiver settings, so the data is generated as I need it.

### Description/Use Case
STLT specific schemas will be used to generate HL7 data as required by the STLT.  Most STLTs have requirements to suppress certain fields, hard code certain values, etc.  After routing, the conversion will use the STLTs schema to generate the data to be sent.

### Risks/Impacts/Considerations
* 

### Dev Notes:
* Note that routing has not been developed yet, so you will need to setup the code to do some fake/hard coded routing.  CO is the first STLT.
* Likely will need to modify the setting table to have values for schema name and schema path for universal pipeline. 
* This should be a receiver with topic 'full-elr'

### Acceptance Criteria
* The FHIR Engine uses a receiver's setting to determine which schema to use for the FHIR to HL7 conversion in the pipeline.
* The FHIR Engine uses the receiver's specified schema from the receiver settings to generate HL7 data from FHIR.

PR #6901, opened 2022-10-03
6582 SLTL-Specific schemas 

TurkeyFried:
This PR updates the FHIR translator to load custom schemas as defined inside a receiver.

Test Steps:
1. Run unit tests. A new test has been added that verifies custom schemas for CO.

## Changes
- Added support for a list of receiver names on FHIR translations.
- Added tests to check for different amounts of receivers.

## Linked Issues
#6582


carlosfelix2 (changes requested):

prime-router/src/main/kotlin/Receiver.kt
```diff
@@ -63,6 +63,14 @@ open class Receiver(
      * the date time would have the offset applied to it
      */
     val dateTimeFormat: DateUtilities.DateTimeFormat? = DateUtilities.DateTimeFormat.OFFSET,
+    /**
+     * Name of the file containing translation customizations for this receiver
+     */
+    val schemaFileName: String? = null,
```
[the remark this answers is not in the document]
TurkeyFried: Fixed

prime-router/src/main/kotlin/Receiver.kt
```diff
@@ -63,6 +63,10 @@ open class Receiver(
      * the date time would have the offset applied to it
      */
     val dateTimeFormat: DateUtilities.DateTimeFormat? = DateUtilities.DateTimeFormat.OFFSET,
+    /**
+     * Location of the custom configuration files for this receiver
+     */
+    val schemaFolderPath: String? = null
```
carlosfelix2, line 69: Remove the path as well.  We will just use the schema name only.  The goal is that when we move the schemas to the DB then there will be no such thing as a path.  For not the path is part of the schema name.

carlosfelix2 (changes requested):

prime-router/src/main/kotlin/fhirengine/engine/FHIRTranslator.kt
```diff
@@ -61,8 +61,8 @@ class FHIRTranslator(
                 // We only process receivers that are active and for this pipeline.
                 if (receiver != null && receiver.topic == Topic.FULL_ELR.json_val) {
                     val converter = FhirToHl7Converter(
-                        "ORU_R01-base",
-                        "metadata/hl7_mapping/ORU_R01"
+                        receiver.schemaName,
```
carlosfelix2, line 64: You will need to add the new constructor I added in my play branch and the check as well.  Check the code I mentioned a while back

carlosfelix2 (approved):

PR #6928, opened 2022-10-05
Removes RSErrorBoundary use for ValueSets mutations

kevinhaube:
This PR reverts the use of error boundaries when catching error from mutation calls. Queries will still rely on RSErrorBoundary for error UI, but on-page error UIs are dependent on error information that the error boundary cannot pass into its children. (Will discuss in eng sync). 

## Context (from Slack)
We can chat in eng sync tomorrow about it, but as I work on the Waters API conversion to react-query, I’m finding that our useMutation error UIs aren’t compatible with the ErrorBoundary pattern the same way our useQuery error UIs are.

```
<ErrorBoundary>
    <Component />
<ErrorBoundary>

const Component = () => {
    // useMutation hook call here
    // Render logic here
}
```

When you trigger the mutation and it returns an error, the boundary catches this. Because the state object changes in the ErrorBoundary, that and all its children are re-rendered, so we cannot show on-page error messaging conditional to that response, we can either replace the entire component with error UI (how we handle query errors) or attempt to show it as a banner above the UI.

This is what the latter solution would render out as, and why it’s a problem:

```
<ErrorBoundary> <-- Caught the error
    <ConditionallyRenderedErrorUI /> <-- Rendered from ErrorBoundary
    <Component /> <--- Refreshed, no idea it errored before this
</ErrorBoundary>
```

Since it’s refreshing that child, there’s no stored error state, and no good way for the boundary to pass it in (plus, that’s a bit of an anti-pattern as far as React is concerned), thus no way to display the info on the page.

In the case of the FileHandler, this would look like a banner above with the network error message, and an empty FileHandler UI waiting for you to drag a file in, again, rather than those errors/warnings returned from the API response that created the error state in the first place.

All that’s to say that I’ll put a PR up tomorrow that rolls back error boundary use with mutations ONLY, keeping boundary usage for normal query fetches. Luckily only one thing is affected: Value Sets. Should be an easy PR.

Test Steps:

## Changes

### Testing
- [x] Tested locally?
- [x] Added tests?

### Process

## Linked Issues
- Fixes #issue

## To Be Done
- #issue

    - new endpoint A
    - new endpoint B


frontend-react/src/hooks/UseValueSets.test.ts
```diff
@@ -89,9 +89,10 @@ describe("useValueSetUpdate", () => {
 
     test("returns trigger and loading indicator", async () => {
         const { result } = renderWithQueryWrapper();
-        const { saveData, isSaving } = result.current;
+        const { saveData, isSaving, saveError } = result.current;
         expect(isSaving).toEqual(false);
         expect(saveData).toBeInstanceOf(Function);
+        expect(saveError).toBeNull();
```
kevinhaube, line 95: We have, and can use, this error from mutation hooks, but because of how we handle the UI for mutation failures, we'll likely rarely use it. Regardless, I added it back so we have access to it, and tests are written to ensure we have that access over the evolution of the hook.

doug-s-nava (commented):
fwiw lgtm

(but have not tested)

penny-lischer (approved):
LGTM

Issue #6634, opened 2022-09-06
Remove the name property from the FHIR to HL7 library schemas


carlosfelix2:
### User Story:
As a developer, I do not want to have redundant properties in the schema, so it is easier to generate.

### Description/Use Case
The schema name property was a copy of the COVID schema name, but this property is not being used as the filename is being used for the name.  Remove the unneeded property.

### Risks/Impacts/Considerations
* 

### Dev Notes:
* Need to remove from all schemas including the ones in the test folder.
* Need to remove the name from the code as well.
* We need to keep a name in the code for debugging purposes.  Delete the name from the YML files, but use the filename as the name of the schema in code.

### Acceptance Criteria
* The FHIR to HL7 library schemas do not use a name property.

PR #6929, opened 2022-10-05
6634 remove name property from FHIRtoHL7 conversion

TurkeyFried:
This PR gets rid of the `name` field from configuration files. This field was primarily used for debugging, and is now replaced with references that use the file's name.

Test Steps:
1. Run the converter with the `fhirdata` command. e.g. `./prime fhirdata --input-file src/testIntegration/resources/datatests/HL7_to_FHIR/sample_co_1_20220518-0001.fhir -s metadata/hl7_mapping/ORU_R01/ORU_R01-base.yml --output-format=HL7`.
2. Verify the output is valid HL7.
3. Check the debug output and verify it's readable.

## Changes
- Remove enforcing of a `name` field for the conversion config files
- Remove all existing uses of the `name` field
- Change logger calls in the changed methods to still be readable

## Linked Issues
- #6634

### DIFF
--- a/prime-router/metadata/hl7_mapping/ORU_R01/ORU_R01-base.yml
+++ b/prime-router/metadata/hl7_mapping/ORU_R01/ORU_R01-base.yml
@@ -1,12 +1,10 @@
-name: ORU-R01-Base
 hl7Type: ORU_R01
 hl7Version: 2.5.1
 constants:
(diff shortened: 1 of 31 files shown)


TurkeyFried:
The changes in this PR overreached and caused issues with some tests. See #6935 for a saner version

PR #7038, opened 2022-10-17
Cle/bundle destination

### LINKED ISSUE #7035: Phase 1 Universal Pipeline hardcoded routing
### User Story:
As reportstream, I want to hard code CO-PHD as the receiver for phase 1 of Universal Pipeline so I can delivery HCA data.

### Description/Use Case
We need the routing step to calculate its own receivers and add them to the bundle so the translator can just use that data to translate for STLTs.

* Add tests

### Dev Notes:
* Will need to ensure the bundle has targets that have the receiver names

### Acceptance Criteria
* The routing is hard coded in the FHIRRouter, not in the translator
* The FHIR bundle contains target information

clediggins-usds:
This PR moves the hard coded phase 1 routing from FHIRTranslator to FHIRRouter, gets the 'target' elements into the bundle, and adds tests

Test Steps:
1. Run automated tests
OR
1. Run a message through universal pipeline 
2. Verify that a 'send' message for co-phd.full-elr-hl7' is generated

## Changes
- put targets into bundle in the FHIRRouter
- pull targets out of bundle in the FHIRTranslator
- add tests

### Testing
- [x] Tested locally?
- [x] Added tests?

Resolves #7035


prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
```diff
@@ -50,6 +58,17 @@ class FHIRRouter(
             // pull fhir document and parse FHIR document
             val fhirDocument = FhirTranscoder.decode(message.downloadContent())
 
+            // TODO: routing would need to happen here to generate a list of the receivers
+            val hardCodedPhase1ReceiverName = "co-phd.full-elr-hl7"
+            var receiver = settings.findReceiver(hardCodedPhase1ReceiverName)
+            if (receiver == null)
+                logger.error("Receiver $hardCodedPhase1ReceiverName was not found in the settings.")
```
carlosfelix2, line 65: You cannot continue the logic at this point because the receiver is null.  Throw exception?

prime-router/src/main/kotlin/fhirengine/engine/FHIRTranslator.kt
```diff
@@ -51,8 +53,8 @@ class FHIRTranslator(
             // track input report
             actionHistory.trackExistingInputReport(message.reportId)
 
-            // todo: iterate over each receiver, translating on a per-receiver basis - for phase 1, hard coded to CO
-            val receivers = listOf("co-phd.full-elr-hl7")
+            var provenance = bundle.entry.first { it.resource.resourceType.name == "Provenance" }.resource as Provenance
+            var receivers = provenance.target.map { (it.resource as Endpoint).identifier[0].value }
```
carlosfelix2, line 57: Looks like this can be val instead of var

carlosfelix2 (changes requested):

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
clediggins-usds: Updated

TurkeyFried (approved):
Tests pass, new code is documented, looks good!

Issue #6931, opened 2022-10-05
[Message ID][Search][UI][API] Complete


penny-lischer:
## Problem statement

As an internal admin, I would like to be able to see messages that I search upon as well as the details of a message when selecting a messageId from the Admin -> Message ID Search page.

## What you need to know
Mural with initiative break-down https://app.mural.co/t/primedatainput6410/m/primedatainput6410/1660849046968/e257020850ab71f57b31120e27ec06203ae9c22b?sender=u3fea607b46e84058d3224917

## Acceptance criteria

As an internal admin, I am able to see a list of messages and its detail

## TODO:
- [x] Hook up the front-end to get the data from the newly created API 
- [x] Unit tests

## Designs
https://www.figma.com/file/fKl2M96ylkxYezwjVVUSIX/Internal-Admin-%26-Tools?node-id=411%3A1068

PR #7061, opened 2022-10-18
Experience/6930 Message Id Detail

penny-lischer:
This PR ...

Test Steps:
1. Add the "message-tracker" feature flag
2. Click on the Admin -> Message Id Search dropdown
3. Click search (no need for input since we are currently using mock data)
4. Click on a message Id in the table
5. View the message details

## Changes
- Added Message Id details including Sender, Warnings, Errors, Quality Filters
- Updated the feature flag name and renames files from "message-id-search" to "message-tracker"

Screen Shot 2022-10-18 at 12 53 01 PM

### Testing
- [x] Tested locally?
- [X] Added tests?

### Process


frontend-react/src/components/MessageTracker/MessageDetails.tsx
```diff
@@ -0,0 +1,185 @@
...
+                className="grid-container margin-bottom-10"
+                data-testid="container"
+            >
+                <div className="margin-bottom-5">
+                    <h1>Message Details</h1>
+                    <Button onClick={() => navigate(-1)} type="button">
+                        Back
+                    </Button>
+                </div>
+                {withCatchAndSuspense(<MessageDetailsSender />)}
```
kevinhaube, line 171: Maybe a design question, but I'm curious if we'd want a spinner for each or one spinner until all data loads? Cc: @fmccaf1 

Regardless, I definitely think we'll want those catches on all of them!

frontend-react/src/components/MessageTracker/MessageTracker.tsx
```diff
@@ -0,0 +1,162 @@
...
+// Main component.
+export function MessageTracker() {
+    const [searchFilter, setSearchFilter] = useState("");
+    const [isLoading, setIsLoading] = useState(false);
+    const [messagesData, setMessagesData] = useState<MessageListResource[]>([]);
+
+    const searchMessageId = async () => {
+        setIsLoading(true);
+
+        // TODO: make API call here to get data
```
kevinhaube, line 102: Is this for prototyping purposes, or are we going to need to fetch different from how we do with react-query?

frontend-react/src/components/MessageTracker/QualityFilters.tsx
```diff
@@ -0,0 +1,57 @@
+import React from "react";
+
+import { Destination } from "../../resources/ActionDetailsResource";
+
+//TODO: do we want to create a new interface here or use the existing Destination?
```
kevinhaube, line 5: Even if we stick with the old Destination, if it's something that's coming back from the API, I vote for moving it into `src/config/endpoints/[some file that makes sense to put it in].ts`. It'll make the cleanup of those rest-hooks Resource classes a lot easier when we come around to it 😃

frontend-react/src/components/MessageTracker/QualityFilters.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+export const QualityFilters = ({
+    qualityFilters,
+}: QualityFiltersDisplayProps) => {
+    return (
+        <>
+            <section className="margin-bottom-5">
+                <h3>Jurisdictions:</h3>
+                {qualityFilters?.map((d) => (
+                    <React.Fragment key={d.organization_id}>
```
kevinhaube, line 18: Today I learned you can use the `key` prop on a fragment. 😮

frontend-react/src/components/MessageTracker/WarningsErrors.tsx
```diff
@@ -0,0 +1,69 @@
+import React from "react";
+
+// TODO: move this interface into the resources directory
```
kevinhaube, line 3: Doug had started moving network data interfaces like this into `src/config/endpoints/*.ts` where we store the endpoints they return from. There's a few examples in there already, but I think that Resources folder is heavily rest-hooks related so moving things we still use out of there helps with that cleanup later.

frontend-react/src/components/MessageTracker/WarningsErrors.tsx
```diff
@@ -0,0 +1,69 @@
...
+    title: string;
+    data: Warnings[];
+};
+
+export const WarningsErrors = ({ title, data }: WarningsErrorsDisplayProps) => {
+    return (
+        <>
+            <section className="margin-bottom-5">
+                <h3>{title}</h3>
+                <table
```
kevinhaube, line 21: Is this something we could use the `Table` component for? The [DeliveryFacilitiesTable](https://github.com/CDCgov/prime-reportstream/blob/master/frontend-react/src/pages/deliveries/details/DeliveryFacilitiesTable.tsx) is a good example of a simpler table config.

kevinhaube (commented):
First pass on the code, gonna take a poke around the UI in the morning. This looks great!

frontend-react/src/components/MessageTracker/MessageDetails.tsx
penny-lischer: Good call, lets discuss.

frontend-react/src/components/MessageTracker/WarningsErrors.tsx
penny-lischer: I thought about it and didn't implement it for reasons. Will discuss with you and Aurora

kevinhaube: Sounds good! 👍🏼

frontend-react/src/components/MessageTracker/QualityFilters.tsx
kevinhaube: Super clutch -- I didn't know `<React.Fragment>` would render the same as `<>` so I've always had to wrap my fragments in a component and key the component instead. Glad I saw this so I don't have to anymore!

frontend-react/src/components/MessageTracker/MessageDetails.tsx
kevinhaube: I reached out to Florence to get some input for that discussion as well, noting that neither really slow down our fetching process and either one can be its own useful experience (i.e. seeing data faster vs seeing more complete data all at once). I'll paste her reply here once I have it!

penny-lischer:
@kevinhaube will make changes as part of #6931

ahay-agile6 (approved):

PR #7069, opened 2022-10-19
Move rarely used Delete Setting feature down into Edit page

TomNUSDS:
Refactor only. Prep for adding Check Settings feature. Step 1 is to move the Delete out of list and down into setting edit.
- Move Delete Setting
- Refactor confirmation modal to be more generic
- Add missing unit tests for existing components.

Delete button removed from list components:
image

image

Delete button added to edit components (Send and Receive):
image

Moved the confirmation modal into a generic component that should be easy to reuse:
image

How to use from comment section:
```
Wrapper on top of ModalRef to do simple confirmation modal.
Step 1 - Add component to end of page:
            <ModalConfirmDialog
                id={"deleteConfirm"}
                onConfirm={confirmDelete}
                ref={modalRef}
            ></ModalConfirmDialog>

Step 2 - Set up handler to display the dialog.
         Note the messages are supplied here instead of step 1 for flexibility.
   const modalRef = useRef<ModalConfirmRef>(null);
   const ShowDeleteConfirm = (itemId: string) => {
        modalRef?.current?.showModal({
            title: "Confirm Delete",
            message:
                "Deleting a setting will only mark it deleted. It can be accessed via the revision history",
            okButtonText: "Delete",
            itemId: itemId,
        });
   const confirmDelete = async (deleteItemId: string) => {
   // handle Confirm Button click action here
   };
   ```

### CODE REVIEW

**Comment by TomNUSDS** [comment]
SonarCloud is not happy because `54.9% Coverage on New Code` but it's really just moving old code around that had 0% coverage tests. We'll need to port the code to the network api to reach 80%.

**Review by kevinhaube — APPROVED** [review]
Nice, this adds so many tests, screw SonarCloud's judgement. You get a pass in my books 👍🏼 ✅

File: frontend-react/src/components/Admin/EditReceiverSettings.tsx [hunk 1]
```diff
@@ -131,7 +131,10 @@ const EditReceiverSettingsForm: React.FC<EditReceiverSettingsFormProps> = ({
             setOrgReceiverSettingsNewJson(
                 JSON.stringify(orgReceiverSettings, jsonSortReplacer, 2)
             );
-            if (latestResponse?.version !== orgReceiverSettings?.version) {
+            if (
+                action === "edit" &&
```
TomNUSDS, line 135: Only do the "did someone edit this while I was working on it" check if doing and edit operation (Not for clone)

frontend-react/src/components/Admin/EditReceiverSettings.tsx
```diff
@@ -350,14 +353,16 @@ const EditReceiverSettingsForm: React.FC<EditReceiverSettingsFormProps> = ({
...
                     <Grid col={6}>
-                        <Button
-                            type={"button"}
-                            secondary={true}
-                            data-testid={"receiverSettingDeleteButton"}
-                            onClick={() => ShowDeleteConfirm(receivername)}
-                        >
-                            Delete...
-                        </Button>
+                        {action === "edit" ? (
```
TomNUSDS, line 356: Don't show Delete if this is a newly cloned entry

prime-router/src/main/kotlin/azure/SettingsFacade.kt
```diff
@@ -285,8 +285,8 @@ class SettingsFacade(
 
             db.insertDeletedSettingAndChildren(current.settingId, claims.userName, OffsetDateTime.now(), txn)
             db.deactivateSettingAndChildren(current.settingId, txn)
-
-            Pair(AccessResult.SUCCESS, "")
+            // returned content-type is json/application, so return empty json not empty string
+            Pair(AccessResult.SUCCESS, "{}")
```
TomNUSDS, line 289: I'm not sure Delete was even working in staging? Maybe something changed in the client library, but if the content-type returned is json, then it always try to JSON.parse() it... even if it's empty. The best fix is to just return empty json.

TomNUSDS:
Can you compare my latest edits?

https://github.com/CDCgov/prime-reportstream/pull/7069/commits/30df050c644c5cf2b7733b931fbfd5f3a9d7696f

kevinhaube (approved):
Found one rogue debugger in that commit, but the rest looks great to me!

Issue #6941, opened 2022-10-07
Tighten the CLIA ID regular expression for the getId and getIdType FHIR Path functions


carlosfelix2:
### User Story:
As a developer, I want any code that deals with CLIA to do it properly, so there are no issues in the decoding/encoding process of data.

### Description/Use Case
Initially when the regex was implemented we did not know the CLIA has a more specific pattern to it than implemented.  CLIA can have a patter like `11D1111111` or `D1D1111111`, where the number 1 is a digit 0-9, and the letter D is the literal letter D.  We assumed there could be letters anywhere in the CLIA.  Need to change our code so we are actually looking at this pattern.  Here is a description from the web:
The first three characters are the two-digit state code followed by the letter D. The remaining seven digits are the unique CLIA system number assigned to the provider.
The exception here is that those that start with D (e.g. D1D1111111) are DOD type CLIAs that we also need to accept.

### Acceptance Criteria
* The custom FHIR Path functions that check for CLIA detect a properly formatted CLIA of 11D1111111, where all 1s are digits.
* The custom FHIR Path functions that check for CLIA detect a properly formatted CLIA of D1D1111111, where all 1s are digits.
* The unit tests are updated test for proper CLIAs.

PR #7072, opened 2022-10-19
6941 Updated Regex for CLIA checks

TurkeyFried:
This PR updates the Regex for CLIA checks to be more strict.

Test Steps:
1. Run test suite and verify that `ConstantResolverTests` passes all tests.

## Changes
- Updated the Regex for CLIA checks.
- Added new tests cases to check for bad CLIA numbers.

## Linked Issues
- #6941


prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/ConstantResolver.kt
```diff
@@ -339,9 +340,11 @@ object CustomFHIRFunctions {
     private val oidRegex = """^[0-2](.(0|[1-9]\d*))+$""".toRegex()
 
     /**
-     * Regex to identify CLIAs.  CLIAs are just 10 alphanumeric characters.
+     * Regex to identify CLIAs.
+     * CLIAs are defined as "10-digit alphanumeric numbers, with the "D"in the third position.
+     * One exception is DoD numbers, where the first character is also "D".
      */
-    private val cliaRegex = """^[a-zA-Z\d]{10}$""".toRegex()
+    private val cliaRegex = """^[dD\d]\dD\d{7}$""".toRegex()
```
carlosfelix2, line 347: I think this is wrong with the `dD`.  Do we want to allow for lowercase characters too?  Either here or uppercase the string when the regex is used?

carlosfelix2, line 347: Let's change something here.  From an earlier discussion in Slack about it, maybe lets do any uppercase letter or a digit for those two positions.  The reason is that they said a valid clia can be like Z1D111111 for DOD.  Might as well allow any letter since no one has a spec for this.

prime-router/src/test/kotlin/fhirengine/translation/hl7/utils/ConstantResolverTests.kt
```diff
@@ -257,5 +257,19 @@ class ConstantResolverTests {
         id = CustomFHIRFunctions.getIdType(mutableListOf(cliaId))
         assertThat(id.size).isEqualTo(1)
         assertThat(id[0].primitiveValue()).isEqualTo(cliaType)
+
+        cliaId.value = "D5D9458360" // DoD-style CLIA
```
carlosfelix2, line 261: Add lowercase letters tests.

carlosfelix2 (changes requested):

prime-router/src/test/kotlin/fhirengine/translation/hl7/utils/ConstantResolverTests.kt
TurkeyFried: done

carlosfelix2 (approved):

Issue #7048, opened 2022-10-18
Universal Pipeline does not route data to inactive receivers


carlosfelix2:
### User Story:
As a ReportStream Administrator, I want to be able to disable the sending of data to a UP receiver, so I can manage who receives data.

### Description/Use Case
We want to be able to disable a receiver from receiving data in case there are any issues.

### Acceptance Criteria
* The UP does not route data to a receiver if the receiver's customer status is set to inactive.

PR #7153, opened 2022-10-26
7048 Universal Pipeline - Skip Inactive Receivers

TurkeyFried:
This PR adds an early check for a potential receiver's status, and skips those that are set to `INACTIVE`

Test Steps:
1. Run automated tests, specifically `FhirRouterTests`.
2. To test manually, submit a report. After receiving a response for that submission including the `destinations` array, go to `public.setting` on the database and set these destinations to `is_active=false`.
3. Submit the same report again and verify it's not being sent to the inactive destinations.

## Changes
- Change the `CustomerStatus` check during the `addReceivers` step to skip the loop if it's set to `INACTIVE`
- Add test case and data to cover the new logic

## Linked Issues
- #7048


prime-router/src/test/kotlin/fhirengine/engine/FhirRouterTests.kt
```diff
@@ -230,20 +235,33 @@ class FhirRouterTests {
...
+        val receiversIn = listOf(oneOrganization.receivers[1])
+
+        // act
+        FHIRBundleHelpers.addReceivers(bundle, receiversIn)
+
+        // assert
+        val provenance = bundle.entry.first { it.resource.resourceType.name == "Provenance" }.resource as Provenance
+        val outs = provenance.target
+        val receiversOut = outs.map { (it.resource as Endpoint).identifier[0].value }
+        assert(receiversOut.isEmpty())
```
carlosfelix2, line 265: Shouldn't we have here one receiver since one is testing.

TurkeyFried: not in that case since the test only uses the inactive receiver. I will add another test case to check a mixed scenario

prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
```diff
@@ -26,12 +26,13 @@ object FHIRBundleHelpers {
         // Create the list of target receivers to be added to the Provenance of the bundle
         val targetList = mutableListOf<Reference>()
         receiverList.forEach { receiver ->
+            if (receiver.customerStatus == CustomerStatus.INACTIVE) return@forEach
```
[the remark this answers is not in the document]
TurkeyFried: updated!

clediggins-usds (approved):

PR #7173, opened 2022-10-28
Cle/7043 juris filter

clediggins-usds:
This PR makes universal pipeline routing use fhirPath instead of being hardcoded.

Test Steps:
1. Run tests
OR
1. Set up a local receiver with filters with topic 'full-elr' and add fhirPath to the jurisdisctionalFilter section
2. Submit a message with client headers for a full-elr sender
3. Observe result

## Changes
- Update organization.yml file with fhirPath filter for co-phd.full-elr-hl7
- Update receiver to allow for jurisdictional filters in test cases
- Update fhirRouter to read and apply the fhirPath. Finds receiver filters first, then org filters if there are no receiver filters
- Added tests

### Testing
- [x] Tested locally?
- [x] Added tests?


prime-router/src/main/kotlin/transport/RESTTransport.kt
```diff
@@ -427,4 +427,4 @@ class RESTTransport(private val httpClient: HttpClient? = null) : ITransport {
                 headers.append(entry.key, entry.value)
             }
     }
-}
+}
```
clediggins-usds, line 430: I don't know why this was missing. ktLint caught it after a master merge.

TurkeyFried (approved):
Tests pass, code is documented and makes sense

carlosfelix2 (changes requested):

carlosfelix2 (approved):

Issue #6139, opened 2022-07-18
Detect duplicate element names in FHIR to HL7 schema


carlosfelix2:
### User Story:
As a administrator, I want to know if I am using duplicate element names in a FHIR to HL7 schema, so there are no name conflicts if a schema is later extended.

### Description/Use Case
When we are extending schemas, we have the ability to override an element by name, but if we make a mistake an use a name more than once then there could be issues with the extension changing the wrong fields.  We just need to make sure we do not reuse element names in a schema.

### Dev Notes:
* Thinking maybe throw a SchemaException if an element name is duplicated.

### Acceptance Criteria
* A FHIR to HL7 schema cannot have an element with the name of another element.
* An error should be generated if a duplicate element name is found.

PR #7192, opened 2022-11-01
6139 FHIRtoHL7  Duplicate Element Name Protection

TurkeyFried:
This PR prevents multiple elements in a schema from having the same name.

Test Steps:
1. Run the converter with the `fhirdata` command. e.g. `./prime fhirdata --input-file src/testIntegration/resources/datatests/HL7_to_FHIR/sample_co_1_20220518-0001.fhir -s metadata/hl7_mapping/ORU_R01/ORU_R01-base.yml --output-format=HL7`.
2. Verify the output is valid HL7.
3. Open any FHIRtoHL7 config (i.e. `patient-visit.yml`) and change the `name` value on multiple elements to be the same.
4. Repeat step 1, but now an error will appear pointing at the culprit.

## Changes
- Added a safety check for duplicate element names with an exception that shows the troublemakers.
- Added a new test case to cover this.

## Linked Issues
- #6139


prime-router/src/main/kotlin/fhirengine/translation/hl7/FhirToHl7Converter.kt
```diff
@@ -98,6 +98,14 @@ class FhirToHl7Converter(
         // Add any schema level constants to the context
         // We need to create a new context, so constants exist only within their specific schema tree
         val schemaContext = CustomContext.addConstants(schema.constants, context)
+
+        // check for duplicate named schema elements before processing
+        val dupes =
```
carlosfelix2, line 103: You want to do this check while validating the schema and not when processing data.

prime-router/src/test/kotlin/fhirengine/translation/hl7/FhirToHl7ConverterTests.kt
```diff
@@ -363,5 +363,26 @@ class FhirToHl7ConverterTests {
...
+
+        val dupe = ConfigSchemaElement(
+            "iMustBeUnique",
+            value = listOf(pathWithValue),
+            hl7Spec = listOf("MSH-12")
+        )
+        schema = ConfigSchema(
+            hl7Type = "ORU_R01",
+            hl7Version = "2.5.1",
+            elements = mutableListOf(element, dupe)
```
carlosfelix2, line 382: Add a test where you check for duplicate elements in a schema containing a schema.  Also add another to make sure the same element name in one leg of a schema tree is not a duplicate in another leg.

carlosfelix2 (changes requested):

prime-router/src/test/kotlin/fhirengine/translation/hl7/FhirToHl7ConverterTests.kt
TurkeyFried: Added!

prime-router/src/main/kotlin/fhirengine/translation/hl7/FhirToHl7Converter.kt
TurkeyFried: Moved the check up the chain

prime-router/src/main/kotlin/fhirengine/translation/hl7/FhirToHl7Converter.kt
```diff
@@ -78,6 +78,14 @@ class FhirToHl7Converter(
         // Sanity check, but at this point we know we have a good schema
         check(message != null)
         terser = Terser(message)
+        val dupes = (
+            schemaRef.elements.filter { it.name != null } +
+                schemaRef.elements.flatMap { it.schemaRef?.elements ?: emptyList() }
+            ).groupingBy { it.name }.eachCount().filter { it.value > 1 }
+
+        if (dupes.isNotEmpty()) { // value is the number of matches
+            throw SchemaException("Schema ${schemaRef.name} had multiple elements with the same name: $dupes")
```
carlosfelix2, line 87: Change to: Schema x HAS ...

carlosfelix2: The $dupes portion shows this: Schema ORU_R01-base had multiple elements with the same name: {set-id=2}
What is set-id=2?  Can we cleanup that last portion to make it more readable?  Not sure what the 2 is.

prime-router/src/main/kotlin/fhirengine/translation/hl7/FhirToHl7Converter.kt
carlosfelix2: I do not see the move.  It is still done in the convert.  SHould be done as part of the schema validate.

carlosfelix2 (changes requested):
Tested and works great.

prime-router/metadata/hl7_mapping/ORU_R01/base/datatype/xon-organization.yml
```diff
@@ -3,19 +3,19 @@ elements:
     value: [ '%resource.name' ]
     hl7Spec: [ '%{hl7OrgField}-1' ]
 
-  - name: xon-organization-type
+  - name: xon-organization-name-type
```
carlosfelix2, line 6: Looks like there is one more dupe.  Look for set-id in patient.  Thanks for cleaning all of these.

prime-router/src/main/kotlin/fhirengine/translation/hl7/FhirToHl7Converter.kt
TurkeyFried: Updated it to be clearer

prime-router/metadata/hl7_mapping/ORU_R01/base/datatype/xon-organization.yml
TurkeyFried: Fixed all missing dupes

carlosfelix2 (approved):

prime-router/src/main/kotlin/fhirengine/translation/hl7/FhirToHl7Converter.kt
```diff
@@ -98,6 +106,7 @@ class FhirToHl7Converter(
         // Add any schema level constants to the context
         // We need to create a new context, so constants exist only within their specific schema tree
         val schemaContext = CustomContext.addConstants(schema.constants, context)
+
```
jimmyfagan, line 109: Is there some kind of standard for these kinds of formatting changes on the team? Have we discussed using a common formatter/linter?

TurkeyFried: We're using klinter with some standard settings that are stored....somewhere. On Intellij you should get lots of notices after installing the plugins

PR #7223, opened 2022-11-03
Add RI and MO to live state map and list

TomNUSDS:
This PR work with Ott to add two new states to the map and list as live.


oslynn (approved):

kevinhaube (approved):
Thanks!

PR #7228, opened 2022-11-03
Metadata Tables Cleanup Proposal

MauriceReeves-usds:
This PR contains a proposal for updating and normalizing our metadata tables. The outcome of this ticket is going to be a ticket that will be for the actual work, once we have solicited feedback from the team.


prime-router/docs/proposals/0017-metadata-tables-cleanup.md
```diff
@@ -0,0 +1,441 @@
...
+    , facility_namespace_id             varchar(512) null
+    , facility_street                   varchar(512) null
+    , facility_street2                  varchar(512) null
+    , facility_city                     varchar(128) null
+    , facility_postal_code              varchar(50) null
+    , facility_callback_number          varchar(512) null
+    , facility_county_id                bigint null references county(county_id)
+    , facility_state_id                 bigint null references state(state_id)
+    , facility_country_id               bigint null references country(country_id)
+    , has_valid_clia                    boolean not null default true
```
GilmoreA6, line 188: how will this be determined? What will be the source of truth for if something is a valid CLIA?. Will we use the existing regex that checks the format or some other source?

prime-router/docs/proposals/0017-metadata-tables-cleanup.md
```diff
@@ -0,0 +1,441 @@
...
+    'The city for the provider';
+comment on column provider.provider_postal_code is
+    'The postal code of the provider';
+comment on column provider.provider_county_id is
+    'The county ID for the provider';
+comment on column provider.provider_state_id is
+    'The state ID for the provider';
+comment on column provider.provider_country_id is
+    'The country ID for the provider';
+comment on column provider.created_at is 'The timestamp for the record creation';
```
GilmoreA6, line 255: what would modify a provider record in this table? Would last_modified ever be something other than created_at?

prime-router/docs/proposals/0017-metadata-tables-cleanup.md
```diff
@@ -0,0 +1,441 @@
...
+comment on column county.county_fips_code is
+    'The numeric code for the county per FIPS';
+comment on column county.state_id is
+    'The foreign key for the state a county is in. Per the Census Bureau no county can cross state lines';
+
+
+-- a facility could exist in multiple roles: ordering facility, testing facility,
+-- but in the future it could also be a servicing facility for vaccines or patient care
+-- at a hospital
+create table facility_role
```
GilmoreA6, line 152: I would consider just folding this into the facility table. I'm not sure we get much advantage from having this as a separate table.

prime-router/docs/proposals/0017-metadata-tables-cleanup.md
```diff
@@ -0,0 +1,441 @@
...
+    , test_set_id                       int not null default 1
+    , reason_for_study_code             varchar(64) null
+    , reason_for_study                  varchar(512) null
+    -- TODO: should we split out the test results to a distinct table? can a test have more than one result?
+    , test_performed_code               varchar(64) not null
+    , test_performed                    varchar(512) not null
+    , test_result_code                  varchar(64) not null
+    , test_result                       varchar(512) not null
+    , test_result_status                varchar(64) not null
+
```
GilmoreA6, line 348: We would want to track ordering provider here as well

prime-router/docs/proposals/0017-metadata-tables-cleanup.md
```diff
@@ -0,0 +1,441 @@
...
+    elr_message_ordered_test_id         bigserial primary key
+    , elr_message_id                    bigint not null references elr_message(elr_message_id)
+    , test_order_id                     varchar(512) not null
+    , test_ordered_code                 varchar(64) not null
+    , test_ordered_long_name            varchar(1024) not null
+    , test_ordered_encoding_version     varchar(64) null
+    , test_set_id                       int not null default 1
+    , reason_for_study_code             varchar(64) null
+    , reason_for_study                  varchar(512) null
+    -- TODO: should we split out the test results to a distinct table? can a test have more than one result?
```
GilmoreA6, line 342: I don't know that we split out results to a distinct table unless it becomes necessary for performance. A test can have definitely have more than one result for the same message id. ex: there are many resp viral panels that test both flu and sars/Covid. where you would have one order (test ordered loinc 95380-2) but four results (test performed LOINCs -94500-6, 94502-2, 92142-9, 92141-1)

prime-router/docs/proposals/0017-metadata-tables-cleanup.md
```diff
@@ -0,0 +1,441 @@
...
+    , patient_age                       varchar(64) null
+);
+
+comment on table elr_message is
+    'Our main ELR message table. This represents a single ORU_R01 message, though the data does not need to necessarily come from HL7.';
+comment on column elr_message.elr_message_id is
+    'A sequential primary key for the message table';
+
+
+-- there can be many AOE questions for a test. let's catch them here
```
GilmoreA6, line 312: Thought: If we have multiple OBRs in a single message and multiple of them have associated AOEs does this become an issue. I don't think so, but worth thinking about.

prime-router/docs/proposals/0017-metadata-tables-cleanup.md
```diff
@@ -0,0 +1,441 @@
...
+
+comment on table country is
+    'A table of countries for the messages we get. These should correspond to the ISO-3166 list';
+comment on column country.country_id is
+    'The primary key for the country table';
+comment on column country.country_name is
+    'The name of the country per ISO-3166, not the official state name value';
+comment on column country.country_abbreviation is
+    'This the alpha-3 code value from HL70399 or ISO-3166, for example USA not US';
+
```
sliu1000, line 102: Is timezone auto generated?

prime-router/docs/proposals/0017-metadata-tables-cleanup/elr-metadata-table-changes.sql
```diff
@@ -0,0 +1,381 @@
...
+insert into facility_role (facility_role) values ('testing');
+insert into facility_role (facility_role) values ('reporting');
+
+
+-- a unified list of facilities
+-- TODO: should we separate out the address from the facility? can a facility have more than one address?
+create table facility
+(
+    facility_id                         bigserial not null primary key
+    , facility_name                     varchar(512) not null
```
sliu1000, line 98: We've had one issue in the past where facility names were too long, and character number allowed was a lot lower than 512.

prime-router/docs/proposals/0017-metadata-tables-cleanup.md
MauriceReeves-usds: @GilmoreA6 - Good question. I think it will be a combination of doing actual lookups at some point on the CLIAs, as well as us setting the values manually for some testing facilities.

prime-router/docs/proposals/0017-metadata-tables-cleanup.md
MauriceReeves-usds: We are trying to keep all date times in UTC in the DB so we don’t have to worry about time zone.

prime-router/docs/proposals/0017-metadata-tables-cleanup.md
brick-green: towards this question: can a facility have more than one address?

ORC-22 is [repeatable](https://hl7-definition.caristix.com/v2/HL7v2.5.1/Fields/ORC.22)

A tricky part is that [ORC-21](https://hl7-definition.caristix.com/v2/HL7v2.5.1/Fields/ORC.21) and [ORC-22](https://hl7-definition.caristix.com/v2/HL7v2.5.1/Fields/ORC.22) are also repeatable. But I don't know if we can rely on senders to ensure that Ordering Facility Name [1] corresponds to Ordering Facility Address [1] and Ordering Facility Phone [1] then Ordering Facility Name [2] corresponds to Ordering Facility Address [2] and Ordering Facility Phone [2]

I think having repeats in these fields is an edge case. More likely is that a facility will change addresses or sometimes the address will be of the specific lab building then other times it will be the general address for the entire facility. The same could be said of the phone number

prime-router/docs/proposals/0017-metadata-tables-cleanup.md
```diff
@@ -0,0 +1,441 @@
...
+    - Create a flag in the database as a configuration setting that controls whether to insert into the legacy table or not 
+3. Create a script that will migrate existing legacy data from the `covid_result_metadata` and `elr_result_metadata` tables into the new table structure
+4. Create two materialized views, one named `covid_result_metadata` and one named `elr_result_metadata` that will exactly replicate the layout of the current tables in order to prevent breakage of any dashboards in Metabase or existing queries
+    - As part of this, the two existing tables will be renamed but preserved for some period of time until we can be sure we have not lost any data
+
+#### Phase 2
+This work will happen as a second step, hopefully in Q1 2023, and reflect a move towards more independent microservices. On ingestion, ReportStream will invoke a separate Azure function that will queue up messages to process, which will then be saved into our metadata tables. The goal is to separate the concerns of processing from the goal of tracking metadata from the messages we receive.
```
oslynn, line 59: Love this idea of moving towards a microservices architecture.  Perhaps, we can take this chance to refactor and introduce the SOLID Principle (Single Responsibility, Open and Closed, Liskov Substitution, Interface Segregation, Dependency Inversion) into our class implementation.   Also, we need to update most of our unit and integration tests as well.

prime-router/docs/proposals/0017-metadata-tables-cleanup.md
```diff
@@ -0,0 +1,441 @@
...
+
+
+## Further Work
+Once feedback has been provided about the structure of the tables and the plan above, any proposed changes will go through an acceptance/rejection process, and then the actual changes will be put into a PR for a final review. This document has been created on 3 November 2022, and will be reviewed at the developer sync on Monday, 7 November 2022 with a targeted proposal acceptance date of 14 November 2022. At that point the work will begin on Phase 1 with an expected completion date of 21 November 2022. Start and completion date for Phase 2 TBD.
+
+It should also be noted that this work is not to make a system of record for the CDC or for our STLT partners. We are still only acting as a hub for the transferring of data. That said, having the ability to provide usage and demographic data to our senders and receivers means we provide a richer and better experience for our users, and it empowers our support staff.
+
+## Appendix 1
+
+```postgresql
```
clediggins-usds, line 78: Is there a definitive set of requirement we are trying to meet here? I know the covid metadata table is used and gives guidance on that, but the elr metadata table is not used and may not have all the data / may have extra data

prime-router/docs/proposals/0017-metadata-tables-cleanup.md
jimduff-usds: Yeah - I'm not seeing any FK to provider - maybe this got missed.

@MauriceReeves-usds , For both FK to provider and to facility, shouldn't that be in the elr_message table, and not here?   Like, it feels like a single ELR message will always be from a single facility.  No?

prime-router/docs/proposals/0017-metadata-tables-cleanup.md
```diff
@@ -0,0 +1,441 @@
...
+
+-- add our primary key
+alter table elr_message_specimen
+    add constraint elr_message_specimen_pk
+        primary key (elr_message_id, test_order_id, specimen_id)
+;
+
+
+-- the facilities for an elr message
+create table elr_message_facilities
```
jimduff-usds, line 420: BLAH.   This seems way too complicated.  One message has multiple facilities?  BLAH.

prime-router/docs/proposals/0017-metadata-tables-cleanup/elr-metadata-table-changes.sql
```diff
@@ -0,0 +1,381 @@
+/**
```
jimduff-usds, line 1: Oops, I didn't realize the SQL was all repeated in the proposal and in a .sql doc.

jimduff-usds (commented):
This is a massive and excellent effort - THANK YOU., THANK YOU.

I have two additional big comments:  

1. Currently we only insert into covid_result_metadata for _incoming_ data.  I think we should also insert metadata for the _outgoing_ metadata as well.    Then when the state calls us up about a report they got, we can immediately see the metadata we sent them, without having to go thru any lineage query contortions.
2. We already have problems that folks don't understand our schema.   That problem will get much worse.

chris-kuryak:
@LaFreda1 Do you know where we stand on this?

chris-kuryak:
@arnejduranovic Are we moving off of Metadata tables? Can we close this ticket?

arnejduranovic:
Closing because we do not expect to merge this anytime soon. Waiting on product requirements to determine future of metadata storage.

PR #7302, opened 2022-11-08
Cle/7044 qual filter

clediggins-usds:
This PR adds in the ability to add FHIRPath filter expressions for receivers and organizations, as well as implements defaults for those filters in the Universal Pipeline that matches the functionality present in the covid pipeline

Test Steps:
1. Run tests
OR
2. Set up a receiver and sender that hits the UP
3. Add fhirpath jurisFilter
4. run a full-elr message through the UP
5. verify the destinations are correct

## Changes
- Add some basic fhirPath filter expressions to organizations.yml for full elr receiver
- Update unit test constructors for Receiver and Organization to accept filters
- Add default filter experssions to FhirRouter
- implement filter checking in sequential order in fhirRouter
- implement updating of fhir bundle to add destinations
- implement getting proper combination of receiver/org filteres in fhirrouter
- add tests for all of it
- add .fhir test files

### Testing
- [x] Tested locally?
- [x] Added tests?


prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
```diff
@@ -183,21 +198,139 @@ class FHIRRouter(
...
+        val processingModeDefaultResult: Boolean by lazy {
+            evaluateFilterCondition(
+                processingModeFilterDefault, bundle,
+                false
+            )
+        }
+
+        fullElrReceivers.forEach { receiver ->
+            // get the receiver's organization, since we need to be able to find/combine the correct filters
+            val orgFilters = settings.findOrganization(receiver.organizationName)!!.filters
```
jimmyfagan, line 231: (Question) Hadn't heard of `!!` before seeing this, why do we want a null pointer exception here in the case where we don't find an organization? How is that surfaced?

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
```diff
@@ -183,21 +198,139 @@ class FHIRRouter(
...
+            passes = passes &&
+                evaluateFilterCondition(
+                    getProcessingModeFilter(receiver, orgFilters), bundle,
+                    processingModeDefaultResult
+                )
+
+            // if all filters pass, add this receiver to the list of valid receivers
+            if (passes) {
+                listOfReceivers.add(receiver)
+            }
```
jimmyfagan, line 262: (Optional) Since we're breaking this down already, both readability and debug-ability could maybe be further enhanced by having separate variables for `jurisFilterPasses`, `qualityFilterPasses`, `routingFilterPasses`, and `modeFilterPasses`, and either `and`ing them at the end to evaluate `passes`, or making the "if all filters pass" condition be `jurisFilterPasses && qualityFilterPasses && routingFilterPasses && modeFilterPasses` and not using `passes` at all. Not familiar with logging norms, but that would also allow us to easily log which filter conditions passed or failed. Just an idea, feel free to keep it as it is.

prime-router/src/test/kotlin/fhirengine/engine/fhirRouterTests/BundleUpdateTests.kt
```diff
@@ -0,0 +1,156 @@
...
+                                "system": "http://terminology.hl7.org/CodeSystem/v2-0003",
+                                "code": "R01",
+                                "display": "ORU_R01"
+                            }
+                        ]
+                    }
+                }
+            }
+        ]
+    }"""
```
jimmyfagan, line 100: (Question) Do we manually generate FHIR and other input for tests, or do we pull those from somewhere?

prime-router/src/test/kotlin/fhirengine/engine/fhirRouterTests/RoutingTests.kt
```diff
@@ -0,0 +1,348 @@
...
+        every { accessSpy.insertTask(any(), bodyFormat.toString(), bodyUrl, any()) }.returns(Unit)
+        every { actionHistory.trackCreatedReport(any(), any(), any()) }.returns(Unit)
+        every { actionHistory.trackExistingInputReport(any()) }.returns(Unit)
+        every { queueMock.sendMessage(any(), any()) }
+            .returns(Unit)
+        every { FHIRBundleHelpers.addReceivers(any(), any()) } returns Unit
+        every { engine.getJurisFilters(any(), any()) } returns jurisFilter
+        every { engine.getQualityFilters(any(), any()) } returns qualFilter
+        every { engine.getRoutingFilter(any(), any()) } returns routingFilter
+        every { engine.getProcessingModeFilter(any(), any()) } returns procModeFilter
```
jimmyfagan, line 186: Not sure exactly how the mocking works, but would this still work if we extracted a helper function (something like `setUpMockResponses`) for setting these up so we could use that instead of having this copy/pasted in each test? I can try to test that out after finishing the review, will post an update if I do.

jimmyfagan (approved):
Thanks for adding me to the review! Left a handful of questions and comments, nothing that should block merging though, so marking approved.

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
```diff
@@ -39,6 +42,39 @@ class FHIRRouter(
...
+    )
+
+    /**
+     * Default Rule:
+     *  Must have a processing mode id of 'P'
+     */
+    val processingModeFilterDefault: ReportStreamFilter = listOf(
+        "Bundle.entry.resource.ofType(MessageHeader).meta" +
+            ".extension('https://reportstream.cdc.gov/fhir/StructureDefinition/source-processing-id')" +
+            ".value.coding.code = 'P'"
```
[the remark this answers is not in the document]
clediggins-usds: In this context, P is 'production'. The valid values per HL7 spec are 'D', 'T', and 'P' representing 'Debug', 'Test', and 'Production' respectively. The default filter for processing mode is to only allow 'production' data to pass through, but it can be overridden during onboarding to allow test data thought.

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
```diff
@@ -39,6 +42,39 @@ class FHIRRouter(
     queue: QueueAccess = QueueAccess,
 ) : FHIREngine(metadata, settings, db, blob, queue) {
 
+    /**
+     * Default Rules:
+     *   Must have message ID, patient last name, patient first name, DOB, specimen type
+     *   At least one of patient street, patient zip code, patient phone number, patient email
+     *   At least one of order test date, specimen collection date/time, test result date
+     */
+    val qualityFilterDefault: ReportStreamFilter = listOf(
```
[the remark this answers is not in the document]
clediggins-usds: The default use cases are matching the current defaults that are used in the covid/monkeypox pipeline. I am not certain of the source of them, but they do seem to make contextual sense.

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
clediggins-usds: settings.findOrganization(<name>) is used in a number of places. In this case we are using a defined name that we know will not be null, so use !! to indicate to kotlin that it is not nullable even though the return type is. This same underlying method is used during authentication to look for the organization contained within the client header of the request. Since that is user-provided input it may return null.

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
clediggins-usds: Originally I had it as you suggest, but in this case it is fall-through logic - this filters are in a particular order and if any filter higher up fails the ones below will not be run since it would short circuit past them. We don't want to run expensive complicated filters on a Bundle if it doesn't pass, for example, jurisdictional filter. In this code we are running ALL potential receivers against a bundle coming through so want to do the least amount of work necessary.

prime-router/src/test/kotlin/fhirengine/engine/fhirRouterTests/BundleUpdateTests.kt
clediggins-usds: Whatever works best for you. I generated these by running an HL7 message through the universal pipeline, downloading the resulting FHIR, and picking the parts I wanted.

prime-router/src/test/kotlin/fhirengine/engine/fhirRouterTests/RoutingTests.kt
clediggins-usds: We could do some more advanced mocking, but we typically try to NOT share much between tests since each is intended to be a completely separate unit test and we've been bitten before by having shared setup break unintended tests when a business rule change for a single use case comes through.

clediggins-usds: There is also some funky shared resources and parallelism going on when we run our tests in github actions that can negatively impact other tests if there is shared code.

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
jimmyfagan: Ah gotcha, so basically `!!` is used to label something as non-nullable, and if it turns out it is null it'll throw an exception, but the exception is not the reason for using `!!` generally, the exception is just kind of a necessary part of handling it that way. And we can't change the method return type to be non-null, because there are other cases where it should be null. I think I understand. Even knowing it'll not be null, I could see us adding a quick catch as a safety net, but if you're not worried about it, then that's probably overkill.

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
jimmyfagan: Ah nice, I overlooked that short-circuiting, I like it.

prime-router/src/test/kotlin/fhirengine/engine/fhirRouterTests/RoutingTests.kt
jimmyfagan: Ah fair enough, good to know!

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
```diff
@@ -183,21 +198,139 @@ class FHIRRouter(
...
+            // Get the applicable filters, either receiver or organization level if there are no receiver filters
+
+            // NOTE: these could all be combined into a single `if` statement, but that would reduce readability and
+            // ability to debug this section so is being left split out like this
+
+            // JURIS FILTER
+            //  default: allowNone
+            var passes = evaluateFilterCondition(getJurisFilters(receiver, orgFilters), bundle, false)
+            // QUALITY FILTER
+            //  default: must have message id, patient last name, patient first name, dob, specimen type
```
carlosfelix2, line 242: Remove this comment from here and add it where the default is defined.

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
```diff
@@ -183,21 +198,139 @@ class FHIRRouter(
...
+            //           must have at least one of patient street, zip code, phone number, email
+            //           must have at least one of order test date, specimen collection date/time, test result date
+            passes = passes &&
```
carlosfelix2, line 245: This is SUPER EASY to read.  Thank you!

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
```diff
@@ -183,21 +198,139 @@ class FHIRRouter(
...
+        bundle: Bundle,
+        defaultResponse: Boolean
+    ): Boolean {
+        // the filter needs to check all expressions passed in, or if the filter is null or empty it will return the
+        // default response
+        return if (filter.isNullOrEmpty())
+            defaultResponse
+        else
+            filter.all {
+                FhirPathUtils.evaluateCondition(CustomContext(bundle, bundle), bundle, bundle, it)
```
carlosfelix2, line 284: FYI: This will be where you pass in your constants (in the CustomContext) once you read them.

prime-router/src/test/resources/fhirengine/engine/routerDefaults/qual_test_0.fhir
```diff
@@ -0,0 +1,2017 @@
+{
```
carlosfelix2, line 1: Is there any way to rename these test files to have something more descriptive on the name, or perhaps a README.md to just list what the different files test?

carlosfelix2 (commented):
Still need to test it...

prime-router/settings/organizations.yml
```diff
@@ -789,9 +789,12 @@
   filters: # default filters applied for all receivers of this organization, if not overridden by receiver filter
     - topic: "full-elr"
       jurisdictionalFilter:
+        # this is equivalent to orEquals(ordering_facility_state, CO, patient_state, CO)
```
carlosfelix2, line 792: Here is a multiline YAML that works:
```
    - topic: "full-elr"
      jurisdictionalFilter:
        # this is equivalent to orEquals(ordering_facility_state, CO, patient_state, CO)
        - >-
          (Bundle.entry.resource.ofType(ServiceRequest)[0].requester.resolve().organization.resolve().address.state.exists() and 
          Bundle.entry.resource.ofType(ServiceRequest)[0].requester.resolve().organization.resolve().address.state = 'CO') or 
          (Bundle.entry.resource.ofType(Patient).address.exists() and Bundle.entry.resource.ofType(Patient).address.state = 'CO')
        # TODO If we had constants defined, this is how this would look
#        - "%performerState = 'CO' or %patientState = 'CO'"
      qualityFilter:
        - >- 
          Bundle.entry.resource.ofType(Provenance).exists() and 
          Bundle.entry.resource.ofType(Provenance).activity.coding.code = 'R01'
```

This is what it looks like in the DB:
```
"filters": [
    {
      "topic": "full-elr",
      "qualityFilter": [
        "Bundle.entry.resource.ofType(Provenance).exists() and  Bundle.entry.resource.ofType(Provenance).activity.coding.code = 'R01'"
      ],
      "routingFilter": null,
      "jurisdictionalFilter": [
        "(Bundle.entry.resource.ofType(ServiceRequest)[0].requester.resolve().organization.resolve().address.state.exists() and  Bundle.entry.resource.ofType(ServiceRequest)[0].requester.resolve().organization.resolve().address.state = 'CO') or  (Bundle.entry.resource.ofType(Patient).address.exists() and Bundle.entry.resource.ofType(Patient).address.state = 'CO')"
      ],
```

carlosfelix2 (approved):

PR #7316, opened 2022-11-09
Allow test result in any OBX iteration

### LINKED ISSUE #6733: Handle result in any OBX Set ID
## Problem statement

Currently ReportStream requires that the COVID-19 result appears in the first OBX instance (typically Set ID = 1). Providence results may appear in any Set ID. While atypical for our senders, this is valid and allowed by the HL7 spec. So we need to be able to accommodate this.

This is a blocker for full onboarding of Providence.

## What you need to know

Examples are in keybase under Providence.

- sender_info/Providence/2022-08-26/ELR_Raw_message.txt
- sender_info/Providence/2022-08-26/POC_Raw_message.txt

Note this samples will generate some other warning and filter issues. The related warning is "The code N for field OBX-5 is invalid."

You can also reorder the OBX segments in a working message to achieve the same outcome.

## Acceptance criteria

- [ ] Message with a result appearing in an OBX instance other than the first is successful.

brick-green:
This PR adds functionality to accept HL7 messages where the test result is not in the first observation iteration. HL7 does not require a certain order for the observations. But currently RS assumes the first OBX segment will contain the test result. These changes checks the first OBX. If the LOINC code in that OBX-3 field is not present in the LIVD table "Test Performed LOINC Code" column, then the remaining OBX segments are searched. If an OBX is found with a OBX-3 value in the LIVD table then that observation group is moved the top of the grouping. If an OBX is not identified then the message is not changed and moves forward in the process. 

Test Steps:
1. pull and package
2. `./prime test`
3. `./gradlew testIntegration`
4. Send test file
     - `curl -X POST -H "client: neurolab" -H "Content-Type: application/hl7-v2" --data-binary "PATH/TO/test_file.txt" "http://localhost:7071/api/reports"`
     - [test_file.txt](https://github.com/CDCgov/prime-reportstream/files/9974956/test_file.txt)
5. Check output HL7 to ensure that the test result observation group(OBX and NTE) was moved to the beginning of the list and that the OBX Set IDs were re-numbered

## Changes
- add function organizeObservationOrder()
- add function checkLIVDValueExists()

### Testing
- [X] Tested locally?
- [X] Added `test checkLIVDValueExists` to HL7SerializerTests
- [X] Added `test confirmObservationOrder` to HL7SerializerTests

## Linked Issues
- Fixes #6733


MauriceReeves-usds (approved):
This looks good to me. I like the approach you took with checking for the LOINC value. Did you check this with a monkeypox message? We don't use the LIVD table for that, so I'm curious what this would do in that case.

brick-green:
> This looks good to me. I like the approach you took with checking for the LOINC value. Did you check this with a monkeypox message? We don't use the LIVD table for that, so I'm curious what this would do in that case.

I did not try it with a MPX message but I did try it with a message with no test result. It only had AOE questions. In that case it left the message untouched and let it go along with processing.

PR #7337, opened 2022-11-10
optimize demo environment provisioning

JosiahSiegel:
This PR optimizes the demo environment provisioning process by removing unnecessary redundancy.

## Changes
- Do not restore db if databases already exist and they are recently restored.
- Remove invalid Terraform "ignore_changes"
- fix dns host file for demo1


jeremy-page (approved):
Awesome, thank you

Issue #5702, opened 2022-06-07
Tech Debt: Pull Serializers out of BlobAccess


clediggins-usds:
As a ReportStream developer, I want BlobAccess to solely have functions for accessing blob storage so I have more testable code.

### Description/Use Case
To make blob access easily mockable and follow SRP, it should not know anything about serializer. Pull the serializers out of the class and pass them in, make BlobAccess just access the blob - write bytestreams / read bytestream

### Risks/Impacts/Considerations
* Depending on how many places this is located, a 'get report bytes' function may need to be written in a utility class.

### Dev Notes:
* Add tests for any new utility functions created

### Acceptance Criteria
* BlobAccess does not know about serializers, and does not reference them.

PR #7346, opened 2022-11-13
5702 Pull Serializers out of BlobAccess

TurkeyFried:
This PR takes the report serializing functionality out of BlobAccess.

Test Steps:
1. Verify all tests pass
2. Manually submit reports, and verify they are stored as blobs in your Azure Storage

## Changes
- Serialization logic that used to be on `BlobAccess` is now in the newly created class `ReportWriter`
- Added and updated test cases to improve coverage
- Updated references to BlobAccess where it was needed to maintain the same functionality.

## Linked Issues
- #5702


prime-router/src/main/kotlin/azure/BlobAccess.kt
```diff
@@ -39,66 +34,29 @@ class BlobAccess(
...
 
     /**
-     * Upload the [report] to the blob store using the [action] to determine a folder as needed.  A [subfolderName]
-     * is optional and is added as a prefix to the blob filename.
+     * Upload the [report] to the blob store using the [action] to determine a folder as needed.
+     * A [subfolderName] is optional and is added as a prefix to the blob filename.
      * @return the information about the uploaded blob
      */
-    fun generateBodyAndUploadReport(
+    fun uploadReport(
```
carlosfelix2, line 49: I think this function can be moved to the companion object

prime-router/src/main/kotlin/azure/BlobAccess.kt
```diff
@@ -21,10 +19,7 @@ import java.security.MessageDigest
 
 const val defaultBlobContainerName = "reports"
 
-class BlobAccess(
-    private val csvSerializer: CsvSerializer? = null,
-    private val hl7Serializer: Hl7Serializer? = null
-) : Logging {
+class BlobAccess : Logging {
```
carlosfelix2, line 22: There is a possibility this can now be an object class, but it would suck to change all the references throughout the code.

prime-router/src/main/kotlin/azure/ReportWriter.kt
```diff
@@ -0,0 +1,43 @@
+package gov.cdc.prime.router.azure
+
+import gov.cdc.prime.router.Report
+import gov.cdc.prime.router.common.BaseEngine
+import gov.cdc.prime.router.serializers.CsvSerializer
+import gov.cdc.prime.router.serializers.Hl7Serializer
+import java.io.ByteArrayOutputStream
+
+class ReportWriter {
+    companion object {
```
carlosfelix2, line 10: Make this into an object class and remove companion object

carlosfelix2 (changes requested):

prime-router/src/main/kotlin/azure/WorkflowEngine.kt
```diff
@@ -259,7 +261,7 @@ class WorkflowEngine(
      */
     fun handleReportEvent(
         messageEvent: ReportEvent,
-        updateBlock: (header: Header, retryToken: RetryToken?, txn: Configuration?) -> ReportEvent,
+        updateBlock: (header: Header, retryToken: RetryToken?, txn: Configuration?) -> ReportEvent
```
jimmyfagan, line 264: Here and in a lot of other spots, the formatting change is removing the trailing comma. From the [Kotlin coding conventions](https://kotlinlang.org/docs/coding-conventions.html#trailing-commas):

> The Kotlin style guide encourages the use of trailing commas at the declaration site and leaves it at your discretion for the call site.

I personally prefer trailing commas wherever allowed by the language, but it sounds like particularly for function definitions, the standard is to keep them. It would be nice to get to the bottom of what is leading to these formatting changes, whether it is `ktlint` or some other general formatter. Unless there's a project specific standard established somewhere, it seems like we should have settings that match the published Kotlin conventions. This is outside the scope of the issue though, so no need to get too hung up on it right now.

jimmyfagan (commented):
Just had one comment related to the formatting changes, overall the changes look good to me as far as I can tell, but I need to look up some of the Kotlin elements that are coming into play here like `companion` and `object`. Thanks for adding me to the review!

prime-router/src/main/kotlin/azure/ReportWriter.kt
TurkeyFried: Done

prime-router/src/main/kotlin/azure/WorkflowEngine.kt
TurkeyFried: I'm not sure when our ktlint was last updated.. I'm gonna check that, might be work making a ticket to refresh it

prime-router/src/main/kotlin/azure/BlobAccess.kt
TurkeyFried: Doing this requires a couple changes, may be better to save for later

prime-router/src/main/kotlin/azure/BlobAccess.kt
TurkeyFried: I wanna do that but I'm sure that would end up delaying the PR quite a bit...

PR #7384, opened 2022-11-16
Use Topic enum on Receiver

### LINKED ISSUE #5282: Use Topic enum in Receiver
### Description/Use Case
In the work for #5050 a new enum was introduced to track Topics. It currently has two values - FULL_ELR and COVID_19. Receiver should also use this enum instead of the String topic value.

* Tests may need to be updated.

### Dev Notes:
* All constructors for Receiver will need to be updated
* Tests may need to be updated
* The enum serializes/deserializes properly to the existing string "covid-19" so that should not be a properly

### Acceptance Criteria
* Receivers no longer use the string "covid-19" as a topic value.
* Full pipeline still works

jimmyfagan:
This PR uses the existing `Topic` enum to replace several hard-coded topic strings in `Receiver` constructors throughout the code. Since there were also some `Receiver` constructors used in tests with values other than `monkeypox`, `covid-19`, and `full-elr`, I also added two new enum values to `Topic` for `TEST` and `CSV-TESTS`. Doing so allowed the existing tests to run basically the same way as before; if it is preferable to remove those and restructure some of the associated tests, I can go back to try to do that.

Test Steps:
1. Run unit tests
2. Run integration tests
3. Run Postman requests
4. Run smoke tests

## Changes
- Added new `Topic.TEST` and `Topic.CSV-TESTS`
- Updated `Receiver.topic` to be a `Topic` rather than a `String`; replaced hard coded String values with enums
- Changed `Schema.topic` and `ReportStreamFilters.topic` to also take `Topic` values (updated a few schema files to match our enum options
- There were some automatic formatting changes, I've decided to keep those for now, and we can determine separately how to sync up on those moving forward.

## Linked Issues
Resolves #5282


prime-router/src/main/kotlin/ReportStreamFilter.kt
```diff
@@ -105,7 +106,6 @@ data class ReportStreamFilters(
         val defaultFiltersByTopic: Map<String, ReportStreamFilters> = mapOf(
             defaultCovid19Filters.topic to defaultCovid19Filters,
             defaultMonkeypoxFilters.topic to defaultMonkeypoxFilters,
-            defaultCsvFileTestFilters.topic to defaultCsvFileTestFilters,
```
jimmyfagan, line 108: Not sure how this happened, or how it didn't break a test, but I'll add it back in and try to figure that out.

prime-router/src/main/kotlin/ReportStreamFilter.kt
```diff
@@ -68,31 +69,31 @@ data class ReportStreamFilters(
             "isValidCLIA(testing_lab_clia,reporting_facility_clia)",
         )
         private val defaultCovid19Filters = ReportStreamFilters(
-            topic = "covid-19",
+            topic = Topic.COVID_19,
```
carlosfelix2, line 72: Oh, so nice!  Thx

prime-router/src/main/kotlin/SettingsProvider.kt
```diff
@@ -52,5 +52,11 @@ enum class Topic(val json_val: String) {
     COVID_19("covid-19"),
 
     @JsonProperty("monkeypox")
-    MONKEYPOX("monkeypox")
+    MONKEYPOX("monkeypox"),
+
+    @JsonProperty("CsvFileTests-topic")
```
carlosfelix2, line 57: I wonder if this is even needed.  Can you ask someone like Jim or Mo?

carlosfelix2:
On your description, the test section is not what you ran, but what you are telling others to run for testing.  Obviously you run this too.

Include to run the smoke tests.

carlosfelix2 (approved):

prime-router/src/main/kotlin/SettingsProvider.kt
jimmyfagan: @MauriceReeves-usds Any thoughts on this? In short, do we need/want to keep separate default filters and tests for topic `"CsvFileTests-topic"`?

MauriceReeves-usds: I am not sure on this. This might be related to the CSV uploader, which means we may not need it now. Go ahead and leave it in, but make a tech debt ticket to investigate if we need this topic, and if we don't, remove it. That way it's tracked, and can be worked, but the rest of your changes can move forward.

prime-router/src/main/kotlin/cli/main.kt
```diff
@@ -226,7 +227,7 @@ class GenerateDocs : CliktCommand(
         }
         return Schema(
             fromSchema.baseName,
-            "covid-19",
+            Topic.COVID_19,
```
MauriceReeves-usds, line 230: This is not your bug @jimmyfagan but looking at this, the topic shouldn't be hardcoded here when generating schema. We should pull the topic from the schema we're generating documentation for. If we are creating a new schema to write documentation for (not even sure why we are building new) then we should copy the topic from schema. Can you fix this as part of your work here please?

MauriceReeves-usds (changes requested):
@jimmyfagan great change and overdue. Thank you for accomplishing this. I saw one bug that your PR uncovered. Not your fault, but since you're in that area, if you can fix it, that would be good. See my response to your question as well. Great job!

prime-router/src/main/kotlin/cli/main.kt
jimmyfagan: Ah good call, fixed now, thanks for catching that!

prime-router/src/main/kotlin/SettingsProvider.kt
jimmyfagan: Sounds good, will do, thanks!

jimmyfagan:
Small update here, the original description said that I did _not_ update the type of `Schema.topic` or `ReportStreamFilters.topic`, but at this point, those have been updated too, and all tests still pass. Also have addressed all code review feedback so far, so this is ready for re-review if anybody has been holding off. Thanks!

prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -361,7 +365,7 @@ class ActionHistory(
         // check that we're dealing with an external file
         val clientSource = report.sources.firstOrNull { it is ClientSource }
         if (clientSource != null) {
-            when (report.schema.topic.lowercase()) {
+            when (report.schema.topic.json_val.lowercase()) {
```
clediggins-usds, line 368: This should be using the enum instead of json_val and lowercasing it.

prime-router/src/main/kotlin/cli/main.kt
```diff
@@ -226,7 +226,7 @@ class GenerateDocs : CliktCommand(
         }
         return Schema(
             fromSchema.baseName,
-            "covid-19",
+            fromSchema.topic,
```
clediggins-usds, line 229: This is a functional change. @carlosfelix2 are you familiar enough with this part of the CLI to confirm this will have no adverse effects?

clediggins-usds (changes requested):

prime-router/src/main/kotlin/cli/main.kt
jimmyfagan: The comment was hidden since it is outdated, but this functional change was in response to Mo's comment [here](https://github.com/CDCgov/prime-reportstream/pull/7384#discussion_r1026503480), so I think we should be okay, but Carlos, still feel free to confirm that it's alright.

prime-router/src/main/kotlin/azure/ActionHistory.kt
jimmyfagan: Good call, updated in latest commit.

jimmyfagan:
Seems like the bot might keep complaining because there aren't updates to most of the schema files, but we should be correct at this point, so I'll just ignore that. Only expecting updates to 3 schema documentation files, the rest should be the same as before.

prime-router/src/main/kotlin/cli/main.kt
clediggins-usds: Just saw Mo's comment. Looks good.

clediggins-usds (approved):

jimmyfagan:
Had to do a force push since I had unsigned commits on the branch without realizing it and that was the simplest way I could go back to fix it. Did a compare to make sure it matched the reviewed code, so shouldn't need an additional review.

Issue #7047, opened 2022-10-18
reverseQualityFilter property in receiver settings negates the result of the quality filter in the Universal Pipeline


carlosfelix2:
### User Story:
As a ReportStream Administrator, I want to be able to set reverse the value from a quality filter by setting a flag, so I don't have to develop a complex negation of an existing filter.

### Description/Use Case
The COVID pipeline has a receiver flag called `reverseQualityFilter` that simply takes the result of the quality filter and negates its value.  This was done because it was complicated to develop a negation logic for the custom filters.  For the UP, we are using FHIR paths, which has the negation logic, but we want to support this flag at least for now since the OO team is used to it.  We may deprecate this in the future.

### Risks/Impacts/Considerations
* This could have unintended consequences when used with the default filter, but does still need to work for default filter.

### Dev Notes:
* Start in FhirRouter.kt, see approximately line 245
* The Receiver object has the property 'reverseTheQualityFilter' which will already be set, and the receiver object is already present in the FHIRRouter where the filters are being checked.

### Acceptance Criteria
* A UP receiver that has the reverseQualityFilter flag set to true negates the result of the quality filter for the given receiver (which could be an organization or default filter)

PR #7388, opened 2022-11-16
7047 - Added functionality to support the reverseQualityFilter flag

victor-chaparro:
Added functionality to support the reverseQualityFilter flag for receivers

Test Steps:
1. Submit a Full ELR HL7 using reports API - /api/reports?processing=async&default=aaa
2. Capture the Id of the submitted report
3. Verify Full ELR HL7 destination using report history API - /waters/report/{{reportId}}/history
4. Capture the receiver being routed
5. Update the reverseTheQualityFilter flag on this receiver inside organizations.yaml
6. Run ./gradlew reloadSettings to update settings for this receiver
7. Repeat steps 1 and 2
8. Verify the destination the previous destination is not present when calling the report history API

## Changes
- Added optional parameter reverseFilter to the evaluateFilterCondition
- Passing reverseTheQualityFilter flag when evaluating quality filters

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- Fixes #issue

## To Be Done
- #7047


jimmyfagan (approved):
Looks good to me, just had one comment on the topic of interaction between `reverseFilter` and `defaultResponse` to consider, either way, marking approved.

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
```diff
@@ -243,7 +245,12 @@ class FHIRRouter(
             //           must have at least one of patient street, zip code, phone number, email
             //           must have at least one of order test date, specimen collection date/time, test result date
             passes = passes &&
-                evaluateFilterCondition(getQualityFilters(receiver, orgFilters), bundle, qualFilterDefaultResult)
+                evaluateFilterCondition(
+                    getQualityFilters(receiver, orgFilters),
+                    bundle,
+                    qualFilterDefaultResult,
+                    receiver.reverseTheQualityFilter
```
carlosfelix2, line 252: You also need a unit test to test that the receiver's config is used correctly here.

carlosfelix2 (changes requested):

carlosfelix2:
Set the label of this PR to pipeline

carlosfelix2:
The test steps section is not what you ran, but what you are telling others to run.  Obviously you ran this too.

carlosfelix2:
The smoke tests are not really testing this new feature you added.  Perhaps add steps to change the reverse filter of a setting to manually test the flag works.  Hint: submit a full elr HL7, see what receiver is goes to, then change the flag in the DB or settings, send again, then see the data not go to that receiver.

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
```diff
@@ -273,16 +281,19 @@ class FHIRRouter(
     internal fun evaluateFilterCondition(
```
TurkeyFried, line 281: The doc for this method could be updated to explain the new param

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
```diff
@@ -273,16 +281,19 @@ class FHIRRouter(
...
         bundle: Bundle,
-        defaultResponse: Boolean
+        defaultResponse: Boolean,
+        reverseFilter: Boolean = false
     ): Boolean {
         // the filter needs to check all expressions passed in, or if the filter is null or empty it will return the
         // default response
-        return if (filter.isNullOrEmpty())
+        val result: Boolean = if (filter.isNullOrEmpty()) {
             defaultResponse
```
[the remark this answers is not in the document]
jimmyfagan: Not the filters, the default _response_; the caller is presumably providing a default response to be "what do we do if there isn't a filter to use", so I don't think we should negate that.

TurkeyFried (changes requested):
The code changes look good. My only comment is that the PR needs to either have manual testing steps or the code needs a Smoke (or Integration) test.

TurkeyFried (approved):
The updated test steps are good. I also saw this working on a pair earlier so thumbs up!

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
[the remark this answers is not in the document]
jimmyfagan: Discussed this during standup today, and it sounds like the default response should be negated too, so this is doing what is wanted as it is; I'll close this conversation.

prime-router/src/test/kotlin/fhirengine/engine/fhirRouterTests/DefaultFilterTests.kt
```diff
@@ -172,6 +170,27 @@ class DefaultFilterTests {
...
+        val one = Schema(name = "None", topic = "full-elr", elements = emptyList())
+        val metadata = Metadata(schema = one)
+
+        val engine = spyk(makeFhirEngine(metadata, settings, TaskAction.route) as FHIRRouter)
+        val bundle = FhirTranscoder.decode(fhirCodeP)
+
+        // act
+        val useDefaultResult = engine.evaluateFilterCondition(emptyList(), bundle, false, true)
+        // assert
+        assert(useDefaultResult)
```
carlosfelix2, line 186: Use assertThat instead in all your tests.  This is the program directed unit test framework. Assert is now only used if you cannot use assertThat which is rare.  See assertk

carlosfelix2 (changes requested):

prime-router/src/test/kotlin/fhirengine/engine/fhirRouterTests/DefaultFilterTests.kt
jimmyfagan: Is that requirement meant to improve readability or is there some other reason? Is there any kind of document or decision record that we could reference to have a better idea of these kinds of standards, or is it more of an informal shared understanding at this point?

victor-chaparro: Updated unit tests to user assertThat. Ran unit tests after making these changes and they're all passing.

prime-router/src/test/kotlin/fhirengine/engine/fhirRouterTests/DefaultFilterTests.kt
```diff
@@ -170,7 +171,28 @@ class DefaultFilterTests {
...
+        val engine = spyk(makeFhirEngine(metadata, settings, TaskAction.route) as FHIRRouter)
+        val bundle = FhirTranscoder.decode(fhirCodeP)
+
+        // act
+        val useDefaultResult = engine.evaluateFilterCondition(emptyList(), bundle, false, true)
+        // assert
+        assertThat(useDefaultResult).isTrue()
+
+        // act
+        val useBundleResult = engine.evaluateFilterCondition(engine.processingModeFilterDefault, bundle, false, false)
```
carlosfelix2, line 193: Not asking you to change this right now, but there is no need for the useBundleResult val.  Just put the code inside the assertThat().  Same for all the other below.  I know you did not write those.

prime-router/src/main/kotlin/Receiver.kt
```diff
@@ -87,7 +87,8 @@ open class Receiver(
         jurisdictionalFilter: ReportStreamFilter = emptyList(),
         qualityFilter: ReportStreamFilter = emptyList(),
         routingFilter: ReportStreamFilter = emptyList(),
-        processingModeFilter: ReportStreamFilter = emptyList()
+        processingModeFilter: ReportStreamFilter = emptyList(),
+        reverseTheQualityFilter: Boolean = false
```
carlosfelix2, line 91: Update the organizations.yml to add `reverseTheQualityFilter: false` to the "full-elr-hl7" receiver

carlosfelix2 (approved):

Issue #7409, opened 2022-11-18
Ensure that all azure functions that pull items off the queue and TASK table confirm that there is no next_action mismatch


jimduff-usds:
We have seen cases with the old `process` azure function where the item in the queue says to 'process' but the row in the TASK table says next_action == 'NONE', that is, its already been processed.

This implies there is a rare race condition problem in the way we are using the queues, wherein things can get processed twice.

For all Azure Functions that are queue triggered, this can be _detected_ by having the Azure Function do a sanity check that the TASK next_action correctly matches the Event.

This change has already been made in the PR to fix #7404.  The code change example is found in WorkflowEngine.kt lines 553 to 564 (at the time of this writing).

This ticket is to simply doublecheck the other azure functions, and add a similar logging/fix as needed.

Yes, you are correct that this does not fix the race condition, but it does workaround it.

bishoyayoub:
Needs to be updated with template.

PR #7410, opened 2022-11-18
Abort ProcessFunction if the TASK table and queue entry don't match

### LINKED ISSUE #7404: Prevent re-'process'ing reports that have already been processed.
It appears that work can sometimes stay on our 'process' queue even though that work has already been completed.

This ticket does not fix the problem that the work item is not properly removed from the queue once its complete.

It does however prevent the bug from causing damage.

The fix is to take advantage of the fact that we track work in two places, thank goodness, in both the queue, and in the TASK table.

If the ProcessFunction begins to work on a task, and if that task has next_action != 'process', then the bug has occurred - the work has already been done.  We can just abort.

jimduff-usds:
We track work-waiting-to-be-done in two places:    It gets queued in an Azure queue, and it gets a row in the TASK table in the database.

There appears to be an issue wherein in rare cases queued items are picked up and processed twice.    Examples of this are documented in #7404.  

The first successful run will update the TASK table, and change the "next_action" to NONE.    So, this double-procesing can be prevented by having the ProcessFunction check that the TASK table next_action value is PROCESS -- that is, the expected next_action in the TASK table matches the expected action found in the queue.   It was probably just an oversight that we were not checking this already - its just a basic assertion / sanity check.

This PR adds a check in ProcessFunction that the next_action in the TASK table is indeed 'process'.

Test Steps:
1. A test has been added to the unit tests.    Further manual testing is not needed.

### Testing
- [X] Tested locally?
- [X] Added tests?

### Process

## Linked Issues
- Fixes #7404 

## To Be Done
- We should check that the other functions do this sanity check!   See #7409

### DIFF
--- a/prime-router/src/main/kotlin/azure/WorkflowEngine.kt
+++ b/prime-router/src/main/kotlin/azure/WorkflowEngine.kt
@@ -549,9 +549,7 @@ class WorkflowEngine(
         db.transact { txn ->
             val task = db.fetchAndLockTask(messageEvent.reportId, txn)
 
-            val blobContent = BlobAccess.downloadBlob(task.bodyUrl)
             val currentAction = Event.EventAction.parseQueueMessage(task.nextAction.literal)
-
             if (currentAction != Event.EventAction.PROCESS) {
                 // As of this writing we are not sure why this bug occurs.  However, this at least prevents it from
                 // causing trouble.
@@ -565,6 +563,8 @@ class WorkflowEngine(
                 // Also, no reason to update the TASK table:  its already marked as done.
                 return@transact
             }
+
+            val blobContent = BlobAccess.downloadBlob(task.bodyUrl)
 
             val report = csvSerializer.readInternal(
                 task.schemaName,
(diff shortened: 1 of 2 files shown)


prime-router/src/main/kotlin/azure/WorkflowEngine.kt
```diff
@@ -549,8 +549,22 @@ class WorkflowEngine(
...
+                // causing trouble.
+                logger.error(
+                    "ProcessQueueFailure: The 'process' queue wants to " +
+                        "process report ${messageEvent.reportId}, " +
+                        "but the TASK table shows its next_action is $currentAction. " +
+                        "Its likely this report was processed earlier, but not removed from the 'process' queue."
+                )
+                // don't throw an exception, as that'll just make this erroneous process action run again.
+                // Also, no reason to update the TASK table:  its already marked as done.
+                return@transact
```
carlosfelix2, line 564: I think it may make more sense to throw an exception here instead of a return to interrupt the execution, but I do not know what happens if you do.

jimmyfagan (approved):
Thanks for adding me to review, good practice for seeing how tests are laid out. Looks good to me, marking approved!

PR #7439, opened 2022-11-21
Allow api/token senders to send parameters in the body

### LINKED ISSUE #7364: Token API arguments
## Problem statement

The token API parameters appear to be passed in the URL instead of as parameters to POST.

## What you need to know

- The token API [spec](https://github.com/CDCgov/prime-reportstream/blob/master/prime-router/docs/api/token.yml) specifies that `parameters:` are `in: path`
- The token API [spec](https://github.com/CDCgov/prime-reportstream/blob/master/prime-router/docs/api/token.yml) is syntactically incorrect yielding errors:
```
Semantic error at paths./api/token.post.parameters.0.name
Path parameter "client_assertion" must have the corresponding {client_assertion} segment in the "/api/token" path
Jump to line 17
Semantic error at paths./api/token.post.parameters.1.name
Path parameter "scope" must have the corresponding {scope} segment in the "/api/token" path
Jump to line 24
```

- The suggested curl command in the [example code](https://github.com/CDCgov/prime-reportstream/blob/master/prime-router/examples/generate-jwt-python/generate-jwt.py) forces a POST but does not use `-d, --data <data>` to pass parameters via POST
- ...

## Acceptance criteria

- [ ] ...

## To do

- [ ] ...

jimduff-usds:
This PR allows senders to send parameters in the body.

Test Steps:
1. Follow the steps here thru the end of STEP 2:  https://github.com/CDCgov/prime-reportstream/blob/master/prime-router/docs/playbooks/how-to-use-token-auth.md

In STEP 2, you run this:  `./prime sender reqtoken --private-key my-rsa-keypair.pem --scope "waters.*.report" --name waters.default`

That `./prime sender reqtoken` command now sends the params in the payload.   If it runs without error and prints out something like this
`{"sub":"waters.*.report_f0ec1e01-000a-449c-a956-b0030c396927","access_token":"really looooong string","token_type":"bearer","expires_in":300,"expires_at_seconds":1669068930,"scope":"waters.*.report"}`

**then it worked!**

## Changes
- Allowed parameters to the api/token endpoint to be send both in the payload and in the URL.

### Testing
- [X] Tested locally?
- [Y] Added tests?

### Process

## Linked Issues
- Fixes #7364 

## To Be Done
- #7440, ticket to get our customers to change.


prime-router/src/test/kotlin/azure/TokenFunctionTests.kt
```diff
@@ -299,7 +299,32 @@ class TokenFunctionTests {
...
         // Invoke
-        var response = TokenFunction(UnitTestUtils.simpleMetadata).report(httpRequestMessage)
+        var response = TokenFunction(UnitTestUtils.simpleMetadata).token(httpRequestMessage)
+        // Verify
+        assertThat(response.getStatus()).isEqualTo(HttpStatus.OK)
+    }
+
+    @Test
+    fun `Test success using parameters in body`() {
+
```
jimduff-usds, line 309: ... and this tests the new way.

prime-router/src/main/kotlin/azure/TokenFunction.kt
```diff
@@ -16,8 +16,15 @@ import gov.cdc.prime.router.tokens.FindReportStreamSecretInVault
 import gov.cdc.prime.router.tokens.FindSenderKeyInSettings
 import gov.cdc.prime.router.tokens.Scope
 import gov.cdc.prime.router.tokens.Server2ServerAuthentication
+import org.apache.http.client.utils.URLEncodedUtils
```
jimduff-usds, line 19: Thanks to @TomNUSDS for finding this most excellent utils library.

prime-router/src/main/kotlin/tokens/SenderUtils.kt
```diff
@@ -39,7 +37,12 @@ class SenderUtils {
             return jws
         }
```
jimduff-usds, line 39: Note:   SenderUtils.kt is a suite of little functions that are useful when you want to write kotlin code that pretends to be someone being a Sender to ReportStream.   Useful for testing, and for creating little example programs.

I had to make some minor tweaks to the methods here so that I could send parameters in URL or in the body.

prime-router/src/main/kotlin/cli/tests/AuthTests.kt
```diff
@@ -528,10 +528,9 @@ class Server2ServerAuthTests : CoolTest() {
         val baseUrl = environment.url.toString() + HttpUtilities.tokenApi
         val privateKey = SenderUtils.readPrivateKeyPem(privateKeyStr)
         val senderSignedJWT = SenderUtils.generateSenderToken(sender, baseUrl, privateKey, kid)
-        val senderTokenUrl =
-            SenderUtils.generateSenderUrl(settingsEnv, senderSignedJWT, scope)
-
-        return HttpUtilities.postHttp(senderTokenUrl.toString(), "".toByteArray())
+        val senderTokenUrl = settingsEnv.formUrl("api/token").toString()
+        val body = SenderUtils.generateSenderUrlParameterString(senderSignedJWT, scope)
+        return HttpUtilities.postHttp(senderTokenUrl, body.toByteArray())
```
jimduff-usds, line 533: Note line 533 now sends the parameters in the body.

prime-router/src/main/kotlin/azure/TokenFunction.kt
```diff
@@ -41,9 +48,21 @@ class TokenFunction(val metadata: Metadata = Metadata.getInstance()) : Logging {
         // &grant_type=client_credentials
         // &client_assertion_type=urn:ietf:params:oauth:client-assertion-type:jwt-bearer
         // &client_assertion=verylong.signed.jwtstring
-        val clientAssertion = request.queryParameters["client_assertion"]
+        //
```
TomNUSDS, line 51: Maybe mention in the comments that `Content-Type:` should be `'application/x-www-form-urlencoded; charset=utf-8'` or `application/x-www-form-urlencoded` and that this format of sending parameters in the body is how form data is POSTed to a server?

prime-router/docs/api/token.yml
```diff
@@ -9,29 +9,29 @@ info:
 paths:
   /api/token:
     post:
-      description: generate a new token for the requested scope
+      description: Generate and return a new token for the requested scope
       security:
         - OAuth2: [ user ]
-      parameters:
```
jimduff-usds, line 15: The openapi spec has been updated to reflect that parameters should be sent in the body.

prime-router/src/main/kotlin/azure/TokenFunction.kt
jimduff-usds: Done.

jimduff-usds (commented):
The release notes and the Programmer's guide and the openapi spec have all been updated to reflect that api/token params should now be sent in the body.

prime-router/docs/api/token.yml
```diff
@@ -9,29 +9,29 @@ info:
...
+                  description: The value should always be "client_credentials"
+                client_assertion_type:
+                  type: string
+                  description:   The value should always be "urn:ietf:params:oauth:client-assertion-type:jwt-bearer"
+                client_assertion:
+                  type: string
+                  description: signed JWT from the sender. See https://github.com/CDCgov/prime-reportstream/blob/master/prime-router/docs/playbooks/how-to-use-token-auth.md
       responses:
         '200':
           description: the token was created and returned
```
TomNUSDS, line 37: From the documentation, I'm still not clear what's being returned for a 200. Is it json with some fields and one of which is a standard token? Is it just the token.

FYI. Some REST client libraries will throw errors if the response type is `application/json` but the content is not valid json (and an empty response is not valid json). 

The one our FE has recently switched to using is such a REST library.

prime-router/src/main/kotlin/cli/tests/AuthTests.kt
TomNUSDS: See my comment in the token.yml about making sure the response content type (header) matches the format of the data returned. 

Some client libraries will throw exceptions if the content type is `application/json` but the body isn't valid json.

IMO: REST API should always return json and put content into fields even if it's just a single value. The helps "future proof" the client code because it generically parses json then just pulls out the fields it understands.

TomNUSDS (commented):
Looks great.

I did have one lingering question about the response format for the API call.

snesm (changes requested):
Programmer's Guide probably needs new version on cover page: "VERSION 2.3 –  AUGUST 2022"
and file name.

jimduff-usds:
> Programmer's Guide probably needs new version on cover page: "VERSION 2.3 – AUGUST 2022" and file name.

Leaving for now - not sure why we have this versionedin the

jimduff-usds:
> 

@snesm , good catch.  I've updated the python example and fixed the Programmer's guide version.

PR #7455, opened 2022-11-23
Check is very slow on staging, added "please wait" message

TomNUSDS:
Before this change, the `Start check` button would disable, but the status area was blank.

With this change, the status area shows `Starting... (this can take a while)`

image


kevinhaube (approved):
Why does it take so long?! 😢

PR #7521, opened 2022-11-30
Fetch Refactor: Replace NewApi: Cleanup

### LINKED ISSUE #6527: Fetch Refactor: Replace NewApi: Cleanup
## Problem statement

As a report stream developer I want to replace all instances of usage of NewApi and useRequestConfig with react query based hooks.

This ticket accomplishes replacement for any remaining files and tests not related to any particular usage of this pattern, such as NewApi and useRequestConfig themselves

## What you need to know

Assume all prior tickets removing usage of these tools are complete

## Acceptance criteria

- [ ] usage of NewApi and useRequestConfig on all mentioned files is replaced with new react query based functionality

## To do

- [ ] ...

jpandersen87:
This PR removes NewApi and directly related files as they have been deprecated and replaced in usage. Includes moving an embedded hook into its own hook file (with test).

Test Steps:
1. Log in
2. Use any organization with senders and receivers (tx-doh, simple-report, etc.)
3. Verify that pages continue to work as expected with data fetches (daily-data, submissions, etc.)

## Changes
- NewApi and directly related files removed
- useAdminSafeOrgName moved and renamed to useAdminSafeOrganizationName (with test)

### Testing
- [x] Tested locally?
- [x] Added tests?

### Process

## Linked Issues
- Fixes #6527


frontend-react/src/hooks/UseAdminSafeOrganizationName.test.ts
```diff
@@ -0,0 +1,72 @@
...
+            dispatch: () => {},
+            initialized: true,
+        });
+        const { activeMembership } = useSessionContext();
+        const { result } = renderHook(() =>
+            useAdminSafeOrganizationName(activeMembership?.parsedName)
+        );
+        expect(result.current).toEqual(undefined);
+    });
+    test("returns correct client organization", async () => {
```
stephenkao, line 34: Minor point, but I don't think this test and [the next one](https://github.com/CDCgov/prime-reportstream/pull/7521/files#diff-9c94215ec20e8aa9cff0f56f854318feb86d24eabd23c6deab16ca9ec693da1dR53) need to be `async` since they're not `await`ing anything.

frontend-react/src/hooks/UseAdminSafeOrganizationName.ts
```diff
@@ -0,0 +1,19 @@
...
+export enum Organizations {
+    PRIMEADMINS = "PrimeAdmins",
+    IGNORE = "ignore",
+}
+
+/** Helper hook that converts `PrimeAdmins` to `ignore`
+ * @todo Ticket to make PrimeAdmins an RS org {@link https://github.com/CDCgov/prime-reportstream/issues/4140 #4140}
+ * @param orgName {string|undefined} Active membership `parsedName` */
+export const useAdminSafeOrganizationName = (orgName: string | undefined) => {
+    return useMemo(
```
stephenkao, line 12: This is a very, very pedantic point, but I'm wondering if we need to memoize this since the logic below is basically a simple ternary; I don't know if we'd really benefit much from memoization since a new `orgName` getting passed in would result in a new recomputation (and likely re-render)🤔

frontend-react/src/hooks/UseAdminSafeOrganizationName.ts
```diff
@@ -0,0 +1,19 @@
+import { useMemo } from "react";
+
+export enum Organizations {
+    PRIMEADMINS = "PrimeAdmins",
+    IGNORE = "ignore",
+}
+
+/** Helper hook that converts `PrimeAdmins` to `ignore`
```
stephenkao, line 8: Thanks for this comment! 🙏

frontend-react/src/hooks/UseAdminSafeOrganizationName.ts
jpandersen87: I also thought the same thing when copy+pasting this from the original file. I didn't want to change anything from the original function from this move initially.

frontend-react/src/hooks/UseAdminSafeOrganizationName.ts
jpandersen87: Credit to the original author 😆

kevinhaube: :D

frontend-react/src/hooks/UseAdminSafeOrganizationName.ts
kevinhaube: Great callout! We were something of a new React team, so we've got a lot of debt in here like "a memo that may not need momization" from bad habits or wonky internet solutions. Definitely want to keep eyes open for chances to fix this, but if the fix becomes cumbersome, reel free to relegate this to a bug/tech-debt item and we can prioritize it in our work that way!

frontend-react/src/hooks/UseAdminSafeOrganizationName.test.ts
```diff
@@ -0,0 +1,72 @@
...
+import { useSessionContext } from "../contexts/SessionContext";
+
+import { MemberType } from "./UseOktaMemberships";
+import {
+    Organizations,
+    useAdminSafeOrganizationName,
+} from "./UseAdminSafeOrganizationName";
+
+describe("useAdminSafeOrganizationName", () => {
+    beforeAll(() => {
```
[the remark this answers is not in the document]
jpandersen87: Agreed. Simplified the tests. Since a parameter is required, I removed the undefined test.

frontend-react/src/hooks/UseAdminSafeOrganizationName.test.ts
```diff
@@ -0,0 +1,22 @@
+import { renderHook } from "@testing-library/react-hooks";
+
+import {
+    Organizations,
+    useAdminSafeOrganizationName,
+} from "./UseAdminSafeOrganizationName";
+
+describe("useAdminSafeOrganizationName", () => {
```
stephenkao, line 8: Thanks for simplifying!  💆

frontend-react/src/hooks/UseAdminSafeOrganizationName.ts
stephenkao: Gotcha, sounds good to me!  I do think we can remove the memoization here with no discernible impact, but it's not a huge deal for this case.  (If we remove the memoization, we could also consider making this just a simple function rather than a hook too, but that's out of the scope of this pull request!)

More generally, though, the biggest risks in my mind from over-memoization are 1) data staleness and 2) very, very, very slight performance hit -- the former of which is much more important to deal with from a UX perspective.  BUT again, it isn't a huge deal for this specific case, and we can just keep an eye on this going forward too!

stephenkao (commented):
The code looks good to me (and I love seeing all that red from deleted files 🔥 🔥 🔥), but I'll defer to someone who has more context to make the final call here!

stephenkao (approved):
Tested locally, and everything seems to work correctly as far as I can tell!  Still happy to defer to others who may have more context over potential impact as I'm still getting familiarized with the codebase, though :)

kevinhaube (approved):
Code looks good, functions on my machine, **BUT** something in the actions seems fishy; failed Azure CLI login. @jeremy-page

PR #7539, opened 2022-12-02
Move the project over to Java 17 and Azure 4

MauriceReeves-usds:
This PR moves us to Java 17 and upgrades the Azure functions image and libraries to version 4.

This is a draft PR so we can test it and make any other updates that we want along the way. There's going to be a lot more work involved with this, though it passes initial local testing.


jack-h-wang:
Closing in favor of #10534

PR #7543, opened 2022-12-02
7402 - Added support for ingesting FHIR bundle through reports endpoint

### LINKED ISSUE #7402: Support ingesting a single FHIR Bundle through the reports endpoint using the UP
### User Story:
As a sender, I want to be able to submit my data as a FHIR bundle, so I can submit all reportable conditions.

### Description/Use Case
We want to be able to support the receiving of FHIR as a report from a sender. This is only for a single FHIR bundle right now.

* We know that any change from the existing RS FHIR bundle format can break us, but the direction we are taking right now is that we are only taking the same FHIR Bundle we are using right now.

### Dev Notes:
* This could be implemented by simply detecting the type of data being submitted (use the content type and verify the right data was sent for that type) then bypassing the HL7 to FHIR conversion.
* Use the proper FHIR content type (research this online)

### Acceptance Criteria
* A sender can submit a RS FHIR bundle to the Universal Pipeline (UP) via the reports endpoint.
* The submitted FHIR bundle is processed (converted, routed, translated, sent) by the UP.
* The reports endpoint returns a success API response if a FHIR Bundle was accepted by the system.
* The reports endpoint returns an error API response (e.g. id=null) if a submission was not accepted by the system (e.g. sent an HL7 file using the FHIR content type).
* Receive and Convert steps are still used with the underlying logs and lineage information stored the same as the incoming HL7 to the universal pipeline.

victor-chaparro:
This PR allows for FHIR bundles to be passed in through the reports endpoint. This for now is using the same FHIR that ReportStream would normally deal with from after the convert step on HL7, and doesn't mean we can broadly accept any FHIR input.

Test Steps:

1. Send a POST request to send a FHIR report
    - Using Postman
        - [Test 7402.postman_collection.json](https://github.com/CDCgov/prime-reportstream/files/10145306/Test.7402.postman_collection.txt)
    - Via curl command
        - [temp.txt](https://github.com/CDCgov/prime-reportstream/files/10145318/temp.txt)

3. Verify that the request was accepted
4. Look at the database to verify that the FHIR report has gone through the expected actions (look at reports and lineage)

## Changes
- Edited `organizations.yml` to add sender for `ignore-fhir` (plus many automatic formatting updates)
- Added a mimeType for `Sender.Format.FHIR` and split out the behavior in `ELRReceiver` depending on the format of the sender
- Added a mimeType for `Report.Format.FHIR` and a function that can return a `Report.Format` from a file extension. Also a function which can return a `Report.Format` from a `blobURL`
- Split out HL7-specific parts of the convert step into a separate function, and make a similar function for FHIR-specific parts of the convert step
- Updated the FHIR converter to read the format from the blobURL and call the appropriate function for the format-specific parts of the convert step
- Added unit tests

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- Resolves #7402


jimmyfagan:
Once again, there are lots of formatting changes that make this changeset appear bigger than it actually is. We think that even though we expect we'll need to make changes during review here, this is ready enough to put it out there and start getting feedback.

prime-router/src/main/kotlin/fhirengine/engine/FHIRConverter.kt
```diff
@@ -144,10 +139,52 @@ class FHIRConverter(
...
                 )
             }
-        } catch (e: IllegalArgumentException) {
+        }
+    }
+
+    /**
+     * Converts an HL7 message into FHIR bundles
+     *
+     * [message] incoming HL7 message to be converted into FHIR
```
carlosfelix2, line 148: Change this kdoc, so the args are inline with the rest of the paragraph.  This is the way the program wants to do it, and yes, there are older examples in many places.

prime-router/src/main/kotlin/fhirengine/engine/FHIRConverter.kt
```diff
@@ -144,10 +139,52 @@ class FHIRConverter(
...
+        // create the hl7 reader
+        val hl7Reader = HL7Reader(actionLogger)
+        // get the hl7 from the blob store
+        val hl7messages = hl7Reader.getMessages(message.downloadContent())
+
+        val bundles = if (actionLogger.hasErrors()) {
+            val e = java.lang.IllegalArgumentException(actionLogger.errors.joinToString("\n") { it.detail.message })
             logger.error(e)
             actionLogger.error(InvalidReportMessage(e.message ?: ""))
+            listOf()
```
carlosfelix2, line 166: Use emptyList()

prime-router/src/main/kotlin/fhirengine/engine/FHIRConverter.kt
```diff
@@ -144,10 +139,52 @@ class FHIRConverter(
...
+        message: RawSubmission,
+        actionLogger: ActionLogger
+    ): List<Bundle> {
+        // create the hl7 reader
+        val hl7Reader = HL7Reader(actionLogger)
+        // get the hl7 from the blob store
+        val hl7messages = hl7Reader.getMessages(message.downloadContent())
+
+        val bundles = if (actionLogger.hasErrors()) {
+            val e = java.lang.IllegalArgumentException(actionLogger.errors.joinToString("\n") { it.detail.message })
```
carlosfelix2, line 163: maybe change the order here.  Create a val with the message then pass that to the logger and actionLogger.  There is no real exception here besides the message that was created.  Then last throw the exception with the message.  We talked about removing this exception, but that can be another ticket (let's create the ticket).

prime-router/src/main/kotlin/fhirengine/engine/FHIRConverter.kt
```diff
@@ -32,39 +33,33 @@ class FHIRConverter(
...
-
-            if (actionLogger.hasErrors()) {
-                throw java.lang.IllegalArgumentException(actionLogger.errors.joinToString("\n") { it.detail.message })
-            }
-
-            // use fhir transcoder to turn hl7 into FHIR
-            val fhirBundles = hl7messages.map {
-                HL7toFhirTranslator.getInstance().translate(it)
-            }
+        val format = Report.getFormatFromBlobURL(message.blobURL)
```
carlosfelix2, line 52: I do not see a reason not to do this, but when we add the new upcoming functionality then we will need to access the DB and get the whole report info, so this code may be redundant then.

prime-router/src/main/kotlin/fhirengine/engine/FHIRConverter.kt
```diff
@@ -144,10 +139,52 @@ class FHIRConverter(
...
+            hl7messages.map {
+                HL7toFhirTranslator.getInstance().translate(it)
+            }
         }
+
+        return bundles
+    }
+
+    /**
+     * Decodes a FHIR message and returns it as list of bundles
```
carlosfelix2, line 178: In line these args in the kdoc

prime-router/src/main/kotlin/Report.kt
```diff
@@ -173,6 +173,15 @@ class Report : Logging {
                     CSV
                 }
             }
+
+            fun valueOfFromExt(ext: String): Format {
+                return when (ext) {
```
carlosfelix2, line 178: I think it may be safer to add case insensitive check here

prime-router/src/test/kotlin/azure/BlobAccessTests.kt
```diff
@@ -25,5 +25,51 @@ class BlobAccessTests {
...
         }
+
+        filename = BlobAccess.BlobInfo.getBlobFilename("")
+        assertThat(filename).isEqualTo("")
+    }
+
+    @Test
+    fun `blob info extension`() {
+        var extension = BlobAccess.BlobInfo
+            .getBlobFileExtension("http://127.0.0.1:10000/devstoreaccount1/reports/ready/pima-az-phd.elr/filename.csv")
```
carlosfelix2, line 36: You are testing outside of the unit you have to test.  You do not need to test each specific extension here, just with/without extension.

carlosfelix2 (changes requested):

prime-router/src/main/kotlin/Report.kt
```diff
@@ -1576,5 +1611,17 @@ class Report : Logging {
...
+        /**
+         * Gets a file format of a blob based on a blobURL
+         *
+         * [blobURL] URL of the blob in question
+         *
+         * @return a Report.Format representing the appropriate format
+         */
+        fun getFormatFromBlobURL(blobURL: String): Format {
+            val extension = BlobAccess.BlobInfo.getBlobFileExtension(blobURL)
+            return Format.valueOfFromExt(extension)
```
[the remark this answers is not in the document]
jimmyfagan: As it is now, `getBlobFileExtension` will return `""`, and then `valueOfFromExt` will throw an exception trying to get the format from that empty string file extension. But we can adjust that if there's a different way you'd recommend.

prime-router/src/main/kotlin/Sender.kt
```diff
@@ -124,6 +124,7 @@ abstract class Sender(
     enum class Format(val mimeType: String) {
         CSV("text/csv"),
         HL7("application/hl7-v2"),
+        FHIR("application/fhir+json")
```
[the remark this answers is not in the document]
jimmyfagan: Ok, we'll mention this during standup tomorrow to determine how to proceed.

prime-router/src/main/kotlin/azure/BlobAccess.kt
```diff
@@ -38,6 +38,17 @@ class BlobAccess : Logging {
...
             }
+
+            /**
+             * Get a file extension from a [blobUrl]
+             * @return the blob filename
+             * @throws MalformedURLException if the blob URL is malformed
+             */
+            fun getBlobFileExtension(blobUrl: String): String {
+                return if (blobUrl.isNotBlank()) {
+                    FilenameUtils.getExtension(URL(URLDecoder.decode(blobUrl, Charset.defaultCharset())).path)
```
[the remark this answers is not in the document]
jimmyfagan: That's true, let's plan on talking about this as a group tomorrow, if it's straightforward to get the report from the report id here and get the format from the report instead, then we don't need to get the extension from the blob url. If it's not straightforward, then maybe we can keep this method for now, and later on when we'll eventually need the report here anyway, we can get ride of this.

prime-router/src/main/kotlin/fhirengine/engine/FHIRConverter.kt
jimmyfagan: This is related to the conversation [here](https://github.com/CDCgov/prime-reportstream/pull/7543#discussion_r1040111421), so this will probably be resolved with whatever comes out of that.

prime-router/src/test/kotlin/fhirengine/engine/FhirConverterTests.kt
```diff
@@ -124,10 +174,54 @@ class FhirConverterTests {
...
+        val actionLogger = spyk(ActionLogger())
+        val engine = spyk(makeFhirEngine(metadata, settings, TaskAction.process) as FHIRConverter)
+        val message = spyk(RawSubmission(UUID.randomUUID(), "http://blobstore.example/file.hl7", "test", "test-sender"))
+
+        every { message.downloadContent() }
+            .returns(File("src/test/resources/fhirengine/engine/valid_data.fhir").readText())
+
+        val result = engine.getBundlesFromHL7(message, actionLogger)
+        assertThat(result).isEmpty()
+        // assert
```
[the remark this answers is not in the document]
jimmyfagan: Nope, unless it's a program expectation to have these, I'm going to just remove `// assert` altogether here.

prime-router/src/main/kotlin/fhirengine/engine/FHIRConverter.kt
```diff
@@ -144,10 +139,52 @@ class FHIRConverter(
...
+     *
+     * [message] incoming FHIR message to be decoded into FHIR bundle
+     *
+     * @return a list containing a FHIR bundle
+     */
+    internal fun getBundlesFromFHIR(
```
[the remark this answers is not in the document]
jimmyfagan: How bout `getContentFromFHIR` and `getContentFromHL7`?

jimmyfagan:
@carlosfelix2 In our recent commits, we added some basic validation in the receive step for FHIR (in response to discussion this morning), and we've addressed all of your comments (some of which will require some follow-up discussion tomorrow to determine how to proceed). So we think this is ready for another review at this point!

prime-router/src/main/kotlin/Report.kt
```diff
@@ -173,6 +173,15 @@ class Report : Logging {
                     CSV
                 }
             }
+
+            fun valueOfFromExt(ext: String): Format {
```
carlosfelix2, line 177: Add kdoc

prime-router/src/main/kotlin/Report.kt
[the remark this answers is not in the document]
carlosfelix2: You need to make sure this is caught correctly even as unlikely that it will be, but it will make sure the code does not blow up

prime-router/src/main/kotlin/azure/BlobAccess.kt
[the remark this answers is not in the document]
carlosfelix2: It is all good for now.  leave as is

carlosfelix2 (approved):

PR #7555, opened 2022-12-05
7301-Permissions/FeatureFlags at organization level.

penny-lischer:
This feature allows the database to maintain a record of an organization's available permissions to features.

Test Steps:
1. Within prime-router directory...
2. run `./gradlew flywayMigrate` 
3. run `./gradlew reloadTables`
4. run `docker-compose down --remove-orphans`
5. run `docker-compose -f docker-compose.yml -f docker-compose.build.yml up -d postgresql`
6. run `./gradlew clean package`
7. run `PRIME_ENVIRONMENT=local && docker-compose -f docker-compose.yml -f docker-compose.build.yml up -d`
8. You can then test the API calls via rest client
GET permissions for an organization - `/api/organization/{orgName}/permissions`
GET all permissions - `/api/permissions`
PUT - `/api/permission/{id}`
POST a permission -` /api/permission`
POST a permission associated with an organization -` /api/organization/{orgName}/permission/{id}`
DELETE a Permission - `/api/permission/{id}`
DELETE a permission associated with an organization -` /api/organization/{orgName}/permission/{id}`

## Changes
- Creation of crud operations.
- Creation of new permission tables.

Feature_Permission_Flow

### Testing
- [x] Tested locally?
- [x] Added tests?


prime-router/src/main/kotlin/azure/DatabaseAccess.kt
```diff
@@ -1352,6 +1355,106 @@ class DatabaseAccess(private val create: DSLContext) : Logging {
...
+        val ctx = if (txn != null) DSL.using(txn) else create
+        return ctx.select(PERMISSION.asterisk())
+            .from(PERMISSION)
+            .join(PERMISSION_ORGANIZATION)
+            .on(PERMISSION.PERMISSION_ID.eq(PERMISSION_ORGANIZATION.PERMISSION_ID))
+            .where(
+                PERMISSION_ORGANIZATION.ORGANIZATION_NAME.eq(organizationName)
+                    .and(PERMISSION.ENABLED.eq(true))
+            )
+            .limit(100)
```
stephenkao, line 1376: Would it make sense to make the limit an optional parameter instead?  I realize 100 might be a biggggg upper bound we don't expect to reach through API calls, but it might make more sense to pass back all Permission records for the given Organization (especially since that's stated in the comment above :P) and manually limit ourselves with a new parameter instead.  What do you think about this?

prime-router/src/main/kotlin/azure/DatabaseAccess.kt
```diff
@@ -1352,6 +1355,106 @@ class DatabaseAccess(private val create: DSLContext) : Logging {
             .fetchInto(Action::class.java)
     }
 
+    /**
+     * Returns all permissions associated with an organization by name
+     * @param organizationName an exact organization name
+     * @param txn an optional database transaction
```
stephenkao, line 1361: This question is just for my own edification, but do we need to start a transaction for single statements like one `SELECT`?

stephenkao: Oh, duh -- it's not executed immediately at this point.  I should have just read a tiny bit further before typing that comment!  😅

prime-router/src/main/kotlin/azure/DatabaseAccess.kt
penny-lischer: Sure, sounds good.

penny-lischer: To be honest we probably don't even need a limit. We only have 4 flags currently and they aren't suppose to stick around long term.

prime-router/src/main/resources/db/migration/V58__create_permissions_tables.sql
```diff
@@ -0,0 +1,38 @@
...
+permission_organization links the permission the organization has access to.
+It has a many(permissions) to one(organization) relationship.
+*/
+CREATE TABLE permission_organization (
+    -- Key
+     permission_organization_id SERIAL PRIMARY KEY,
+     permission_id SERIAL REFERENCES permission(permission_id) ON DELETE CASCADE,
+
+    -- Value
+     organization_name VARCHAR(63)
```
stephenkao, line 37: VERY minor and potentially moot point depending on how we're populating this table, but is it possible for an Organization to update its name after creation?  I wonder if `organization_id` might be a more robust identifier just since it's numeric and less subject to typos 🤔

penny-lischer: We can't use the id b/c of the way in which the org's get created. This is why we are using orgName.

ahay-agile6 (changes requested):
a few questions

prime-router/src/main/kotlin/azure/DatabaseAccess.kt
```diff
@@ -1352,6 +1355,106 @@ class DatabaseAccess(private val create: DSLContext) : Logging {
...
+            .limit(100)
+            .fetchInto(Permission::class.java)
+    }
+
+    /**
+     * Returns a permission
+     * @param id an exact permission id
+     * @param txn an optional database transaction
+     */
+    fun fetchPermissionById(
```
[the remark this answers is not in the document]
kevinhaube: Ooo, great question actually. Just taking stock of what I see as I'm browsing:

- [X] Create flag
- [X] Remove flag
- [X] Update flag
- [X] Delete flag
- [X] Read org flags
- [ ] Add flag to org
- [ ] Remove flag from org

prime-router/src/main/kotlin/azure/DatabaseAccess.kt
```diff
@@ -1352,6 +1355,106 @@ class DatabaseAccess(private val create: DSLContext) : Logging {
...
+     * Deletes a permission
+     * @param id an exact permission id
+     * @param txn an optional database transaction
+     */
+    fun deletePermissionById(
+        id: Int,
+        txn: DataAccessTransaction? = null
+    ) {
+        val ctx = if (txn != null) DSL.using(txn) else create
+        ctx.deleteFrom(PERMISSION)
```
kevinhaube, line 1453: @ahay-agile6 Do you know if we have a soft-delete vs hard-delete rule/norm? Always something I consider when seeing delete functions

kevinhaube (commented):
Left some comments/questions, this is looking good Penny!!

prime-router/src/main/resources/db/migration/V58__create_permissions_tables.sql
```diff
@@ -0,0 +1,38 @@
...
+Follow this style guide https://about.gitlab.com/handbook/business-ops/data-team/platform/sql-style-guide/
+use VARCHAR(63) for names in organization and schema
+
+Copy a version of this comment into the next migration
+*/
+
+/*
+    Permissions are type of features that can be associated with an organization.
+*/
+CREATE TABLE permission (
```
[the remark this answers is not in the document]
penny-lischer: The reason I didn't was b/c I noticed that was the pattern in the dbs - singular. @ahay-agile6 should I go agains the grain :)?

prime-router/src/main/kotlin/azure/DatabaseAccess.kt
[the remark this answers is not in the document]
penny-lischer: We need to add the PermissionOrganization CRUD operations. This was just to check that a permission exists before updating or deleting an existing permission.

prime-router/src/main/kotlin/azure/DatabaseAccess.kt
penny-lischer: @kevinhaube I think were good since we have an enable flag for each permission. To address your comment about the norm I noticed that we mostly have hard deletes.

kevinhaube: Awesome! Thanks for checking on this

ahay-agile6 (approved):

PR #7584, opened 2022-12-07
Experience/7256: example of replica db connection and usage

ahay-agile6:
example of how to use a replica db connection at the class level and in the method

### DIFF
--- a/prime-router/src/main/kotlin/azure/DatabaseAccess.kt
+++ b/prime-router/src/main/kotlin/azure/DatabaseAccess.kt
@@ -59,6 +59,9 @@ import java.util.UUID
 import javax.sql.DataSource
 
 const val databaseVariable = "POSTGRES_URL"
+const val replicaDatabaseVariable = "POSTGRES_REPLICA_URL"
+// TODO: Add POSTGRES_REPLICA_USER env var
+// TODO: Add POSTGRES_REPLICA_PASSWORD env var
 const val userVariable = "POSTGRES_USER"
 const val passwordVariable = "POSTGRES_PASSWORD"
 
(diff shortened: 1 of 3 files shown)


prime-router/src/main/kotlin/azure/DatabaseAccess.kt
```diff
@@ -1643,5 +1633,52 @@ class DatabaseAccess(private val create: DSLContext) : Logging {
                 )
                 .execute()
         }
+
+        private fun createHikariConfig(useReplica: Boolean = false): HikariConfig {
+            val password = System.getenv(passwordVariable)
+            val user = System.getenv(userVariable)
+            val primaryDatabaseUrl = System.getenv(databaseVariable)
+            val replicaDatabaseUrl = System.getenv(replicaDatabaseVariable)
```
ahay-agile6, line 1641: TODO: use replica db user credentials!

prime-router/src/main/kotlin/azure/DatabaseAccess.kt
```diff
@@ -1643,5 +1633,52 @@ class DatabaseAccess(private val create: DSLContext) : Logging {
...
+                        "Attempting to use the replica database connection but the connection string does not exist. " +
+                            "Falling back to the primary database connection"
+                    )
+                primaryDatabaseUrl
+            } else {
+                primaryDatabaseUrl
+            }
+            val config = HikariConfig()
+            config.jdbcUrl = databaseUrl
+            config.username = user
```
ahay-agile6, line 1656: TODO: use replica db user credentials!

stephenkao:
We're not going forward with a replica for the purpose of speeding up our readonly queries.  While connecting to the replica isn't out of the question down the line for downtime coverage, we're having discussions about a more scalable solution to query against documents instead.  Closing!

PR #7608, opened 2022-12-08
Experience-7518: Add session-based AI track event hook

stephenkao:
This changeset adds a thin wrapper around App Insights' `useTrackEvent` hook, which just incorporates session data by default.  As we further refine our analytics needs (and what's available to us through the App Insights dashboard), we can continue to add data to these calls so they'll be tracked by default.

Using this hook is the same as `useTrackEvent`:

```jsx
import useTrackEventWithSession from "./UseTrackEventWithSession";

function Button() {
    // pass in the name of the event on hook invocation
    const trackEvent = useTrackEventWithSession("Button Click")

    // pass in whatever data you want to track to the returned track function
    return (
        <button type="button" onClick={ () => trackEvent({ a: 1, b: 2 }) }>Click me!</button>
    );
}
```

This PR ...

**If you are suggesting a fix for a currently exploitable issue, please disclose the issue to the prime-reportstream team directly outside of GitHub instead of filing a PR, so we may immediately patch the affected systems before a disclosure. See [SECURITY.md/Reporting a Vulnerability](https://github.com/CDCgov/prime-reportstream/blob/master/SECURITY.md#reporting-a-vulnerability) for more information.**

Test Steps:
1. *Include steps to test these changes*

## Changes
- *Include a comprehensive list of changes in this PR*
- *(For web UI changes) Include screenshots/video of changes*

### Testing
- [ ] Tested locally?
- [ ] Ran `./prime test` or `./gradlew testSmoke` against local Docker ReportStream container?
- [ ] (For Changes to /frontend-react/...) Ran `npm run lint:write`? 
- [ ] Added tests?

### Process
- [ ] Are there licensing issues with any new dependencies introduced?
- [ ] Includes a summary of what a code reviewer should test/verify?
- [ ] Updated the release notes?
- [ ] Database changes are submitted as a separate PR?
- [ ] DevOps team has been notified if PR requires ops support?

## Linked Issues
- Fixes #issue

## To Be Done
*Create GitHub issues to track the work remaining, if any*
- #issue

## Specific Security-related subjects a reviewer should pay specific attention to
- Does this PR introduce new endpoints?
    - new endpoint A
    - new endpoint B
- Does this PR include changes in authentication and/or authorization of existing endpoints?
- Does this change introduce new dependencies that need vetting?
- Does this change require changes to our infrastructure?
- Does logging contain sensitive data?
- Does this PR include or remove any sensitive information itself?

[...]

PR #7614, opened 2022-12-08
Proposal: Send activeMembership data with all AI track events

stephenkao:
This changeset updates the AI track calls to always send `activeMembership` -- it should be applicable to `PageviewData`, `PageviewPerformanceData`, and `RemoteDependencyData` since those are active by default, but it should also apply to `EventData` and `MetricData` if we choose to include manually tracked events as well.  (I'll double-check this, though!)

A couple of questions:
- Are there any issues with sending this information on _every_ tracked event?
- Do we need to flatten this data so we only send primitive types?  Right now it's just a stringified object, but I don't have access to the AI dashboard so I don't know what our aggregation capabilities are😅

Test Steps:
1. Log in
2. Start navigating around
3. Check the network tab to see `track` requests and ensure that `activeMembership` is being sent along -- it should be under `baseData.data.properties.activeMembership`:

Screenshot 2022-12-08 at 11 11 16 AM

### Testing
- [ ] Tested locally?
- [ ] Added tests?

### Process

## Linked Issues
- Fixes #issue

## To Be Done
- #issue

    - new endpoint A
    - new endpoint B

### DIFF
--- a/frontend-react/src/TelemetryService.ts
+++ b/frontend-react/src/TelemetryService.ts
@@ -1,6 +1,8 @@
 import { ApplicationInsights } from "@microsoft/applicationinsights-web";
 import { ReactPlugin } from "@microsoft/applicationinsights-react-js";
 
+import { getSessionMembershipState } from "./utils/SessionStorageTools";
+
 let reactPlugin: ReactPlugin | null = null;
 let appInsights: ApplicationInsights | null = null;
 
@@ -30,6 +32,14 @@ const createTelemetryService = () => {
         });
         // Initialize for use in ReportStream
         appInsights.loadAppInsights();
+
+        // Add active membership information to all tracking events
+        appInsights.addTelemetryInitializer((envelope) => {
+            const { activeMembership } = getSessionMembershipState() || {};
+            if (activeMembership) {
+                (envelope.data as any).activeMembership = activeMembership;
+            }
+        });
     };
 
     return {


frontend-react/src/TelemetryService.ts
```diff
@@ -1,9 +1,15 @@
...
 import { ReactPlugin } from "@microsoft/applicationinsights-react-js";
 
+import { getSessionMembershipState } from "./utils/SessionStorageTools";
+
+import { appQueryClient } from "./network/QueryClients";
+
 let reactPlugin: ReactPlugin | null = null;
 let appInsights: ApplicationInsights | null = null;
 
+(window as any).appQueryClient = appQueryClient;
```
stephenkao, line 11: Oops, will remove :)

kevinhaube (commented):
>Are there any issues with sending this information on every tracked event?

The first possible downside I could see is sending a bunch of "undefined"s for logged-out users. That said, not sure how it affects queries. We can tinker with it!

>Do we need to flatten this data so we only send primitive types? Right now it's just a stringified object, but I don't have access to the AI dashboard so I don't know what our aggregation capabilities are😅

It seemed like yesterday it was grouping numeric values and string values from these objects accordingly! I think it stringifies it and then parses it again on the Azure side. We'll check!

stephenkao:
> > Are there any issues with sending this information on every tracked event?
> 
> The first possible downside I could see is sending a bunch of "undefined"s for logged-out users. That said, not sure how it affects queries. We can tinker with it!

I think that's fine since we can't really send anything else for logged-out users.  We'll at least have the AI-generated session id for logged-out users, but there's nothing else available to us if the user isn't logged in! 🤔 

> > Do we need to flatten this data so we only send primitive types? Right now it's just a stringified object, but I don't have access to the AI dashboard so I don't know what our aggregation capabilities are😅
> 
> It seemed like yesterday it was grouping numeric values and string values from these objects accordingly! I think it stringifies it and then parses it again on the Azure side. We'll check!

Yep!  My question was about whether having a stringified object here would make it harder for us to aggregate on the AI dashboard side (like if we'd run into issues with whatever the equivalent of `GROUP BY` is on the dashboard).  I think I remember seeing the properties get properly parsed from the stringified JSON, but would you mind double-checking if you could do something like group by `properties.activeMembership.parsedName` on the AI dashboard?  (Might be [the `summarize` operator](https://learn.microsoft.com/en-us/azure/data-explorer/kusto/query/summarizeoperator)?)

stephenkao:
Confirmed that this works with the `useTrackEvent` hook out of the box as well, so I can close https://github.com/CDCgov/prime-reportstream/pull/7608 if we choose to go forward with this changeset!

Screenshot 2022-12-08 at 1 50 23 PM

stephenkao:
@kevinhaube Now that we've verified that nested fields in the stringified JSON can work, can I get an approval on this?  :P  Otherwise, is there anything else we need here?

kevinhaube (approved):
🥳

PR #7635, opened 2022-12-09
Experience/7358

### LINKED ISSUE #7358: [BUG] Sort in Table component not sorting
The sort in the Table component is not working as expected

**Impact**
Cannot sort on tables with pagination

**To Reproduce**
1. Select an organization that will return more that 20 records within Submissions. ie. Ignore
2. Go to 'Submissions'
3. Click on the 'Date/time' column heading 
4. Notice that the data does not get sorted

**Expected behavior**
Data should get sorted by Date/time in ASC order.
* Unit tests need to be created

**Screenshots**
/api/waters/org/ignore/submissions?pageSize=61&cursor=3000-01-01T00:00:00.000Z&endcursor=2000-01-01T00:00:00.000Z&sort=ASC&showfailed=false

**Logs**
N/A

arnejduranovic:
7358 Update calls to Submission and Delivery endpoints to pass correct params

Test Steps:

## Changes

The root cause of #7358 was we were calling the APIs with the wrong parameters. so:

- Fix table sorting for "Daily Data"
  - Update API call to DeliveryFunction
- Fix table sorting for "Submissions"
  - Update API call to SubmissionFunction

### Testing
- [x] Tested locally?
- [x] Added tests? - Decided with team we will not add tests at this time. The correct way to test that we are passing the correct params to the API is to actually call the API, which we don't have the infrastructure to support at this time.

### Process

## Linked Issues
- Fixes #7358

## To Be Done
- #7637 : This is confirmed on staging as well, so defect was not introduced as part of this ticket

### DIFF
--- a/frontend-react/src/hooks/network/History/DeliveryHooks.ts
+++ b/frontend-react/src/hooks/network/History/DeliveryHooks.ts
@@ -67,7 +67,7 @@ const useOrgDeliveries = (service?: string) => {
         (currentCursor: string, numResults: number) => {
             const fetcher = generateFetcher();
             // The `cursor` parameter is always the "high" value of the range
-            const cursor = sortOrder === "DESC" ? currentCursor : rangeFrom;
+            const cursor = sortOrder === "DESC" ? rangeTo : rangeFrom;
 
             return fetcher(getOrgDeliveries, {
                 segments: {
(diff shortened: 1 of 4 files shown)


stephenkao (approved):
Looks great and works locally!  Thanks for the fix!  🙌 🙌 🙌 

I definitely agree that the double API call is out of the scope of this pull request so we can create a follow-up ticket and tackle that separately!

PR #7638, opened 2022-12-09
7556 [Permission] [API] Documentation

penny-lischer:
This PR adds documentation for the Permission related APIs in the application

Test Steps:
1. place docs/api/permissions.yml in an open api inspector (intellij has one built in)
2. we can now see something like this
Screenshot 2022-12-12 at 3 06 46 PM

3. There is now documentation about the classes and DatabaseAccess apis that was added for the Permission (Feature Flag) feature


kevinhaube (commented):
Looks good to me! Tagging @jimduff-usds in for a review from the API/Platform side

prime-router/docs/api/permissions.yml
```diff
@@ -0,0 +1,163 @@
...
+          description: a valid permissions id
+          required: true
+          schema:
+            type: integer
+      responses:
+        '400':
+          description: bad request
+        '401':
+          description: unauthorized
+  /api/permission:
```
arnejduranovic, line 82: I'm a bit confused on some of these endpoints. Using this one as an example, shouldn't we allow the caller to set different permission parameters, such as name and description, for example?

The #/components/schemas/Permission object is only referenced in the responses?

arnejduranovic (changes requested):

prime-router/docs/api/permissions.yml
arnejduranovic: If you look at reports.yml, it allows the client to pass in query parameters in its post, which is what I would expect to see here?

penny-lischer: yup, my bad - will update

penny-lischer: This endpoint can allow for updating all fields but our use case is only for enabled currently and thus why it can be confusing. I can make it so that it accepts the other fields but lets discuss.

penny-lischer:
@jimduff-usds yes a permission is a feature flag. When the team discussed the architecture of this we did reference the current management of claims in ReportStream. Would you like me to set up a meeting to discuss further?

prime-router/src/main/kotlin/azure/DatabaseAccess.kt
```diff
@@ -1421,6 +1421,8 @@ class DatabaseAccess(private val create: DSLContext) : Logging {
     ) {
         val ctx = if (txn != null) DSL.using(txn) else create
         ctx.update(PERMISSION)
+            .set(PERMISSION.NAME, permission.name)
```
arnejduranovic, line 1424: Does this change mean the client HAS to provide the name and description otherwise it will get set to empty? If so, I think instead of permission.name we should check for null or empty and if so write the current value.

arnejduranovic (changes requested):

prime-router/src/main/kotlin/azure/DatabaseAccess.kt
penny-lischer: I was thinking the UI would pass the name, description, and enabled values always. This means some/all could be modified, otherwise it would be passing the original values. Thoughts?

arnejduranovic: Sure, if the front-end took it upon itself to do that then we don't have any problems, in that particular case. But from the perspective of the API it would be better if it checked for empty/null before setting it, since the API can't control what the client sends and the API knows that empty/null values are not valid. For example, what if someone calls the API via postman or a third party application and forgets to fill out all the values?

Now that I think about it more, we would probably have to do a DB read to get the current value, right? So instead of `we should check for null or empty and if so write the current value.` I should have said `we should check for null or empty and if so DON'T set that value in the query syntax. ` This probably means you would have to do the .set calls on separate lines as part of a conditional.

As a side note, if we did query params we could probably mark the params as required and the azure function would reject the request automatically, saving us from some work. But, in my opinion, name and description should be optional anyways.

arnejduranovic: Also, I can see this being part of a different ticket, so feel free to create one and I can approve this PR, since it is just for documentation.

penny-lischer: @arnejduranovic makes sense. Thanks!

arnejduranovic (approved):

kevinhaube (commented):
Will this remove my requested changes? Let's find out

jimduff-usds (commented):
I'd like to discuss all of this through as a group, but here's my thoughts:

- First, FeatureFlags can be their own thing, and I'm thinking maybe we don't try to conflate them with the more general concept of "Permissions".

- Are we OK with FeatureFlags being per-Organization and not per-User?   That is, we can turn on/off features for entire Orgs, and not for individual users in that Org.  

- If yes, then I'd propose simply adding a FeatureFlag List to the Organization setting object yaml, which can contain zero or more FeatureFlags.   No permission API needed, No permission table nor permission_organization table needed.   Instead, the react code would just retrieve the entire Organization Setting, and get the feature flags.  This should be very fast, and maybe you are already doing this?

- To add a FeatureFlag to an Organization, you just edit their setting in the existing Settings Edit feature.    I think you'll need to add a UI widget to support this.

- Minor note if you do want Feature Flags per user, here is one way this might be done:  Remember the per-Org Admin level access I mentioned?  If desired, its possible we could have feature flags that only kick-in if you both 1) have that feature flag in your Org, and 2) you are an Admin within your Org.

kevinhaube:
>First, FeatureFlags can be their own thing, and I'm thinking maybe we don't try to conflate them with the more general concept of "Permissions".

Agree, the flags should only control show/hide on the front-end, not grant authorization. Authorization to and endpoint should remain a scope-based permission as you outlined the system today.

>Are we OK with FeatureFlags being per-Organization and not per-User? That is, we can turn on/off features for entire Orgs, and not for individual users in that Org.

For now, yes, as many orgs only contain 1-4 web app users anyway (according to Okta group populations). No per-user requirements have come up in the current iteration of the web app, but migration towards a user-based FeatureFlag paid service in the future might be on the table if the requirement arrises.

>Instead, the react code would just retrieve the entire Organization Setting, and get the feature flags. This should be very fast, and maybe you are already doing this?

>To add a FeatureFlag to an Organization, you just edit their setting in the existing Settings Edit feature. I think you'll need to add a UI widget to support this.

The issue is the need to update them, add them, and remove them from those settings reliably. Without some list of legitimate feature flags - say if we kept them instead as front end enum - we could potentially run into a situation where a feature flag is no longer in _our_ enum, thus not a valid input, but hasn't yet been removed from a setting. In creating the FeatureFlags table, the idea was it replaces what we _used_ to do, which is track these via a front-end enum and just plug them into our localStorage through an Admin exclusive interface. 

Our desire is to no longer manage these from the codebase, but from the Admin exclusive front-end, and only require installation of the logic in the app to check for the flag. 

Hope all of this helps with the convo! Chat next week :)

snesm:
@jimduff-usds commented
> If a feature flag is not a permission, then perhaps we should change the name of the API from /api/permission to /api/featureflag.

@kevinhaube commented
> Agree, the flags should only control show/hide on the front-end, not grant authorization. Authorization to and endpoint should remain a scope-based permission as you outlined the system today.

Since this is not an authorization/permission, but rather control over features in the user interface, can this be renamed something like UIControl?

penny-lischer:
Closing this due to change in architecture per Platform Refinement meeting on 12/19/22

PR #7639, opened 2022-12-09
7448 [Bug] Json Diff Editor gets out-of-sync when doing edits

penny-lischer:
Updated onChange method of JSON diff editor to reformat the text so that the changed text gets highlighted correctly

Test Steps:
1. Go into Edit Settings for `ignore` (`admin/orgsettings/org/ignore`)
2. Get into the Json Diff editor via the `Preview Save...` button
3. Change the Right Hand editable side by editing an existing field... but do NOT use correct json lint format. For example: change `"filters": [],` to `"filters": ["1","2"],` (which really should be:
```
"filters": [
    "1",
    "2"],
```
to be correctly formatted.
Result: The highlighting for the differences is now correct.
Screenshot 2022-12-09 at 1 53 24 PM

### Testing
- [x] Tested locally?
- [ ] Added tests?


frontend-react/src/components/EditableCompare.tsx
```diff
@@ -147,7 +147,9 @@ export const EditableCompare = forwardRef(
         // on change, we highlight the errors
         const onChangeHandler = useCallback(
             (newText: string) => {
-                setTextAreaContent(newText);
+                setTextAreaContent(
+                    JSON.stringify(JSON.parse(newText), null, 2)
```
kevinhaube, line 151: Because our app errors in production can trigger PagerDuty, we should be `try/catch`-ing anything that might throw like `JSON.stringify` and `JSON.parse` and using `console.warn` to "mute" it for PagerDuty.

We're working on keeping logs of warnings and errors in Azure Logs, too, so we aren't sacrificing any useful debugging/internal error info by this change. :) 

We did this recently in the TableFilters component [here](https://github.com/CDCgov/prime-reportstream/blob/84610483d1e5b5b5b971a86790fa6247664754c2/frontend-react/src/components/Table/TableFilters.tsx#L54-L69)

kevinhaube (changes requested):
Nice bug catch! Just a minor change to keep PagerDuty as painless as possible!

@stephenkao This component currently has no tests. How do you wanna handle incrementally improving unit test coverage? I could ticket it, or we can hack at it ourselves!

frontend-react/src/components/EditableCompare.tsx
stephenkao: Regarding error-handling, I think we actually want to do something different here.  The changeset in this pull request is changing the behavior of the component -- originally, you could type in invalid JSON and still update the textarea, but now `textAreaContent` fails to update with invalid JSON because of `JSON.stringify` and `JSON.parse`.

I'm not sure what the optimal UX would be in this case, but my thought here would be to:
- Allow users to type in any content (even invalid JSON)
- If the content is valid, then update the textarea with properly highlighted content
- If the content is not valid, then show the toast error like before
Screenshot 2022-12-12 at 10 22 14 AM

stephenkao:
> @stephenkao This component currently has no tests. How do you wanna handle incrementally improving unit test coverage? I could ticket it, or we can hack at it ourselves!

My feeling is that we should try to add/update tests whenever we change any functionality, even  (maybe especially?) if the update includes a bugfix so as to mitigate future regressions that might be similar -- if the fix and the test are in the same changeset, it becomes easier to 1) understand what the issue is under the hood, 2) see the different cases the new fix is testing against (if applicable), and 3) revert them both at once if the fix actually ends up breaking something later on.  There are definitely exceptions to this since we can't write tests for every possible situation, but I think we should err on the side of more tests -- within reason.

frontend-react/src/components/EditableCompare.tsx
stephenkao: @penny-lischer Yep, I know :) I'm saying that this changeset changes the behavior because it doesn't allow for any invalid JSON to get added _because of_ the errors that come out of `JSON.parse`.

See master:


Versus this branch:


Note that I'm not able to type in anything because `JSON.parse` is throwing an error since what I'm typing in is invalid JSON.  It'd accept it if I copied and pasted valid JSON as one change, but I don't think this would be common in our user flows 🤔

frontend-react/src/components/EditableCompare.tsx
```diff
@@ -127,6 +127,19 @@ export const EditableCompare = forwardRef(
...
 
+                // reformat the modified text in case it is not in correct JSON format
+                // then stick it back into the edited box. This ensures the highlight works correctly.
+                try {
+                    const reformatModifiedText = JSON.stringify(
+                        JSON.parse(modifiedText),
+                        null,
+                        2
+                    );
+                    setTextAreaContent(reformatModifiedText);
```
stephenkao, line 138: Hmmm it's a bit odd to me that we're calling `setTextAreaContent` in `refreshDiffCallback`, which is already called immediately after `setTextAreaContent` in [`onChangeHandler`](https://github.com/CDCgov/prime-reportstream/blob/master/frontend-react/src/components/EditableCompare.tsx#L150-L151).  It miiiiiiight also be poor UX to replace what the user put in with a pretty-printed JSON (i.e., `JSON.stringify(whatever, null, 2)`) -- even though that'll guarantee the error demarcations are correct, it automatically changes the user input and prevents spaces and newlines from being added when the current JSON is valid:


I think there's a bigger bug here due to the fact that the diffing is based on an indentation of two spaces (see [JsonSourceMap](https://github.com/CDCgov/prime-reportstream/blob/master/frontend-react/src/utils/DiffCompare/JsonSourceMap.ts#L86) and [JsonDiffer](https://github.com/CDCgov/prime-reportstream/blob/master/frontend-react/src/utils/DiffCompare/JsonDiffer.ts#L276-L277)), and user input may include less or more whitespace than that -- which is valid JSON as long as the structural pieces are all valid.

In my opinion, this is a fairly small UX bug that would require a much larger restructure to address so I'd be in favor of punting on it and coming back to it later if we need.  What are your thoughts?  (I'd also be in favor of looking into an external library to handle this so we don't have to maintain the logic ourselves, but that's a different conversation! :))

TomNUSDS: > I'd also be in favor of looking into an external library to handle this so we don't have to maintain the logic ourselves, but that's a different conversation! :)

History: I investigated other solutions and was not able to find any that didn't come with their own set of issues. Also, the first iteration of the code was a simple text differ, you might see it referenced in places. It wasn't removed because it could still be handy for other data formats (FHIR?).

UI wise, is it possible to click directly on the Save button with invalid json in the edit area? 

One possible solutions that uses a fairly trivial UI change:
1. when the edit area gets focus, remove the highlighting (so you don't have to worry about keeping it in sync) and add a "check syntax" button somewhere (maybe overlay it in the upper left of the edit area).
2. When the "check syntax" is clicked, then apply formatting and syntax highlighting and syntax error checking and remove/disable the button.
3. Run "check syntax" function when the Save button is clicked to make sure there are no json syntax mistakes before calling the POST API.

stephenkao: Thanks for the context, @TomNUSDS!  Yeah, I think moving it to a more user-initiated event like clicking a new 'Check Syntax' button would be better -- the issue was that the highlighting and reformatting was happening while the user could still be typing was a jarring experience so taking it out as its own action would solve this issue.  @penny-lischer what are your thoughts about this?

penny-lischer: Sure

Issue #7427, opened 2022-11-21
UP returns specific errors when bad HL7 data is submitted


carlosfelix2:
### User Story:
As a sender, I want specific error messages if my data was not accepted, so I can fix it.

### Description/Use Case
The UP is using the HAPI HL7 v2 library to read in HL7 data.  Currently we are returning a generic error message if the HL7 data is invalid.  We need to include more details if they are available.

### Risks/Impacts/Considerations
* The error messages provided by the HAPI library may not be too specific, but it will be better than nothing.

### Dev Notes:
* Note that errors in decoding are given at the report level and not the individual HL7 message level (like when you get multiple messages in a BATCH).  Most if not all the time we will receive single HL7 messages.
* Try to keep the error message as clean as possible, meaning, you do not want to be sending an exception as the message, just a test message a user could understand.
* See HL7Reader.kt:33
* Note the errors are logged in the action_log table correctly.
* An example HL7 error would be a missing required field (e.g. missing the message type)

### Acceptance Criteria
* The UP provides the error details from the HL7 v2 decoding errors if available from the HAPI v2 HL7 library.

PR #7661, opened 2022-12-12
Victor/7427 hl7 errors

victor-chaparro:
This PR adds the ability for UP return specific errors when bad HL7 data is submitted. Currently our implementation of the HL7Reader has validation turned off and will only throw an error if it fails to parse the HL7 message. Because of this the amount of bad data you can send is very minimal. Once validation is turned on we can check if more specific error messages need to be added.

Test Steps:
1. Submit bad data to the reports API. Sample bad data can be found in the unit tests added to `HL7ReaderTests.kt` on this PR.
2. Test required field is missing. Sample  HL7 message can be found in unit test `'test decoding of bad HL7 message - MSH Version ID missing'`
3. API should return this error message: "Required field missing: Can't find version ID - MSH.12 is null"
4. Test  Data type error. Sample  HL7 message can be found in unit test `test decoding of bad HL7 message - OBX Wrong data type`
5. API should return this error message: "Data type error: 'test' in record 1 is invalid for version 2.5.1 at OBX-2(0)"
6. I tried sending other types of bad data, but nothing would throw an error. I tried sending duplicate lines, taking out required fields, bad data for dates, gender, and country codes.

## Changes
- Added logic to `HL7Reader.kt` to log and return more specific errors when bad HL7 data is submitted.

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- Fixes #issue

## To Be Done
- #7427


TurkeyFried:
Positive first impression, the tests are working. I'm gonna play with it and finish the review tomorrow.

prime-router/src/main/kotlin/fhirengine/utils/HL7Reader.kt
```diff
@@ -40,6 +44,40 @@ class HL7Reader(private val actionLogger: ActionLogger) : Logging {
...
+                    ErrorCode.DATA_TYPE_ERROR.code -> actionLogger.error(
+                        InvalidReportMessage("Data type error: ${rootCause.message}")
+                    )
+                    else -> actionLogger.error(InvalidReportMessage("Failed to parse message"))
+                }
+            }
+
+            else -> {
+                actionLogger.error(InvalidReportMessage("Failed to parse message"))
+            }
```
jimmyfagan, line 77: (Optional) An option to make this function more concise would be to initialize a `var errMessage = "InvalidReportMessage("Failed to parse message")` or `var errMessageStr = "Failed to parse message"` at the top, then update that var if we have a more specific message, and log that `errMessage` either way at the end. It would allow us to get rid of the duplicate else cases at 76 and 71, but would involve a tiny bit of extra operations by having to init and overwrite the var in some cases.

jimmyfagan (approved):
Looks great! I left one optional comment, nothing important, just an option to consider, feel free to ignore it if you prefer it as it is!

prime-router/src/main/kotlin/fhirengine/utils/HL7Reader.kt
victor-chaparro: I tried making that change, but `when` requires an else statement... I did turn the error message into a variable to make the function more readable.

TurkeyFried (approved):

clediggins-usds (changes requested):

prime-router/src/test/kotlin/fhirengine/utils/HL7ReaderTests.kt
```diff
@@ -130,4 +131,33 @@ OBX|1|CWE|94558-4^SARS-CoV-2 (COVID-19) Ag [Presence] in Respiratory specimen by
         assertThat(timestamp).isNotNull()
         assertThat(timestamp).isEqualTo(SimpleDateFormat("yyyyMMddHHmmssZ").parse(timestampStr))
     }
+
```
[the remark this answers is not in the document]
victor-chaparro: I tried many different scenarios to try to get a validation exception, but I only got them when I turned on validation on the Hl7InputStreamMessageIterator. I left the catch in there so it's ready for whenever we add validation. Is there a scenario you know of where I can send an invalid HL7? tried sending duplicate lines, taking out required fields, bad data for dates, gender, and country codes and nothing would throw an error.
There's already a test that passes an entirely invalid HL7, but that throws a different exception and that's covered in that test.

clediggins-usds (approved):

PR #7670, opened 2022-12-13
Revert "Experience/7358"

arnejduranovic:
Reverts CDCgov/prime-reportstream#7635

Currently this sorting fix causes a new issue: The if a user navigates past the 5th page the pages stop updating with values. This is because I messed up the calculation of the "cursor" value. Will revert this PR so we don't introduce this issue to production and will work on a fix.


kevinhaube (approved):
🥾 🐞

PR #7682, opened 2022-12-14
7052 - UP shows routing filter results in submission history

### LINKED ISSUE #7052: Submission History contains routing filter result for Universal Pipeline submissions
### User Story:
As a Sender, I want to see the list of receivers my data was sent to via the Universal Pipeline with the result of the routing filter, so I can track my data.

### Description/Use Case
Senders can access the submissions history API to get the submission result for a given report that contains which receivers the report was sent to.  We need to implement the destinations list in the response of the API for data processed through the Universal Pipeline.

### Dev Notes:
* Refer to the COVID pipeline to see how the destinations are store in the database.
* Once the destinations are stored it is possible you do not have to do anything to the API itself.

### Acceptance Criteria
* A sender can see the results of any routing filter evaluated for a given receiver in the Submissions History API result for a given report.

jimmyfagan:
These changes continue on the work from #7051 to ensure that the routing filter results show up in the submission history. There was a slight bug that was preventing this from working, but once that was resolved, this worked basically out-of-the-box.

Test Steps:
1. Add a routing filter to `organizations.yml`
2. Reload settings and run ReportStream
3. Send a report that will fail to pass the routing filter
4. Using the report id from the result of the report submission, query the submission history

<details>
<summary>You should see a result like this</summary>

```
{
    "id": "884c19ae-1499-4313-9ed5-eae8e604756e",
    "submissionId": 37,
    "overallStatus": "Not Delivering",
    "timestamp": "2022-12-14T02:11:29.136Z",
    "plannedCompletionAt": null,
    "actualCompletionAt": null,
    "sender": "ignore.ignore-full-elr",
    "reportItemCount": 1,
    "errorCount": 0,
    "warningCount": 0,
    "httpStatus": 201,
    "destinations": [
        {
            "organization": "FOR TESTING ONLY",
            "organization_id": "ignore",
            "service": "FULL_ELR",
            "itemCount": 0,
            "filteredReportRows": [
                "For ignore.FULL_ELR, filter [matches(test_result_status, C, P)][] filtered out item 1234d1d1-95fe-462c-8ac6-46728dba581c"
            ],
            "filteredReportItems": [
                {
                    "filterType": "ROUTING_FILTER",
                    "filterName": "[matches(test_result_status, C, P)]",
                    "filteredTrackingElement": "1234d1d1-95fe-462c-8ac6-46728dba581c",
                    "filterArgs": [],
                    "message": "For ignore.FULL_ELR, filter [matches(test_result_status, C, P)][] filtered out item 1234d1d1-95fe-462c-8ac6-46728dba581c"
                }
            ],
            "sentReports": [],
            "downloadedReports": []
        }
    ],
    "actionName": "receive",
    "externalName": null,
    "reportId": "884c19ae-1499-4313-9ed5-eae8e604756e",
    "topic": "full-elr",
    "errors": [],
    "warnings": [],
    "destinationCount": 0
}
```

</details>

## Changes
- Every time I edit the `organizations.yml` file, my IDE fixes the formatting, and it makes it hard to keep track of what I changed, so I'm including **format-only changes** in that file to prevent that from continuing to happen in the future.
- In `FHIRRouter.kt`, we were logging filter results multiple times since a non-passing filter at one step would continue to hit all of the other log points meant for the other filter types. This also meant that we were logging non-passing juris filters, and that we often didn't even have all of the variables set we need for the log messages on other filters (including routing). I added some continues for the `!passes` cases for each filter, which fixes this issue, and enhances readability. Victor also added a helper function which consolidates some duplicated code.
- Victor added several tests, including ones which would have caught this issue

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- Fixes #7052


prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
```diff
@@ -332,42 +332,46 @@ class FHIRRouter(
             // JURIS FILTER
             //  default: allowNone
             var passes = evaluateFilterCondition(getJurisFilters(receiver, orgFilters), bundle, false)
+            if (!passes) {
+                return@forEach
```
clediggins-usds, line 336: Multiple returns can result in spaghetti code, especially if methods get longer and more complicated. Please do not do internal returns unless absolutely necessary for a programmatic reason. Add in additional 'if' statements where needed to not run logic unless needed.

This previously had short circuit logic but that was removed during a previous merge.

clediggins-usds (changes requested):

jimmyfagan:
To follow up on a few discussions on this PR:

- We had a huddle and restructured `applyFilters` to be closer to what it originally was. We decided that while early returns don't need to be prohibited generally, this other form was more concise and readable in this case.
- I double checked to make sure the formatting changes to `organizations.yml` don't affect the ability to load settings. We're all good on that front.

I'm going to add a few quick short unit tests, and then will resubmit this for review.

clediggins-usds (approved):

jimmyfagan:
Finally was able to verify smoke tests pass, also added one more commit with an additional unit test. Will leave this open til somebody has a chance to review the latest including all commits, and then will merge in.

clediggins-usds (approved):

Issue #7558, opened 2022-12-05
Errors in the evaluation of UP filter expressions should be recorded


carlosfelix2:
### User Story:
As an administrator, I want ReportStream to log any issues evaluating receiver filter expressions, so I can trace why of data was not sent to a receiver.

### Description/Use Case
Filter expressions for receivers are automatically evaluated to being false if the result of the FHIR Path expression does not evaluate to a boolean.  If someone makes a mistake in an expression, then the filter simply return false and the data will not be sent to a receiver.  There is a warning that gets written to the logs, but those logs are pretty much hard to get to in production.  

### Dev Notes:
* The reason to auto evaluate a bad expression to False is that there is really no reason for the system to retry re-evaluating the filters again when you know 100% that they will fail again.
* But, it is not easy to see that there was a problem.
* I think if there is a syntax error in the FHIR Path then an exception is thrown, but don't remember if it goes into a retry or not.
* Need consistency in this behavior.
* We log a warning instead of an error as we can continue evaluating filters for other receivers.  Maybe this should be an error, but careful with changing the behavior for other uses (e.g. for a schema it is just a warning as we want to generate the data anyways). 
* The Submission History API response has not included this level of errors, so need to define how this will look in the response.

### Acceptance Criteria
* The UP logs an action log error for a given report if there is a syntax error in a receiver filter FHIR Path expression.
* The UP logs an action log error for a given report if there a receiver filter FHIR Path expression does not evaluate to a Boolean expression.
* The Submission History API response includes filter FHIR Expression errors.

TurkeyFried:
From discussion: These messages are not actually useful for the average API user. Instead, these should be logged but not returned in the API response.

TurkeyFried:
From discussion: These messages are not actually useful for the average API user. Instead, these should be logged but not returned in the API response.

TurkeyFried:
From discussion: These messages are not actually useful for the average API user. Instead, these should be logged but not returned in the API response.

PR #7687, opened 2022-12-14
7558 UP store filter expression errors

TurkeyFried:
This PR adds logging support for errors found while evaluating receiver filter expressions in the Universal Pipeline, without making them viewable on the Submissions API.

Test Steps:
1. On your `organizations.yml` file, change a `qualityFilter:` entry for a receiver to an unparsable expression (e.g. `#2!dej2+1))k))(`) . Make sure to run `reloadSettings` on Gradle after saving the changes.
2. Submit a Full ELR report via Postman (or whichever you use)
3. After receiving a `submissionId`, use that to check the History API until the response includes a list of `destinations`.
4. On receiving a response, scroll to the bottom of the output. Check the `errors` array and verify it's empty.
5. Go to the database and check in the `action_log` table for a new entry that matches the issue with the given expression.
6. Repeat steps 1-5, but this time use an expression that returns a non-Boolean value (e.g. `1+2`). Verify the new error text looks correct.

## Changes
- Add filter logging for any errors during the `evaluateCondition` step on the Universal Pipeline
- Use simplified messages on the API-viewable data

## Linked Issues
- #7558


prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
```diff
@@ -175,6 +182,7 @@ class FHIRRouter(
         actionHistory: ActionHistory
     ) {
         logger.trace("Processing HL7 data for FHIR conversion.")
+        this.actionLogger = actionLogger
```
carlosfelix2, line 185: I feel like this should be set at the constructor if you are setting a class property. Can that be done?

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirPathUtils.kt
```diff
@@ -103,12 +104,14 @@ object FhirPathUtils : Logging {
...
         } catch (e: Exception) {
             // This is due to a bug in at least the extension() function
-            logger.error(
-                "Unknown error while evaluating FHIR expression $expression for condition. " +
-                    "Setting value of condition to false.",
-                e
-            )
-            false
+            val msg =  when (e) {
+                is FHIRLexerException -> "Syntax error on filter expression"
```
carlosfelix2, line 108: At this level, this can be used in many places. Change this to "Syntax error in FHIR Path expression $expression

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirPathUtils.kt
```diff
@@ -103,12 +104,14 @@ object FhirPathUtils : Logging {
...
+                is SchemaException -> e.message ?: "Condition error on filter expression"
```
carlosfelix2, line 109: Same here, change this to FHIR Path expression and add the expression.

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirPathUtils.kt
```diff
@@ -103,12 +104,14 @@ object FhirPathUtils : Logging {
...
+                else -> "Unknown error while evaluating FHIR expression $expression for condition. " +
```
carlosfelix2, line 110: "... FHIR Path expression ..."

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirPathUtils.kt
```diff
@@ -103,12 +104,14 @@ object FhirPathUtils : Logging {
...
+                    "Setting value of condition to false."
+            }
+            logger.error(msg, e)
+            throw SchemaException(msg)
```
carlosfelix2, line 114: This is not setting the final result to false.  Are we still doing that?

carlosfelix2 (dismissed):

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirPathUtils.kt
TurkeyFried: Since the method is only used in 2 places, I thought it made more sense to have it return the exception if there is one, and then handle that exception appropriately on the calling function

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirPathUtils.kt
TurkeyFried: done

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirPathUtils.kt
TurkeyFried: done

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirPathUtils.kt
TurkeyFried: done

TurkeyFried:
I'd forgotten to update the PR information with changes from discussions. It should now be more accurate.

JessicaWNava:
Looks good to me, thanks Luis for helping me through some of the glitches I had trying to get it to work!

TurkeyFried:
Disabling auto-merge so I can pull changes without accidentally merging it yet

jimmyfagan (approved):
Looks good, I want to try to get confirmation that the user shouldn't see anything about this in the submission history at standup today, and then Carlos has one unresolved conversation, but I think aside from that, this looks good from my perspective.

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
TurkeyFried: I don't see an easy way to do this without refactoring on FHIREngine and the other classes that inherit from it

PR #7699, opened 2022-12-15
Cle/7202 up in smoke

clediggins-usds:
This PR adds a smoke test for universal pipeline

Test Steps:
1. ./gradlew testSmoke

## Changes
- add universal pipeline sender for ignore
- updated fileUtilities to take in schema name instead of sender
- added end2endUniversalPipeline test
- updated pollForProcessResult to make it more task action agnostic
- update pollForStepResult to make it more task action agnostic
- updated queryForStep results to make it task action agnostic
- added examineStepResponse to look at the DetailedSubmissionHistory for each step
- modified examplePostResponse to make it accept both covid-19 and full-elr topics
- added universal pipeline receiver
- updated stepActionResultQuery to make it task agnostic

### Testing
- [x] Tested locally?
- [x] Added tests?


prime-router/settings/organizations.yml
```diff
@@ -2504,6 +2504,26 @@
         port: 22
         filePath: ./upload
         credentialName: DEFAULT-SFTP
+    - name: UNIVERSAL_PIPELINE
+      externalName: Ignore FHIR Universal Pipeline
+      organizationName: ignore
+      topic: full-elr
+      customerStatus: active
+      jurisdictionalFilter: [ "%performerState = 'CO' or %patientState = 'CO'" ]
```
carlosfelix2, line 2512: Historically, this has been always set to something like `matches(ordering_facility_county, HL7_BATCH)` for receivers used in the smoke tests. Could we do that here too?

prime-router/src/main/kotlin/cli/tests/BasicTests.kt
```diff
@@ -65,13 +65,104 @@ class Ping : CoolTest() {
     }
 }
 
+class End2EndUniversalPipeline : CoolTest() {
```
carlosfelix2, line 68: Need kdoc

carlosfelix2 (changes requested):
I shutdown the SFTP server and I still got a success on the test.  See normal end2end test

prime-router/src/main/kotlin/cli/tests/TestReportStream.kt
```diff
@@ -911,26 +991,68 @@ abstract class CoolTest {
...
             reportId: ReportId,
             receivingOrgSvc: String? = null,
             receivingOrg: String? = null,
             action: TaskAction,
+            isUniversalPipeline: Boolean
         ): Int? {
             val ctx = DSL.using(txn)
-            val sql = """select count(*)
+            return if (isUniversalPipeline) {
+                val sql = """
```
JessicaWNava, line 1022: Thank you for letting me rubber duck with you on this, it actually makes sense when it definitely would not have otherwise.

JessicaWNava (commented):
Tested and it worked. At least to the level I understand, the changes look good. Thank you for all of the cleanup you did along the way as well!

TurkeyFried (approved):
Tests pass, I see the kdoc has been added, looks good!

carlosfelix2 (approved):

PR #7710, opened 2022-12-17
side scrolling tables

### LINKED ISSUE #4926: Tables need to side-scroll in narrow viewports
## Problem statement
On narrow layouts, like mobile phones, the tables on the RS website do not side-scroll. They currently cause overflow scrolling.

## What you need to know
USWDS scrolling table: https://designsystem.digital.gov/components/table/

## Acceptance criteria
- [ ] Users can side-scroll data on a table when in narrow viewports.

## To do
- [ ] Implement USWDS scrolling on tables

etanb:
This PR fixes tables to have the `scrolling` attribute turned on when a user is on a table or mobile width device.

Test Steps:
1. Open on a mobile device, or device `1024px` or under in viewport width
2. Go to your admin page.
3. Scroll table left or right by simple dragging

## Changes
- New `util` that a developer can pass in what device they want to check for (`mobile` || `tablet` || `desktop`)
- Change the `<Table/>` component to allow scroll on certain device width

https://user-images.githubusercontent.com/516877/208568720-575c3138-56f7-47ac-a323-00f7dcb07f2e.mp4

### Testing
- [ ] Tested locally?
- [ ] Added tests?

### Process

## Linked Issues
- Fixes #4926 

## To Be Done
*NEED CONFIRM FROM REVIEWER BEFORE PROCEEDING* ⚠️ 

I ONLY applied my solution to the `OrgsTable.tsx` page since I wanted to confirm we liked this solution and didn't need any tweaks before proceeding to adding to the other many instances of our usage of the `<Table/>` component.

## Known Issues

I did not add a `onResize` handler so if you open the able on Desktop, then resize the browser to be smaller, the table will _not_ change. I did this on purpose since the vast majority of people, besides devs, will not resize the browser and use the app in whatever format they opened it in. `onResize` handler is expensive and will fire constantly, but if there's a case for it, I can create a debounced version of the handler.


kevinhaube (commented):
Just wanna make sure we tried using the USWDS solution first: `usa-table-container--scrollable` should handle the window logic via css!

[USWDS Tables](https://designsystem.digital.gov/components/table/), a few sections down. Expanding the "Component Code" section gives you all the goodies! That's basically free boilerplate since we utilize their css styles.

stephenkao:
> Just wanna make sure we tried using the USWDS solution first: `usa-table-container--scrollable` should handle the window logic via css!
> 
> [USWDS Tables](https://designsystem.digital.gov/components/table/), a few sections down. Expanding the "Component Code" section gives you all the goodies! That's basically free boilerplate since we utilize their css styles.

@kevinhaube It looks like that's what [the `scrollable` prop does](https://github.com/trussworks/react-uswds/blob/main/src/components/Table/Table.tsx#L68), so mission accomplished :)

frontend-react/src/utils/ScreenWidth.ts
```diff
@@ -0,0 +1,11 @@
+export default function isScreenWidth(size: "mobile" | "tablet" | "desktop") {
+    const currentScreenWidth = window.innerWidth;
+
+    if (size === "mobile") {
+        return currentScreenWidth <= 480;
```
stephenkao, line 5: Hmmm this is a minor point, but I'm a little wary of hardcoded values like this in our JavaScript when they're actually defined [in the USWDS CSS](https://github.com/uswds/uswds/blob/develop/packages/uswds-core/src/styles/tokens/units/spacing.scss).  We could write a utility function to determine which breakpoint we're currently on: add an invisible element, add a breakpoint-specific modifier class, check if it's applied.  Otherwise, I think we should make these values enums so that 1) they're in one place for easier modification and 2) we can reference them to pass into this function -- and possibly others in the future.  

What do you think?

etanb: It seems like we define our own set of breakpoints in `_uswds_settings.scss` with:

https://github.com/CDCgov/prime-reportstream/blob/dc43aea6730b930634c5732f018fa464b1d376bf/frontend-react/src/styles/_uswds_settings.scss#L48-L51

So I could see defining those breakpoints in one place like you suggested!

etanb:
+1 to what Stephen said about the `scrollable` prop. 

On a more important note, and to make things more complicated, what do y'all think of going `stacked` for smaller view widths? 

Screen Shot 2022-12-20 at 9 29 06 AM

This would require refactoring `Table` across the site to utilize the `data-label` property across all `<td>` but it obviously looks quite nice. What do we think?

[Link for reference.](https://designsystem.digital.gov/components/table/#component-variants-table)

frontend-react/src/utils/ScreenWidth.ts
```diff
@@ -0,0 +1,11 @@
+export default function isScreenWidth(size: "mobile" | "tablet" | "desktop") {
```
stephenkao, line 1: Minor point, but it might be nice to add convenience functions like `isMobile`, `isTablet`, `isDesktop` so we don't have to always know what to pass into this.  This function can still be the thing we use under the hood, though:

```js
export function isMobile() {
  return isScreenWidth("mobile");
}

export function isTablet() {
  return isScreenWidth("tablet");
}

...
```

Also, I think if this is going to be a general utility file for screen width operations, it may not make sense to export a single function as a default.

stephenkao, line 1: Would you mind adding tests for this?

Could be as easy as something like:
```js
import isScreenWidth, { AvailableScreenWidths } from "./ScreenWidth";

// UNTESTED
describe("isScreenWidth", () => {
    function setWindowWidth(width: number) {
        // I think innerWidth is a read-only property so we can get around it this way 🤷 
        Object.defineProperty(window, "innerWidth", {
            configurable: true,
            writable: true,
            value: width,
        });
    }
    const originalWindowWidth = window.innerWidth;

    afterAll(() => {
        setWindowWidth(originalWindowWidth);
    });

    describe("when on a mobile viewport", () => {
        beforeEach(() => {
            setWindowWidth(480);
        });

        test("it only returns true for 'mobile'", () => {
            expect(isScreenWidth("mobile")).toEqual(true);
            expect(isScreenWidth("tablet")).toEqual(false);
            expect(isScreenWidth("desktop")).toEqual(false);
        });
    });

   describe("when on a tablet viewport", () => {
        beforeEach(() => {
            setWindowWidth(768);
        });

         test("it only returns true for 'tablet'", () => { ... });
   });

   // ...and so on
});

```

stephenkao:
> +1 to what Stephen said about the `scrollable` prop.
> 
> On a more important note, and to make things more complicated, what do y'all think of going `stacked` for smaller view widths?
> 
> Screen Shot 2022-12-20 at 9 29 06 AM
> 
> This would require refactoring `Table` across the site to utilize the `data-label` property across all `<td>` but it obviously looks quite nice. What do we think?
> 
> [Link for reference.](https://designsystem.digital.gov/components/table/#component-variants-table)

I think we can stack tables on smaller viewports, but my vote would be to reach out to design for their green light and break that out into a separate ticket so we don't have to complicate this one.  Whatchu think?

frontend-react/src/utils/ScreenWidth.ts
etanb: This totally makes sense -- shouldn't have to rely on the developer to know the potential values, should just export them as semantic functions.

etanb: Sounds totally reasonable!

etanb:
Synced up with the design team, we're going to go with the `scrollable` option for now and in the new year we can revisit how we use these components for a potential overhaul.

etanb:
I know this PR should not have turned into this lengthy conversation that it has, and in my heart I knew that there _must_ be a simpler answer than all this, and I'm wondering if we're ok if I just add the following CSS to `usa-table`:

```
display: block;
overflow-x: auto;
```

Which makes the `Table` component naturally scrollable on smaller width BUT doesn't affect the desktop width like adding the prop `scrollable` does. From my cursory research, the `scrollable` prop adds `overflow` rules to the _parent_ component of the actual `<table/>` element which is why it's so wide. Benefits:

1. We add the aforementioned addition to `_uswds_overrides.scss` and now all of our `Table` components are inherently scrollable. No need to go through every `Table` and add special props.
2. We don't need any unneeded browser width detection with JS.

Demo:


https://user-images.githubusercontent.com/516877/210905826-2ceaeb43-c85e-45bc-9778-4c880101bc30.mp4


Sound good @stephenkao @kevinhaube ?

kevinhaube:
@etanb Your hunch is right! Check out [this documentation](https://designsystem.digital.gov/components/table/) and how they achieve scrollable tables in the component code. It looks like we can utilize uswds's styles instead of having to extend them:

`<div class="usa-table-container--scrollable" tabindex="0">` seems to be the key!

kevinhaube:
@etanb Oh, yikes! 😬 Good catch on that. Should we also propose a change to uswds with this case as Exhibit A? I get they prefer their stacked tables for mobile, but this seems like it should already be natural behavior; y'know, responsive design principles and all!

etanb:
@kevinhaube can file an issue against USWDS!

etanb:
we'll see if if they take to it: https://github.com/uswds/uswds/issues/5096

stephenkao (approved):
Schiiiiiimple :)

PR #7713, opened 2022-12-19
Experience 7448: Fix highlighting UX

### LINKED ISSUE #7448: [Bug] Json Diff Editor gets out-of-sync when doing edits
Step:
1. Go into Edit Settings for `ignore` (`admin/orgsettings/org/ignore`)
2. Get into the Json Diff editor via the `Preview Save...` button
3. Change the Right Hand editable side by editing an existing field... but do NOT use correct json lint format. For example: change `"filters": [],` to `"filters": ["1","2"],` (which really should be:
```
"filters": [
    "1",
    "2"],
```
to be correctly formatted.
Result: The highlighting for the differences isn't correct.
image

What's happening: the Json Diffing logic assumes the json is formatted a particular way. If the format (not syntax, but the format like a linter cares about) isn't "correct", then the highlight doesn't align correctly.

We can see why the highlighting is wrong by reformatting the json:
image

Possible solutions:
 - Disable highlighting while the user is editing the json field and only refresh when it loses focus. At that point, it's probably ok to reformat the json, too.
 - Maybe have a "refresh" button floating in side text edit area on the upper right.

stephenkao:
This changeset updates the Organization settings highlighting UX flow to:
- Disable highlighting while the user is updating the textarea
- Add a 'Check syntax' button to reformat + highlight errors/changes
- Update the blur listener to reformat + highlight

It also includes a few minor/refactory updates to the code, which I'll point out in comments.

Test Steps:
1. Log in as a PrimeAdmin
2. Start the flow to edit an Organization's settings
3. Ensure that the 'Save' button is enabled by default (since settings should be valid from the first fetch)
4. Update the textarea with changes
  a. When the textarea has valid JSON:
    i. Ensure the 'Save' button is enabled
    ii. On blurring OR clicking the 'Check syntax' button, ensure that the JSON is properly formatted and changes are highlighted in yellow
  b. When the textarea has invalid JSON:
    i.  Ensure the 'Save' button is disabled
    ii. On blurring OR clicking the 'Check syntax' button, ensure that the errors are highlighted in red
5. Ensure that submissions are persisted (for valid fields)

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- Fixes #7448

### DIFF
--- a/frontend-react/src/components/Admin/CompareJsonModal.test.tsx
+++ b/frontend-react/src/components/Admin/CompareJsonModal.test.tsx
@@ -1,0 +1,237 @@
+import React, { useRef } from "react";
+import { fireEvent, render, screen } from "@testing-library/react";
+
+import {
+    ConfirmSaveSettingModal,
+    ConfirmSaveSettingModalRef,
+    CompareSettingsModalProps,
+} from "./CompareJsonModal";
+
+describe("ConfirmSaveSettingModal", () => {
+    const VALID_JSON = JSON.stringify({ a: 1 });
+    const VALID_JSON_UPDATED = JSON.stringify({ a: 1, b: 2 });
+    const INVALID_JSON = "{ nope }";
+    let errorDiffNode: HTMLElement | null;
+    let textareaNode: HTMLElement;
+    let saveButtonNode: HTMLElement;
+    let checkSyntaxButtonNode: HTMLElement;
+
+    function TestWrapper(props?: Partial<CompareSettingsModalProps>) {
+        const confirmModalRef = useRef<ConfirmSaveSettingModalRef>(null);
+
+        return (
+            <ConfirmSaveSettingModal
+                uniquid={new Date().getTime().toString()}
+                onConfirm={jest.fn()}
+                ref={confirmModalRef}
+                oldjson={VALID_JSON}
+                newjson={VALID_JSON}
+                {...props}
+            />
+        );
+    }
+
+    function renderComponent(props: Partial<CompareSettingsModalProps> = {}) {
+        render(<TestWrapper {...props} />);
+
+        textareaNode = screen.getByTestId("EditableCompare__textarea");
+        saveButtonNode = screen.getByText("Save");
+        checkSyntaxButtonNode = screen.getByText("Check syntax");
+    }
+
+    afterEach(() => {
+        jest.clearAllMocks();
+    });
+
+    describe("on initial mount", () => {
+        describe("when the updated JSON is valid", () => {
+            beforeEach(() => {
+                renderComponent();
+            });
+
+            test("enables the save button", () => {
+                expect(saveButtonNode).toBeEnabled();
+            });
+        });
+
+        describe("when the updated JSON is invalid", () => {
+            beforeEach(() => {
+                renderComponent({
+                    newjson: INVALID_JSON,
+                });
+            });
+
+            test("disables the save button", () => {
+                expect(saveButtonNode).toBeDisabled();
+            });
+        });
+    });
+
+    describe("on change", () => {
+        describe("when the updated JSON is valid", () => {
+            beforeEach(() => {
+                renderComponent();
+
+                fireEvent.change(textareaNode, {
+                    target: { value: VALID_JSON_UPDATED },
+                });
+            });
+
+            test("enables the save button", () => {
+                expect(saveButtonNode).toBeEnabled();
+            });
+        });
+
+        describe("when the updated JSON is invalid", () => {
+            beforeEach(() => {
+                renderComponent();
+
+                fireEvent.change(textareaNode, {
+                    target: { value: INVALID_JSON },
+                });
+            });
+
+            test("disables the save button", () => {
+                expect(saveButtonNode).toBeDisabled();
+            });
+        });
+    });
+
+    describe("on blur", () => {
+        describe("when the updated JSON is valid", () => {
+            beforeEach(() => {
+                renderComponent();
+
+                fireEvent.change(textareaNode, {
+                    target: { value: VALID_JSON_UPDATED },
+                });
+                fireEvent.blur(textareaNode);
+
+                errorDiffNode = screen.queryByTestId(
+                    "EditableCompare__errorDiff"
+                );
+            });
+
+            test("does not render an error diff", () => {
+                expect(errorDiffNode).toBeNull();
+            });
+
+            test("does not render an error toast", () => {
+                expect(
+                    screen.queryByText(/JSON data generated/)
+                ).not.toBeInTheDocument();
+            });
+        });
+
+        describe("when the updated JSON is invalid", () => {
+            const consoleTraceSpy = jest.fn();
+
+            beforeEach(() => {
+                renderComponent();
+
+                jest.spyOn(console, "trace").mockImplementationOnce(
+                    consoleTraceSpy
+                );
+
+                fireEvent.change(textareaNode, {
+                    target: { value: INVALID_JSON },
+                });
+                fireEvent.blur(textareaNode);
+
+                errorDiffNode = screen.getByTestId(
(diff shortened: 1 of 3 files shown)


frontend-react/src/components/EditableCompare.tsx
```diff
@@ -48,7 +49,7 @@ interface EditableCompareProps {
 export const EditableCompare = forwardRef(
     // allows for functions on components (useImperativeHandle)
     (
-        props: EditableCompareProps,
+        { jsonDiffMode, modified, onChange, original }: EditableCompareProps,
```
stephenkao, line 52: Destructuring `props` here because using `props` as a dependency for hooks is bad practice since the reference will change if _any_ prop changes.  I've updated the dependencies accordingly below.

frontend-react/src/components/EditableCompare.tsx
```diff
@@ -66,43 +67,12 @@ export const EditableCompare = forwardRef(
         // the API call into this forwardRef well say if json is valid (to enable save button)
         const [isValidSyntax, setIsValidSyntax] = useState(true);
 
-        useImperativeHandle(
```
stephenkao, line 69: Moved this [down](https://github.com/CDCgov/prime-reportstream/pull/7713/files#diff-9549b9587cf857fa88a11d9c5d5a2ad10d106f0846b78007cd5901d000c038b8R147) so we don't have an issue with referencing functions created underneath it in the code.  (As a VERY superficial aside, this is generally why I like using function declarations instead of function expressions since they're hoisted up to the top of the scope, but this doesn't apply to memoized callbacks since those will always be expression!)

frontend-react/src/components/EditableCompare.tsx
```diff
@@ -66,43 +67,12 @@ export const EditableCompare = forwardRef(
...
-                isValidSyntax() {
-                    return isValidSyntax;
-                },
-            }),
-            // onChangeHandler appears below, remove from deps
-            // eslint-disable-next-line
-            [textAreaContent]
-        );
-
-        const turnOffSpellCheckSwigglies = () => {
```
stephenkao, line 92: I'm not sure why we needed a function for this since there's a [`spellcheck` HTML attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/spellcheck) (`spellCheck` as a React prop).  I've updated the textarea [below](https://github.com/CDCgov/prime-reportstream/pull/7713/files#diff-9549b9587cf857fa88a11d9c5d5a2ad10d106f0846b78007cd5901d000c038b8R228), but let me know if there's was a reason for this function!

frontend-react/src/components/EditableCompare.tsx
```diff
@@ -115,12 +85,13 @@ export const EditableCompare = forwardRef(
 
                     const threeParts = splitOn(modifiedText, start, end);
                     // we're using HTML5's <s> tag to show error, style sets background to red.
-                    const errorHtml = `${threeParts[0]}<s>${threeParts[1]}</s>${threeParts[2]}`;
+                    const errorHtml = `${threeParts[0]}<s data-testid="EditableCompare__errorDiff">${threeParts[1]}</s>${threeParts[2]}`;
```
stephenkao, line 88: Added this test id so we can find it in tests: https://github.com/CDCgov/prime-reportstream/pull/7713/files#diff-cd109c0ecef22f0c55512e1aad07a73668fc622e5793c1e7fa1dc6aec2f7df6eR110

frontend-react/src/components/EditableCompare.tsx
```diff
@@ -207,6 +219,8 @@ export const EditableCompare = forwardRef(
                                 onBlur={(e) => {
                                     onBlurHandler(e?.target?.value || "");
```
kevinhaube, line 220: This makes our new button irrelevant, no? Perhaps instead of the button and the `onBlur`, could we debounce an `onChange` event so that it'll only diff after an idle of like 3 seconds? The UX as is feels a bit odd; like, halfway between trying to be automatic and manual.

kevinhaube (commented):
Non-blocking UX nit

frontend-react/src/components/EditableCompare.tsx
stephenkao: Sure, that's a fair point!  For context, this came out of a discussion between Penny, Tom, and me [in this pull request](https://github.com/CDCgov/prime-reportstream/pull/7639).  I can update this so that the blur doesn't handle the automatic reformatting/highlighting, but I think a timed update -- even with a debounce -- could introduce more confusion unless we had helper text saying when the update would be happening.  Plus, the user will lose their cursor positioning if we automatically reformat when the textarea has focus.

penny-lischer:
I see one minor issue when saving before the user clicks on the 'Check Syntax". A JSON error will appear if the JSON has not been reformatted. 

https://user-images.githubusercontent.com/102491809/208483134-494b5a8e-4c34-4814-bc42-7759f287bfa4.mp4

frontend-react/src/components/Admin/CompareJsonModal.test.tsx
```diff
@@ -0,0 +1,237 @@
...
+                    });
+
+                    errorDiffNode = screen.queryByTestId(
+                        "EditableCompare__errorDiff"
+                    );
+                    expect(errorDiffNode).toBeNull();
+                });
+            });
+
+            test("when the user starts typing again", () => {});
```
penny-lischer, line 177: Was this left here as a placeholder?

frontend-react/src/components/EditableCompare.tsx
penny-lischer: I would't go that route. If you want to remove the onBlur then ensure that the Save is disabled anytime the textarea is changed so that the user has to click on the "Check syntax" button.

kevinhaube: Yeah, after reading the issues with the old UX, definitely more in favor of losing `onBlur` and disabling `Save` until they `Check Syntax`! 💯

frontend-react/src/components/Admin/CompareJsonModal.test.tsx
stephenkao: Whoops, nope!  That's just my eyes failing me :) Thanks for catching -- will remove!

stephenkao: Removed in this commit: https://github.com/CDCgov/prime-reportstream/pull/7713/commits/f58b30724999df81bf11614f398e907531f53e8e

frontend-react/src/components/EditableCompare.tsx
stephenkao: Thanks, everyone!  I just pushed up a commit here to only enable the 'Save' button if the 'Check syntax' shows that it's valid JSON: https://github.com/CDCgov/prime-reportstream/pull/7713/commits/f58b30724999df81bf11614f398e907531f53e8e



Let me know what y'all think! 🙏

stephenkao: Oh, duh -- need to update the new tests!  Feel free to take a look at the code, and I'll push up the updated tests in the meantime.

penny-lischer (approved):
LGTM

TomNUSDS (approved):
LG(reat)TM!

Unit tests, useful comments and plenty of fixing techdebt as you went. Thanks!

frontend-react/src/components/EditableCompare.tsx
TomNUSDS: I vaguely remember struggling to get the HTML attribute approach to work, but maybe that issue long resolved?

See: https://stackoverflow.com/questions/49079148/reactjs-remove-spellcheck-attribute

stephenkao:
Thanks, everyone!  Will merge after tests pass 🙏

PR #7727, opened 2022-12-19
Sync login state across multiple tabs #closes #6868

### LINKED ISSUE #6868: Allow users to maintain login status between tabs
## Problem statement

As a logged in report stream user I would like to be able to open a report stream page in a new tab without having to log in again.

## What you need to know

The solution to this is storing login status in localStorage rather than sessionStorage. Our current system relies on sessionStorage as the source of truth for log in status, and this information is not shared between tabs, while localStorage is.

Logging in or out on one tab should also reflect on any other open tabs.

## Acceptance criteria

- Remain logged in when opening a new tab while logged in
- All open tabs log in or out when logging in or out
  - i.e. if I login/logout on one tab, other tabs should login/logout too
- No refreshes should be performed to sync a tab
  - Ideally no navigations either (except for login/logout redirect as appropriate)

## Design

1. Introduce a `session` broadcast channel which will be used to broadcast `login` and `logout `messages when the user logs in or out.

2. Listen to the broadcast channel and react accordingly (redirect as appropriate and ensure app state reflects session state)
  1. okta's `updateAuthState` function handled this nicely

## Notes

- Currently the [code](https://github.com/CDCgov/prime-reportstream/blob/thetaurean/6868-allow-users-to-maintain-login-status-between-tabs/frontend-react/src/App.tsx#L69) handling broadcast messages does not differentiate between `login` and `logout` messages because the handling code is the same in either case.
- On logout we always trigger other tabs to logout, even if logout failed. Maybe we don't want to trigger in some cases? Relevant finally statement [here](https://github.com/CDCgov/prime-reportstream/blob/thetaurean/6868-allow-users-to-maintain-login-status-between-tabs/frontend-react/src/utils/UserUtils.ts#L6)

thetaurean:
This PR adds authentication state synchronization between tabs to our frontend web app. Using localStorage to share session data and broadcast channels to trigger session events. See linked issue for further info on design et. al.

Test Steps:
1. Login on a tab
2. Find a link (i.e. "Product") and open it in a new tab
    - **ASSERT** You are still logged in on the new tab
3. Click logout on either tab
    - **ASSERT** You are now logged out of the other tab
4. Login on either tab
    - **ASSERT** You are now logged in on the other tab

## Changes
- Updated okta configuration to use `localStorage` instead of `sessionStorage`
- Emit `login` and `logout` messages on new `session` broadcast channel
- Watch the `session` broadcast channel and call `updateAuthState` when a message is rx'd

### Testing
- [X] Tested locally?
- [ ] ~~Ran `./prime test` or `./gradlew testSmoke` against local Docker ReportStream container?~~ **No backend changes**
- [ ] ~~Added tests?~~ **No backend changes**

### Process
- [ ] ~~Are there licensing issues with any new dependencies introduced?~~ **No new dependencies**
- [X] Includes a summary of what a code reviewer should test/verify? **Yes**
- [ ] Updated the release notes? **Should I do this?**
- [ ] ~~Database changes are submitted as a separate PR?~~ **No database changes**
- [ ] ~~DevOps team has been notified if PR requires ops support?~~ **None needed (that I'm aware of)**

## Linked Issues
- Fixes #6868

## Notes

- Currently the [code](https://github.com/CDCgov/prime-reportstream/blob/thetaurean/6868-allow-users-to-maintain-login-status-between-tabs/frontend-react/src/App.tsx#L69) handling broadcast messages does not differentiate between `login` and `logout` messages because the handling code is the same in either case.
- On logout we always trigger other tabs to logout, even if logout failed. Maybe we don't want to trigger in some cases? Relevant finally statement [here](https://github.com/CDCgov/prime-reportstream/blob/thetaurean/6868-allow-users-to-maintain-login-status-between-tabs/frontend-react/src/utils/UserUtils.ts#L6)

    - new endpoint A
    - new endpoint B


kevinhaube:
Hey, this looks awesome! I don't think I knew about `BroadcastChannel` until this. I'm out for the next few days, and I wanna make sure it gets a proper review, so I'm gonna pass it over to @stephenkao 🚀

thetaurean:
> Hey, this looks awesome! I don't think I knew about `BroadcastChannel` until this. I'm out for the next few days, and I wanna make sure it gets a proper review, so I'm gonna pass it over to @stephenkao 🚀

My first time using BroadcastChannel as well actually! Thanks for the prompt reply and passing it on.
Enjoy your time off the next few days!

frontend-react/src/App.tsx
```diff
@@ -64,6 +64,12 @@ const App = () => {
         navigate(toRelativeUrl(originalUri, window.location.origin));
     };
 
+    // watch the session broadcast channel to handle login/logout on other tabs
+    const bc = new BroadcastChannel("session");
```
stephenkao, line 68: Ah, this'll run on every re-render of the App component, which happens on navigation, context change, and auth state change -- so we'll be creating new instances of BroadcastChannel and adding a listener each time.  Let's use a `useRef` and a `useEffect` so it'll only do that once:

```jsx
// only create one BroadcastChannel instance for this component
const broadcastChannelRef = useRef(new BroadcastChannel("session"));

// only bind the message listener on mount
useEffect(() => {
  broadcastChannelRef.current.onmessage = () => {
    OKTA_AUTH.authStateManager.updateAuthState();
  };

  // unbind the message listener on unmount
  return () => broadcastChannelRef.current.close();
}, []);
```

frontend-react/src/App.tsx
```diff
@@ -64,6 +64,12 @@ const App = () => {
...
+    bc.onmessage = function () {
```
stephenkao, line 69: Since BroadcastChannel can be used for any event, let's narrow it down by checking the name:

```jsx
bc.onmessage = (event) => {
  // could also do something with event.data, which is 'login' or 'logout'
  if (event.currentTarget.name === 'session') {
     OKTA_AUTH.authStateManager.updateAuthState();
  }
};

// Side note: I think we could also use attachEventListener, 
// which is more modern since on<event> nomenclature was from the olden days of HTML event-binding:
bc.addEventListener('message', (event) => {
  if (event.currentTarget.name === 'session') {
     OKTA_AUTH.authStateManager.updateAuthState();
  }
});
```

ALSO, what do you think about enumifying the 'session', 'login', and 'logout'?  At least to prevent typos going forward :)

frontend-react/src/utils/UserUtils.ts
```diff
@@ -3,12 +3,15 @@ import { OktaAuth } from "@okta/okta-auth-js";
 function logout(oktaAuth: OktaAuth): void {
     if (oktaAuth?.signOut) {
         try {
-            oktaAuth.signOut();
+            oktaAuth.signOut().finally(() => {
+                const bc = new BroadcastChannel("session");
```
stephenkao, line 7: Hmmmm now that I see this, can we just create ONE BroadcastChannel for the entire lifetime of the app?  That way, we won't have to worry about needing to unbind/close [up here](https://github.com/CDCgov/prime-reportstream/pull/7727/files#r1053662183), and we'll have one "global" instance to funnel our session events through.

frontend-react/src/App.tsx
thetaurean: 👍

frontend-react/src/App.tsx
thetaurean: Great point on enumifying, do you have a good suggestion for where that enum should live? Perhaps `SessionStorageTools` or `UserUtils` ?

Should not be any need to check for `session`. The BroadcastChannel is named on instantiation (`new BroadcastChannel("session")`) so these event listeners will only ever receive `session` messages. 

I totally agree we might want to check for `logout` vs `login` (as mentioned in Notes), but for now those are the only two actions and they are both handled the same, so I vote we add that check later on.

frontend-react/src/utils/UserUtils.ts
thetaurean: 👍 Will find a place for the ref & its instantiation 🦖 . Perhaps the `AuthStateManager`'s data? 🤔

frontend-react/src/App.tsx
stephenkao: SessionStorageTools makes sense to me since that's likely the only place we'll be using this for now!

The idea of checking for `session` was because we're creating two `session` instances of BroadcastChannel [here](https://github.com/CDCgov/prime-reportstream/pull/7727/files#diff-f69b57e78f568071b4f3cab66ca28cc47c5efae1b0959c5f4eef858538b1ac02R18) and [here](https://github.com/CDCgov/prime-reportstream/pull/7727/files#diff-64a2066f31389cb7d53d6360524cfcb49c7fecd153a2db15db0aa49b92f954efR7) -- if we update the name for whatever reason and forget to change it in one place, that'll fail to respond properly to the events we're sending.  HOWEVER, this is a moot point if we only choose [to create one instance](https://github.com/CDCgov/prime-reportstream/pull/7727/files#r1053669075) like we're discussing below!

frontend-react/src/utils/UserUtils.ts
stephenkao: In my mind, there are a couple of ways we can handle this:

// Method 1: instantiate as soon as JavaScript executes
// This naturally ensures that we'll only create one BroadcastChannel for the lifetime of the app,
// but it can be a little bit harder for test environments since it relies on a "global" (not really global, just not in the scope of any other function)
```js
// in UserUtils.ts
export const sessionBroadcastChannel = new BroadcastChannel("session")
sessionBroadcastChannel.addEventListener(...);

// then we can import sessionBroadcastChannel if we want to do anything else with it
// in SomeOtherFile.ts
import { sessionBroadcastChannel } from './UserUtils';
sessionBroadcastChannel.addEventListener(...);
sessionBroadcastChannel.close();
```

// Method 2: wrap the instantiation and close in functions
// This is safer, and we can add checks to prevent duplicity, but it requires a bit more code:
```js
// in UserUtils.ts

// initialize to null in the "global" scope
let sessionBroadcastChannel = null;

export function initializeSessionBroadcastChannel() {
  // don't reinitialize
  if (sessionBroadcastChannel) {
    return sessionBroadcastChannel;
  }

  sessionBroadcastChannel = new BroadcastChannel("session");
  return sessionBroadcastChannel;
}

export function closeSessionBroadcastChannel() {
  sessionBroadcastChannel?.close(); // close channel and mark for garbage collection
  sessionBroadcastChannel = null;
}

export function getSessionBroadcastChannel() {
  return sessionBroadcastChannel;
}

// we'll have to import and invoke these functions instead of the straight BroadcastChannel instance
// in SomeOtherFile.ts
import { initializeSessionBroadcastChannel } from './UserUtils';

broadcastChannel = initializeSessionBroadcastChannel();
broadcastChannel.addEventListener(...)
```

We could add other conveniences to respond to `login` and `logout` events automatically, but we gotta go to sprint kickoff now!  Happy to chat more about this later, though!  And sorry for the wall o' text 😆

frontend-react/src/App.tsx
```diff
@@ -64,6 +64,8 @@ const App = () => {
         navigate(toRelativeUrl(originalUri, window.location.origin));
     };
 
+    initializeSessionBroadcastChannel(OKTA_AUTH); // for cross-tab login/logout
```
stephenkao, line 67: Let's move this call outside of the component body.  It's not a big deal because we're early terminating if we've already initialized, but there's no reason to call this on each re-render of the component:

```js
initializeSessionBroadcastChannel(OKTA_AUTH);

const App = () => { ... };
```

frontend-react/src/utils/UserUtils.ts
```diff
@@ -1,14 +1,49 @@
...
+
+export function initializeSessionBroadcastChannel(oktaAuth: OktaAuth) {
+    if (sessionBroadcastChannel) {
+        return sessionBroadcastChannel; // don't reinitialize
+    }
+
+    sessionBroadcastChannel = new BroadcastChannel(SESSION_CHANNEL.NAME);
+    sessionBroadcastChannel.onmessage = function (e) {
+        oktaAuth.authStateManager.updateAuthState();
+        if (e.data === "logout") closeSessionBroadcastChannel();
```
stephenkao, line 34: Can we use the enum here?

```js
if (e.data === SESSION_CHANNEL.LOGOUT)
```

frontend-react/src/utils/UserUtils.ts
```diff
@@ -1,14 +1,49 @@
 import { OktaAuth } from "@okta/okta-auth-js";
 
+let sessionBroadcastChannel: BroadcastChannel | null = null;
+
+export enum SESSION_CHANNEL {
```
stephenkao, line 5: Sorry, I think this is on me, but it looks like the convention in this codebase is to use PascalCase for enums:

```js
export enum SessionChannel { ... }
```

Also, I don't think it makes sense to have the channel name and events grouped in the same enum since they're used for different things:

```js
// used to identify the channel
// (all caps used here to signify a top-level constant variable)
export const SESSION_CHANNEL_NAME = "session";

// used to identify events passed through the channel
export enum SessionEvent {
    Login = "login",
    Logout = "logout"
}
```

stephenkao (commented):
Left a few more comments, but this is looking good to me overall!

This might be out of the scope of this pull request, but one question I have is how much of the website do we want to react to the changing of the LocalStorage data?  Say if you log in as an admin and have two tabs open.  Then in tab one, you impersonate 'abbott' -- tab one updates properly to display that you're impersonating 'abbott', but tab two still says 'PrimeAdmins':

Screenshot 2022-12-28 at 3 24 32 PM

Again, this may be out of the scope of this pull request, we may want to set up another listener on the channel so we can re-render accordingly.  (cc @kevinhaube for your thoughts!)

frontend-react/src/App.tsx
thetaurean: Agreed good call 👍

frontend-react/src/utils/UserUtils.ts
thetaurean: 👍

frontend-react/src/utils/UserUtils.ts
thetaurean: Whoopsie missed one (:

thetaurean:
Thanks for all the great feedback @stephenkao !

And hmm good question on how much of the site is sync'd. In your example use case I like the ability of retaining two separate identities rather than having to incognito if I need that.

Usually I've seen this used to refresh data on one tab changed by another, as a workaround or in addition to websockets. Definitely a good topic for product to chew on.

Another thought, this is a great use case for redux stores although I know hooks are the du jour. There's a nice plugin that can sync stores across tabs while leaving some alone as desired. But I'm off on a tangent :P.

stephenkao (approved):
Looks good to me.  We can file a ticket to handle the outdated UI issue at some point, but in my mind it's a super low-priority item that we don't need to address anytime soon.  

Thanks for handling this!

thetaurean:
> 

mmmm brain food. Good notes and conversation.

👍 Was a delight contributing to this codebase.

PR #7736, opened 2022-12-20
7139 - prevent org names with special characters/reserved words

### LINKED ISSUE #7139: New Settings names (and Clone) should prevent any names with special characters (e.g. `.`)
Lots of code on the server depend on specific naming conventions. If Setting names (or Org names) include a specific set of character (or even words), then random problems will happen.

Ideally, this would be enforced at the API/pipeline layer, but a good place to start is the frontend UI.

 - [x] Exclude following characters from all names: 
  - `"."` (period) because it's often used as a delimiter for `[org].[setting]` 
  - `" "` (space)
  - `"/"` (forward slash)
  - Any non-ascii characters. (e.g. UTF8)
  - URL reserved characters: [e.g. `&`, `+`, `,`, `=`, `?`, etc]. 

The best approach might be to call a urlEncode function on the string and verify it doesn't change.

 - [x] Prohibited names:
  - `"all"` (`api/check` issue)
  - `"revs"` (`org/{organizationName}/senders/revs` issue)
  - `"organization"`, `"sender"`, `"receiver"`, `"org"`, `"list"` because they are often used as cmd actions.
  
Bonus Lookup table:
 - [ ] Exclude all URL reserved characters (see above for how).
 - [ ] Exclude Lookup table named `"list"` (`/api/lookuptables/` issues)

penny-lischer:
Fixes #7139 This PR ...

Prevents org names with special characters/reserved words

Test Steps:
1. Log into RS
2. Edit an Organization
3. Select 'New' under 'Organization Sender Settings' or 'Organization Receiver Settings'
4. Try to enter a non-ascii character or a prohibited word and click 'Save'
5. You should not be able to save.
6. Click 'Cancel'
7. 'Clone' an existing setting
8. Try to enter a non-ascii character or a reserved word and click 'Save'
9. You should not be able to save.
*prohibited words - 
        "sender",
        "receiver",
        "org",
        "organization",
        "list",
        "all", 
        "revs"

## Changes
- Added logic when Saving
Screenshot 2022-12-20 at 2 12 46 PM

### Testing
- [x] Tested locally?
- [x] Added tests?


frontend-react/src/utils/misc.ts
```diff
@@ -46,6 +46,38 @@ export const checkJson = (
...
+    orgName: string
+): { prohibited: boolean; errorMsg: string } => {
+    const prohibitedNames = [
+        "sender",
+        "receiver",
+        "org",
+        "organization",
+        "list",
+        "all", //api/check issue
+        "revs", //org/{organizationName}/senders/revs issue
```
TomNUSDS, line 63: nice catch!

frontend-react/src/utils/misc.ts
```diff
@@ -46,6 +46,38 @@ export const checkJson = (
     }
 };
 
+export function includesSpecialChars(text: string): boolean {
+    return /[^a-z\d\s]/gi.test(text);
+}
+
```
TomNUSDS, line 52: Probably some unit tests would be useful?

I think `includesSpecialChars("a\ttest")` returns false (e.g. someone copy/pastes a tab or linefeed character)

Also,
`encodeURI("a\ttest") !== "a\ttest"` returns true. But calling encodeURI maybe is slightly more expensive?


There also might need to be a `trim()` before checking? Not sure what level that might happen at.

frontend-react/src/utils/misc.ts
```diff
@@ -46,6 +46,38 @@ export const checkJson = (
...
+        "sender",
+        "receiver",
+        "org",
+        "organization",
+        "list",
+        "all", //api/check issue
+        "revs", //org/{organizationName}/senders/revs issue
+    ];
+
+    if (prohibitedNames.indexOf(orgName) > -1) {
```
TomNUSDS, line 66: I _think_ this is a case sensitive check. So would we want to allow `All`? (I'm not sure, but probably not.)

Maybe the orgName should be lowerCased at the start of this function so it doesn't depend on it happening higher up?

frontend-react/src/components/Admin/EditSenderSettings.test.tsx
```diff
@@ -89,8 +93,52 @@ describe("EditSenderSettings", () => {
...
+            jest.spyOn(console, "trace").mockImplementationOnce(
+                consoleTraceSpy
+            );
+        });
+
+        afterEach(() => {
+            jest.resetAllMocks();
+        });
+
+        test("should display an error if name value is prohibited", async () => {
```
stephenkao, line 114: Minor point, but this doesn't need to be `async` since nothing's being `awaited` :)

frontend-react/src/components/Admin/EditSenderSettings.test.tsx
```diff
@@ -89,8 +93,52 @@ describe("EditSenderSettings", () => {
...
+            fireEvent.change(nameField, {
+                target: { value: "Organization" },
+            });
+            expect(nameField).toHaveValue("Organization");
+
+            fireEvent.click(editJsonAndSaveButton);
+            expect(consoleTraceSpy).toHaveBeenCalled();
+        });
+
+        test("should display an error if name value contains a non-ascii char", () => {
```
stephenkao, line 124: Very, very, very pedantic point, but these are technically valid ASCII characters.  Might be better to name it a "non-alphanumeric" character, but a whitespace is also non-alphanumeric 😆

frontend-react/src/utils/misc.ts
```diff
@@ -46,6 +46,38 @@ export const checkJson = (
     }
 };
 
+export function includesSpecialChars(text: string): boolean {
```
stephenkao, line 49: Is it worth it to combine this with `isProhibitedOrgName`?  It's a pretty specific regular expression, and the name of the function is a little generic 🤔

frontend-react/src/utils/misc.ts
```diff
@@ -46,6 +46,38 @@ export const checkJson = (
...
+    return /[^a-z\d\s]/gi.test(text);
```
stephenkao, line 50: Does this need to be global?  (re: the `g` flag at the end)

frontend-react/src/components/Admin/EditSenderSettings.test.tsx
TomNUSDS: Yeah, I was wondering if "Non-url-escapable-characters" would be a better way to describe it? IDK. it's also the same syntax that's allowed for a (compiler) variable name (e.g. can't start with a number).

frontend-react/src/utils/misc.ts
TomNUSDS: Good point since the same rules apply for a new org name in theory.

And thinking about it, maybe also Value Sets?

frontend-react/src/utils/misc.ts
TomNUSDS: Good point since the same rules apply for a new org name in theory.

And thinking about it, maybe also Value Sets?

frontend-react/src/components/Admin/EditSenderSettings.test.tsx
penny-lischer: will remove

frontend-react/src/utils/misc.ts
penny-lischer: I can combine it but had it as its own function if we were to use it elsewhere and also for testing purposes. I felt that it was unique enough to separate.

frontend-react/src/utils/misc.ts
penny-lischer: No, not necessary

frontend-react/src/utils/misc.ts
stephenkao: Sure, not a big deal in my mind at this point.  I do think it's worth updating the name of the function to be a little more indicative of what it's checking for, but also not a blocker!

stephenkao (approved):
Looks good to me!

PR #7779, opened 2022-12-28
Experience-7607: Capture FE exceptions in Azure

### LINKED ISSUE #7607: Capturing exceptions from the React app in Azure (trackException)
## Problem statement

Our React application is likely to throw critical errors when utilizing third party packages to support core functionality. We do our due diligence, but software can be unpredictable, and logs are a key site reliability tool we can use to prevent production outages of functionality. 

To capture a log of warnings and errors to aid us in site maintenance and bug troubleshooting, **we should implement Azure App Insights to capture the output of any console warning and error.**

## What you need to know

Our sister application, SimpleReport, seems to have hijacked the console methods and implemented a "track exception" function for app insights when it's a warning or error: [code snippet](https://github.com/CDCgov/prime-simplereport/blob/main/frontend/src/app/TelemetryService.ts#L64-L87)

Azure has _some_ kind of default logging of these that has been triggering PagerDuty for the smallest JavaScript errors. Through conversation, @jeremy-page informed us we can use a string to filter _some_ front end messages from PagerDuty. Work with him or the DevOps team to outline an approach that lets us utilize errors and warnings, **and prevents PagerDuty from receiving unnecessary alerts**.

## Acceptance criteria

- [ ] Azure App Insights maintains a log of exceptions thrown by the React applications (console.warn, console.error)
- [ ] PagerDuty will not be alerted for unnecessary errors in the React application
- [ ] PagerDuty **will** be notified for unexpected failures that require attention

## To do

- [ ] _Use this section to outline your approach as you work, if you'd like!_

stephenkao:
This PR ...

This changeset reverts the revert in https://github.com/CDCgov/prime-reportstream/pull/7740.  There are two main changes from that revert:

1) Always pass an instance of Error to `trackException`
Originally, we were passing in either an Error or a string depending on the arguments `console.error` was called with.  However, `exception` should be an instance of [Error or IAutoExceptionTelemetry](https://github.com/microsoft/ApplicationInsights-JS/blob/master/shared/AppInsightsCommon/src/Interfaces/IExceptionTelemetry.ts#L31) or we'll end up with a `not_specified` for the exception message.

2) Only tracking error-level messages
This is a deviation from SimpleReport so I'm happy to discuss and update if needed, but my feeling is that we don't have to track warnings just yet until we have a proper funnel or response to unexpected behaviors.  Errors are much more significant to track, and while they -- like warnings -- can be triggered from anywhere (by third-party scripts, by browsers themselves, et cetera), they're generally more actionable or intentional than warnings.

Also included:
- Some minor cleanup
- Rewritten tests + new ones to check for the changes

Test Steps:
1. (before starting local Webpack server) Populate `REACT_APP_APPINSIGHTS_KEY`, `REACT_APP_APPLICATIONINSIGHTS_CONNECTION_STRING` env vars
2. Open up the console
3. Trigger info, log, and warning and verify they're getting tracked as **events** with the proper data
Screenshot 2022-12-28 at 10 29 17 AM
Screenshot 2022-12-28 at 10 29 29 AM
Screenshot 2022-12-28 at 10 30 42 AM

4. Trigger error and verify it's getting tracked as an **exception** with the proper data
Screenshot 2022-12-28 at 10 28 36 AM
Screenshot 2022-12-28 at 10 29 07 AM
Screenshot 2022-12-28 at 10 29 51 AM

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- Fixes #7607

### DIFF
--- a/frontend-react/src/TelemetryService.test.ts
+++ b/frontend-react/src/TelemetryService.test.ts
@@ -1,0 +1,90 @@
+import { SeverityLevel } from "@microsoft/applicationinsights-web";
+
+import { ai, getAppInsights, withInsights } from "./TelemetryService";
+
+jest.mock("@microsoft/applicationinsights-web", () => {
+    return {
+        ...jest.requireActual("@microsoft/applicationinsights-web"),
+        ApplicationInsights: function () {
+            return {
+                loadAppInsights() {},
+                trackEvent: jest.fn(),
+                trackException: jest.fn(),
+                addTelemetryInitializer: jest.fn(),
+            };
+        },
+    };
+});
+
+const oldEnv = process.env.REACT_APP_APPLICATIONINSIGHTS_CONNECTION_STRING;
+
+describe("TelemetryService", () => {
+    beforeEach(() => {
+        jest.spyOn(console, "warn").mockImplementation(() => {});
+        jest.spyOn(console, "log").mockImplementation(() => {});
+        jest.spyOn(console, "error").mockImplementation(() => {});
+        jest.spyOn(ai, "initialize");
+    });
+
+    afterEach(() => {
+        jest.resetAllMocks();
+        process.env.REACT_APP_APPLICATIONINSIGHTS_CONNECTION_STRING = oldEnv;
+    });
+
+    it("initializes the appInsights service", () => {
+        process.env.REACT_APP_APPLICATIONINSIGHTS_CONNECTION_STRING =
+            "fake-connection-string";
+        ai.initialize();
+        expect(getAppInsights()).not.toBe(null);
+    });
+
+    it("calls app insights on console methods", () => {
+        process.env.REACT_APP_APPLICATIONINSIGHTS_CONNECTION_STRING =
+            "fake-connection-string";
+        const appInsights = getAppInsights();
+        withInsights(console);
+        const message = "hello there";
+        console.log(message);
+        expect(appInsights?.trackEvent).toBeCalledWith({
+            name: "LOG - hello there",
+            properties: {
+                severityLevel: SeverityLevel.Information,
+                message,
+                additionalInformation: undefined,
+            },
+        });
+
+        const warning = "some warning";
+        const data = { oh: "no" };
+        console.warn(warning, data);
+        expect(appInsights?.trackException).toBeCalledWith({
+            id: "some warning",
+            severityLevel: SeverityLevel.Warning,
+            properties: {
+                additionalInformation: JSON.stringify([data]),
+            },
+        });
+
+        const error = new Error("bad news");
+        console.error(error);
+        expect(appInsights?.trackException).toBeCalledWith({
+            exception: error,
+            id: error.message,
+            severityLevel: SeverityLevel.Error,
+            properties: {
+                additionalInformation: undefined,
+            },
+        });
+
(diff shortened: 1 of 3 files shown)


frontend-react/src/TelemetryService.test.ts
```diff
@@ -0,0 +1,325 @@
...
+        });
+
+        afterAll(() => {
+            jest.resetAllMocks();
+
+            process.env.REACT_APP_APPLICATIONINSIGHTS_CONNECTION_STRING =
+                oldConnectionString;
+        });
+
+        describe("when calling console.info", () => {
```
stephenkao, line 160: I can use `describe.each`/`test.each` to reduce some redundancy here, but I generally try not to over-optimize in tests to make them as readable/updatable as possible.  I can update this if it's an issue, though!

frontend-react/src/TelemetryService.test.ts
```diff
@@ -0,0 +1,325 @@
...
+                        id: error.message,
+                        severityLevel: SeverityLevel.Error,
+                        properties: {
+                            additionalInformation: undefined,
+                        },
+                    });
+                });
+            });
+
+            describe("when passing in a String as the first parameter", () => {
```
stephenkao, line 289: This was the case that was causing the `not_specified` errors to be added, so I'm just casting to an Error instance now.  I'm curious if SimpleReport has encountered this at all since they're allowing `undefined` [here](https://github.com/CDCgov/prime-simplereport/blob/main/frontend/src/app/TelemetryService.ts#L65)!

@kevinhaube @penny-lischer if you get a chance, would you mind checking in on the Azure side to see if no longer getting `not_specified` errors?  I don't have access so the most I can do is verify that we're sending the proper request payload 😬

frontend-react/src/TelemetryService.ts
```diff
@@ -54,12 +57,77 @@ const createTelemetryService = () => {
...
+}
+
 export function getAppInsightsHeaders(): { [key: string]: string } {
     return {
-        "x-ms-session-id": getAppInsightsSessionId(),
+        "x-ms-session-id": appInsights?.context.getSessionId() || "",
     };
 }
 
-function getAppInsightsSessionId(): string {
```
stephenkao, line 63: This doesn't really need to be a separate function since 1) it's only used in one place and 2) it wasn't even being exported.  🤷

frontend-react/src/TelemetryService.test.ts
```diff
@@ -0,0 +1,325 @@
...
+            ({
+                ai,
+                getAppInsights,
+                getAppInsightsHeaders,
+                withInsights,
+            } = require("./TelemetryService"));
+        });
+    });
+
+    describe("#initialize", () => {
```
stephenkao, line 46: Backfilled a bunch of tests along the way!  The ones in [this describe](https://github.com/CDCgov/prime-reportstream/pull/7779/files#diff-d9d00c1fd91c8fd306ae3d32c0ecd6620d304cf3638678f8bfacf3ff9f5285beR136) are the only ones that are actually relevant to this changeset, but feel free to review as needed!

penny-lischer (approved):
LGTM

kevinhaube (approved):

Issue #7730, opened 2022-12-20
Create a single custom Link component to replace all usage of <a> tags and other Links within RS


etanb:
## Problem statement

We have a few ways of navigating around the RS site, we have `<NavLink/>`, `<Link/>` and raw `<a/>` tags that use `DOMPurify` to clean the HREF. The data that then gets provided to these links are either strings hard-coded into those components or referenced from site.json. There's just a lot of repetition OR inconsistency in our links overall that could benefit from them all living in a single component so that when we link to things, we're consistent! 

In addition, many of our links use `target="_blank"` to open them in a new tab, which is fine, but we should be adding `rel="noopener noreferrer"` for security as explained in this post: https://getfrontend.tips/always-use-noopener-or-noreferrer-for-links-opened-in-new-tabs/

## Acceptance criteria

- [ ] A single <RSLink /> component that handles all links on the RS website.
- [ ] A custom lint config that warns devs about using raw <a> tags?

Issue #7627, opened 2022-12-09
Session duration with membership metadata


kevinhaube:
## Problem statment

Usage statistics for our website are a mystery. We know that people hit the site, because we see unique session ids coming through production app insights. We **don't** know, however, how long they're using the site for, nor can we relate their session ID to a set of actions. To better understand how long different types of users are on the site, **app insights should show a breakdown of average session times per memberType**, and when observing the tracked event, I should be able to pull a session ID and query the `AppEvents` table for event logs.

## Acceptance criteria

- [X] Front-end data is sent with unique session id
- [x] Session length data is sent to App Insights with session id
- [x] Azure can query the `customEvents` table and find the actions of a single session
- [x] Dashboard item shows the average time spent on the app for each memberType

kevinhaube:
Identified #7730 as the blocker, adding blocked label and starting on the fix.

etanb:
@kevinhaube would love to sync up with you on this ticket!

etanb:
Talked with Chris, and for this MVP, the concept of a session will be the following:

- begins when a user first loads up the ReportStream website
- we ignore the concept of logged in and logged out
- we will attach the member type to the response
- if possible, pause session timer when user navigates to a different tab
- ends when a user navigates away from the page which includes: closing browser, closing tab

We will also want a view in the Azure dashboard that aggregates session length counts into buckets (0-1 min, 1-5 min, 5 - 10 min, 10 - 20 min, Over 20 min) to see the distribution -- Chris.

Let me know if this sounds good to everyone! cc @stephenkao and @kevinhaube for visibility

kevinhaube:
Nice!! Love the idea of binning the times rather than taking a general average for each member type. Gives us granularity!

stephenkao:
@etanb Makes sense to me as well!

One thing--

> we will attach the member type to the response

We're already [adding the user's `activeMembership` to every AI event tracked](https://github.com/CDCgov/prime-reportstream/blob/master/frontend-react/src/TelemetryService.ts#L43) (including automatic ones like PageViewData and RemoteDependencyData) -- is there anything more we need to do here?

chris-kuryak:
@stephenkao @penny-lischer 
I am not familiar with the technical aspect of this as I'm unsure if sessions are built into AI, or if Etan is creating something new here. Any insights you can lend here would be appreciated.

I assume that if we are passing membership data with every AI event we track, then it should automatically be queryable with session data on the AI side of things?

[...]

PR #7780, opened 2022-12-28
Add new `USLink` components

kevinhaube:
This PR adds component code **that will be implemented in the next PR** to solve for links in our React app refreshing the DOM at the `App` level (the blocking behavior for #7627)

Test Steps:
1. Run the test suite to ensure they pass!

## Changes
- Introduces `USLink`, `USExtLink`, `USCrumbLink`, and `USNavLink`
- Adds tests for all new components

### Testing
- [X] Tested locally?
- [X] Added tests?

### Process

## Linked Issues
- First PR for #7730 

## To Be Done
- #issue

    - new endpoint A
    - new endpoint B

### DIFF
--- a/frontend-react/src/components/USLink.test.tsx
+++ b/frontend-react/src/components/USLink.test.tsx
@@ -54,4 +54,7 @@ describe("USNavLink", () => {
         const link = screen.getByRole("link");
         expect(link).toHaveClass("usa-nav__link");
     });
+    // Not sure that we have a way, nor a need, to test the activeClassName
+    // value. It's assumed that react-router-dom is testing that as a part of
+    // their package.
 });
--- a/frontend-react/src/components/USLink.test.tsx
+++ b/frontend-react/src/components/USLink.test.tsx
@@ -1,0 +1,57 @@
+import { render, screen } from "@testing-library/react";
+import { BrowserRouter } from "react-router-dom";
+
+import { renderWithRouter } from "../utils/CustomRenderUtils";
+
+import { USCrumbLink, USExtLink, USLink, USNavLink } from "./USLink";
+
+describe("USLink", () => {
+    test("renders without error", () => {
+        renderWithRouter(<USLink href={"/some/url"}>My Link</USLink>);
+        const link = screen.getByRole("link");
+        expect(link).toHaveClass("usa-link");
+        expect(link).toHaveTextContent("My Link");
+    });
+    test("renders with additional className values", () => {
+        renderWithRouter(
+            <USLink href={"/some/url"} className={"my-custom-class"}>
+                My Link
+            </USLink>
+        );
+        const link = screen.getByRole("link");
+        expect(link).toHaveClass("usa-link");
+        expect(link).toHaveClass("my-custom-class");
+    });
+    /** Specialization of USLink */
+    describe("USExtLink", () => {
+        test("renders with external link class", () => {
+            renderWithRouter(<USExtLink href={"/some/url"}>My Link</USExtLink>);
+            const link = screen.getByRole("link");
+            expect(link).toHaveClass("usa-link");
+            expect(link).toHaveClass("usa-link--external");
+        });
+    });
+    /** Specialization of USLink */
+    describe("USCrumbLink", () => {
+        test("renders with breadcrumb link class", () => {
+            renderWithRouter(
+                <USCrumbLink href={"/some/url"}>My Link</USCrumbLink>
+            );
+            const link = screen.getByRole("link");
+            expect(link).toHaveClass("usa-link");
+            expect(link).toHaveClass("usa-breadcrumb__link");
+        });
+    });
+});
+/** Specialization of NavLink from react-router-dom */
+describe("USNavLink", () => {
+    test("renders without error", () => {
+        render(
+            <BrowserRouter>
+                <USNavLink href={"/some/url"}>Navigation Link</USNavLink>
+            </BrowserRouter>
+        );
+        const link = screen.getByRole("link");
+        expect(link).toHaveClass("usa-nav__link");
+    });
+});
--- a/frontend-react/src/components/USLink.tsx
+++ b/frontend-react/src/components/USLink.tsx
@@ -1,5 +1,5 @@
 import React, { PropsWithChildren } from "react";
-import { NavLink } from "react-router-dom";
+import { Link, NavLink } from "react-router-dom";
 
 // Known issue with the `PropsWithChildren generic in React 18,
 // so I wrote this in a way that we can just remove `<any>` and be
@@ -13,9 +13,9 @@ type USLinkProps = Omit<USNavLinkProps, "activeClassName">;
 
 /** A single link for rendering standard links */
 export const USLink = ({ children, href, className }: USLinkProps) => (
-    <a href={href} className={`usa-link ${className}`}>
+    <Link to={href} className={`usa-link ${className}`}>
         {children}
-    </a>
+    </Link>
 );
 
 /** A single link for rendering external links */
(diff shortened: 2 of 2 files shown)


penny-lischer (approved):
LGTM

frontend-react/src/components/USLink.tsx
```diff
@@ -0,0 +1,54 @@
...
+// okay if we update our React version.
+interface USNavLinkProps extends PropsWithChildren<any> {
+    href: string;
+    className?: string;
+    activeClassName?: string;
+}
+type USLinkProps = Omit<USNavLinkProps, "activeClassName">;
+
+/** A single link for rendering standard links */
+export const USLink = ({ children, href, className }: USLinkProps) => (
```
stephenkao, line 15: Just a thought: for generic components like this, would it be useful if we have a single component with an enum-able `variant` (or whatever we could call it) prop as opposed to multiple components?  I feel as though it might get a little hard to maintain if we have a separate component for every state 🤔

frontend-react/src/components/USLink.tsx
```diff
@@ -0,0 +1,54 @@
+import React, { PropsWithChildren } from "react";
+import { Link, NavLink } from "react-router-dom";
+
+// Known issue with the `PropsWithChildren generic in React 18,
```
stephenkao, line 4: This is still an issue if we use function declarations for components?  We also might not even need to extend `PropsWithChildren` in the first place if we just declare `children` as a prop on its own 🤷 

```jsx
interface USNavLinkProps {
    href: string;
    className?: string;
    activeClassName?: string;
    children?: React.ReactNode
}

export function USLink({ children, href, className }: USNavLinkProps) {
  return (...);
}
```

kevinhaube: PropsWithChildren is the preferred way to extend props with React so they manage what "child" means and we remain current; the ideal method is packing our interface custom shape into the generic, but I did it like this in case we moved to React 18 before they patched that bug. 😩

kevinhaube: But so far as function declaration vs jsx vs text... ReactNode covers it all so PropsWithChildren should be fine here in that regard

frontend-react/src/components/USLink.tsx
kevinhaube: I think I originally did do something like that, but I think React bloggers are getting to me. I was reading the React docs "Inheritance vs Composition", and there was a section about "Specializations" that used this pattern and I liked it :)

https://reactjs.org/docs/composition-vs-inheritance.html#specialization

kevinhaube: I'll put you on the actual refactor PR and we can see how things ended up and whether we like it or not. Should have the PR ready later tonight! I also changed some of the component code to use more robust types for stuff like "noreferrer noopener"

frontend-react/src/components/USLink.tsx
stephenkao: Right, but I'm wondering why we even need to use `PropsWithChildren` in the first place since it's just a union between the provided generic type and an object with an optional `ReactNode` for `children`: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L800.  It feels like listing `children` directly in our prop type/interface would be clearer/less circuitous.

frontend-react/src/components/USLink.tsx
stephenkao: Sounds good to me!  I'm still a bit wary about having a separate class for every visual state since right now we're basically just changing the `className` prop (the `to` versus `href` could be abstracted away by checking against react-router context).  My main concern is the maintenance of these separate components for every state -- making a global change (e.g., adding a new prop) for the link components would require us to update each one.  

(There's also the potential for byte bloat in the resultant bundle since each of these would get transpiled into its own `createElement` with its own JSX, but that's significantly less important than longterm maintenance in my mind!)

frontend-react/src/components/USLink.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+            expect(link).toHaveClass("usa-link");
+            expect(link).toHaveClass("usa-breadcrumb__link");
+        });
+    });
+});
+/** Specialization of NavLink from react-router-dom */
+describe("USNavLink", () => {
+    test("renders without error", () => {
+        render(
+            <BrowserRouter>
```
stephenkao, line 50: Out of curiosity, why does this need an explicit `BrowserRouter` instead of using the `renderWithRouter` utility?

frontend-react/src/components/USLink.tsx
kevinhaube: Ah, yes! The updated branch (checkout experience/7730/whatever-i-named-the-new-one) and take a peek at that USLink.tsx file :) Solves the concern by relying again on React's definition of AnchorHTMLAttributes and everything should be more centralized in USLink as it's the only one adapting props to a Link component from react-router-dom

frontend-react/src/components/USLink.tsx
kevinhaube: Yeah, it would reduce the need for import and such, but my mind goes to the points of change; if React changes everywhere we used it changes too, but we lose that if we define it ourselves. 

Let's generalize this and consider it in norming! Something like "when do we rely on our packages types vs. defining our own, and why?" 🙂 I think "why" is the sauce here; I'm sure there's stuff we aren't thinking of.

frontend-react/src/components/USLink.tsx
stephenkao: Hmmmm my feeling here is that this particular use case still doesn't need more than one component with different variants because all of this is slightly different output with the same underlying responsibilities:
1. Determine if rendering within a router context
2. Determine if rendering an absolute url
3. Apply a different `className` prop based on the variant
4. Apply different attributes based on the variant (if applicable)
5. Forward native HTML attributes (`target`, `rel`, et cetera) to the element (like even though it doesn't make a lot of sense in most cases, Link/NavLink can also open new tabs so who are we to restrict the consuming component's needs :P)

Putting it in Storybook terms, these would just be different stories of the same component rather than stories of different components 🤓

But aside from this, I thiiiiiiiink there's an issue with the code in the new branch since Link/NavLink would end up like `<a href="https://reportstream.cdc.gov/https://www.google.com">whatever</a>` when it's rendered if passed an absolute url.  I'm thinking we should handle that properly under the hood so we can use the same component(s) all over the place.

kevinhaube: >Determine if rendering within a router context
>Determine if rendering an absolute url

This decision is much easier made by a human, and the less complex logic packed into a component, the lest testing required. Declaring `USExtLink` and `USNavLink` means we can declare their necessary styles, props, and base component rather than infer them from a `variant`.

>Apply a different className prop based on the variant
>Apply different attributes based on the variant (if applicable)

This, again, could be handled by `variant` parsing logic but why? Putting it into a new declaration that proxies the base `USLink` and pre-applies the right styles and props is less dynamic logic to test and still maintains the practice of reusing code at the `USLink` level. It's also a lot easier read than parsing through a `variant` switch statement, then seeing what else that may affect.

>Forward native HTML attributes (target, rel, et cetera) to the element

This is `USLink`'s purpose; each of the specializations (sans `USNavLink` as it relies on a different base for extended Router functionality like `activeClassName`) declaratively applies the necessary special attributes, and pipes the rest down to the base component to also be applied. 

---

`Link` was an issue, so I've changed it back to an `a` tag, but due to its use case, I **don't** think we should be considering any external `NavLink`s. The only true added utility that we have with it right now is using its `activeClassName` prop and inner logic to give `ReportStreamHeader` navigation items the proper active styles (which, as a side effect, this refactor will actually _fix_ considering we don't use the active styles right now like we should).

PR #7793, opened 2022-12-29
Kotlin Data Types to Typescript Codegen Plugin

jpandersen87:
This PR adds a new kotlin project to the repo that creates a gradle plugin for exporting kotlin data types to typescript. This plugin will take into account classes/fields/methods annotated with JsonIgnore/JsonIgnoreType. It adds this plugin to the prime-router project with some initial pending classes selected for export.  Typescript definitions are written to frontend-react/typings/api-codegen.d.ts. The plugin supplies a new gradle task called generateTypescriptDefinitions that is set to run as a dependency of the classes lifecycle task. The expected workflow would be that any changes to kotlin data structures that are being exported to the frontend will be automatically converted to typescript when rebuilt. As those types begin to be used on the frontend, any breaking changes to those structures will cause the frontend to fail to build.

Also includes minor changes to better enforce prime-router's kotlin version to transitive packages and eager task mitigation.

Test Steps:
1. Find a suitable kotlin data class that is used for the azure API and either apply the @TsExport annotation to it, or manually add its fully qualified name to the manualClass list in  prime-router's build.gradle.kts
2. Run the build task
3. Verify that the generateTypescriptDefinitions task runs in the console output
4. Verify that the data class and all related types have been added to the api-codegen.d.ts file

## Changes
- tsGeneratePlugin project added
- prime-router includes tsGeneratePlugin via includeBuild in settings.gradle.kts
- tsGeneratePlugin configured for usage on prime-router
- some initial kotlin classes have been configured for export

### Testing
- [x] Tested locally?
- [x] Added tests?

### Process


prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirPathUtils.kt
```diff
@@ -104,11 +104,12 @@ object FhirPathUtils : Logging {
             }
         } catch (e: Exception) {
             // This is due to a bug in at least the extension() function
-            val msg =  when (e) {
+            val msg = when (e) {
                 is FHIRLexerException -> "Syntax error in FHIR Path expression $expression"
                 is SchemaException -> e.message ?: "Condition error in FHIR Path expression $expression"
-                else -> "Unknown error while evaluating FHIR Path expression $expression for condition. " +
-                    "Setting value of condition to false."
+                else ->
```
kevinhaube, line 110: Try merging `master` in; I believe this is what I saw yesterday and put the PR up for

frontend-react/src/typings/api-codegen.d.ts
```diff
@@ -0,0 +1,150 @@
...
+    jurisdictionalFilter: string[] | undefined;
+    processingModeFilter: string[] | undefined;
+    qualityFilter: string[] | undefined;
+    routingFilter: string[] | undefined;
+    topic: Topic;
+}
+
+type Jurisdiction = "FEDERAL" | "STATE" | "COUNTY";
+
+interface Organization {
```
kevinhaube, line 13: I was Googling to try and figure out the benefit of `.d.ts` over exporting a type/interface from a `.ts` file. Here's what I found:

>If you define a file with an exported interface (that will not output any js), you should do write it in a .d.ts file. In that way, it will be excluded by the compiler when generating modules.

Is my understanding correct in saying "if we move our types/interfaces to type declaration files rather than exporting them, we can cut our static build size down"?

frontend-react/package.json
```diff
@@ -82,7 +82,7 @@
         "eject": "react-scripts eject",
         "lint": "npm-run-all -p lint:eslint",
         "lint:write": "npm-run-all -p lint:eslint:write",
-        "lint:eslint": "eslint \"src/**/*.{js,ts,jsx,tsx}\"",
+        "lint:eslint": "eslint \"src/**/*.{js,ts,jsx,tsx}\" --ignore-pattern \"src/typings/api-codegen.d.ts\"",
```
kevinhaube, line 85: Do we need this _and_ the line down on `196`? And was there a specific reason we didn't want to lint the generated code?

jpandersen87: Yes, the lint:eslint command doesn't take the ignore rule into account on 196 (the IDE does though I think).

frontend-react/src/typings/api-codegen.d.ts
jpandersen87: All typescript types/interfaces (except enum I think) go away on build. If your file is purely typescript typing then it's preferred to have it as a d.ts file.

frontend-react/package.json
kevinhaube: Ahhh, interesting 🤔 I wonder why that is. Good catch!

frontend-react/src/typings/api-codegen.d.ts
jpandersen87: I was wrong in this scenario. d.ts files are intended only as type declarations to corresponding .js files. So in this situation, the codegen should be creating .ts files. Will fix.

arnejduranovic (commented):
Is it possible to include tsGeneratePlugin as a third-party dependency via gradle or something? I don't really like adding a new library that we have to maintain and manually update directly in the RS project...

jpandersen87:
@arnejduranovic I can't seem to reply to your comment. Do you mean break the plugin off into its own CDCgov repo and pull from that?

frontend-react/package.json
stephenkao: I think the issue is that the file that's generated is at `src/typings/api-codgen.ts`.  If we change line 196 to `"src/typings/api-codegen.ts"` (note the `src/` path as well), we shooooooould be able to get rid of this ignore 🤔

frontend-react/src/typings/api-codegen.ts
```diff
@@ -0,0 +1,150 @@
...
+interface Organization {
+    countyName: string | undefined;
+    description: string;
+    filters: ReportStreamFilters[] | undefined;
+    jurisdiction: Jurisdiction;
+    name: string;
+    stateCode: string | undefined;
+}
+
+interface TemporalAccessor {
```
stephenkao, line 22: Huhhhh interesting that these Java interfaces are exposed here as empty objects.  I guess it makes sense for proper inheritance, but it just feels odd 😆

prime-router/build.gradle.kts
```diff
@@ -821,5 +825,30 @@ dependencies {
...
+
+typescriptGenerator.apply {
+    outputPath.set(project.projectDir.toPath().resolve("../frontend-react/src/typings/api-codegen.ts"))
+    classPath.set(layout.files(project.sourceSets.main.get().runtimeClasspath))
+    annotation.set(
+        gov.cdc.prime.tsGenerateLibrary.TsExportAnnotationConfig(
+            "gov.cdc.prime.router"
+        )
+    )
+    manualClasses.set(
```
stephenkao, line 840: Oh, nice nice nice

arnejduranovic:
> @arnejduranovic I can't seem to reply to your comment. Do you mean break the plugin off into its own CDCgov repo and pull from that?

@jpandersen87 Yes, we would move the library to its own repo but also, we would need to "publish" the plugin (jar) to a, probably private, software artifact repository, like Nexus. Once the jar is published, we can pull it into our project as detailed here https://docs.gradle.org/current/userguide/plugins.html#sec:custom_plugin_repositories. Do we have any related prior art to reference here?

The main reason I advocate for this change is because this plugin does not seem to be specific to this project and can be reused elsewhere. I would like to separate the maintenance of this plugin from the RepoStream project itself. I would be interested in other opinions as well. @kevinhaube @stephenkao 

Otherwise, I think your changes look good.

stephenkao:
> > @arnejduranovic I can't seem to reply to your comment. Do you mean break the plugin off into its own CDCgov repo and pull from that?
> 
> @jpandersen87 Yes, we would move the library to its own repo but also, we would need to "publish" the plugin (jar) to a, probably private, software artifact repository, like Nexus. Once the jar is published, we can pull it into our project as detailed here https://docs.gradle.org/current/userguide/plugins.html#sec:custom_plugin_repositories. Do we have any related prior art to reference here?
> 
> The main reason I advocate for this change is because this plugin does not seem to be specific to this project and can be reused elsewhere. I would like to separate the maintenance of this plugin from the RepoStream project itself. I would be interested in other opinions as well. @kevinhaube @stephenkao
> 
> Otherwise, I think your changes look good.

I agree with @arnejduranovic about this.  It's not something that needs to be kept specifically in this repository, and breaking it out into a separate codebase should make for easier, more targeted maintenance going forward.  It'd also be great to have configuration options for things like namespaces we want to include, the destination for the generated types file, and the like.  I don't have experience with Nexus so I'll defer to him on that, though 😅 

One question I have about our intended process is: who's in charge of maintaining the generated file?  If the Pipeline or Platform team makes a structural change to our entities, are they expected to regenerate the file and commit it -- and how do we escalate if there's a front-end build failure resulting from it?

frontend-react/src/typings/api-codegen.d.ts
kevinhaube: Ahh, like in the case of retrofitting a package with a `@types/package`. That's where I'd seen `.d.ts` used for the most part; never really had a case for it in my own work so this is good to know! Thank you

kevinhaube:
>One question I have about our intended process is: who's in charge of maintaining the generated file? If the Pipeline or Platform team makes a structural change to our entities, are they expected to regenerate the file and commit it -- and how do we escalate if there's a front-end build failure resulting from it?

@stephenkao I know our pre-commit hooks (or maybe it's GHA) currently have checks for generated docs within the `prime-router` directory related to _something_. Perhaps we could add this in as a part of a pre-commit hook.

>The main reason I advocate for this change is because this plugin does not seem to be specific to this project and can be reused elsewhere. I would like to separate the maintenance of this plugin from the RepoStream project itself. I would be interested in other opinions as well.

@arnejduranovic Absolutely! As for maintenance, I'd look to USDS and the CDC on guidance for who maintains open-source packages in situations like this. I know reliance on a single entity (i.e. a single contractor like Nava, Fearless) tends to be a bad model, so I'd love to hear if they have any standards around how OSS packages keep up-to-date, or _if_ they've even considered it. If not, we might be breaking some ground here! haha :)

frontend-react/src/typings/api-codegen.ts
```diff
@@ -0,0 +1,150 @@
...
+
+interface TranslatorConfiguration extends TranslatorProperties {
+    type: string;
+}
+
+interface TransportType {
+    type: string;
+}
+
+interface Receiver {
```
kevinhaube, line 100: Noticed this during your demo in eng sync: nothing actually exports from here since we ere using `.d.ts` prior

stephenkao:
@jpandersen87 While the plan for the gradle plugin is still being figured out, I think we could create a separate pull request to commit just the new type declarations to unblock the remaining react-query work.  What do you think?

jpandersen87:
Closing this as a new repo has been made for the plugin. Will create a new PR with config/dependency updates to use the plugin when ready.

PR #7796, opened 2022-12-29
Refactor: One `Link` to rule them all

### LINKED ISSUE #2772: Add USWDS skipnav above gov website banner
** Research and implement if low effort, if high effort discuss with team **

**Describe the bug**
As a user with a screen reader, I should be able to skip over the "This is an official gov website ..." banner and accordion with an anchor link and get right to the content.

Currently exists on 11ty site under `_includes/banner.liquid`. I would have just *added it* instead of creating this bug ticket, but I'm not totally sure how the gov banner component is being pulled in now.

**Impact**
Is a USWDS recommended accessibility feature.

**To Reproduce**

**Expected behavior**
Banner markup should look like:
```
<a class="usa-skipnav" href="#main-content">Skip to main content</a>
<section class="usa-banner" aria-label="Official government website"> ... </section
```

**Screenshots**

**Logs**
If applicable, please attach logs to help describe your problem.

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

### LINKED ISSUE #7730: Create a single custom Link component to replace all usage of <a> tags and other Links within RS

kevinhaube:
This PR implements a `USLink` component system, including specialty uses like `USExtLink`, to replace all rogue `a`, `Link` and `NavLink` uses with a consolidated application of styles and security. It updates all found links, though it has its limitations:

- We cannot use relative links within React Router from markdown pages (fixed if we add `MDX` support).
- We cannot use relative links within sections of `Home` that use `dangerouslySetInnerHTML` (though, I'd argue we shouldn't be using this method _at all_).

Test Steps:
1. Search the Repository for uses of `a`, `Link`, and `NavLink`. Your results should show the only uses of those tags within the `USLink.tsx` file
2. Ensure that all Navigation links are working in our header, including pages, the org selector, and "log out"
3. Ensure all links are working from public site pages (See known issues for details)
4. Ensure all links are working from Daily Data, Submission History, Value Sets, Last Mile Failures, and Receiver Status Dashboard tables
5. Ensure all links are working within the Admin Organizations table (the admin landing page)

>**NOTE:** Some links are wrapping `button` elements; ensure you are testing buttons that indicate navigation to a new component, such as "Edit" on the Admin Organizations table, or "New" in the organization settings UI.

## Changes
- Updates uses of `a`, `Link`, and `NavLink` to `USLink` and its specializations.
- Adds "Skip Nav" link 

Site link
Screen Shot 2022-12-29 at 4 52 37 PM

External link
Screen Shot 2022-12-29 at 4 50 33 PM

Navigation link w/ active style
Screen Shot 2022-12-29 at 4 50 45 PM

Special case where buttons are links:
Screen Shot 2022-12-29 at 4 49 38 PM

### Testing
- [X] Tested locally?
- [X] Added tests?

### Process

## Linked Issues
- Fixes #7730 
- Fixes #2772 

## To Be Done
- #issue

    - new endpoint A
    - new endpoint B


frontend-react/src/components/FileHandlers/FileHandlerMessaging.tsx
```diff
@@ -151,24 +151,17 @@ export const RequestedChangesDisplay = ({
                 <h5 className="margin-bottom-1">Resources</h5>
                 <ul className={"margin-0"}>
                     <li>
-                        <NavLink
+                        <USLink
                             target="_blank"
```
kevinhaube, line 155: This one is notably _NOT_ an external link, but still opens `_blank`. Great example of the level of customization we have of the base component. The value add in something like `USExtLink` is the application of off-site link security measures and proper styling to indicate the link takes you away from ReportStream.

frontend-react/src/components/SenderModeBanner.tsx
```diff
@@ -25,9 +26,9 @@ const BannerContent = () => {
                         <div className="grid-col-fill tablet:grid-col-auto">
                             <b>Onboarding: </b> Your account is not yet sending
                             data to your public health authority.{" "}
-                            <NavLink to={path}>
+                            <USNavLink href={path}>
```
kevinhaube, line 29: Woops! Should be a `USLink`.

frontend-react/src/components/Table/ColumnData.tsx
```diff
@@ -74,16 +75,15 @@ export const ColumnData = ({
         // Render column value as NavLink
         const feature = columnConfig?.feature as LinkableColumn;
         return tableData(
-            <NavLink
-                className="usa-link"
-                to={`${feature.linkBasePath || ""}${
+            <USLink
```
kevinhaube, line 78: Great time to address `Link` and `NavLink`: from what I can find/read, the _only_ time we should utilize `NavLink` is for navigation headers such as our nav bar, a side nav display, or anything requiring the ability to display an `active` set of styles.

frontend-react/src/components/header/DropdownNav.tsx
```diff
@@ -40,7 +40,7 @@ export const DropdownNav = ({ label, root, directories }: DropdownNavProps) => {
         };
     }, []);
     const navMenu = directories.map((dir) => (
-        <NavLink to={`${dir.root}/${dir.slug}`}>{dir.title}</NavLink>
+        <USLink href={`${dir.root}/${dir.slug}`}>{dir.title}</USLink>
```
kevinhaube, line 43: Furthermore, the sub-item of a dropdown is _not_ a `NavLink` because uswds does not have an active style for a chosen dropdown navigation item.

frontend-react/src/pages/deliveries/Table/ReportLink.tsx
```diff
@@ -62,11 +62,14 @@ function ReportLink(props: Props) {
 
     if (!props.button) {
         return (
-            <a href="/" onClick={handleClick} className="usa-link">
+            <button
+                onClick={handleClick}
```
kevinhaube, line 66: As a general rule, no link should have `onClick` behavior other than navigating to a page. This will always be a case for a `button`

kevinhaube (commented):
Still fixing up some styling faux pas; few comments to help with reviewing.

frontend-react/src/components/Table/ColumnData.tsx
stephenkao: Yep, NavLink is just a souped-up Link: https://github.com/remix-run/react-router/blob/main/packages/react-router-dom/index.tsx#L485-L543

It's surprisingly a lot of code to detect/implement current path activeness 😆

frontend-react/src/components/USLink.tsx
```diff
@@ -1,38 +1,116 @@
...
+    Omit<CustomLinkProps, "activeClassName">;
+type USNavLinkProps = Pick<AnchorHTMLAttributes<{}>, "href"> & CustomLinkProps;
 
-/** A single link for rendering standard links */
-export const USLink = ({ children, href, className }: USLinkProps) => (
-    <Link to={href} className={`usa-link ${className}`}>
-        {children}
-    </Link>
-);
+/** Helper class to protect against the string literal "undefined" appearing in styles
```
stephenkao, line 17: Do we need this?  We have the `classnames` package already, so we could just use that instead:

```js
const className1 = undefined;
const className2 = null;
const className3 = "style-c";
classnames("style-a style-b", className1, className2, className3); // returns "style-a style-b style-c"
```

frontend-react/src/components/USLink.tsx
```diff
@@ -1,38 +1,116 @@
-import React, { PropsWithChildren } from "react";
+import React, { AnchorHTMLAttributes } from "react";
 import { Link, NavLink } from "react-router-dom";
 
 // Known issue with the `PropsWithChildren generic in React 18,
```
stephenkao, line 4: Is this comment needed anymore?

frontend-react/src/components/USLink.tsx
kevinhaube: I had no idea we even had that! Saves me time and trouble (or, I guess, _would have_ saved me) 😂 I'll update this

frontend-react/src/components/USLink.tsx
kevinhaube: Ah, not _this_ wording, but I did mean to change it to reflect why we aren't using the `PropsWithChildren` type in general. Also not sure why I added `React.ReactNode[]` -- can't think of a time we'd need more than one element inside the link tag.

frontend-react/src/components/FileHandlers/FileHandlerMessaging.tsx
stephenkao: Very small small point: I agree that we don't need `noreferrer` here since it's an internal link, but my feeling is that we should still add `rel="noopener"` whenever we have `target="_blank"` since that makes it so the new tab doesn't share the render process with the original tab.  It's not usually something we really have to worry about since the site isn't intensive on the client side, but I figure why not split off the processes anyway 🤷

frontend-react/src/components/FileHandlers/FileHandlerMessaging.tsx
```diff
@@ -151,24 +151,17 @@ export const RequestedChangesDisplay = ({
...
                     <li>
-                        <a
-                            className={"usa-link--external"}
-                            target={"_blank"}
-                            href={
-                                "https://www.cdc.gov/csels/dls/sars-cov-2-livd-codes.html"
-                            }
-                            rel="noreferrer"
-                        >
+                        <USExtLink href="https://www.cdc.gov/csels/dls/sars-cov-2-livd-codes.html">
```
stephenkao, line 162: May be out of the scope of this pull request, but can we add a URL constants file for all the external links?  We could also do that for internal links for programmatic routing, but that's less straightforward.

frontend-react/src/components/SenderModeBanner.tsx
```diff
@@ -1,19 +1,19 @@
...
+import { USLink } from "./USLink";
+
 const isNotActive = (val: string | undefined): boolean => {
     return val === "testing" || val === "inactive";
 };
 
 const BannerContent = () => {
     const { senderDetail: sender, senderIsLoading: loading } =
         useSenderResource();
-    const path = "/getting-started/testing-facilities/overview";
```
stephenkao, line 16: Seems like an unnecessary variable?  Glad we're simplifying it! 😆

frontend-react/src/components/USLink.tsx
```diff
@@ -1,38 +1,103 @@
...
-    </Link>
-);
+/** A single link for rendering standard links. Uses a `Link` by default
+ * but adding `anchor` will make this a generic anchor tag.
+ * @example
+ * <USLink href="/page">To Page</USLink> // uses <Link> from react-router-dom
+ * <USLink anchor href="#this-section-on-my-page">To Section</USLink> // uses <a>
+ * */
+export const USLink = ({
+    anchor = false,
```
stephenkao, line 25: I'm wondering if we can infer whether to use an `<a>` versus `<Link>` if 1) the component is rendered within React Router context or 2) the `href` starts with `http(s)://`, `#`, `mailto:`, `tel:` (or whatever other protocol we might want to use) 🤔 

Just a thought, though.  I don't think it hurts to have this override prop!

frontend-react/src/components/USLink.tsx
```diff
@@ -1,38 +1,103 @@
...
+    children,
+    className,
+    href,
```
stephenkao, line 28: VERY minor point, but I'm a little wary of calling this prop `href` because it's not technically a 1:1 with the underlying `href` HTML attribute for Links.  If we were to add `basename`s to our routing, something like `href="/whatever"` could ultimately turn into `href="/base/name/whatever"` when rendered.

frontend-react/src/components/USLink.tsx
```diff
@@ -1,38 +1,103 @@
...
+    ...anchorHTMLAttributes
+}: USLinkProps) => {
+    return !anchor ? (
```
stephenkao, line 31: Small thing, but to reduce a bit of redundancy/future maintenance, we could just toggle the tag:

```jsx
const Tag = anchor ? 'a' : Link;

return (
  <Tag to={ .... }>{ children }</Tag>  
);
```

frontend-react/src/components/USLink.tsx
```diff
@@ -41,11 +106,11 @@ export const USNavLink = ({
 }: USNavLinkProps) => {
     return (
         <NavLink
-            to={href}
+            to={href || ""}
             className={({ isActive }) =>
                 isActive
-                    ? `usa-nav__link usa-current ${activeClassName || ""}`
-                    : `usa-nav__link ${className || ""}`
+                    ? classnames("usa-nav__link usa-current", activeClassName)
```
stephenkao, line 112: Another benefit for `classnames` is that you can pass an object to toggle class names:

```jsx
className={ classnames("usa-nav__link", { "usa-current": isActive, [activeClassName]: isActive }) }
```

frontend-react/src/components/header/ReportStreamHeader.tsx
```diff
@@ -6,54 +6,36 @@ import {
...
 import { AdminDropdown } from "./DropdownNav";
 
 const { IS_PREVIEW, CLIENT_ENV } = config;
 
 const ProductIA = () => (
-    <NavLink
-        to="/product"
-        key="product"
-        data-attribute="hidden"
-        hidden={true}
```
stephenkao, line 28: Huh!  I'm guessing there was no need for this since it's getting removed now, but do you know what the intention of this prop was?  I guess it was just accompanying the `data-attribute` but not really doing anything 😬

frontend-react/src/components/header/SignInOrUser.tsx
```diff
@@ -24,23 +24,24 @@ const SignInOrUser = () => {
         <div className="prime-user-account">
             <span id="emailUser">{user}</span>
             <br />
-            <a
-                href="/"
+            {/* Okta handles our logoutRedirectUri, this should _not_ be an anchor tag! */}
+            <button
```
stephenkao, line 28: Can we use Trussworks' Button component?  It looks like they have an `unstyled` prop for this: https://trussworks.github.io/react-uswds/?path=/story/components-button--unstyled

Also, let's add a `type="button"` since the default is `type="submit"`, which would cause some weirdness if this were nested in a form for whatever reason :P

frontend-react/src/components/header/SignInOrUser.tsx
```diff
@@ -24,23 +24,24 @@ const SignInOrUser = () => {
...
-                className="usa-link"
+                className="usa-button usa-button--unstyled"
             >
                 Log out
-            </a>
+            </button>
         </div>
     ) : (
-        <NavLink to="/login">
+        <USLink href="/login">
```
stephenkao, line 39: I realize this existed beforehand so it's out of the scope of this changeset, but it's invalid html to have a `<button>` inside an `<a>`.  Could we turn that into a navigation in an `onClick`?

stephenkao:
It looks like `react-markdown` has a way for us to specify custom components for the markdown equivalent: https://remarkjs.github.io/react-markdown/#components

Maybe we could use that instead of waiting for MDX?  (Though to be completely transparent, I'm not a huge fan of MDX since it requires more transformation than it's worth in my opinion 😆)

frontend-react/src/components/USLink.tsx
kevinhaube: YES! I was trying to Google how to do this because I'd seen it before in some React code but couldn't find anyone answering it. Will try to implement it this way

frontend-react/src/components/header/ReportStreamHeader.tsx
kevinhaube: I, for the life of me, could not find a use. Nor was it actually _hidden_ in the UI. 🤷🏼 I figured it was copy/paste from the startup days.

frontend-react/src/components/header/SignInOrUser.tsx
kevinhaube: We could, but this ties back to legacy concerns of how unresponsive Trussworks is in that repo. 😬 Think this ties into the research and discussion around component consolidation - a question of can we trust that library.

frontend-react/src/components/header/SignInOrUser.tsx
kevinhaube: Yep! I believe we have a few of these in here that were quick jobs in the refactor that need to be revisited. I don't mind going back and just nixing them all for an `onClick(() => navigate("..."))` in the button itself. There was also some dodgy CSS overrides I had to write to get styling right because of the bad nesting. Would love to rip it out and never let it see the light of `master`

frontend-react/src/components/header/SignInOrUser.tsx
stephenkao: Right, but we're already using Trussworks' component system in our production code.  I agree that we still need to have that discussion down the line (I'm curious as to what specific issues there are and why we can't implement workaround components like USLink when their components fail us), but I'd argue that a migration down the line would be easier if we have uniformity to start with -- even if that means that we continue using their components in the meantime.

frontend-react/src/components/FileHandlers/FileHandlerMessaging.tsx
kevinhaube: What kind of constraints? I do plan on taking the DOM Purify bit up to the component level as cleansing external links seems to be a good practice we follow on external links. This link, for example, would have no purification on it as of now.

frontend-react/src/components/SenderModeBanner.tsx
kevinhaube: I mean, hey...sometimes you just gotta get those ugly strings in line 😂 but yeah, felt redundant.

frontend-react/src/components/USLink.tsx
kevinhaube: Hmm, great point about the inference! Unless it's hidden better than Waldo himself, we don't use `tel:`. I think it's safe to condense it to just `starts with "http" || "#" || "mailto:"`. That said, in the spirit of no pushes needed for this kinda thing, is there a definitive list of protocols we'd want to make an enumeration of for this detection? Don't see why we shouldn't do it all in one swift commit. :)

frontend-react/src/components/header/SignInOrUser.tsx
kevinhaube: Yeah, at a glance, looks like we aren't using nearly as many `<button>`s (notably, on `master` only found 2). I'll change this over; do you mind ticketing the conversion of the other 2 so we can get someone on those and get that uniformity? Saves a bit of trouble on that research ticket :)

frontend-react/src/components/FileHandlers/FileHandlerMessaging.tsx
stephenkao: Ah constants, not constraints :) I just mean something simple like:

```js
// in src/utils/urls.ts -- or wherever!
// also I'm terrible at naming variables 
export const CDC_COVID_LIVD_CODES_URL = "https://www.cdc.gov/csels/dls/sars-cov-2-livd-codes.html";
```

That way we can keep track of all external URLs in one place for easier maintenance.  In my opinion, internal paths can be handled separately since they might have wildcards and interpolation, but down to discuss this too!

frontend-react/src/components/USLink.tsx
stephenkao: `href`s can basically take any URI scheme so it's unfortunately a pretty big list 😬 I'd say that the most popular ones that we'd likely use are the ones we've already discussed, though.  Plus a `javascript:` one just in case (even though that's generally not great practice!).

So anything that starts with:
- `http(s)://`
- `mailto:`
- `tel:` (I'm thinking we might as well add it just in case 🤷)
- `javascript:`
- `#`

But again, definitely not a blocking point!  Just some food for thought :)

frontend-react/src/components/header/ReportStreamHeader.tsx
stephenkao: Yeah, I have no idea 😆 I know that this can be used for inputs for bot honeypots, but as far as I know `hidden` isn't a valid prop for NavLink so I have no idea why it was applied.  Either way, good to get rid of it!

frontend-react/src/components/header/SignInOrUser.tsx
stephenkao: Sure, I'll add a ticket now and link it to the consolidation ticket!  Thanks!

frontend-react/src/components/header/SignInOrUser.tsx
stephenkao: No problem!  As far as I know, modern browsers have become a lot more lenient so nothing like this explicitly breaks the UX.  It's just not great for accessibility since an element that can be actioned shouldn't be nested into _another_ element that can be actioned.

I can add a ticket for this, and we can come back to it later to keep your changeset specific to the replacement of links! :)

stephenkao: Feel free to update!  https://app.zenhub.com/workspaces/experience-607d9d5e68b95200150fec37/issues/gh/cdcgov/prime-reportstream/7838

frontend-react/src/components/USLink.tsx
kevinhaube: I'll go ahead and ticket/link this one as well as a good follow-up ticket :)

frontend-react/src/components/FileHandlers/FileHandlerMessaging.tsx
kevinhaube: #7840

frontend-react/src/components/USLink.tsx
kevinhaube: So, in the screenshot below, I tried the above implementation and we get an issue with testing where the tag rendered is _not_ recognized as a link if it's an `anchor` tag. That gives me some concern for accessibility even though it's clear that we're rendering it as `<a>` in the DOM.

Screen Shot 2023-01-04 at 11 04 50 AM

And then I realized, due to the `to` vs `href` difference between `Link` and `a`, we end up rendering links that don't link anywhere. Are you aware of a way to conditionally name that prop? If not, I think the current version might be our best bet.

Screen Shot 2023-01-04 at 11 10 41 AM

kevinhaube:
@stephenkao 
>It looks like react-markdown has a way for us to specify custom components for the markdown equivalent: https://remarkjs.github.io/react-markdown/#components

Yes! Love this! Though, with the knowledge that we're going to be using constants for external links in #7840 and parsing of said URLs in the `USLink` component in #7841, I think we should hold off so our logic is easier to configure for the markdown plugin. Right now, we'd need to parse it all out inside the components block:

```typescript/jsx
<ReactMarkdown
  components={{
    a: (props) => { /* Some logic crammed in to parse all this stuff */ }
  }}
>
```

frontend-react/src/components/USLink.tsx
stephenkao: Nah, this was a pedantic point about clarity of naming -- I was just wary of making it seem like the `href` prop would be guaranteed to be 1:1 with the rendered `href` data attribute, which isn't necessarily correct in the router context:

```jsx
<Router basename="/blah">
    <USLink href="/whatever">Whatever</USLink>
</Router>
```

The anchor that's actually rendered is `<a href="/blah/whatever">Whatever</a>` as opposed to `<a href="/whatever">Whatever</a>` so the `href` prop could be a little misleading since it's not exactly the resultant `href` attribute value.  Definitely not a blocking point, though!

frontend-react/src/components/USLink.tsx
stephenkao: Hehehe oops, duh!  I think we could do something like this, though:

```jsx
const props = {
    ...otherProps,
    [anchor ? 'a' : 'to']: href
};

return <Tag { ...props } />;
```

Not a big deal, but I think it helps to cut down redundancy and potential for typos if we add more props 🤷

stephenkao:
> @stephenkao
> 
> > It looks like react-markdown has a way for us to specify custom components for the markdown equivalent: https://remarkjs.github.io/react-markdown/#components
> 
> Yes! Love this! Though, with the knowledge that we're going to be using constants for external links in #7840 and parsing of said URLs in the `USLink` component in #7841, I think we should hold off so our logic is easier to configure for the markdown plugin. Right now, we'd need to parse it all out inside the components block:
> 
> ```
> <ReactMarkdown
>   components={{
>     a: (props) => { /* Some logic crammed in to parse all this stuff */ }
>   }}
> >
> ```

Yeah, definitely agree with holding off on this -- I can add a follow-up ticket to test it out!

stephenkao (approved):
A few smalllllll/non-blocking remaining points, but this looks great to me!  Thanks for taking this on!  🙌

frontend-react/src/components/USLink.tsx
kevinhaube: I was trying to do this inline with JSX and it hated it; gonna try this! :)

kevinhaube: Bah humbug...still doesn't like it at the JSX level. Tried both inline and as the `toOrHref` var.

Screen Shot 2023-01-04 at 11 51 19 AM

frontend-react/src/components/header/SignInOrUser.tsx
kevinhaube: Oops! Forgot to resolve this; just reviewed the ticket and it looks good! Just added `needs refinement` tag

frontend-react/src/components/USLink.tsx
kevinhaube: Ahhh, got it!

PR #7805, opened 2022-12-30
7629 Added Analytics service to TrackAppInsightEvent

penny-lischer:
This PR ...

Test Steps:
1. Log into RS
2. Navigate to either Daily Data or Submission History
3. Select start and to dates and click Filter
4. App Insights should capture the event name ("Daily Data | Table Filter", "Submission History | Table Filter") along with the selected dates within the Logs or Usage Events

*If you are running locally you will need to set the following keys in your .env.development file:
REACT_APP_APPINSIGHTS_KEY
REACT_APP_APPLICATIONINSIGHTS_CONNECTION_STRING

## Changes
- Added Analytics service to TrackAppInsightEvent.
- Updated TableFilters.tsx to track event when filtering.
- Created enum to store Feature types

### Testing
- [x] Tested locally?
- [ ] Added tests?


frontend-react/src/TelemetryService.ts
```diff
@@ -6,6 +6,10 @@ import { getSessionMembershipState } from "./utils/SessionStorageTools";
 let reactPlugin: ReactPlugin | null = null;
 let appInsights: ApplicationInsights | null = null;
 
+export function getAppInsights() {
+    return appInsights;
```
kevinhaube, line 10: This should be worked into the `AppInsightsProvider`. Right now it's payload is `ai.reactPlugin` but if we provide the whole `ai` objects, we can destructure it in a component like `const { appInsights } = useAppInsightsContext()`

frontend-react/src/utils/Analytics.ts
```diff
@@ -0,0 +1,20 @@
+import { IEventTelemetry } from "@microsoft/applicationinsights-web";
+
+import { getAppInsights } from "../TelemetryService";
+
+export enum EventName {
```
kevinhaube, line 5: Could we instead rely on `FeatureType` to provide the names of features and `EventName` to provide the event name itself, and mash them together in a string? If we change `FeatureType.DAILY_DATA`, for example, we'd want to ensure we are getting the right `Feature | Event` string

frontend-react/src/AppRouter.tsx
```diff
@@ -33,6 +33,13 @@ import { AdminRevHistoryWithAuth } from "./pages/admin/AdminRevHistory";
 import { ErrorNoPage } from "./pages/error/legacy-content/ErrorNoPage";
 import { MessageDetailsWithAuth } from "./components/MessageTracker/MessageDetails";
 
+export enum FeatureType {
```
kevinhaube, line 36: Just a nit, non-blocking: to keep with the `EventName` nomenclature, could this be `FeatureName`?

kevinhaube (dismissed):
Mainly want to keep the use of insights through the provider everywhere we can, and ensure our feature events are completely enumerated

frontend-react/src/utils/Analytics.ts
penny-lischer: That was my intent initially :)

frontend-react/src/TelemetryService.ts
stephenkao: My opinion is that it's fine to keep this as-is.  `appInsights` is something that's already initialized outside of React, and we don't need to tightly couple this instance with rendering.  Plus, this approach allows us to create functions that can be invoked inside or outside of React rendering whereas destructuring it from Context will prohibit that.

frontend-react/src/utils/Analytics.ts
```diff
@@ -0,0 +1,20 @@
...
+    DAILY_DATA_TABLE_FILTER = "Daily Data | Table Filter",
+    SUBMISSIONS_TABLE_FILTER = "Submission History | Table Filter",
+}
+
+export const TrackAppInsightEvent = (
```
stephenkao, line 10: Very small point, but function names in JavaScript conventionally start with lowercase letters: `trackAppInsightEvent`.

frontend-react/src/utils/Analytics.ts
stephenkao: Should this be named `PageName` instead of `FeatureType` and split up `FeatureType.ADMIN` then?  It looks like the intention is to organize them by page, in which case 'feature' feels a little vague.

frontend-react/src/components/Table/TableFilters.tsx
```diff
@@ -54,6 +52,17 @@ function TableFilters({ filterManager, cursorManager }: SubmissionFilterProps) {
...
+
+            const eventName = featureEvent;
+            const eventData = {
+                tableFilter: {
+                    startRange: from,
+                    endRange: to,
+                },
+            };
+
+            TrackAppInsightEvent(eventName, eventData);
```
stephenkao, line 64: Since TableFilters seems intended to be a generic component that supposedly could be rendered anywhere (_possibly_ without wanting to track analytics), it might be better to turn this into an `onFilterClick` prop:

```jsx
function TableFilters({ ...otherStuff, onFilterClick }) {
    // other stuff

    const applyToFilterManager = () => {
        updateRange();

        // call onFilterClick with the specified range
        onFilterClick({ from, to })
    };
}
```

Then we specify the tracking event in the parent components wherever we want to track:
```jsx
// in DeliveriesTable.tsx
<TableFilters
    ...otherStuff
    onFilterClick={ ({ from, to }) => TrackAppInsightEventName(FeatureType.DAILY_DATA, { tableFilter: { startRange: from, endRange: to } }) }
/>

// in SubmissionTable.tsx
<TableFilters
    ...otherStuff
    onFilterClick={ ({ from, to }) => TrackAppInsightEventName(FeatureType.SUBMISSIONS, { tableFilter: { startRange: from, endRange: to } }) }
/>
```

If we wanted to consolidate this into a utility function, we could also do that:

```jsx
function trackFilterClickEvent(eventName, { from, to })) {
    TrackAppInsightEventName(eventName, { tableFilter: { startRange: from, endRange: to } });
}
```

It's a small thing right now since TableFilters isn't used a lot of places, but this sort of abstraction would help to keep the generic TableFilters less coupled to specific use cases and more flexible.

penny-lischer: Sounds good!

frontend-react/src/utils/Analytics.ts
stephenkao: Penny and I chatted about this yesterday, but for posterity:

My thought here was that features can exist across pages so having the insight into "feature A being used on page 1" versus "feature A being used on page 2" might be helpful.  But this isn't a blocking point in my mind, and we can address it in the future if we ever need to!

frontend-react/src/TelemetryService.ts
kevinhaube: Hm 🤔 good point, we wouldn't have to pipe those values from components into external functions. That considered, the antithetical question: do we even need the provider?? Might be something leftover from setup instructions that we don't utilize. My only concern is that I think using the hooks of the React AppInsights library might rely on the provider, but I haven't had time to look into that specific concern.

stephenkao (changes requested):
Just holding out for the decoupling in [this comment](https://github.com/CDCgov/prime-reportstream/pull/7805#discussion_r1060939687)!

kevinhaube (commented):
Good stuff! Thanks for addressing the comments

frontend-react/src/TelemetryService.ts
stephenkao: My feeling about this sort of stuff is if we can extricate logic easily outside of rendering, we should do it.  There's no reason to tightly couple something that exists entirely outside the render cycle to React, and doing so would unnecessarily limit our ability to track analytics events outside of the Provider Context.

The way that the custom track event function is set up is simpler than using a hook too:
```jsx
// as-is, just call the track function
TrackAppInsightEvent(...);

// with hooks
const trackFn = useTrackEvent(...);
trackFn(...);
```

As it is right now, we also don't have to worry about the track function reference changing between renders since it's defined once for the entirety of the application lifespan.  The reference for the return function from `useTrackEvent` could potentially change, though to be frank I haven't personally confirmed that.

Ultimately, I don't think it hurts to keep the AppInsightsProvider in the render tree, but it doesn't really _provide_ (:trollface:) much value in my opinion.

frontend-react/.env.development
```diff
@@ -18,3 +18,6 @@ REACT_APP_ENV=development
 INLINE_RUNTIME_CHUNK=false
 ; this will generate noise when packages have bad scss source maps
 GENERATE_SOURCEMAP=false
+
+REACT_APP_APPINSIGHTS_KEY=7ce107ea-c24a-49e2-a05f-2aaeb2e82f29
```
stephenkao, line 22: I filed a DevOps ticket for a dedicated dev environment for testing analytics stuff locally, but it's admittedly pretty low priority 😬  Mind removing this for the time being?

frontend-react/src/components/Table/TableFilters.tsx
```diff
@@ -18,23 +18,19 @@ export enum StyleClass {
...
-
 interface SubmissionFilterProps {
     filterManager: FilterManager;
     cursorManager?: CursorManager;
+    onFilterClick?: any;
 }
 
 /* This helper ensures start range values are inclusive
  * of the day set in the date picker. */
 const inclusiveDateString = (originalDate: string) => {
```
stephenkao, line 29: Let's export this function and add a couple of tests! 🙏

frontend-react/src/components/Table/TableFilters.tsx
```diff
@@ -18,23 +18,19 @@ export enum StyleClass {
...
-    END_RANGE = "end-range",
-    CURSOR = "cursor",
-    SORT_ORDER = "sort-order",
-    PAGE_SIZE = "page-size",
-}
-
 interface SubmissionFilterProps {
     filterManager: FilterManager;
     cursorManager?: CursorManager;
+    onFilterClick?: any;
```
stephenkao, line 24: Let's try not to add any `any`s in our types and type this function:

```ts
// untested!
onFilterClick?: ({ from: string, to: string }) => void
```

frontend-react/src/components/Table/Table.test.tsx
```diff
@@ -269,7 +269,7 @@ describe("Table, filter integration tests", () => {
             defaultCursor
         );
         // Checking for inclusive date
-        expect(screen.getByText(/cursor:/)).toHaveTextContent(/23:59:59.000Z/);
```
stephenkao, line 272: Is the difference here going to be an issue anywhere?

frontend-react/src/components/Table/TableFilters.tsx
stephenkao: We could also add a custom type for the parameter structure here!  Might be overkill, but up to you!

frontend-react/src/pages/submissions/SubmissionTable.tsx
```diff
@@ -37,6 +39,7 @@ const SubmissionTableContent: React.FC<SubmissionTableContentProps> = ({
     const transformDate = (s: string) => {
         return new Date(s).toLocaleString();
     };
+    const featureEvent = `${FeatureName.SUBMISSIONS} | ${EventName.TABLE_FILTER}`;
```
stephenkao, line 42: Same small, non-blockign point here about hoisting up this variable outside of the component!

frontend-react/src/pages/submissions/SubmissionTable.tsx
```diff
@@ -37,6 +39,7 @@ const SubmissionTableContent: React.FC<SubmissionTableContentProps> = ({
     const transformDate = (s: string) => {
```
stephenkao, line 39: Oh, also!  Unrelated and I realize it existed beforehand so we can make this change in a separate pull request, but this function could be moved out since it's getting recreated on every render.

frontend-react/src/utils/Analytics.ts
```diff
@@ -0,0 +1,19 @@
...
+export const trackAppInsightEvent = (
+    eventName: string = "",
+    eventData: any
+) => {
+    const appInsights = getAppInsights();
+    const telemetryEvent: IEventTelemetry = {
+        name: eventName,
+    };
+
+    if (eventName !== "") appInsights?.trackEvent(telemetryEvent, eventData);
```
stephenkao, line 18: If we make `eventName` a required string/remove the fallback [above](https://github.com/CDCgov/prime-reportstream/pull/7805/files#diff-6fd707ed4d605aa3f4eec2edc8899327ec74ed71587dffa749987e5a7ea591ddR10), we shouldn't need to check for an empty string.

stephenkao, line 18: Out of curiosity, is there any difference on the AI side in specifying `eventData` as `customProperties` versus `properties` in the IEventTelemetry object?

```jsx
appInsights?.trackEvent({
    name: telemetryEvent
    properties: eventData
});
```

frontend-react/.env.development
penny-lischer: Yup, was not intended to be part of the PR.

frontend-react/src/components/Table/Table.test.tsx
penny-lischer: It shouldn't be since this is what is being sent as the 'to' date in the filter to return the table results.

frontend-react/src/utils/Analytics.ts
```diff
@@ -0,0 +1,19 @@
...
+
+import { getAppInsights } from "../TelemetryService";
+
+export enum EventName {
+    TABLE_FILTER = "Table Filter",
+}
+
+export const trackAppInsightEvent = (
+    eventName: string = "",
+    eventData: any
```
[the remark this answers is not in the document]
penny-lischer: haha

frontend-react/src/pages/deliveries/Table/DeliveriesTable.tsx
```diff
@@ -68,6 +70,7 @@ const DeliveriesTableContent: React.FC<DeliveriesTableContentProps> = ({
     serviceReportsList,
 }) => {
     const { oktaToken, activeMembership } = useSessionContext();
+    const featureEvent = `${FeatureName.DAILY_DATA} | ${EventName.TABLE_FILTER}`;
```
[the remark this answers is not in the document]
penny-lischer: This causes compilations issue:
Uncaught ReferenceError: Cannot access 'FeatureName' before initialization
    at Module.FeatureName (App.tsx:111:1)
    at ./src/pages/deliveries/Table/DeliveriesTable.tsx (DeliveriesTable.tsx:28:1)

frontend-react/src/utils/Analytics.ts
penny-lischer: No difference. I'll change mine to be consistent with TelemetryService

frontend-react/src/components/Table/TableFilters.tsx
penny-lischer: This is already exported at the end of file. We currently have one test, I will ad another.

PR #7814, opened 2023-01-03
[Bug]: Site Wide: Navigating through the application buttons does not…

### LINKED ISSUE #4327: [Bug]: Site Wide: Navigating through the application buttons does not scroll you to the top of destination page
## Describe the bug
Generally within the app if you navigate to any other page it will try to keep you at whatever vertical position you had scrolled to before clicking the button- often forcing the user to constantly keep scrolling up as they use the application.

## Need to Know
This may be "us" using the React Router incorrectly.
React Router version is now v6.

## Acceptance Criteria
- [ ] When clicking one of the page interaction buttons (eg, Edit, Clone, Cancel, etc) you are navigated to the top of the subsequent page
- [ ] Verify problem is solved site-wide

## Impact
Harmless but very annoying. Could eventually lead to carpal tunnel syndrome.

## To Reproduce
Login as admin, go to Admin->Organization Settings
Choose the "waters" org from the list
Scroll down to a Receiver setting and click Edit or Clone
Observe that you are halfway down the page

## Screenshots
1. scrolled down to giang2-clone receiver setting

2. Navigated to edit page after clicking Edit on giang2-clone

jpandersen87:
… scroll you to the top of destination page

This PR adds a ScrollToTop component as a sibling to App (not parent of, so as not to cause unnecessary re-renders) that will forcefully set the window to scroll back to the top on page changes.

Test Steps:
1. Login as admin, go to Admin->Organization Settings
2. Choose the "waters" org from the list
3. Scroll down to a Receiver setting and click Edit or Clone
4. Observe that you are NOT halfway down the page

## Changes
- Created ScrollToTop component
- Added it as sibling to App

### Testing
- [x] Tested locally?
- [ ] Added tests?

### Process

## Linked Issues
- Fixes #4327


frontend-react/src/ScrollToTop.tsx
```diff
@@ -0,0 +1,18 @@
+import { useEffect } from "react";
+import { useLocation } from "react-router-dom";
+
+export const ScrollToTop = () => {
+    const { pathname } = useLocation();
+
+    useEffect(() => {
+        const canControlScrollRestoration =
+            "scrollRestoration" in window.history;
+        if (canControlScrollRestoration) {
```
kevinhaube, line 10: Could you explain this logic to a n00b? I'm not familiar with `scrollRestoration` as a value of `window.history`, and not sure where we _shouldn't_ have it scroll to top

jpandersen87: It seems to be a recent thing that browsers are implementing. See https://developer.mozilla.org/en-US/docs/Web/API/History/scrollRestoration

stephenkao: Does setting the scroll restoration mode need to be applied whenever the `pathname` changes or can it be applied on mount?

stephenkao: I haven't used scroll restoration either so I may be lacking context over its usage, but do we need it if we're explicitly calling `window.scrollTo(0, 0);` on each navigation below?

frontend-react/src/ScrollToTop.tsx
```diff
@@ -0,0 +1,18 @@
...
+            window.history.scrollRestoration = "manual";
+        }
+
+        window.scrollTo(0, 0);
+    }, [pathname]);
+
+    return null;
```
stephenkao, line 17: Hmmmm!!!  I guess since we're not actually rendering anything and we're already using a `useEffect`, we could just effectively make this a custom hook that's invoked in the App component instead of needing to "render" a component 🤔

jpandersen87: This is true. The SO version that I copied returned children but at the same time said not to use it as a parent due to re-renders which is why I null'd it. As long as the hook's returned result never changes (thus never causing a re-render in the invoker) it would make a cleaner implementation for sure.

frontend-react/src/ScrollToTop.tsx
jpandersen87: I moved the scrollRestoration portion to an useEffect init block

stephenkao (approved):
Looks good to me!

PR #7827, opened 2023-01-03
Breadcrumbs on Submissions Detail don't link back to previous URL

### LINKED ISSUE #7712: Breadcrumbs on Submissions Detail don't link back to previous URL
## Describe the bug
When using the message ID search feature for Admins, the Report ID links to `/submissions/{reportId}` to display submission details. The breadcrumb to go back, however, links to `/submissions`, the list page for submissions from an organization.

## Impact
Breaks the UI if Admins navigate using the breadcrumb.
Can be bypassed using browser "Back" button.

## To Reproduce
_Requires #7711 to merge_

1. Login at staging.reportstream.cdc.gov
2. Under the Admin dropdown in navigation, click Feature Flags
3. Add `message-tracker` to your active Feature Flags
4. Under the Admin dropdown in navigation, click Message ID Tracker
5. Search `014228`
6. The message will appear in the results, click on the Report ID (this is the part dependent on #7711's fix)
7. The url change to `/submissions/{reportId}`, when results load, observe the breadcrumb `Submissions` (hard-coded)
8. Click it
9. Submissions UI will break because of a **known issue** where `PrimeAdmins` is not a valid organization, and thus have no submissions

## known workaround

10. Replicate 4-7 again
11. Click the browser "back" button.
12. UI is okay and back to the message ID search

## Expected behavior
In step 7, we should observe `Message ID Search` if coming from that page, and `Submission History` if coming from that page.
In step 8, we should be taken to the proper previous URL

## Acceptance Criteria

- [ ] When clicking on the breadcrumb, it takes you to the expected page, not back to /submissions
- [ ] Breadcrumb link name matches correct previous page

jpandersen87:
This PR adds context to the link state for submission reports on the message tracker page and a new previousPage prop to Crumbs so that breadcrumbs can display a "Return to Message Tracker ID Search" button on top of the existing breadcrumbs for the submissions sitemap area. Minor updates were made to the Table component to support passing link state.

Test Steps:
1. Login
2. Under the Admin dropdown in navigation, click Feature Flags
3. Add message-tracker to your active Feature Flags
4. Under the Admin dropdown in navigation, click Message ID Tracker
5. Search (ex: 014228, 798189, etc.)
6. The message will appear in the results, click on the Report ID
7. Verify that the breadcrumbs shows a back button to Message Tracker on top of the breadcrumbs

## Changes
- "from" property added to link state for reports on message-tracker
- Table component and related types updated to support passing link state

### Testing
- [x] Tested locally?
- [ ] Added tests?

### Process

## Linked Issues
- Fixes #7712


frontend-react/src/pages/submissions/SubmissionDetails.tsx
```diff
@@ -127,8 +127,16 @@ function SubmissionDetailsContent() {
...
+    const submissionCrumb = { label: "Submissions", path: "/submissions" };
+    const messageSearchCrumb = {
+        label: "Message Tracker",
+        path: "/admin/message-tracker",
+    };
     const crumbs: CrumbConfig[] = [
-        { label: "Submissions", path: "/submissions" },
+        (location.state as any).from === "/admin/message-tracker"
+            ? messageSearchCrumb
+            : submissionCrumb,
```
kevinhaube, line 139: From what I've seen in design, we almost certainly have a more generalized need for this behavior. Can we work this into the `Crumbs` component so any use of Breadcrumbs could utilize it?

An immediate benefit would be navigation between various resource pages in our `Product`, `Resources` and `Support` public pages. Right now, their crumbs are hard-coded to go back to the index of each particular page. So, in the case of navigating from `index a > page a` to `index b > page b` your backwards crumb will not take you back to `page a`, it'll take you to `index b`, and then you're lost or reliant on the browser back button.

I'll leave it to @stephenkao and you to determine if it's in/out of scope here but flagging for posterity!

stephenkao: Hmmmm actually, this might be a question for @jillian-hammer and @chris-kuryak, but I think there are two separate concepts here:

- Breadcrumbs, which are named absolute links to illustrate website taxonomy
- Back button, which are arbitrary links to return to the point of origin

Clicking on a breadcrumb that says X should always go to X (hard link) whereas clicking on a back button should return you to where you just were (pop from history).

Like GitHub's file explorer is an example of breadcrumbs:
Screenshot 2023-01-05 at 11 02 32
(Clicking on "styles" will always navigate to the "styles" directory in GitHub regardless of where you navigated from.)

In this particular case, I think we're trying to force a breadcrumb to be a back button -- so why not just change this to be a back button (with a dynamic "from" path)?

kevinhaube: Hmm, I think my bias is towards trialing back buttons! But maybe we could consult Insights about it 👀

jpandersen87: Stephen brings up a good point. If we were to stick to the breadcrumb philosophy we would create a new route that goes /message-tracker/:submissionid which would then have a static breadcrumb back to message-tracker rather than going into submission's subroute.

stephenkao: Let's bring this up on the ticket to get feedback from Jillian and Chris!

frontend-react/src/components/IconButton/IconButton.tsx
```diff
@@ -0,0 +1,46 @@
+import { Button } from "@trussworks/react-uswds";
```
stephenkao, line 1: This is something that @etanb and I chatted about yesterday, but I think we're going to have to start thinking about our directory structure for more scalability.  Specifically for atomic components like this, it'd be great to put them in a `core/` or `common/` directory so that we know they're either very widely used throughout our site or wrapping/replacing Trussworks' components.

Off the top of my head, my feeling is that we can start making "domains" (from domain-driven design) that'll have a structure like:

```
components/  # where all components for this domain live
tests/               # where all tests for this domain live
utils/                # where all utility functions/classes for this domain live
hooks/             # where all hooks for this domain live
```

For our use case, I'd imagine it to be something like:
```
frontend-react/
  src/
    --- new domain --- 
    core/
      components/
        IconButton.tsx
      tests/
        IconButton.test.tsx
        DateTimeUtils.test.ts
      utils/
        DateTimeUtils.ts
      hooks/

    --- new domain --- 
    analytics/
      components/
      tests/
        Analytics.test.ts
      utils/
        Analytics.ts
      hooks/
        UseCustomTrackEvent.ts # or whatever if we'd ever need that 🤷

    --- new domain ---
    organizations/ # idk maybe?
```

What do you think about this?  I can also bring it up in Slack so we can chat more about it as a team!

frontend-react/src/components/IconButton/IconButton.tsx
```diff
@@ -0,0 +1,46 @@
...
+     */
+    small?: boolean;
+    /**
+     * @deprecated since 1.9.0
+     */
+    icon?: boolean;
+    unstyled?: boolean;
+}
+
+export interface IconButtonProps extends ButtonProps {}
```
stephenkao, line 33: Do we need this?

frontend-react/src/components/Crumbs.tsx
```diff
@@ -11,28 +16,49 @@ interface CrumbConfig {
...
+const Crumbs = ({ crumbList, noPadding, previousPage }: CrumbsProps) => {
+    if (crumbList || previousPage) {
         return (
             <div className={!noPadding ? "grid-container" : ""}>
-                <BreadcrumbBar>
-                    {crumbList?.map((crumbConfig) => (
-                        <Breadcrumb
-                            key={`${crumbConfig.label}`}
-                            current={!crumbConfig?.path}
+                {previousPage !== undefined ? (
```
stephenkao, line 26: Minor point and far, far, far from a blocker, but we can do `Boolean(previousPage)` to be a little more exclusive of all falsy types.  I realize TypeScript should help prevent this, but this'll help make it a little more robust in case we do eventually want to allow passing other falsy values (`null`|`false`|'""`, et cetera) down the line 🤷

frontend-react/src/components/Crumbs.tsx
```diff
@@ -11,28 +16,49 @@ interface CrumbConfig {
...
-                            )}
-                        </Breadcrumb>
-                    ))}
-                </BreadcrumbBar>
+                            <IconArrowBack />
+                        </IconButton>
+                        Return to {previousPage}
+                    </div>
+                ) : null}
+                {crumbList !== undefined && crumbList.length > 0 ? (
```
stephenkao, line 38: This can be shortened to `crumbList?.length > 0` 🧹

frontend-react/src/components/IconButton/IconButton.tsx
jpandersen87: I'm torn between a component folder per component containing all their relevant files (component, test, storybook story, etc.) or those types of files having their own dedicated folder. Both makes sense, but I do like having a component folder that I know I can go directly to get all relevant just for that specifically.

jpandersen87: For clarity, this:

`frontend-react/
    src/
        core/
            components/
                IconButton/
                    IconButton.tsx
                    IconButton.test.tsx
                    IconButton.story.tsx
...`

frontend-react/src/components/IconButton/IconButton.tsx
jpandersen87: I feel it's good practice for every component to have a props interface exposed, even if in this case there's nothing added (yet) on top of ButtonProps. However, I can drop if desired.

frontend-react/src/components/Crumbs.tsx
```diff
@@ -11,28 +16,49 @@ interface CrumbConfig {
...
-                    {crumbList?.map((crumbConfig) => (
-                        <Breadcrumb
-                            key={`${crumbConfig.label}`}
-                            current={!crumbConfig?.path}
+                {previousPage !== undefined ? (
+                    <div className="font-sans-lg line-height-sans-4 margin-top-8">
+                        <IconButton
+                            size="big"
+                            type="button"
+                            onClick={() => window.history.back()}
```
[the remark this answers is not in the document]
jpandersen87: Technically that would work, though seems weird since I would be passing the event as an argument to the back function (in this case it would technically work since back doesn't use arguments).

frontend-react/src/components/Crumbs.tsx
jpandersen87: I do like the explicit expressiveness of it. Is there an eslint rule that we could use for this as a norm?

jpandersen87: Confusingly, typescript's Boolean is not typed correctly to allow called it as a function?

frontend-react/src/components/Crumbs.tsx
jpandersen87: Unfortunately typescript will still assume crumbList is possibly undefined later without an explicit undefined check. I presume this is possibly why your Boolean() suggestion in another comment isn't allowed.

frontend-react/src/components/IconButton/IconButton.tsx
stephenkao: Sure, that sounds good to me!  I don't have a strong preference of how domains are organized internally (as long as they're structured well) -- just that we have proper separation of concerns in the filesystem!

How about we hold off on directory reorganization for now and bring it up during our Experience eng sync on Thursday to get everyone involved?  If we decide on a more dramatic change like this, we can use the new pattern for any new components and do a gradual migration for existing ones.  What do you think?

frontend-react/src/components/IconButton/IconButton.tsx
stephenkao: Sure, not a problem in my mind!

frontend-react/src/components/IconButton/IconButton.tsx
jpandersen87: Absolutely. Would you like me to fold IconButton into the components folder directly for consistency in the current norm?

frontend-react/src/components/Crumbs.tsx
stephenkao: I'm not sure about an ESLint rule (we can look around, though!), but that's very odd about TypeScript 🤔 🤔 🤔 

It seems okay locally for me!  What's the error that's coming up for you?

frontend-react/src/components/Crumbs.tsx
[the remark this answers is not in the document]
stephenkao: Definitely not an issue -- just thinking about [the discussion Kevin and I had yesterday](https://github.com/CDCgov/prime-reportstream/pull/7883#issuecomment-1374998121).  This might be a case in which we could just pass the reference straight through rather than wrapping it in an anonymous function, but not a huge deal either way!

frontend-react/src/components/Crumbs.tsx
jpandersen87: Ok, now it's not complaining. Can't explain what happened there😕. Will commit the change

frontend-react/src/components/Crumbs.tsx
stephenkao: Oh, TypeScript why ya dumb!  That's the entire point of safe navigation 🤦 

No worries -- carry on then!

frontend-react/src/components/IconButton/IconButton.tsx
stephenkao: Ah, good point!  Yeah, if you wouldn't mind!  I think that would be good just so it'll be more consistent as we take on a possible reorganization in the future 🙏

frontend-react/src/components/Crumbs.tsx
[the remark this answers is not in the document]
jpandersen87: Yeah I've always been a bit confused on best practices with event handler anonymous functions when the supposed golden rule is to not use them to prevent memory churn in re-renders. Is it that big of a problem on event handlers? Do we need to pull out hook tooling to memo every single one? etc.

frontend-react/src/components/Crumbs.tsx
```diff
@@ -11,28 +16,49 @@ interface CrumbConfig {
...
-                </BreadcrumbBar>
+                            <IconArrowBack />
+                        </IconButton>
+                        Return to {previousPage}
+                    </div>
+                ) : null}
+                {crumbList !== undefined && crumbList.length > 0 ? (
+                    <>
+                        <hr />
+                        <BreadcrumbBar>
```
stephenkao, line 41: Really small thing, but there's a bit of vertical asymmetry on less-than-desktop viewports since `usa-breadcrumb` has bottom padding of 1rem 🤔

Screenshot 2023-01-10 at 13 25 49

Not a blocker in my mind, though!

stephenkao (approved):
Lots of tiny, nitpicky points, but this looks great!  Thanks for all your work on this and for looking into the automated routes 🙏

PR #7861, opened 2023-01-05
Experience-7857: Add ReactQueryDevtools for query/mutation debugging

### LINKED ISSUE #7857: Add ReactQueryDevTools for debugging query/mutation statuses
## Problem statement

We don't have an easy way of inspecting the current status (mount state, staleness, count, et cetera) of react-query's active queries/mutations other than manually keeping track of the network tab or globally exposing the query cache.

## What you need to know
react-query offers a [devtools](https://tanstack.com/query/v4/docs/react/devtools) utility that would help us do that.  Even though there's no risk in enabling it on production environments, we can make sure it's only enabled in development mode.

stephenkao:
This PR ...

This changeset adds the ReactQueryDevtools utility to our codebase to let us more easily debug query/mutation statuses in development mode.  When the devtools are open, you can see the statuses of queries and mutations ("mountedness" and staleness), and you can retrigger them or even delete them from the cache entirely.

There's also a possibility of lazy-loading the devtools in production mode if we need that, but I think we can just include it in development for now.

Test Steps:
1. Start up the local Webpack dev server: `yarn run start:localdev`
2. Load up any page
3. Look for the devtools icon on the lower-left corner
Screenshot 2023-01-05 at 10 29 12
4. Start making requests through react-query and see the devtools populate
Screenshot 2023-01-05 at 10 29 03

### Testing
- [x] Tested locally?
<strike>- [ ] Added tests?</strike>

## Linked Issues
- Fixes #7857

### DIFF
--- a/frontend-react/src/components/AppWrapper.tsx
+++ b/frontend-react/src/components/AppWrapper.tsx
@@ -2,6 +2,7 @@ import { PropsWithChildren } from "react";
 import { Security } from "@okta/okta-react";
 import { OktaAuth } from "@okta/okta-auth-js";
 import { QueryClientProvider } from "@tanstack/react-query";
+import { ReactQueryDevtools } from "@tanstack/react-query-devtools";
 import { AppInsightsContext } from "@microsoft/applicationinsights-react-js";
 
 import SessionProvider, { OktaHook } from "../contexts/SessionContext";
@@ -38,6 +39,7 @@ export const AppWrapper = ({
                                 {children}
                             </FeatureFlagProvider>
                         </AuthorizedFetchProvider>
+                        <ReactQueryDevtools initialIsOpen={false} />
                     </QueryClientProvider>
                 </SessionProvider>
             </AppInsightsProvider>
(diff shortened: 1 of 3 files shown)


frontend-react/src/components/AppWrapper.tsx
```diff
@@ -38,6 +39,7 @@ export const AppWrapper = ({
                                 {children}
                             </FeatureFlagProvider>
                         </AuthorizedFetchProvider>
+                        <ReactQueryDevtools initialIsOpen={false} />
```
kevinhaube, line 42: Does this automagically know it's in DEV, or do we need an env condition?

stephenkao: It's only included when `NODE_ENV === 'development'`:

> By default, React Query Devtools are only included in bundles when process.env.NODE_ENV === 'development', so you don't need to worry about excluding them during a production build.

There's [an explicit method](https://tanstack.com/query/v4/docs/react/devtools#devtools-in-production) to include them in production, but I think having it available locally first should be the first step.

jpandersen87:
👍

jpandersen87 (approved):

PR #7883, opened 2023-01-06
Change `button` to `Button` for consistency

### LINKED ISSUE #7837: Replace native HTML button elements with Trussworks' Button component
## Problem statement

We have a few places where we're using native HTML `button` elements instead of Trussworks' `Button` element.  Since we're already using the latter more frequently in the codebase, we should convert the native HTML elements so they're consistent across the board.  This'll also help with any migration should we choose to migrate away from Trussworks' component system.

This is relevant to [6652](https://app.zenhub.com/workspaces/experience-607d9d5e68b95200150fec37/issues/gh/cdcgov/prime-reportstream/6652).

## What you need to know

There are two places in the code that we're using native HTML buttons:
- [Pagination.tsx](https://github.com/CDCgov/prime-reportstream/blob/master/frontend-react/src/components/Table/Pagination.tsx#L35)
- [ColumnData.tsx](https://github.com/CDCgov/prime-reportstream/blob/master/frontend-react/src/components/Table/ColumnData.tsx#L103)

## Acceptance criteria

- [ ] The `button`s are replaced with `Button`s with no user-facing impact

## Related links

- [GitHub discussion](https://github.com/CDCgov/prime-reportstream/pull/7796#discussion_r1060918942)

kevinhaube:
This PR swaps all HTML element `button`s with the Trussworks exported `Button` component throughout the application.

Test Steps:
> It is excpeted that _any_ button linking to a page on our own site/app should _not_ refresh the application. Keep an eye out for an app refresh. This would be considered a bug!
1. Ensure public site content buttons function properly 
2. Ensure table buttons (including filtering and pagination) function properly
3. Ensure error page buttons function properly (try navigating to a non-sanctioned route like `/ads;jfkla;`

## Changes
- Swaps all `button` invocations to `Button`

### Testing
- [X] Tested locally?
- [ ] Added tests?

### Process

## Linked Issues
- Fixes #7837 

## To Be Done
- #issue

    - new endpoint A
    - new endpoint B

### DIFF
--- a/frontend-react/src/components/Table/ColumnData.tsx
+++ b/frontend-react/src/components/Table/ColumnData.tsx
@@ -1,4 +1,5 @@
 import React, { ReactNode } from "react";
+import { Button } from "@trussworks/react-uswds";
 
 import { USLink } from "../USLink";
 
@@ -100,12 +101,9 @@ export const ColumnData = ({
             return action();
         };
         return tableData(
-            <button
-                className="usa-link bg-transparent border-transparent"
-                onClick={() => doAction()}
-            >
+            <Button type="button" unstyled onClick={() => doAction()}>
                 {displayValue}
-            </button>
+            </Button>
         );
     }
 
--- a/frontend-react/src/components/Table/Pagination.tsx
+++ b/frontend-react/src/components/Table/Pagination.tsx
@@ -32,17 +32,22 @@ const PaginationPageNumber: React.FC<PaginationPageNumberProps> = ({
 }) => {
     return (
         <li className="usa-pagination__item usa-pagination__page-no">
-            <button
+            {/* Using `unstyled` and custom classes is a method used by Trussworks,
+            but they do not export a `PaginationButton` component, so we have to rewrite
+            it here.
+            See: https://github.com/trussworks/react-uswds/blob/main/src/components/Pagination/Pagination.tsx */}
+            <Button
+                type="button"
+                unstyled
                 {...(isCurrentPage && { "aria-current": "page" })}
                 aria-label={`${isLastPage ? "last page, " : ""}Page ${pageNum}`}
-                className={classnames(
-                    ["usa-pagination__button", "rs-pagination-no-button"],
-                    { "usa-current": isCurrentPage }
-                )}
+                className={classnames(["usa-pagination__button"], {
+                    "usa-current": isCurrentPage,
+                })}
                 onClick={() => setSelectedPage(pageNum)}
             >
                 {pageNum}
-            </button>
+            </Button>
         </li>
     );
 };
--- a/frontend-react/src/pages/About.tsx
+++ b/frontend-react/src/pages/About.tsx
@@ -1,11 +1,14 @@
 import DOMPurify from "dompurify";
+import { Button } from "@trussworks/react-uswds";
+import { useNavigate } from "react-router-dom";
 
 import { BasicHelmet } from "../components/header/BasicHelmet";
-import { USExtLink, USLink } from "../components/USLink";
+import { USExtLink } from "../components/USLink";
 
 import site from "./../content/site.json";
 
 export const About = () => {
+    const navigate = useNavigate();
     return (
         <>
             <div className="grid-container rs-documentation usa-prose desktop:margin-top-6">
@@ -18,14 +21,20 @@ export const About = () => {
                     labs, and more through a single connection.
                 </h2>
                 <p>
-                    <button className="usa-button usa-button--outline">
-                        <USLink href="/product">
-                            Learn more about ReportStream{" "}
-                        </USLink>
-                    </button>
-                    <button className="usa-button usa-button--outline">
-                        <USLink href="/support/contact">Contact us</USLink>
-                    </button>
+                    <Button
+                        type="button"
+                        outline
+                        onClick={() => navigate("/product")}
+                    >
+                        Learn more about ReportStream
+                    </Button>
+                    <Button
+                        type="button"
+                        outline
+                        onClick={() => navigate("/support/contact")}
+                    >
+                        Contact us
+                    </Button>
                 </p>
                 <hr />
                 <p>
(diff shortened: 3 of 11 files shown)


kevinhaube:
@stephenkao Thoughts on the code smells? It seems like a new rule, haven't seen it before, but it's alternatively like any use of an arrow function within a prop, like `onClick` to be created as a callback.

```typescript
<Button onClick={() => navigate(-1)}>Back</Button> // "Bad" method
// Where navigateBack = useCallback(() => navigate(-1), [])
<Button onClick={navigateBack}>Back</Button> // "Good" method
```

While on the surface this looks like a valid performance issue, I'm not so sure. The `navigate` function [is already a callback](https://github.com/remix-run/react-router/blob/50a6f10493e6f4e2bf059d0ff288601b446441ca/packages/react-router/lib/hooks.tsx#L178-L218) who's refresh is handled by the `react-router` code.

stephenkao:
> @stephenkao Thoughts on the code smells? It seems like a new rule, haven't seen it before, but it hates any use of an arrow function within a prop, like `onClick`.
> 
> ```ts
> <Button onClick={() => navigate(-1)}>Back</Button> // "Bad" method
> // Where navigateBack = useCallback(() => navigate(-1), [])
> <Button onClick={navigateBack}>Back</Button> // "Good" method
> ```
> 
> While on the surface this looks like a valid performance issue, I'm not so sure. The `navigate` function [is already a callback](https://github.com/remix-run/react-router/blob/50a6f10493e6f4e2bf059d0ff288601b446441ca/packages/react-router/lib/hooks.tsx#L178-L218) who's refresh is handled by the `react-router` code. To my knowledge, `onClick` has required a `() => {}` wrapper as to not trigger the function on render of the button, so I could try removing it and see if that's still the case. I think the fact it's a callback already may actually negate the need for that arrow function wrapper. We'll see!

Yeahhhhhh, I don't think this is a good one-size-fits-all rule.  `useCallback` can have [negative performance implications](https://kentcdodds.com/blog/usememo-and-usecallback) given that it has an initial cost to memory complexity (though it's small), and it's not often that we need to memoize _functions_ since garbage collection should take care of unreferenced ones in most cases.  I'd be in favor of disabling this rule and just make sure we catch opportunities for memoization manually during code review.

kevinhaube:
@stephenkao Tried the `navigate` function without the wrapper, and got: 
```bash
TS2322: Type 'void' is not assignable to type 'MouseEventHandler<HTMLButtonElement> | undefined'.  index.d.ts(1446, 9): The expected type comes from property 'onClick' which is declared here on type 'IntrinsicAttributes & ButtonProps & ClassAttributes<HTMLButtonElement> & ButtonHTMLAttributes<HTMLButtonElement>'
```
Feels like we can go ahead and ignore the smells. Not sure how we disable them for SonarCloud tho

frontend-react/src/components/Table/ColumnData.tsx
```diff
@@ -100,12 +101,9 @@ export const ColumnData = ({
             return action();
         };
         return tableData(
-            <button
-                className="usa-link bg-transparent border-transparent"
-                onClick={() => doAction()}
-            >
+            <Button type="button" unstyled onClick={() => doAction()}>
```
stephenkao, line 104: Could this be simplified to `onClick={ doAction }` or does TypeScript complain?

kevinhaube: Yep, worked fine!

frontend-react/src/components/Table/Pagination.tsx
```diff
@@ -32,17 +32,22 @@ const PaginationPageNumber: React.FC<PaginationPageNumberProps> = ({
...
+            <Button
+                type="button"
+                unstyled
                 {...(isCurrentPage && { "aria-current": "page" })}
                 aria-label={`${isLastPage ? "last page, " : ""}Page ${pageNum}`}
-                className={classnames(
-                    ["usa-pagination__button", "rs-pagination-no-button"],
-                    { "usa-current": isCurrentPage }
-                )}
+                className={classnames(["usa-pagination__button"], {
```
stephenkao, line 44: VERY minor point, but this doesn't need to be an array:

```
className={ classnames("usa-pagination__button", { ... }) }
```

`classnames` is a [variadic function](https://github.com/JedWatson/classnames/blob/main/index.js#L16) anyway!

frontend-react/src/pages/About.tsx
```diff
@@ -18,14 +21,20 @@ export const About = () => {
...
                 <p>
-                    <button className="usa-button usa-button--outline">
-                        <USLink href="/product">
-                            Learn more about ReportStream{" "}
-                        </USLink>
-                    </button>
-                    <button className="usa-button usa-button--outline">
-                        <USLink href="/support/contact">Contact us</USLink>
-                    </button>
+                    <Button
```
stephenkao, line 24: Woohoo!  Thanks for fixing this as well!

frontend-react/src/components/Table/__snapshots__/Pagination.test.tsx.snap
```diff
@@ -47,7 +47,9 @@ exports[`Pagination Renders the expected markup when the current page is between
       >
         <button
           aria-label="Page 1"
-          class="usa-pagination__button rs-pagination-no-button"
```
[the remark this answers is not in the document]
kevinhaube: This is the _only_ snapshot test I know of. A former engineer used them during their brief time here, but we've never found them useful. In fact, this created a pain when updating tests because it's not in line with our standard testing methods.

frontend-react/src/pages/About.tsx
kevinhaube: Tried to catch all those nasty links-in-buttons :D

frontend-react/src/components/Table/Pagination.tsx
kevinhaube: Oh, good catch!

Issue #7754, opened 2022-12-22
Redesign Message Tracker Details based on internal usability findings


chris-kuryak:
## Problem statement

We want to implement some changes to how the data is conveyed on the Message Tracker to improve the visual hierarchy and overall readability based on internal usability findings.

## What you need to know
- The new design includes two primary sections -- one about the Sender and one list the Receivers -- with more emphasis on actionable data.  See the [Figma](https://www.figma.com/proto/6mwI5ac6rprACKDzDo4Ady/ReportStream-Workspace-%7C-2023?node-id=393%3A4746&starting-point-node-id=393%3A4746) for reference.
- Style Guide specs in [Figma](https://www.figma.com/file/6mwI5ac6rprACKDzDo4Ady/ReportStream-Workspace-%7C-2023?node-id=708%3A8503&t=ZQx7bNZo5AQSeFij-1)
- **NOTE:** Do not include Performing Facility field in this MVP.

## Acceptance criteria

- [x] All sections
   - [x] Divided by a horizontal border/rule
   - [x] All sizes and spacings match the design

--- 

- [x] Title section
  - [x] Page title is renamed to `Message ID Search`

---

- [x] Sender section
  - [x] Shows updated data:
    - [x] Submitter: currently `Sender Name`
    ~~- [ ] Performing Facility: net new field~~  removed from AC as this new field caused tech issues
    - [x] Performing Facility field is removed (even though it is shown in the Figma file)
    - [x] Incoming Report ID: keep as-is
    - [x] Date/Time Submitted: existing field (`submittedDate`) not yet displayed
    - [x] File Location: existing field (`fileUrl`) but needs to be parsed
    - [x] Incoming File Name: existing field (`fileUrl`) but needs to be parsed
  - [x] Warnings and errors are grouped in [Accordions](https://designsystem.digital.gov/components/accordion/), each with the respective counts

PR #7885, opened 2023-01-06
Add performing facility to Message ID Details

### LINKED ISSUE #7850: Update Message structure to include additional data
## Problem statement

We need to add some additional fields to the messages response payload so we can display them in the redesigned Message Tracker.  Specifically, the missing fields are:

- Performing Facility

## What you need to know

- For `Performing Facility`:
  - See [processMesageDetailRequest](https://github.com/CDCgov/prime-reportstream/blob/master/prime-router/src/main/kotlin/azure/MessagesFunctions.kt#L156) as a starting point
  - We should return `testing_facility_name` as part of the payload

## Acceptance criteria

- [ ] The message detail endpoint (`/api/message/<report_index>`) includes the above fields.

kevinhaube:
This PR adds the `testingLabName` field to the response for `/api/message/<report_index>`

Sample Response (test data in my local env)
```json
{
  "id" : 16,
  "messageId" : "797620",
  "sender" : "simple_report.default",
  "submittedDate" : "2022-08-31T13:57:49.609785",
  "reportId" : "342a21b3-fa85-4b88-9717-49e55ead1205",
  "fileName" : null,
  "fileUrl" : "http://localhost:10000/devstoreaccount1/reports/receive%2Fsimple_report.default%2Fpdi-covid-19-342a21b3-fa85-4b88-9717-49e55ead1205-20220831135748.csv",
  "testingLabName" : "Any lab USA", // New value!
  "warnings" : [ ],
  "errors" : [ ],
  "receiverData" : [ ] // Omitted for brevity
}
```

Test Steps:
>**NOTE:** Test data in your local environment, and a local database explorer like DBeaver is necessary for locating a valid message ID in the covid metadata table.
1. Use the `/api/message/<report_index>` endpoint with a valid message ID to ensure the property is returned

## Changes
- Adds `testingLabName` as a string in the JSON response for the message details endpoint

### Testing
- [X] Tested locally?
- [ ] Added tests?

### Process

## Linked Issues
- Fixes #7850 

## To Be Done
- #issue

    - new endpoint A
    - new endpoint B


prime-router/docs/api/messages.yml
```diff
@@ -76,6 +76,9 @@ components:
         fileUrl:
           description: the blobstore url of the file the message belongs to
           type: string
+        testingLabName:
```
arnejduranovic, line 79: The endpoints that return the `Message` object, `/api/message/{id}` and `/api/messages`, both appear to be NOT covid specific. These endpoints, as documented here, functionally do not have to rely on the covid database table, although currently they do. Adding `testingLabName` here would now make the `messages` endpoints tightly coupled to the covid table, which I think would be a significant step backwards, seeing as we would like to move away from the covid pipeline. 

Furthermore, I do not see why a seemingly generic feature such as "message tracker" needs to be coupled to covid at all. If "message tracker" needs to show covid-specific information, I would propose we add a NEW endpoint that can return a NEW covid-table-specific message object, like `CovidMessage`, for example. Then, if it determines the message in question is a covid message, the message tracker feature can call the new endpoint in addition to the current endpoint to get ALL the covid-specific data, not just `testingLabName`. If the O+O team currently wants "testingLabName" in the message tracker, it seems reasonable that they may request additional data be provided in the future. Furthermore, it goes without saying, we should update the query to use the `elr_result_metadata` table instead of covid table. This method has the following advantages over the current approach:

1. The current messages endpoints remain decoupled from any specific "topic" and, as such, can be used for other topics/pipelines
2. Adding a new endpoint for covid messages means we can add all the covid specific data-points you want without breaking any good practices, since it would be specific to covid

I propose we close this PR and pass this off to the platform team, if time permits. If this is very high priority and needs to be merged, let me know and that platform team can do the proposed updates at another time. If this is the route we take, then I suppose we can just consider the endpoints in this file the "covid specific" ones and platform team can create a new set of "generic elr" message endpoints.

arnejduranovic (changes requested):

prime-router/docs/api/messages.yml
kevinhaube: @chris-kuryak Can we get a product perspective?
A quick TL;DR

This addition was scoped as a blocker for the UI updates in #7754 but Arnej has brought up a great point, that this feature is hyperspecific to `covid-19` messages. I'd double check this with O&O, but I believe we _do_ have at least one production user sending `full-elr` and a few more that send `monkeypox` messages. 

Arnej's approach of utilizing the general table would open up our message ID details feature to these other non-covid topics, however, there would be additional scoping needed to figure out what else is encompassed in that change. Similarly, we aren't using the general table for Message _Search_, so this would be an update to two endpoints and their underlying functionality.

The question(s) at hand:

1. Though it wasn't originally, should this in scope for our current Message Tracker iteration?
2. If YES, is this a separate unit of work or should we consider this a blocker for our current Message Tracker UI updates?

@arnejduranovic @stephenkao Feel free to add in any details I missed!

Chris, as a heads up you're not in `prime-reportstream-platform` on Slack, where this thread originated. I tried to import you, but I was not authorized. USDS has to do it. :)

stephenkao: Thanks for following up, all!  This sounds good to me with two points:

- My feeling is that we can start on the Message Tracker redesign without waiting for our decision here and just omit the `Performing Facility` field for now.  It's only a single field, and I don't think that should block our progress on the majority of our work here.
- If/when we go forward with updating the backend to use the `elr_result_metadata` table, I think we'll just need to include the `topic` from the `messageSearch` function so we know which endpoint (COVID, flu, mpox, et cetera) to send a request to.  That works for me!

kevinhaube: >If/when we go forward with updating the backend to use the elr_result_metadata table, I think we'll just need to include the topic from the messageSearch function so we know which endpoint (COVID, flu, mpox, et cetera) to send a request to. That works for me!

Oh, I was thinking the goal was homogeny, rather than any need to distinguish a topic. Perhaps an optional front-end filter with this functionality would be useful, but I see the value in also not filtering by topic as, in cases where someone like SimpleReport transitions from covid-19 to full-elr, that'd require further work to differentiate which is which on their end.

chris-kuryak: @arnejduranovic - This is a high priority ask from the O+O team, so I would like to get them something ASAP, which informs my guidance below.

@kevinhaube @stephenkao - I agree with Stephen here, let's just remove the Performing Facility field for now and get this redesigned tool into the hands of O+O ASAP. And then depending on Jillian's response below, we can ping the O+O engineers to let them know the change and ask for feedback regarding how crucial it would be to have this field moving forward. And if it is crucial, we'll start a new ticket to add it in a future sprint.
**What do you two say to that?**

@jillian-hammer - Due to some technical issues, I think it is best for us to remove the Performing Facility field from this MVP launch and figure out a solution to include it at a later date. Was that field a Highly Valuable Field as requested by the O+O team during your research, or do you think we can proceed without it for now?

kevinhaube: Sounds good to me! I can close this PR for now, and leave the branch open. If O&O decides it's crucial, we can put this in, if not, I'd advocate for generalizing Message ID Search and Details endpoints to query the universal table with the expectation of receiving _all_, including covid-19, messages, _then_ build onto it with any new fields we need.

kevinhaube:
Closing for now. Will re-open **IF** O&O finds this to be critical for this iteration. If not, efforts will be made by Platform to generalize this endpoint for retrieval of _all_ messages, not strictly those within the covid metadata table. :)

prime-router/docs/api/messages.yml
stephenkao: @arnejduranovic We've updated the Experience tickets to omit this change as a requirement for now.  When you get a chance, would you mind adding a Platform ticket with your proposal linking back to this discussion for context?

arnejduranovic: Created a new ticket here to capture the proposed work: https://github.com/CDCgov/prime-reportstream/issues/7892. Please note that this ticket should only be done if O+O really wants that COVID data in the message tracker, have we confirmed that is the case? Or can we just keep it not specific to any topic?

stephenkao: Thanks, Arnej!  We're going to follow up with O+O to see if this field is needed or if they think any COVID-specific data will be useful in the future.  I can follow up on that ticket once we reach a decision, but yeah, let's hold off on any work on this till then.

PR #7887, opened 2023-01-06
Experience-7886: Fix memberships.has in admin check

### LINKED ISSUE #7886: Admin memberships check causes an memberships.has is not a function when not a Map
## Describe the bug

We have [a check in SessionContext](https://github.com/CDCgov/prime-reportstream/blob/master/frontend-react/src/contexts/SessionContext.tsx#L51) to determine if a user is an admin.  That breaks if `memberships` is falsy/not a `Map`.

## Impact on ReportStream

If users encounter this error, they won't be able to use the site unless they explicitly clear their local storage.

## Context
We can change this to check for `activeMembership.memberType == 'prime-admin'` since that'll never change with the removal of that feature in [7844](https://app.zenhub.com/workspaces/experience-607d9d5e68b95200150fec37/issues/gh/cdcgov/prime-reportstream/7844).

stephenkao:
This PR ...

This changeset addresses a bug wherein the `memberships.has` check for an admin status fails because `memberships` is not a Map.  This requires users to clear their local storage in order to continue using the site.

Since we no longer have the ability to mock a user's member type, we can just check if their active membership is an admin member type instead.

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- Fixes #7886


kevinhaube (approved):
Looks good, and solves our `memberships.has` issue! For some context, that `memberships` property itself _could_ go away if we know that nobody will be a member of two OktaGroups.

We added it back in the day, because so many admins, for some reason, were members of more than one group, and [we default to your first organization](https://github.com/CDCgov/prime-reportstream/blob/master/frontend-react/src/hooks/UseOktaMemberships.ts#L134-L135) in the Okta `groups` claim. 

I think a more _accurate_ behavior to ensure no admin's misconfiguration in Okta ruins their experience would be to check there, "hey, is this user a part of our Admin group", and if so, set that, if not, default to the first. Then, we could do away with `memberships` because nowhere on the site do we expect or allow users to swap to another organization except through PrimeAdmin tools. 

Just a thought for cleanup of `useOktaMemberships` and our `MembershipState`!

penny-lischer:
@stephenkao thanks for the fix on this!

stephenkao:
> Looks good, and solves our `memberships.has` issue! For some context, that `memberships` property itself _could_ go away if we know that nobody will be a member of two OktaGroups.
> 
> We added it back in the day, because so many admins, for some reason, were members of more than one group, and [we default to your first organization](https://github.com/CDCgov/prime-reportstream/blob/master/frontend-react/src/hooks/UseOktaMemberships.ts#L134-L135) in the Okta `groups` claim.
> 
> I think a more _accurate_ behavior to ensure no admin's misconfiguration in Okta ruins their experience would be to check there, "hey, is this user a part of our Admin group", and if so, set that, if not, default to the first. Then, we could do away with `memberships` because nowhere on the site do we expect or allow users to swap to another organization except through PrimeAdmin tools.
> 
> Just a thought for cleanup of `useOktaMemberships` and our `MembershipState`!

Good point!  I definitely agree we can remove `memberships` + clean up our organization impersonation going forward, but that's out of the scope of this pull request.  I'll file a separate ticket for that, though!

PR #7923, opened 2023-01-10
ServiceSettings in MembershipState

### LINKED ISSUE #7328: Add Senders hook
## Problem statement

Organizations in ReportStream have "services", such as Sender services and Receiver services. These are what we call "Senders" or "Receivers" internally for short. Because an Organization may have one, the other, or both, and they are required parameters for many features, we should have an way to fetch and cache the Organization-specific Senders and Receivers.

## What to know

- We're going to be disabling redundant refetches as part of [this changeset](https://github.com/CDCgov/prime-reportstream/pull/7928/files#r1080627826)

- We already have a [`useOrganizationReceivers` hook](https://github.com/CDCgov/prime-reportstream/blob/master/frontend-react/src/hooks/UseOrganizationReceivers.ts), and we have [the endpoint configuration to fetch Senders](https://github.com/CDCgov/prime-reportstream/blob/master/frontend-react/src/config/endpoints/settings.ts#L65).

To that point, we can add a hooks:
- `useOrganizationSenders` to fetch an Organization's Senders

## Acceptance criteria
With this hook:
- [ ] Senders are fetched once for the current Organization when needed

kevinhaube:
This PR adds code that fetches a user's sender and receiver service settings and provides them across the app at the SessionContext level.

Test Steps:
1. Log in to RS locally
2. Use React DevTools (Components > search for SessionProvider) or localStorage to view current `activeMembership` values
3. Use the org switcher functionality to change to any org
4. Ensure the `activeMembership` updates with the fetched senders and/or receivers for the given org.

## Changes
- Adds `useMemberServices`
- Integrates the above new hook with `useOktaMemberships`
- Unit and integration testing for new hook

### Testing
- [X] Tested locally?
- [x] Added tests?

### Process

## Linked Issues
- Fixes #7328 

## To Be Done
- #issue

    - new endpoint A
    - new endpoint B


frontend-react/src/__mocks__/SettingsMockServer.ts
```diff
@@ -1,16 +1,43 @@
...
+};
+
 const handlers = [
     rest.put(
         `http://localhost:3000${senderSettingsUrl("abbott", "user1234")}`,
         (req, res, ctx) => {
             return res(ctx.status(200));
         }
     ),
+    rest.get(
```
kevinhaube, line 23: Anyone got a good way to consolidate these?? I _think_ I can just replace `senders` and `receivers` with `:serviceType` and never actually measure it in the response. Senders and Receivers follow the same data shape, but my concern is that the variable path would open the door to some hard-to-track testing issues down the line. So I left it as WET as I could here haha

frontend-react/src/components/AppWrapper.tsx
```diff
@@ -32,16 +32,16 @@ export const AppWrapper = ({
     return (
         <Security oktaAuth={oktaAuth} restoreOriginalUri={restoreOriginalUri}>
             <AppInsightsProvider>
-                <SessionProvider oktaHook={oktaHook}>
-                    <QueryClientProvider client={appQueryClient}>
+                <QueryClientProvider client={appQueryClient}>
+                    <SessionProvider oktaHook={oktaHook}>
```
kevinhaube, line 36: This ended up being a moot point since we didn't use a useQuery fetch, but I stand by it as a cleanup measure. Nothing inside QueryClientProvider relies on SessionProvider. With the new arrangement, all the external providers (from libraries) are up top, and ours are nested within.

frontend-react/src/hooks/network/ServiceHooks.ts
```diff
@@ -0,0 +1,101 @@
...
+/** Fetches membership services (senders, receivers) from the ReportStream API. Updates
+ * whenever the given {@link MembershipState} or AccessToken (Okta) are updated.
+ * @remarks ONLY FOR USE WITHIN {@link SessionContext} */
+export const useMemberServices = (
+    state: MembershipState,
+    token: AccessToken | undefined
+): ServiceState => {
+    const [senders, setSenders] = useState<RSService[] | undefined>();
+    const [receivers, setReceivers] = useState<RSService[] | undefined>();
+    const [activeService, setActiveService] = useState<string>("default");
```
kevinhaube, line 32: Defaulting to `"default"` as _so much_ of the site relies on it. This will change when we actually hook up features to this. Daily Data is unbroken by this, and Validate has been modified to read the value from here, as it was already a consumer of the `activeMembership.service` value.

frontend-react/src/hooks/network/ServiceHooks.ts
```diff
@@ -0,0 +1,101 @@
...
+        const fetchData = async () => {
+            const { senders, receivers } = servicesEndpoints;
+            const senderServiceResults: RSService[] = await fetcher<
+                RSService[]
+            >(senders, fetcherOptions);
+            const receiverServiceResults: RSService[] = await fetcher<
+                RSService[]
+            >(receivers, fetcherOptions);
+            setSenders(senderServiceResults);
+            setReceivers(receiverServiceResults);
```
kevinhaube, line 79: There's a _lot_ of data in these settings that do not pertain at all to a session. I'm thinking there needs to be a cleansing of these objects in each array that cuts them down to what we actually need (i.e. the true shape of `RSService` ([see here](https://github.com/CDCgov/prime-reportstream/pull/7923/files#diff-a72d42ba97677d060a8ea1988d66b9d9bca5cfed183eb9a12c08b17fd84d3ef4R9-R14))

kevinhaube: Also going to swap to a `Promise.all()` on this

kevinhaube: Screen Shot 2023-01-11 at 5 14 56 PM
Screen Shot 2023-01-11 at 5 15 35 PM

frontend-react/src/contexts/SessionContext.tsx
```diff
@@ -43,8 +44,8 @@ const SessionProvider = ({
         state: { activeMembership, initialized },
         dispatch,
     } = useOktaMemberships(authState);
-    /* This logic is a for when admins have other orgs present on their Okta claims
-     * that interfere with the activeMembership.memberType "soft" check */
+    /* TODO: Remove - logic can be easily checked by accessing the session context
```
stephenkao, line 47: Hmmm I feel like it's still a nice convenience value in Context, no?  We could just check `memberType` wherever we need it, but we'd have to add a comparison against `MemberType.PRIME_ADMIN` every time.  My feeling is that it's easier just to have this boolean set in one place even if it's technically redundant 🤷

frontend-react/src/hooks/UseOktaMemberships.test.ts
```diff
@@ -212,6 +235,16 @@ describe("useOktaMemberships", () => {
 });
 
 describe("helper functions", () => {
+    describe("getTypeOfGroup", () => {
+        test("returns the right type of group", () => {
+            const admin = getTypeOfGroup("DHPrimeAdmins");
```
stephenkao, line 240: PEDANTIC POINT, but we could get rid of the variables and just asset against direct invocations:

```js
expect(getTypeOfGroup("DHPrimeAdmins")).toEqual("prime-admin")
expect(getTypeOfGroup("DHSender_test-group")).toEqual("sender")
expect(getTypeOfGroup("DHtest-group")).toEqual("receiver")
```

Also, we may want to add another test case or two for the fall-through:
```js
expect(getTypeOfGroup("")).toEqual("non-standard")
expect(getTypeOfGroup("whatever")).toEqual("non-standard")
```
This case may not actually be encountered in production, but it's still a possibility in our code so I feel that warrants a test 🤷

frontend-react/src/hooks/UseOktaMemberships.ts
```diff
@@ -155,7 +160,7 @@ const calculateNewState = (
                 ),
                 initialized: true,
             };
-        case MembershipActionType.ADMIN_OVERRIDE:
+        case MembershipActionType.UPDATE_MEMBERSHIP:
```
stephenkao, line 163: What's the reason for this renaming?  I feel like `ADMIN_OVERRIDE` makes more sense since the intention here is to update the override value in LocalStorage even though it updates the global membership state all the time [in the reducer](https://github.com/CDCgov/prime-reportstream/blob/master/frontend-react/src/hooks/UseOktaMemberships.ts#L203).

It shouldn't be part of this changeset, but I'm thinking eventually we should keep `global-membership-state` as the canonical "true" user value whereas `global-organization-override` can be anything that an admin would be impersonating.  Our membership hooks/functions look like they already prioritize the override value so that might not be too huge a change.  But right now, it feels like their usage is kind of muddled 🤔 

--- 

I also wonder if we should really be accessing LocalStorage as frequently as we are since its operations are synchronous or if we can persist state in memory -- but this is a really small point in my mind; it's not like our user states are changing a thousand times per second or anything, in which case we may have a severe case of multiple personality disorder.

frontend-react/src/hooks/UseOktaMemberships.test.ts
kevinhaube: Oh crap, good catch! We should definitely be testing "non-standard" as it's the catch-all for misconfigured groups, or groups like `simple_report`, purely existing for auth, but do not need functionality given to them in our web application

frontend-react/src/contexts/SessionContext.tsx
kevinhaube: It's just weird to have it only for admins, and _not_ have an equivalent for other member types. I was thinking it made more sense to condense the conditionals on member type to instances like `AuthElement` or that `requiresServices` HOF idea (related to gating admins from pages requiring mocked state)

stephenkao: In that case, I'd argue for an `is*` predicate for each of our member types 🤷 

```js
isAdmin
isSender
isReceiver
```

I don't think HOC/HOF will have enough granularity to handle alllllll cases down the component hierarchy -- especially in the cases of individually swapped-in components or hooks.  These convenience predicates could also be used in HOCs/HOFs, but they'll give us more flexibility.

frontend-react/src/hooks/UseOktaMemberships.ts
stephenkao: Right, my follow-up point was about splitting up the logic here so that only the `global-organization-override` value would be updated rather than the `global-membership-state` as well -- the latter would essentially be static on login, and the former would change when we impersonate an Organization.  Our in-memory abstractions can factor in both memberships wherein the override would take priority, but I think it'd help to treat them functionally separately in LocalStorage.

frontend-react/src/hooks/network/ServiceHooks.ts
```diff
@@ -0,0 +1,109 @@
...
+ * whenever the given {@link MembershipState} or AccessToken (Okta) are updated.
+ * @remarks ONLY FOR USE WITHIN {@link SessionContext} */
+export const useMemberServices = (
+    state: MembershipState,
+    token: AccessToken | undefined
+): ServiceState => {
+    const [senders, setSenders] = useState<RSService[] | undefined>();
+    const [receivers, setReceivers] = useState<RSService[] | undefined>();
+    const [activeService, setActiveService] = useState<string>("default");
+    /* Because this is used at the SessionContext > useOktaMemberships level, it does not have access
```
stephenkao, line 42: Hmmm I'm wondering if we need to invoke it directly in useOktaMemberships or if we can use a lazy-fetch approach instead wherein useMemberServices can be used anywhere we need to have access to the list of services -- we could leverage 1) prepopulating the react-query cache through values in LocalStorage if applicable, 2) add the organization `parsedName` as part of the query key, and 3) make any hook that requires the list of services a dependent query or hide anything visual behind a loader.  This'd help decouple this from useOktaMemberships and obviate the need for the custom fetch since any invocations would be within the SessionContext 🤔

frontend-react/src/hooks/network/ServiceHooks.ts
```diff
@@ -0,0 +1,109 @@
...
+                ]);
+            setSenders(cleanServicesArray(senderServiceResults));
+            setReceivers(cleanServicesArray(receiverServiceResults));
+        };
+        if (hasAllNecessaryVariablesForFetch) {
+            if (state?.activeMembership?.parsedName !== "PrimeAdmins") {
+                fetchData().catch((e: RSNetworkError) => console.error(e));
+            }
+        } else {
+            console.log("Did not fetch member services");
```
stephenkao, line 94: Why do we need to log this out?

frontend-react/src/hooks/network/ServiceHooks.ts
```diff
@@ -0,0 +1,109 @@
...
+            const [receiverServiceResults, senderServiceResults] =
+                await Promise.all([
+                    fetcher<RSService[]>(receivers, fetcherOptions),
+                    fetcher<RSService[]>(senders, fetcherOptions),
+                ]);
+            setSenders(cleanServicesArray(senderServiceResults));
+            setReceivers(cleanServicesArray(receiverServiceResults));
+        };
+        if (hasAllNecessaryVariablesForFetch) {
+            if (state?.activeMembership?.parsedName !== "PrimeAdmins") {
```
stephenkao, line 90: Can we just disable these fetches if we're an admin?

frontend-react/src/hooks/network/ServiceHooks.ts
```diff
@@ -0,0 +1,109 @@
+import {
+    Dispatch,
+    SetStateAction,
+    useCallback,
+    useEffect,
+    useMemo,
+    useState,
+} from "react";
+import { AccessToken } from "@okta/okta-auth-js";
+import _ from "lodash";
```
stephenkao, line 10: Let's only import what we need from lodash:

```js
import pick from "lodash.pick"
```

Historically, lodash has been bad with tree-shaking so `import _ from "lodash"` would load in the entire library into the bundle.  `import specificThing from lodash.specificThing` makes it so we only include what we need in the bundle.  (Not sure if this is still the case, but it couldn't hurt -- especially since we don't have the larger `lodash` dependency listed in package.json.)

frontend-react/src/hooks/UseOktaMemberships.ts
```diff
@@ -217,6 +222,27 @@ export const useOktaMemberships = (
     const token = authState?.accessToken;
     const organizations = authState?.accessToken?.claims?.organization;
 
+    const memberServices = useMemberServices(state, token);
+    // Update services when these values change
+    useEffect(() => {
```
stephenkao, line 227: This is definitely out of the scope of this pull request, but I think it'd be a good idea to revisit our useEffect usages across the site.  There are a lot of cases in which we're using effects to trigger some other event based on a computed change, but with React v18's strict mode that's technically a discouraged behavior.

For this particular case, I think we could attach a [subscription](https://github.com/okta/okta-auth-js#authstatemanagersubscribehandler) to the OktaAuth instance and trigger whatever we need that way 🤷 

HOWEVER, like I said, this is definitely outside the scope of this pull request -- it's just something that I'd like the team to keep in mind when applicable.

frontend-react/src/contexts/SessionContext.tsx
kevinhaube: Oh, I think what you're suggesting here goes well with what I'm suggesting! I'm not saying we need a catch-all HOF, I'm just saying using the checks in a consistent way (i.e. not having _only_ an `isAdmin` while relying on `membeType === xyz` for the other types) is preferred.

I'll add in the other two types, and I can make a separate PR implementing those like I'll do for implementing services :)

frontend-react/src/hooks/UseOktaMemberships.ts
kevinhaube: I might need to refresh my brain on how we're using localStorage, but I don't think having two versions of membership state really serves a benefit, _yet_. Considering there's no real "PrimeAdmins" organization settings, the admins themselves don't benefit from having their default parsed-from-token settings; it's only when they override that they really have any benefit for an admin.

frontend-react/src/hooks/network/ServiceHooks.ts
kevinhaube: Given some thought, we _could_ use React Query for this given the provider swap in the PR, but just not with the AuthFetchProvider tooling (i.e. `rsUseQuery`); we'd still have to do the auth leg work. As for lazy fetching, I'm not seeing the benefit -- it'd make the first load of a service-based feature like Daily Data equally as long as it takes now, wouldn't it?

kevinhaube: After toying around with using `useQuery` for this, it comes with a slew of issues to troubleshoot. I vote we kick that can since this implementation works fine.

frontend-react/src/contexts/SessionContext.tsx
kevinhaube: Actually, to avoid scope creep, I'll alter the TODO here and make a ticket to link to instead

frontend-react/src/hooks/UseOktaMemberships.ts
stephenkao: Fair point.  I was thinking it could be useful for logical separation of the "canonical" user versus an impersonated user, but to that point, we could just use `memberType` for the time being.  I guess I don't really understand why we keep both `global-organization-override` AND `global-membership-state` when the `activeMembership` stuff is kept the same in both -- seems redundant 🤔

frontend-react/src/hooks/network/ServiceHooks.ts
stephenkao: The idea here would be only loading what we need/making requests to the API on demand rather than at user login and treating services as an association rather than conflating them with session logic.  As far as I can tell, we don't need services to be loaded "as part of" the user as soon as they're logged in since they're not directly related to authentication (if we did, we should update the initial login payload to include them).  I don't think these two things need to be tightly coupled together 🤔

kevinhaube: My thought was that showing/hiding navigation UI based on available services was a more generalized approach than segregating experience by `memberType` as we do now; it'd remove the need to infer from an organization name in Okta what features should appear, and mean that any user with an account is a `MemberType.STANDARD` unless they're a `PRIME_ADMIN`. I've never been a fan of our need to parse a name prefix since it creates another nomenclature standard that we need to maintain.

So, in practice, a user is in a single `DHmy-org` group in Okta, and the show/hide of any features depends on whether they are given one or more settings for a type of service. To an extent, including pipeline functionality, _all_ features do require a service, either sender or receiver.

frontend-react/src/hooks/UseOktaMemberships.ts
stephenkao: Sounds good -- I can do some initial investigation on where it's used and ticket the cleanup work, but for now I'd say we can punt on this point!

frontend-react/src/hooks/network/ServiceHooks.ts
stephenkao: Right, but this still seems like we're conflating two separate concepts here: authentication versus associations.  This is how I understand how things are currently laid out (feel free to correct me if I'm wrong here!):

- An end user _is an_ Organization
- An Organization _has many_ services
- A service _has many_ reports, files, et cetera

And there are two types of settings: one at an Organization level and one at a service level.  (A similar analogy I can think of is having different profiles on a social media platform: the user logs in with one set of credentials but can swap between different profiles, each with their own memberships, followers, et cetera.)  I think we should keep these logically separate because conflating them means that we're going to be treating an Organization like a service and vice versa when they really operate on two different levels.  Toggling features based on services isn't a need I'm aware of, and given the recent discussion and implementation behind feature flags, it seems like we'd only want to do this on an Organization level anyway.

Additionally, bypassing react-query here means that we're not going to be able to take advantage of the query cache.  We're going to be making the potentially time-consuming requests to fetch the senders and receivers _every_ time we impersonate a new Organization, even if we had already impersonated that Organization beforehand:
- Log in as admin --> no sender/receiver fetches
- Impersonate `abbott` --> immediate, uncached sender/receiver fetches
- Impersonate `ignore` --> immediate, uncached sender/receiver fetches
- Impersonate `abbott` again --> immediate, uncached sender/receiver fetches
Making these requests lazy react-query queries means that 1) we won't be making immediate requests until the user needs them and 2) we won't be making redundant requests since that data will already exist on the client side.

stephenkao: Also, I agree with you about wanting to not base user ability on the prefix of the group, but we already have working functional precedent on it -- in both the front-end code and in the back-end code (for [scopes](https://github.com/CDCgov/prime-reportstream/blob/master/prime-router/src/main/kotlin/tokens/Scope.kt#L176-L204) and [authentication](https://github.com/CDCgov/prime-reportstream/blob/master/prime-router/src/main/kotlin/tokens/OktaAuthentication.kt#L92-L135)).  If we want to deviate from this pattern, I'd argue that it should be a joint movement on both parts of the stack or we'll be introducing unexpected behavior.

kevinhaube:
@stephenkao Closing this in favor of the "stash after first fetch" approach. Tooling is added via PR #8021 and the follow-on work (because it requires a good bit of refactor) is Issue #8022

PR #7928, opened 2023-01-11
Experience-7891: Disable organization-specific fetches as admin

### LINKED ISSUE #7890: Inform admins of necessary organization mocking on feature pages
## Problem statement

As a member of the `PrimeAdmins` group, when accessing pages like _Submission History_, _Daily Data_, and _Validator_, I'm required to be mocked as an organization with services (i.e. `sender` and `receiver` settings) required for each feature. PrimeAdmins does not have any organizational services supplied from settings, nor does it have organizational settings. **When navigating to one of these pages, the API call should not be triggered at all. Instead, the UI should inform me that I need to mock my organization before using this feature.**

## What we know

PrimeAdmins does not need to be an organization in the settings database table as no data gets transmitted from a PrimeAdmin user through the pipeline. 

Previously, all PrimeAdmin work defaulted to our testing group, `ignore`, but this behavior was removed due to it being confusing for PrimeAdmin users.

We have tools in place that currently perform this behavior, but are reliant on the API call failing rather than stopping it altogether.

We can use `activeMembership.parsedName` to accurately halt on pages requiring mocked membership if the value is set to `PrimeAdmin`

**We should also consider admins mocked as a sender organization accessing a receiver feature.** This may be tricky to parse with current membership data at the session level, but perhaps #7328 will help?

## Acceptance criteria

- [x] No API calls should go out for features requiring a mocked `parsedName` for PrimeAdmins
- [x] No API calls should go out for sender features when an admin is mocked as a receiver
- [x] No API calls should go out for receiver features when an admin is mocked as a sender
- [ ] In the 3 above cases, the UI should inform them which type of org they should be mocked as to access the feature

### LINKED ISSUE #7891: Uncaught not-found exceptions
## Describe the bug
There are six of these just on Jan 8, 2023 alone:

```
Event time	1/8/2023, 10:53:11.431 AM (Local time)	
Message	Uncaught not-found: Request failed with status code 404	
Exception type	Uncaught not-found: Request failed with status code 404	
Failed method	Unknown	
Custom Properties
typeName	String	
activeMembership	{"parsedName":"PrimeAdmins","memberType":"prime-admin"}	
errorSrc	window.onerror@https://reportstream.cdc.gov/static/js/main.7827669e.js:2:1074562	
url	https://reportstream.cdc.gov/static/js/main.7827669e.js	
lineNumber	2	
columnNumber	1074562	
message	Uncaught not-found: Request failed with status code 404

not-found: Request failed with status code 404
    at https://reportstream.cdc.gov/static/js/main.7827669e.js:2:1583989
```

## Impact on ReportStream

We aim for zero exceptions in ReportStream production.

## Steps to reproduce
- Log in as an admin
- Navigate to either /daily-data or /submissions, both of which are pages that will fail on requesting Organization-specific data

stephenkao:
This changeset disables a few Organization-specific requests to mitigate the number of 404s we're getting as admins try to fetch Organization resources:
- Organization settings
- Deliveries
- Submissions

There's not a one-size-fits-all solution here given the different fetch mechanisms we already have and how the data is rendered, so I tried to add minimal changes to prevent disrupting anything down the line.  I think going forward, we can make a generic `useOrganizationQuery` hook (or whatever) that'll automatically be disabled if the user is logged in as an admin without impersonating an Organization.  There are a lot of layers with our fetching that in my opinion should be untangled, but that's out of the scope of this pull request.

Test Steps:
1. Log in as an admin
2. Open the network tab
3. Without impersonating an organization:
  a. Navigate to `/submissions` --> ensure no GET request is made for PrimeAdmins Organization settings or submissions 
  b. Navigate to `/daily-data` --> ensure no GET request is made for for PrimeAdmins Organization settings or deliveries
4. While impersonating an organization:
  a. Navigate to `/submissions` --> ensure GET requests are made for selected Organization settings and submissions
  b. Navigate to `/daily-data` --> ensure GET requests are made for selected Organization settings and deliveries

## Changes

Added a small visual change for the pages to account for this case:

Screenshot 2023-01-18 at 17 09 47
Screenshot 2023-01-18 at 17 09 49

Valid user behavior is unaffected:
Screenshot 2023-01-18 at 17 09 54
Screenshot 2023-01-18 at 17 09 59

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- Fixes #7891 
- Fixes #7890

### DIFF
--- a/frontend-react/src/components/FileHandlers/FileHandlerMessaging.tsx
+++ b/frontend-react/src/components/FileHandlers/FileHandlerMessaging.tsx
@@ -5,10 +5,10 @@ import {
     formattedDateFromTimestamp,
     timeZoneAbbreviated,
 } from "../../utils/DateTimeUtils";
-import { StaticAlert } from "../StaticAlert";
+import { StaticAlert, StaticAlertType } from "../StaticAlert";
 import {
-    ResponseError,
     ErrorCodeTranslation,
+    ResponseError,
 } from "../../config/endpoints/waters";
 import { Destination } from "../../resources/ActionDetailsResource";
 import { USExtLink, USLink } from "../USLink";
@@ -38,7 +38,7 @@ export const FileSuccessDisplay = ({
     return (
         <>
             <StaticAlert
-                type={"success slim"}
+                type={[StaticAlertType.Success, StaticAlertType.Slim]}
                 heading={heading}
                 message={message}
             />
@@ -129,7 +129,10 @@ export const RequestedChangesDisplay = ({
     handlerType,
 }: RequestedChangesDisplayProps) => {
     const alertType = useMemo(
-        () => (title === RequestLevel.WARNING ? "warning" : "error"),
+        () =>
+            title === RequestLevel.WARNING
+                ? StaticAlertType.Warning
+                : StaticAlertType.Error,
         [title]
     );
     const showTable =
@@ -208,7 +211,13 @@ interface FileWarningBannerProps {
 }
 
 export const FileWarningBanner = ({ message }: FileWarningBannerProps) => {
-    return <StaticAlert type={"warning"} heading="Warning" message={message} />;
+    return (
+        <StaticAlert
+            type={StaticAlertType.Warning}
+            heading="Warning"
+            message={message}
+        />
+    );
 };
 
 interface ErrorRowProps {
@@ -251,7 +260,7 @@ export const FileQualityFilterDisplay = ({
     return (
         <>
             <StaticAlert
-                type={"error slim"}
+                type={[StaticAlertType.Error, StaticAlertType.Slim]}
                 heading={heading}
                 message={message}
             />
--- a/frontend-react/src/components/alerts/AdminFetchAlert.tsx
+++ b/frontend-react/src/components/alerts/AdminFetchAlert.tsx
@@ -1,0 +1,13 @@
+import React from "react";
+
+import { StaticAlert, StaticAlertType } from "../StaticAlert";
+
+export default function AdminFetchAlert() {
+    return (
+        <StaticAlert
+            type={StaticAlertType.Error}
+            heading="Cannot fetch Organization data as admin"
+            message="Please try again as an Organization"
+        />
+    );
+}
--- a/frontend-react/src/hooks/UseOrganizationReceiversFeed.ts
+++ b/frontend-react/src/hooks/UseOrganizationReceiversFeed.ts
@@ -10,11 +10,16 @@ interface ReceiverFeeds {
     services: RSReceiver[];
     activeService: RSReceiver | undefined;
     setActiveService: Dispatch<SetStateAction<RSReceiver | undefined>>;
+    isDisabled: boolean;
 }
 /** Fetches a list of receivers for your active organization, and provides a controller to switch
  * between them */
 export const useOrganizationReceiversFeed = (): ReceiverFeeds => {
-    const { data: receivers, isLoading } = useOrganizationReceivers();
+    const {
+        data: receivers,
+        isLoading,
+        fetchStatus,
+    } = useOrganizationReceivers();
     const [active, setActive] = useState<RSReceiver | undefined>();
 
     useEffect(() => {
@@ -29,9 +34,10 @@ export const useOrganizationReceiversFeed = (): ReceiverFeeds => {
     }, [receivers]);
 
     return {
-        loadingServices: isLoading,
+        loadingServices: isLoading && fetchStatus !== "idle",
         services: receivers || [],
         activeService: active,
         setActiveService: setActive,
+        isDisabled: isLoading && fetchStatus === "idle",
     };
 };
--- a/frontend-react/src/hooks/network/History/DeliveryHooks.ts
+++ b/frontend-react/src/hooks/network/History/DeliveryHooks.ts
@@ -1,7 +1,10 @@
 import { useCallback, useMemo } from "react";
 import { AccessToken } from "@okta/okta-auth-js";
 
-import { useAdminSafeOrganizationName } from "../../UseAdminSafeOrganizationName";
+import {
+    Organizations,
+    useAdminSafeOrganizationName,
+} from "../../UseAdminSafeOrganizationName";
 import { useAuthorizedFetch } from "../../../contexts/AuthorizedFetchContext";
 import {
     deliveriesEndpoints,
@@ -65,6 +68,11 @@ const useOrgDeliveries = (service?: string) => {
 
     const fetchResults = useCallback(
         (currentCursor: string, numResults: number) => {
+            // HACK: return empty results if requesting as an admin
+            if (activeMembership?.parsedName === Organizations.PRIMEADMINS) {
+                return Promise.resolve<RSDelivery[]>([]);
+            }
+
             const fetcher = generateFetcher();
             return fetcher(getOrgDeliveries, {
                 segments: {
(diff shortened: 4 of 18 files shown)


frontend-react/src/pages/submissions/SubmissionTable.test.tsx
```diff
@@ -73,4 +74,28 @@ describe("SubmissionTable", () => {
...
+                    service: "",
+                },
+                dispatch: () => {},
+                initialized: true,
+            });
+
+            renderWithResolver(<SubmissionTable />, []);
+        });
+
+        test("renders a warning about not being able to request submission history", async () => {
```
stephenkao, line 93: This test outputs a test warning about state changes because of the [`dispatch` in usePagination](https://github.com/CDCgov/prime-reportstream/blob/master/frontend-react/src/hooks/UsePagination.ts#L397).  I don't think this can be avoided without making further changes to our source code so my feeling is that it's a tolerable warning.  I also feel like the usePagination hook could be replaced with a custom useQuery so we're not having to maintain the event bus -- but this is also outside the scope of this pull request.  I'll add a ticket or two with some thoughts.

stephenkao:
Whoops, still need to fix one more test.

frontend-react/src/hooks/network/History/DeliveryHooks.ts
```diff
@@ -65,6 +68,11 @@ const useOrgDeliveries = (service?: string) => {
 
     const fetchResults = useCallback(
         (currentCursor: string, numResults: number) => {
+            // HACK: return empty results if requesting as an admin
+            if (activeMembership?.parsedName === Organizations.PRIMEADMINS) {
+                return Promise.resolve<RSDelivery[]>([]);
```
kevinhaube, line 73: This will negate the need for [lines 56-58](https://github.com/CDCgov/prime-reportstream/pull/7928/files#diff-57a7939d0bedc4fa4cbe35c6e3cc2a4a0671869ac0aad27cb2df3ef7a5d8aacbR56-R58) which adapted any Admin to the "ignore" org.

Also wondering if there is any way we can generalize this functionality for the future, since services (senders/receivers) will be driven by the organization name. Maybe once that change merges, we could have a `withServices` HOC/HOF to render a fallback UI?

stephenkao: My thought here was to that they're two separate entities -- `PrimeAdmins` with no services and `ignore` as a "legitimate" Organization with services.  If we want to always make `PrimeAdmins` map to `ignore`, then shouldn't we get rid of `PrimeAdmins` entirely and just handle all admin session with `ignore`?  Otherwise, they should be treated as separate entities entirely.  I can update this in any case, but I feel like our logic around this is unintuitive right now 🤔

And yep, that's what I noted in the pull request description:

>  I think going forward, we can make a generic useOrganizationQuery hook (or whatever) that'll automatically be disabled if the user is logged in as an admin without impersonating an Organization. There are a lot of layers with our fetching that in my opinion should be untangled, but that's out of the scope of this pull request.

The changes in this pull request are intended to be minimally intrusive and temporary workarounds while the react-query migration is still underway.  I think adding a HOC at this point would be furthering the workaround logic when we could have an easier solution with a single place where we set `enabled: false` in our react-query queries if an admin isn't impersonating an Organization for Organization-specific requests.  But again, that's dependent on the migration to react-query.  What are your thoughts?

kevinhaube: >If we want to always make PrimeAdmins map to ignore, then shouldn't we get rid of PrimeAdmins entirely and just handle all admin session with ignore?

Oh, my comment is saying the antithesis of this: we _don't ever_ want to map one org to another, admin or otherwise, imo. :) And your change here on line 72 is going to negate any need for that conditional above. What's happening in this, I think, is:

- Sees I'm admin, "ah just change it in the outbound network call to `ignore`"
- Sees I'm admin, "ah don't send any outbound network call"

stephenkao: Ah, got it -- I think we're on the same page then.  If I understand you correctly, we should really remove [this hook](https://github.com/CDCgov/prime-reportstream/blob/master/frontend-react/src/hooks/UseAdminSafeOrganizationName.ts) and any associated/similar logic.  If so, I can update so as to _not_ map to `ignore` here and make a ticket to clean up other places we're using this logic.

frontend-react/src/components/alerts/AdminFetchAlert.tsx
```diff
@@ -0,0 +1,13 @@
+import React from "react";
```
stephenkao, line 1: I feel like it's not really worth it to test this out since it's using StaticAlert logic verbatim 🤷

frontend-react/src/components/FileHandlers/FileHandlerMessaging.tsx
```diff
@@ -38,7 +38,7 @@ export const FileSuccessDisplay = ({
     return (
         <>
             <StaticAlert
-                type={"success slim"}
+                type={[StaticAlertType.Success, StaticAlertType.Slim]}
```
stephenkao, line 41: Added some Enums for StaticAlert `type`s -- why not 🤷 🤷 🤷

frontend-react/src/pages/deliveries/Table/DeliveriesTable.test.tsx
```diff
@@ -107,106 +108,168 @@ describe("DeliveriesTable", () => {
     });
 });
 
-describe("DeliveriesTableWithNumberedPagination - with data", () => {
```
stephenkao, line 110: A lot of this is just reorganizing/scoping tests.  [This one](https://github.com/CDCgov/prime-reportstream/pull/7928/files#diff-27e69cc0f92ac1293b833154b6fe60e745b0bd48c510faecd23188247518eddbR226) is the only net new test here!

frontend-react/src/hooks/UseOrganizationReceiversFeed.ts
```diff
@@ -29,9 +34,10 @@ export const useOrganizationReceiversFeed = (): ReceiverFeeds => {
     }, [receivers]);
 
     return {
-        loadingServices: isLoading,
+        loadingServices: isLoading && fetchStatus !== "idle",
         services: receivers || [],
         activeService: active,
         setActiveService: setActive,
+        isDisabled: isLoading && fetchStatus === "idle",
```
stephenkao, line 41: I really hate that react-query queries don't pass through the `enabled` status as an `isEnabled`.  It feels like such a missed opportunity...especially since we have to grapple with two load states like this 😩

frontend-react/src/network/QueryClients.ts
```diff
@@ -2,7 +2,14 @@ import { QueryClient } from "@tanstack/react-query";
 
 export const appQueryClient = new QueryClient({
     defaultOptions: {
-        queries: { suspense: true, useErrorBoundary: true },
+        queries: {
+            suspense: true,
+            useErrorBoundary: true,
+            retry: false,
```
stephenkao, line 8: No more retries + we're okay with stale queries since our data isn't up-to-the-minute accurate anyway.  The fewer fetches, the better 🙏

kevinhaube: This is the _real_ star of the show, here 😍

frontend-react/src/hooks/UseOrganizationReceiversFeed.ts
kevinhaube: So what you're saying is ... `enabled != isEnabled` but `enabled == !isEnabled` 😂

kevinhaube (approved):
Works for me! Noting that we still hit the case of an Admin mocking as one user type hitting errors for features meant for another user type: out of scope for this change, but worth considering when determining an approach to #7328 (tangential: #8022)

frontend-react/src/hooks/UseOrganizationReceiversFeed.ts
stephenkao: Hahahah EXACTLY 😆

frontend-react/src/network/QueryClients.ts
stephenkao: Seriously!  I haven't been so excited for configuration since......ever 😆

PR #7942, opened 2023-01-12
7580 Bundle FHIR Updates

TurkeyFried:
This PR updates the HL7-FHIR mapping for MSH to Bundle based on spec updates.

The relevant spec is available here: [HL7 Segment - FHIR R4: MSH[Bundle]](https://docs.google.com/spreadsheets/d/1oOy6JW4Mv5VL8DsoA05bYZCUa1JqP4GCYNPF5GJTLxQ/edit#gid=0)

Test Steps:
1. Build this branch and run the unit and integration tests. Verify they all pass. Alternatively you can verify the Github Actions build passes. Note the HL7 to FHIR to HL7 integration test data was not changed, so the result of our HL7 to FHIR to HL7 conversion remains the same = no data changes.
2. Optional: Manually translate a HL7 document, and at the top of the resulting FHIR file, check that the Bundle has an `identifier/system` value and that it matches the text in the PR changes.

## Changes
- Added new `system` value to the identifier for Bundles
- Updated a data sample for reference

## Linked Issues
- #7580


prime-router/src/main/kotlin/fhirengine/translation/HL7toFhirTranslator.kt
```diff
@@ -102,5 +102,6 @@ class HL7toFhirTranslator internal constructor(
         // The HL7 message ID
         val mshSegment = hl7Message["MSH"] as MSH
         bundle.identifier.value = mshSegment.messageControlID.value
+        bundle.identifier.system = "https://reportstream.cdc.gov/prime-router"
```
TurkeyFried, line 105: DISCUSS: What should this _actually_ say?

jimmyfagan: This seems good to me 🤷

prime-router/src/testIntegration/resources/datatests/HL7_to_FHIR/sample_co_1_20220518-0001.fhir
```diff
@@ -1,14 +1,15 @@
 {
   "resourceType" : "Bundle",
-  "id" : "1673379887653974000.2d3730ba-0420-4ead-a33c-878fcecf5941",
+  "id" : "1674153054109185000.9fae9996-785b-428d-a4f8-5d6d0de32cc6",
   "meta" : {
-    "lastUpdated" : "2023-01-10T13:44:47.663-06:00"
+    "lastUpdated" : "2023-01-19T13:30:54.124-05:00"
   },
   "identifier" : {
+    "system" : "https://reportstream.cdc.gov/prime-router",
```
TurkeyFried, line 8: This is the new field

JessicaWNava (commented):
Tested it locally and the change worked. LGTM!

prime-router/src/testIntegration/resources/datatests/HL7_to_FHIR/sample_co_1_20220518-0001.fhir
jimmyfagan: Is this not checked in the tests? I'm just unclear of why this wasn't needed in other `.fhir` files for the tests to pass.

jimmyfagan (approved):
Took me forever to figure out why we're adding this, but just read your comments on the original issue, and this makes sense now. Looks good, glad to see it's so simple! Just left one question, but the tests seem to be passing, so it's probably all good.

PR #7951, opened 2023-01-13
double polling time for smoke test step checking

clediggins-usds:
This PR doubles the polling time for smoke test steps, since staging is inconsistent.


GilmoreA6 (approved):
Approved

PR #7983, opened 2023-01-17
Experience: add proposal for domain-driven directory structure

stephenkao:
This changeset adds a quick proposal for introducing a domain-driven directory structure in frontend-react.  The Experience team already spoke about this in person last week, but in the spirit of documenting larger decisions, we might as well add a proposal for it!

The idea here is to mimic the usage of [prime-router/docs/proposals](https://github.com/CDCgov/prime-reportstream/tree/master/prime-router/docs/proposals), just in frontend-react.


etanb:
I'm very pro this idea! But if we're going to get more granular, I'd love to break out filetypes too. I'm not a huge fan that we combine our `.test` files with their associated files, like in the `/Admin` folder, there are SO many files, that having their associated `.test` files is just cumbersome. I think it would be better if we had separate `/test` and `/storybook` folders so we really have a separation of concerns style structure.

I'm also wondering if we can move things like `alerts`, `headers`, and `tooltips` to a `/shared` folder. Seems like some components get a folder like `Table` and some just live at the root directory so some directives there would be nice for organization.

stephenkao:
> I'm very pro this idea! But if we're going to get more granular, I'd love to break out filetypes too. I'm not a huge fan that we combine our `.test` files with their associated files, like in the `/Admin` folder, there are SO many files, that having their associated `.test` files is just cumbersome. I think it would be better if we had separate `/test` and `/storybook` folders so we really have a separation of concerns style structure.

The colocation of the component files and their tests/styles/stories/et cetera was something we discussed [here](https://github.com/CDCgov/prime-reportstream/pull/7827#discussion_r1065899389).  I personally don't really mind what approach we take as long as the file structure is predictable, scalable (potentially for nested domains), and overall organized by functionally related code so @jpandersen87 @penny-lischer @kevinhaube, do you have any thoughts about this?

> I'm also wondering if we can move things like `alerts`, `headers`, and `tooltips` to a `/shared` folder. Seems like some components get a folder like `Table` and some just live at the root directory so some directives there would be nice for organization.

Yep!  That's what I was thinking about for the [`common` domain](https://github.com/CDCgov/prime-reportstream/pull/7983/files#diff-0bb6054758d1cf78f83b7e22559669449ee057c1479d7aac2adf621c569a3b07R115), but it might also be nice to have a separate `styleguide` domain specifically for custom components that we create.  The latter may be overkill, but it could be good to have that separation of concerns there -- like "these are components specifically deviating from Trussworks/USWDS defaults" 🤷 Let's go with a common/core/shared for now and break it out later if needed!

jpandersen87:
Having related files (stories, test, etc.) co-located with the entity in question (ex: component) was my personal bias on not having to navigate to other folders in the file browser when working on them. However, if the majority of us work better with them filtered into their own dedicated folders then that's perfectly fine too.

kevinhaube:
I'm fairly indifferent as far as test files: on one hand, it's _another_ directory to maintain the structure of, but on the other, it keeps working directories a lot cleaner when scrolling. I tend to be a keyboard-shortcut-to-find-a-file-and-open-it type person. That said, I'm a fan of consolidating standards, so if we need to break out Storybook into `/stories` I think we should follow suit and use Jest's `__tests__` and `__mocks__` directory model as well.

etanb:
_If_ we do keep stores, tests etc with their components, then I'd advocate for a more granular folder structure so like instead of `/Admin` having all these components, each individual component, like `AdminFormEdit.tsx`, could then have it's own folder, which would then have all associated files so:

```
/AdminFormEdit
../AdminFormEdit.tsx
../AdminFormEdit.scss
../AdminFormEdit.test.tsx
../AdminFormEdit.stories.tsx
```

jpandersen87:
Agreed. I seem to have misread the proposal and am only now realizing it didn't include the folder per entity concept. That's on me. The colocation idea only works with dedicated entity folders.

frontend-react/docs/proposals/0002-domain-driven-directory-structure.md
```diff
@@ -0,0 +1,148 @@
...
+- `decorators`: all domain-specific React decorators
+- `utils`: all domain-specific utilities, tests, and endpoint resources
+
+A typical structure will look like the following:
+
+```
+src/
+  <domain>/
+    components/
+      ComponentA.tsx
```
kevinhaube, line 75: One gripe I have with this is that it feels like it'll step on code reuse. When I consider React as a framework, I think of `components` and `hooks` being two unique "tools" to build with, not specific to any single domain on the site, like a page or feature. Consider building an IKEA shelf vs hanging a picture frame. Both require a hammer, one requires nails, and one requires dowels (that IKEA always so kindly gives you one less of than is required, but I digress). Shelves and pictures can live anywhere in a house, though...do we own a "living room hammer" and a "bedroom hammer"? No, because that's silly!

The nature of React's tools is reusability, and it gets hard to isolate where something is and whether it's utility applies elsewhere when it's domain-specific. Taking that hammer analogy one step further, there's many types of hammers to do many jobs, but the core of each hammer is the same: handle, head. If we abstracted things from each domain into an "abstract" domain, and build from each of those, we run the risk of over-abstraction (at least in my experience). 

I'd propose something more akin to this:

```
src/
    components/
        ... (sub directories as they make sense)
    hooks/
        ... (sub directories as they make sense)
    app/
        <domain>/
            DomainPage.tsx
            ...
``` 

This would still maintain these core tools of React as reusable pieces of code rather than one domain borrowing from another domain, and give us the ability to further consider what _types_ of "hammers" we need, so to speak, while keeping domain-specific code as the least reusable part of the front-end codebase, like a page that composes hooks and components in such a way that is _only_ meant for use at a specific route.

kevinhaube: And as a quick follow up, for an example: consider what'd happen if we locked hooks into a domain, but that domain - say, an unused page/feature - no longer gets used. When it comes time to delete that domain from the repository, that subdirectory is tied into other parts of the codebase if we've reused HookA or ComponentB somewhere in the application other than that directory.

That's mainly what I mean when I said (then edited out) "it steps on code reusability". I changed the wording to fit my point above better to "it sends the message that reusability isn't our goal", but in this example, I think "steps on" is much more apt.

stephenkao:
> Agreed. I seem to have misread the proposal and am only now realizing it didn't include the folder per entity concept. That's on me. The colocation idea only works with dedicated entity folders. The main entry point (ex the component file) can be index.tsx for less verbosity on import.

So correct me if I'm wrong, but you're both proposing something like the following:

```
<domain_name>/
  components/
    AdminFormEdit/
      index.tsx
      AdminFormEdit.test.tsx
      AdminFormEdit.scss
      AdminFormEdit.stories.tsx
```

That's fine with me as long as the entity folders don't contain any nested entities:

```
<domain_name>/
  components/
    AdminFormEdit/
      index.tsx
      AdminFormEdit.test.tsx
      AdminFormEdit.scss
      AdminFormEdit.stories.tsx
      AdminFormButton/
        index.tsx
        AdminFormButton.test.tsx
        AdminFormButton.scss
        AdminFormButton.stories.tsx    
```

In this case, I'd argue that the button should be brought to the same depth in `components`:

```
<domain_name>/
  components/
    AdminFormEdit/
      index.tsx
      AdminFormEdit.test.tsx
      AdminFormEdit.scss
      AdminFormEdit.stories.tsx
    AdminFormButton/
      index.tsx
      AdminFormButton.test.tsx
      AdminFormButton.scss
      AdminFormButton.stories.tsx    
```

The reason for this would at least be twofold: the directory structure is recursively deterministic (so there's no question on when to add a new directory), and it keeps the directory structure as flat as possible to surface components that may be generalizable across the domain.  What do y'all think?

jpandersen87:
Absolutely agree that a component folder should be the end of a nesting structure (aka flat).

frontend-react/docs/proposals/0002-domain-driven-directory-structure.md
stephenkao: I don't think anyone's arguing against reusability in our codebase, just that the way that our files are organized right now isn't going to do us any favors as we continue building in it.  One of the benefits for a domain-driven folder structure is for the ease of colocation and maintenance in understanding that a file in any particular domain is _likely_ to used in or affect another file in the same domain.

Domains are intended to operate as separate modules, yes, but outside of the `common`/`core`/`shared` domain, there realistically should be minimal reuse between domains; anything generic or atomic enough should be elevated to said domain.  Working in a larger React app means that we're going to be building a lot of "unique" components/hooks frequently, but if they're all built on top of the same atomic parts, they're just going to be consumers of the same underlying logic rather than producing anything novel.  (Just look at  `Admin/`, `header/`, and `MessageTracker/` as an example of this.)  And if we find ourselves repeating the same patterns over and over again, then in my opinion that'll warrant a cleanup ticket to escalate into a `common`/`core`/`shared` file.

My feeling here is that real reusability isn't something that can be enforced through a directory structure; there's nothing stopping engineers from reinventing the wheel ad nauseam.  (This is how we ended up with the NPM community.)  Reusability is something we need to keep top of mind as we're building out our core component system and making the code's capabilities known to design/product during planning.  Instead, the directory structure can help to build a mental model of where certain logic lives and to enforce better patterns around encapsulation.

kevinhaube: >Domains are intended to operate as separate modules, yes, but outside of the common/core/shared domain, there realistically should be minimal reuse between domains

Ahh, I think I misunderstood domains as "partitioned areas of the site" not "partitioned areas of functionality" from the way the file structure was shaped. Common, core, and shared domains are exactly what I'd considered the "as it makes sense" subdirectories of hooks and components. 

>My feeling here is that real reusability isn't something that can be enforced through a directory structure; there's nothing stopping engineers from reinventing the wheel ad nauseam.

Agreed, though my statement was leaning more towards the philosophical approach; sort of a question of "what are we saying _without_ words when we display code segmented by domain-specific partitions (again, under the assumption that a "common/core/shared" domain didn't exist in this file structure). I think those non-verbal comms are what help newcomers to a project see our intent.

stephenkao: Yeah, exactly!  Domains are more about consolidation for evergreen concepts like authentication, Organizations, Senders/Receivers, admin stuff...things that won't be going away anytime soon unless ReportStream dramatically changes its mission statement -- at which point...we may just want to start an entirely new codebase 😆

I'll update the proposal to be a little clearer about this!

jpandersen87:
I'm fine with the folder concept applying across the board but we can keep it limited to components as well.

etanb:
I think things being conceptually scoped, ie all these files are for the `<Table/>` component, makes sense to me, and anything shared (ie generalizable + usable in another place within the code), whether it's a hook or a `<Tooltip/>` component, living in a higher level folder. 

IE, I share your concern of being "overly verbose" and making sure things are directly symmetrical folder-wise vs. something flatter.

kevinhaube (approved):
My vote lies with domain-driven given the info/discussion in this PR!

PR #7995, opened 2023-01-18
Session Analytics

etanb:
This PR tracks a user's session length (in seconds) and upon idle or unload (closing browser in some form), sends that session length to Azure. 

A user session, as agreed upon with Chris, is the following:

- When a user opens ReportStream
- Ignores logged in / logged out state
- If a user swaps to another tab OR minimizes, pause the session
- Only when a user closes the browser OR idles long enough that our `useIdleTimer` kicks in, is a tracking event sent to Azure

Test Steps:
1. Since this is a draft PR, I left the envars in `frontend-react/.env.development` but you NEED to have `REACT_APP_APPLICATIONINSIGHTS_CONNECTION_STRING` set correctly to see events
2. Tool around on the site, and then if you add a `debugger` statement in `AppRouter` in the `onUnload` function, you can see the event as you close the page
3. You can also see the aggregate time, and how the time pausing works, if you just add a `console.log` of `sessionTimeAggregate` and see how it increases / pauses 

## Changes
The biggest set of changes is in `AppRouter` as we are adding two `eventListeners`, one on `visibilitychange` and one on `beforeunload`. I had to add this code to `AppRouter` as all the Context variables we're utilizing live in the `SessionContext` and the `AppRouter` is the first larger component that is instantiated AFTER `SessionContext` is.

### Testing
- [ ] Tested locally?
- [ ] Added tests?

## Linked Issues
https://app.zenhub.com/workspaces/experience-607d9d5e68b95200150fec37/issues/gh/cdcgov/prime-reportstream/7627

## To Be Done
- #issue


etanb:
@stephenkao @penny-lischer would love y'all help getting this over the finish line.

The final steps here are structuring the session data correctly to be consumed by Azure. The even I'm firing looks like so:

```
trackAppInsightEvent("session", {
    sessionLength: sessionTimeAggregate,
});
```

The key is `session` and the value is the time in seconds. I can round the value if need be. Is this correct / sufficient for our Azure dashboards needs? If so, would love help with actually creating the dashboard in Azure!

frontend-react/.env.development
```diff
@@ -16,4 +16,7 @@ REACT_APP_CLIENT_ENV=development
 REACT_APP_ENV=development
 
 INLINE_RUNTIME_CHUNK=false
-GENERATE_SOURCEMAP=true
\ No newline at end of file
+GENERATE_SOURCEMAP=true
+
+; App Insights connection
+REACT_APP_APPLICATIONINSIGHTS_CONNECTION_STRING=InstrumentationKey=7ce107ea-c24a-49e2-a05f-2aaeb2e82f29;IngestionEndpoint=https://eastus-8.in.applicationinsights.azure.com/;LiveEndpoint=https://eastus.livediagnostics.monitor.azure.com/
```
stephenkao, line 22: Let's unstage this change :)

stephenkao:
> @stephenkao @penny-lischer would love y'all help getting this over the finish line.
> 
> The final steps here are structuring the session data correctly to be consumed by Azure. The even I'm firing looks like so:
> 
> ```
> trackAppInsightEvent("session", {
>     sessionLength: sessionTimeAggregate,
> });
> ```
> 
> The key is `session` and the value is the time in seconds. I can round the value if need be. Is this correct / sufficient for our Azure dashboards needs? If so, would love help with actually creating the dashboard in Azure!

I think minutes would actually be better since it's unlikely for a typical user flow to be less than one minute, but that's more of a question for @chris-kuryak.  @penny-lischer would definitely be your go-to for dashboard work since she's been owning that (and actually has access to Azure 😆)

frontend-react/src/AppRouter.tsx
```diff
@@ -41,6 +43,50 @@ export enum FeatureName {
...
+    const {
+        sessionStartTime,
+        setSessionStartTime,
+        sessionTimeAggregate,
+        setSessionTimeAggregate,
+    } = useSessionContext();
+
+    useEffect(() => {
+        const onUnload = () => {
+            trackAppInsightEvent("session", {
```
stephenkao, line 55: Can we add [an enum](https://github.com/CDCgov/prime-reportstream/blob/master/frontend-react/src/utils/Analytics.ts#L3) for this event name?

frontend-react/src/AppRouter.tsx
```diff
@@ -41,6 +43,50 @@ export enum FeatureName {
...
+        window.addEventListener("visibilitychange", onVisibilityChange);
+        return () => {
+            // We MUST remove the event listeners as the dependency values
+            // update or else we'll have 100s of event listeners on the page
+            // all with stale values
+            window.removeEventListener("beforeunload", onUnload);
+            window.removeEventListener("visibilitychange", onVisibilityChange);
+        };
+    }, [
+        sessionStartTime,
```
stephenkao, line 85: Why does this effect need to rerun whenever these change?  I think we could just make it a one-time* run to bind these effects on mount with `[]`, no?

* Not exactly a one-time run with React v18 strict mode, but it should be good enough for us since we're not there yet 😆

frontend-react/src/contexts/SessionContext.tsx
```diff
@@ -51,17 +60,33 @@ const SessionProvider = ({
...
+        });
+    };
+
+    const updateSessionTimeAggregate = (timeInSeconds: number) => {
+        setState((prevState) => {
+            return { ...prevState, sessionTimeAggregate: timeInSeconds };
+        });
+    };
+
+    const [state, setState] = useState<RSSessionContext>({
```
stephenkao, line 75: Hmmm I don't really think this needs to exist in SessionContext.  Every time the state changes here (including the time updates) the reference to `state` will update, which can cause a full rerender of the entire SessionContext.

I think `visibilityState` stuff can just exist in App.tsx since that's the top-level component since `visibilityState` events only need to be bound in one place for a client-side app.  Or if we want to split it out, we can make it a separate hook 🤷

frontend-react/src/AppRouter.tsx
penny-lischer: Yes, please add to Analytics.ts -> EventName

etanb:
> I think minutes would actually be better since it's unlikely for a typical user flow to be less than one minute, but that's more of a question for @chris-kuryak. @penny-lischer would definitely be your go-to for dashboard work since she's been owning that (and actually has access to Azure 😆)

Chris and I chatted about this, seconds are the preferred unit here.

etanb:
After spending time with @penny-lischer on Azure dashboards, and finally gaining access myself, everything in this PR is now setup and reporting correctly. Here's the sample query in Azure:

```
let usg_events = dynamic(["*"]);
let mainTable = union customEvents
    | where timestamp > ago(1h)
    | where isempty(operation_SyntheticSource) //filtering out HTTP requests made by bots
    | extend name =replace("\n", "", name)
    | extend name =replace("\r", "", name)
    | where '*' in (usg_events) or name in (usg_events);
let queryTable = mainTable;
let splitTable =  () {
    queryTable
    | extend timeInSeconds = todynamic(tostring(customDimensions["sessionLength"]))
    | evaluate bag_unpack(timeInSeconds)
    | order by timestamp desc
};
splitTable
```

And in Azure, you can see the event represented:

Screen Shot 2023-01-19 at 11 04 16 AM

frontend-react/src/AppRouter.tsx
```diff
@@ -41,6 +43,52 @@ export enum FeatureName {
...
+        setSessionStartTime,
+        sessionTimeAggregate,
+        setSessionTimeAggregate,
+    } = useSessionContext();
+
+    useEffect(() => {
+        const onUnload = () => {
+            trackAppInsightEvent(EventName.SESSION, {
+                sessionLength:
+                    (new Date().getTime() - sessionStartTime.getTime()) / 1000 +
```
penny-lischer, line 57: Can we create a function to calculate the sessionLength since its used in multiple places?

frontend-react/src/AppRouter.tsx
stephenkao: Bumping this question 🐢

frontend-react/src/contexts/SessionContext.tsx
stephenkao: Bumping this point 🐢

frontend-react/src/AppRouter.tsx
etanb: Test comment!

stephenkao: I think the issue here is that we're expecting a state change to flush out a re-render, but nothing visual is actually updating hence the stale values.  For this purpose, I'd recommend using `ref`s instead:

```ts
export const AppRouter = () => {
    // I think it's easier to treat these as `getTime()` return values (milliseconds)
    // so 1) there's a consistent unit and 2) we can perform basic math operations on them
    // and then convert to seconds (x / 1000) when we actually need to submit to AI
    const sessionStartTime = useRef<number>(new Date().getTime());
    const sessionTimeAggregate = useRef<number>(0);

    useEffect(() => {
        const onUnload = () => {
            trackAppInsightEvent(EventName.SESSION, {
                sessionLength:
                    (new Date().getTime() -
                        sessionStartTime.current +
                        sessionTimeAggregate.current) /
                    1000,
            });
        };

        const onVisibilityChange = () => {
            if (document.visibilityState === "hidden") {
                sessionTimeAggregate.current =
                    new Date().getTime() -
                    sessionStartTime.current +
                    sessionTimeAggregate.current;
            } else if (document.visibilityState === "visible") {
                sessionStartTime.current = new Date().getTime();
            }
        };

        // bind + unbind stuff here
    }, []);

    //...otherStuff
};
```

This is untested, but I think that should work since refs exist outside of the state reconciler!

frontend-react/src/contexts/SessionContext.tsx
stephenkao: My feeling is that we can continue to treat SessionContext as the place where we maintain auth state, but session _timing_ is conceptually separate from that -- like I don't think we should have to update all of the tests in this changeset to accommodate the updated SessionContext value 🤔 I think if we want to keep it encapsulated, we could break it out into a separate hook like `useSessionTimings` that'd 1) maintain the start time and session time sum, 2) return functions to start/pause/reset the session time, 3) bind to the document visibility state on render.

What are your thoughts about this?

frontend-react/src/utils/Analytics.ts
```diff
@@ -2,6 +2,7 @@ import { getAppInsights } from "../TelemetryService";
 
 export enum EventName {
     TABLE_FILTER = "Table Filter",
+    SESSION = "Session",
```
stephenkao, line 5: VERY minor point, but should this be `SESSION_DURATION` + `"Session Duration"`?  We may want to add other session-related events in the future.

frontend-react/src/contexts/SessionContext.tsx
etanb: Chatted offline, but decided to remove all the work within `SessionContext` and just have a single instance of `sessionTime` exist within `App.tsx` so we don't have to share any variables across the app.

penny-lischer (approved):
LGTM

stephenkao (approved):
Sorry for the delay!  Looks good to me as well -- thanks for circling back on the feedback! ❤️

etanb:
With final approval of sample Azure Dashboard from @chris-kuryak which is saved in my personal Dashboards within Azure

Screen Shot 2023-01-25 at 1 58 09 PM

PR #8028, opened 2023-01-20
RSEndpoints refinement and new react-query hooks that get auth withou…

### LINKED ISSUE #8027: RSEndpoints refinement and new react-query hooks that get auth without context
As a developer I would like to rely on our RSEndpoint class more for interactions with the RS API. Also, not being reliant on react contexts to get current auth information (at this moment, okta tokens) would allow for easier coding of static functions that interact with the API as we would not be bound by the react lifecycle.

jpandersen87:
RSEndpoints refinement and new react-query hooks that get auth without context.

Fixes #8027


jpandersen87:
Closing in favor of alternative PR.

PR #8030, opened 2023-01-21
Update package.json to support upgrade to Typescript 4.9

### LINKED ISSUE #8029: Update package.json to support upgrade to Typescript 4.9
Typescript 4.9 was released recently and with it comes support with a new keyword: "satisfies". Our dependency react-scripts allows us to freely use any 4.X version, however we also need to update other dependencies to support this new keyword (babel, eslint, and prettier) that are normally controlled by react-scripts. This will require overriding react-scripts via yarn's "resolutions" property.

jpandersen87:
This PR declares the necessary dependencies under "resolutions" to allow full use of typescript 4.9 features. Also has some cleanup on dependencies. Those that should have been put under dev have been moved. Redundant declarations (seemingly handled primarily by another dependency at the requested version we had) removed.

Fixes #8029


frontend-react/package.json
```diff
@@ -13,23 +13,12 @@
         "@rest-hooks/rest": "^3.0.3",
         "@tanstack/react-query": "^4.20.9",
         "@tanstack/react-query-devtools": "^4.20.9",
-        "@testing-library/user-event": "^13.5.0",
```
stephenkao, line 16: Thanks for moving these!

stephenkao (approved):
Looks great to me!  Thanks for the cleanup along the way!

penny-lischer (approved):
LGTM

PR #8057, opened 2023-01-24
7776 - Create Events to track success of Public File Validation

penny-lischer:
This PR ...

Tracks how the File Validator is used as well as captures errors and warnings.

Test Steps:
1. Log into RS
2. Select an Federal Organization
3. Navigate to Admin -> Validate 
4. Upload a .csv or HL7 file
5. Check the logs in Azure to verify that the following were recorded in the customEvent table:
warningCount
errorCount
fileType
selected schema
sender

### Testing
- [x] Tested locally?
- [x] Added tests?


stephenkao (approved):
Easy!  Looks great to me!

stephenkao (changes requested):
Oh, actually!  I think the track will fire on every re-render -- one sec

frontend-react/src/components/FileHandlers/FileHandler.tsx
```diff
@@ -250,6 +251,19 @@ const FileHandler = ({
         (reportId || overallStatus === OverallStatus.VALID) &&
         !hasQualityFilterMessages;
 
+    // Once file is submitted, send to analytics
+    if (submitted && !isWorking) {
```
stephenkao, line 255: Can we move this block to after [the `sendFile` above](https://github.com/CDCgov/prime-reportstream/pull/8057/files#diff-90daad087bcd5ce4dac7318290253991cf7355a69bdc37793105cdb8b08e7ae1R182)?  It'd ensure we'd only track the event once rather than on every render when `submitted && !isWorking`.

penny-lischer: @stephenkao  It would need to be after submitted. The submitted is not declared until after the sendFile. Options?

stephenkao: `sendFile` from useWatersUploader [actually submits it to the `/validate` endpoint](https://github.com/CDCgov/prime-reportstream/blob/master/frontend-react/src/hooks/network/WatersHooks.ts#L25-L52), right?  `submitted` there is just a derived value that happens after a re-render, but we'll have the response from the `async sendFile` call:

```js
const response =  await sendFile(...);

// can access response.errorCount, response.warningCount, et cetera
```

This may actually need to be wrapped in a `try-catch` since I'm pretty sure invalid validations come back as an error code rather than a 200, though!

penny-lischer: @stephenkao good catch!

frontend-react/src/components/FileHandlers/FileHandler.tsx
```diff
@@ -137,6 +137,18 @@ const FileHandler = ({
         validateOnly
     );
 
+    if ((reportId || errors.length || overallStatus) && !isWorking) {
```
stephenkao, line 140: Ah!  So what I mean was something more this:

```ts
const handleSubmit = async (event: React.FormEvent<HTMLFormElement>) => {
    // other stuff    
    dispatch({ type: FileHandlerActionType.PREPARE_FOR_REQUEST });
   

    let eventData;

    try {
        const response = await sendFile({
            contentType: contentType,
            fileContent: fileContent,
            fileName: fileName,
            client: client,
        });

        eventData = {
            warningCount: response?.warnings?.length,
            errorCount: response?.errors?.length,
        };
    } catch (e: any) {
        if (e.data) {
            eventData = {
                warningCount: e.data.warningCount,
                errorCount: e.data.errorCount,
            };
        }
    }

    if (eventData) {
        trackAppInsightEvent(EventName.FILE_VALIDATOR, {
            schema: sender?.schemaName,
            fileType: fileType,
            sender: organization?.name,
            ...eventData,
        });
    }
};
```

The reason for this is that it's better to invoke functions we don't intend to run on every render in an event handler (`onChange`, `onClick`, et cetera).  There are of course some exceptions to this, but this runs the risk of firing `trackAppInsightEvent` on every render when `(reportId || errors.length || overallStatus) && !isWorking)` is true, even if there wasn't a state change here.

penny-lischer: Thanks @stephenkao . This is great.

frontend-react/src/components/FileHandlers/FileHandler.tsx
```diff
@@ -178,12 +179,40 @@ const FileHandler = ({
...
+        } catch (e: any) {
+            if (e.data) {
+                eventData = {
+                    warningCount: e.data.warningCount,
+                    errorCount: e.data.errorCount,
+                };
+            }
+        }
+
+        if (eventData) {
```
stephenkao, line 206: Ah, sorry -- I think we might only want to track this event when `validateOnly` is true too.  Otherwise, we'll be firing it when the user's actually uploading to send rather than just validating.  We'd probably want to track that too, but that's a question for @chris-kuryak.  Will follow up on the ticket!

frontend-react/src/components/FileHandlers/FileHandler.test.tsx
```diff
@@ -577,5 +618,130 @@ describe("FileHandler", () => {
                 type: FileHandlerActionType.RESET,
             });
         });
+
+        describe("trackEvent", () => {
+            test("gets called on submit if FileHandlerType is type VALIDATION", async () => {
+                mockUseSenderResource.mockReturnValue({
```
stephenkao, line 624: Some of this setup stuff could be pulled into a common function, like `renderComponent`:

```js
function renderComponent(fileHandlerType) {
    mockUserSenderResource.mockReturnValue(...);
    // ...other stuff
    renderWithFullAppContext(
        <FileHandler
            handlerType={fileHandlerType}
            /* other props */
         />
    );
}
```

Then just use that for each of the tests:

```js
describe("trackEvent", () => {
    test("gets called on submit if FileHandlerType is type VALIDATION", async () => {
         renderComponent(FileHandlerType.VALIDATION);

         // run validation expectations
    });

    test("gets called on submit if FileHandlerType is type UPLOAD", async () => {
         renderComponent(FileHandlerType.UPLOAD);

         // run upload/send expectations
    });
});
```

I also generally think it's better to keep test setup in `beforeEach`/`beforeAll` hooks, but definitely not a blocker.  None of this is a blocker at all, actually 😆

stephenkao (approved):
One more tiny comment, but this looks good to me regardless!  Thank you!

PR #8066, opened 2023-01-25
React Component ADR

etanb:
This is a proposal for how we approach building React components in the front-end of ReportStream. This _is not_ calling into question USWDS as a basis for our components, but how to work with UWSWDS, Trussworks' React library and fully custom components going forwards.


frontend-react/docs/proposals/0003-uswds-react-components.md
```diff
@@ -0,0 +1,251 @@
...
+
+## Problem Statement
+
+- We rely on [USWDS components](https://designsystem.digital.gov/components/overview/) for the basis of much of the code in ReportStream. This creates a rigidity in our codebase since our designs require extensions on their library and contributing to their codebase has proved a cumbersome and untimely process.
+- On top of the base USWDS library, we also use [Trussworks React component library](https://github.com/trussworks/react-uswds). This adds another level of rigidity to the stack since we aren't interacting with the base USWDS library, we're only exposed to whatever `props` trussworks exposes for us.
+- From-the-ground-up customized components while offering complete control, could move us too far off the standardized, accessible components offered by the USWDS + would be quite a lot of work to create and maintain.
+
+## Decision Drivers
+
+- Recent experience adding a simple addition to USWDS showed that we can't rely on them to update their library according to our needs.
```
stephenkao, line 17: Something related to these points is that Trussworks's MO is to provide very thin, almost logic-less wrappers around the basic USWDS markup; they specifically don't want to deviate too far from the basic markup.  This forces anyone who wants to support more complex UX to either finagle a lot of logic to wrap/consume their components or just create their own custom components.

frontend-react/docs/proposals/0003-uswds-react-components.md
```diff
@@ -0,0 +1,251 @@
...
+- Extremely actively maintained with an associated [Trussworks Storybook](https://trussworks.github.io/react-uswds/) which would make diffing our extended components very easy.
+- Follows our currently implemented solutions of extending components, so least overhead in terms of code changes.
+- Ideal for stylistic changes to Trussworks React components. Can simply use the scoped `.module` syntax to create new dynamic styles.
+- No need to add styles to our convoluted overrides file `_uswds_overrides.scss` anymore.
+
+#### Cons
+- Not ideal for complex functional extensions of React components.
+- Can get convoluted quickly since we're essentially wrapping a wrapper.
+- Potentially easy breaking changes as we'd be two levels from USWDS.
+- Still need to manually override USWDS styles on occasion using `!important`.
```
stephenkao, line 49: Related but not dependent on our decision in this ADR: I think we should avoid `!important`s as much as possible and instead try to find a scalable pattern that'll allow us to override USWDS defaults through specificity rather than the `!important` steamroller.  Basically, my feeling is that we should be keeping specificity as low as possible so styles can be more easily overridden downstream.

frontend-react/docs/proposals/0003-uswds-react-components.md
```diff
@@ -0,0 +1,251 @@
...
+<a name="option-two"></a>
+### Option 2: Utilize USWDS Base Library Only
+
+#### Description
+
+Utilize the raw [USWDS library](https://github.com/uswds/uswds) which would remove a level of abstraction to our components. We'd then wrap their components and create a React component that interacts directly with their code which will allow us full customization of props and functionality. CSS will still need to be overridden at times, but can be component-scoped.
+
+#### Pros
+
+- 1:1 relationship with USWDS which means we can upgrade our dependencies without having to rely on Trussworks upgrading theirs.
```
stephenkao, line 89: 🙏

frontend-react/docs/proposals/0003-uswds-react-components.md
```diff
@@ -0,0 +1,251 @@
...
+    // with all USWDS CSS classes instead
+    const USWDSTableClasses = {
+      borderless: 'usa-table--borderless',
+      compact: 'usa-table--compact',
+      striped: 'usa-table--striped',
+      stickyHeader: 'rs-table--sticky-header',
+    }
+    
+    // The classnames library recursively flattens arrays
+    const classes = classnames("usa-table", Object.keys(props).map((class) => USWDSTableClasses[class]));
```
stephenkao, line 115: I know this is a hypothetical component, but I feel like it'd be clearer to pass this as an object to `classnames`:

```jsx
const classes = classnames("usa-table", {
    "usa-table--borderless": borderless,
    "usa-table--compact": compact,
    ...
});
```

This'd still be usable if we had a hash or enum of available styles provided by the USWDS too:

```jsx
const classes = classnames("usa-table", {
    [UswdsStyle.Borderless]: borderless,
    [UswdsStyle.Compact]: compact,
    ...
});
```

We could also play around with the concept of `variant`s as well.  But just a thought! 🤷

frontend-react/docs/proposals/0003-uswds-react-components.md
```diff
@@ -0,0 +1,251 @@
...
+Many of the USWDS components are not very complicated from a Design, animation and overall front-end point of view. Re-creating them from the ground up, especially potentially very complicated components like ReportStream's proposed new tables, would give us granular control and better testing.
+
+#### Pros
+
+- Complete control over presentation and functionality of our components.
+- Can set the standard for other government applications and how they create accessible, usable and beautiful web components for their users.
+- Zero dependencies on external teams to fix broken code or add certain features.
+
+#### Cons
+- Have to maintain, top-to-bottom, the codebase including: testing, accessibility, optimization and everything else that comes with maintaining a library of sorts.
```
stephenkao, line 150: And having to maintain documentation!  😪

frontend-react/docs/proposals/0003-uswds-react-components.md
```diff
@@ -0,0 +1,251 @@
...
+```js
+import styles from "./RSTable.module.css";
+import classnames from "classnames";
+
+export const RSTable = ({ styles, dataHeader, dataBody }) => {
+    // The classnames library recursively flattens arrays
+    const classes = classnames("rs-table", styles);
+
+    return (
+      <div className={classes}>
```
stephenkao, line 166: I know it may not always be applicable, but I think we should still try to align with specific tags like `<table>`, `<button>`, et cetera for custom components as much as possible -- or if we can't, we should opt for using `aria-role`.  Maintaining HTML semantics is still important for accessibility/assistive technology 🙏

frontend-react/docs/proposals/0003-uswds-react-components.md
```diff
@@ -0,0 +1,251 @@
...
+
+- [ ] My component's fundamentals are based on the USWDS design system.
+- [ ] My component deviates from a USWDS-based component in deeper ways that include functionality and style.
+- [ ] My component can be represented by extending a single or sewing together multiple Trussworks USWDS React component(s).
+
+**[Option 3 (Greenstart Customized Components)](#option-three):**
+
+- [ ] My component's fundamentals are *not* based on the USWDS design system.
+- [ ] My component is very complicated and *cannot* be represented by extending multiple Trussworks USWDS React components.
+- [ ] My Engineering, Product and Design teams understand that a fundamentally new component will be added to our ecosystem.
```
stephenkao, line 236: 🙏

frontend-react/docs/proposals/0003-uswds-react-components.md
etanb: I totally agree. I would _much_ prefer if we could just use CSS specificity to overwrite any required CSS changes. I guess I was just going showcasing the worst case scenario (and should've been more explicit about that) if we had a direct collision and needed to use `!important`. Can add a summary of this conversation to the ADR.

frontend-react/docs/proposals/0003-uswds-react-components.md
etanb: I'm all for being explicit! So you're saying to forego the iteration with `Object.keys(props)` and just being explicit with the `USWDSTableClasses` <-> 'className` mapping? Totally cool with that.

frontend-react/docs/proposals/0003-uswds-react-components.md
etanb: Totally! Will add a mention.

frontend-react/docs/proposals/0003-uswds-react-components.md
etanb: Yeah, good point. I just chose a not-obviously-a-table component to show that the developer, in this example, made a unique choice. I can make a note here about this conversation, where if a developer is going the fully custom React component they would have to do exactly that -- justify their decision between `<table/>` and using `aria-role`.

frontend-react/docs/proposals/0003-uswds-react-components.md
stephenkao: I think I remember reading an issue in which they said this flat out -- will do some digging to find it again if it exists!  Otherwise, they have a [Non-Goals section](https://github.com/trussworks/react-uswds#non-goals) that basically says that the components are purely visual as opposed to having any more functionality 🤔

frontend-react/docs/proposals/0003-uswds-react-components.md
stephenkao: Makes sense to me!

frontend-react/docs/proposals/0003-uswds-react-components.md
stephenkao: Yeah, I think just being explicit helps out here because it's slightly less mental overhead to process :)

stephenkao:
This looks good to me @etanb!  Is there anything holding it back in draft status or is it ready for approval?  I don't think we need to reach a formal decision or anything -- like you state in the doc, there's no one-size-fits-all solution to this, so it could just be a reference point for patterns going forward.  What are your thoughts?

etanb:
@stephenkao Yup, we're all good to go! Just giving people ample opportunity to comment on the draft stage. Would love to use this as a blueprint going forwards for the proposed rebuild of the frontend codebase.

frontend-react/docs/proposals/0003-uswds-react-components.md
```diff
@@ -0,0 +1,257 @@
+# [0003: Creating New React Components]
+
+* Status: [proposed | rejected | accepted | deprecated | superseded by [0005](0005-example.md)]
```
stephenkao, line 3: I think we can change this to `accepted` since we're basically sticking with our existing approach, just with a little more formality :)

stephenkao (approved):
Looks good to me!  Thank for adding this!

PR #8074, opened 2023-01-25
Add ordering_provider_name_type_code and testing_lab_name_type_code to sender schemas

brick-green:
This PR adds two fields to the sender schemas: ordering_provider_name_type_code and testing_lab_name_type_code. The purpose of these fields is to add values to OBR-16.10 and OBX-23.2 respectively. These fields were requested by PA as part of their onboarding but will be sent with all STLT reporting with this PR. Both fields are type codes and should only be present when their primary fields(OBR-16.1 and OBX-23.1) have values.

Test Steps:
1. Pull an HL7 and CSV file for testing
     - sample HL7 file: `./prime sender-files --env prod --message-id 20230124100151`
     - sample CSV file: `./prime sender-files --env prod --message-id C80402368X+33+20230125`
2. add settings for PA: [pa-phd.txt](https://github.com/CDCgov/prime-reportstream/files/10513907/pa-phd.txt)
3. run curl to submit test files to docker
4. check output files for OBR-16.10 and OBX-23.2. Both values should be "L"

### Testing
- [X] Tested locally?


oslynn (approved):

oslynn:
Looks good, @brick-green

prime-router/metadata/schemas/upload-covid-19.schema
```diff
@@ -358,6 +359,8 @@ elements:
       cardinality: ONE 
       csvFields: [{name: testing_lab_name}]
 
+    - name: testing_lab_name_type_code
```
GilmoreA6, line 362: It definitely doesn't hurt to have this here but I think as long as the element is present in the COVID-19 schema or whatever schema the receiver is using you get the same result whether this is here or not

prime-router/metadata/schemas/upload-covid-19.schema
```diff
@@ -334,17 +334,18 @@ elements:
...
-      cardinality: ONE 
+      cardinality: ONE
       csvFields: [{name: ordering_provider_zip_code}]
 
     - name: ordering_provider_phone_number
-      cardinality: ONE 
+      cardinality: ONE
       csvFields: [{name: ordering_provider_phone_number}]
 
+    - name: ordering_provider_name_type_code
```
GilmoreA6, line 348: Same as below.

adegolier (approved):
Already bothered Brick with the questions I had, from my unfamiliarity. Seems fine.

prime-router/metadata/schemas/upload-covid-19.schema
adegolier: If this is true, I think there's some documentation that needs to change somewhere because I swear I read that "based on" requires you to specify elements you want to include.

GilmoreA6 (approved):
LGTM

PR #8090, opened 2023-01-26
6528 utils

jpandersen87:
This PR adds the changes to utils, api config, mocks, etc. for upcoming PRs for hook updates.


frontend-react/src/TelemetryService.ts
```diff
@@ -61,7 +61,11 @@ export function getAppInsights() {
     return appInsights;
 }
 
-export function getAppInsightsHeaders(): { [key: string]: string } {
+export interface AppInsightsHeaders {
```
stephenkao, line 64: 🙏

frontend-react/src/config/api/QueryClients.ts
```diff
@@ -0,0 +1,11 @@
+import { QueryClient } from "@tanstack/react-query";
+
+export const appQueryClient = new QueryClient({
```
stephenkao, line 3: Just wanted to make sure [these changes](https://github.com/CDCgov/prime-reportstream/blob/master/frontend-react/src/network/QueryClients.ts#L3) are captured here!

frontend-react/src/config/api/types.ts
```diff
@@ -0,0 +1,475 @@
...
+ */
+export enum ErrorCodeTranslation {
+    INVALID_HL7_MSG_VALIDATION = "Invalid entry for field.",
+    INVALID_MSG_PARSE_DATE = "Invalid entry for field. Reformat to either the HL7 v2.4 TS or ISO 8601 standard format.",
+    INVALID_HL7_MSG_FORMAT_INVALID = "",
+    INVALID_MSG_PARSE_TELEPHONE = "The string supplied is not a valid phone number. Reformat to a 10-digit phone number (e.g. (555) 555-5555).",
+    UNKNOWN = "",
+}
+
+declare global {
```
stephenkao, line 23: Minor point, but it might be nice to spread these types spread out across a few files just to make additions and maintenance a bit easier.  We could still declare them as global types from said files -- my feeling is that they can just be grouped around concepts/domains.  Definitely not a blocker, though!  Just thinking aloud about future ideas 🤔

jpandersen87: The plugin is only going to be able to export a single file so this was done in an attempt to get adjusted to the idea.

frontend-react/src/config/api/RSEndpoint.ts
```diff
@@ -0,0 +1,296 @@
...
+
+export type RSEndpointFetchers<T extends RSEndpoint<any>> = {
+    [P in string & keyof T["meta"]["methods"] as Lowercase<P>]: (
+        args: Partial<Omit<RSOptionsWithSegments<T>, "method">>
+    ) => P extends HTTPMethod
+        ? Promise<RSEndpointMethodRequestType<T, P>>
+        : never;
+};
+
+export class RSEndpoint<E extends EndpointConfig = any> {
```
stephenkao, line 150: Hmmm!  I've been mulling this over for a bit so forgive the wall of text that's inevitably incoming.

I'm wary about building more logic on top of RSEndpoint; this is starting to conflate the singular purpose of 1) what configuration is needed to make the request with 2) actually making the request -- and I'd argue that should be much more in the purview of react-query specifically.  I realize that the `queryFn` is likely intended to be used in `useQuery` calls as the fetcher, but I don't think we need to be adding fetcher logic to RSEndpoint.

My feeling is that there are a few concerns that can be handled separately:
- Generating request configuration, which includes fetch options (HTTP method, path interpolation) and react-query options (query key and query key prefixes, default values)
- Constructing the fetcher, which is just a function that calls Axios given a subset of the above
-  Wrapping `useQuery` and `useMutation` to accept custom parameters

For instance:

```js
// given a string url potentially with dynamic segments, return a string interpolated from a params object
// can also distinguish segment params versus query params
function getFullUrl(url, params) { ... }

// given request options, make a request via Axios
function createFetch({ 
    url, 
    method, // HTTP request method
    body, // optional body
    headers, // optional headers
    params  // optional query parameters
}) { ... }

// custom wrapper around useQuery given optional query config
function useResourceQuery({
    params,
    url,
    config, // optional resource-specific query options
    headers, // optional headers if needed on a per-endpoint level
    method, // can be omitted or defaulted to `GET` -- useQuery can be used for simple POSTs but unlikely in our usage (https://github.com/TanStack/query/discussions/1107)
}) { 
    // also may need to handle query parameters, depending on how getDynamicUrl works
    const fullUrl = getDynamicUrl(url, params);
    
    const query = useQuery({
        queryKey: [url, params],
        queryFn: () => createFetch({ url: fullUrl, method, headers }),
        config
    });

    return {
        ...query,
        // it'd also be nice to return some important things like queryKey and config.enable that aren't returned by default for whatever reason
        queryKey,
        isEnabled: config.enabled,
        ...
    };
}
```

We could also wrap useResourceQuery with another hook that'd handle authentication if needed:

```jsx
function useAuthenticatedResourceQuery({ ...otherStuff }) {
    const bearerToken = ...; // can get from LocalStorage or from SessionContext

    return useResourceQuery({ 
        ...otherStuff, 
        headers: { authorization: "Bearer blahblahblah" }
    });
}
```

And this composability would be useful to us in case if, say, we wanted to disable certain queries based on auth status:

```jsx
function useSenderOnlyResourceQuery({ ...otherStuff, config }) {
    const isOnlyPrimeAdminAndNotImpersonatingSender = ...; // can get from LocalStorage or from SessionContext

    return useResourceQuery({
        ...otherStuff,
        config: { ...config, enabled: !isOnlyPrimeAdminAndNotImpersonatingSender }
    });
}
```

I honestly don't think this is too far off from what's here; it's just a different, flatter approach with more emphasis on invoking functions as needed to accommodate flexible requirements rather than needing an upfront configuration.  However, I think it's important for us to keep our patterns more focused around the idea of composition to 1) play nicely with the very composition-forward React but also 2) to have better separation of concerns for more flexibility and easier debugging, testing, implementation, et cetera.

---

As an aside, I'm also wondering how useful RSEndpoint really is to us.  If we have a custom `useQuery` and `useMutation`, we should allow them to accept custom parameters that we can run the path interpolation against as another first-class function -- in which case, we could really downgrade RSEndpoint into simple functions that return Axios config options 🤷  However, this is a separate conversation from the above!

--- 

ANYWAY, my brain is completely shot today so I'm going to stop there.  I'm happy to chat more about this tomorrow if you'd like!

frontend-react/src/config/api/types.ts
stephenkao: Ah, I see I see!  No worries then :)

frontend-react/src/config/api/RSEndpoint.ts
```diff
@@ -0,0 +1,296 @@
...
+    PATCH: "PATCH",
+} as const;
+export type HTTPMethod = (typeof HTTPMethods)[keyof typeof HTTPMethods];
+
+/**
+ * Copied from react-router for internal use with Endpoints.
+ * Please use react-router's types directly for react-router
+ * work.
+ */
+export type ParamParseSegment<Segment extends string> =
```
stephenkao, line 77: I'm a bit wary of the fact that we're copying this from react-router source code.  It'd be one thing if this (and the path interpolation) were explicitly exported by react-router, but having to maintain this type separately feels like unnecessary maintenance on our end.  (It also looks like [there's been some simplification](https://github.com/remix-run/react-router/pull/9090](https://github.com/remix-run/react-router/blob/main/packages/router/utils.ts#L218)) in the latest version?)

Regarding the path interpolation, I wonder if it'd just be simpler to have all of the paths be functions that, at the time of request, could take in certain parameters:

```js
new RSEndpoint({
    path(pathArg1, pathArg2, pathArg3, params) {
       const basePath = `/api/${arg1}/${arg2}/${arg3}`;
       if (params) {
         return `${basePath}?${transformQueryParams(params)}`;
       } 

       return basePath;
    },
    ...otherStuff
});
```

`toDynamicUrl` is interesting logic, but my feeling is that it's more complicated than it needs to be because 1) we really only need an exploded string path returned and 2) missing segments can just be enforced through argument types to something like the `path` method above.  It's also specifically around dynamic path components, not dynamic query parameters -- so why not make everything dynamic?  

What are your thoughts about this?

frontend-react/src/components/MessageTracker/WarningsErrors.tsx
```diff
@@ -1,6 +1,5 @@
 import React from "react";
 
-import { WarningErrorDetail } from "../../config/endpoints/messageTracker";
```
penny-lischer, line 3: If we remove this the `WarningsErrorsDisplayProps` will break

jpandersen87: All of the API data types were moved to one file that is globally declared for the project and do not need to be imported.

frontend-react/src/components/MessageTracker/QualityFilters.tsx
```diff
@@ -1,7 +1,5 @@
 import React from "react";
 
-import { QualityFilter } from "../../config/endpoints/messageTracker";
```
penny-lischer, line 3: If we remove this won't the `QualityFiltersDisplayProps` break?

frontend-react/src/config/api/RSEndpoint.test.ts
```diff
@@ -0,0 +1,179 @@
...
+    // const axiosInstance = jest.spyOn(Axios,"default");
+    // const warnInstance = jest.spyOn(console, "warn");
+
+    const tokens = {
+        accessToken: mockSenderAccessToken as unknown as RecursiveMutable<
+            typeof mockSenderAccessToken
+        >,
+    };
+    const { activeMembership } = getMembershipsFromToken(tokens.accessToken);
+    // const mockResponse = Promise.resolve({foo: "bar"}) as any;
```
penny-lischer, line 139: Were these comments left by mistake?

frontend-react/src/config/api/RSEndpoint.ts
penny-lischer: Agree with Stephen, can this be simplified?

Issue #8012, opened 2023-01-18
Migrate to okta-auth-js v7 (EOL 2023-09-30)


stephenkao:
## Problem statement

We're on v6 of [okta-auth-js](https://github.com/okta/okta-auth-js), and it has an [end of life](https://github.com/okta/okta-auth-js#release-status) scheduled for 2023-09-30.  There won't be a hard shut-off date, but there will no longer be active maintenance for v6 after that date.

## What you need to know

v7 is stable, but it'll require more investigation to see how it'll change our usage and how it'll affect the [React integration](https://github.com/okta/okta-react).

## Acceptance criteria

- [ ] Okta is upgraded to v7 and continues to work

stephenkao:
cc @jpandersen87

PR #8116, opened 2023-01-31
Migrate to okta-auth-js v7 (EOL 2023-09-30) #8012

jpandersen87:
This PR updates our Okta dependencies to their latest versions and implements the minimum required changes to work with our existing code. The latest version of okta-auth-js has built-in support for syncing auth across tabs and thus our in-house solution is no longer needed (it also handles token refreshes at a single window source to prevent flooding okta with refresh requests). OF IMPORTANT NOTE: I have to use the "useClassicEngine" flag for the sign-in widget. Classic engine is deprecated and marked for removal. @chris-kuryak this documentation will be relevant on migrating to their new engine: [Migrating to OIE](https://developer.okta.com/docs/guides/migrate-to-oie/).

Closes #8038


frontend-react/src/App.tsx
```diff
@@ -24,56 +22,48 @@ import { ErrorPage } from "./pages/error/ErrorPage";
...
-const OKTA_AUTH = new OktaAuth(oktaAuthConfig);
+import { trackAppInsightEvent, EventName } from "./utils/Analytics";
+import { getOktaAuth } from "./oktaConfig";
+import { logout } from "./utils/UserUtils";
 
 const { APP_ENV } = config;
+const OKTA_AUTH = getOktaAuth();
 
-initializeSessionBroadcastChannel(OKTA_AUTH); // for cross-tab login/logout
+let sessionStartTime = new Date().getTime();
```
etanb, line 32: Can we defer on these changes until we come to a consensus as a team with Chris and co?

chris-kuryak: Yes, sounds like we could use a conversation on this so I can get aligned. If we don't get to it in 16th min, perhaps the Exp Eng sync on Thurs? @stephenkao

frontend-react/package.json
```diff
@@ -86,7 +86,8 @@
             "[/\\\\]node_modules[/\\\\](?!(react-markdown|vfile|vfile-message|markdown-table|unist-.*|unified|bail|is-plain-obj|trough|remark-.*|rehype-.*|html-void-elements|hast-util-.*|zwitch|hast-to-hyperscript|hastscript|web-namespaces|mdast-util-.*|escape-string-regexp|micromark.*|decode-named-character-reference|character-entities|property-information|hast-util-whitespace|space-separated-tokens|comma-separated-tokens|pretty-bytes|ccount|mdast-util-gfm|gemoji)).+\\.(js|jsx|mjs|cjs|ts|tsx)$"
         ],
         "moduleNameMapper": {
-            "axios": "axios/dist/node/axios.cjs"
+            "axios": "axios/dist/node/axios.cjs",
+            "\\.(css|less)$": "<rootDir>/src/__mocks__/styleMock.js"
```
stephenkao, line 90: Let's add `scss` in here as well -- we'll probably need it!

Also, TINY POINT, but I'm assuming we can use this as a general catch-all for any static asset mock (like .jpg, .png, .md, et cetera) so we might want to just name this something like `staticAssetMock.js` instead.  Not a huge deal, though 🤷

frontend-react/src/oktaConfig.ts
```diff
@@ -11,12 +12,15 @@ const oktaAuthConfig: OktaAuthOptions = {
...
+        storage: APP_ENV === "test" ? "memory" : "localStorage",
     },
     scopes: ["openid", "email"],
+    services: {
+        autoRenew: false,
+    },
 };
 
-const oktaSignInConfig = {
+const oktaSignInConfig: WidgetOptions = {
```
stephenkao, line 23: Thanks for typing this!

frontend-react/src/pages/Login.tsx
```diff
@@ -2,22 +2,35 @@ import React from "react";
...
+import { Navigate } from "react-router-dom";
 
 import OktaSignInWidget from "../components/OktaSignInWidget";
 import { oktaSignInConfig } from "../oktaConfig";
 import { useSessionContext } from "../contexts/SessionContext";
 import { MembershipActionType } from "../hooks/UseOktaMemberships";
 import { USLink } from "../components/USLink";
-import { getSessionBroadcastChannel, SessionEvent } from "../utils/UserUtils";
+
+const MonitoringAlert = () => {
```
stephenkao, line 13: Ah, thanks for pulling this out! 🙏

frontend-react/src/utils/UserUtils.ts
```diff
@@ -1,47 +1,19 @@
...
-            });
+            await oktaAuth.signOut();
         } catch (e) {
             console.trace(e);
         }
     }
     localStorage.clear();
 }
 
-export function initializeSessionBroadcastChannel(oktaAuth: OktaAuth) {
```
stephenkao, line 25: 🔥 🔥 🔥

stephenkao:
Blech, I'm getting an uncaught error in a Promise on logout:

Screenshot 2023-02-01 at 17 07 04

I think it's due to [this lovely block](https://github.com/CDCgov/prime-reportstream/blob/master/frontend-react/src/components/header/SignInOrUser.tsx#L15-L21) that's spamming Okta just to get the email.  But I think we could just replace it with:

```js
const email =  oktaAuth?.authStateManager?.getAuthState()?.idToken?.claims?.email;
```

No need to make a request for data we already have 🤷 

I think you're already on [8141](https://app.zenhub.com/workspaces/experience-607d9d5e68b95200150fec37/issues/gh/cdcgov/prime-reportstream/8141) so we might be able to kill two birds with one stone with this fix!

stephenkao:
[One small bug](https://github.com/CDCgov/prime-reportstream/pull/8116#issuecomment-1412806015), but this looks good to me once that's address!  Thanks for hopping on this so quickly!

etanb:
@stephenkao is the issue of [this block you mentioned](https://github.com/CDCgov/prime-reportstream/blob/master/frontend-react/src/components/header/SignInOrUser.tsx#L15-L21), that it's being called multiple times?

```
useEffect(() => {
        if (authState?.isAuthenticated) {
            oktaAuth
                .getUser()
                .then((cl) => setUser(cl.email ? cl.email : "unknown user"));
        }
    });
```

If we add an empty dependency array `[]`, it'll only be called once instead of 3 times (that I saw locally):

```
useEffect(() => {
        if (authState?.isAuthenticated) {
            oktaAuth
                .getUser()
                .then((cl) => setUser(cl.email ? cl.email : "unknown user"));
        }
    }, []);
```

jpandersen87:
We're gonna pause this PR and get this one merged first which handles the user profile in a better way: https://github.com/CDCgov/prime-reportstream/pull/8159

frontend-react/yarn.lock
```diff
@@ -1857,30 +1855,33 @@
...
     "@types/jqueryui" "^1.12.16"
     "@types/q" "^1.5.5"
     "@types/selectize" "^0.12.35"
     "@types/underscore" "^1.11.4"
+    chokidar "^3.5.1"
     clipboard "^1.5.16"
     cross-fetch "^3.1.5"
+    ejs "^3.1.7"
     handlebars "^4.7.7"
+    jquery "^3.6.0"
```
stephenkao, line 1884: Blech, I'm not a fan of this, but it's a necessary hit for the signin widget.  Looks like there's [an open issue](https://github.com/okta/okta-signin-widget/issues/512) to remove it, but that was from almost three years ago 🤔 

Obviously not a blocker for this since we're just upgrading to get off the EOL version.  Something we can keep an eye on for future enhancements when we have more tooling around front-end performance.

stephenkao (approved):
Tested locally and looks good to me!

PR #8122, opened 2023-01-31
Experience-7948: Update validate API endpoint to allow Senders to specify schemas

### LINKED ISSUE #7948: Update the validate API call from the Validation Tool
## Problem statement
Given that the validation tool will allow users to select between schemas and Sender settings, we need to update our front-end call to the validate function to include their choice.

## What you need to know
- The validation API endpoint [is going to be updated](https://app.zenhub.com/workspaces/platform-638a6fe2aacec620dafdac38/issues/gh/cdcgov/prime-reportstream/7927) to accept either 1) one of the three master schemas (HL7, CSV, CSV OTC) or 2) a Sender setting.  
- useWatersUploader already factors in a [client header](https://github.com/CDCgov/prime-reportstream/blob/master/frontend-react/src/hooks/network/WatersHooks.ts#L39) for the Organization name.  This will be updated on the backend to allow for a chosen schema as well.

## Acceptance criteria
- [ ] [useWatersUploader](https://github.com/CDCgov/prime-reportstream/blob/master/frontend-react/src/hooks/network/WatersHooks.ts#L19) for `validate` should be updated to pass the chosen schema or Sender settings as part of the header, which will be informed by [7773](https://app.zenhub.com/workspaces/experience-607d9d5e68b95200150fec37/issues/gh/cdcgov/prime-reportstream/7773)

stephenkao:
This changeset updates the `validate` endpoint to allow Senders to specify schemas.  The use case here is that we want to let authenticated Senders validate against their predefined schemas but also to validate against one of the three master schemas (CSV, CSV OTC, HL7) that will be provided as options on the front end.  This required a bit of reworking around when to use the real Sender versus the dummy Sender:

Use a dummy Sender when:
- `client` is not provided
- `client`, `schema`, and `format` are all provided but don't match up with Sender settings

Use a real Sender when:
- only `client` is provided (backwards compatibility with current functionality)
- `client`, `schema`, and `format` are all provided and match up with Sender settings

The reason for this is that we're still abiding by Sender-specific settings during validation (specifically around [`allowDuplicates`](https://github.com/CDCgov/prime-reportstream/blob/master/prime-router/src/main/kotlin/azure/ValidateFunction.kt#L100)), and we need to pull in the real Sender settings for that.  Otherwise, there'd be different error responses between this flow and the actual upload to send, which would be poor UX and would kind of defeat the point of this feature.  I also considered two other options to account for this, but they wouldn't work out as far as I know:

- Always use the dummy Sender.  This would simplify a lot of logic, but it'd bypass Sender-specific settings needed during validation.
- Pull in a Sender by the specified schema.  I don't think would actually be the right move because we're not indexing on `values.schema`/`values.format` on the `setting` table so we can't guarantee that we'd only get one unique Sender per pair.  Plus, there'd be potential performance implications since there isn't an index there -- not a huge deal at the moment since we have <3000 records in the `setting` table now (I think), but still not a scalable solution.

If anyone has any suggestions, I'm all ears!

To test this out:
- Call the endpoint with matching `client`, `schema`, and `format` --> ensure that the validation is running against the real Sender (can flip `allowDuplicates` as a way to check)
- Call the endpoint with only `client` --> ensure that the validation is running against the real Sender
- Call the endpoint with mismatched `client`, `schema`, and `format` --> ensure that the validation is running against the dummy Sender
- Call the endpoint with only `schema` and `format` --> ensure that the validation is running against the dummy Sender

### Testing
- [x] Tested locally?
<strike>- [ ] (For Changes to /frontend-react/...) Ran `npm run lint:write`? </strike>
- [x] Added tests?

## Linked Issues
- Fixes #7948

### DIFF
--- a/prime-router/src/main/kotlin/azure/RequestFunction.kt
+++ b/prime-router/src/main/kotlin/azure/RequestFunction.kt
@@ -181,7 +181,7 @@ abstract class RequestFunction(
      */
     @Throws(InvalidParameterException::class)
     internal fun getDummySender(schemaName: String?, formatName: String?): TopicSender {
-        val errMsgPrefix = "No client found in header so expected valid " +
+        val errMsgPrefix = "Expected valid " +
             "'$SCHEMA_PARAMETER' and '$FORMAT_PARAMETER' query parameters but found error: "
         if (schemaName != null && formatName != null) {
             val schema = workflowEngine.metadata.findSchema(schemaName)
@@ -200,7 +200,7 @@ abstract class RequestFunction(
                 schema.topic
             )
         } else {
-            throw InvalidParameterException("$errMsgPrefix 'SchemaName' and 'format' parameters must not be null")
+            throw InvalidParameterException("$errMsgPrefix 'schema' and 'format' parameters must not be null")
         }
     }
 }
--- a/prime-router/src/main/kotlin/azure/ValidateFunction.kt
+++ b/prime-router/src/main/kotlin/azure/ValidateFunction.kt
@@ -74,24 +74,22 @@ class ValidateFunction(
                 }
             }
 
-            if (sender == null) {
-                sender =
-                    try {
-                        getDummySender(schema, format)
-                    } catch (e: InvalidParameterException) {
-                        return HttpUtilities.bad(request, e.message.toString())
-                    }
-            }
+            sender = sender ?: getDummySender(schema, format)
 
             actionHistory.trackActionParams(request)
-            processRequest(request, sender as Sender)
+            processRequest(request, sender)
         } catch (ex: Exception) {
-            if (ex.message != null) {
-                logger.error(ex.message!!, ex)
-            } else {
-                logger.error(ex)
+            when (ex) {
+                is InvalidParameterException -> HttpUtilities.bad(request, ex.message.toString())
+                else -> {
+                    if (ex.message != null) {
+                        logger.error(ex.message!!, ex)
+                    } else {
+                        logger.error(ex)
+                    }
+                    HttpUtilities.internalErrorResponse(request)
+                }
             }
-            HttpUtilities.internalErrorResponse(request)
         }
     }
 
--- a/prime-router/src/main/kotlin/azure/ValidateFunction.kt
+++ b/prime-router/src/main/kotlin/azure/ValidateFunction.kt
@@ -14,6 +14,7 @@ import gov.cdc.prime.router.ActionLog
 import gov.cdc.prime.router.ActionLogLevel
 import gov.cdc.prime.router.InvalidReportMessage
 import gov.cdc.prime.router.Sender
+import gov.cdc.prime.router.TopicSender
 import gov.cdc.prime.router.Translator
 import gov.cdc.prime.router.ValidationReceiver
 import gov.cdc.prime.router.azure.db.enums.TaskAction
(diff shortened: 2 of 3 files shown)


prime-router/src/test/kotlin/azure/ValidateFunctionTests.kt
```diff
@@ -192,6 +206,155 @@ class ValidateFunctionTests {
         verify(exactly = 0) { validateFunc.processRequest(any(), any()) }
     }
 
+    @Test
+    fun `test validate endpoint only client`() {
+        val metadata = UnitTestUtils.simpleMetadata
```
stephenkao, line 211: I definitely realize this is continuing some redundancy/repeat logic between test cases, so I'm happy to clean this up if it's a blocker.  Just let me know!

prime-router/src/main/kotlin/azure/RequestFunction.kt
```diff
@@ -181,7 +181,7 @@ abstract class RequestFunction(
      */
     @Throws(InvalidParameterException::class)
     internal fun getDummySender(schemaName: String?, formatName: String?): TopicSender {
-        val errMsgPrefix = "No client found in header so expected valid " +
```
stephenkao, line 184: Small clarifications of the error messages now that we have different use cases to account for 🤷

prime-router/src/main/kotlin/azure/ValidateFunction.kt
```diff
@@ -50,23 +51,40 @@ class ValidateFunction(
...
+                    (
+                        (clientSender as TopicSender).schemaName == schema &&
+                            Sender.Format.valueOf(format) == clientSender.format
+                        )
+                ) {
+                    sender = clientSender
                 }
             }
+
+            if (sender == null) {
```
arnejduranovic, line 77: One thing I thought of here is maybe it's more idiomatic/cleaner if we replace this whole block with just `sender = sender ?: getDummySender(schema, format)` and then in the general catch at the bottom you can do a `WHEN` expression on the error type?

```
 catch (ex: Exception) {
            when (ex) {
                is InvalidParameterException -> HttpUtilities.bad(request, ex.message.toString())
                else -> {
                    if (ex.message != null) {
                        logger.error(ex.message!!, ex)
                    } else {
                        logger.error(ex)
                    }
                    HttpUtilities.internalErrorResponse(request)
                }
            }
        }
```
Only update if you think its worthwhile :)

prime-router/src/main/kotlin/azure/ValidateFunction.kt
```diff
@@ -50,23 +51,40 @@ class ValidateFunction(
...
-                        request.queryParameters.getOrDefault(SCHEMA_PARAMETER, null),
-                        request.queryParameters.getOrDefault(FORMAT_PARAMETER, null)
-                    )
-                } catch (e: InvalidParameterException) {
-                    return HttpUtilities.bad(request, e.message.toString())
+
+                // only use client Sender in two cases
+                // 1) only client is sent (not schema or format) - backwards compatibility
+                // 2) client is sent and matches up with provided schema and format to use Sender settings
+                if (
```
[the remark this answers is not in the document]
stephenkao: Ah, good point!  My thought process here was about the separation of the schema itself versus the Sender; using the dummy Sender in this case would just be validated against the schema without incorporating any Sender settings like `allowDuplicates` -- or any other settings we could potentially factor in down the line.  If we wanted to incorporate Sender settings, I think we should change it and remove the `client` header to just pull in the appropriate Sender by `schema` + `format` without using a dummy Sender at all.  But I don't think we can/should guarantee uniqueness by `schema` + `format` on the `setting` table 🤔

prime-router/src/main/kotlin/azure/ValidateFunction.kt
stephenkao: Oooh fantastic!  Will definitely update --thank you!

prime-router/src/main/kotlin/azure/ValidateFunction.kt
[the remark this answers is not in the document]
arnejduranovic: That makes sense to me. I won't push for any other changes here.

arnejduranovic (approved):

stephenkao:
Closing because we're going to be handling this on the client side now 🙏

Issue #8141, opened 2023-01-31
Reduce the number of `userinfo` calls to Okta


stephenkao:
## Describe the bug
We're making a lot of `userinfo` calls to Okta, which looks to be caused by repetitive calls on re-renders in the [SignInOrUser component](https://github.com/CDCgov/prime-reportstream/blob/master/frontend-react/src/components/header/SignInOrUser.tsx#L15-L21) (happens on _every_ render because there's no dependency array).  This calls `getUser`, an alias to `token.getUserInfo`, which appears [to make this call](https://github.com/okta/okta-auth-js/blob/a57a689fedf4484d694865f7a0c04b14837f9bf3/lib/oidc/getUserInfo.ts#L39).

## Impact on ReportStream
No user impact at the moment, but this could potentially result in our getting throttled by or locked out of the Okta API if we start to have more active users.

## Steps to reproduce
1. Log in through Okta
2. Open the Network tab
3. Navigate around the site
4. See a bunch of `userinfo` calls on each re-render

## Expected behavior
We should only make one `userinfo` call upon auth change and memoize it.

PR #8159, opened 2023-02-01
Reduce the number of userinfo calls to Okta #8141

jpandersen87:
This PR moves the logic of requesting the okta user profile from the SignInOrUser component (located in upper-right corner of page UI) to the SessionContext component. On okta changes, if authenticated, the component will request the latest version of the user's profile and set the "user" property appropriately in the context (or will set to undefined if auth is lost). This update also memo's the context value for SessionContext (depending on all of it's properties so as to update appropriately). The makeOktaHook function has had a bug fix to create a static object instead of always creating a new object in it's wrapper function (stopping an infinite loop in react components in testing).


frontend-react/src/contexts/SessionContext.tsx
```diff
@@ -48,17 +55,40 @@ const SessionProvider = ({
     const isAdminStrictCheck = useMemo(() => {
         return activeMembership?.memberType === MemberType.PRIME_ADMIN;
     }, [activeMembership?.memberType]);
+    const [user, setUser] = useState<UserClaims<CustomUserClaims>>();
+
+    useEffect(() => {
```
stephenkao, line 60: Rather than making a fetch inside an effect, could we just subscribe to an auth state change?  With the global `OKTA_AUTH` instance getting exported now, we have access to the AuthStateManager:

```tsx
const [authState, setAuthState] = useState(
    OKTA_AUTH.authStateManager.getAuthState()
);

...other stuff

function updateAuthState(_authState: AuthState) {
    setAuthState(_authState);
}

useEffect(() => {
    OKTA_AUTH.authStateManager.subscribe(updateAuthState);

    return () => OKTA_AUTH.authStateManager.unsubscribe(updateAuthState);
}, []);

const context = useMemo(
    () => ({
        oktaToken: authState?.accessToken,
        activeMembership,
        isAdminStrictCheck,
        dispatch,
        initialized: authState !== null && !!initialized,
        user: authState?.idToken?.claims,
    }),
    [
        activeMembership,
        authState,
        dispatch,
        initialized,
        isAdminStrictCheck
    ]
);
```

This should work for logged in --> logged out and logged out --> logged in state changes.  We already have synchronous access to the user claims after an auth state change, so I don't think we should need to make a subsequent request 🤔

jpandersen87: That asks the question: do we foresee needing OIDC extended user profile information in the future? If we do, we need the getUser call as that hits the OIDC userinfo endpoint. If we don't, we can rely strictly on the claims in the ID token. Right now the userinfo endpoint seems to only be returning sub, email, and email_verified properties. I think only the email_verified claim is not given in the original ID token.

stephenkao: That's a great point!  My feeling is that we likely wouldn't need anything other than an email -- at the moment, we only care about the `claims.organization` since a user is acting on behalf of an Organization.  I can't imagine there being any features being specifically around a user rather than an Organization on the website...plus, just skimming through the users on the production Okta, we don't have any of that information anyway 😬 

But ultimately, I think this would be a better question for @chris-kuryak!  Do you foresee our needing to access any extended user information from Okta at some point in the future?  [Here](https://developer.okta.com/docs/reference/api/oidc/#userinfo)'s an example of what kind of information we could get.  However, it seems like we'd have to change our onboarding process to have users fill that extended information in or we'd have to collect it ourselves since it doesn't look like we're doing that now.

jpandersen87: If we do decide that we need OIDC user profiles, I'll likely update this PR to create a hook that handles the logic of when to call getHook that the SessionProvider component can use.

stephenkao: That sounds like a plan to me.  In any case, I don't think this is a blocker -- just a suggestion for simplification if we can afford it!

frontend-react/src/utils/UserUtils.ts
```diff
@@ -9,16 +11,13 @@ export enum SessionEvent {
...
-function logout(oktaAuth: OktaAuth): void {
-    if (oktaAuth?.signOut) {
-        try {
-            oktaAuth.signOut().finally(() => {
-                getSessionBroadcastChannel()?.postMessage(SessionEvent.LOGOUT);
-            });
-        } catch (e) {
-            console.trace(e);
-        }
+async function logout(): Promise<void> {
```
stephenkao, line 14: 🙏

stephenkao (approved):
Just that one point about not making a separate request to `/userinfo` if we don't need, but that's not a blocker.  I'd say we can go with this solution as-is since it's preserving the same functionality as it exists today, just with substantially fewer requests to the Okta API :)

Thanks so much for digging into this!

frontend-react/src/contexts/SessionContext.tsx
chris-kuryak: @stephenkao Good question. Yes, currently, I think you are correct that email is all we need for Okta. However, know we do collect more information on our users in SmartSheet, which is essentially our CRM. And in there, we have information around name, location, etc. So I am unsure if we'd need to keep the information in both places, and I think this actually may be a question best answered with O+O team's input.

However, let me ask a couple follow-up questions:
1) How permanent of a decision is this? If we decide we don't need to access the other user information, is this a change we could make in the future to start accessing more of the user info stored in Okta.
2) And when you say "access" the user information (like name, location, etc), what would be potential purposes/uses for doing that through Okta? Is that so we could say, perhaps, display the persons first name on their Dashboard like "Welcome Chris from Dell!"

stephenkao: 1. Definitely not a permanent decision on the code side, but we should start collecting data in Okta during onboarding as soon as possible if there's a potential we'll want to use it down the road.  The code here is just making a request for whatever data we have stored on Okta users so if we don't collect that data at the start, we'd just fetch empty values.
2. Yep, exactly -- right now, we're just using email, but there's potential to use other data like timezone, address, and phone number to inform future features of the site.
3. Just one HTTP request to the Okta API, which isn't bad.  I was just thinking we could simplify it since we already have the email address available without making the request, but not a problem if we want to keep it!

I think we can go forward with keeping the request -- no harm in keeping it for now :)

chris-kuryak: @stephenkao 
Actually, I would lean the opposite direction for now. Let's keep it simple until we have confirmation we need it. Any harm in that? Especially if we could add it back in later.

I know we don't currently ask for extra information on Okta account creation, but I know Sharon has it in SmartSheet, and I think we could manually add it if we wanted to. HOWEVER, I don't see any point in putting that in now without confirmation.

So I say simplify for now and I will bring it up at next O+O/Exp sync. Sound good?

stephenkao: For posterity, we chatted about this in standup and are okay with removing the fetch for now!

frontend-react/src/contexts/SessionContext.tsx
```diff
@@ -49,16 +50,20 @@ const SessionProvider = ({
         return activeMembership?.memberType === MemberType.PRIME_ADMIN;
     }, [activeMembership?.memberType]);
 
+    const context = useMemo(
```
stephenkao, line 53: Minor point, but since this entire Context value is going to be memoized, would you mind removing the `useMemo` for `isAdminStrictCheck` above?  Since we're no longer allowing users to toggle between different `member_type` values, we can assume that any changes to `activeMembership` will result in a reference change so `isAdminStrictCheck` would be recomputed anyway.  (Plus, I thiiiiink the  old `member_type` change feature was just spreading on to a new object so the reference would change anyway 🤷)

frontend-react/src/utils/CustomRenderUtils.tsx
```diff
@@ -15,15 +15,18 @@ import { mockToken } from "./TestUtils";
...
 export const makeOktaHook = (_init?: Partial<IOktaContext>): OktaHook => {
-    return () => ({
+    const result = {
         authState: {
             accessToken: mockToken(),
             ..._init?.authState,
         },
         oktaAuth: {
             ..._init?.oktaAuth,
+            getUser:
```
stephenkao, line 25: I think we can remove this now, no?

stephenkao (approved):
Added one more quick comment about a test, but this looks great regardless!  Thanks again for hopping on this so quickly again!

PR #8163, opened 2023-02-02
Move DOMPurify.sanitize into USExtLink#7843

jpandersen87:
This PR centralizes the sanitization of urls passed for USLink/USExtLink/USCrumbLink to within USLink (USExtLink and USCrumbLink wrap around USLink). This makes sanitizing urls before passing to the components unnecessary.


etanb:
@jpandersen87 can we combine this `dependabot` update #8037  with this PR since they're directly related?

jpandersen87:
> @jpandersen87 can we combine this `dependabot` update #8037 with this PR since they're directly related?

Great idea. I merged those changes.

frontend-react/src/components/USLink.tsx
```diff
@@ -32,7 +33,7 @@ export const USLink = ({
 }: USLinkProps) => {
     return !anchor ? (
         <Link
-            to={href || ""}
+            to={href ? DOMPurify.sanitize(href) : ""}
```
etanb, line 36: Should there even be a case where `href` is empty or falsy here? I don't think (and I know this is how it was previously structured) the null check on `href` should be within the `<Link>` itself. I'd think this should look something like:

```
export const USLink = ({
    anchor = false,
    children,
    className,
    href,
    state,
    ...anchorHTMLAttributes
}: USLinkProps) => {
    if (!href) return;

    return (
        <>
            {!anchor ? (
                <Link
                    to={DOMPurify.sanitize(href)}
                    className={classnames("usa-link", className)}
                    state={state}
                    {...anchorHTMLAttributes}
                >
                    {children}
                </Link>
            ) : (
                <a
                    href={DOMPurify.sanitize(href)}
                    className={classnames("usa-link", className)}
                    {...anchorHTMLAttributes}
                >
                    {children}
                </a>
            )}
            ;
        </>
    );
};
```

Or if for some reason we need an empty string based `<Link>` we can just pull it out as a variable since the `<a>` should have it sanitized as well:

```
export const USLink = ({
    anchor = false,
    children,
    className,
    href,
    state,
    ...anchorHTMLAttributes
}: USLinkProps) => {
    const sanitizedLink = href ? DOMPurify.sanitize(href) : "";

    return (
        <>
            {!anchor ? (
                <Link
                    to={sanitizedLink}
                    className={classnames("usa-link", className)}
                    state={state}
                    {...anchorHTMLAttributes}
                >
                    {children}
                </Link>
            ) : (
                <a
                    href={sanitizedLink}
                    className={classnames("usa-link", className)}
                    {...anchorHTMLAttributes}
                >
                    {children}
                </a>
            )}
        </>
    );
};
```

jpandersen87: I did have the same thought when seeing this code for the first time. I'm not against requiring the href or just sanitizing it no matter what (even if an empt string) if empty strings are allowed downstream.

frontend-react/package.json
```diff
@@ -17,7 +17,7 @@
         "@xmldom/xmldom": "0.8.6",
         "axios": "^1.2.1",
         "classnames": "^2.3.1",
-        "dompurify": "^2.3.4",
+        "dompurify": "^2.4.3",
```
stephenkao, line 20: Hopefully none of us is dyslexic :)

frontend-react/src/components/USLink.tsx
stephenkao: `href` is technically an optional attribute for anchors, and a `href`-less anchor actually has a valid use [as a placeholder link](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element) like a disabled button:

> If the [a](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element) element has no [href](https://html.spec.whatwg.org/multipage/links.html#attr-hyperlink-href) attribute, then the element [represents](https://html.spec.whatwg.org/multipage/dom.html#represents) a placeholder for where a link might otherwise have been placed, if it had been relevant, consisting of just the element's contents.

Whether or not that's actually useful to us is another question, but my vote would be something like #2 so we'd err on the side of allowing what's valid in HTML.

--- 

Also just in case it's helpful, you can enable syntax highlighting for your code in GitHub by adding a file extension at the end of the first set of backticks:

```tsx
// with ```tsx for TypeScript
type Whatever = {
   Blah: 1,
   Bleh: 2
};
```

```kt
// with ```kt for Kotlin
fun main() {
    println("beep beep")
}
```

stephenkao (approved):
I agree with @etanb's point, but this looks great to me otherwise!

Issue #8147, opened 2023-02-01
Update new ITEM table during receive step


arnejduranovic:
Refers to #8032

arnejduranovic:
The Universal Pipeline already treats the item lineage table as an "item" table and there is no immediate need for an item table in the old pipeline, since message tracker already works for covid via the covid metadata table, and that is essentially all the old pipeline is and will be used for.

Will potentially revisit/rethink this when working on universal search engine.

Issue #8033, opened 2023-01-23
API users want a messageID API that works for all topics / datatypes


jimduff-usds:
The current messageID API only works for covid.  This ticket is to make it work for all future datatypes, by using the new ITEM table, created in #8032.

This ticket includes some tech debt cleanup on the messageID API as well.
#### REST API Name change
The concept of `message` is HL7 specific.   Generalize to the term `item` which can be a message, a row, or a bundle, as follows:
- Change `api/messages/search?messageID=xxx` to `api/items?trackingId=xxx`.
- Change `api/message/{covid_results_metadata_id}` to `api/item/{item_id}` 
- Fix the documentation too (docs/api/message.yml, release notes, kotlin code).
- Rename Message to something else (probably not Item, as that'll be generated by Jooq).  Maybe ItemHistory?

#### Search on item_id, not covid_results_metadata_id

- The "id' returned in the new ItemHistory object, (returned by the api/items endpoint) should now always be the item_id, the primary key of the item table.
- For each item found by any query in the ITEM table, also return its metadata. 

#### Other improvements
- Add 'topic' to the ItemHistory object so we know which metadata table it came from.  THis is found in the ReportFile table.

#### Tech debt cleanup:
- The .kt files for this need to move to src/kotlin/history/azure.  messageID queries are history API queries, so they properly belong in history/azure.
- ItemHistory (was Message) should be its own class to parallel DeliveryHistory.kt etc.

### Implementation Notes

The linking of ITEM  - via Topic - to the corresponding ***_Metadata table can be accomplished in several different ways.   
- returns just the generic ITEM data, then requiring a second, topic-specific query, to get the *_metadata.
- using a fancy table inheritance design.    If queries are going to go across topics, or return a large number of ITEMs, this is the most performant solution.  However, I think its its own separate ticket.
- Require the user to send a topic parameter.   this is a nice MVP solution.  Use the topic to determine which metadata table to join with.     A query can only get data for one topic.
-

avnieldravid:
Waiting on Uinversal Search design to create Epic.

PR #8184, opened 2023-02-03
Create ITEM table and link to related tables

### LINKED ISSUE #8032: Create an ITEM table, in support of the messageID API
### Problem

Currently we do not have any table in the system that has one row per submitted ITEM.   The closest we have is ITEM_LINEAGE.  It has these flaws:

- Data is only inserted into ITEM_LINEAGE if there is both a parent and a child report.   So ITEMS that get completely filtered out never appear in ITEM_LINEAGE and so are not well tracked in our system.  (They do appear in COVID_RESULT_METADATA, which is one reason why the messageID query is using that table)
- As we add more datatypes (eg, full_elr, covid, etc), we need a table per item that knows nothing about what datatype it is.  This is the ITEM table.
- Then all the datatype specific metadata tables can simply FK to the ITEM table.

### Acceptance Criteria

- [x]  The new ITEM table is nicely designed 
- [x] ITEM TABLE COLUMNS: item_id, report_id, created_at
- [x] ITEM_LINEAGE, COVID_RESULT_METADATA, and ELR_RESULT_METADATA all FK to ITEM. All these tables need an ITEM_ID column added to them (FK)
- [x] ITEM FKs to REPORT_FILE.
- [ ] All descendant ITEM_LINEAGE entries (not just for the initially received report, but also for the routed, batched, and sent reports) have the proper FK back to the original ITEM_ID) - This will be a part of #8147
- [x] No backfill of old items; that can be a separate ticket, or we can choose not to do it.
- [x] No changes to messageID api (yet) - separate ticket.

### Notes

See #8031 for more background on this need.
See #8147 for populating the data
See also #8033 for the follow-on work to change the messageID queries to use this.

arnejduranovic:
Create a new `item` table in the database and links it to `report_file`, `covid_result_metadata`, and `elr_result_metadata`. This should help us implement a non-covid-specific message tracker API, as well as possibly simplifying queries like `report_facilities(UUID)`

Test Steps:
1. run `./gradlew clean package` and verify migration passes

## Changes
- Create `item` table and its relationships
- Wherever the POJOs are used, pass `null` for the item id, for now

### Testing
- [ ] Tested locally?
- [ ] Added tests?

### Process

## Linked Issues
- Fixes #8032 

## To Be Done
- #8147
- #8033

### DIFF
--- a/prime-router/src/main/resources/db/migration/V61__create_item_table.sql
+++ b/prime-router/src/main/resources/db/migration/V61__create_item_table.sql
@@ -1,0 +1,36 @@
+/*
+This SQL creates the tables of the DB. The Flyway tool applies this migration to create the database
+
+Follow this style guide https://about.gitlab.com/handbook/business-ops/data-team/platform/sql-style-guide/
+use VARCHAR(63) for names in organization and schema
+
+Copy a version of this comment into the next migration
+*/
+
+/*
+    An item can be a covid test result or any other message (HL7 v2), row (CSV), bundle (FHIR), et. al.
+*/
+CREATE TABLE item (
+    -- Key
+                            id SERIAL PRIMARY KEY,
+
+    -- Value
+                            report_id UUID,
+                            created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
+    -- Foreign Key
+                            CONSTRAINT report_id
+                                FOREIGN KEY(report_id)
+                                REFERENCES report_file(report_id)
+);
+
+/*
+    Add FK references to ITEM table in related tables
+ */
+ALTER TABLE item_lineage ADD COLUMN item_id INT
+    CONSTRAINT item_lineage_item_id_fkey REFERENCES item (id);
+
+ALTER TABLE covid_result_metadata ADD COLUMN item_id INT
+    CONSTRAINT covid_result_metadata_item_id_fkey REFERENCES item (id);
+
+ALTER TABLE elr_result_metadata ADD COLUMN item_id INT
+    CONSTRAINT elr_result_metadata_item_id_fkey REFERENCES item (id);
--- a/prime-router/src/main/kotlin/Report.kt
+++ b/prime-router/src/main/kotlin/Report.kt
@@ -1422,7 +1422,8 @@ class Report : Logging {
                     grandParentTrackingValue,
                     null,
                     null,
-                    itemHash
+                    itemHash,
+                    null
                 )
             } else {
                 val trackingElementValue =
@@ -1436,7 +1437,8 @@ class Report : Logging {
                     trackingElementValue,
                     null,
                     null,
-                    itemHash
+                    itemHash,
+                    null
                 )
             }
         }
@@ -1467,7 +1469,8 @@ class Report : Logging {
                         it.trackingId,
                         it.transportResult,
                         null,
-                        it.itemHash
+                        it.itemHash,
+                        null
                     )
             }
             val retval = mutableListOf<ItemLineage>()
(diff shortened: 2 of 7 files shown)


prime-router/src/main/resources/db/migration/V61__create_item_table.sql
```diff
@@ -0,0 +1,36 @@
...
+*/
+
+/*
+    An item can be a covid test result or any other message (HL7 v2), row (CSV), bundle (FHIR), et. al.
+*/
+CREATE TABLE item (
+    -- Key
+                            id SERIAL PRIMARY KEY,
+
+    -- Value
```
arnejduranovic, line 17: Would it be appropriate/valid/helpful to add `tracking_id` here. `tracking_id` exists in item_lineage table.

prime-router/src/main/resources/db/migration/V61__create_item_table.sql
```diff
@@ -0,0 +1,36 @@
...
+                            CONSTRAINT report_id
+                                FOREIGN KEY(report_id)
+                                REFERENCES report_file(report_id)
+);
+
+/*
+    Add FK references to ITEM table in related tables
+ */
+ALTER TABLE item_lineage ADD COLUMN item_id INT
+    CONSTRAINT item_lineage_item_id_fkey REFERENCES item (id);
```
arnejduranovic, line 30: No data integrity rules here (DELETE/UPDATE CASCADE) because once written, our data is not getting modified and I'm scared to mess data up. Is this OK?

arnejduranovic: My vote is to do the update/delete cascade for these tables, but interested in thoughts.

carlosfelix2: I would do it. The moment someone deletes a row somewhere else then it may/will create problems

prime-router/src/main/resources/db/migration/V61__create_item_table.sql
```diff
@@ -0,0 +1,36 @@
...
+                            id SERIAL PRIMARY KEY,
+
+    -- Value
+                            report_id UUID,
+                            created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
+    -- Foreign Key
+                            CONSTRAINT report_id
+                                FOREIGN KEY(report_id)
+                                REFERENCES report_file(report_id)
+);
```
carlosfelix2, line 24: I don't understand this table. report ID is already a PK in report_file. Why have a table that just have report IDs?

carlosfelix2 (changes requested):

prime-router/src/main/resources/db/migration/V61__create_item_table.sql
```diff
@@ -0,0 +1,36 @@
...
+Follow this style guide https://about.gitlab.com/handbook/business-ops/data-team/platform/sql-style-guide/
+use VARCHAR(63) for names in organization and schema
+
+Copy a version of this comment into the next migration
+*/
+
+/*
+    An item can be a covid test result or any other message (HL7 v2), row (CSV), bundle (FHIR), et. al.
+*/
+CREATE TABLE item (
```
carlosfelix2, line 13: Don't really like this short name. Maybe something more descriptive?

arnejduranovic:
Closing this PR and associated ticket. 

The Universal Pipeline already treats the item lineage table as an "item" table and there is no immediate need for an item table in the old pipeline, since message tracker already works for covid via the covid metadata table, and that is essentially all the old pipeline is and will be used for.

Will potentially revisit/rethink this when working on universal search engine.

PR #8190, opened 2023-02-03
7754 - Redesign Message Tracker Sender Details

penny-lischer:
This PR ...

Improves the visual hierarchy and overall readability based on internal usability findings.

Test Steps:
1. Log into RS
2. Add 'message-tracker' feature flag
3. Navigate to the message tracker feature
4. Enter a message Id in the search (Can be found via the covid_result_metadata table)
5. Select a message Id
6. View design changes

## Changes
Screenshot 2023-02-03 at 2 41 54 PM
Screenshot 2023-02-03 at 2 42 03 PM

### Testing
- [x] Tested locally?
- [x] Added tests?


frontend-react/src/components/MessageTracker/MessageDetails.tsx
```diff
@@ -15,57 +18,154 @@ type MessageDetailsProps = {
...
+    fileLocation: string;
+    sendingOrg: string;
+    fileName: string;
+} => {
+    let fileLocationFromStr = "";
+    let sendingOrgFromStr = "";
+    let fileNameFromStr = "";
+
+    let fileReportsLocation = fileLocation.split("reports/").pop();
+    if (fileLocation.includes("reports/") && fileReportsLocation) {
```
penny-lischer, line 33: Do we want to display the filename even if "reports/" is not included in the fileLocation?

frontend-react/src/styles/_uswds_overrides.scss
```diff
@@ -110,6 +112,27 @@ main {
   overflow-x: auto;
 }
 
+// Message Tracker specific styles for Accordion component
+.message-tracker {
```
penny-lischer, line 116: Should this class live here or be moved to another scss file?

frontend-react/src/components/MessageTracker/MessageDetails.test.tsx
```diff
@@ -141,14 +141,15 @@ const MOCK_EMPTY_MESSAGE_DETAIL = {
     errors: [],
     receiverData: [],
 };
-const MOCK_MESSAGE_DETAIL = {
+let MOCK_MESSAGE_DETAIL = {
```
stephenkao, line 144: We can keep this a `const` and just spread whenever we need to override anything inside:

```
// usually better practice to name this something like "BASE" or "DEFAULT" like BASE_MESSAGE_DETAIL or DEFAULT_MESSAGE_DETAIL
const DEFAULT_MESSAGE_DETAIL = { ... };

...

test("does not return fileLocation, sendingOrg, and fileName", () => {
    const mockMessageDetails = {
        ...DEFAULT_MESSAGE_DETAIL,
        fileUrl: "https://azurite:10000/devstoreaccount1/receive%2Fsimple_report.csvuploader%2Fupload-covid-19-c33f9d36-9e5b-44eb-9368-218d88f3a7d1-20230131190253.csv";
    };  
});
```

The reason for this is at least twofold:
1. React state changes are triggered on _reference_ changes, not value changes.  For example, even if `MOCK_MESSAGE_DETAIL.fileUrl` is mutated, it won't register as a state change because `MOCK_MESSAGE_DETAIL` still has the same reference in memory.
2. It prevents state changes from "leaking" between test cases.  Like if writing over `MOCK_MESSAGE_DETAIL.fileUrl` [here](https://github.com/CDCgov/prime-reportstream/pull/8190/files#diff-d758df7c75cb3671680fdd77a2b4dc67c7968b61ad5428fb89bd6f923584a3aaR213) means that that's permanently changed for **all test cases** unless it's explicitly overridden in other tests.  Creating a new object for each test case makes it so that any potential mutation is fully contained in the test case.  (This increases memory complexity, but in Jest tests/internal use that's definitely tolerable.  I'd honestly also say that's pretty tolerable in React overall because lost references will just be garbage collected, and it's rare that this'll have significant performance implications unless we're implementing something in a really, really unperformant way anyway 🤷)


--- 
As a side note, you can still mutate nested fields in an object even if declared with a `const`; it's just the reference to the object that just can't change:

```js
const obj = {};
obj.a = "whatever"; // valid
obj = {}; // invalid
```

---

[Here](https://blog.logrocket.com/immutability-react-should-you-mutate-objects/)'s a good article about mutability/immutability in the context of React rendering in case you're interested!

frontend-react/src/components/MessageTracker/MessageDetails.test.tsx
```diff
@@ -173,21 +174,66 @@ describe("RSMessageDetail component", () => {
...
 
-    test("renders expected content", async () => {
+    test("renders expected content", () => {
         mockUseMessageDetails.mockReturnValueOnce({
             messageDetails: MOCK_MESSAGE_DETAIL as RSMessageDetail,
         });
         render(<MessageDetails />);
 
-        expect(screen.getByText(/Message ID/)).toBeInTheDocument();
+        expect(screen.getByText("Message ID")).toBeInTheDocument();
```
stephenkao, line 183: Small thing, but I think it'd be a good idea to start using `toBeVisible` rather than `toBeInTheDocument` when we're looking for things that are supposed to exist in the DOM.  The former checks for "pseudo-actual" visibility (certain styles like `display`, `visibility`, and opacity) whereas the latter just checks that it exists in the markup.

frontend-react/src/components/MessageTracker/MessageDetails.test.tsx
```diff
@@ -173,21 +174,66 @@ describe("RSMessageDetail component", () => {
         expect(mockUseMessageDetails).toHaveBeenCalledWith(TEST_ID);
     });
 
-    test("renders expected content", async () => {
+    test("renders expected content", () => {
         mockUseMessageDetails.mockReturnValueOnce({
-            messageDetails: MOCK_MESSAGE_DETAIL as RSMessageDetail,
+            messageDetails: DEFAULT_MESSAGE_DETAIL as RSMessageDetail,
```
stephenkao, line 179: Let's try to avoid downcasting (`as RSMessageDetail`) and be upfront with the type during declaration:

```ts
const DEFAULT_MESSAGE_DETAIL: RSMessageDetail = {
    ...
};
```

That way, it's clear to TypeScript that this object will always be an instance of `RSMessageDetail`, and no casting is required ever again :)

frontend-react/src/components/MessageTracker/MessageDetails.tsx
etanb: I'd love to learn more about this function and it's intended design. It seems to rely on the formatting of the url with `%2F` to indicate: location, sending organization and file name in that order. Is there a way we can restructure the backend response to form this into URL parameters? Then we could at least use `new URLSearchParams` to parse out the data much more easily / type it.

In regards specifically to `display the filename even if "reports/" is not included` I would think so, but what does that URL look like? Just the sending org + the filename? Is the data ALWAYS structured that 3 part URL === fileLocation, sendingOrg and fileName and 2 part URL === fileLocation and sendingOrg?

frontend-react/src/components/MessageTracker/MessageDetails.test.tsx
penny-lischer: Will do

frontend-react/src/styles/_uswds_overrides.scss
etanb: Are these changes across all `<Accordion />` components across the sites? If so, we should remove the `.message-tracker` scoping to the styles. Otherwise, I'd recommend we make either a `MessageTracker.module.scss` file or we can add to `_components.scss` file and we'll add component-scoped CSS in Stephen's rewrite.

etanb:
Small thing, but are the potential error messages we'll receive be very long? If so, they make the accordion table scroll across which is a funny experience (will try to upload images later, still not working at the moment). We can easily add `word-break: break-word;` to the `<td>` elements to make the text more readable.

frontend-react/src/components/MessageTracker/MessageDetails.tsx
```diff
@@ -15,57 +18,154 @@ type MessageDetailsProps = {
...
         <>
             <div
                 className="grid-container margin-bottom-10"
                 data-testid="container"
             >
-                <div className="margin-bottom-5">
-                    <h1>Message Details</h1>
+                <div>
+                    <h2>Message ID Search</h2>
                     <Button onClick={() => navigate(-1)} type="button">
```
etanb, line 90: In case a user got here directly / from a different page, is it worth actually setting the "Back" functionality to explicitly go to `/admin/message-tracker`?

frontend-react/src/components/MessageTracker/MessageDetails.test.tsx
```diff
@@ -167,27 +168,72 @@ jest.mock("react-router-dom", () => ({
...
-            messageDetails: MOCK_EMPTY_MESSAGE_DETAIL as RSMessageDetail,
+            messageDetails: MOCK_EMPTY_MESSAGE_DETAIL,
         });
         render(<MessageDetails />);
         expect(mockUseMessageDetails).toHaveBeenCalledWith(TEST_ID);
     });
 
-    test("renders expected content", async () => {
+    test("renders expected content", () => {
         mockUseMessageDetails.mockReturnValueOnce({
```
stephenkao, line 178: This is a much larger point of discussion for the entire team so no changes needed at this point, but I think we should move all test setup/teardown work into `beforeEach`/`beforeAll` and `afterEach`/`afterAll` hooks and only/mostly running expectations in the `test` blocks.  The reason for this is that it makes it much easier to isolate expectations based on specific setup conditions and elevate common setup/teardown that would otherwise be redundant.

Again, no changes needed at this point, but I wanted to throw this thought out there!

frontend-react/src/components/MessageTracker/MessageDetails.tsx
penny-lischer: Yes, but we have a story coming to remove the back button and use breadcrumb instead.

frontend-react/src/styles/_uswds_overrides.scss
penny-lischer: Will move.

frontend-react/src/components/MessageTracker/MessageDetails.tsx
penny-lischer: @stephenkao thoughts on this?

penny-lischer:
> Small thing, but are the potential error messages we'll receive be very long? If so, they make the accordion table scroll across which is a funny experience (will try to upload images later, still not working at the moment). We can easily add `word-break: break-word;` to the `<td>` elements to make the text more readable.

I dont see the horizontal scroll with longer error messages. Can you please provide me with an example?

frontend-react/src/components/MessageTracker/MessageDetails.tsx
stephenkao: Good point!  There's definitely some cleanup we can do with this, but I think we can assume this format will be okay for the time being -- as far as I know, the structure of this URL is pretty set (at least with the COVID pipeline) since that signifies the step in the pipeline.  We could also honestly probably just grab whatever's after the last unencoded slash instead of splitting on `reports` 🤷

frontend-react/src/components/MessageTracker/MessageDetails.tsx
```diff
@@ -15,57 +18,157 @@ type MessageDetailsProps = {
...
+    sendingOrg: string;
+    fileName: string;
+} => {
+    let fileLocationFromStr = "";
+    let sendingOrgFromStr = "";
+    let fileNameFromStr = "";
+
+    let fileReportsLocation = fileLocation.split("reports/").pop();
+    if (fileLocation.includes("reports/") && fileReportsLocation) {
+        fileLocationFromStr = fileReportsLocation.split("%2F")[0].toUpperCase();
```
stephenkao, line 34: We're creating a new array on each invocation of `split` here -- we can just keep the one reference around and pick from it:

```tsx
[fileLocation, sendingOrg, fileName] = fileReportsLocation.split("%2F");
```

Or if we want to make all this logic a one-liner:

```tsx
[fileLocation, sendingOrg, fileName] = fileLocation.split("/").slice(-1)[0].split("%2F");
```

Regardless, let's wrap this in a `try-catch` and return empty string just in case there's a format issue, though!

frontend-react/src/components/MessageTracker/MessageDetails.tsx
```diff
@@ -15,57 +18,157 @@ type MessageDetailsProps = {
     id: string | undefined;
 };
 
+const parseFileLocation = (
```
stephenkao, line 21: Would you mind adding a few unit tests for this function?

frontend-react/src/components/MessageTracker/MessageDetails.tsx
penny-lischer: @stephenkao  I'm splitting on `reports/ `since I need to grab the folder location, sending org, and filename which all come after `reports/ `. 
If I grab whatever's after the last unencoded slash, I will only get the filename.

stephenkao: Hmmmm isn't the URL structure like this?

```
https://azurite:10000/devstoreaccount1/reports/receive%2Fsimple_report.csvuploader%2Fupload-covid-19-c33f9d36-9e5b-44eb-9368-218d88f3a7d1-20230131190253.csv
```

If we grab the last fragment when splitting on `'/'`, we should get `receive%2Fsimple_report.csvuploader%2Fupload-covid-19-c33f9d36-9e5b-44eb-9368-218d88f3a7d1-20230131190253.csv` -- then we can just split on the encoded `'%2F'` to get the folder location, Sender, and filename, right?

frontend-react/src/components/MessageTracker/MessageDetails.tsx
```diff
@@ -15,57 +18,157 @@ type MessageDetailsProps = {
...
-                        content={messageDetails!.fileName}
-                    ></DetailItem>
-                    <DetailItem
-                        item={"Incoming File URL"}
-                        content={messageDetails!.fileUrl}
-                    ></DetailItem>
+                    <h2 className="font-sans-lg margin-bottom-4">Sender:</h2>
+                    <div className="font-sans-sm bg-gray-5 radius-lg padding-top-3 padding-bottom-4 padding-left-3 padding-right-3">
+                        <DetailItem
+                            item={"Submitter"}
```
stephenkao, line 109: This also existed beforehand, but we don't need to wrap string props in curly braces:

```jsx
item="Submitted"
```

frontend-react/src/components/MessageTracker/MessageDetails.tsx
```diff
@@ -15,57 +18,157 @@ type MessageDetailsProps = {
...
                     </Button>
                 </div>
-                <DetailItem
-                    item={"Message ID"}
-                    content={messageDetails!.messageId}
-                ></DetailItem>
+                <hr className="margin-top-4 margin-bottom-4" />
+                <div className="display-flex flex-column">
+                    <span className="text-base font-sans-sm">Message ID</span>
+                    <h1 className="margin-top-1">
```
stephenkao, line 100: Let's try to use a single direction for vertical margins: either top or bottom, but not both.  [Collapsing margins](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Box_Model/Mastering_margin_collapsing) can be a pain to work around, especially when we're trying to establish a predictable vertical rhythm between elements on the page.

frontend-react/src/components/MessageTracker/MessageDetails.tsx
penny-lischer: Yup, sorry I misread what you meant previously. Will update.

frontend-react/src/components/MessageTracker/MessageDetails.tsx
```diff
@@ -1,12 +1,16 @@
 import React from "react";
+import moment from "moment";
```
stephenkao, line 2: Let's try to deprecate our usage of the `moment` dependency -- it adds a huge amount to our payload size, and the maintainers even say that [the project is done](https://momentjs.com/docs/#/-project-status/).

Instead, we can use JavaScript's native `Intl`:

```jsx
// can declare this in the "global" scope outside of components
const dateTimeFormatter = new Intl.DateTimeFormat("en-US", { month: "2-digit", day: "2-digit", year: "numeric", hour: "2-digit", minute: "2-digit", second: "2-digit" })

// call format() whenever you need a formatted date:
content={ dateTimeFormatter.format(messageDetails!.submittedDate) }  
```

Once we've removed all references to `moment` in our codebase, we can add a ticket to remove it from our dependencies and shed a lot of bundle weight :)

frontend-react/src/components/MessageTracker/MessageDetails.tsx
```diff
@@ -15,57 +19,132 @@ type MessageDetailsProps = {
     id: string | undefined;
 };
 
+const dataToAccordionItems = (props: {
```
stephenkao, line 22: This function feels a bit odd to me, but I can't think of a better way to pass the necessary data and the custom WarningsErrors component to the Accordion right now 🤔

frontend-react/src/components/MessageTracker/MessageDetails.tsx
```diff
@@ -15,57 +19,132 @@ type MessageDetailsProps = {
...
+    });
+
+    return results;
+};
+
 export function MessageDetails() {
     const navigate = useNavigate();
     const { id } = useParams<MessageDetailsProps>();
     const { messageDetails } = useMessageDetails(id!!);
+    const warnings: WarningError[] = messageDetails
```
stephenkao, line 50: These three variables can be simplified a bit with safe navigation:

```tsx
const warnings = messageDetails?.warnings || []; // I think the WarningError type can be inferred?
const errors = messageDetails?.errors || [];           // same here?
const fileUrl = messageDetails?.fileUrl || "";
```

frontend-react/src/components/MessageTracker/MessageDetails.tsx
```diff
@@ -15,57 +19,132 @@ type MessageDetailsProps = {
...
                 className="grid-container margin-bottom-10"
                 data-testid="container"
             >
-                <div className="margin-bottom-5">
-                    <h1>Message Details</h1>
-                    <Button onClick={() => navigate(-1)} type="button">
+                <div>
+                    <h2>Message ID Search</h2>
+                    <Button
+                        onClick={() => navigate("/admin/message-tracker")}
```
stephenkao, line 67: Not a huge deal, but since this is navigating somewhere, I think we should use a link instead...but our USLink component always applies link-specific styles 🤔 

Maybe we can just use react-router's Link and add the `usa-button` modifier class for now?  But we'd likely want to expand USLink to adopt button classes down the line 🤔 🤔 🤔

frontend-react/src/utils/misc.ts
```diff
@@ -174,3 +174,31 @@ export const groupBy = <T>(
...
+    folderLocation: string;
+    sendingOrg: string;
+    fileName: string;
+} => {
+    let fileLocationFromStr = "";
+    let sendingOrgFromStr = "";
+    let fileNameFromStr = "";
+
+    const fileReportsLocation = urlFileLocation.split("/").pop();
+    if (fileReportsLocation?.split("%2F").length === 3) {
```
stephenkao, line 193: Because of the four `split` calls, we're still creating four separate arrays from splitting on the same string.  I think we can simplify this to be the following:

```jsx
const fileReportsLocation = urlFileLocation.split("/").pop() || "";
const [folderLocation, sendingOrg, fileName] = fileReportsLocation.split("%2F");
        
if (!(folderLocation && sendingOrg && fileName)) {
    return {
        folderLocation: "",
        sendingOrg: "",
        fileName: ""
    }
}

return {
    folderLocation,
    sendingOrg,
    fileName
};
```

frontend-react/src/styles/_uswds_overrides.scss
```diff
@@ -19,6 +19,8 @@ i.e.
 @include u-padding-right('05');
 ----------------------------------------
 */
+$bg-primary: #005ea2;
```
stephenkao, line 22: Hmmm I realize these hardcoded values already existed beforehand, but the hex code is already available from the USWDS CSS as `primary`.  Like we could just use `color('primary')` instead of creating a new variable 🤷

frontend-react/src/utils/misc.test.ts
```diff
@@ -218,3 +219,25 @@ describe("isProhibitedName", () => {
...
+            "https://azurite:10000/devstoreaccount1/reports/receive%2Fsimple_report.csvuploader%2Fupload-covid-19-c33f9d36-9e5b-44eb-9368-218d88f3a7d1-20230131190253.csv"
+        );
+        expect(folderLocation).toEqual("RECEIVE");
+        expect(sendingOrg).toEqual("simple_report.csvuploader");
+        expect(fileName).toEqual(
+            "upload-covid-19-c33f9d36-9e5b-44eb-9368-218d88f3a7d1-20230131190253.csv"
+        );
+    });
+
+    test("returns empty strings when string is missing three '/'", () => {
```
stephenkao, line 235: Maybe another test that'll verify that all three strings are empty when there aren't three fragments split on `%2F`?

frontend-react/src/components/MessageTracker/MessageDetails.test.tsx
```diff
@@ -167,27 +168,76 @@ jest.mock("react-router-dom", () => ({
...
+        expect(screen.getByText(/Incoming Report ID/)).toBeVisible();
         expect(
             screen.getByText(/29038fca-e521-4af8-82ac-6b9fafd0fd58/)
-        ).toBeInTheDocument();
+        ).toBeVisible();
+        expect(screen.getAllByText("Date/Time Submitted")[0]).toBeVisible();
+        expect(screen.getByText("9/28/2022, 12:00:00 am")).toBeVisible();
+        expect(screen.getByText("File Location")).toBeVisible();
+        expect(screen.getByText("RECEIVE")).toBeVisible();
+        expect(screen.getByText("/ simple_report.csvuploader")).toBeVisible();
```
stephenkao, line 195: Just so we don't have to check explicitly for a slash and space (which feels a little fragile), we could just render the slash with horizontal:

```jsx
<span className="font-mono-sm border-1px bg-primary-lighter radius-md padding-top-4px padding-bottom-4px padding-left-1 padding-right-1">{folderLocation}</span>
<span className="padding-x-1">/</span>
<span>{sendingOrg}</span>
```

stephenkao (approved):
Bunch of small comments, but this looks good to me!

frontend-react/src/components/MessageTracker/MessageDetails.test.tsx
etanb: Worth making a ticket for @stephenkao ?

frontend-react/src/utils/misc.ts
etanb: Might as well instantiate an object with those values at the beginning like:

```
const fileDetails = { 
  folderLocation: "", 
  sendingOrg: "", 
  fileName: "" 
}
const fileReportsLocation = urlFileLocation.split("/").pop() || ""; 
const [folderLocation, sendingOrg, fileName] = fileReportsLocation.split("%2F");

return {
  ...fileDetails,
  ...folderLocation && { folderLocation: folderLocation },
  ...sendingOrg && { sendingOrg: sendingOrg },
  ...fileName && { fileName: fileName },
}
```

Since we're cool with sending empty strings back up, we might as well send back _any_ data we have. Not just when all 3 are present. What do y'all think?

etanb: Or I guess (sorry for getting into the weeds here) even fewer lines:

```
const fileReportsLocation = urlFileLocation.split("/").pop() || ""; 
const [folderLocation, sendingOrg, fileName] = fileReportsLocation.split("%2F");

return {
  folderLocation : folderLocation ? folderLocation : "",
  sendingOrg: sendingOrg ? sendingOrg : "",
  fileName: fileName ? fileName : "",
}
```

Sometimes being too clever, is too much 😄

etanb (approved):
Looks good! Thanks so much for taking this on. We can handle any nits later on if need be.

frontend-react/src/utils/misc.ts
etanb: I can't resist, we can make it _even simpler_:

```
const fileReportsLocation = urlFileLocation.split("/").pop() || ""; 
const [folderLocation = "", sendingOrg = "", fileName = ""] = fileReportsLocation.split("%2F");

return {
  folderLocation ,
  sendingOrg,
  fileName,
}
```

frontend-react/src/components/MessageTracker/MessageDetails.test.tsx
penny-lischer: @etanb made a ticket [here](https://app.zenhub.com/workspaces/experience-607d9d5e68b95200150fec37/issues/gh/cdcgov/prime-reportstream/8279)

frontend-react/src/components/MessageTracker/MessageDetails.tsx
penny-lischer: I think we should leave it for now. I hesitated making this change to begin with since this will be going away. We will be converting this section to a breadcrumb.

frontend-react/src/utils/misc.ts
stephenkao: Ah, good point!  I think it should be an all-or-nothing thing so maybe we should update this function to return null if all three fragments aren't available.  Like it doesn't make sense for us to show the bucket if we can't identify a file name -- shouldn't happen because the url should be formatted properly, but better to be safe since the check here is fragile (as we've discussed previously).  What do y'all think?

frontend-react/src/components/MessageTracker/MessageDetails.test.tsx
penny-lischer: I will update the slash after testing different scenarios.

frontend-react/src/components/MessageTracker/MessageDetails.tsx
stephenkao: Sure, makes sense to me!

frontend-react/src/utils/misc.ts
```diff
@@ -134,12 +134,12 @@ export function formatDate(date: string): string {
 }
 
 /*
-  for strings in machine readable form:
+  for strings in machine-readable form:
     * camel cased
     * inconsistent caps
-    * whitespace deliminted by - or _
```
stephenkao, line 140: I read this as "deli-minted" 🥩 💵

frontend-react/src/utils/misc.ts
etanb: If we have the data I'd say we should show it BUT are you saying if there are data fragments, then the data might be mislabeled since the structure is fragile?

stephenkao: Yeah, exactly.  It might also not make sense to show individual pieces of the following:

Screenshot 2023-02-10 at 15 56 31

Like if we only have the `RECEIVE` part, that doesn't really seem too important or actionable for the end user.

If we want to, we could tighten this function up even more to ensure that we're capturing the right pieces:
- Folder location is in one of (receive|process|send|et cetera)
- Sending Organization takes the form of something like `([a-zA-Z0-9])\.([a-zA-Z0-9])`
- File name is a UUID + extension

But I think splitting for now is fine for now, and we can test it against real data to see if there are other edge cases to handle.  Whatchu think?

etanb: If so, if we need all three, then maybe:

```
const fileDetails = { 
  folderLocation: "", 
  sendingOrg: "", 
  fileName: "" 
}
const fileReportsLocation = urlFileLocation.split("/").pop() || ""; 
const fileReportsData =  fileReportsLocation.split("%2F");

if (fileReportsData.length === 3) {
  [fileDetails.folderLocation, fileDetails.sendingOrg, fileDetails.fileName] = fileReportsData;
}

return fileDetails;
```

Issue #8200, opened 2023-02-06
Add end-to-end tests for validate page


stephenkao:
## Problem statement

With the validator page going public, we should add some end-to-end tests to cover the three basic flows:

- As a logged-out user, see three master schemas and run validation
- As an admin, see three master schemas and run validation
- As a Sender, see four schemas (three master schemas + Sender-designated schema) and run validation

## Acceptance criteria

- [ ] A new Cypress test is added for a logged-out user validation flow
- [ ] A new Cypress test is added for an admin validation flow
- [ ] A new Cypress test is added for a Sender validation flow

chris-kuryak:
This is a good place to start in creating E2E tests on the rest of the site moving forward.

KeithNava:
The scope for this original ticket has changed quite a bit but the End Goal is the same - Get an End To End testing framework pattern up and running. See: https://github.com/CDCgov/prime-reportstream/pull/10881

I'll be create new tickets and adding End to End testing to acceptance criteria after a full round of code review

PR #8201, opened 2023-02-06
Experience-7773: Add schema dropdown on validator page

### LINKED ISSUE #7773: Update validator page to allow for selecting schema or Sender settings
## Problem statement
In this MVP, we have three possible master schemas to validate against with the public validation tool.  Since the users may or not be logged in, we have no way of knowing which schema they'll want to validate their uploads against.  Additionally, Organizations have their own Sender settings against which they may want to validate.  Consequently, we'll need to provide a select box to choose among all available options on the validation page while they're uploading their file.

## What you need to know
We currently have three "master schemas" that we can validate against:
- [HL7](https://github.com/CDCgov/prime-reportstream/blob/master/prime-router/metadata/schemas/HL7/hl7-ingest-covid-19-training.schema)
- [CSV](https://github.com/CDCgov/prime-reportstream/blob/master/prime-router/metadata/schemas/upload-covid-19.schema)
- [CSV OTC](https://github.com/CDCgov/prime-reportstream/blob/master/prime-router/metadata/schemas/csv-otc-covid-19.schema)

## Acceptance criteria
**For all users**
- [x] There's a [select box](https://designsystem.digital.gov/components/select) accompanying the validator tool
- [x] The schema choice can be pre-selected via unique query parameter in the URL
  - [x] If the schema choice is invalid or not applicable for the currently logged-in user, there should be no initial selection
- [x] The validation button should be disabled until the user selects an option and chooses a file for upload

**For an unauthenticated user**
- [x] The dropdown component should be rendered with the three default options

**For an authenticated user**
- [x] The dropdown component should be rendered with all available Sender settings for the user _in addition to_ the three default options
- [x] If the user has a designated schema, the dropdown should default to that

## Additional information
- We may want to hide the tooltip behind a `public-validator` feature flag if it'd be confusing to admins as we're preparing to make the validator page public.
- [FileHandler](https://github.com/CDCgov/prime-reportstream/blob/master/frontend-react/src/components/FileHandlers/FileHandler.tsx) is also rendered in [UploadToPipeline](https://github.com/CDCgov/prime-reportstream/blob/master/frontend-react/src/pages/UploadToPipeline.tsx), but that's locked behind the now-unused `user-upload` feature flag.  We should be okay to update that with the select dropdown for the time being, and we can revisit if we need `user-upload` to be turned on for whatever reason.

stephenkao:
This PR ...

This changeset adds a dropdown to the validator page to allow the user to select between schema options to validate against.  To accommodate this, I've added a `useSenderSchemaOptions` hook that always returns the three "master" schemas" in addition to the designated Sender schema if the current user is logged is as a Sender.

This also includes a decent amount of cleanup/rewrites because the logic in useFileHandler, FileHandler, and FileHandlerForm is oddly coupled together.  Will comment below!

To generate test files to validate against different schemas, you can use the CLI tool:
```
./prime data --input-fake <item_count> --input-schema <schema> --output <output_destination>
```

like:

```
./prime data --input-fake 10 --input-schema hl7/hcintegrations-covid-19 --output ~/Desktop/whatever.hl7
```

This seems to generate lots of duplicate columns so those have to be removed or validation will complain about those, but this is a starting point for test data!

Test Steps:
As an admin:
1. Log in
2. Go to http://localhost:3000/file-handler/validate
3. Verify that the three master schemas are available options
4. Upload a file to validate against the selected schemas --> ensure you get the right success/warning/error messages

As a Sender:
1. Log in
2. Go to http://localhost:3000/file-handler/validate
3. Verify that the three master schemas AND the Sender schema are available options
4. Upload a file to validate against the selected schemas --> ensure you get the right success/warning/error messages

This can also be tested against a logged-out user by changing [this line](https://github.com/CDCgov/prime-reportstream/blob/master/frontend-react/src/pages/Validate.tsx#L23) to be `export const ValidateWithAuth = Validate;` just to get around the authentication stuff.  The validate call will still work.

## Changes
As an admin or unauthenticated user:
Screenshot 2023-02-06 at 11 18 46

As a Sender:
Screenshot 2023-02-06 at 11 19 10

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- Fixes #7773

## To Be Done
- #8200

### DIFF
--- a/frontend-react/src/components/FileHandlers/FileHandler.test.tsx
+++ b/frontend-react/src/components/FileHandlers/FileHandler.test.tsx
@@ -1,4 +1,4 @@
-import { screen, fireEvent, waitFor } from "@testing-library/react";
+import { fireEvent, screen } from "@testing-library/react";
 
 import {
     renderWithFullAppContext,
@@ -9,39 +9,44 @@ import {
     ResponseError,
     WatersResponse,
 } from "../../config/endpoints/waters";
+import * as useFileHandlerExports from "../../hooks/UseFileHandler";
 import {
-    INITIAL_STATE,
+    FileHandlerState,
     FileType,
-    FileHandlerActionType,
+    INITIAL_STATE,
 } from "../../hooks/UseFileHandler";
-import { formattedDateFromTimestamp } from "../../utils/DateTimeUtils";
-import { mockUseWatersUploader } from "../../hooks/network/__mocks__/WatersHooks";
-import { mockUseSenderResource } from "../../hooks/__mocks__/UseSenderResource";
-import { RSSender } from "../../config/endpoints/settings";
 import { mockAppInsights } from "../../utils/__mocks__/ApplicationInsights";
 import { EventName } from "../../utils/Analytics";
+import * as useSenderSchemaOptionsExports from "../../senders/hooks/UseSenderSchemaOptions";
+import {
+    STANDARD_SCHEMA_OPTIONS,
+    UseSenderSchemaOptionsHookResult,
+} from "../../senders/hooks/UseSenderSchemaOptions";
+import * as useWatersUploaderExports from "../../hooks/network/WatersHooks";
+import {
+    UseWatersUploaderResult,
+    UseWatersUploaderSendFileMutation,
+} from "../../hooks/network/WatersHooks";
+import * as useSessionContextExports from "../../contexts/SessionContext";
+import { RSSessionContext } from "../../contexts/SessionContext";
+import * as useSenderResourceExports from "../../hooks/UseSenderResource";
+import { UseSenderResourceHookResult } from "../../hooks/UseSenderResource";
+import { MembershipSettings, MemberType } from "../../hooks/UseOktaMemberships";
+import { CustomerStatus, Format } from "../../utils/TemporarySettingsAPITypes";
+import { RSSender } from "../../config/endpoints/settings";
 
-import FileHandler from "./FileHandler";
+import FileHandler, {
+    getClientHeader,
+    UPLOAD_PROMPT_DESCRIPTIONS,
+} from "./FileHandler";
 
-let fakeState = {};
-
-const hl7Sender: RSSender = {
-    name: "default",
-    organizationName: "hcintegrations",
-    format: "HL7",
-    customerStatus: "active",
-    schemaName: "hl7/hcintegrations-covid-19",
-    processingType: "sync",
-    allowDuplicates: true,
-    topic: "covid-19",
-};
 const mockSendFile: WatersResponse = {
     id: "",
     submissionId: 1,
     overallStatus: OverallStatus.VALID,
     sender: "",
     errorCount: 0,
-    warningCount: 4,
+    warningCount: 2,
     httpStatus: 200,
     errors: [],
     warnings: [
@@ -67,21 +72,6 @@ const mockSendFile: WatersResponse = {
         },
     ],
 };
-
-const mockState = (state: any) => (fakeState = state);
-const fakeStateProvider = () => fakeState;
-const mockDispatch = jest.fn();
-
-jest.mock("../../hooks/UseFileHandler", () => ({
-    ...jest.requireActual("../../hooks/UseFileHandler"),
-    default: () => {
-        return {
-            state: fakeStateProvider(),
-            dispatch: mockDispatch,
-        };
-    },
-    __esModule: true,
-}));
 
 jest.mock("../../hooks/UseOrganizationSettings", () => ({
     useOrganizationSettings: () => {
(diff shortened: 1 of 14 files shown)


frontend-react/src/components/FileHandlers/FileHandler.test.tsx
```diff
@@ -1,4 +1,4 @@
-import { screen, fireEvent, waitFor } from "@testing-library/react";
+import { screen, fireEvent } from "@testing-library/react";
```
stephenkao, line 1: This entire file has been rewritten (see [`mockUseFileHandler`](https://github.com/CDCgov/prime-reportstream/pull/8201/files#diff-350227b22d21936220afceaa87bbde022566cf5fe7ab301e53da9e0dcbc715b5R126), [`mockUseSenderSchemaOptions`](https://github.com/CDCgov/prime-reportstream/pull/8201/files#diff-350227b22d21936220afceaa87bbde022566cf5fe7ab301e53da9e0dcbc715b5R138), et cetera) as convenience functions to reduce redundancy in rendering the component.  Also, there were some tests that were either 1) not particularly useful or 2) not within the purview of this component.  For example, [checking the alert class](https://github.com/CDCgov/prime-reportstream/pull/8201/files#diff-350227b22d21936220afceaa87bbde022566cf5fe7ab301e53da9e0dcbc715b5L273) is not really useful -- we should have a data attribute if we really need to know the "type" of alert so as to not couple CSS with JavaScript logic -- and that's the responsibility of the StaticAlert component anyway.

frontend-react/src/components/FileHandlers/FileHandler.test.tsx
```diff
@@ -133,545 +119,442 @@ const fileChangeEvent = {
 } as React.ChangeEvent<HTMLInputElement>;
 
 describe("FileHandler", () => {
-    test("renders a spinner while loading sender / organization", async () => {
```
stephenkao, line 136: A lottttt of our tests right now are `async` by default, but if we're mocking the return values of our hooks (as opposed to using MSW), we can run expectations synchronously because the end result of the render will be synchronous.  Not a huge deal, but we don't need to be wrapping everything in Promises 🤷

frontend-react/src/hooks/UseCreateFetch.ts
```diff
@@ -39,20 +39,12 @@ function createTypeWrapperForAuthorizedFetch(
     ): Promise<T> {
         const headerOverrides = options?.headers || {};
         const headers = { ...authHeaders, ...headerOverrides };
-        // this system assumes that we want to be making authenticated
-        // requests whenever possible
-        if (!headers.authorization || headers.authorization.length < 8) {
```
stephenkao, line 44: I just took this out since it feels redundant.  If we make an unauthenticated request that shouldn't be handled, we'll get an error code anyhow 🤷

frontend-react/src/hooks/UseSenderResource.ts
```diff
@@ -23,13 +30,14 @@ export const useSenderResource = () => {
             authorizedFetch,
         ]
     );
-    const { data, isLoading } = rsUseQuery(
+    const { data, isLoading, isInitialLoading } = rsUseQuery(
```
stephenkao, line 33: Added `isInitialLoading` to check if the query is disabled or not.  I'd like to add `isEnabled` to our useQuery wrapper, but that would result in a lot of test files getting changed 😬

frontend-react/src/hooks/network/WatersHooks.ts
```diff
@@ -33,6 +45,10 @@ export const useWatersUploader = (
                 client: client,
             },
             data: fileContent,
+            params: {
+                schema: "standard/standard-covid-19",
```
stephenkao, line 49: Oops, need to fix this!

frontend-react/src/pages/Validate.tsx
```diff
@@ -6,16 +6,7 @@ import { AuthElement } from "../components/AuthElement";
 import { withCatch } from "../components/RSErrorBoundary";
 
 const Validate = () => {
-    return withCatch(
-        <FileHandler
-            headingText="ReportStream File Validator"
```
stephenkao, line 11: Removed all of these props since the FileHandler component doesn't need to handle customized text for upload-to-send anymore.

frontend-react/src/senders/hooks/UseSenderSchemaOptions/index.ts
```diff
@@ -0,0 +1,64 @@
+import { useSenderResource } from "../../../hooks/UseSenderResource";
```
stephenkao, line 1: This is using the new domain structure we discussed:

```
src/
  senders/
    hooks/
      UseSenderSchemaOptions/
        index.ts
        UseSenderSchemaOptions.test.ts
```

@jpandersen87 @etanb Thoughts on this?  One tiny gripe I have is that Intellij doesn't automatically the index.ts file when I type `UseSenderSchemaOptions` -- it opens up the directory instead 😆

jpandersen87: Can you elaborate on this?

jpandersen87 (changes requested):
Except for the one area you already marked to fix, this looks good to me.

etanb:
@stephenkao do you have samples of the formats for each of the 3 schemas? Just so I can test each scenario. Or, if this is a silly/n00b question, point me in the right direction to find such information?

frontend-react/src/components/FileHandlers/FileHandlerForm.tsx
```diff
@@ -35,7 +41,12 @@ export const FileHandlerForm = ({
...
     submitText,
+    schemaOptions,
+    selectedSchemaOption,
+    onSchemaChange,
 }: FileHandlerFormProps) => {
+    const isDisabled = fileName.length === 0 || !selectedSchemaOption;
+
     return (
         <Form onSubmit={(e) => handleSubmit(e)} className="rs-full-width-form">
             {!submitted && (
```
etanb, line 52: Not sure if this is outside the scope of this PR, but it seems like we're avoiding usage of `<FileInput />` usage of the `accept` prop that you can [see in practice in their Storybook](https://trussworks.github.io/react-uswds/?path=/story/components-file-input--accept-text-and-pdf).

This would help us avoid a lot of extra logic of determining selection AFTER the fact. It's quite weird that I can upload a file that is NOT supported (a `.pdf`) and it persists in the UI.

frontend-react/src/components/FileHandlers/FileHandler.tsx
```diff
@@ -149,7 +142,7 @@ const FileHandler = ({
         event.preventDefault();
 
         if (fileContent.length === 0) {
-            showError(`No File Contents To ${submitText}`);
+            showError(`No file contents To validate`);
```
etanb, line 145: nit: doesn't need to be be a template literal since there's no interpolation happening.

frontend-react/src/components/FileHandlers/FileHandlerForm.tsx
```diff
@@ -67,9 +78,28 @@ export const FileHandlerForm = ({
                 </div>
                 <div className="grid-col flex-1" />
                 <div className="grid-col flex-1 display-flex flex-column flex-align-end">
+                    <Dropdown
```
penny-lischer, line 81: Probably out of scope but would be nice for the dropdown to be hidden once the file has been validated. The user experience seems a little clunky.

penny-lischer:
Bug in dropdown...
Screenshot 2023-02-06 at 2 42 38 PM

penny-lischer:
Table width got messed up b/c of the dropdown...
Screenshot 2023-02-06 at 2 39 04 PM

frontend-react/src/components/FileHandlers/FileHandler.tsx
```diff
@@ -182,7 +175,7 @@ const FileHandler = ({
         if (eventData) {
             trackAppInsightEvent(EventName.FILE_VALIDATOR, {
                 fileValidator: {
-                    schema: sender?.schemaName,
+                    schema: selectedSchemaOption?.value,
```
penny-lischer, line 178: This is being passed through as the BASE_SCHEMA_OPTIONS.value i.e.'csv-otc-covid-19'. Is that what we want or do we want to pass the key i.e. 'CSV_OTC'?
If we leave it as is then we will need to transform it within Azure.

stephenkao:
> @stephenkao do you have samples of the formats for each of the 3 schemas? Just so I can test each scenario. Or, if this is a silly/n00b question, point me in the right direction to find such information?

We chatted about this on Slack, but just for posterity --

This isn't a silly question at all!  You can generate test files with the CLI tool:

```
./prime data --input-fake <num_items> --input-schema <schema_name> --output <output_destination>
```

It looks like this generates a number of duplicate columns so validation will fail on the stricter schemas without removing the duplicates, but it's a good starting place!

Thanks for calling this out!  I'll update the commit description to include these instructions for future reference :)

frontend-react/src/senders/hooks/UseSenderSchemaOptions/index.ts
stephenkao: Sure!  Here's a little GIF:



It just opens up the directory in the sidebar rather than the file proper -- I'd have to type `UseSenderSchemaOptions/index.ts` to open the latter.  Not a huge deal, just a bit of an annoyance on my end :P

frontend-react/src/components/FileHandlers/FileHandlerForm.tsx
stephenkao: Sure, can definitely add this!  I do want to note that since this is up to the browsers to interpret MIME types, there's a possibility of lesser-used files like HL7 (and maybe other formats in the future) not being respected by this attribute.  I don't think it'd hurt to add it in now so it's something we can just keep an eye out for in the future!

frontend-react/src/components/FileHandlers/FileHandler.tsx
stephenkao: Good catch -- will update!

frontend-react/src/components/FileHandlers/FileHandlerForm.tsx
stephenkao: Sure, I can definitely update that!

frontend-react/src/components/FileHandlers/FileHandler.tsx
stephenkao: Yes, this is intentional because that's how the schemas are stored in memory -- see [Metadata.kt:202](https://github.com/CDCgov/prime-reportstream/blob/master/prime-router/src/main/kotlin/Metadata.kt#L202) where we're looking for the schemas.  `schemaStore` is a hash map whose keys are strings based on the schema filepaths like "hl7/hcintegrations-covid-19", "upload-covid-19", et cetera.

Regarding Azure, do you mean the App Insights side?  I thiiiiiink it should be fine to leave as-is since we're passing both the schema name and the format -- this'll give us granularity to see which schemas and file types are the most used/problematic.  What do you think?

stephenkao:
> Table width got messed up b/c of the dropdown... Screenshot 2023-02-06 at 2 39 04 PM

I don't think that's caused by this changeset because the dropdown is in a separate row than the warning/error tables -- it looks like it's actually caused by the shorter text in the first column.  If you add more text into it, the table will expand to be full width.  This is likely related to the table style change from a while back that we have ticketed to fix.

stephenkao:
@jpandersen87 @etanb @penny-lischer just pushed up changes to address the above comments:

- Removing the hardcoded `format`/`schema` in [useWatersUploader](https://github.com/CDCgov/prime-reportstream/pull/8201/files#diff-37657e2387928452ba76d474388df583083ff250b55bd87ad290bcfa9856e3b2R56-R59)
- Deduplicating schema values [in options](https://github.com/CDCgov/prime-reportstream/pull/8201/files#diff-7baf450a8973abd9bdc0d7ba9334e493b6db2bb3dac086fc8cf792e7886617caR53-R64)
- [Removing the dropdown](https://github.com/CDCgov/prime-reportstream/pull/8201/files#diff-2cae5b746e955e3ca267b4083b2710fac31fdf7892a39355d3470b81a2c9d3d3R91) after submission
- Adding the [`accept` value](https://github.com/CDCgov/prime-reportstream/pull/8201/files#diff-2cae5b746e955e3ca267b4083b2710fac31fdf7892a39355d3470b81a2c9d3d3R53-R55) to limit the uploadable file types based on schema

Also added [a convenience `getClientHeader` function](https://github.com/CDCgov/prime-reportstream/pull/8201/files#diff-90daad087bcd5ce4dac7318290253991cf7355a69bdc37793105cdb8b08e7ae1R77) for getting the `client` value to send as part of the headers.  Long story short, we don't want to pass this value if the selected schema doesn't match the Sender's designated schema because of [the `allowDuplicates` value on the Sender](https://github.com/CDCgov/prime-reportstream/blob/8726c72903f817ab7bcb80dcd096d47de6590a84/prime-router/src/main/kotlin/azure/ValidateFunction.kt#L100).

Let me know what y'all think!

frontend-react/src/components/FileHandlers/FileHandler.tsx
penny-lischer: I'll show @chris-kuryak how the schema is returned in App Insights and let you know if we need to rework.

frontend-react/src/senders/hooks/UseSenderSchemaOptions/index.ts
```diff
@@ -0,0 +1,70 @@
...
+import { FileType } from "../../../hooks/UseFileHandler";
+
+export enum StandardSchema {
+    CSV = "upload-covid-19",
+    CSV_OTC = "csv-otc-covid-19",
+    HL7 = "standard/standard-covid-19",
```
penny-lischer, line 7: Just a reminder to change this to 'standard-covid-19'

stephenkao: Ah, this should actually be `standard/standard-covid-19` -- it's based on the filepaths in [metadata/schemas](https://github.com/CDCgov/prime-reportstream/tree/master/prime-router/metadata/schemas), just normalized/lowercased.  So `WATERS/waters-covid-19` is `waters/waters-covid-19`, `Experity/experity-covid-19`  is `experity/experity-covid-19`, `standard/standard-covid-19` is `standard/standard-covid-19`, et cetera.

You can check out `schemaStore.keys` in [Metadata.kt](https://github.com/CDCgov/prime-reportstream/blob/master/prime-router/src/main/kotlin/Metadata.kt#L201-L203) to check up on the normalized names:

Screenshot 2023-02-08 at 20 53 34

frontend-react/src/hooks/UseFileHandler.ts
```diff
@@ -223,18 +233,40 @@ function reducer(
                 payload as RequestCompletePayload
             );
             return { ...state, ...requestCompleteState };
+        case FileHandlerActionType.SCHEMA_SELECTED:
+            const selectedSchemaOption = payload as SchemaOption | null;
+
+            // reset anything related to the file if the selected schema format
```
stephenkao, line 239: @penny-lischer Updated to only clear file stuff if the format is different from the selected file, and `fileInputResetValue` below should prevent that batch file issue we were running into, but I can't reproduce it locally to check.  Let me know what you think!

jpandersen87 (approved):
Sonar is a little rude and thinks your code is smelly but i'll let it pass :P

frontend-react/src/hooks/UseFileHandler.ts
penny-lischer: It's working now but I still think the user experience is wonky and maybe the dropdown to select a schema should be placed above the 'Choose from folder' section. Design question.

penny-lischer (approved):
[https://www.lgtmgifs.com/gifs/yugi-thumb](url)

stephenkao:
We're okay holding off on further visual changes till the design is finalized so I'm gonna mergeeeeee 🚀 🚀 🚀 

Thanks, everyone!

PR #8208, opened 2023-02-06
Provide memberType boolean checks from SessionProvider#7987

### LINKED ISSUE #7987: Provide memberType boolean checks from SessionProvider
## Problem statement

We use the check `if (activeMembership.memberType === "...") { ... }` in numerous places within the React app. **To consolidate the logic, we can provide booleans from the session level that do this check for us**, and use those in places we need to check member types, such as `AuthElement`. 

## What you need to know

Stems from this discussion: https://github.com/CDCgov/prime-reportstream/pull/7923#discussion_r1069614277

## Acceptance criteria

- [ ] SessionProvider can be utilized to retrieve boolean checks for member types
- [ ] Provided booleans are used in place of current conditionals within our React app

## To do

- [ ] ...

jpandersen87:
This PR adds "isX" membership helper boolean properties to the session context. Fixes #7987


penny-lischer (approved):
LGTM

frontend-react/src/utils/PermissionsUtils.ts
```diff
@@ -35,4 +35,18 @@ const permissionCheck = (
...
 
+export interface RSUserPermissions {
+    isUserAdmin: boolean;
+    isUserSender: boolean;
+    isUserReceiver: boolean;
+}
+
+export function getUserPermissions(user?: RSUserClaims): RSUserPermissions {
+    return {
+        isUserAdmin: !!user?.organization.find((o) => isAdmin(o)),
```
stephenkao, line 46: Ah, I think `organization` is on `accessToken.claims` as opposed to `idToken.claims` (where we're getting `user` [above](https://github.com/CDCgov/prime-reportstream/pull/8208/files#diff-d10ebe8bd144dc544c78e474f801eb75d711a874b34f565b7c3cd6119848685aR60)), right?

stephenkao: Also, VERY PEDANTIC POINT, but we could simplify this to a single iteration as opposed to at most three iterations.  This is a sub-microscopic point since our users shouldn't be assigned to more than a couple of Organization at most at any time so definitely not a problem :)

frontend-react/src/utils/PermissionsUtils.ts
```diff
@@ -35,4 +35,18 @@ const permissionCheck = (
     }
 };
 
+export interface RSUserPermissions {
+    isUserAdmin: boolean;
+    isUserSender: boolean;
+    isUserReceiver: boolean;
+}
+
+export function getUserPermissions(user?: RSUserClaims): RSUserPermissions {
```
stephenkao, line 44: Would you mind adding a few tests for the different state return values?

frontend-react/src/utils/PermissionsUtils.ts
stephenkao: Oooh yeah looks like the first point here is causing a critical bug:
Screenshot 2023-02-08 at 21 27 50

stephenkao (changes requested):
Looks like there's a blocking bug [here](https://github.com/CDCgov/prime-reportstream/pull/8208/files#r1100497424) -- mind checking on that?

stephenkao (approved):
Looks great!  This'll make things super convenient!

PR #8240, opened 2023-02-07
Kotlin version update

arnejduranovic:
Update Kotlin version to 1.8.0 and removes use of deprecated enum.


jimduff-usds (approved):
Thank you!

carlosfelix2 (approved):

prime-router/src/main/kotlin/azure/CheckFunction.kt
```diff
@@ -17,8 +17,6 @@ import gov.cdc.prime.router.azure.db.enums.SettingType
 import gov.cdc.prime.router.common.BaseEngine
 import gov.cdc.prime.router.common.JacksonMapperUtilities
 import gov.cdc.prime.router.credentials.RestCredential
-import gov.cdc.prime.router.tokens.AuthenticatedClaims
```
jimduff-usds, line 20: @arnejduranovic , this change is not correct.   Some code that used these classes was erroneously commented out.   I asked Ott to correct that, and he has merged that fix into master, and now these two imports are actually needed.    So you can remove these changes to CheckFunction from your PR.

Issue #7841, opened 2023-01-04
USLink parses url for protocols


kevinhaube:
## Problem statement

The `to` prop for a `USLink` can contain many unique values, from a relative route to a `mailto:` link. When using the component, we provide an override `anchor` prop that lets us dictate whether something should render as an `a` or `Link` element. **To provide a better experience using the `USLink` component, we should implement logic to parse the following, and correctly set the element to either an anchor or Link.**

## What you need to know

Origin of this work: https://github.com/CDCgov/prime-reportstream/pull/7796#discussion_r1060902074
From the above PR comments:
- `http(s)://`
- `mailto:`
- `tel:`
- `javascript:`
- `#`

## Acceptance criteria

- [ ] USLink can auto-detect which element is needed
- [ ] USLink no longer requires `anchor` prop

## To do

- [ ] ...

kevinhaube:
@stephenkao Heads up, looks like we should move away from `javascript:` links anyway.
![Screen Shot 2023-01-06 at 10.23.24 AM.png](https://images.zenhubusercontent.com/611a907ec3e013d0e33e58a7/b3bd7e2c-0f82-4ac2-9619-8bb771cc5c54)

jpandersen87:
Closed by #8254

PR #8254, opened 2023-02-08
Systems and Settings: "On this page" navigation links don't work

### LINKED ISSUE #8204: Systems and Settings: "On this page" navigation links don't work
## Describe the bug
On the [Systems and Settings page](https://reportstream.cdc.gov/resources/system-and-settings), the "On this page" section lists several links to different sections of the page. But those links don't work and do not take you anywhere after clicking on them.

## Impact on ReportStream
Better usability on public facing side of the website.

## Steps to reproduce
1. Go to [Systems and Settings page](https://reportstream.cdc.gov/resources/system-and-settings)
2. Navigate to the "On this page:" section 
3. Click on "Data Security and Storage" for example
4. Nothing happens

## Expected behavior
When clicking on one of the title sections under "On this page:", you should be taken to that section of the page. For example, clicking on "Data access" should take you to the Data access section at the bottom of the page.

## Resources
For reference, here is the section in question.

## Logs
N/A

#### Desktop
 - OS: macOS Monterey
 - Browser: Chrome
 - Version: 109.0.5414.87 (Official Build) (arm64)

#### Smartphone
N/A

## Additional context
N/A

jpandersen87:
Fixes #8204 #7841

This PR enhances USLinks to automatically detect relative/absolute RS urls to properly use react-router's Link instead of an anchor element.


etanb:
As part of this, can remove the usage of the `anchor` prop as well just for clean up, ie:

https://github.com/CDCgov/prime-reportstream/blob/af9421f48d571c0a1256478a0c13686f3c3af86d/frontend-react/src/pages/resources/legacy-page-content/ElrChecklist.tsx#L54

frontend-react/src/components/USLink.tsx
```diff
@@ -23,14 +23,14 @@ type USNavLinkProps = Pick<AnchorHTMLAttributes<{}>, "href"> & CustomLinkProps;
...
     href,
     state,
     ...anchorHTMLAttributes
 }: USLinkProps) => {
-    return !anchor ? (
+    const isAnchor = href?.startsWith("#");
+    return !isAnchor ? (
         <Link
             to={href || ""}
             className={classnames("usa-link", className)}
```
penny-lischer, line 36: Can the 'anchor' prop be removed on ln 75?

jpandersen87: Thanks @etanb and @penny-lischer for the call out. I thought I removed that prop from the interface and was expecting typescript errors as an indicator of its use to remove but I didn't. Will fix.

frontend-react/src/components/USLink.tsx
```diff
@@ -20,17 +19,17 @@ type USNavLinkProps = Pick<AnchorHTMLAttributes<{}>, "href"> & CustomLinkProps;
...
-    anchor = false,
     children,
     className,
     href,
     state,
     ...anchorHTMLAttributes
 }: USLinkProps) => {
-    return !anchor ? (
+    const isAnchor = href?.startsWith("#");
+    return !isAnchor ? (
```
etanb, line 32: nit: can we reverse this logic so the `truthy` value renders first?

jpandersen87:
@etanb you're in luck because there was another ticket covering making all possible links to "just work" which in essence is just checking if the url is ours or not and using the appropriate component.

frontend-react/src/components/USLink.tsx
```diff
@@ -16,23 +18,31 @@ type USLinkProps = AnchorHTMLAttributes<{}> &
...
     ...anchorHTMLAttributes
 }: USLinkProps) => {
-    return !anchor ? (
+    const isRoute =
+        href && (href.startsWith("/") || RS_DOMAIN_REGEX.test(href));
+
+    return isRoute ? (
         <Link
-            to={href || ""}
+            to={href?.replace(RS_DOMAIN_REGEX, "") || ""}
```
penny-lischer, line 45: Can we add a unit test for this?

jpandersen87: Sure can. Will update asap.

jpandersen87:
@stephenkao I don't see any way to tell sonar that the three instances of non-secure http urls is for testing purposes. Should I just remove those tests?

stephenkao:
> @stephenkao I don't see any way to tell sonar that the three instances of non-secure http urls is for testing purposes. Should I just remove those tests?

Nah, I think it's worth keeping those tests in -- we can ignore SonarCloud here :)

frontend-react/src/components/USLink.tsx
```diff
@@ -72,7 +88,6 @@ export const USExtLink = ({
 }: Omit<USLinkProps, "anchor" | "rel" | "target">) => {
```
stephenkao, line 88: Let's remove `anchor` here too!

frontend-react/src/config/index.ts
```diff
@@ -16,6 +16,7 @@ const config = {
     DEFAULT_FEATURE_FLAGS: DEFAULT_FEATURE_FLAGS as string[],
     IS_PREVIEW: envVars.OKTA_URL?.match(/oktapreview.com/) !== null,
     API_ROOT: `${envVars.RS_API_URL}/api`,
+    RS_DOMAIN: "reportstream.cdc.gov",
```
stephenkao, line 19: Hmmm is this something we can add to the .env.* files as a `REACT_APP_*` variable?  I think it'd be good to keep this dynamic based on the domain the app is currently deployed on.

frontend-react/src/components/USLink.tsx
```diff
@@ -1,13 +1,15 @@
 import React, { AnchorHTMLAttributes } from "react";
 import { Link, NavLink } from "react-router-dom";
 import classnames from "classnames";
+import { escapeRegExp } from "lodash";
```
stephenkao, line 4: Ah, `lodash` isn't a direct dependency for our project so we should add either `lodash.escapeRegExp` or`lodash` to our package.json.  I'd op for the former just because I've been burned before with their bundle sizes.  I know that lodash was trying to be more modular, though, but ya know...lingering bad tastes :P

etanb: +1 to the former as well! Or just manually escape those characters :D

stephenkao:
@jpandersen87 Mind rebasing on top of `master`?  I think there are minimal changes once this is up-to-date with latest 🙏

frontend-react/src/components/USLink.tsx
```diff
@@ -17,25 +16,51 @@ type USLinkProps = AnchorHTMLAttributes<{}> &
...
 
+/**
+ * Stateless function to get route href from href that could be
+ * absolute, shorthand, relative, and/or a non-route.
+ * Attempt to parse href as URL (taking into account "//" shorthand).
+ * If it errors, then assume its a relative url (aka route). If it
+ * parses, then verify its an absolute route through origins.
+ */
+export function getHrefRoute(href?: string): string | undefined {
+    if (href === undefined) return undefined;
```
stephenkao, line 27: I think this check might be redundant since it'll be caught as an error and returned [below](https://github.com/CDCgov/prime-reportstream/pull/8254/files#diff-ca9c2c77fb047cac23c0b9a0b5740bc1f961ac5fcc83bea960f6537c3ff341ccR39), no?

frontend-react/src/components/USLink.tsx
```diff
@@ -17,25 +16,51 @@ type USLinkProps = AnchorHTMLAttributes<{}> &
...
     className,
     href,
     state,
     ...anchorHTMLAttributes
 }: USLinkProps) => {
     const sanitizedHref = href ? DOMPurify.sanitize(href) : href;
+    const routeHref = getHrefRoute(sanitizedHref);
 
-    return !anchor && sanitizedHref ? (
+    return routeHref !== undefined ? (
```
stephenkao, line 61: VERY PEDANTIC AND SUBJECTIVE POINT, but would you mind changing this to be `return Boolean(routeHref)`?  I feel like it's often easier to reason about with positive booleans, but that's just my simple brain getting confused with booleans 😆

stephenkao (approved):
Two more very, very, very minor non-blocking points, but this looks good regardless!

frontend-react/src/components/USLink.tsx
jpandersen87: TypeScript correctly gets angry at possibly passing undefined to URL, which is not allowed.

frontend-react/src/components/USLink.tsx
jpandersen87: The differentiation on undefined is actually important here. An empty string is considered the current page (aka route) but undefined means you're not using a URL (aka not a route) so you should get an anchor. Blind Boolean/negating will consider an empty string as false.

frontend-react/src/components/USLink.tsx
stephenkao: Oh, damn you perfectly correct TypeScript!  Ignore me then :)

frontend-react/src/components/USLink.tsx
stephenkao: Ah, yes -- you're definitely right here!

PR #8256, opened 2023-02-08
Move off of react-helmet

### LINKED ISSUE #8081: Move off of react-helmet
## Problem statement

[react-helmet](https://github.com/nfl/react-helmet) is a dependency we use for dynamic document header content (metatags, title).  However, it hasn't been updated in over two years, and it isn't thread-safe so we should consider moving off of it to another alternative.

## What you need to know

- [react-helmet-async](https://github.com/staylor/react-helmet-async) is an actively maintained and thread-safe fork of react-helmet, and while there are some usage differences (like needing to wrap everything in `<HelmetProvider>`), its usage is by and large the same.

## Acceptance criteria

- [ ] We no longer have `react-helmet` as a dependency and have moved to a 
- [ ] Document header content can continue to be updated dynamically

jpandersen87:
Fixes #8081

This PR replaces our usage of the abandoned react-helmet library with the forked react-helmet-async library. It also creates a "BaseWrapper" that should be used on all test renders.


frontend-react/src/utils/CustomRenderUtils.tsx
```diff
@@ -40,54 +41,80 @@ export const makeOktaHook = (_init?: Partial<IOktaContext>): OktaHook => {
...
+            <BaseWrapper>
+                <QueryClientProvider client={client}>
+                    <AuthorizedFetchProvider initializedOverride={true}>
+                        {children}
+                    </AuthorizedFetchProvider>
+                </QueryClientProvider>
+            </BaseWrapper>
         );
 
+export const BaseWrapper = ({ children }: PropsWithChildren<{}>) => (
```
stephenkao, line 78: Honestly, I feel like it'd simplify a lottttt if we just make one big wrapper that includes all Providers with sensible default wherever needed.  I mean, that's how every literally every component is rendered on our site so I can't imagine that this'd be too problematic, but I've learned to not make assumptions in this codebase 😆 

What are your thoughts about this, though?  No changes needed here -- just thinking aloud!

jpandersen87: I agree and am in favor of going that route if someone with better testing experience than I sees that as best.

frontend-react/package.json
```diff
@@ -30,7 +30,7 @@
         "react": "<18.0.0",
         "react-device-detect": "^2.0.0",
         "react-dom": "<18.0.0",
-        "react-helmet": "^6.1.0",
```
stephenkao, line 33: Good riddance, ya thread-unsafe animal!!!!!

frontend-react/src/components/AppWrapper.tsx
```diff
@@ -30,19 +31,24 @@ export const AppWrapper = ({
...
-                            <FeatureFlagProvider>
-                                {children}
-                            </FeatureFlagProvider>
-                        </AuthorizedFetchProvider>
-                        <ReactQueryDevtools initialIsOpen={false} />
-                    </QueryClientProvider>
-                </SessionProvider>
-            </AppInsightsProvider>
-        </Security>
+        <HelmetProvider>
```
stephenkao, line 34: 🙏

frontend-react/src/components/header/BasicHelmet.test.tsx
```diff
@@ -1,4 +1,6 @@
-import { render, waitFor } from "@testing-library/react";
```
stephenkao, line 1: This existed beforehand so I'm not commenting on the state of your pull request, but I don't really understand the purpose of this test.  Seems like we're testing basic functionality of React and react-helmet(-async) functionality, which isn't really in our purview of maintenance 🤔 

Like if either React or Helmet breaks and doesn't render this component correctly, that's not something we can fix 🤷 🤷 🤷

stephenkao (approved):
Looks and works great!  I love it when a dependency swap is (generally) 1:1 like this 🙌

frontend-react/src/components/header/BasicHelmet.test.tsx
jpandersen87: I would go as far as saying I don't understand the purpose of the BaseHelmet at all. You're not saving worthwhile time/kb on it currently vs just calling Helmet directly with a title.

stephenkao: 💯 💯 💯 💯 💯 💯 💯 

Want to add a fast-follow cleanup ticket to remove it?  😈

PR #8258, opened 2023-02-08
Unnest buttons in anchors, anchors in buttons

### LINKED ISSUE #7838: Unnest buttons in anchors, anchors in buttons
## Problem statement

We have a couple of places where a `<button>` is nested in an `<a>` or vice versa, and having an interactive element nested in another is bad practice for accessibility.

## What you need to know
This will require the engineer to look for occurrences in the codebase and make their best judgment call based on the context when converting between elements.

- If we just need a simple navigation, an anchor is likely the best choice.
- If a complex click handler is needed, a button is likely the best choice.  If a navigation is after the rest of the click, it can be done with a `history.push()` or `window.location.href = x;`

## Acceptance criteria

- [ ] All buttons nested within anchors are converted into the appropriate flattened element
- [ ] All anchors nested within buttons are converted into the appropriate flattened element

## Relevant links

- [Github discussion](https://github.com/CDCgov/prime-reportstream/pull/7796#discussion_r1060922186)

jpandersen87:
Fixes #7838

This PR:
- creates a new component (USLinkButton) that applies button styling to a link
- replaces improper link-wrapped button with USLinkButton

No other instances of improper nesting of anchor/button combinations were found.


frontend-react/src/components/USLink.tsx
```diff
@@ -53,6 +54,52 @@ export const USLink = ({
...
+    secondary,
+    accentStyle,
+    base,
+    outline,
+    inverse,
+    size,
+    unstyled,
+    ...anchorHTMLAttributes
+}: USLinkButtonProps) => {
+    const linkClassname = classnames(
```
stephenkao, line 76: `classnames` can accept an object so we don't have to use all these ternaries:

```jsx
classnames("usa-button", {
    "usa-button--secondary": secondary,
    [`usa-button--accent-${accentStyle}`]: accentStyle,
    ....
}, className);
```

frontend-react/src/components/USLink.test.tsx
```diff
@@ -55,3 +61,48 @@ describe("USNavLink", () => {
...
 });
+
+describe("USLinkButton", () => {
+    test("boolean button styles applied", () => {
+        const view = renderWithRouter(
+            <USLinkButton secondary base outline inverse unstyled>
+                Test
+            </USLinkButton>
+        );
+        expect(view.container.children[0]).toHaveClass(
```
stephenkao, line 72: Small point, but can this be found with a `getByText` or even `getByRole`?   If there's something visual to be queried for, I think we should do that instead of relying on the return renderer structure (even though that's unfortunately necessary sometimes 😬)

frontend-react/src/components/USLink.test.tsx
```diff
@@ -55,3 +61,48 @@ describe("USNavLink", () => {
...
+            const classNamePrefix = enumPropMap[key as keyof typeof enumProps];
+            const prop = { [key]: value };
+            const className =
+                classNamePrefix === ""
+                    ? `usa-button--${value}`
+                    : `usa-button--${classNamePrefix}-${value}`;
+            const view = renderWithRouter(
+                <USLinkButton {...prop}>Test</USLinkButton>
+            );
+            expect(view.container.children[0]).toHaveClass(
```
stephenkao, line 102: Same point here about the `getBy`s!

frontend-react/src/components/USLink.tsx
jpandersen87: I knew there was a cleaner way to do it, but couldn't remember. Ty!

stephenkao (approved):
Minor points as always, but nothing blocking.  Thanks for the fix!

frontend-react/src/components/USLink.tsx
```diff
@@ -42,50 +43,85 @@ export function getHrefRoute(href?: string): string | undefined {
...
 
-/** A single link for rendering standard links. Uses a `Link` by default
- * but adding `anchor` will make this a generic anchor tag.
- * @example
- * <USLink href="/page">To Page</USLink> // uses <Link> from react-router-dom
- * <USLink href="#this-section-on-my-page">To Section</USLink> // uses <a>
- * */
-export const USLink = ({
+export interface SafeLinkProps extends React.AnchorHTMLAttributes<Element> {
+    fallbackComponent?: React.FunctionComponent<
```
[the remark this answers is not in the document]
jpandersen87: Leftover from early iteration work. Will remove.

PR #8261, opened 2023-02-08
7933 Submission history API supports bulk FHIR submissions

### LINKED ISSUE #7933: Submission history API supports bulk FHIR submissions
### User Story:
As sender, I want to be able to retrieve submission details for FHIR bulk submissions, so I can get accurate reports of my submitted data.

### Description/Use Case
The submissions history API and submission detail API need to provide accurate data on FHIR bulk submissions, so they show all the reports included in the bulk submission. This is similar to having the system receive an HL7 batch or CSV file with multiple lines.

* The queries for the API may need tweaking to support this.

### Dev Notes:
* This is similar to having the system receive an HL7 batch or CSV file with multiple lines, so look at what those responses contain and make sure the FHIR bulk submission version of it looks similar.

### Acceptance Criteria
* A sender is capable of retrieving submission history (the list of submissions) that includes FHIR bulk submissions.
* A sender is capable of retrieving submission details for a given FHIR bulk submission that includes all related FHIR bundles in the original submission (how does this show up in the response?).

victor-chaparro:
This PR updates the Submission History API to support bulk FHIR submissions. Currently the API would return each report as a separate destination. SubmissionHistory was updated to consolidate all the reports into their respective destination.

Test Steps:
1. Send a bulk FHIR submission with two or more bundles via the reports API.
2. Call the Submission History API with the report id you submitted
3. Verify bundles are being routed to the right destination, and no duplicate destinations are created

## Changes
- Updated `enrichWithRoutingAndTranslationActions` in `SubmissionHistory.kt` to handle multiple reports
- Added unit tests for new added functionality

### Testing
- [x] Tested locally?
- [ x Added tests?

## Linked Issues
- Fixes #7933


prime-router/src/main/kotlin/history/SubmissionHistory.kt
```diff
@@ -341,41 +341,61 @@ class DetailedSubmissionHistory(
...
+                    val index = indexOfFirst {
+                        it.organizationId == receiverNameSegments[0] && it.service == receiverNameSegments[1]
+                    }
+                    if (index >= 0) {
+                        get(index).filteredReportRows =
+                            get(index).filteredReportRows.add(filterReport)
+                        get(index).filteredReportItems =
+                            get(index).filteredReportItems.add(ReportStreamFilterResultForResponse(filterResult))
+                        get(index).itemCountBeforeQualFilter = get(index).itemCountBeforeQualFilter?.plus(
+                            filterResult.originalCount
```
jimmyfagan, line 383: This might be more readable by saving the destination at this index to a local variable like `existingDestination` instead of using `get(index)`.

prime-router/src/main/kotlin/history/SubmissionHistory.kt
```diff
@@ -341,41 +341,61 @@ class DetailedSubmissionHistory(
...
+                    val receiverNameSegments = filterResult.receiverName.split(Sender.fullNameSeparator)
+                    val index = indexOfFirst {
+                        it.organizationId == receiverNameSegments[0] && it.service == receiverNameSegments[1]
+                    }
+                    if (index >= 0) {
+                        get(index).filteredReportRows =
+                            get(index).filteredReportRows.add(filterReport)
+                        get(index).filteredReportItems =
+                            get(index).filteredReportItems.add(ReportStreamFilterResultForResponse(filterResult))
+                        get(index).itemCountBeforeQualFilter = get(index).itemCountBeforeQualFilter?.plus(
```
jimmyfagan, line 382: If I'm reading this correctly, if `itemCountBeforeQualFilter` is null, it'll stay null. Is that what we want, or should we set it to 1 in that case?

prime-router/src/main/kotlin/history/SubmissionHistory.kt
```diff
@@ -341,41 +341,61 @@ class DetailedSubmissionHistory(
...
-            val filteredReportRows = filterLogs.map { log -> log.detail.message }
-            val filteredReportItems = filterLogs.map { log ->
-                ReportStreamFilterResultForResponse(log.detail as ReportStreamFilterResult)
-            }
+        // local function to safe add elements to nullable immutable list
+        fun <T> List<T>?.add(element: T): List<T> {
+            val tempMutableList = this?.toMutableList() ?: arrayListOf()
+            tempMutableList.add(element)
+            return tempMutableList
+        }
```
jimmyfagan, line 366: Do the lists we're using this for need to stay immutable? If not, maybe we can make them mutable and use the normal `add`. If they do need to stay immutable, it might avoid some confusion to name this `addToImmutableList` or something, to make it clear that if a `mutableList` calls `add` it'll still use the usual one. That's 100% optional, I just got kind of confused reading this earlier because I thought this function was being used for the `add` at line 355.

jimmyfagan (commented):
Looks good! I left some comments and questions, I'll re-review to approve it after you get a chance to respond to those.

prime-router/src/main/kotlin/history/SubmissionHistory.kt
```diff
@@ -341,41 +341,61 @@ class DetailedSubmissionHistory(
...
-                        destinations[index].itemCountBeforeQualFilter?.plus(
-                            descendantDest.itemCountBeforeQualFilter ?: 0
-                        )
-                } else {
-                    destinations += descendantDest
+                with(destinations) {
+                    val index = indexOfFirst {
+                        it.organizationId == descendantDest.organizationId && it.service == descendantDest.service
+                    }
+                    if (index >= 0) {
```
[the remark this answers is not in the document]
carlosfelix2: index is the item index which starts at 1. There was a big effort a while back to standardize the code to use 1 instead of 0, so the database and API responses match.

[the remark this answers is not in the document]
carlosfelix2: >=0 is right as the index search in line 345 can return a 0.

prime-router/src/main/kotlin/history/SubmissionHistory.kt
```diff
@@ -341,41 +341,61 @@ class DetailedSubmissionHistory(
...
-                if (index >= 0) {
-                    destinations[index].itemCount += descendantDest.itemCount
-                    destinations[index].itemCountBeforeQualFilter =
-                        destinations[index].itemCountBeforeQualFilter?.plus(
-                            descendantDest.itemCountBeforeQualFilter ?: 0
-                        )
-                } else {
-                    destinations += descendantDest
+                with(destinations) {
+                    val index = indexOfFirst {
```
carlosfelix2, line 345: This is OK, but you could use a filter or firstOrNull instead of dealing with the index number

prime-router/src/main/kotlin/history/SubmissionHistory.kt
```diff
@@ -341,41 +341,61 @@ class DetailedSubmissionHistory(
...
-        // to return the expected output, we create dummy destinations that are only used in enrichment
-        // these dummies are created using the receivers defined in the logs
-        if (destinations.isEmpty() && filterLogs.isNotEmpty()) {
-            val filteredReportRows = filterLogs.map { log -> log.detail.message }
-            val filteredReportItems = filterLogs.map { log ->
-                ReportStreamFilterResultForResponse(log.detail as ReportStreamFilterResult)
-            }
+        // local function to safe add elements to nullable immutable list
+        fun <T> List<T>?.add(element: T): List<T> {
+            val tempMutableList = this?.toMutableList() ?: arrayListOf()
```
carlosfelix2, line 363: Should arrayListOf() really be emptyList()?

prime-router/src/main/kotlin/history/SubmissionHistory.kt
```diff
@@ -341,41 +341,61 @@ class DetailedSubmissionHistory(
...
 
-        // if all the items were filtered out, no translate actions are generated (thus no destinations)
-        // to return the expected output, we create dummy destinations that are only used in enrichment
-        // these dummies are created using the receivers defined in the logs
-        if (destinations.isEmpty() && filterLogs.isNotEmpty()) {
-            val filteredReportRows = filterLogs.map { log -> log.detail.message }
-            val filteredReportItems = filterLogs.map { log ->
-                ReportStreamFilterResultForResponse(log.detail as ReportStreamFilterResult)
-            }
+        // local function to safe add elements to nullable immutable list
```
carlosfelix2, line 361: Change this comment to be kdoc for the function

prime-router/src/main/kotlin/history/SubmissionHistory.kt
```diff
@@ -341,41 +341,61 @@ class DetailedSubmissionHistory(
...
-                        null,
-                        0,
-                        filterResult.originalCount,
-                    )
-                )
+                with(destinations) {
+                    val filterResult = log.detail as ReportStreamFilterResult
+                    val filterReport = log.detail.message
+                    val receiverNameSegments = filterResult.receiverName.split(Sender.fullNameSeparator)
+                    val index = indexOfFirst {
```
carlosfelix2, line 374: Same comment as above, you could use a filter or firstOrNull here.

prime-router/src/main/kotlin/history/SubmissionHistory.kt
```diff
@@ -341,41 +341,61 @@ class DetailedSubmissionHistory(
...
-                        receiverNameSegments[1],
-                        filteredReportRows,
-                        filteredReportItems,
-                        null,
-                        0,
-                        filterResult.originalCount,
-                    )
-                )
+                with(destinations) {
+                    val filterResult = log.detail as ReportStreamFilterResult
```
carlosfelix2, line 371: Maybe add a `check` here to make sure log.detail is indeed that type

prime-router/src/main/kotlin/history/SubmissionHistory.kt
```diff
@@ -341,41 +341,61 @@ class DetailedSubmissionHistory(
...
-                            descendantDest.itemCountBeforeQualFilter ?: 0
-                        )
-                } else {
-                    destinations += descendantDest
+                with(destinations) {
+                    val index = indexOfFirst {
+                        it.organizationId == descendantDest.organizationId && it.service == descendantDest.service
+                    }
+                    if (index >= 0) {
+                        get(index).itemCount += descendantDest.itemCount
```
carlosfelix2, line 349: Don't repeat get(index). Set it to a variable or replace the index with a filter

prime-router/src/main/kotlin/history/SubmissionHistory.kt
```diff
@@ -341,41 +341,61 @@ class DetailedSubmissionHistory(
...
-                )
+                with(destinations) {
+                    val filterResult = log.detail as ReportStreamFilterResult
+                    val filterReport = log.detail.message
+                    val receiverNameSegments = filterResult.receiverName.split(Sender.fullNameSeparator)
+                    val index = indexOfFirst {
+                        it.organizationId == receiverNameSegments[0] && it.service == receiverNameSegments[1]
+                    }
+                    if (index >= 0) {
+                        get(index).filteredReportRows =
```
carlosfelix2, line 378: Don't repeat get(index). Set it to a variable or replace the index with a filter

prime-router/src/main/kotlin/history/SubmissionHistory.kt
carlosfelix2: ☝️

prime-router/src/test/kotlin/history/SubmissionHistoryTests.kt
```diff
@@ -1225,4 +1206,118 @@ class SubmissionHistoryTests {
...
+                    TaskAction.route,
+                    OffsetDateTime.now(),
+                    HttpStatus.OK.value(),
+                    null,
+                    logs
+                ),
+            )
+        )
+
+        testEnrich.run {
```
carlosfelix2, line 1314: This feels you are only testing happy day scenarios. Are there any error states to test? The code coverage shows you are missing some branches of the code. Consider refactoring some of the code out to its own functions if that makes it easier for testing.

carlosfelix2 (changes requested):
Does this need for unit tests? Otherwise it looks good

prime-router/src/main/kotlin/history/SubmissionHistory.kt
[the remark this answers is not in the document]
jimmyfagan: If it returns a 0, wouldn't that mean we found the destination at the first position in the list? I guess my question is more about why we are adding a new destination when we find an existing one at index `0`, instead of only adding one if line 345 returns `-1`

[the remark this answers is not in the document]
victor-chaparro: This was updated to use firstOrNull, but the if statement  was including 0, so if it returns 0 it would update that destination not add a new one.

prime-router/src/main/kotlin/history/SubmissionHistory.kt
victor-chaparro: Good call, updated to use existingDestination

prime-router/src/main/kotlin/history/SubmissionHistory.kt
victor-chaparro: This is increasing the existing `itemCountBeforeQualFilter` by whatever `itemCountBeforeQualFilter` is on the next destination. itemCountBeforeQualFilter can be null on the next destination and you can't add null so have to add 0 if null.

prime-router/src/main/kotlin/history/SubmissionHistory.kt
victor-chaparro: They don't have to stay immutable, that's why I was wondering if it was bad practice to have mutable lists as properties. Maybe we can discuss that at standup to see what everybody thinks about that. I updated them to mutable and it does it make it simpler. But if that turns out to be problem I can update them back.

prime-router/src/main/kotlin/history/SubmissionHistory.kt
victor-chaparro: Using firstOrNull now

prime-router/src/main/kotlin/history/SubmissionHistory.kt
victor-chaparro: Updated list to be mutable and longer need this function. Let me know if I need to change them back to immutable.

prime-router/src/main/kotlin/history/SubmissionHistory.kt
victor-chaparro: No longer need this function if lists are mutable.

prime-router/src/main/kotlin/history/SubmissionHistory.kt
victor-chaparro: using firstOrNull now

prime-router/src/main/kotlin/history/SubmissionHistory.kt
victor-chaparro: Added check

prime-router/src/main/kotlin/history/SubmissionHistory.kt
victor-chaparro: Using firstOrNull now.

prime-router/src/test/kotlin/history/SubmissionHistoryTests.kt
victor-chaparro: Refactored code into separate functions to make it more readable and consistent with the other enrich functions.  When I debug the unit test, it shows it goes through all the branches of the function, but I also see the coverage is missing some of that. So not sure what's happening there...

prime-router/src/main/kotlin/history/SubmissionHistory.kt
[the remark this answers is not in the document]
jimmyfagan: Oh wow, with fresh eyes I'm seeing that again and it makes perfect sense, I think I'm just used to seeing `index > -1` and got confused, that was my bad 😅

prime-router/src/main/kotlin/history/SubmissionHistory.kt
jimmyfagan: Gotcha, let me start a new convo with the updated code, still have a question on this.

prime-router/src/main/kotlin/history/SubmissionHistory.kt
jimmyfagan: Sounds like a plan 👍

prime-router/src/main/kotlin/history/SubmissionHistory.kt
```diff
@@ -321,61 +326,79 @@ class DetailedSubmissionHistory(
...
+
+                    if (existingDestination.filteredReportItems == null) {
+                        existingDestination.filteredReportItems = mutableListOf(filterResultResponse)
+                    } else {
+                        existingDestination.filteredReportItems!!.add(filterResultResponse)
+                    }
+
+                    existingDestination.itemCountBeforeQualFilter = existingDestination.itemCountBeforeQualFilter?.plus(
+                        filterResult.originalCount
                     )
```
jimmyfagan, line 388: In my other comment, what I was getting at is that if `existingDestination.itemCountBeforeQualFilter` is null, then `existingDestination.itemCountBeforeQualFilter?.plus(anything)` is going to be null as well. I think the change below would increment it treating a starting point of null as 0. It might be that we don't expect it to ever be null, if so then this isn't necessary.
```suggestion
                    existingDestination.itemCountBeforeQualFilter = (existingDestination.itemCountBeforeQualFilter ?: 0).plus(
                        filterResult.originalCount
                    )
```

victor-chaparro: I see that now. It would so much easier if these weren't nullables. I added an extra `?:` to set it to `filterResult.originalCount` if null. I did the same for the other case.

jimmyfagan (approved):

carlosfelix2 (approved):

PR #8263, opened 2023-02-08
prime cli fake input generation

### LINKED ISSUE #7897: prime cli fake test file generation is broke.
Using csv-otc-covid-19.schema, The command Fake file generates wrong values for Patient_state and Patient_county.  It also adds extra columns patient_state to the file.

To duplicate the issue run the following command:
./prime data --input-fake 2 --input-schema csv-otc-covid-19 --output PATH/TO/pdi-covid-19-ri.csv --target-states RI

Generated Output file contains wrong Patient_state and Patient_county instead of RI.

adegolier:
This PR ...

Test Steps:
1. Run command `./prime data --input-fake 50 --input-schema csv-otc-covid-19 --output PATH/TO/pdi-covid-19-ri.csv --target-states RI`
2. Open the output file.
3. Verify that the the produced csv file contains only one patient_state column, that it is RHODE ISLAND.
4. Verify that the patient_county column is mostly actual counties with the occasional random string.

## Changes
- Changed CsvSerializer.kt to only include one (1) of each named csv field and to prefer non-TEXT types when available.
- Changed FakeReport.kt to properly choose target state from a valueset
### Testing
- [X] Tested locally?
- [ ] Added tests?

## Linked Issues
- Fixes #7897


prime-router/src/main/kotlin/FakeReport.kt
```diff
@@ -35,41 +35,6 @@ class FakeDataService : Logging {
         context: FakeReport.RowContext,
     ): String {
         val faker = context.faker
-        // creates fake text data
```
adegolier, line 38: Note: This wasn't actually deleted, I just moved it to the bottom so I could call other generation functions from within it

brick-green (commented):
checked the output file. These two fields (equipment_model_name, test_performed_code) should have values from the LIVD table.

Then I saw that processing_mode_code had values of "ACTIVE". That seems like it's using the display value from the altValueSet. Instead it should use the code.

adegolier:
Added special handling for equipment_model_name and test_performed_code when they are TEXT type fields, but leaving ACTIVE as is. Went over it with @GilmoreA6 and we came to the conclusion that ACTIVE is the right choice there because for fake data (i.e. the file we'd receiver from a sender) we'd see the "display", for the internal csv is when we'd be seeing the "code".

brick-green (approved):
Looks good!

PR #8272, opened 2023-02-09
Jessica/8094 add condition filtering setting

JessicaWNava:
This PR makes it so that an error is thrown if organizations.yml has a conditionFilter added to a receiver under a topic that is anything other than full-elr.

Test Steps:
1. Add the following to a full-elr receiver in organizations.yml. Run ./gradlew reloadSettings, it should succeed.
        conditionFilter:
                - "fail"
2. Do the same, but for a topic that is not full-elr. Run ./gradlew reloadSettings, it should fail.

## Changes
 -Added a condition filter and have it fail when it is added to a receiver with a topic other than full-elr

### Testing
- [x] Tested locally?
- [x] Added tests?

### Process

## Linked Issues
- Fixes #issue 8094

## To Be Done
- #issue

- Does this PR introduce new endpoints? No
    - new endpoint A
    - new endpoint B
- Does this PR include changes in authentication and/or authorization of existing endpoints? No
- Does this change introduce new dependencies that need vetting? no
- Does this change require changes to our infrastructure? no
- Does logging contain sensitive data? no
- Does this PR include or remove any sensitive information itself? no


prime-router/src/main/kotlin/Translator.kt
```diff
@@ -133,6 +133,17 @@ class Translator(private val metadata: Metadata, private val settings: SettingsP
...
+            organization,
+            ReportStreamFilterType.CONDITION_FILTER,
+            trackingElement,
+            doLogging = true
+        )
+        if (conditionFilteredReport.isEmpty()) return conditionFilteredReport
+
         // Do processingModeFiltering on the routingFilteredReport
         val processingModeFilteredReport = filterByOneFilterType(
             routingFilteredReport,
```
jimmyfagan, line 149: This may not matter since I _think_ the list should be empty at this point either way, but this should maybe now be
```suggestion
            conditionFilteredReport,
```
for consistency with each of the other calls.

prime-router/src/main/kotlin/Receiver.kt
```diff
@@ -24,6 +24,7 @@ import java.time.ZoneId
  * receiver does not want, based on who sent it.  However, it's available for any general purpose use.
  * @param processingModeFilter defines the filters that is normally set to remove test and debug data.
  * @param reverseTheQualityFilter If this is true, then do the NOT of 'qualityFilter'.  Like a 'grep -v'
+ * @param conditionFilter defines the filters that select the conditions that a STILT wants to receive
```
jimmyfagan, line 27:
```suggestion
 * @param conditionFilter defines the filters that select the conditions that a STLT wants to receive
```

prime-router/src/main/kotlin/Translator.kt
jimmyfagan: Re-read this, and no it's not empty at this point, so I think this is more of a required change.

prime-router/src/main/kotlin/Receiver.kt
```diff
@@ -247,6 +252,11 @@ open class Receiver(
                 }
             }
         }
+
+        if (topic != Topic.FULL_ELR && conditionFilter.isNotEmpty() && conditionFilter[0].isNotEmpty()) {
+            return "Invalid value for condition filter. " +
+                "Cannot have a condition filter in pipelines other than universal pipeline."
```
jimmyfagan, line 258: `Invalid value` seems to me to be a little bit misleading, the value isn't necessarily invalid, it's more that using the field at all is prohibited. I might prefer a message like `Condition filter only allowed for receivers with topic "full_elr", but one was found for receiver $fullName`. That's not great either, so consider this optional, what you have is probably fine.

prime-router/src/main/kotlin/Receiver.kt
```diff
@@ -247,6 +252,11 @@ open class Receiver(
                 }
             }
         }
+
+        if (topic != Topic.FULL_ELR && conditionFilter.isNotEmpty() && conditionFilter[0].isNotEmpty()) {
```
jimmyfagan, line 256: This function overall seems kinda inefficient and confusing, but we've already checked the topic at line 236 so we probably can assume that it's not FULL_ELR here. I guess if that's likely to change, then it's good defensive programming to check again at this point, it's just hard for me to tell since I can't really understand how this function is intended to be used (It's currently only used in test cases 🤷 )

prime-router/src/test/kotlin/ReceiverTests.kt
```diff
@@ -12,6 +13,43 @@ import kotlin.test.Test
...
+            externalName = "Ignore ELR"
+        )
+
+        assertThat(
+            receiver.consistencyErrorMessage(
+                UnitTestUtils.simpleMetadata
+            )
+        ).isEqualTo(
+            "Invalid value for condition filter. " +
+                "Cannot have a condition filter in pipelines other than universal pipeline."
```
jimmyfagan, line 34: (optional) The approach to error message checking in tests seems to vary and be kind of a judgment call, but you might want to consider changing this to use `messageContains` or `isNotNull` to make this less dependent on the specific error message text.

prime-router/src/test/kotlin/ReceiverTests.kt
```diff
@@ -12,6 +13,43 @@ import kotlin.test.Test
...
+            name = "elr",
+            organizationName = "IGNORE",
+            topic = Topic.FULL_ELR,
+            customerStatus = CustomerStatus.INACTIVE,
+            conditionFilter = listOf("blah"),
+            translation = translatorConfig,
+            externalName = "Ignore ELR"
+        )
+
+        assertThat(receiver.consistencyErrorMessage(UnitTestUtils.simpleMetadata)).isEqualTo(null)
```
jimmyfagan, line 50: I see both `isEqualTo(null)` and `isNull()` used in the repo, but `isNull()` is probably a little more concise.

prime-router/src/test/kotlin/fhirengine/engine/fhirRouterTests/GetFilterTests.kt
```diff
@@ -85,14 +85,16 @@ class GetFilterTests {
...
+            conditionFilter = listOf("testOrgCondition")
         ),
         ReportStreamFilters(
             topic = Topic.COVID_19,
             jurisdictionalFilter = listOf("testCovidJuris"),
             qualityFilter = listOf("testCovidQuality"),
             routingFilter = listOf("testCovidRouting"),
-            processingModeFilter = listOf("testCovidProcMode")
+            processingModeFilter = listOf("testCovidProcMode"),
+            conditionFilter = listOf("testCovidCondition")
```
jimmyfagan, line 97: Why was this needed?

jimmyfagan (commented):
Left a few comments/questions (some were left as "single comments" outside of the review by mistake), but overall this looks good!

prime-router/src/main/kotlin/Receiver.kt
JessicaWNava: So this function gets used when we call ./gradlew reload tables. I kept the FULL_ELR check since there is a todo labeling the above check as temporary.

prime-router/src/test/kotlin/ReceiverTests.kt
JessicaWNava: I can't get messageContains to work for some reason, and I don't feel like isNotNull is enough since it should be returning a different error than the one checked for so I think I am going to stick with what I have. It would be really nice if we used a different test library that had better support for things like this.

prime-router/src/main/kotlin/Receiver.kt
jimmyfagan: Good call on planning for the TODO, but I'm having a hard time verifying that this function is used in the gradle tasks, I can't seem to hit breakpoints in this function when I run reload tables or settings. I do see that it fails to reload settings when a covid receiver has a condition filter, but I'm still not seeing anything go through this function still. Any tips?

prime-router/src/test/kotlin/fhirengine/engine/fhirRouterTests/GetFilterTests.kt
jimmyfagan: Ah I see, if you update `ReportStreamFilters` primary constructor in `ReportStreamFilter.kt` to set a default value for `conditionFilter` to be `null`, then you can omit this in this case and in some of the other cases where you're explicitly setting it to null.

prime-router/src/main/kotlin/Receiver.kt
JessicaWNava: For me, it only hit when I ran it from the command line. I think running it from IntelliJ kills the Attach to Container process because it can only run one at a time.

prime-router/src/test/kotlin/fhirengine/engine/fhirRouterTests/GetFilterTests.kt
JessicaWNava: Good call, thanks for that. Sorry about the bolding. Idk what that was about.

prime-router/src/main/kotlin/Receiver.kt
jimmyfagan: Ah okay, that worked, thanks!

prime-router/src/main/kotlin/ReportStreamFilter.kt
```diff
@@ -44,6 +45,7 @@ data class ReportStreamFilters(
     val qualityFilter: ReportStreamFilter?,
     val routingFilter: ReportStreamFilter?,
     val processingModeFilter: ReportStreamFilter?,
+    val conditionFilter: ReportStreamFilter? = null
```
jimmyfagan, line 48: Missed this the first time, let's just add some kdoc for this new param.

jimmyfagan (approved):
Looks good, one minor kdoc catch, but no other comments at this point.

carlosfelix2 (approved):

prime-router/settings/organizations.yml
```diff
@@ -186,6 +187,8 @@
       topic: full-elr
       customerStatus: active
       jurisdictionalFilter: [ "%performerState = 'IG' or %patientState = 'IG'" ]
+      conditionFilter:
```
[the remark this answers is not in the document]
JessicaWNava: Shouldn't it fail if anything is put in a condition filter that is not full-elr and not fail if there is a filter for full-elr? This should not have been checked in, that was just test data.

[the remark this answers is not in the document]
carlosfelix2: Yes, but the problem is that fail is not a FHIR Path expression and hence you will get an exception when you start using that condition. So, "fail" is an incorrect expression. Not like it matters a lot right this sec, but it is incorrect.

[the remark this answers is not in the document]
JessicaWNava: Carlos Slack response: Yes, but the problem is that fail is not a FHIR Path expression and hence you will get an exception when you start using that condition. So, "fail" is an incorrect expression. Not like it matters a lot right this sec, but it is incorrect.

PR #8283, opened 2023-02-09
Remove replaceShorthand and update resolveConstant to return a list

### LINKED ISSUE #8106: Remove custom UP code to support constants with multiple FHIR values
### User Story:
As a developer, I do not want duplicate code if a library has the functionality, so our code is clean.

### Description/Use Case
We had to implement a piece of custom code, so we could support having FHIR constants that resulted in multiple FHIR values. The FHIR structure library has been updated, so constants can support multiple values and hence we no longer need the custom code. Remove the custom code in favor of the code in the library.

* Need to make sure that the library supports our use case 100%

### Dev Notes:
* FhirPathCustomResolver.resolveConstant needs to support returning collections
* Need to remove the custom code that is in FHIRRouter.kt related to the FHIR short hands begin replaced in the string.  The short hands are now to be passed as constants in the app context for the FHIR Path expression evaluations.
* Need to make sure that the short hands work as expected in the FHIR Path expressions.
* Need to remove unit tests that relate to the custom code

### Acceptance Criteria
* FHIR Path Short Hand constants defined for routing filters are available for substitution in FHIR Path expressions used in the routing filters.
*  FHIR Path Short Hand constants defined for routing filters can support collections.

arnejduranovic:
This PR removes `replaceShorthand` and lets the FHIR engine do the work by ensuring the constants are passed to the appContext. It also updates `FhirPathCustomResolver.resolveConstant` to allow the return of multiple values, instead of just taking the first value.

Test Steps:
1. Run the following via postman and set the app to debug. See that `resolveConstant` returns the correct values for state (IG).
[Tests.postman_collection.txt](https://github.com/CDCgov/prime-reportstream/files/10701871/Tests.postman_collection.txt)

## Changes
- remove `replaceShorthand` function
- update `FhirPathCustomResolver.resolveConstant`

### Testing
- [x] Tested locally?
- [x] Added tests? - I determined no additional unit tests were necessary, the existing tests cover everything related to `resolveConstant`

## Linked Issues
- Fixes #8106


prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/ConstantResolver.kt
```diff
@@ -130,24 +129,25 @@ class FhirPathCustomResolver : FHIRPathEngine.IEvaluationContext, Logging {
...
-                // TODO: Support multiple values with the updated function return type and remove our custom solution
-                throw SchemaException("Constant $name must resolve to one value, but had ${values.size}.")
             } else {
-                logger.trace("Evaluated FHIR Path constant $name to: ${values[0]}")
-                // Convert string constants that are whole integers to Integer type to facilitate math operations
-                if (values[0] is StringType && StringUtils.isNumeric(values[0].primitiveValue())) {
-                    IntegerType(values[0].primitiveValue())
-                } else values[0]
+                logger.trace("Evaluated FHIR Path constant $name to: $values")
+                val cleanedValues = mutableListOf<Base>()
```
arnejduranovic, line 139: If anyone has a nicer way to do this chunk of code in Kotlin let me know

PatriciaPerozoUSDS: I don't have a better way to do this but I do think its worth adding this comment `// Convert string constants that are whole integers to Integer type to facilitate math operations` back to explain why you're doing this.

PatriciaPerozoUSDS (commented):
I think it's worth adding a unit test for the case when FHIRPathUtils.evaluate actually returns multiple values. Its new functionality that should be reflected in the the unit test spec

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/ConstantResolver.kt
jimmyfagan: I think you could do something like 
```
val cleanedValues = values.map {
    if (it is StringType && ...) {
        IntegerType(it.primitiveValue()
    } else {
        it
    }
}
```
Also at that point, you probably don't need to even save it to a new val `cleanedValues`, you could just leave that expression.

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/ConstantResolver.kt
```diff
@@ -130,24 +129,25 @@ class FhirPathCustomResolver : FHIRPathEngine.IEvaluationContext, Logging {
...
+                for (value in values) {
+                    if (value is StringType && StringUtils.isNumeric(values[0].primitiveValue())) {
```
jimmyfagan, line 141: Was the intention for this to be `value` instead of `values[0]`?

jimmyfagan (commented):
Looks good, left a few comments for small ways to make this a little more readable/clean, and one thing that _might_ be a mistake with a quick fix. I think I need to read the ticket more to get a refresher on the reasoning for this, but I'll do that outside of the review.

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/ConstantResolver.kt
carlosfelix2, line 141: You need to add unit tests to test this code with multiple values both strings that can be numeric and not numeric.

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/ConstantResolver.kt
```diff
@@ -130,24 +129,25 @@ class FhirPathCustomResolver : FHIRPathEngine.IEvaluationContext, Logging {
         }
 
         // Evaluate the constant before it is used.
-        val value = if (constantValue.isNullOrBlank()) null
+        val values = if (constantValue.isNullOrBlank()) null
         else {
```
carlosfelix2, line 133: Add a unit test to test that you can actually get a collection out of this function.

carlosfelix2 (changes requested):

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/ConstantResolver.kt
arnejduranovic: There already exist unit tests that check the types:

`assertThat(result[0] is IntegerType).isTrue()`

`assertThat(result[0]).isInstanceOf(StringType::class.java)`

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/ConstantResolver.kt
arnejduranovic: To make sure I understand this request, you want to add a test that tests that a third-party library, `FHIRPathEngine`, can return multiple values, even if its signature says it returns a list? In my opinion this test would test a third party library, something I think we shouldn't be doing. I do think we should add another test, per Patricia's comment. In this test, I would mock `FhirPathUtils.evaluate` to return two items: a numeric String and a non-Numeric String and test that our method, `resolveConstant`, handles it correctly. Do you agree? Any additional thoughts?

carlosfelix2: Nope. Don't test the third party library, but you have code here that used to convert one string to a numeric value and not can do multiple. So you need some test to test that. You always assume the third party library works. Here you will need some FHIR Path that will return a collection of string that have numbers on it. Let me know if you need help there.

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/ConstantResolver.kt
arnejduranovic: Yes, this is a bug, the new unit test I just added should catch this in the future. Thanks for all the kotlin tips!

arnejduranovic: Also check this in the new unit test I wrote.

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/ConstantResolver.kt
arnejduranovic: So I just committed a new test called `test fhir path resolver multiple values`. I think that's what you're asking for but please let me know if I'm missing something.

prime-router/src/test/kotlin/fhirengine/translation/hl7/utils/ConstantResolverTests.kt
```diff
@@ -113,6 +115,29 @@ class ConstantResolverTests {
...
+    @Test
+    fun `test fhir path resolver multiple values`() {
+        val integerValue = 99
+        val stringValue = "Ninety-Nine"
+
+        mockkObject(FhirPathUtils)
+        every { FhirPathUtils.evaluate(any(), any(), any(), any()) } returns
+            listOf<Base>(StringType(stringValue), StringType(integerValue.toString()))
+
+        val constants = sortedMapOf("const1" to "'value1'") // this does not matter but context wants something
```
carlosfelix2, line 127: You could just say CustomContext()

carlosfelix2 (approved):

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/ConstantResolver.kt
jimmyfagan: Looks good in latest commit, marking this conversation resolved.

jimmyfagan (approved):

prime-router/src/test/kotlin/fhirengine/translation/hl7/utils/ConstantResolverTests.kt
arnejduranovic: I can't because I need it to get through the first part of the `resolveConstant` function, the part that calculates "constantValue". If I don't set anything, then the code we are trying to test does not run. This tells me that this function does two separate things and should be broken up to be more cohesive, but I didn't find it egregious enough to warrant a refactor/rewrite. The other tests cover the other cases and the other parts of the function. 

The code coverage report is good, tests all the lines of the class we are testing.

PR #8284, opened 2023-02-09
7545 UP Routing function sends message to Translate function even when data will go to no receivers

### LINKED ISSUE #7545: UP Routing function sends message to Translate function even when data will go to no receivers
There is a need for some queries to see the step runs.  Will need to figure out if there is anything to do here.  Potentially, the proper thing to do here is not to put a message on the translate queue, not translate task, and make sure the generate report next action is NULL.  This is not causing any issues and it is a TINY inefficiency.

### User Story:
As a administrator, I don't want the UP to send data to the Translate function if no receivers will receive data, so it is easier to track where the data went.

### Description/Use Case
The UP sends a message to the Translate function regardless if data will go to a receiver.  The result is that the Translate function does not do anything (which is not bad), but it is an unnecessary message and unnecessary resources used.

### Dev Notes:
* Just change the data or the filters, so data goes nowhere, submit it, then see in the logs how the translate function gets a message, but nothing happens.
* Ensure that the Task, Action, and Report all reflect that there is no next step needed. 
* Ensure that nothing breaks and there is no expectation of a 'translate' step for each submission going through the UP.

### Acceptance Criteria
* The UP only sends messages to the Translate function when data is to be sent a receivers.

jack-h-wang:
This PR adds an alternate routing process for FHIR bundles with no receivers that skips sending a message to the translate function, eliminating a minor inefficiency.

Test Steps:
1. Use the `Single HL7 Full ELR` test to submit data to be routed. Ensure the data sent does not have a valid receiver.
2. Retrieve the report `id` from the data returned from the API call.
3. Allow time for the data to be routed, then confirm no calls to the translate function for this report lineage on the `report_file`, `action`, and `task` tables.

## Changes
- Changed the FHIR routing process to omit steps to create a message for and send a message to the translation queue. This event is tracked as `EventAction.NONE` rather than `EventAction.TRANSLATE`.
- Added `EventAction.NONE` as a possible action for `blobURL` creation. This should prevent primary key constraint violations if other `EventAction` types cause the `other` string to be written to `blobURL`.

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- Fixes #7545

### DIFF
--- a/prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
+++ b/prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
@@ -205,74 +205,116 @@ class FHIRRouter(
             // add the receivers, if any, to the fhir bundle
             if (listOfReceivers.isNotEmpty()) {
                 FHIRBundleHelpers.addReceivers(bundle, listOfReceivers)
-            }
 
-            // create item lineage
-            report.itemLineages = listOf(
-                ItemLineage(
-                    null,
-                    message.reportId,
-                    1,
-                    report.id,
-                    1,
-                    null,
-                    null,
-                    null,
-                    report.getItemHashForRow(1)
+                // create item lineage
+                report.itemLineages = listOf(
+                    ItemLineage(
+                        null,
+                        message.reportId,
+                        1,
+                        report.id,
+                        1,
+                        null,
+                        null,
+                        null,
+                        report.getItemHashForRow(1)
+                    )
                 )
-            )
 
-            // create translate event
-            val translateEvent = ProcessEvent(
-                Event.EventAction.TRANSLATE,
-                message.reportId,
-                Options.None,
-                emptyMap(),
-                emptyList()
-            )
+                // create translate event
+                val translateEvent = ProcessEvent(
+                    Event.EventAction.TRANSLATE,
+                    message.reportId,
+                    Options.None,
+                    emptyMap(),
+                    emptyList()
+                )
 
-            // upload new copy to blobstore
-            val bodyBytes = FhirTranscoder.encode(bundle).toByteArray()
-            val blobInfo = BlobAccess.uploadBody(
-                Report.Format.FHIR,
-                bodyBytes,
-                report.name,
-                message.blobSubFolderName,
-                translateEvent.eventAction
-            )
+                // upload new copy to blobstore
+                val bodyBytes = FhirTranscoder.encode(bundle).toByteArray()
+                val blobInfo = BlobAccess.uploadBody(
+                    Report.Format.FHIR,
+                    bodyBytes,
+                    report.name,
+                    message.blobSubFolderName,
+                    translateEvent.eventAction
+                )
 
-            // ensure tracking is set
-            actionHistory.trackCreatedReport(translateEvent, report, blobInfo)
+                // ensure tracking is set
+                actionHistory.trackCreatedReport(translateEvent, report, blobInfo)
 
-            // insert translate task into Task table
-            this.insertTranslateTask(
-                report,
-                report.bodyFormat.toString(),
-                blobInfo.blobUrl,
-                translateEvent
-            )
-
-            // nullify the previous task next_action
-            db.updateTask(
-                message.reportId,
-                TaskAction.none,
-                null,
-                null,
-                finishedField = Tables.TASK.ROUTED_AT,
-                null
-            )
-
-            // move to translation (send to <elrTranslationQueueName> queue). This passes the same message on, but
-            //  the destinations have been updated in the FHIR
-            this.queue.sendMessage(
-                elrTranslationQueueName,
-                RawSubmission(
-                    report.id,
+                // insert translate task into Task table
+                this.insertTranslateTask(
+                    report,
+                    report.bodyFormat.toString(),
                     blobInfo.blobUrl,
-                    BlobAccess.digestToString(blobInfo.digest),
-                    message.blobSubFolderName
-                ).serialize()
-            )
+                    translateEvent
+                )
+
+                // nullify the previous task next_action
+                db.updateTask(
+                    message.reportId,
+                    TaskAction.none,
+                    null,
+                    null,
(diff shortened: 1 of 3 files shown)


prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
```diff
@@ -222,15 +220,29 @@ class FHIRRouter(
                 )
             )
 
-            // create translate event
-            val translateEvent = ProcessEvent(
-                Event.EventAction.TRANSLATE,
+            // create none event
+            var translateEvent = ProcessEvent(
```
carlosfelix2, line 224: So, to better read this code, how about your create and if/else. One branch will use translate event. The other will use something like a noneEvent. Right now the reuse of the translateEvent variable confuses the flow a bit.  This will also clear up the multiple ifs you have even if you have to have a few lines of the same code.

carlosfelix2 (changes requested):

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
```diff
@@ -244,13 +256,15 @@ class FHIRRouter(
             // ensure tracking is set
             actionHistory.trackCreatedReport(translateEvent, report, blobInfo)
```
jimmyfagan, line 257: (queestion) Similar to above, do we still need to track the event if it is `NONE`?

jimmyfagan (commented):
Good work! Not too familiar with some of this process, so I left a few questions, also left a suggestion. Depending on the answers to the questions, I might have additional suggestions, but this looks like it does what we want it to.

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
```diff
@@ -222,15 +220,29 @@ class FHIRRouter(
...
+                    message.reportId,
+                    Options.None,
+                    emptyMap(),
+                    emptyList()
+                )
+            }
+
             // upload new copy to blobstore
             val bodyBytes = FhirTranscoder.encode(bundle).toByteArray()
             val blobInfo = BlobAccess.uploadBody(
```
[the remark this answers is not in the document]
jack-h-wang: The `report_file` table has a `body_url` field that corresponds to a blob. The table has a constraint `report_file_body_url_key` that requires each `body_url` in the database to be unique. An earlier draft of the code changes in this PR attempted to reuse a previous blob, but in testing this caused a database error due to violating that constraint.

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
jack-h-wang: `trackCreatedReport` is responsible for updates to the `report_file` database table. Without performing this call, there is no record of what actions the routing step performed; it would appear as if the bundle was not routed.

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
jack-h-wang: Code reworked as suggested in latest commit.

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
```diff
@@ -222,15 +220,29 @@ class FHIRRouter(
...
+
+                // create translate event - used instead of none event
+                translateEvent = ProcessEvent(
+                    Event.EventAction.TRANSLATE,
+                    message.reportId,
+                    Options.None,
+                    emptyMap(),
+                    emptyList()
+                )
+            }
```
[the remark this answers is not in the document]
jack-h-wang: The latest commit no longer uses the multiple `hasReceivers` checks, but I liked the `nextEvent` name and switched to it.

jimmyfagan (approved):
There's a tradeoff with the reworking, a bit of duplicate code now, but also a bit easier to follow what happens in each case, so I think it's worth it to keep it like this. Optionally could consider pulling common elements into [local functions](https://kotlinlang.org/docs/functions.html#local-functions), or pull out functions roughly for `finishRoutingProcessWithReceivers` and `finishRoutingProcessWithoutReceivers`, but not a requirement, and might not be worth it for the relatively small amount of duplicated code.

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
```diff
@@ -190,57 +182,100 @@ class FHIRRouter(
...
+                this.queue.sendMessage(
+                    elrTranslationQueueName,
+                    RawSubmission(
+                        report.id,
+                        blobInfo.blobUrl,
+                        BlobAccess.digestToString(blobInfo.digest),
+                        message.blobSubFolderName
+                    ).serialize()
+                )
+            } else {
```
carlosfelix2, line 245: Excellent! THank you!

carlosfelix2 (changes requested):
Code works as expected. Tested locally. There is however one issue. The submission history does not say that the process was complete. You can see in this example that it says the status is received. You need to figure out what is the proper status for this and update it too. I don't think this should be too complicated.
```
{
    "id": "260a223d-3bd6-412f-9716-63c616a8c306",
    "submissionId": 949,
    "overallStatus": "Received",
    "timestamp": "2023-02-17T15:17:39.573Z",
    "plannedCompletionAt": null,
    "actualCompletionAt": null,
    "sender": "ignore.ignore-full-elr",
    "reportItemCount": 1,
    "errorCount": 0,
    "warningCount": 0,
    "httpStatus": 201,
    "destinations": [],
    "actionName": "receive",
    "externalName": null,
    "reportId": "260a223d-3bd6-412f-9716-63c616a8c306",
    "topic": "full-elr",
    "errors": [],
    "warnings": [],
    "destinationCount": 0
}
```

carlosfelix2 (approved):

jack-h-wang:
@carlosfelix2 

Created #8451 for the Submission History behavior.

PR #8295, opened 2023-02-13
Experience/8157/redesign message tracker table

### LINKED ISSUE #8157: Redesign Message Tracker Details Receiver Table
## Problem statement

We want to implement some changes to how the data is conveyed on the Message Tracker to improve the visual hierarchy and overall readability based on internal usability findings.

## What you need to know

- The new design includes two primary sections -- one about the Sender and one list the Receivers -- with more emphasis on actionable data. See the [Figma](https://www.figma.com/proto/6mwI5ac6rprACKDzDo4Ady/ReportStream-Workspace-%7C-2023?page-id=283%3A1792&node-id=393%3A4529&starting-point-node-id=393%3A4746&scaling=min-zoom) for reference.

- Style Guide specs in [Figma](https://www.figma.com/file/6mwI5ac6rprACKDzDo4Ady/ReportStream-Workspace-%7C-2023?node-id=708%3A8503&t=FKL6GYx93mG1f06N-0)

## Acceptance criteria

- [ ] Receivers section
  - [ ] Each Receiver's data is organized into a [table row](https://designsystem.digital.gov/components/table/) with the following columns:
    - [ ] Name (`receivingOrg`)
    - [ ] Service (`receivingOrgSvc`)
    - [ ] Date (`createdAt`)
    - [ ] Report ID (`reportId`)
    - [ ] File Location 
      - [ ] Parent (parsed from `fileUrl`)
      - [ ] Child (parsed from `fileUrl`)
      - [ ] File Name (parsed from `fileUrl`)
    - [ ] Transport Results (`transportResult` or `N/A` if null)
  - [ ] Each row is sortable by Name, Service, Date, Parent, Child, File Name, and Transport Results
  - [ ] Each table cell truncates longer content
  - [ ] Each table cell can be clicked to expand to reveal full content

etanb:
This PR redesigns the table component for Receivers within Message Details.

**To access:**

- Log into RS
- Add `'message-tracker'` feature flag
- Navigate to the message tracker feature from the Admin menu
- Enter a message Id in the search (Can be found via the `covid_result_metadata` table)
- Select a message Id
- View design changes

**Changes Preview:** **UPDATED AS OF 2/21/23**

Screenshot 2023-02-21 at 5 11 49 PM

_Updated table with new sub header File Location, updated USWDS colors, font sizes, and sortability arrows_

Screenshot 2023-02-21 at 5 11 57 PM

_There are 3 sort options,`None`, `Ascending` and `Descending`. Click on a column header will enumerate through these options while highlighting the active column._

Screenshot 2023-02-21 at 5 12 09 PM

_Upon click of table cells with longer date (Report Id, File Name and Transport Results), we pop out the data into a modal for easy viewing and copying._

Closes #8157


frontend-react/src/components/MessageTracker/MessageDetails.test.tsx
```diff
@@ -191,7 +191,7 @@ describe("RSMessageDetail component", () => {
         ).toBeVisible();
         expect(screen.getAllByText("Date/Time Submitted")[0]).toBeVisible();
         expect(screen.getByText("09/28/2022, 10:21:33 PM")).toBeVisible();
-        expect(screen.getByText("File Location")).toBeVisible();
+        expect(screen.getAllByText("File Location")[0]).toBeVisible();
```
etanb, line 194: There are multiple elements on the page with `File Location` in the name now so we must get the array of them and index.

frontend-react/src/components/MessageTracker/MessageDetails.tsx
```diff
@@ -179,10 +179,17 @@ export function MessageDetails() {
...
+                    <MessageReceivers
+                        receiverDetails={messageDetails.receiverData}
+                    />
+                ) : (
+                    <>
+                        <h2>Receivers:</h2>
+                        <h3>No Data to Display</h3>
+                    </>
+                )}
             </div>
```
etanb, line 193: I pulled the logic out from the `MessageRecievers` component to determine if there was data since its makes the component much more stateless and easy to handle.

frontend-react/src/components/MessageTracker/MessageReceivers.test.tsx
```diff
@@ -13,79 +14,65 @@ describe("MessageReceivers component", () => {
...
         expect(
-            screen.getByText(
-                "For ak-phd.elr, filter isValidCLIA[testing_lab_clia, reporting_facility_clia] filtered out item Alaska1"
-            )
-        ).toBeInTheDocument();
+            screen
+                .getByRole("table")
+                .getElementsByClassName("message-receiver-break-word")[0]
+        ).toHaveTextContent("400eae4e-b24d-45aa-bc5c-4d96a0bfef96");
     });
```
etanb, line 77: This test is showing that upon click of `Report Id`, the order of the data has changed from:

1st: `578eae4e-b24d-45aa-bc5c-4d96a0bfef96`
2nd: `400eae4e-b24d-45aa-bc5c-4d96a0bfef96`

TO:

1st: `400eae4e-b24d-45aa-bc5c-4d96a0bfef96`
2nd: `578eae4e-b24d-45aa-bc5c-4d96a0bfef96`

frontend-react/src/styles/_uswds_settings.scss
```diff
@@ -64,6 +64,9 @@ $flex-wrap-settings: (
   responsive: true
 );
 
+// Add all USWDS colors: https://designsystem.digital.gov/design-tokens/color/system-tokens/
+$global-color-palettes: "palette-color";
+
```
etanb, line 69: This adds all the needed colors from `https://designsystem.digital.gov/design-tokens/color/system-tokens/` that design uses across their Figma.

etanb:
FYI: Have reviewed this PR with Design/Jillian and has received their approval 👍

jillian-hammer:
thanks for setting up the preview screenshots @etanb, and for replacing the arrow icons. these look great! 😍 

a few of notes:

- icons look too small overall, I think they should be bumped up for better legibility (see design comp reference below)
- icons have inconsistent sizes (particularly the Transport Results icon)
- when `Ascending` and `Descending` the icon jumps up in size, but it should be remain the same size as in the `None` state 
- the space between the column header and arrow icon should be consistent between all of the columns (for example, it's closer to "Report ID" than it is to "Sub"). in the design, they are aligned to the right of the text by 8px.

i'm sure these are tricky bc of the columns being narrow – happy to talk through it. thank you!

**Design reference below 👇** 
A) Icon size & alignment reference
image

B) Icon size remaining the same when `Ascending`
image

frontend-react/src/styles/_components.scss
```diff
@@ -9,6 +11,43 @@ hr {
     padding: 0.25rem 0.25rem;
 }
 
+// MessageDetails.tsx
+.usa-table thead th {
```
stephenkao, line 15: Hmmm can you wrap all of these rules in a Message Tracker-specific modifier class?  We don't want these rules to leak out and apply globally to other tables -- at least not until we're using one unified table component.

I realize there's some `message-receivers`-specific classes below, but I think it'd be safer to wrap everything and only unwrap when we're sure that these are styles we want to apply globally.

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
```diff
@@ -1,62 +1,479 @@
...
+    Date: "Date",
+    ReportId: "Report Id",
+    Main: "Main",
+    Sub: "Sub",
+    FileName: "File Name",
+    TransportResults: "Transport Results",
+};
+
+const FilterOptionsEnum = {
+    None: "",
```
stephenkao, line 67: Should this be consistently uppercased?  `NONE`?

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
```diff
@@ -1,62 +1,479 @@
...
+                    modalRef?.current?.toggleModal();
+                }}
+            >
+                {receiver.FileName}
+            </td>
+            <td
+                className={`message-receiver-break-word ${checkForActiveColumn(
+                    ColumnDataEnum.TransportResults
+                )}`}
+                onClick={() => {
```
stephenkao, line 199: Minor point, but `td` isn't an interactive element for accessibility.  Any chance this could be turned into an unstyled `<button>` instead?

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
```diff
@@ -1,62 +1,479 @@
...
+                                        receiverItem.transportResult;
+                                break;
+                        }
+                    }
+                    return formattedData;
+                }
+            ),
+        [receiverDetails]
+    );
+    const sortedData = useMemo(
```
stephenkao, line 324: I think this could also be generalized with the [data format](https://github.com/CDCgov/prime-reportstream/pull/8295/files#r1108850755), no?  Like if someone clicks on the 0th header, we'd just sort the array of tuples.  The generalized sort function could include type checking and/or casting as needed 🤔

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
```diff
@@ -1,62 +1,479 @@
...
+    Service: "Service",
+    Date: "Date",
+    ReportId: "Report Id",
+    Main: "Main",
+    Sub: "Sub",
+    FileName: "File Name",
+    TransportResults: "Transport Results",
+};
+
+const FilterOptionsEnum = {
```
stephenkao, line 66: Out of curiosity, why are these consts rather than real enums?

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
```diff
@@ -1,62 +1,479 @@
...
+                setActiveColumnSortOrder(FilterOptionsEnum.DESC);
+                break;
+            case activeColumnSortOrder === FilterOptionsEnum.DESC:
+                setActiveColumnSortOrder(FilterOptionsEnum.None);
+                break;
+        }
+    };
+    return (
+        <th
+            className={
```
stephenkao, line 110: Can use `classnames` for this:

```tsx
className={ classnames("active-col-header: isCurrentlyActiveColumn && activeColumnSortOrder !== FilterOptionsEnum.None) }
```

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
```diff
@@ -1,62 +1,479 @@
...
+    const isCurrentlyActiveColumn = columnHeaderTitle === activeColumn;
+    const handleColHeaderClick = () => {
+        if (!isCurrentlyActiveColumn) {
+            // Reset active column and sort order on new column click
+            setActiveColumn(columnHeaderTitle);
+            setActiveColumnSortOrder(FilterOptionsEnum.None);
+        }
+
+        // Explicitly set the proceeding sort order
+        switch (true) {
```
stephenkao, line 96: Minor point, but I feel like an `if-else` would be clearer here 🤔

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
```diff
@@ -1,62 +1,479 @@
...
+                {receiver.TransportResults}
+            </td>
+        </tr>
+    );
 };
 
 export const MessageReceivers = ({ receiverDetails }: MessageReceiverProps) => {
+    const modalRef = useRef<ModalRef>(null);
+    const [modalText, setModalText] = useState({ title: "", body: "" });
+    const [activeColumn, setActiveColumn] = useState("");
```
stephenkao, line 216: Do we need to save active column and active column sort order separately or could we just do something like storing them both in one state object?

```tsx
const [activeSortColumn, setActiveSortColumn] = useState(null);
// setting one column
setActiveSortColumn({ column: "Name", order: "ASC" });
// resetting
setActiveSortColumn(null);
```

I think it's easier to reason about if we treat all sort operations as an atomic change (both sorted column and sort order) 🤷

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
```diff
@@ -1,62 +1,479 @@
...
+    const [activeColumnSortOrder, setActiveColumnSortOrder] = useState("");
+    const filterIcon = useMemo(
+        () =>
+            activeColumnSortOrder === FilterOptionsEnum.ASC ? (
+                <Icon.ArrowUpward />
+            ) : activeColumnSortOrder === FilterOptionsEnum.DESC ? (
+                <Icon.ArrowDownward />
+            ) : (
+                <Icon.SortArrow />
```
stephenkao, line 225: Also, minor point, but would you mind unnesting this ternary?  I think an `if-else if-else` or even a `case` could be clearer. 

We could enable [an ESLint rule](https://eslint.org/docs/latest/rules/no-nested-ternary) for preventing nested ternaries too 🤷

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
```diff
@@ -1,62 +1,479 @@
...
+        </tr>
+    );
 };
 
 export const MessageReceivers = ({ receiverDetails }: MessageReceiverProps) => {
+    const modalRef = useRef<ModalRef>(null);
+    const [modalText, setModalText] = useState({ title: "", body: "" });
+    const [activeColumn, setActiveColumn] = useState("");
+    const [activeColumnSortOrder, setActiveColumnSortOrder] = useState("");
+    const filterIcon = useMemo(
```
stephenkao, line 218: Why do we need to pass `filterIcon` down into the headers?  Don't the headers already have enough information to infer that themselves with `activeColumn` and `activeColumnSortOrder`?

frontend-react/src/styles/_components.scss
```diff
@@ -1,4 +1,6 @@
 $theme-hr-border-color: "base-lighter";
+$bg-col-header: #73B3E7;
```
stephenkao, line 2: Can we use `color("primary-light")` and `color("primary-lighter")` instead of introducing new hardcoded hex codes?

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
```diff
@@ -1,62 +1,479 @@
...
+                        title: `${ColumnDataEnum.ReportId}:`,
+                        body: receiver.ReportId,
+                    });
+                    modalRef?.current?.toggleModal();
+                }}
+            >
+                {receiver.ReportId}
+            </td>
+            <td className={checkForActiveColumn(ColumnDataEnum.Main)}>
+                <p
```
stephenkao, line 168: Might be nice to pull out any custom/non-verbatim table cell logic into separate components.  This could be specified on the top level when the Table component is rendered like:

`[null, null, null, StatusBadge, null, ...]`

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
```diff
@@ -1,62 +1,479 @@
...
+                    }`}
+                >
+                    {uppercaseMain}
+                </p>
+            </td>
+            <td className={checkForActiveColumn(ColumnDataEnum.Sub)}>
+                {receiver.Sub}
+            </td>
+            <td
+                className={`message-receiver-break-word white-space-pre ${checkForActiveColumn(
```
stephenkao, line 182: Is `white-space-pre` a real class?  🤔

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
```diff
@@ -1,62 +1,479 @@
...
+                </thead>
+                <tbody className="message-receivers-table">
+                    {sortedData.map(
+                        (receiver: NormalizedReceiverData, index) => (
+                            <MessageReceiversRow
+                                key={index}
+                                receiver={receiver}
+                                activeColumn={activeColumn}
+                                activeColumnSortOrder={activeColumnSortOrder}
+                                setModalText={setModalText}
```
stephenkao, line 464: Rather than pass `setModalText` and `modalRef` down directly, can we just wrap that in a function?

```jsx
function openModal({ title, body }) {
    setModalText({ title, body });
    modalRef?.current?.toggleModal();
}
```

Then we could generalize this to an optional `onCellClick` for the individual cells wherever needed 🤷

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
```diff
@@ -1,62 +1,479 @@
...
+                {receiver.Service}
+            </td>
+            <td className={checkForActiveColumn(ColumnDataEnum.Date)}>
+                {receiver.Date}
+            </td>
+            <td
+                className={`message-receiver-break-word ${checkForActiveColumn(
+                    ColumnDataEnum.ReportId
+                )}`}
+                onClick={() => {
```
penny-lischer, line 157: Shouldn't we be truncating the field and then only showing modal if needed?

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
etanb: I was thinking about that, but I've really only seen that pattern when updating a large state object with `useReducer`. From what I've seen in the app and in [React's docs](https://reactjs.org/docs/hooks-state.html#tip-using-multiple-state-variables), `useState` is super explicit in updating a very _specific_ bit of state even if it feels excessive.

frontend-react/src/styles/_components.scss
etanb: Yeah! Good call, that should be standard practice until we use CSS modules.

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
etanb: Totally, in my mind I was thinking `none` is literally, no string / an inherently falsy value.

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
etanb: I think that would affect the table layout, I was thinking I could just add `role="button"`?

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
etanb: I was getting some strange TS errors about using them as a `key` in places, and from what I see in TS, the TS `enum` just compiles down to an object anyways. I can swap these back to `enum` and debug.

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
etanb: Sounds good, can refactor my dynamic class implementation across the board.

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
etanb: it's true, it's only 3 options, just matching my other sort function. can change to `if-else`!

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
etanb: Done with `filterIcon` refactor!

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
etanb: Yes! Good catch. `filterIcon` was created before I refactored things to the way they are.

frontend-react/src/styles/_components.scss
etanb: We actually have access to `$blue-10` and `$blue-30` now! Will change.

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
etanb: Changed to USWDS' class.

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
etanb: Talked to Jillian and figured a path forwards here!

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
stephenkao: No worries -- these could also be problematic if they're needed at compile-time versus runtime.  If that's the case, then keeping at consts is fine, but the `Enum` suffix is just a bit unnecessary :)

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
```diff
@@ -1,62 +1,479 @@
...
+    //     Name: string;
+    //     Service: string;
+    //     Date: string;
+    //     ReportId: string;
+    //     Main: string;
+    //     Sub: string;
+    //     FileName: string;
+    //     TransportResults: string;
+    // }
+    const normalizedData = useMemo(
```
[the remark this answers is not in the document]
etanb: Continuing this discussion in the backend updates in this ticket: https://app.zenhub.com/workspaces/experience-607d9d5e68b95200150fec37/issues/gh/cdcgov/prime-reportstream/8281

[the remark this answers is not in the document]
etanb: We'd ultimately update all of this when we update our generalized approach to tables: https://app.zenhub.com/workspaces/experience-607d9d5e68b95200150fec37/issues/gh/cdcgov/prime-reportstream/8280

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
```diff
@@ -1,62 +1,511 @@
...
+        // Explicitly set the proceeding sort order
+        if (activeColumnSortOrder === FilterOptionsEnum.NONE) {
+            setActiveColumnSortOrder(FilterOptionsEnum.ASC);
+        } else if (activeColumnSortOrder === FilterOptionsEnum.ASC) {
+            setActiveColumnSortOrder(FilterOptionsEnum.DESC);
+        } else if (activeColumnSortOrder === FilterOptionsEnum.DESC) {
+            setActiveColumnSortOrder(FilterOptionsEnum.NONE);
+        }
+    };
+    const sortIcon = useMemo(() => {
```
stephenkao, line 103: I thought I left a comment about this earlier, but I guess not!  😅 

Instead of memoizing, we could just infer the icon type directly instead of rendering early:
```tsx
let SortIcon = Icon.SortArrow;
if (isCurrentlyActiveColumn && activeColumnSortOrder === FilterOptionsEnum.ASC) {
    SortIcon = Icon.ArrowUpward;
} else if (isCurrentlyActiveColumn && activeColumnSortOrder === FilterOptionsEnum.DESC) {
    SortIcon = Icon.ArrowDownward;
}
```

Then just render `<SortIcon>` in the JSX below 🤷

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
```diff
@@ -1,62 +1,511 @@
...
+const MessageReceiversRow = ({
+    receiver,
+    activeColumn,
+    activeColumnSortOrder,
+    setModalText,
+    modalRef,
+}: MessageReceiversRowProps) => {
+    // We highlight the entire active column so within the row,
+    // we need to know which column is active
+    const checkForActiveColumn = (colName: string) =>
```
stephenkao, line 152: Small point, but the name of this function is a little misleading.  Could it be named something like `getColumnClasses`?

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
stephenkao: Might be nice to break out this onClick into a function:

```jsx
function handleCellClick(title, body)
    setModalText({ title, body });
    modalRef?.current.toggleModal();
}
```

Then just use that instead of redefining it each time below:

```jsx
<td
    onClick={ () => handleCellClick(ColumnDataEnum.ReportId, receiver.reportId) }
>
```

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
```diff
@@ -1,62 +1,511 @@
...
+                    )}
+                >
+                    {uppercaseMain}
+                </p>
+            </td>
+            <td className={checkForActiveColumn(ColumnDataEnum.Sub)}>
+                {receiver.Sub}
+            </td>
+            <td
+                role="button"
```
stephenkao, line 204: Oh!  One issue with going with `role="button"` is that we'll have to update `tabIndex="0"` for keyboard navigation.  This is handled by default by interactive elements like buttons and links, so we should use those first if possible.

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
stephenkao: Right sorry, I meant that the cell content could be wrapped in a button since it looks like it's all just phrasing content 🤔

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
stephenkao: Etan and I chatted about this separately, and this isn't a huge deal either way.  My rationale was coming from state updates getting swallowed up [in React's batching](https://github.com/facebook/react/issues/14259) (like lots of `setState` calls sequentially), but it should be fine here since it's happening in an event-handler.

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
etanb: Added to our discussion on what features to add for a genericized table component: https://app.zenhub.com/workspaces/experience-607d9d5e68b95200150fec37/issues/gh/cdcgov/prime-reportstream/8280

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
etanb: Added to our discussion on what features to add for a genericized table component: https://app.zenhub.com/workspaces/experience-607d9d5e68b95200150fec37/issues/gh/cdcgov/prime-reportstream/8280

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
etanb: Added to our discussion on what features to add for a genericized table component: https://app.zenhub.com/workspaces/experience-607d9d5e68b95200150fec37/issues/gh/cdcgov/prime-reportstream/8280

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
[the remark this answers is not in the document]
etanb: Added to our discussion on what features to add for a genericized table component: https://app.zenhub.com/workspaces/experience-607d9d5e68b95200150fec37/issues/gh/cdcgov/prime-reportstream/8280

penny-lischer (approved):
LGTM once other comments have been addressed.

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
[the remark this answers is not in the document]
stephenkao: Etan and I chatted about this, and we can hold off on the first point for the wider generic Table discussion, but I still think the data normalization and sorting can be hoisted out of the component:

```tsx
function normalizeReceiverDetails(receiverDetails) { ... }
function sortReceiverDetails(...) { ... }

Then just call those functions in the component:

```tsx
function MessageReceivers({ receiverDetails }) {
    const normalizedData = useMemo(() => normalizeReceiverDetails(receiverDetails), [...]);
    
    ...

    // also can we just sort on an event?
    <th onClick={ () => sortReceiverDetails()></th>
}

File: frontend-react/src/components/MessageTracker/MessageReceivers.tsx [hunk 24]
```diff
@@ -1,62 +1,511 @@
...
+                        (accumulator: any, currentValue: string) => {
+                            accumulator[currentValue] = "N/A";
+                            return accumulator;
+                        },
+                        {}
+                    );
+                    for (const key of Object.keys(ColumnDataEnum)) {
+                        const columnTitle =
+                            ColumnDataEnum[key as keyof typeof ColumnDataEnum];
+                        switch (true) {
```
stephenkao, line 284: I feel like `if-else`s are clearer than `switch (true)` 🤔

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
```diff
@@ -1,62 +1,511 @@
...
+                            case columnTitle === ColumnDataEnum.Name:
```
stephenkao, line 285: Can this be simplified a bit?  Essentially, we have a few different groups of data mappers/normalizers:

- No-ops/1:1s: Name, Service, ReportId, TransportResults
- Date formatter: Date
- File url formatter: Main, Sub, FileName

We're always going to need to have a data mapper to transform response data for what we want to display, but I think we can cut down on some repeated code.

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
```diff
@@ -1,62 +1,511 @@
...
+                                        );
+
+                                break;
+                            case columnTitle === ColumnDataEnum.ReportId:
+                                if (receiverItem.reportId)
+                                    formattedData.ReportId =
+                                        receiverItem.reportId;
+                                break;
+                            case columnTitle === ColumnDataEnum.Main:
+                                const { folderLocation } = parseFileLocation(
```
stephenkao, line 309: Similar to [the point above](https://github.com/CDCgov/prime-reportstream/pull/8295/files#r1114986589), we don't need to call `parseFileLocation` three times; we can call it once and destructure from that, no?

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
```diff
@@ -1,62 +1,511 @@
...
+                        },
+                        {}
+                    );
+                    for (const key of Object.keys(ColumnDataEnum)) {
+                        const columnTitle =
+                            ColumnDataEnum[key as keyof typeof ColumnDataEnum];
+                        switch (true) {
+                            case columnTitle === ColumnDataEnum.Name:
+                                if (receiverItem.receivingOrg)
+                                    formattedData.Name =
```
stephenkao, line 287: How come the property name in `formattedData` is title-cased?

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
```diff
@@ -1,62 +1,511 @@
...
+                                columnHeaderTitle={ColumnDataEnum.ReportId}
+                                activeColumn={activeColumn}
+                                setActiveColumn={setActiveColumn}
+                                activeColumnSortOrder={activeColumnSortOrder}
+                                setActiveColumnSortOrder={
+                                    setActiveColumnSortOrder
+                                }
+                                rowSpan={2}
+                            />
+                            <th colSpan={3}>
```
stephenkao, line 436: Definitely not for now, but along the lines of what we were talking about for coming up with a standard structure to pass to the table, I wonder if we could do something like recursively nest them so we could programmatically iterate over them when rendering:

```tsx
const headers = [{
    title: "Name"
}, {
    title: "File Location",
    headers: [{
        title: "Transport Results"
    }, {
        title: "Main"
    }, {
        title: "Sub"
    }]
}];
```

idk something like that 🤷

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
```diff
@@ -1,62 +1,511 @@
...
-                    <DetailItem
-                        item={"Date/Time Submitted"}
-                        content={formattedDateFromTimestamp(
-                            receiver.createdAt,
-                            "MMMM DD YYYY"
+                    <thead>
+                        <tr>
+                            <MessageReceiversCol
+                                columnHeaderTitle={ColumnDataEnum.Name}
+                                activeColumn={activeColumn}
```
stephenkao, line 398: For common props like this, we could just store them in a spread variable:

```tsx
const commonProps = {
    activeColumn,
    setActiveColumn,
    activeColumnSortOrder,
    setActiveColumnSortOrder
};

<MessageReceiversCol
    { ...commonProps }
    columnHeaderTitle={ColumnDataEnum.Name}
    rowSpan={2}
/>
<MessageReceiversCol
    { ...commonProps }
    columnHeaderTitle={ColumnDataEnum.Service}
    rowSpan={2}
/>
...
```

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
```diff
@@ -1,62 +1,511 @@
...
+    const isCurrentlyActiveColumn = columnHeaderTitle === activeColumn;
+    const handleColHeaderClick = () => {
+        if (!isCurrentlyActiveColumn) {
+            // Reset active column and sort order on new column click
+            setActiveColumn(columnHeaderTitle);
+            setActiveColumnSortOrder(FilterOptionsEnum.NONE);
+        }
+
+        // Explicitly set the proceeding sort order
+        if (activeColumnSortOrder === FilterOptionsEnum.NONE) {
```
stephenkao, line 95: This is a MINOR thing, but this state machine works for setting the sort order in the same column, but not necessarily when you click on another column:

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
```diff
@@ -1,62 +1,511 @@
...
+            >
+                {receiver.TransportResults}
+            </td>
+        </tr>
+    );
 };
 
 export const MessageReceivers = ({ receiverDetails }: MessageReceiverProps) => {
+    const modalRef = useRef<ModalRef>(null);
+    const [modalText, setModalText] = useState({ title: "", body: "" });
```
stephenkao, line 239: Oooooh already using state for an object 😆 

(Nah, I get it -- title and body should both change atomically :))

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
etanb: Yup, I like that.

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
etanb: You got me! My thinking was when it's changing a piece of how something works vs. strings. Also, as you said `title` and `body` MUST be changed at the same time, while column name and sort direction can change independent of each other. 🤷

frontend-react/src/styles/_components.scss
```diff
@@ -11,6 +11,55 @@ hr {
     padding: 0.25rem 0.25rem;
 }
 
+// MessageReceivers.tsx
+.message-receivers-table {
+  .usa-table thead th {
+    cursor: pointer;
+    user-select: none;
+    background-color: #f0f0f0;
```
stephenkao, line 19: Can this use `color('base-lightest')` instead?

frontend-react/src/styles/_components.scss
```diff
@@ -11,6 +11,55 @@ hr {
     padding: 0.25rem 0.25rem;
 }
 
+// MessageReceivers.tsx
+.message-receivers-table {
+  .usa-table thead th {
```
stephenkao, line 16: TINIEST THING, but would you mind updating the indentations to be four spaces?

frontend-react/src/styles/_uswds_settings.scss
```diff
@@ -59,4 +59,6 @@
   $flex-wrap-settings: (
     responsive: true
   ),
-);
+
+  $global-color-palettes: "palette-color",
```
stephenkao, line 63: Mind adding a comment here to say that it exposes USWDS tokens for us?

frontend-react/yarn.lock
```diff
@@ -3217,7 +3217,7 @@
     "@webassemblyjs/ast" "1.11.1"
     "@xtuc/long" "4.2.2"
 
-"@xmldom/xmldom@^0.8.3":
+"@xmldom/xmldom@0.8.6", "@xmldom/xmldom@^0.8.3":
```
stephenkao, line 3220: This can be removed from this changeset.  It's not directly related to your changes, and it should be rectified on `master` already 🙏

frontend-react/src/components/MessageTracker/MessageReceivers.test.tsx
stephenkao: I added a comment about this [here](https://github.com/CDCgov/prime-reportstream/pull/8295/files#r1114978097), but I think we should break out the sorting logic from the component body so we can test the sorting proper versus the component proper.  If this test fails, there could be a number of reasons:
- Something failed to render
- Class names changed
- Role changed
- Fake data changed
- Sorting logic changed
- ...probably a bunch of other things

If we broke out the sorting logic (and data mapping logic for that matter!) and saw failing tests for that, we'd know exactly what's failing rather than having to debug the rendering in addition to the sorting.

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
etanb: Sounds good!

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
etanb: Yeah, sounds (semantically) good!

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
etanb: Excellent call out! Added.

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
etanb: I can actually optimize this `switch` statement to it's perfect form: 

```
switch (columnTitle) {
  case ColumnDataEnum.Name:
  case ColumnDataEnum.Name:
// etc
}
```

which I think is actually easier to read than a ton of if-else statements. In addition, from what I've read online, `switch` statements will be read faster in that the compiler will "go to" the `true` condition immediately instead of with if-else ladder, it would have to descend one-at-a-time before getting to the "truthy" value.

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
etanb: In code yes, but during runtime, my thinking was that it would actually be called MORE often since if we pulled `parseFileLocation` to the very top, EVERY time `normalizedData` is re-rendered, the `parseFileLocation` would be called instead of just for its proper 3 cases. Is that over thinking it?

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
etanb: because they're the `keys` from `ColumnDataEnum`

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
etanb: I like this! Will change.

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
etanb: Nice catch, fell into the classic react no-no of changing `useState` and reading the updated value in the same call.

frontend-react/src/styles/_components.scss
etanb: Will use color token `gray-5`

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
penny-lischer: Agree and could have other properties for each object i.e.
{
        field: 'reportId',
        label: 'Report Id',
        colSpan: ColSpanValue,
        sortable: true,
        sortField: sortFields.name
      }

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
stephenkao: Fantastic -- definitely would be best to use your proposed changed here :)

Regarding whether or not it's best to use `switch` versus `if-else`...this is a hotly debated topic 😆 Compiled languages have more definitive answers for this, but I'd say the performance benefits are negligible in JavaScript -- especially since the interpretation/underlying implementation differs between browsers -- so we should opt for human readability whenever.  

[Here](https://jsben.ch/5w6eT)'s a quick comparison if you're curious!  (Also RIP jsperf?)

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
stephenkao: Ah, nope -- you're right :)  I'm the one overthinking it here 😆

PR #8314, opened 2023-02-13
Remove react-device-detect

### LINKED ISSUE #8313: Remove react-device-detect
## Problem statement

We use the `react-device-detect` specifically to check  [if the user is using Internet Explorer](https://github.com/CDCgov/prime-reportstream/blob/master/frontend-react/src/App.tsx#L107), and given IE's usage, we don't need to support it any longer.

## What you need to know

- Right now, we don't need to do browser or device detection.  If we do, there are [other alternatives](https://developer.mozilla.org/en-US/docs/Web/HTTP/Browser_detection_using_the_user_agent) we can use that are strictly in-browser features that we don't need to use a third-party package to check.

## Acceptance criteria

- [ ] `react-device-detect` is removed from our dependencies and code

jpandersen87:
Fixes #8313

This PR removes the react-device-detect dependency and replaces the usage of its isIE function with a getIsIE function supplied by us.


frontend-react/src/utils/GetIsIE.ts
```diff
@@ -0,0 +1,4 @@
+export function getIsIE() {
```
stephenkao, line 1: Rather than creating a utility function strictly for detecting Internet Explorer, I wonder if this could be broadened to `(d|D)evice.ts` where we could eventually house functionality to detect other browsers and/or mobile devices as well.  And maybe even a `getBrowser` convenience function down the line.  Whatchu think?

Also, I think we can hoist the IE regex out since we're not using the global flag:

```tsx
const IE_REGEX = /Trident\/|MSIE/;

export function getIsIE() {
    return IE_REGEX.test(window.navigator.userAgent);
}
```

jpandersen87: I feel at that rate you've broadened back to needing a library that keeps up to date with devices.

stephenkao: That's a fair point, though I'd argue that with that file it'd be more gradual and opt-in rather than having an entire library with 99% of it unused.

In any case, I feel like it's unnecessary to have an entire file devoted to checking against a single regular expression.  Not a blocker, but it seems unnecessary 🤷 

(Also, obligatory "mind adding some tests?" comment  😆)

jpandersen87: With IE behind us we should be programmatically making choices based on available features (which should be getting polyfilled anyways) and device hardware capabilities (screen size, dpi, etc.) and not necessarily the device/browser itself.

stephenkao: Sure, maybe device.ts is a misnomer, but "feature" is a very overloaded term -- my thought here was that we'd have a utility file for any sort of the aforementioned detection going forward.  But again, not a blocker at this point.

frontend-react/src/App.tsx
```diff
@@ -104,7 +104,7 @@ const App = () => {
         debounce: 500,
     });
 
-    if (isIE) return <ErrorUnsupportedBrowser />;
+    if (getIsIE()) return <ErrorUnsupportedBrowser />;
```
stephenkao, line 107: Minor point, but the useragent shouldn't be changing for the app's lifetime so we can hoist this out of the render.

frontend-react/src/utils/GetIsIE.test.ts
```diff
@@ -0,0 +1,18 @@
...
+
+describe("GetIsIE", () => {
+    test("detects ie 10", () => {
+        expect(
+            getIsIE(
+                "Mozilla/5.0 (compatible; MSIE 10.0; Windows NT 6.1; Trident/6.0)"
+            )
+        ).toBeTruthy();
+    });
+    test("detects ie 11", () => {
```
stephenkao, line 11: Mind adding a falsy return value case or two?

stephenkao (approved):
Looks good!  Thanks for flagging this in the first place!

PR #8332, opened 2023-02-14
Fix delivery & submission history filtering inconsistencies

### LINKED ISSUE #8019: Can't query receiver sender channel on submission history API
## Describe the bug

Querying a specific receiver sender channel on the submission history endpoint returns submissions for all sender channels.

### Given:
- organization `foo` with sender channels `s1` and `s2`
- organization `bar` with receiver `r1` (no senders)

### Expectations

1. `/api/waters/org/foo/submissions` should return submissions for both s1 and s2.
2. `/api/waters/org/foo.s1/submissions` should return submissions for s1  only.  None for s2.
3. `/api/waters/org/bar/submissions` should return an empty array, not an error.  The org exists, but has no submissions.
4. `/api/waters/org/buzz/submissions` should return a 404 error - no such org 'blarg-org'
5. `/api/waters/org/foo.buzz/submissions` should return a 404 - no such sender 'blarg'
6. `/api/waters/org/bar.r1/submissions` should return a 404 - no such sender 'r1'.   (r1 is a valid receiver not a valid sender)

## Acceptance criteria

- [x] The submissions endpoint does the above.
- [x] There are nice unit tests for all the above variations
- [x] There should be nothing special about the sender called 'default'
- [x] The Experience team has been notified of the change and the change is coordinated with them.
- [x] The openapi specs are documented
- [x] This is a customer facing change, so it must go in the release-notes.md.
- [X] ~~Determine if the deliveries API has similar issues.  If so, ticket that work separately.~~ It does. Ticket.

## Impact on ReportStream
- Current implementation drives the Experience team bonkers.

### LINKED ISSUE #8322: Can't query receiver deliveries for all sender channels
## Describe the bug

Querying deliveries for all receivers of a sender channel does not work.

### Given:
- organization `foo` with receiver `r1` and `r2`
  - `r1` with deliveries
  - `r2` with no deliveries
- organization `bar` with no deliveries

### Expectations

1. `/api/waters/org/foo/deliveries` should return submissions for both r1 and r2.
2. `/api/waters/org/foo.r1/deliveries` should return submissions for r1 only.  None for r2.
3. `/api/waters/org/bar/deliveries` should return an empty array, not an error.  The org exists, but has no deliveries.
4. `/api/waters/org/buzz/deliveries` should return a 404 error - no such org 'buzz'
5. `/api/waters/org/foo.blarg/submissions` should return a 404 - no such sender 'blarg'

## Acceptance criteria

- [x] The deliveries endpoint does the above.
- [ ] There are nice unit tests for all the above variations
- [x] There should be nothing special about the sender called 'default'
- [x] The Experience team has been notified of the change and the change is coordinated with them.
- [x] The openapi specs are documented
- [x] This is a customer facing change, so it must go in the release-notes.md.

thetaurean:
This PR fixes inconsistent filtering behavior in the delivery and submission history endpoints.

Test Steps:
1. Ensure smoke test data is present in the database
2. Run the backend
3. `GET` the following URLs in your tool of choice and verify the result

| curl | Result |
| ---- | ------ |
| `/api/waters/org/ignore/submissions` | submission history for all senders |
| `/api/waters/org/ignore.ignore-full-elr/submissions` | 🆕 submission history for `ignore-full-elr` senders only |
| `/api/waters/org/ignore.sdfljkd/submissions` | 404 |
| `/api/waters/org/ignore/deliveries` | 🆕 delivery history for all receivers |
| `/api/waters/org/ignore.CSV/deliveries` | delivery history for `CSV` receivers |
| `/api/waters/org/ignore.sdfljkd/deliveries` | 404 |

### Old Behavior:
- Submission histories for all sending services was being returned no matter what.
  - e.g. `GET waters/someorg.somesender/submissions` would return all senders for someorg 
- Delivery histories for all receiving services could not be queried (only a specific service)
  - e.g. `GET waters/someorg/deliveries` resulted in 404 for a valid org

### New Behavior (also Release Notes):
- Submissions history can now be retrieved for a specific sender
  - e.g. `GET waters/someorg.somesender/submissions` returns only submissions from `somesender`
- Delivery histories for all receiving services can now be queried
  - e.g. `GET waters/someorg/deliveries` returns delivery history for all receivers of `someorg`
 
### Testing
- [X] Tested locally?
- [x] Added tests for submissions
- [ ] Added tests for deliveries

### Process
- [x] ~~Updated~~ Drafted the release notes?

## Linked Issues
- Fixes #8019 
- Fixes #8322 

## TODO

- New: #8424


prime-router/src/main/kotlin/history/azure/DeliveryFunction.kt
```diff
@@ -45,9 +47,13 @@ class DeliveryFunction(
...
-        val receiver = workflowEngine.settings.findReceiver(organization)
-        receivingOrgSvc = receiver?.name
-        return receiver?.organizationName
+        return if (organization.contains(Sender.fullNameSeparator)) {
+            workflowEngine.settings.findReceiver(organization).also { receivingOrgSvc = it?.name }
+        } else {
+            workflowEngine.settings.receivers.find {
+                it.organizationName == organization && it.customerStatus == CustomerStatus.ACTIVE
+            }
+        }?.organizationName
```
thetaurean, line 56: If a fully qualified `org.Receiver` string was supplied, find that receiver's org name & store its name for further filtering. Otherwise, find an active receiver for that org to validate the org name, but don't store the receiver name.

**We may want to change the filtering here (or get last instance instead of first)**

prime-router/src/main/kotlin/history/azure/SubmissionFunction.kt
```diff
@@ -26,14 +27,21 @@ class SubmissionFunction(
...
      * Get the correct name for an organization sender based on the name.
      *
      * @param organization Name of organization and service
      * @return Name for the organization
      */
     override fun getOrgName(organization: String): String? {
-        return workflowEngine.settings.findSender(organization)?.organizationName
+        return workflowEngine.settings.findSender(organization).also {
+            if (organization.contains(Sender.fullNameSeparator)) sendingOrgSvc = it?.name
+        }?.organizationName
```
thetaurean, line 44: If a fully qualified org.sender string was supplied, store the validated sender name to use during our query later on.

prime-router/docs/api/waters-delivery.yml
```diff
@@ -16,6 +16,39 @@ servers:
...
+            application/json:
+              schema:
+                $ref: '#/components/schemas/GetDeliveriesResponse'
+        '400':
+          description: bad request
+        '401':
+          description: unauthorized
+        '404':
+          description: organization does not exist
   /waters/org/{organization}.{client}/deliveries:
```
thetaurean, line 52: This is the actual change. The diff is wonky.

Is there a better way to reference an existing endpoint with an additional optional parameter?
From my research this seemed to be the correct way. Swagger does not permit optional path params.

Same goes for `waters-submissions.yml`

prime-router/src/main/kotlin/history/azure/DeliveryFunction.kt
```diff
@@ -45,9 +47,13 @@ class DeliveryFunction(
      * @return Name for the organization
      */
     override fun getOrgName(organization: String): String? {
-        val receiver = workflowEngine.settings.findReceiver(organization)
-        receivingOrgSvc = receiver?.name
-        return receiver?.organizationName
+        return if (organization.contains(Sender.fullNameSeparator)) {
```
stephenkao, line 50: They're literally the same character so no logic would change, but shouldn't this be `Receiver.fullNameSeparator` given the context?

prime-router/src/main/kotlin/history/azure/DeliveryFunction.kt
```diff
@@ -45,9 +47,13 @@ class DeliveryFunction(
...
+            workflowEngine.settings.findReceiver(organization).also { receivingOrgSvc = it?.name }
+        } else {
+            workflowEngine.settings.receivers.find {
```
stephenkao, line 53: Out of curiosity, why are we checking for active status here and not [in SubmissionFunction](https://github.com/CDCgov/prime-reportstream/pull/8332/files#diff-c78040f89c39c1a4b5c02d72cfb1f1430853317a722a5f299f4eecb56dc0b405R41)?  Senders and Receivers both should be able to be disabled, no?

stephenkao: The reason I'm asking is -- and this is definitely outside the scope of this pull request -- if the logic could be the same for both types of services, it feels like there's an opportunity for some consolidation here, just bifurcated by service type.  Basically:

- Find the service (Sender|Receiver) by Organization name
- Check if there's a period separator
    - If there is, store the latter half as an instance variable
    - If there isn't find the name of the service by type + Organization name

As far as I can tell, it looks like there's just [some fallback logic](https://github.com/CDCgov/prime-reportstream/blob/master/prime-router/src/main/kotlin/Sender.kt#L206-L213) for Senders if there aren't two fragments in the specified Organization name whereas Receivers [would error](https://github.com/CDCgov/prime-reportstream/blob/master/prime-router/src/main/kotlin/Receiver.kt#L272-L278).

What are your thoughts?

stephenkao:
Would you mind adding some tests specifically for the `orgName.serviceName` fix you're adding here?  It looks like we're already checking against other Organizations [here](https://github.com/CDCgov/prime-reportstream/blob/master/prime-router/src/test/kotlin/history/azure/SubmissionFunctionTests.kt#L337) and [here](https://github.com/CDCgov/prime-reportstream/blob/master/prime-router/src/test/kotlin/history/azure/DeliveryFunctionTests.kt#L373), but it looks like we're missing tests that ensure data doesn't leak between services within the same Organization 🤔

prime-router/src/main/kotlin/history/azure/SubmissionFunction.kt
```diff
@@ -59,7 +67,7 @@ class SubmissionFunction(
 
         return submissionsFacade.findSubmissionsAsJson(
             userOrgName,
-            null, // currently, sending org client is not used but the functionality is there
```
stephenkao, line 62: 🙏

prime-router/src/main/kotlin/history/azure/DeliveryFunction.kt
thetaurean: I agree, but in `Receiver.kt:parseFullName()`, `Sender.fullNameSeparator` is also used. Intentionally or otherwise, I'm unsure.

I'm happy to change it in both places. However, I have a conceptualized refactor that would render that moot. More on that in a later reply.

prime-router/src/main/kotlin/history/azure/DeliveryFunction.kt
thetaurean: Thanks for the insightful questions!

All great observations, I share your desire to consolidate. This went through several revisions and I settled on treading lightly because:
- other approaches impacted functions used elsewhere in the project
- the classes mentioned are due for some refactoring:
  - `getOrgName` is actually a validator/getter (and that logic likely belongs elsewhere as you mentioned)
  - there's a strong case for a mix-in design pattern here to avoid some unnecessary OO pain
- as you mentioned, it felt out of scope for an already scope++ issue
- one more reason my neurons are playing hard-to-get with right now

> Out of curiosity, why are we checking for active status here and not [in SubmissionFunction](https://github.com/CDCgov/prime-reportstream/pull/8332/files#diff-c78040f89c39c1a4b5c02d72cfb1f1430853317a722a5f299f4eecb56dc0b405R41)? Senders and Receivers both should be able to be disabled, no?

In SubmissionFunction, we always have a fully qualified org name to retrieve the Sender (because `default` is always a valid Sender for an org). Following that callstack, we call `findSender()` -> `findSetting()` and the filter IS_ACTIVE on [699](https://github.com/CDCgov/prime-reportstream/blob/master/prime-router/src/main/kotlin/azure/DatabaseAccess.kt#L699).

In DeliveryFunction there is no equivalent fallback, so we need to peruse the list of Receivers which includes ACTIVE and INACTIVE customers. With that said, this comment may have been buried:

> **We may want to change the filtering here (or get last instance instead of first)**

I'm not sure that filter (or any at all) is appropriate, but intent was to match behaviors and only look at active Settings. Regardless, in this logic branch we are not filtering by sending service, so the sender retrieved is irrelevant as it's just used to validate the organization name (as long as it doesn't refer to an out-of-date orgname).

I will add a TODO and create a ticket for a refactor here that will cover consolidating this logic and ensuring no breakage in impacted callers.

thetaurean:
> Would you mind adding some tests specifically for the `orgName.serviceName` fix you're adding here?  It looks like we're already checking against other Organizations [here](https://github.com/CDCgov/prime-reportstream/blob/master/prime-router/src/test/kotlin/history/azure/SubmissionFunctionTests.kt#L337) and [here](https://github.com/CDCgov/prime-reportstream/blob/master/prime-router/src/test/kotlin/history/azure/DeliveryFunctionTests.kt#L373), but it looks like we're missing tests that ensure data doesn't leak between services within the same Organization 🤔 

Great idea w/ great timing! I am still wrapping up tests and will definitely add this case.

arnejduranovic:
It says this ticket fixes https://github.com/CDCgov/prime-reportstream/issues/8322 but that story is not groomed. It has to at least have acceptance criteria.

arnejduranovic:
The 404 error for submission when you provide a bad sender name returns: 

`{"error": "ignore.fgfgf: unknown ReportStream user"}`

Should we change "user" to "sender"?

prime-router/src/main/kotlin/history/azure/DeliveryFunction.kt
arnejduranovic: I agree with Stephen. My concern is if we create new tickets the work described above will never get done. If you are going to push this work, then you'll need to have created and linked the new tickets before approval.

JessicaWNava:
The endpoints worked as described for me once I took off the `/api` part. Just had the one comment about optional parameters. After that and some of other people's comments are resolved, I'll be happy to approve this!

thetaurean:
> The 404 error for submission when you provide a bad sender name returns:
> 
> `{"error": "ignore.fgfgf: unknown ReportStream user"}`
> 
> Should we change "user" to "sender"?
> 
> Names are case sensitive, should this be the case? for example, `ignore.hl7` does not work.

They should be case sensitive as far as I know. That would mirror behavior elsewhere.
Error message has been updated.

prime-router/src/main/kotlin/history/azure/DeliveryFunction.kt
thetaurean: I have created a ticket and added it to the TODO for the refactoring / consolidating work.
It still needs grooming, will be done today.

arnejduranovic (approved):

prime-router/docs/api/waters-delivery.yml
thetaurean: True for query, header, and cookie parameters; but not for path parameters.

> # Path Parameters
> Also remember to add required: true, because path parameters are always required.

Which is a bummer tbh :(

JessicaWNava (approved):
Looks good now!

Issue #7998, opened 2023-01-18
Typing non-numeric characters into date range filter fields errors out


stephenkao:
## Describe the bug
On the /daily-data and /submissions pages, typing a non-numeric character into the date range filter causes an error:

![Screenshot 2023-01-17 at 20.49.59.png](https://images.zenhubusercontent.com/5999b3b38f62dc7798c36942/2896e750-b638-4c06-b721-5825e2fb0a5a)

## Impact on ReportStream
This forces users to navigate to a different page and come back if they accidentally type in a non-numeric value.

## Steps to reproduce
1. Go to /daily-data or /submissions
2. Type a letter into one of the date range inputs
3. See the error

## Expected behavior
The date range inputs should handle non-numeric characters, but it should either disable the filter button or have an error message and still allow the user to correct the values.

PR #8348, opened 2023-02-15
7998 date range filter bug fix

penny-lischer:
This PR ...

Fixes bug - Typing non-numeric characters into date range filter fields errors out

Test Steps:
1. Log into RS
2. Select a State(For Daily Data) or a Federal(For Submissions) organization
3. Go to Daily Data and enter in a non-numeric character into the start range and/or end range
4. Click Filter
5. Notice that an error is not thrown and the table lists the default result set

## Changes
Added check for valid date before calling toISOString.
Added startDateHint to DateRangePicker
Screenshot 2023-02-15 at 3 25 44 PM

### Testing
- [x] Tested locally?
- [x] Added tests?

### Fixes
- #7998

### DIFF
--- a/frontend-react/src/components/Table/TableFilters.tsx
+++ b/frontend-react/src/components/Table/TableFilters.tsx
@@ -12,6 +12,7 @@ import {
     FALLBACK_TO,
     RangeSettingsActionType,
 } from "../../hooks/filters/UseDateRange";
+import { showError } from "../AlertNotifications";
 
 export enum StyleClass {
     CONTAINER = "grid-container filter-container",
@@ -26,9 +27,10 @@ interface SubmissionFilterProps {
 
 /* This helper ensures start range values are inclusive
  * of the day set in the date picker. */
-const inclusiveDateString = (originalDate: string) => {
-    let inclusiveDate = new Date(originalDate);
-    return new Date(inclusiveDate.setUTCHours(23, 59, 59, 999)).toISOString();
+const inclusiveDateString = (originalDateStr: string) => {
+    return new Date(
+        new Date(originalDateStr).setUTCHours(23, 59, 59, 999)
+    ).toISOString();
 };
 
 /* This component contains the UI for selecting query parameters.
@@ -47,11 +49,15 @@ function TableFilters({
      * state to hold these so updates don't render immediately after setting a filter */
     const [rangeFrom, setRangeFrom] = useState<string>(FALLBACK_FROM);
     const [rangeTo, setRangeTo] = useState<string>(FALLBACK_TO);
+    let from: string;
+    let to: string;
 
-    let from = !isNaN(Date.parse(rangeFrom))
-        ? new Date(rangeFrom).toISOString()
-        : "";
-    let to = !isNaN(Date.parse(rangeTo)) ? inclusiveDateString(rangeTo) : "";
+    try {
+        from = new Date(rangeFrom).toISOString();
+        to = inclusiveDateString(rangeTo);
+    } catch (e: any) {
+        showError("Invalid date value.");
+    }
 
     const updateRange = () => {
         filterManager.updateRange({
@@ -96,7 +102,7 @@ function TableFilters({
                         id: "start-date",
                         name: "start-date-picker",
                         onChange: (val?: string) => {
-                            if (val) {
+                            if (val && Date.parse(val)) {
                                 setRangeFrom(val);
                             }
                         },
@@ -107,7 +113,7 @@ function TableFilters({
                         id: "end-date",
                         name: "end-date-picker",
                         onChange: (val?: string) => {
-                            if (val) {
+                            if (val && Date.parse(val)) {
                                 setRangeTo(val);
                             }
                         },
--- a/frontend-react/src/components/Table/TableFilters.tsx
+++ b/frontend-react/src/components/Table/TableFilters.tsx
@@ -27,10 +27,8 @@ interface SubmissionFilterProps {
 /* This helper ensures start range values are inclusive
  * of the day set in the date picker. */
 const inclusiveDateString = (originalDate: string) => {
-    let inclusiveDateDate = new Date(originalDate);
-    return new Date(
-        inclusiveDateDate.setUTCHours(23, 59, 59, 999)
-    ).toISOString();
+    let inclusiveDate = new Date(originalDate);
+    return new Date(inclusiveDate.setUTCHours(23, 59, 59, 999)).toISOString();
 };
 
 /* This component contains the UI for selecting query parameters.
@@ -50,8 +48,10 @@ function TableFilters({
     const [rangeFrom, setRangeFrom] = useState<string>(FALLBACK_FROM);
     const [rangeTo, setRangeTo] = useState<string>(FALLBACK_TO);
 
-    let from = new Date(rangeFrom).toISOString();
-    let to = inclusiveDateString(rangeTo);
+    let from = !isNaN(Date.parse(rangeFrom))
+        ? new Date(rangeFrom).toISOString()
+        : "";
+    let to = !isNaN(Date.parse(rangeTo)) ? inclusiveDateString(rangeTo) : "";
 
     const updateRange = () => {
         filterManager.updateRange({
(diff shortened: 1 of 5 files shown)


frontend-react/src/components/Table/TableFilters.tsx
```diff
@@ -27,10 +27,8 @@ interface SubmissionFilterProps {
 /* This helper ensures start range values are inclusive
  * of the day set in the date picker. */
 const inclusiveDateString = (originalDate: string) => {
-    let inclusiveDateDate = new Date(originalDate);
-    return new Date(
-        inclusiveDateDate.setUTCHours(23, 59, 59, 999)
-    ).toISOString();
+    let inclusiveDate = new Date(originalDate);
```
etanb, line 30: nit: doesn't need to be `let`, can be `const`

frontend-react/src/components/Table/TableFilters.tsx
```diff
@@ -50,8 +48,10 @@ function TableFilters({
     const [rangeFrom, setRangeFrom] = useState<string>(FALLBACK_FROM);
     const [rangeTo, setRangeTo] = useState<string>(FALLBACK_TO);
 
-    let from = new Date(rangeFrom).toISOString();
-    let to = inclusiveDateString(rangeTo);
+    let from = !isNaN(Date.parse(rangeFrom))
+        ? new Date(rangeFrom).toISOString()
+        : "";
+    let to = !isNaN(Date.parse(rangeTo)) ? inclusiveDateString(rangeTo) : "";
```
etanb, line 54: We should just put the raw values `rangeFrom` into `updateRange` function. declaring them as variables and using that to typecheck them isn't ideal, nor is using them outside of a react hook like `useMemo` to make sure it has the proper dependencies.

My recommendation:

Let's do that `Date.parse(input)` type check within the onChange function:

```
onChange: (val?: string) => {
                            if (val && Date.parse(val)) {
                                setRangeTo(val);
                            }
                        },
```

`Date.parse()` returns `NaN` when it's a non-valid date which is falsy.

frontend-react/src/hooks/filters/mocks/MockFilterManager.ts
```diff
@@ -11,7 +11,7 @@ const fakeDispatch = <T>(): Dispatch<T> => {
     };
 };
 
-export const mockFilterManager = {
+export const mockFilterManager: FilterManager = {
```
jpandersen87, line 14: 🎉

jpandersen87 (approved):
once @etanb 's points are addressed, this looks gud to geaux

frontend-react/src/components/Table/TableFilters.tsx
stephenkao: Agree with Etan here about using `const` instead.

Alternatively, we don't even need to assign a variable since it's such a short-lived reference 🤷 
```tsx
return new Date(
    new Date(originalDate).setUTCHours(23, 59, 59, 999)
).toISOString();
```

frontend-react/src/components/Table/TableFilters.tsx
stephenkao: Agreed -- we should be validating/parsing prior to committing to state rather than doing it on the subsequent render 🙏

stephenkao:
@penny-lischer Any chance you could write some tests for the new logic?  Like `when typing in invalid characters into a date range input, expect to not throw an error`

frontend-react/src/components/Table/TableFilters.tsx
```diff
@@ -27,10 +27,8 @@ interface SubmissionFilterProps {
 /* This helper ensures start range values are inclusive
  * of the day set in the date picker. */
 const inclusiveDateString = (originalDate: string) => {
```
stephenkao, line 29: I realize this existed before, but when casting between types, it's nice to use a bit of Hungarian notation so it's clear what the type currently is.  Like `originalDateStr`/`originalDateString` in this case since `originalDate` is a little misleading 🤷 

Not a huge deal at all, just nice for readability 🤷 🤷 🤷

frontend-react/src/hooks/filters/mocks/MockFilterManager.ts
stephenkao: 🎉 but also hopefully 🧹/🔥 soon :)

etanb (approved):
LGTM!

stephenkao:
@penny-lischer @etanb @jpandersen87 Just pushed up [a change](https://github.com/CDCgov/prime-reportstream/pull/8348/commits/fb63c53610f7bcf3817004538dd15c973f83e12b) to just disable the 'Filter' button when the dates aren't well-formed or valid.  Let me know what you think!



I'll make tickets for a couple of follow-up items that are out of the scope of this pull request:
- Mocking the current date for all Jest tests to ensure current Date instances are deterministic
- Adding sensible defaults for the date range filter instead of years 2000 - 3000
- Having the input date range values persist between fetches: this doesn't happen on the current implementation
- Removing FilterManager and CursorManager 🔥

frontend-react/src/components/Table/TableFilters.tsx
```diff
@@ -25,13 +25,14 @@ interface SubmissionFilterProps {
...
 
-/* This helper ensures start range values are inclusive
- * of the day set in the date picker. */
-const inclusiveDateString = (originalDateStr: string) => {
-    return new Date(
-        new Date(originalDateStr).setUTCHours(23, 59, 59, 999)
-    ).toISOString();
-};
+const DATE_RE = /^[0-9]{1,2}\/[0-9]{1,2}\/[0-9]{4}$/;
+
```
etanb, line 29: Does a valid year HAVE to have the full year associated with it? ie, `01/01/2023` vs `01/01/23`?

Also, since we're checking with `Date.parse()` which checks for valid dates (no, months that don't exist, proper formatting etc), what is `DATE_RE` offering on top? Just curious! Might be worth a comment since I'm missing it.

jpandersen87: We should probably favor using momentjs to validate the dates rather than regex. see: https://momentjs.com/docs/#/parsing/string-format/

etanb: Correct me if i'm wrong @stephenkao but we wanted to slowly move away from Moment JS since it's unmaintained now? IE [from their docs](https://momentjs.com/docs/#/-project-status/):

```
We now generally consider Moment to be a legacy project in maintenance mode. It is not dead, but it is indeed done.
```

jpandersen87: Interesting. I was not aware of its transition to maintenance mode. If there is a way to handle validation cleanly with the default Date object then that's def OK, I just assumed it probably wasn't. If it's not, we may need to use momentjs even if temporary if we decide to move to another datetime library instead of waiting for temporal.

stephenkao: > Does a valid year HAVE to have the full year associated with it? ie, `01/01/2023` vs `01/01/23`?
> 
> Also, since we're checking with `Date.parse()` which checks for valid dates (no, months that don't exist, proper formatting etc), what is `DATE_RE` offering on top? Just curious! Might be worth a comment since I'm missing it.

@etanb No problem!  `Date` has different implementations across browsers, and the regular expression is just an extra check to make sure the user is inputting something that can be understood by all of them.  Like `new Date('11')` returns a Date instance of 11/01/2001 in Chrome but throws an `Invalid Date` error in Firefox.  It's not absolutely necessary, but it's just an added check so I can definitely add a comment explaining that + update to allow for two or four year digits.

@jpandersen87 Yeah, I think we should be at least reevaluating moment.js because it doesn't provide much more functionality than what the browser API offers now -- at least not for our use cases.  I'm personally okay with using a regular expression in lieu of moment.js for this particular instance since it's checking against the format of the input rather than the date itself, which is handled through `Date.parse`.  

We still have references to moment.js in our codebase so I'm not advocating for its immediate removal; I just want us to keep in mind that it may not be necessary, and we can reevaluate its inclusion in this project in a few months once we have a better understanding of how much we really need it.  What are your thoughts?

stephenkao: Updated [here](https://github.com/CDCgov/prime-reportstream/pull/8348/commits/1ad2e1d2867e9bd9102f1c4f317d8be449a9603d).  Let me know what y'all think!

stephenkao:
Test test: 

https://app.zenhub.com/workspaces/experience-607d9d5e68b95200150fec37/issues/gh/cdcgov/prime-reportstream/7998
https://github.com/CDCgov/prime-reportstream/issues/7998

etanb (approved):

jpandersen87 (approved):

stephenkao:
Merging now!  Thanks, everyone!

PR #8349, opened 2023-02-16
Upgrade to msw 1.0.x

### LINKED ISSUE #8340: Update msw to 1.0.X
Upgrade msw from 0.36.8 to 1.0.X. Will still not be able to upgrade to Node 18 sadly.

jpandersen87:
Fixes #8340 
This PR:
- updates msw from 0.36.8 to 1.0.X
- provides a new test utility that accepts matcher strings to suppress expected console warnings and errors via jest mockImplementation
- utility returns a function that can be called to restore console to original state
- uses said utility in various areas that tested error and warning handling
- fixes some errors that occurred in tests (these errors were not failure triggers for said tests)
- various tests were fixed due to behavior changes on hook tests from msw update
- changes to QueryClient instantiation to ensure a new client is created on every render to fix new errors likely related to poor isolation
- various hook tests that were bundled together were isolated into their own modules to fix new errors likely related to poor isolation

Hopefully with this update we'll be ready for a simple dependency bump when msw gets node 18 support.


snesm:
Related to #8404

frontend-react/src/__mocks__/WatersMockServer.ts
```diff
@@ -18,12 +18,12 @@ export enum WatersTestHeaderValue {
 const handlers = [
     rest.post(`${RS_API_URL}/api${WatersUrls.VALIDATE}`, (req, res, ctx) => {
         if (
-            req.headers["_headers"][WatersTestHeader.CLIENT] ===
+            req.headers.get(WatersTestHeader.CLIENT) ===
```
stephenkao, line 21: 🙏

frontend-react/src/components/Admin/AdminReceiverDashboard.tsx
```diff
@@ -547,11 +547,13 @@ function renderAllReceiverRows(props: {
                     </div>
                 </Grid>
                 <ScrollSyncPane enabled>
-                    <Grid row className={"horizontal-scroll"}>
-                        <Grid row className={"week-column"}>
-                            {perDayElements}
+                    <>
```
stephenkao, line 550: Out of curiosity, why is this needed?

frontend-react/src/components/Admin/EditSenderSettings.test.tsx
```diff
@@ -33,8 +34,8 @@ jest.mock("react-router-dom", () => ({
     },
     useParams: () => {
         return {
-            orgName: "abbott",
```
stephenkao, line 36: Welp

frontend-react/src/components/Admin/EditSenderSettings.test.tsx
```diff
@@ -64,8 +65,15 @@ const testKeys = JSON.stringify([
 const testProcessingType = "sync";
 
 describe("EditSenderSettings", () => {
-    beforeAll(() => settingsServer.listen());
-    afterEach(() => settingsServer.resetHandlers());
+    beforeAll(() => {
+        settingsServer.listen();
+        settingsServer.use(
+            rest.get(
+                "https://test.prime.cdc.gov/api/settings/organizations/abbott/senders/user1234",
```
stephenkao, line 72: I'm wondering if we should use the `REACT_APP_BACKEND_URL` environment variable for the base path here.  Also, I think it'd be good to make sure this is an invalid base path so we're not actually making calls to test.prime.cdc.gov since that's real.

Not an issue with Jest tests since they should be sandboxed, but I'm thinking we could adopt the pattern for all tests, including Cypress. 🤔

frontend-react/src/components/FileHandlers/FileHandlerMessaging.test.tsx
```diff
@@ -63,6 +64,7 @@ describe("FileSuccessDisplay", () => {
 
 describe("FileErrorDisplay", () => {
     test("renders expected content", async () => {
+        const restore = conditionallySupressConsole("failure:");
```
stephenkao, line 67: Gonna add this to the upcoming tickets so we don't `console.error` on all validator errors 😬

frontend-react/src/components/Admin/AdminReceiverDashboard.tsx
jpandersen87: A react ref error was being thrown that could only be resolved by giving the scroll panel a class-based component.

frontend-react/src/hooks/UseCreateFetch.ts
```diff
@@ -45,11 +45,12 @@ function createTypeWrapperForAuthorizedFetch(
             headers,
         });
 
-        return axios(axiosConfig)
-            .then(({ data }) => data)
-            .catch((e: AxiosError) => {
-                throw new RSNetworkError<T>(e as AxiosError<T>);
-            });
+        try {
```
stephenkao, line 48: 🙏

frontend-react/src/hooks/UseValueSetActivation.test.ts
```diff
@@ -0,0 +1,50 @@
...
+    beforeAll(() => lookupTableServer.listen());
+    afterEach(() => lookupTableServer.resetHandlers());
+    afterAll(() => lookupTableServer.close());
+
+    const renderWithQueryWrapper = () =>
+        renderHook(() => useValueSetActivation(), {
+            wrapper: QueryWrapper(),
+        });
+
+    test("returns trigger and loading indicator", async () => {
```
stephenkao, line 18: Does this need to be `async`?

frontend-react/src/hooks/UseValueSets.ts
```diff
@@ -92,10 +92,15 @@ export const useValueSetsMeta = (
     dataTableName: string = LookupTables.VALUE_SET
 ): ValueSetsMetaResponse => {
     const { authorizedFetch, rsUseQuery } = useAuthorizedFetch<LookupTable[]>();
+    const memoizedDataFetch = useCallback(
```
stephenkao, line 95: Why does this need to be memoized?  Even if the `authorizedFetch` reference changes, react-query should still prevent another request if the query key doesn't change, no?

frontend-react/src/hooks/filters/mocks/MockFilterManager.ts
```diff
@@ -7,7 +7,7 @@ import { PageSettingsAction } from "../UsePages";
 
 const fakeDispatch = <T>(): Dispatch<T> => {
     return (_v: T) => {
-        console.log(_v);
+        // console.log(_v);
```
stephenkao, line 10: Eh?

frontend-react/src/network/QueryClients.ts
```diff
@@ -12,7 +12,12 @@ export const appQueryClient = new QueryClient({
...
     },
 });
-export const testQueryClient = new QueryClient({
-    // to allow for faster testable failures
-    defaultOptions: { queries: { retry: false } },
-});
+export const getTestQueryClient = () =>
+    new QueryClient({
+        // to allow for faster testable failures
+        defaultOptions: {
```
stephenkao, line 18: Since our approach is moving toward instantiating a new QueryClient for each test case, should we use [the same configuration as in production](https://github.com/CDCgov/prime-reportstream/pull/8349/files#diff-aa5131257f5b08b05099cac4d6c0fa3c367c4f16059043512369f72d668d6494R6-R11) so the behavior is the same?

frontend-react/src/utils/TestUtils.ts
```diff
@@ -20,4 +20,35 @@ const mockEvent = (mock?: Partial<any>) => {
     };
 };
 
-export { mockEvent };
+function conditionallySupressConsole(...matchers: string[]) {
```
stephenkao, line 23: Thanks for adding this!  Super useful!

frontend-react/src/utils/TestUtils.ts
```diff
@@ -20,4 +20,35 @@ const mockEvent = (mock?: Partial<any>) => {
...
+            }
+        });
+
+    return () => {
+        jestError.mockRestore();
+        jestWarn.mockRestore();
+    };
+}
+
+export { mockEvent, conditionallySupressConsole };
```
stephenkao, line 54: This is a stupid, pedantic, and useless point, but I feel like inline exports (`export function blahblahblah() { ... }`) are easier to think about than scrolling up/down in the file to make sure they're exported from a specific file.  Tiny, tiny subjective thing, though!

frontend-react/src/utils/TestUtils.ts
stephenkao: That said, I think we should use caution when using this.  It looks like most/all of the warnings/errors in this pull request shouldn't be escalated to the test log, but I'm wary of overusing this just to create a silenced log.  Our first inclination should be to look into the why warnings/errors are getting printed out for a given test and silence them only if there's nothing actually fishy with said test.  We can also explicitly `expect` them if they're intentional, of course.

Not asking for any changes -- just something for the team to keep in mind!

stephenkao (approved):
Looks great!  Thanks for tackling this!

frontend-react/src/utils/TestUtils.ts
jpandersen87: Very true. This is only to be used for scenarios where we are expecting console messages (error handling testing and such) to reduce noise to see real console messages that need addressing. I was debating on whether the return should be a tuple or object containing the spy objects so that it's possible to check the call counts and such.

stephenkao: No worries!  I think it's fine to keep this a separate function named as such so it gets attention called to it during code review.  Otherwise, we can manually set up spies when we're intending to catch warnings/errors.

One small, small thing, though -- I think `suppress` has two `p`s 😆

frontend-react/src/network/QueryClients.ts
jpandersen87: I ran into issues where turning on the suspend feature created errors when testing hooks in isolation.

frontend-react/src/hooks/filters/mocks/MockFilterManager.ts
jpandersen87: The magic of main-merging

frontend-react/src/hooks/UseValueSets.ts
jpandersen87: This was from earlier on when troubleshooting errors (value sets was a real trouble area) and I was throwing everything at the wall to see what would stick (aka stop erroring). I can revert if needed.

frontend-react/src/components/Admin/EditSenderSettings.test.tsx
jpandersen87: Funny enough, I'm pretty sure with this msw update preflight calls were succeeding to test.prime.cdc.gov which only stopped once I started isolating hook tests.

frontend-react/src/hooks/filters/mocks/MockFilterManager.ts
jpandersen87: Woops misread the diff. I commented out the console log to reduce test noise. Should this log stay?

frontend-react/src/components/Admin/AdminReceiverDashboard.tsx
stephenkao: Huh!  That's very odd...I wonder if it has something to do with [`childRef` not being used](https://github.com/okonet/react-scroll-sync/blob/master/src/ScrollSyncPane.js#L36) 🤔 

In any case, if a Fragment works for our cases (and it looks like it does), that's fine enough for me.

frontend-react/src/hooks/filters/mocks/MockFilterManager.ts
stephenkao: Let's remove the log call, but I feel like each of these dispatches could be replaced with a spy instead.  Not a huge deal either way -- hoping we can get off of FilterManager and CursorManager soon.

frontend-react/.env.test
```diff
@@ -3,8 +3,8 @@ REACT_APP_TITLE=CDC Prime ReportStream TEST
 REACT_APP_DESCRIPTION=ReportStream Dashboard Application
 REACT_APP_ICON=https://test.reportstream.cdc.gov/assets/cdc-logo.svg
 
-REACT_APP_BASE_URL=https://test.reportstream.gov
-REACT_APP_BACKEND_URL=https://test.prime.cdc.gov
+REACT_APP_BASE_URL=https://localhost
```
stephenkao, line 6: 🙏

stephenkao (approved):
Looks great again! :)

PR #8392, opened 2023-02-17
Non-breaking visual hiccups from USWDS 3.0 upgrade

### LINKED ISSUE #8361: Non-breaking visual hiccups from USWDS 3.0 upgrade
## Describe the bug
There are some small visual differences between our usage of USWDS 2.0 versus USWDS 3.0.  

One example is the login page, but there could be more:

Live on production:

Before the USWDS 3.0 bump:

## Impact on ReportStream
The impact is minimal since they're likely only visual differences.

## Steps to reproduce
1. Check against the production and the pre-USWDS 3.0 commit to see the differences.

## Expected behavior
There should be no visual differences between production and the pre-USWDS 3.0 commit.  However, what's more important is just making sure the website doesn't appear buggy so we can concentrate on that if there would be larger structural changes otherwise.

jpandersen87:
Fixes #8361

This PR fixes the visual bugs from the react-uswds/uswds dependency update by targeting @uswds/uswds 3.1.0 only. Version 3.2.0+ has breaking changes in their styles which conflicts with the 3.1.0-based css react-uswds outputs. Some changes to scss files were made to conform with 3.X.X expectations.

Before:
image

After:
image


frontend-react/src/styles/_uswds_settings.scss
```diff
@@ -1,73 +1,66 @@
...
 
 $flex-wrap-settings: (
   responsive: true
-);
-
-// COLORS - USDS redefine
-// these are color names we can reference else where and/or redefine usds colors
-$black: #010101;
-$medium-gray: #d9d9d9;
+),
```
etanb, line 64: nit of nits, can these be indented?

jpandersen87: Good catch, thanks. Fixed.

frontend-react/src/global.scss
```diff
@@ -7,17 +7,17 @@
...
-
-@import "@trussworks/react-uswds/lib/index.css";
+@forward "styles/uswds_settings";  // _uswds_settings.scss
+@forward 'uswds';
 
+@use 'uswds';
+@import "~@trussworks/react-uswds/lib/uswds.css";
+@import "styles/uswds_custom_theme.scss";
 @import "styles/uswds_overrides"; // _uswds_overrides.scss
 @import "styles/color"; // _color.scss
```
etanb, line 17: I was going to say, we should add the colors to `_color.scss` here, but this file isn't even used? I checked all 3 of the SASS variables and they're not referenced anywhere. Might be worth removing.

etanb:
@jpandersen87 is it possible to post some before/after photos for posterity? 🙏

jpandersen87:
> @jpandersen87 is it possible to post some before/after photos for posterity? 🙏

before/after added

frontend-react/src/styles/_uswds_settings.scss
```diff
@@ -1,73 +1,63 @@
...
+@use "uswds-core" with (
+  /*
+    These are necessary for the image and font urls referenced in the source
+    files to resolve correctly.
+
+
+      these paths are from where the output css lives, in this case
+              /frontend-react/src/content/generated/global.out.css
+  */
+  $theme-image-path: '@uswds/uswds/img',
```
penny-lischer, line 13: Is this the same image as what was in '@uswds/uswds/dist/img'?

penny-lischer (approved):
If yes, to my comment then LGTM

frontend-react/src/styles/_uswds_settings.scss
jpandersen87: if you mean $theme-hero-image, the png was replaced for a jpg at some point. since we're seemingly not changing the image from what uswds supplies, I just removed the variable from our file.

etanb (approved):
LGTM!

frontend-react/src/styles/_color.scss
```diff
@@ -1,6 +0,0 @@
-// CDC color variables
-// 7/22 Testing these out, not used in production
```
stephenkao, line 2: Uhhhhh 😆

stephenkao (approved):
Looks great to me too!  Thanks for hopping on this!

PR #8430, opened 2023-02-21
Experience-6693: Add doc for front-end test patterns

stephenkao:
This changeset adds an initial doc for test conventions we want to maintain in frontend-react.

Still some todos, which will be broken out into a separate ticket or two:
- Mocking and tautological tests
- Different types of tests (unit, integration, end-to-end) and when to use them
- Jest tips
- Cypress tips

### DIFF
--- a/frontend-react/docs/test-conventions.md
+++ b/frontend-react/docs/test-conventions.md
@@ -1,0 +1,504 @@
+# Testing Conventions
+
+## Guiding Principles
+
+Automated tests are an integral part of ongoing maintenance, and they help to fill in the cracks that manual testing inevitably leaves behind.  Of course, that's not to say that manual testing has no purpose; manual testing and automated testing are both essential at increasing the surface area of overall test coverage.
+
+To that end, we should subject our test code to the same rigor and scrutiny that we would our source code; tests should be clean, deterministic, and readily understandable for the benefit of anyone working in the codebase, regardless of how long they've been working in it.
+
+### Tests should read like a manual for source code
+
+Tests are often one of the first places to look when dealing with familiar code because they offer insight into how specific pieces function in isolation.  Because of this, we should strive to make our tests read like a manual for the related source code: "given this input, I should expect this output" or more practically, "given these conditions, I should expect this behavior."  Most concepts in programming should be able to be boiled down into inputs and outputs, and tests are no different in that regard.
+
+```tsx
+function sum(...numbers: number[]) {
+    return numbers.reduce((total, value) => total + value, 0);
+}
+
+// bad - what does it mean to function correctly?
+describe("sum", () => {
+    test("functions correctly", () => {
+        expect(sum()).toEqual(0);
+
+        expect(sum(1)).toEqual(1);
+        expect(sum(1, 2)).toEqual(3);
+        expect(sum(1, 2, 3)).toEqual(6);
+        expect(sum(1, 10, 100, 1000)).toEqual(1111);
+    });
+});
+
+// good - use descriptions to explain what's actually happening
+describe("sum", () => {
+    describe("with no arguments", () => {
+        test("returns 0", () => {
+            expect(sum()).toEqual(0);
+        });
+    });
+
+    describe("with arguments", () => {
+        test("returns the sum of all the arguments", () => {
+            expect(sum(1)).toEqual(1);
+            expect(sum(1, 2)).toEqual(3);
+            expect(sum(1, 2, 3)).toEqual(6);
+            expect(sum(1, 10, 100, 1000)).toEqual(1111);
+        });
+    });
+})
+```
+
+> Why?
+>
+> Having clearer and more sensible descriptions calls attention to what the purpose of the test actually is, which makes it easier to follow along with the underlying logic or to group similarly intentioned tests together.
+
+
+### Tests should check behavior, not implementation
+
+Especially in the world of front-end code, it's especially important to test our code from a user's perspective -- this means our tests should be prioritizing checking behavior ("what happens") rather than implementation details ("how it happens").
+
+```tsx
+function List({ items }: { items: string[] }) {
+    return (
+        <ul className="bg-light-gray">
+            { items.slice(0, 10).map((item) => (
+                <li className="bg-white" key={item}>{item}</li>
+            )) }
+        </ul>
+    );
+}
+
+// bad - don't test markup or classes, which are more likely to change over time
+test("renders ul with bg-light-gray", () => {
+    expect(ulNode).toHaveClass("bg-light-gray");
+});
+
+// good - check what the user would see
+describe("when there are ten or more items", () => {
+    test("renders ten items", () => {
+        expect(liNodes).toHaveLength(10);
+    });
+});
+```
+
+> Why?
+>
+> Testing behavior keeps tests agnostic to the implementation, which makes them more resistant to changes and refactors.  Ultimately, changing implementation shouldn't automatically imply needing to update tests unless the behavior is also changing.
+
+Somewhat of an exception to the aforementioned rule is pure functions, whose behavior is essentially its implementation:
+
+```tsx
+// source code
+function capitalize(str?: string) {
+    if (!str) {
+        return "";
+    }
+    
+    return str.charAt(0).toUpperCase() + str.substring(1).toLowerCase();
+}
+
+// tests
+describe('capitalizeString', () => {
+    describe("when the provided string is falsy", () => {
+        test("returns an empty string", () => {
+            expect(capitalizeString()).toEqual("");
+            expect(capitalizeString("")).toEqual("");
+        });
+    });
+
+    describe("when the provided string is truthy", () => {
+        test("returns the string with the first character capitalized (if applicable)", () => {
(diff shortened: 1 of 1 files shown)


frontend-react/docs/test-conventions.md
```diff
@@ -0,0 +1,492 @@
...
+        // assume a POST is sent to create a user session
+    }
+    
+    return (
+        <form onSubmit={handleSubmit}>
+            <button type="submit">
+        </form>
+    );
+}
+```
```
jpandersen87, line 70: Is the test-side of this example missing?

frontend-react/docs/test-conventions.md
```diff
@@ -0,0 +1,492 @@
...
+// good
+expect((sum(1, 2, 3))).toEqual(6);
+expect(
+    concatenate(arrayA, arrayB, arrayC)
+).toEqual([1, 2, 3, 4, 5, 6, 7, 8, 9]);
+```
+
+> Why?
+>
+> Tests should read as superficially clearly as possible, and they shouldn't reimplement the underlying logic.
```
jpandersen87, line 142: Should we be clearer here that the expectation is that the compare value, no matter how complex, needs to precalculated?

frontend-react/docs/test-conventions.md
```diff
@@ -0,0 +1,492 @@
...
+> Why?
+>
+> It's still a good idea to keep an eye on the test coverage percentage as loose guidance (70% - 80%) as a metric for confidence in our code, but 100% coverage is a fallacy.  Especially on the client side where much of our code will be fed through a renderer verbatim, it's less important to attain "complete" coverage, and it may encourage an anti-pattern of introducing tautological tests.
+
+Consider the following example:
+
+```tsx
+// given an array of words, return a string that formats them into a human-readable list  
+// ex) ["apple", "banana", "cantaloupe", "durian"] ==> "apple, banana, cantaloupe, and durian"  
+function toSentence(words: string[]) {
```
jpandersen87, line 155: There's a really good existing example case here with the react-query hook's return values (like isLoading) not needing to be tested.

frontend-react/docs/test-conventions.md
```diff
@@ -0,0 +1,492 @@
...
+        test("does not render the fallback message", () => {
+            expect(screen.queryByText("This is a fallback message.")).toBeNull();
+        });    
+    });
+});
+```
+
+> Why?
+>
+> It's redundant to test third-party library logic, even implicitly, because that logic is entirely out of our control and really should already be tested by the maintainers of said library.
```
jpandersen87, line 345: Woops, looks like you cover this here kinda in regards to my react-query example.

frontend-react/docs/test-conventions.md
```diff
@@ -0,0 +1,492 @@
...
+        test("clears the form", () => {
+            ...
+        });
+    });
+});
+```
+
+> Why?
+>
+> By organizing our tests into sets of conditions, it becomes much easier to add in more expectations in the future to account for new features and regressions.
```
jpandersen87, line 431: It might be important to note that before* and after* functions are scoped to the parent describe (at least I think with some experience I had with the msw update).

jpandersen87 (commented):
Some questions but overall is gud!

frontend-react/docs/test-conventions.md
stephenkao: Hahah whoops, pushed up a outdated version of this file.  Fixed!

frontend-react/docs/test-conventions.md
stephenkao: Yeah, that's a good point!  I'd say the compare value doesn't necessarily have to be precalculated as long as it's clear (could be through variable names like `VALID_VALUE`/`TODAY`/`ONE_MILLION_SECONDS`) at first glance and doesn't require the reader of the test to trace through logic.

frontend-react/docs/test-conventions.md
stephenkao: No problem!  Happy to update this for something a bit more practical with `react-query` -- do you have that example ready?  Or is it just something like the following?

```tsx
export function useSomething() {
    const { isLoading } = useQuery(...);

    return {
        isLoading
    };
}

test(useSomethingReturnValue.isLoading).toEqual(true); // redundant to test react-query logic
```

frontend-react/docs/test-conventions.md
stephenkao: Good point!  Updated [here](https://github.com/CDCgov/prime-reportstream/pull/8430/files#diff-42fae80a7a121e5a9d806d505777c76a86e871122ef5bb86be2646e4ae4e5cf6R409).  Let me know if you think even more clarity is needed here!

frontend-react/docs/test-conventions.md
jpandersen87: Pretty much that exactly. We shouldn't test third party logic, but it's OK to rely on it to know when to call an expect (waitFor isLoading to be false).

stephenkao: Makes sense to me!  I'll add that as an example too -- thanks!

frontend-react/docs/test-conventions.md
```diff
@@ -0,0 +1,504 @@
...
+test("renders ul with bg-light-gray", () => {
+    expect(ulNode).toHaveClass("bg-light-gray");
+});
+
+// good - check what the user would see
+describe("when there are ten or more items", () => {
+    test("renders ten items", () => {
+        expect(liNodes).toHaveLength(10);
+    });
+});
```
etanb, line 79: It's true that we should avoid testing CSS classes, but markup is inevitable. Perhaps not something as specific as `expect <span>` or something, but we're always looking for DOM nodes in a way. I wonder if we can lay best practices there? We could even get as explicit as using `data` attributes for nodes we're using and expecting for tests?

frontend-react/docs/test-conventions.md
```diff
@@ -0,0 +1,504 @@
...
+        test("does not render the fallback message", () => {
+            expect(screen.queryByText("This is a fallback message.")).toBeNull();
+        });    
+    });
+});
+```
+
+> Why?
+>
+> It's redundant to test third-party library logic, even implicitly, because that logic is entirely out of our control and really should already be tested by the maintainers of said library.
```
etanb, line 355: Can you elaborate on 3rd party library tests? I totally understand not creating tests that just touch the base library, ie for a `date-picker` to just have a test that says something like "click date component, expect `date-picker` exists", BUT we should be testing how that 3rd party library lives in our UX flow, ie "click date component, select 1/2/23 in `date-picker` expect valid date banner to appear in UI" or some such.

etanb:
Is it worth adding a section about mock data and perhaps redoing our test inputs? I was hoping we could make a single file with all of our test mock data so we can see it in one place. We want our data to be as realistic as possible, and could potentially look into something like [Chance JS](https://chancejs.com/) for generating realistic random strings.

frontend-react/docs/test-conventions.md
stephenkao: Sure, that's a good point; we're always going to be testing markup implicitly, but it's rarely important to test markup _explicitly_.  Will clarify!

stephenkao: Actually, RTL has a great doc about this already: https://testing-library.com/docs/queries/about/#priority

Maybe we could just link to this as our guideline on this point?

frontend-react/docs/test-conventions.md
stephenkao: Yeah, definitely can clarify!  In your example, I'd say that the linking of the date picker to the date banner flow falls under the category of custom logic that should be tested.  I feel like your point and Joseph's point about assumptions are somewhat related so I'll add a block that hopefully covers the two.

stephenkao:
> Is it worth adding a section about mock data and perhaps redoing our test inputs? I was hoping we could make a single file with all of our test mock data so we can see it in one place. We want our data to be as realistic as possible, and could potentially look into something like [Chance JS](https://chancejs.com/) for generating realistic random strings.

Sure, makes sense to add a "mocking" section!  But I think I'm going to split that out into a separate fast-follow ticket -- I'd like to just get these two docs (test conventions + best practices) merged in first so all of us can start contributing to them as needed.

Regarding doing all of our mocks over again, I think we should settle on a mocking library and then replace as we encounter old tests and write new ones.  In my opinion, redoing all the mock data for our tests at once wouldn't be a worthwhile use of our time -- if they're working and testing things correctly right now, there's no need to touch them unless we need to update them.  

For the mocking library, we added [a ticket](https://app.zenhub.com/workspaces/web-experience-607d9d5e68b95200150fec37/issues/gh/cdcgov/prime-reportstream/8191) for Faker + Fishery back then, but I'll update it to just be focused on choosing a mock library.  Personally, I'd push for Faker over Chance just because the former has more types and namespacing, but definitely open to hearing your thoughts!

frontend-react/docs/test-conventions.md
etanb: Yeah! Love that idea. Especially that `getByTestId` both exists AND should be used as a last resort.

PR #8452, opened 2023-02-22
chore: update yarn.lock

stephenkao:
Just noticed this smallllllll update to the yarn.lock when installing on Node v16.19.1


jpandersen87 (approved):

stephenkao:
Thanks, Joseph!

PR #8454, opened 2023-02-22
Engagement/angela/8259 log stlt status codes

### LINKED ISSUE #8259: REST - Log the status code returned from STLT
## Problem statement

Every post to a REST endpoint will be returned with a message and status code. Currently we are logging the message but not the status code. That status code can provide additional clarity in the event that the message is not descriptive.

## Acceptance criteria

- [ ] Status code is logged to action table in http status column

## To do

- [ ] review REST post code in RESTTransport.postReport()
- [ ] write to isolate status code from response and log with action history class

adegolier:
Test Steps:
I modified the existing tests to check for this and verified that the same check fails in master without this merge, but unsure how to manually test and validate. There does not appear to be any integration tests that pass all the way through the RESTTransport logic and into the database to do manual db validation after running tests.

## Changes
- Added a function to Action History to allow tracking by a given message and HTTPStatusCode
- Changed places in RESTTransport.send() to use that function instead of one that only tracked the message
- Changed RESTTransport.postReport() to return a HTTPResponse instead of a String
- Updated tests to handle new return type on postReport() and to check that the status code of the Action in ActionHistory is not null

### Testing
- [ ] Tested locally? - Unsure how to do this
- [X] Added tests? (More like modified existing tests to check for this addition)

## Linked Issues
- Fixes #8259


brick-green (approved):
great stuff!

Issue #7247, opened 2022-11-04
UP Sender Transform can add new data to a FHIR bundle


carlosfelix2:
### User Story:
As a receiver, I want sender data to be corrected and/or normalized, so I can receive correct data.

### Description/Use Case
Similar to the COVID Pipeline, we need to have the ability to transform sender data to be able to fix and/or normalize the data.  One example of this data problem would be the use of different strings to mean the same thing (e.g. Y, y, yes, positive). Some senders may also use non-standard data that is unique to them, so a generic approach for all senders may not work and hence why we need sender specific transforms.

### Risks/Impacts/Considerations
* Adding new data to a FHIR bundle is tricky as the proper data type needs to be created for all parent levels before a value can be assigned and values assigned to a property need to match the data type as well.
* There may be limitations on creating new data for certain resources that we may need special code for, but we can tackle those when we get a use case for them.

### Dev Notes:
* See the code included in the design attached to issue #6761 to create new FHIR properties and resources.  This code is a prototype and requires proper error checks and testing. The design is attached in the comments of the above linked issue.
* Make sure to test various levels of adding resources and properties in your unit tests.
* See FhirBundleHelpers.kt to see how new elements are added during the routing step

### Acceptance Criteria
* The UP sender transform can add non-existing resources and properties to a FHIR bundle.

PR #8456, opened 2023-02-22
UP sender transform existing data

### LINKED ISSUE #7246: UP Sender Transform can transform existing data in non-repeating FHIR resources
### User Story:
As a receiver, I want sender data to be corrected and/or normalized, so I can receive correct data.

### Description/Use Case
Similar to the COVID Pipeline, we need to have the ability to transform sender data to be able to fix and/or normalize the data.  One example of this data problem would be the use of different strings to mean the same thing (e.g. Y, y, yes, positive). Some senders may also use non-standard data that is unique to them, so a generic approach for all senders may not work and hence why we need sender specific transforms.

* 

### Dev Notes:
* Note that this is to update existing data in the FHIR bundle, so you do not have to worry about how to create new FHIR properties at this time.
* Reference the design #6761

### Acceptance Criteria
* The UP sender transform can set an element's focus resource based on a FHIR Path expression.
* The UP sender transform checks for a condition that uses FHIR Path to be true before proceeding to evaluate the rest of a schema element.
* The UP sender transform can set an existing non-repeating (e.g. Patient and not DiagnosticReport) FHIR bundle property to a value specified using FHIR Path.
* The UP sender transform can evaluate multiple value expressions and use the first expression that has a valid non-empty value.

jimmyfagan:
This PR adds the ability to use Fhir Transform Schemas to make adjustments to fhir bundles. Most of the new functionality in `FhirTransformer.kt` is based off of similar functionality in `FhirToHl7Converter.kt`. The `transform` method allows for fhir bundles to be transformed based on a fhir transform schema. As demonstrated by the new tests, this supports the following:

- updating existing properties based on a FHIR Path specified value
- using conditional property updates
- adding some properties to the bundle if they don't yet exist
- using `resource` on a schema element to set the focus resource of the bundle
- using the first of a list of values that has a valid non-empty value for property updates

Test Steps:
1. Since this is another intermediate step along the process of getting this functionality into the Universal Pipeline, at this point, this is only testable through the written tests. There are new tests in `FhirTransformerTests.kt` and `FhirBundleUtilsTests.kt`, and some existing tests were updated (mostly minor changes due to small updates to existing function signatures).

## Changes
- Added `FhirTransformer.kt`, based on `FhirToHl7Converter.kt`, to handle processing of FHIR transform schemas on a bundle
- Common elements were extracted from `FhirToHl7Converter.kt` and placed into `ConfigSchemaProcessor.kt`. Part of this included making a distinction between `getValue` and `getValueAsString`, also changed params for `getFocusResources`. Updated tests accordingly.
- Added `FhirBundleUtils.kt` to support updates using values of different fhir types
- Refactored existing code to change `FHIRTransformSchema` to `FhirTransformSchema` and similar.

### Testing
- [x] Tested locally?
- [x] Added tests?

### Process

## Linked Issues
- Fixes #7246
- Also relevant to #7247, since some (or maybe all) of that work was handled in these changes


prime-router/src/test/kotlin/fhirengine/translation/hl7/FhirTransformerTests.kt
```diff
@@ -0,0 +1,263 @@
...
+import assertk.assertions.isFalse
+import gov.cdc.prime.router.fhirengine.translation.hl7.schema.fhirTransform.FHIRTransformSchemaElement
+import gov.cdc.prime.router.fhirengine.translation.hl7.schema.fhirTransform.FhirTransformSchema
+import gov.cdc.prime.router.fhirengine.translation.hl7.utils.CustomContext
+import gov.cdc.prime.router.fhirengine.translation.hl7.utils.FhirPathUtils
+import org.hl7.fhir.r4.model.Bundle
+import org.hl7.fhir.r4.model.Patient
+import kotlin.test.Test
+
+class FhirTransformerTests {
```
arnejduranovic, line 17: I like these tests, but should we add more unit tests to test all the paths in the other worker functions like `setBundleProperty`? Did you run a code coverage report? What does that look like?

jimmyfagan: Ah yea you're probably right, I wrote tests focused on the Acceptance Criteria, but probably should go back and check code coverage and do more true "Unit" tests for the other functions. Will work on that today, hopefully the rest of this is still okay to review.

arnejduranovic: I like the idea of having tests focused on acceptance criteria! I think adding tests for good code coverage is also very important as well, and it helps me better understand the code as well. I don't think code coverage needs be 100% in all cases but these methods are long and have a lot of logic in them that need to have tests.

jimmyfagan: Agreed, yea I think it was mostly laziness and eagerness to get this out for review, but going back and boosting code coverage actually exposed a few issues, so very much worthwhile!

Many of the helper functions were tested already in `FhirToHl7ConverterTests.kt`, so I pulled some of those out into a new file `ConfigSchemaProcessorTests.kt` and added new tests to cover both `getValue` and `getValueAsString`. I was missing some coverage of valueSet cases for `getValue` so I added tests and made some fixes to the function for that. I also was missing some coverage for adding an extension as part of the fhir transform, so I added tests for that, and then made some fixes to that chunk of code too. Code coverage is back to where I want it to be, although I still don't have really isolated unit tests for some functions in `FhirTransfomer`, so I may still go back and add a few there.

prime-router/src/main/kotlin/fhirengine/translation/hl7/FhirTransformer.kt
```diff
@@ -0,0 +1,225 @@
...
+        focusResources.forEachIndexed { index, singleFocusResource ->
+            // The element context must now get the focus resource
+            elementContext.focusResource = singleFocusResource
+            if (canEvaluate(element, bundle, singleFocusResource, elementContext)) {
+                when {
+                    // If this is a schema then process it.
+                    element.schemaRef != null -> {
+                        // Schema references can have new index references
+                        val indexContext = if (element.resourceIndex.isNullOrBlank()) elementContext
+                        else CustomContext.addConstant(
```
arnejduranovic, line 129: Should this chunk have code coverage?

arnejduranovic (approved):

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirBundleUtils.kt
```diff
@@ -0,0 +1,71 @@
...
+        "id",
+        "instant",
+        "markdown",
+        "oid",
+        "string",
+        "time",
+        "uri",
+        "url",
+        "uuid"
+    )
```
victor-chaparro, line 36: Would it be better to turn this into an enum?

prime-router/src/main/kotlin/fhirengine/translation/hl7/ConfigSchemaProcessor.kt
```diff
@@ -0,0 +1,136 @@
...
+    ): String {
+        var retVal = ""
+        run findValue@{
+            element.value.forEach {
+                val value = if (it.isBlank()) ""
+                else try {
+                    FhirPathUtils.evaluateString(context, focusResource, bundle, it)
+                } catch (e: SchemaException) {
+                    logger.error("Error while getting value for element ${element.name}", e)
+                    ""
```
victor-chaparro, line 31: Missing code coverage

prime-router/src/main/kotlin/fhirengine/translation/hl7/ConfigSchemaProcessor.kt
```diff
@@ -0,0 +1,136 @@
...
+    ): Base? {
+        var retVal: Base? = null
+        run findValue@{
+            element.value.forEach {
+                val value = if (it.isBlank()) emptyList<Base>()
+                else try {
+                    FhirPathUtils.evaluate(context, focusResource, bundle, it)
+                } catch (e: SchemaException) {
+                    logger.error("Error while getting value for element ${element.name}", e)
+                    emptyList<Base>()
```
victor-chaparro, line 69: Missing code coverage

prime-router/src/main/kotlin/fhirengine/translation/hl7/FhirTransformer.kt
```diff
@@ -0,0 +1,225 @@
...
+    constructor(
+        schema: String,
+        strict: Boolean = false,
+    ) : this(
+        schemaRef = fhirTransformSchemaFromFile(
+            FilenameUtils.getName(schema),
+            FilenameUtils.getPathNoEndSeparator(schema)
+        ),
+        strict = strict,
+    )
```
victor-chaparro, line 59: IntelliJ complains that these constructors are never used. Are they needed?

victor-chaparro (commented):
Added some comments. There's some code coverage missing.

prime-router/src/main/kotlin/fhirengine/translation/hl7/ConfigSchemaProcessor.kt
```diff
@@ -0,0 +1,136 @@
...
+     * @return the value for the element or an empty string if no value found
+     */
+    internal fun getValueAsString(
+        element: ConfigSchemaElement,
+        bundle: Bundle,
+        focusResource: Base,
+        context: CustomContext
+    ): String {
+        var retVal = ""
+        run findValue@{
```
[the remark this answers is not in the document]
jimmyfagan: This was copied from the previously existing implementation of `getValue`, so I don't know for sure, but I think this just allows for us to resolve this block early if we find a valid value when going through the `element.value.forEach` iterations. A `break` there would probably achieve the same thing, but I think it's a little more readable by having a very explicit `return@findValue` instead of the somewhat ambiguous `break`.

prime-router/src/main/kotlin/fhirengine/translation/hl7/FhirTransformer.kt
jimmyfagan: I suspect at least some of them will be eventually when this is integrated more into the pipeline, for now this is still only used in tests, so the constructors aren't as needed. After that integration point in a future story, I'll remove any that are still unused.

jimmyfagan:
Responded to a few of the comments, still need to go back and check the missing code coverage, I'll either add more tests or respond to those comments to say why I skipped code coverage of those. I can't remember off the top of my head how intentional some of those skips were, it might have just been aiming for parity with the existing tests for the Hl7toFhirConverter, but that's not a great reason to tolerate lower code coverage if it would be easy enough to cover. Will aim to respond later tonight or first thing tomorrow morning, thanks for the reviews so far!

prime-router/src/main/kotlin/fhirengine/translation/hl7/ConfigSchemaProcessor.kt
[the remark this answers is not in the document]
JessicaWNava: That's a cool Kotlin feature I didn't know about, thanks for the explanation!

prime-router/src/main/kotlin/fhirengine/translation/hl7/FhirTransformer.kt
```diff
@@ -0,0 +1,225 @@
...
+     * [focusResource] in the bundle. Set [debug] to true to enable debug statements to the logs.
+     */
+    private fun transformWithSchema(
+        schema: FhirTransformSchema,
+        bundle: Bundle,
+        focusResource: Base,
+        context: CustomContext = CustomContext(bundle, focusResource),
+        debug: Boolean = false
+    ) {
+        val logLevel = if (debug) Level.INFO else Level.DEBUG
```
JessicaWNava, line 85: Does this need to be flipped where it is Level.DEBUG if debug is true?

prime-router/src/main/kotlin/fhirengine/translation/hl7/FhirTransformer.kt
```diff
@@ -0,0 +1,225 @@
...
+     * [focusResource] in the bundle. Set [debug] to true to enable debug statements to the logs.
+     */
+    internal fun transformBasedOnElement(
+        element: FHIRTransformSchemaElement,
+        bundle: Bundle,
+        focusResource: Base,
+        context: CustomContext,
+        debug: Boolean = false
+    ) {
+        val logLevel = if (element.debug || debug) Level.INFO else Level.DEBUG
```
JessicaWNava, line 107: Same here, should the log levels be flipped based on the if statement?

JessicaWNava (commented):
A couple of minor comments and agree with the code coverage comments Victor made since we are fully relying in tests here. Thank you for all the testing you did add though, that was a lot!

prime-router/src/main/kotlin/fhirengine/translation/hl7/ConfigSchemaProcessor.kt
jimmyfagan: Didn't realize until looking into this more, but the `evaluate`, unlike `evaluateString`, doesn't throw a schema exception that I can find, so I'm just removing the try/catch here.

prime-router/src/main/kotlin/fhirengine/translation/hl7/ConfigSchemaProcessor.kt
jimmyfagan: Will be adding a test to cover this in my next commit.

prime-router/src/main/kotlin/fhirengine/translation/hl7/FhirTransformer.kt
jimmyfagan: It should eventually, but that'll be more relevant to [the story for repeating elements](https://app.zenhub.com/workspaces/platform-6182b02547c1130010f459db/issues/gh/cdcgov/prime-reportstream/7248). Writing tests for that and verifying that things work is probably the main work for that story, so I'll save it for then, but I still adapted this piece of the code over as a starting point.

prime-router/src/main/kotlin/fhirengine/translation/hl7/FhirTransformer.kt
jimmyfagan: Good question, what this is actually doing is setting the level for the log messages, so if `debug` is true, then we're logging things at the informational level so they will be more visible. If `debug` is false, then we log things at the debug level so they'll be less visible. I'm not super familiar with our logger, but that's my understanding of it, and this was following the lead from existing code in `FhirToHl7Converter.kt`.

JessicaWNava (approved):
Looks good, thank you for addressing those comments/making those changes!

victor-chaparro (approved):
Looks good and thanks for all the explanations!

PR #8465, opened 2023-02-23
Update to React v18

### LINKED ISSUE #8338: Upgrade to React v18
With all other major dependency upgrades done, we need to finally update our react to v18. HOWEVER, we cannot immediately use strict mode! We'll continue to use the legacy renderer and in a later ticket make any project changes needed (ex: fix our usages of useEffect) so that we can fully transition to react's new renderer. @stephenkao

jpandersen87:
Fixes #8338 

This PR:
- updates react to v18
- Uses react v18 rendering mode aka concurrent (non-strict)
- updates related react testing libraries to their latest versions
- fixes bugs with tests due to these changes
- unifies the custom test render functions to one
- fixes tests so that they await on signals that indicate re-rendering is complete for userEvents (fixes act() warnings).
  - warnings on state updates after unmount were disabled in react v18

The unified custom render function accepts an options bag that can include restHookFixtures and initialRouteEntries. 
These options allow us to stop mocking the rest-hooks and react-router-dom modules in most cases (partial module mocks for react-router-dom may still be needed for hooks like useNavigation). 

When testing a component that relies on react-router-dom hooks such as useLocation, useParams, or useSearchParams we can provide initialRouteEntries to the render function so that these hooks work naturally without mocking. 

For remaining rest-hook tests we use restHookFixtures to use rest-hooks' built-in mocking functionality. Rest-hooks will conveniently emit a console warning when a rest-hook data fetch occurs that was not matched to a fixture with details to help debugging. 

These changes should resolve existing errors that currently occur in tests (but do not cause a failure) due to erroneous mocks.


frontend-react/package.json
```diff
@@ -188,10 +188,6 @@
   "peerDependencies": {
     "minimist": "^1.2.6"
   },
-  "resolutions": {
```
stephenkao, line 191: 🔥

frontend-react/src/utils/CustomRenderUtils.tsx
```diff
@@ -1,174 +1,92 @@
...
-    ({ children }: PropsWithChildren<{}>) =>
-        (
-            <BaseWrapper>
-                <QueryClientProvider client={client}>
-                    <AuthorizedFetchProvider initializedOverride={true}>
-                        {children}
-                    </AuthorizedFetchProvider>
-                </QueryClientProvider>
-            </BaseWrapper>
+export const AppWrapper = ({
```
stephenkao, line 39: THANK YOU for consolidating all of these 🙏

frontend-react/src/utils/CustomRenderUtils.tsx
```diff
@@ -1,174 +1,92 @@
...
+    initialRouteEntries,
+    restHookFixtures,
+}: AppWrapperOptions = {}) => {
+    // FUTURE_TODO: Replace children with <AppRouter /> if initialRouteEntries after mocking okta users
+    // in tests is made easier for better coverage as we'd be able to test through
+    // any custom route wrappers.
+    // FUTURE_TODO: Remove MockResolver and restHookFixtures when removing react-hooks.
+    return ({ children }: PropsWithChildren<{}>) => {
```
stephenkao, line 47: Minor point, and I know this existed beforehand, but I think we can just define `children` explicitly since it's clearer with props and `PropsWithChildren` doesn't give us anything more than that.

frontend-react/src/utils/CustomRenderUtils.tsx
```diff
@@ -1,174 +1,92 @@
...
-                </QueryClientProvider>
-            </BaseWrapper>
+export const AppWrapper = ({
+    initialRouteEntries,
+    restHookFixtures,
+}: AppWrapperOptions = {}) => {
+    // FUTURE_TODO: Replace children with <AppRouter /> if initialRouteEntries after mocking okta users
+    // in tests is made easier for better coverage as we'd be able to test through
+    // any custom route wrappers.
+    // FUTURE_TODO: Remove MockResolver and restHookFixtures when removing react-hooks.
```
stephenkao, line 46: 🙏

frontend-react/src/components/AuthElement.test.tsx
```diff
@@ -5,15 +5,19 @@ import { mockSessionContext } from "../contexts/__mocks__/SessionContext";
...
+import { renderApp } from "../utils/CustomRenderUtils";
 
 import { AuthElement } from "./AuthElement";
 
 const mockUseNavigate = jest.fn();
 
-jest.mock("react-router", () => ({
-    useNavigate: () => mockUseNavigate,
-}));
+jest.mock("react-router", () => {
```
stephenkao, line 14: Smallest point in the world, but I think we could simplify this even further, no?

```tsx
jest.mock("react-router", () => ({
    ...jest.requireActual("react-router"),
    useNavigate: () => mockUseNavigate
}));
```

frontend-react/src/components/Table/Pagination.test.tsx
```diff
@@ -63,12 +58,12 @@ describe("Pagination", () => {
...
                 setSelectedPage: jest.fn(),
             };
-            renderWithBase(<Pagination {...props} />);
+            renderApp(<Pagination {...props} />);
 
             const list = screen.getByRole("list");
-            const { getAllByRole } = within(list);
-            const items = getAllByRole(ITEM_ROLE);
-            const itemContents = items.map((item) => item.textContent);
+            const itemContents = Array.from(list.children).map(
```
stephenkao, line 64: Ah, I don't think this is exactly equivalent of using `within` -- `children` would be the direct descendants whereas querying with `within` would check for all descendants, no?

Not a huge deal -- it's completely fine if it works for this test, but I'd still err on the side of using `within` in case the we change the markup to have an extra layer between the `list` ancestor and the list items 🤷

frontend-react/src/utils/CustomRenderUtils.tsx
```diff
@@ -1,174 +1,92 @@
...
+
+interface TestRouterProps {
+    children: React.ReactNode;
+    initialEntries?: string[];
+}
+
+/**
+ * Dynamically makes the supplied children the return element for all
+ * routes.
+ * FUTURE_TODO: Remove this once okta user/session mocking is easier
```
stephenkao, line 28: 🙏 Would you mind adding a ticket for this if you haven't already?

jpandersen87:
Not exactly sure why the latest frontend build failed. The jest output shows source code that is NOT the current code.

frontend-react/src/components/Admin/AdminReceiverDashboard.test.tsx
```diff
@@ -1,11 +1,15 @@
...
+import {
+    AdmConnStatusDataType,
+    AdmConnStatusResource,
+} from "../../resources/AdmConnStatusResource";
 import { ErrorPage } from "../../pages/error/ErrorPage";
-import { renderWithBase } from "../../utils/CustomRenderUtils";
+import { renderApp } from "../../utils/CustomRenderUtils";
+import Spinner from "../Spinner";
 
 import { _exportForTesting } from "./AdminReceiverDashboard";
```
stephenkao, line 14: Minor point and COMPLETELY out of the scope of this pull request because it already existed beforehand, but I don't think we should use this pattern going forward; if things need to be exported, they should just be exported 🤷 We should also break large files like this up into smaller files in the future.

frontend-react/src/components/Admin/AdminReceiverDashboard.test.tsx
```diff
@@ -81,19 +85,21 @@ const mockData: AdmConnStatusDataType[] = [
 ];
 // </editor-fold>
```
stephenkao, line 86: We can remove the `editor-fold` stuff since there's no reason to change code for specific IDEs.  Not a huge deal, we can do that cleanup anytime!

frontend-react/src/components/Admin/EditSenderSettings.test.tsx
```diff
@@ -70,53 +70,95 @@ describe("EditSenderSettings", () => {
...
+        await userEvent.type(
+            keysField,
+            testKeys.replaceAll("[", "[[").replaceAll("{", "{{")
+        );
 
         expect(keysField).toHaveValue(testKeys);
     });
 
-    test("should be able to edit processing type field", () => {
+    // DO NOT USE THIS TEST AS AN EXAMPLE. Checking the classList
```
stephenkao, line 116: 🙏

frontend-react/src/components/Table/Pagination.test.tsx
```diff
@@ -9,17 +9,12 @@ import Pagination, {
...
-    // presentation).
-    const ITEM_ROLE = /listitem|presentation/i;
-
     test.each([
         {
             description: "on the first page of an unbounded set",
             slots: [1, 2, 3, 4, 5, 6, OVERFLOW_INDICATOR],
             currentPageNum: 1,
-            expectedItems: ["1", "2", "3", "4", "5", "6", "…", "Next"],
+            expectedItems: ["1", "2", "3", "4", "5", "6", "Next", "…"],
```
stephenkao, line 17: Huhhhhh...interesting that the wrong order didn't cause any issues with the textContent comparison below 🤔

frontend-react/src/components/Table/Table.test.tsx
```diff
@@ -91,19 +86,21 @@ const selectDatesFromRange = async (dayOne: string, dayTwo: string) => {
     const endDatePickerButton = datePickerButtons[1];
 
     /* Select Start Date */
-    await userEvent.click(startDatePickerButton);
-    const newStartDateButton = screen.getByText(`${dayOne}`);
-    await userEvent.click(newStartDateButton);
+    userEvent.click(startDatePickerButton);
```
stephenkao, line 89: Why were the `await`s removed for the `userEvent` interactions?  Aren't they all asynchronous now?

stephenkao:
@jpandersen87 This looks good to me overall, but it's a huge pull request that would benefit from being split up into smaller changesets -- at least for the ease of reviewing from my perspective.  In my opinion, it could be broken down into multiple parts with the core React v18 dependency upgrade as a root branch to work off of:

- Bump to React v18: just the dependency upgrades of react + react-router-dom, initial setup (should be small)
    - Consolidate test renders into `renderApp`: this might actually be able to be done independently of the React v18 bump
    - Essential test fixes (warnings/errors that explicitly fail tests, could be split up by types or domains of warnings/errors)
    - Non-essential test fixes (warnings/errors that don't explicitly fail tests like invalid DOM descendants, could be split up by types or domains of warnings/errors)
    - Test rewrites (changing fundamental logic of tests to get them to pass or make them more robust): splitting this out into a separate changeset helps to call attention to which changes are caused by faulty logic in individual tests versus the test utils

I don't think it's absolutely necessary for all tests to pass _until we're merging the root branch into `master`_; if there are still failing tests for any particular changeset along the way, we can just acknowledge the failures as part of that changeset and fix them in a subsequent one (likely particularly pertinent for new rest-hooks mocks).

I'm happy to chat about this more, but my feeling is that we should be striving for small, incremental changes for faster, more focused reviews during which we can discuss more decisions and possibly catch more bugs -- even if that means maintaining a long-lived root branch off of `master` for a little while.  Let me know what you think!

frontend-react/src/components/Table/Pagination.test.tsx
jpandersen87: That's due to the items array being [...listitems, presentation (if ...)]. the regex that was previously used doesn't seem to be allowed anymore for the query so I had to grab the two role types manually and adjust expectedItems to fit the new order

frontend-react/src/components/Table/Table.test.tsx
jpandersen87: I have to await on the following findBy so that side effects are settled

jpandersen87:
Closing to restart as a layered process

PR #8485, opened 2023-02-24
Experience-8433: Add initial best practices doc

stephenkao:
This changeset is just introducing an initial "best practices" doc for working within the frontend-react side of things.  I'm hoping our team can keep adding to it as we encounter and build out new patterns.

There are still lots and lots of things to write about, and I'll split those up into separate tickets:
- More styleguide conventions
- Using react-query: our approach to resources/collections, our treatment of the query cache (and mocking it in the test conventions doc)
- React patterns and optimizations (hooks versus HOCs, references)
- A billion other things

### DIFF
--- a/frontend-react/docs/best-practices.md
+++ b/frontend-react/docs/best-practices.md
@@ -398,15 +398,15 @@ function UserList({ users, sortProperty }: { users: User[], sortProperty: "age" 
 ```  
 
 
-### Strive for single-purpose ("dumb") components and hooks
+### Strive for single-purpose ("dumb" or purely presentational) components and hooks
 
 > Why?
 >
 > This keeps components and hooks much more flexible in usage, which lends to the React ethos of composition.  Additionally, this breaks complex logic down for easier maintenance; ensuring that each component is focused around one purpose helps break up long logic chains for easier testability, debuggability, and upkeep, and it (usually) also helps keep the components and hooks smaller so they're more likely to be reusable across the codebase.
 
-The concept of "dumb" components has been around in the React ecosystem for a while: components whose sole purpose is to present and display data.  And although this concept was much easier to abide by with the [container/presenter pattern](https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0) before the introduction of hooks, keeping this clear separation of concerns is still important for scalability.
+The concept of "dumb"/purely presentational components has been around in the React ecosystem for a while: components whose sole purpose is to present and display data.  And although this concept was much easier to abide by with the [container/presenter pattern](https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0) before the introduction of hooks, keeping this clear separation of concerns is still important for scalability.
 
-Put simply, **dumb components are presentational whereas smart components are concerned with maintaining events and state.**  When writing a new component, it's important to question whether all the logic should exist in one place or if it would be more maintainable (and possibly readable) split out between different components or utility functions.
+Put simply, **dumb components are presentational whereas container/smart components are concerned with maintaining events and state.**  When writing a new component, it's important to question whether all the logic should exist in one place or if it would be more maintainable (and possibly readable) split out between different components or utility functions.
 
 Take the following example:
 
--- a/frontend-react/docs/best-practices.md
+++ b/frontend-react/docs/best-practices.md
@@ -46,6 +46,8 @@ Branch names should follow the format:
 experience/XXXX/some-short-description-of-the-feature  
 ```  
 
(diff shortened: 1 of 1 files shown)


penny-lischer (approved):
LGTM

frontend-react/docs/best-practices.md
```diff
@@ -0,0 +1,677 @@
...
+---
+
+## Processes
+
+### Branch names
+
+Branch names should follow the format:
+
+```  
+experience/XXXX/some-short-description-of-the-feature
```
etanb, line 46: Should just leave a little comment here what `XXXX` means (the Zenhub ticket number).

frontend-react/docs/best-practices.md
```diff
@@ -0,0 +1,677 @@
...
+# description  
+This changeset includes a new feature that allows users to sort tables by multiple columns.   
+Also included in this changeset:  
+- A few backfilled tests  
+- - I listed this item  
+  
+```  
+
+> Why?
+>
```
etanb, line 72: Any opinion on _size_ or _contents_ of commits? Only commit a fully complete feature vs. extremely piecemeal functions etc.

frontend-react/docs/best-practices.md
```diff
@@ -0,0 +1,677 @@
...
+
+The concept of "dumb" components has been around in the React ecosystem for a while: components whose sole purpose is to present and display data.  And although this concept was much easier to abide by with the [container/presenter pattern](https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0) before the introduction of hooks, keeping this clear separation of concerns is still important for scalability.
+
+Put simply, **dumb components are presentational whereas smart components are concerned with maintaining events and state.**  When writing a new component, it's important to question whether all the logic should exist in one place or if it would be more maintainable (and possibly readable) split out between different components or utility functions.
```
etanb, line 387: might be worth replacing smart / dumb with something like stateful / stateless components or some other more apt descriptor?

frontend-react/docs/best-practices.md
```diff
@@ -0,0 +1,677 @@
...
+    background-color: pink;
+}
+```
+
+> Why?
+>
+> Low-specificity selectors in CSS rules makes them more easily overridden if a specific visual variant or ad hoc style needs to be applied.
+
+---
+
```
etanb, line 672: A small section on `_uswds_overrides.scss` and best practices (classes + element selectors of equal specificty > `!important`) and that, at least for now, component-specific CSS should live in: `_components.scss` and wrapped in a css class describing the component. I was doing:

```
// MessageReceivers.tsx
.message-receivers {
  .big-font {
    font-size: 36px;
  }
}

// End -- MessageReceivers.tsx
```

To indicate which component the CSS applied to in my previous work, but open to suggestions!

frontend-react/docs/best-practices.md
stephenkao: Sure, can do!

frontend-react/docs/best-practices.md
stephenkao: I think it really depends on the logic at hand to be honest.  Sometimes a big changeset is inevitable, especially for essential dependencies and such, so that nothing breaks.  BUT I can add a point about trying to break pull requests up into as many small, atomic pieces as possible -- occasionally branching off a central root branch to be merged back into `master`.

frontend-react/docs/best-practices.md
stephenkao: Hahah "smart" and "dumb" are pretty frequently used terms for components in front-end frameworks!  I've also heard "container" and "presentational" components, which could help to clarify.  Happy to change to that if preferred.

But "stateful" and "stateless" would be misnomers because that'd imply that no presentational component has state -- that's not true since there could be _component_ state through `useState`.  We're really talking about global or mediator state, but I'd argue that this'd still be confusing.

frontend-react/docs/best-practices.md
stephenkao: Sure, definitely can do!

stephenkao:
@etanb Updated with your suggestions [here](https://github.com/CDCgov/prime-reportstream/pull/8485/commits/5a6752117f075f579c0cbf7ae5722d12cacf24c0).  Let me know what you think whenever you get a chance!

frontend-react/docs/best-practices.md
etanb: That's totally fair, I was just trying to set up some meaningful dichotomy. Container and presentational sounds good to me!

etanb (approved):

PR #8537, opened 2023-02-28
Update jest from 27.5.1 to 29.4.3

### LINKED ISSUE #8460: Update jest from 27.5.1 to 29.4.3
Update jest from 27.5.1 to 29.4.3 and jest-watch-typeahead from 1.1.0 to 2.2.0. Determine if jest-resolve needs to be explicitly defined as a dependency. If so, also update to 29.4.3. If not, remove it.

This update will likely break our testing so project changes will need to be made to fix the breakages.

Migration documentation:
- 27 -> 28: https://jestjs.io/docs/28.x/upgrading-to-jest28
- 28 -> 29: https://jestjs.io/docs/upgrading-to-jest29

jpandersen87:
Fixes #8460

This PR updates jest from 27.6.1 to 29.4.3. Jest 28 updated their package.json exports support so that environment globals are fully supported. This means dependencies that resolved before potentially won't anymore and vice-versa depending on their exports when jest attempts to automatically find an export based off of "browser" and "require" globals (with a bad path giving us an ES module thus breaking). The following dependences were affected:
- axios - previously we had to manually map it, but it appears to work naturally now
- rest-hooks and @rest-hooks/* - requires manual map (this is potentially not an issue in their current versions but we're in the process of deprecating this dependency so its version is intentionally out-of-date)
- msw/lib/node - requires manual map

Ejected scripting from CRA needed updates to support the newer jest. Two test modules were updated to fix test bugs causing a false failure.

Some minor update bumps were made to:
- cypress
- eslint
- @typescript-eslint/*
- jest-watch-typeahead
- msw

Some peer dependencies were added to cease warnings from yarn.


jpandersen87:
Going to need a sanity check on the error GH is getting. Context: after the dependency updates a test for DeliveriesTable was failing because the output from `new Date("01-01-2000").toISOString()` changed from "2000-01-01T00:00:00.000Z" to "2000-01-01T**06:00:00.000**Z". I was able to verify this output in a browser console. My assumption is that the jest updates made expected changes to how the browser environment is simulated and date behavior was modified. However, GH is throwing an error because it is still getting the previous output somehow. I have flushed my node_modules to make sure this wasn't a local package manager bug but the test still passes. Can someone check this out locally and confirm?

frontend-react/src/pages/deliveries/Table/DeliveriesTable.test.tsx
```diff
@@ -160,15 +161,16 @@ describe("DeliveriesTableWithNumbered", () => {
...
+                test("Clicking on filter invokes the trackAppInsightEvent", async () => {
+                    await userEvent.click(screen.getByText("Filter"));
 
                     expect(mockAppInsights.trackEvent).toBeCalledWith({
                         name: "Daily Data | Table Filter",
                         properties: {
                             tableFilter: {
                                 endRange: "3000-01-01T23:59:59.999Z",
-                                startRange: "2000-01-01T00:00:00.000Z",
+                                // xx/xx/xxxx returns a 06:00:00 vs xxxx-xx-xx's 00:00:00??
```
penny-lischer, line 172: Was this comment intentional?

jpandersen87: yes, this is confusing behavior

penny-lischer: Agree - Maybee we need a ticket to investigate and fix this. @stephenkao ?

jpandersen87: Discrepancy was due to local timezones, this has been fixed in another PR.

penny-lischer (commented):
LGTM

penny-lischer (approved):

frontend-react/jest.config.ts
```diff
@@ -24,7 +24,13 @@ const config: Config.InitialOptions = {
     moduleNameMapper: {
         "^react-native$": "react-native-web",
         "^.+\\.module\\.(css|sass|scss)$": "identity-obj-proxy",
-        axios: "axios/dist/node/axios.cjs",
```
stephenkao, line 27: 🙏

frontend-react/src/components/FileHandlers/FileHandler.test.tsx
```diff
@@ -274,40 +261,47 @@ describe("FileHandler", () => {
...
+                userEvent.upload(
                     screen.getByTestId("file-input-input"),
-                    fileChangeEvent
+                    fakeFile
                 );
-
                 await screen.findByTestId("file-input-preview-image");
 
-                fireEvent.click(screen.getByText("Validate"));
+                // jsdom seems currently unable to properly handle required file input
```
stephenkao, line 270: Ugh, makes sense but unfortunate.  As an aside, I think it'd be worthwhile to disable native validations in favor of our own form validation logic anyway -- especially for more complex validators -- but thoughts for the future.

PR #8547, opened 2023-03-01
MessageReceivers sortData bug: Fixes #8546

### LINKED ISSUE #8546: BUG: MessageReceivers Component - sortedData uses wrong property name for accessing column data
## Describe the bug
The algorithm for sortedData in the MessageReceivers component is incorrectly accessing A and B column data by its column label string rather than its internal name string (ex: "Report Id" instead of "ReportId"). This causes the algorithm to incorrectly compare both A and B as undefined, causing an incorrect sort.

## Impact on ReportStream
MessageReceivers table is incorrectly sorted upon selecting a column header.

## Steps to reproduce
1. Set a breakpoint inside the sortData function
2. Go to the Message Tracker page
3. Search for a message
4. Click the Report Id column header
5. Once the breakpoint is hit, see value of activeColumn as "Report Id"

OR

1. Set a breakpoint inside the sortData function
2. Run yarn run test with debugging
5. Once the breakpoint is hit, see value of activeColumn as "Report Id"

## Expected behavior
The sortedData function should be accessing Report Id by the "ReportId" property name.

## Suggestions
In addition to the columnHeaderTitle prop, a columnHeader prop consisting of its internal name string should also be passed to the MessageReceiversCol component. The handleColHeaderClick function should call setActiveColumn with this string rather than columnHeaderTitle.

jpandersen87:
Fixes #8546

This PR fixes a bug in sortData in the MessageReceivers component.


frontend-react/src/components/MessageTracker/MessageReceivers.tsx
```diff
@@ -60,7 +61,8 @@ const ColumnDataEnum = {
     Sub: "Sub",
     FileName: "File Name",
     TransportResults: "Transport Results",
-};
+} as const satisfies { [k in keyof NormalizedReceiverData]: string };
+type ColumnDataTitle = (typeof ColumnDataEnum)[keyof typeof ColumnDataEnum];
```
stephenkao, line 65: Hmmm I know this isn't a huge deal because a lot of things are already enumerated, but I'm wary of doing comparisons based on presentational strings like the column titles rather than using a centralized key.  

Given the interface:

```tsx
interface NormalizedReceiverData {
    Name: string;
    Service: string;
    Date: string;
    ReportId: string;
    Main: "Batch" | "Process" | "Ready";
    Sub: string;
    FileName: string;
    TransportResults: string;
}
```
(Side note: I'd still argue these should be starting with lowercase letters since they're not enums themselves but rather descriptors of object properties: `name`, `service`, `date`, `fileName` ...)

Our keys here are `Name`, `Service`, `Date`, `FileName`, et cetera.  To that end, if we're going to use these as keys, that'll make things easier because we can define titles separately:

```tsx
// renamed from ColumnDataEnum for clarity
const ColumnDataTitles = {
    name: "Name",
    service: "Service",
    date: "Date",
    reportId: "Report Id",
    main: "Main",
    sub: "Sub",
    fileName: "File Name",
    transportResults: "Transport Results",
} as const satisfies { [k in keyof NormalizedReceiverData]: string };

// will be one of "Name" | "Service" | "Date" | "FileName" | ...
type NormalizedReceiverDataKey = keyof typeof ColumnDataTitles;
```

Then we can pass these keys around as a prop:

```tsx
interface MessageReceiversColProps {
    columnKey: NormalizedReceiverDataKey;
    columnHeaderTitle: string;
    activeColumnKey?: NormalizedReceiverDataKey;
    setActiveColumnKey?: (columnKey: NormalizedReceiverDataKey) => void;
    ...
}
```

And this'd allow us to set/compare against verbatim strings because `NormalizedReceiverDataKey` would only allow for values in the union of `Name` | `Service` | `Date` | `FileName`.  So ` columnKey === "Name"` would be okay whereas `columnKey === "barf"` would not be.

The point of this would be to use the keys as the centralized identifier of columns rather than titles -- the latter of which are more prone to change.  I realize this would be a bigger change than is needed for the bugfix so I'm okay with breaking that out into a separate changeset if needed.

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
```diff
@@ -329,32 +334,20 @@ export const MessageReceivers = ({ receiverDetails }: MessageReceiverProps) => {
         // which requires a return of -1 0 1. For dates, we need to
         // convert them to a numerical value.
         () =>
-            activeColumnSortOrder !== FilterOptionsEnum.NONE
+            activeColumnSortOrder !== FilterOptionsEnum.NONE && activeColumn
```
stephenkao, line 337: Out of the scope of this pull request, but I think it'd be nice to pull out the normalization logic and the sorting logic out of the render cycle and into more unit-testable functions.  I think that'd make it easier to write a regression test for this bug too 🤔

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
jpandersen87: I think my changes align with what you want here, it's just that i'm using the verbose keyof typeof ColumnDataTitles in all of the relevant areas instead of aliasing it to NormalizedReceiverDataKey (which should def be done for easier code reading).

jpandersen87: Will implement suggestions on lowercase keys, etc.

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
jpandersen87: too bad, you get it meow

frontend-react/src/components/MessageTracker/MessageReceivers.test.tsx
```diff
@@ -1,36 +1,119 @@
 import { screen } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 
+import { ReceiverData } from "../../config/endpoints/messageTracker";
 import { renderWithBase } from "../../utils/CustomRenderUtils";
 
-import { MessageReceivers } from "./MessageReceivers";
+import {
```
etanb, line 7: Thank you for these tests! 🙏

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
```diff
@@ -393,67 +394,71 @@ export const MessageReceivers = ({ receiverDetails }: MessageReceiverProps) => {
...
                                 {...commonProps}
-                                columnHeaderTitle={ColumnDataEnum.Sub}
+                                columnHeader={"sub"}
+                                columnHeaderTitle={ColumnDataTitles.sub}
                                 rowSpan={1}
                             />
                             <MessageReceiversCol
                                 {...commonProps}
-                                columnHeaderTitle={ColumnDataEnum.FileName}
+                                columnHeader={"fileName"}
```
etanb, line 446: If we're going to explicitly use a new property to define the ID for the column, I think we should be explicit with the property name so something like `columnHeaderID` might be helpful and more semantic. 

However, I'm not a huge fan of just passing the raw ID string in since it won't be conceptually "linked" to the actual title. Maybe even something more robust like:

```
const ColumnData = {
    name: { id: "name", title: "Name" },
    service: { id: "service", title: "Service" },
    date: { id: "date", title: "Date" },
    reportId: { id: "reportId", title: "Report Id" },
   // etc...
};
```

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
stephenkao: I guess my point is more pertinent to where we're still running comparisons against the titles rather than the normalized field names instead:

- Determining class name [here](https://github.com/CDCgov/prime-reportstream/pull/8547/files#diff-9b64e4e57cfe7ca23254b7aa8cb8c29482b4f902329f5083e96c5a6be5a4657eR140)
- Assigning values [here](https://github.com/CDCgov/prime-reportstream/pull/8547/files#diff-9b64e4e57cfe7ca23254b7aa8cb8c29482b4f902329f5083e96c5a6be5a4657eR281)

Also, I know that this logic existed beforehand, but I think `formatMessages` could be simplified since we have an unnecessary iteration of `ColumnDataTitles` each invocation:

```ts
export function formatMessages(data: ReceiverData[]): NormalizedReceiverData[] {
    return data.map((receiverItem: ReceiverData): NormalizedReceiverData => {
        const fieldNames = Object.keys(ColumnDataTitles) as NormalizedReceiverKey[];

        const formattedData = fieldNames.reduce(
            (
                acc: Partial<NormalizedReceiverData>,
                fieldName: NormalizedReceiverKey
            ) => {
                let value: string | null;

                switch (fieldName) {
                    case "name":
                        value = receiverItem.receivingOrg;
                        break;
                    case "service":
                        value = receiverItem.receivingOrgSvc;
                        break;
                    case "date":
                        value = dateTimeFormatter.format(
                            new Date(receiverItem.createdAt)
                        );
                        break;
                    case "reportId":
                        value = receiverItem.reportId;
                        break;
                    case "main":
                        const { folderLocation } = parseFileLocation(
                            receiverItem.fileUrl || ""
                        );
                        value = folderLocation;
                        break;
                    case "sub":
                        const { sendingOrg } = parseFileLocation(
                            receiverItem.fileUrl || ""
                        );
                        value = sendingOrg;
                        break;
                    case "fileName":
                        const { fileName } = parseFileLocation(
                            receiverItem.fileUrl || ""
                        );
                        value = fileName;
                        break;
                    case "transportResults":
                        value = receiverItem.transportResult;
                        break;
                }

                return {
                    ...acc,
                    [fieldName]: value || "N/A",
                };
            },
            {}
        );

        return formattedData as NormalizedReceiverData;
    });
}
```

I don't think we need that initial iteration just to prepopulate all fields with "N/A" 🤔

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
stephenkao: Hmmm ah, this might be out of the scope of this pull request since it brings up that fun conversation about a standard structure for our eventual generic Table component.  For this changeset in particular, the [NormalizedReceiverKey](https://github.com/CDCgov/prime-reportstream/pull/8547/files#diff-9b64e4e57cfe7ca23254b7aa8cb8c29482b4f902329f5083e96c5a6be5a4657eR69) is now the thing that links the title to the key; the property name like `name`, `service`, `date` is used to identify the string title in `ColumnDataTitles`, and `columnHeader` can't be anything outside of this set of strings.

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
```diff
@@ -393,67 +394,71 @@ export const MessageReceivers = ({ receiverDetails }: MessageReceiverProps) => {
                         <tr>
                             <MessageReceiversCol
                                 {...commonProps}
-                                columnHeaderTitle={ColumnDataEnum.Name}
+                                columnHeader={"name"}
```
stephenkao, line 397: Might be good to call this prop `columnKey` or something for clarity since it'll be the identifier for any column, whether we're in the context of the headers or an individual data cell.

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
etanb: Yeah, I just want to make sure they're tightly coupled in someway.

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
jpandersen87: should we override the default key property to just be required and use that?

jpandersen87: followup: avoiding the key prop, will use columnKey

stephenkao: For context, my feeling was that we shouldn't overload the `key` prop since that's specifically for React rendering stability/optimizations.  Anything more than that would be outside the purview of this prop (i.e., plucking cell data) so we decided to keep it more explicit with `columnKey`.

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
stephenkao: I'll add a ticket for this!

stephenkao (approved):
Looks good to me!

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
stephenkao: https://app.zenhub.com/workspaces/web-experience-607d9d5e68b95200150fec37/issues/gh/cdcgov/prime-reportstream/8621

PR #8550, opened 2023-03-01
Experience-8549: Fix incorrect setting type history

### LINKED ISSUE #8549: Bug:  Those who do not fix history are doomed to repeat it.
## Describe the bug
The "History" feature for Senders and Receivers does not work

See screenshot.  Note that there are three "History" links.   One for the Organization, one for the Senders, and one for the Receivers.    But all three of those links seem to erroneously point to the Organization history.  

## Impact on ReportStream
Can't get a history of Sender settings changes.  Can't get a history of Receiver settings changes.

## Steps to reproduce
1. Go to the a setting page, and click on the three different History links.

## Expected behavior
- Organizations history shows Organization history (no changes - this already works)
- Sender history shows that Sender's history.  
- Receiver history shows that Receiver's history.

stephenkao:
This changeset updates the setting revision query key to be cached on the Organization and setting type (Sender|Receiver).  Originally, it was being cached on the strings "org" and "settingType" verbatim, which means that all revision histories would get the same response that's in the query cache.  This just updates the query key to be more dynamic.

Test Steps:
1. Log in as an admin
2. Choose an Organization with multiple Sender/Receiver channels to edit
3. Click on the different "History" links (using the back button/without refreshing the page) and verify that only the correct revision histories are shown for the specific setting type (Organization|Sender|Receiver)

### Testing
- [x] Tested locally?
- [ ] Added tests? -- not sure if it's really needed since this is just updating a query key 🤷 

## Linked Issues
- Fixes #8549

### DIFF
--- a/frontend-react/src/network/api/Organizations/SettingRevisions.ts
+++ b/frontend-react/src/network/api/Organizations/SettingRevisions.ts
@@ -39,7 +39,7 @@ export const useSettingRevisionEndpointsQuery = (
         useAuthorizedFetch<SettingRevision[]>();
 
     // get all lookup tables in order to get metadata
-    return rsUseQuery(["org", "settingType"], async () =>
+    return rsUseQuery(["history", params.org, params.settingType], async () =>
         authorizedFetch(settingRevisionEndpoints.getList, {
             segments: params,
         })


jpandersen87 (approved):
yes, use the dynamic query keys luke

etanb (approved):
Tested locally, works as intended!

I do wonder if we should explicitly be passing which piece of history we're looking for and pass that to `useSettingRevisionEndpointsQuery` instead of inferring it from the query params?

stephenkao:
> Tested locally, works as intended!
> 
> I do wonder if we should explicitly be passing which piece of history we're looking for and pass that to `useSettingRevisionEndpointsQuery` instead of inferring it from the query params?

That's already happening [here](https://github.com/CDCgov/prime-reportstream/pull/8550/files#diff-40e2e9542862a5c6d89548617b8c2ff99ecda9a29ea78468e727feac0b207910R29), and they're getting passed as props from path fragments (rather than query parameters) [here](https://github.com/CDCgov/prime-reportstream/blob/master/frontend-react/src/pages/admin/AdminRevHistory.tsx#L197-L198).  The underlying issue is that we were just telling the query cache to cache on the very, very general and literal query key of `["org", "settingType"]` so it was returning the cached response rather than making a new call specific to the Organization/setting type.

etanb:
> That's already happening [here](https://github.com/CDCgov/prime-reportstream/pull/8550/files#diff-40e2e9542862a5c6d89548617b8c2ff99ecda9a29ea78468e727feac0b207910R29), and they're getting passed as props from path fragments (rather than query parameters) [here](https://github.com/CDCgov/prime-reportstream/blob/master/frontend-react/src/pages/admin/AdminRevHistory.tsx#L197-L198). The underlying issue is that we were just telling the query cache to cache on the very, very general and literal query key of ["org", "settingType"] so it was returning the cached response rather than making a new call specific to the Organization/setting type.

Ah, sorry yes, path fragments vs. query params! Looks like they are getting passed down (like you said) but not from [here](https://github.com/CDCgov/prime-reportstream/blob/master/frontend-react/src/pages/admin/AdminRevHistory.tsx#L197-L198), but from here: https://github.com/CDCgov/prime-reportstream/blob/master/frontend-react/src/pages/admin/AdminRevHistory.tsx#L196-L201.

I know out of the scope here, but a little convoluted in how it's passed all the way down...and seems like `org` can be `""` which I assume won't throw a potential error in `rsUseQuery`?

stephenkao:
> 

Those two links are pointing to the same location -- did you mean to post another link? :P

Regarding passing down the props, I agree with you for this specific case because there's only one instance of MainRevHistoryComponent in the entire app, but I think the idea here was that we could use it elsewhere irrespective of the page location.  Put another way, we could simplify it so that MainRevHistoryComponent calls `useParams` directly, but that'd make it impossible to use that component on a different page because it'd always be tied to paths that have to include the `:org` and `:settingType` fragments.

PR #8551, opened 2023-03-01
Update Jest to run in UTC

stephenkao:
This changeset updates our test commands to run in UTC all the time.  This was causing inconsistencies between our machines because timezones are anathema to tests.


penny-lischer (approved):
LGTM

etanb (approved):
LGTM! Is there a place we could document this?

jpandersen87 (approved):
yep, the test passes now. bitten by the datetime programmer's curse

stephenkao:
> LGTM! Is there a place we could document this?

@etanb Sure, where are you thinking the documentation should live?  I personally think having these sorts of environment variables in package.json/Jest config makes our global values clear (and hopefully prevents anyone from having to think about timezones while writing tests), but not problem to call attention to it just in case.  I could add it to the forthcoming test conventions doc if you'd like!

I've also used the simple-but-effective [MockDate](https://github.com/boblauer/MockDate) for things like Storybook, and that's something we could consider using as well.

PR #8561, opened 2023-03-02
8552 Gracefully Handle Routing Filters when a property isn't present in the FHIR Bundle

### LINKED ISSUE #8552: Gracefully Handle Routing Filters when a property isn't present in the FHIR Bundle
User Story.
As an oncall engineer I don't want to be paged unnecessarily.

Description/Use Case
We currently throw an error when we try to evaluate a filter condition and a boolean is not returned. This will cause our on call to be paged. It is expected that there will be filter conditions that evaluate to null because the data doesn't exist in the bundle we were sent and therefore cause us to not return a boolen. 

We also want to better proof the filters that are in our settings. Please update prod, staging, and organizations.yml with the check to see if the resource exists before comparing it.  For example `%performerState.exists() and %performerState = 'CA'`

Risks/Impacts/Considerations

Dev Notes:
* Error thrown here: https://github.com/CDCgov/prime-reportstream/blob/master/prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt#L454 

Acceptance Criteria
UP logs a warning when information isn't present in the FHIR bundle
UP still logs that a filter couldn't be evaluated because information wasn't present

jack-h-wang:
This PR changes the default filters to avoid exceptions when a FHIR bundle property being queried doesn't exist, and reduces the log level of query exceptions so that on call is paged less frequently.

Test Steps:
1. Remove the value of `performerState` or `patientState` from a UP submission. The shorthand definitions can be viewed in `fhirpath_filter_shorthand.csv`.
2. Observe the UP routing function and confirm either an exception is not thrown (if the `organization.yml` updates are in place) or that an exception is written as a warning instead of an error.

## Changes
- Changed the handling of a `SchemaException` in the UP routing function to write a `warn` in the actionLogger instead of `error`.
- Changed the handling of the same so that the `SchemaException` is not rethrown after being logged. Also changed the handling of a call to the query function to not expect an exception to be thrown in this case.
- Renamed the actionLogger message object to better describe the scenario.
- Updated the default queries defined for full ELR receivers to handle the case that the target object may not exist without throwing an exception.

### Testing
- [x] Tested locally?
- [ ] Added tests?

## Linked Issues
- Fixes #8552


prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
```diff
@@ -451,8 +443,8 @@ class FHIRRouter(
                 )
             }
         } catch (e: SchemaException) {
-            actionLogger?.error(InvalidFilterExpressionMessage(e.message ?: ""))
-            throw e
+            actionLogger?.warn(EvaluateFilterConditionErrorMessage(e.message ?: ""))
```
carlosfelix2, line 446: Add a default message instead of just empty string

jimmyfagan (approved):

jimmyfagan (approved):

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
jack-h-wang: A default message is now provided in the error object constructor. There is no need for the UP router function to provide one.

PR #8575, opened 2023-03-02
React v18 update

### LINKED ISSUE #8338: Upgrade to React v18

jpandersen87:
Fixes #8338 

This PR:
- updates react to v18
- uses react v18 rendering mode aka concurrent (non-strict)
- updates related react testing libraries to their latest versions
- fixes bugs with tests due to these changes
  - renderHook is now exported from @testing-library/react
  - testing-library's renderHook no longer provides wait functions in it's return
  - waitFor is used in place of waitForNextUpdate in light of previously mentioned library change
  - testing-library's getByRole no longer allows regex, nor allows constraints in options
  - Table/Pagination nav page item testing logic refactored in light of previously mentioned library change


frontend-react/package.json
```diff
@@ -192,10 +189,6 @@
     "webpack-manifest-plugin": "^5.0.0",
     "workbox-webpack-plugin": "^6.5.4"
   },
-  "resolutions": {
```
stephenkao, line 195: 🔥 🔥 🔥

frontend-react/src/components/Table/Pagination.test.tsx
```diff
@@ -9,55 +9,38 @@ import Pagination, {
...
         {
             description: "when there is only one page",
             slots: [1],
             currentPageNum: 1,
-            expectedItems: ["1"],
         },
     ])(
         "Handles Previous and Next links $description",
-        ({ slots, currentPageNum, expectedItems }) => {
+        ({ slots, currentPageNum }) => {
```
stephenkao, line 40: I feel like the changes here are making it harder to understand exactly what these tests are validating at first glance.  I realize that the custom logic here is to account for the different descendants (and that we also chatted about the direct descendants in the previous pull request), so this might be a good use case for using a common `data-testid` between all of them just so we can grab all buttons and overflow indicators and run expectations against their text content.  What do you think?

stephenkao (approved):
Looks good to me!  Thanks for splitting this up!

PR #8578, opened 2023-03-02
Consolidate test renderers #8292

### LINKED ISSUE #8292: Consolidate test renderers
## Problem statement

We have multiple, generally single-purpose renderers that we use in our test code, and this sort of stratified logic makes our tests more difficult to maintain.  We can consolidate them into two test renderers: one for rendering components and one for rendering hooks.  This would standardize our approach to rendering tests and make sure that no shared values or Context Providers are missing when rendering.

## What you need to know
- Check out [CustomRenderUtils.tsx](https://github.com/CDCgov/prime-reportstream/blob/master/frontend-react/src/utils/CustomRenderUtils.tsx), which contains our many renderers: renderWithRouter, renderWithSession, renderWithCustomWrapper, renderWithFullAppContext, renderWithQueryProvider, renderWithFeatureFlags
- I'm marking this as a 5 because I can imagine that there are some hard dependencies on specific pieces of logic in the test renderers in this code base.  If it turns out to be a lot easier, please feel free to reduce the point value!

## Acceptance criteria
- [ ] All renderWith* renderers are collapsed into two: one for components and one for hooks
- [ ] All `render`/`renderHook` calls are updated to use the new standardized renderers

jpandersen87:
Fixes #8292

This PR unifies the custom test render functions to one and makes minimum changes to tests to account for this. 

The unified custom render function accepts an options bag that can include restHookFixtures and initialRouteEntries.
These options allow us to stop mocking the rest-hooks and react-router-dom modules in most cases (partial module mocks for react-router-dom may still be needed for hooks like useNavigation).

When testing a component that relies on react-router-dom hooks such as useLocation, useParams, or useSearchParams we can provide initialRouteEntries to the render function so that these hooks work naturally without mocking.

For remaining rest-hook tests we use restHookFixtures to use rest-hooks' built-in mocking functionality. Rest-hooks will conveniently emit a console warning when a rest-hook data fetch occurs that was not matched to a fixture with details to help debugging.


frontend-react/src/AppRouter.tsx
```diff
@@ -39,77 +39,76 @@ export enum FeatureName {
...
-        <Route
-            path="/admin/revisionhistory/org/:org/settingtype/:settingType"
-            element={<AdminRevHistoryWithAuth />}
-        />
-        <Route path="/file-handler/validate" element={<ValidateWithAuth />} />
-        {/* Handles any undefined route */}
-        <Route path={"*"} element={<ErrorNoPage />} />
-    </Routes>
-);
+export const appRoutes = [
```
stephenkao, line 42: Having this as an array makes me so much happier than it should 🙏

frontend-react/src/components/Admin/AdminReceiverDashboard.test.tsx
```diff
@@ -1,15 +1,13 @@
...
-import { MemoryRouter } from "react-router-dom";
 
 import { AdmConnStatusDataType } from "../../resources/AdmConnStatusResource";
 import { ErrorPage } from "../../pages/error/ErrorPage";
-import { renderWithBase } from "../../utils/CustomRenderUtils";
+import { renderApp } from "../../utils/CustomRenderUtils";
 
 import { _exportForTesting } from "./AdminReceiverDashboard";
 
-// <editor-fold defaultstate="collapsed" desc="mockData: AdmConnStatusDataType[]">
```
stephenkao, line 12: Thank you for removing these!

frontend-react/src/components/Admin/AdminReceiverDashboard.test.tsx
```diff
@@ -237,30 +235,28 @@ describe("AdminReceiverDashboard tests", () => {
...
-                                _exportForTesting.SuccessRate.ALL_SUCCESSFUL
-                            }
-                            filterErrorText={" "}
-                            filterRowReceiver={"-"}
-                            onDetailsClick={(
-                                _subdata: AdmConnStatusDataType[]
-                            ) => {}}
-                        />
-                    </NetworkErrorBoundary>
+                    {/*eslint-disable-next-line react/jsx-pascal-case*/}
```
stephenkao, line 243: Blech, I might have already left a comment about this on your previous pull request, but we should get rid of the `exportForTesting` pattern in our codebase we encounter it/when we can -- if things need to be exported, they should just be exported.  While it's unavoidable in very rare cases, we shouldn't be formatting our source code around tests.  (Fortunately it's not too commonly used in this codebase, though!)

DEFINITELY DEFINITELY DEFINITELY out of the scope of this pull request -- just wanted to raise it for awareness as we touch files that have this pattern! FYI @penny-lischer @etanb

frontend-react/src/pages/submissions/SubmissionTable.test.tsx
```diff
@@ -1,31 +1,14 @@
...
+import { renderApp } from "../../utils/CustomRenderUtils";
 import { mockSessionContext } from "../../contexts/__mocks__/SessionContext";
 import { MemberType } from "../../hooks/UseOktaMemberships";
 import { Organizations } from "../../hooks/UseAdminSafeOrganizationName";
 
 import SubmissionTable from "./SubmissionTable";
 
-// const { addFeatureFlag, removeFeatureFlag } = _exportForTesting;
-
-// TODO: Move this to CustomRenderUtils.tsx once we stop mocking rest-hooks.
```
stephenkao, line 16: 🔥 🔥 🔥

frontend-react/src/utils/CustomRenderUtils.tsx
```diff
@@ -1,174 +1,100 @@
...
-    ({ children }: PropsWithChildren<{}>) =>
-        (
-            <BaseWrapper>
-                <QueryClientProvider client={client}>
-                    <AuthorizedFetchProvider initializedOverride={true}>
-                        {children}
-                    </AuthorizedFetchProvider>
-                </QueryClientProvider>
-            </BaseWrapper>
+export const AppWrapper = ({
```
stephenkao, line 43: SO MUCH BETTER.  Thank you for this 🙏 🙏 🙏 🙏 

The singularity!
giphy

frontend-react/src/utils/CustomRenderUtils.tsx
```diff
@@ -1,174 +1,100 @@
...
-export const renderWithFeatureFlags = (
-    ui: ReactElement,
-    options?: Omit<RenderOptions, "wrapper">
-) =>
-    render(ui, {
-        wrapper: FeatureFlagWrapper,
-        ...options,
-    });
-
 export * from "@testing-library/react";
```
stephenkao, line 100: Since we're funneling all exports from RTL, we may want to consider exporting `renderApp` as the default `render` (and eventually a default `renderHook`) and just import everything from CustomRenderUtils in our tests.  It's a suggested pattern in the official RTL docs [here](https://testing-library.com/docs/react-testing-library/setup/#custom-render), and it's something I've personally implemented at previous workplaces to pretty good effect.  It's especially useful if we need to extend the return value with, say, a query client from react-query or something.

Not something we need to do now of course!  Just wanted to get your thoughts!

stephenkao (approved):
Looks fantastic.  Thanks for all this consolidation!

frontend-react/src/utils/CustomRenderUtils.tsx
jpandersen87: I noticed this as a recommended pattern when researching tests. It also recommends changing the config so that the folder containing the new testing exports module be loaded at root so that only a simple fixed import statement is needed rather than a relative one (import { render } from "CustomRenderUtils" vs import { render } from "../../../CustomRenderUtils") which is why I held off on making this pattern change until it was discussed. I'm for this change, whether we go the extra step on the root import or not.

frontend-react/src/pages/submissions/SubmissionTable.test.tsx
jpandersen87: This was the catalyst for going forward with the consolidation, since even though we're not done with rest-hooks yet the changes required to tests still using them to use fixtures was minimal 🙂.

frontend-react/src/AppRouter.tsx
jpandersen87: Indeed. Having routes in a reusable data type was always planned, and in fact is the preferred pattern in the current version of react-router-dom (until the next update completely changes the API surface as is usual for that library... yeesh).

PR #8580, opened 2023-03-02
UP sender transform on repeating resources

### LINKED ISSUE #7248: UP Sender Transform can transform existing data for repeating FHIR resources
### User Story:
As a receiver, I want sender data to be corrected and/or normalized, so I can receive correct data.

### Description/Use Case
Similar to the COVID Pipeline, we need to have the ability to transform sender data to be able to fix and/or normalize the data.  One example of this data problem would be the use of different strings to mean the same thing (e.g. Y, y, yes, positive). Some senders may also use non-standard data that is unique to them, so a generic approach for all senders may not work and hence why we need sender specific transforms.

* 

### Dev Notes:
* Note that this is to update existing data in the FHIR bundle, so you do not have to worry about how to create new FHIR properties at this time.
* Per the design, repetitions are handled by using child schemas, resourceIndex and resource, just like the FHIR to HL7 library.
* Reference the design #6761

### Acceptance Criteria
* The UP sender transform can set an existing repeating (e.g. many DiagnosticReport and not Patient) FHIR bundle property to a value specified using FHIR Path.

jimmyfagan:
This PR confirms that UP sender FHIR transforms can edit repeating resources. The changes here mostly involve adding a test, and making minor edits to existing code to allow the test to work, and to clean up some unneeded properties in the `FhirTransformer`. Note that none of the functional code changes are directly relevant to the goals of this ticket, it just turns out that the implementation of the FhirTransformer from [this previous PR](https://github.com/CDCgov/prime-reportstream/pull/8456) was already sufficient for this ticket's acceptance criteria.

Test Steps:
1. Run the automated tests (this is still not integrated into the running UP)

## Changes
- Added a new test to confirm that repeating `ServiceRequest` elements can be updated by the FHIR transformer using the `resourceIndex` on the schema.
- Removed the `ConstantSubstitutor` from `FhirTransformer` since it isn't needed (fhirpath evalutation handles the constants from the context object that gets passed around.
- Adjusted the type conversion in `convertFhirType` to handle conversion from integer types to string compatible types; this was done to allow the new test to pass while using the `resourceIndex` as a value.

### Testing
- [x] Tested locally?
- [x] Added tests?

### Process

## Linked Issues
- Fixes #7248


JessicaWNava (approved):
I am having a hard time how those changes result in repeating resources being transformed, but it seems to work and the code is clean so going to approve.

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirBundleUtils.kt
```diff
@@ -42,7 +42,7 @@ object FhirBundleUtils : Logging {
     fun convertFhirType(value: Base, sourceType: String, targetType: String): Base {
         return if (sourceType == targetType || targetType == "*") {
             value
-        } else if (StringCompatibleType.values().any { it.typeAsString == sourceType }) {
+        } else if (sourceType == "integer" || StringCompatibleType.values().any { it.typeAsString == sourceType }) {
```
arnejduranovic, line 45: Why is "integer" not a string compatible type? I don't like having to "manually" manage additional types here.

jimmyfagan: I went back and forth on this, basically I wanted to support converting from an integer to a string-like type, without wanting to deal with converting from other string-like types to integer because integer is so restrictive. But the same problem exists for some of the other types like date/datetime anyway, so maybe this should be handled the same way as all of those.

arnejduranovic (approved):

PR #8631, opened 2023-03-07
Update to Node 18.15

### LINKED ISSUE #7078: Update to Node 18
## Problem statement

Node 16 LTS ends **September 11, 2023**. Our current node required version is `16.18.0`. We will need to update the app and CI to run on Node 18 by the end of Node 16 LTS.

## What you need to know

- [Node EOL (end of life)](https://endoflife.date/nodejs)
- Using NVM (Node version manager) that has some info you may need to know

## Acceptance criteria

- [ ] ReportStream's front end requires Node 18.x.x
- [ ] Github Actions frontend-build requires Node 18.x.x
- [ ] Update the .nvmrc

## To do

- [ ] ...

jpandersen87:
Fixes #7078

This PR updates the required node version to 18.15.


snesm (changes requested):
Please update frontend-react/README.md as well.

jpandersen87:
> Please update frontend-react/README.md as well.

Thanks for the reminder. Knew my 16.19.2 replace-all missed something!

.github/actions/build-frontend/action.yml
```diff
@@ -28,7 +28,7 @@ runs:
     - name: Use Node.js ${{ matrix.node-version }} with yarn
       uses: actions/setup-node@64ed1c7eab4cce3362f8c340dee64e5eaeef8f7c
       with:
-        node-version: "16.19.1"
+        node-version: "18.14.2"
```
stephenkao, line 31: I wonder if we can just use [`node-version-file`](https://github.com/actions/setup-node/pull/338) so we can just maintain the version in .nvmrc instead of having to keep this updated as well.

stephenkao:
Blech, I'm still running into the proxy issue wherein all requests to localhost:7071 are refused.  BUT it works if I update `proxy` [in package.json](https://github.com/CDCgov/prime-reportstream/blob/master/frontend-react/package.json#L6) to `http://127.0.0.1:7071` so maybe we can just update to that (or `http://0.0.0.1:7071` for Docker) for a more explicit local IP address?

@etanb @penny-lischer Out of curiosity, do you run into this issue with the new version of Node?

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
```diff
@@ -271,7 +271,10 @@ export const MessageReceiversRow = ({
 //     fileName: string;
 //     transportResults: string;
 // }
-export function formatMessages(data: ReceiverData[]): NormalizedReceiverData[] {
+export function formatMessages(
```
stephenkao, line 274: Rather than update our source code to work around the bug in Node v18.13, can we just lock our version down to 18.12.1 for now?  It looks like [the Node v18.15 proposal](https://github.com/nodejs/node/pull/46920) includes [the fix](https://github.com/nodejs/node/commit/1d5058d4bb23897e933aad021e30854e111aff5b) so we can just upgrade directly to that after that's released.  Whatchu think?

jpandersen87: 18.12.1 is likely vulnerable from the CVE's that forced us to update the 16.x branch last month (see: https://nodejs.org/en/blog/vulnerability/february-2023-security-releases/)

jpandersen87: We also still need to be ready for when changes like this happen again in the future for Intl strings as browser vendors expect devs to assume Intl strings can change at any time, unless we're deeming this as a puntable issue.

.github/actions/build-frontend/action.yml
snesm: @JosiahSiegel feasible?

jpandersen87: If devops approves, I feel this would be the right way to go.

jpandersen87: Also, according to the docs (see: https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#node-version-file) it can read the engines.node property in package.json so the .nvmrc file may not be needed.

stephenkao: Oh, fantastic!  Anything automatic like this would be the way to go in my opinion 🙏

frontend-react/src/components/MessageTracker/MessageReceivers.tsx
stephenkao: The security vulnerabilities in 18.12.1 are tolerable in my opinion because we're not running Node anywhere in production; all the JavaScript that we run in production is in the browser.  But actually, this might be a moot point since 18.15 looks like it just released today?

Regarding the future of formatter strings, my opinion is that it isn't good practice to structure source code around test -- like if there weren't a difference between browser and Node implementation, we wouldn't have to pass in a `formatter` argument to `formatMessages`.  And if we were using this formatter in a lot of different places in the code, the amount of manual maintenance would add up.  

Since there was enough developer outrage to cause both browsers and Node to revert -- even if temporarily -- my vote here would be to wait to see what happens in the wider community.  Even if browsers undo their reversion before Node, it wouldn't be a breaking (pun intended?) change; the worst that would happen would be that we wouldn't have line breaks in formatted strings.  Then we'd just have to update our tests whenever Node undoes their reversion.

What are your thoughts?

etanb: > it isn't good practice to structure source code around test

+1 to this!

.github/actions/build-frontend/action.yml
```diff
@@ -28,7 +28,7 @@ runs:
     - name: Use Node.js ${{ matrix.node-version }} with yarn
       uses: actions/setup-node@64ed1c7eab4cce3362f8c340dee64e5eaeef8f7c
       with:
-        node-version: "16.19.1"
+        node-version: "18.15"
```
stephenkao, line 31: Just checking/belaboring the point -- is it not possible to read this from `engines` or the .nvmrc file?

stephenkao (commented):
Looking good to me!  Any reason this can't be taken out of draft?

.github/actions/build-frontend/action.yml
jpandersen87: Just waiting on approval from devops on going that route.

snesm (approved):

stephenkao:
> Blech, I'm still running into the proxy issue wherein all requests to localhost:7071 are refused. BUT it works if I update `proxy` [in package.json](https://github.com/CDCgov/prime-reportstream/blob/master/frontend-react/package.json#L6) to `http://127.0.0.1:7071` so maybe we can just update to that (or `http://0.0.0.0:7071` for Docker) for a more explicit local IP address?
> 
> @etanb @penny-lischer Out of curiosity, do you run into this issue with the new version of Node?

@jpandersen87 I'm still running into the proxy error even with the latest minor version bump.  What are your thoughts about updating that to be an explicit `127.0.0.1` or `0.0.0.0` instead of `localhost`?  Shouldn't adversely affect anything 🤷

jpandersen87:
> > Blech, I'm still running into the proxy issue wherein all requests to localhost:7071 are refused. BUT it works if I update `proxy` [in package.json](https://github.com/CDCgov/prime-reportstream/blob/master/frontend-react/package.json#L6) to `http://127.0.0.1:7071` so maybe we can just update to that (or `http://0.0.0.0:7071` for Docker) for a more explicit local IP address?
> > @etanb @penny-lischer Out of curiosity, do you run into this issue with the new version of Node?
> 
> @jpandersen87 I'm still running into the proxy error even with the latest minor version bump. What are your thoughts about updating that to be an explicit `127.0.0.1` or `0.0.0.0` instead of `localhost`? Shouldn't adversely affect anything 🤷

Explicit ip address commit pushed

stephenkao (approved):
Looks great to me!

PR #8635, opened 2023-03-07
Experience-8026: Widen grid container

stephenkao:
This changeset updates the follow pages' containers to be widescreen (up to 1400px) since they're the most data-heavy:
- Daily Data
- Submissions
- Message Tracker

The Public Validator page will continue to be on 1024px to align with the rest of the public pages.

Additionally, there's some cleanup regarding containers in the codebase across several different pages, and I've updated `grid-container` modifier classes to use the GridContainer component instead.  There are a lot of `grid-container` modifier classes used around the codebase, and it really shouldn't be used this much; it should be used at the highest ancestor node of common width, and all descendants should naturally expand to their own widths inside of that.  The grid container sets a maximum width, yes, but it also includes padding to account for grid gutters -- if this is used too frequently, there'll be a lot of horizontal jaggedness across the website.

**tl;dr In most cases, a grid container should be used at the top of the page and nowhere else.**  There are obviously use cases for nested grid containers, but they should be rare and should be brought up by design rather than used anywhere we need a maximum width -- especially not in components that should be able to be rendered agnostically to the wrapping context.

Daily Data:
Screenshot 2023-03-13 at 10 43 58
Screenshot 2023-03-13 at 10 44 38

Submissions:
Screenshot 2023-03-13 at 10 44 08
Screenshot 2023-03-13 at 10 44 25

Message Tracker:
Screenshot 2023-03-13 at 10 44 52
Screenshot 2023-03-13 at 10 44 57

### DIFF
--- a/frontend-react/src/components/Admin/AdminReceiverDashboard.tsx
+++ b/frontend-react/src/components/Admin/AdminReceiverDashboard.tsx
@@ -875,7 +875,7 @@ export function AdminReceiverDashboard() {
     }, []);
 
     return (
-        <section className="grid-container">
+        <article>
             <h4>Receiver Status Dashboard</h4>
             <section>
                 CRON job results that check if receivers are working.
@@ -1017,7 +1017,7 @@ export function AdminReceiverDashboard() {
             >
                 <ModalInfoRender subData={currentDataForModal} />
             </Modal>
-        </section>
+        </article>
     );
 }
 
--- a/frontend-react/src/components/FileHandlers/FileHandler.tsx
+++ b/frontend-react/src/components/FileHandlers/FileHandler.tsx
@@ -1,4 +1,5 @@
 import React, { useCallback, useEffect, useState } from "react";
+import { GridContainer } from "@trussworks/react-uswds";
 
 import { showError } from "../AlertNotifications";
 import { useSessionContext } from "../../contexts/SessionContext";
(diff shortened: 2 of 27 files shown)


frontend-react/src/components/FileHandlers/FileHandler.tsx
```diff
@@ -260,82 +261,86 @@ function FileHandler() {
     }
 
     return (
-        <div className="grid-container usa-section margin-bottom-10">
```
stephenkao, line 263: No markup changes here until than wrapping in a GridContainer + article.

frontend-react/src/components/Admin/AdminReceiverDashboard.tsx
```diff
@@ -875,7 +875,7 @@ export function AdminReceiverDashboard() {
     }, []);
 
     return (
-        <section className="grid-container">
+        <article>
```
stephenkao, line 878: Articles make more semantic sense than sections for our pages because they're self-contained bodies of content.  Updated a bunch of other pages as well, but we should keep this in mind going forward.

frontend-react/src/components/Table/Table.tsx
```diff
@@ -124,10 +124,7 @@ const Table = ({
         return config.rows;
     }, [config.rows, filterManager?.sortSettings]);
 
-    const wrapperClasses = useMemo(
-        () => `grid-container margin-bottom-10 ${classes}`,
```
stephenkao, line 128: If we want to be able to use atomic components anywhere, we can't tie them down to context-specific styles like the `grid-container` here -- that should be taken care of in an ancestor node, and Table/TableFilters should expand to their natural widths inside of that.

stephenkao:
NOTE: This is still subject to review by @jillian-hammer when it's ready on demo3.

jpandersen87 (approved):

PR #8648, opened 2023-03-07
Improve buffer handling

JessicaWNava:
This PR ...

Test Steps:
1. Just test that you can still send a normal message through the HL7 pipeline
2. Also make sure that if you send it with a message with an exceptionally long message line that you see the warning in the logs.

## Changes
-Improving buffer handling

### Testing
- [x] Tested locally?
- [ ] Added tests?

### Process

## Linked Issues
- Fixes #issue 8595

## To Be Done
- #issue

    - new endpoint A
    - new endpoint B


prime-router/src/main/kotlin/serializers/Hl7Serializer.kt
```diff
@@ -193,7 +193,17 @@ class Hl7Serializer(
             }
 
             if (it.isNotBlank()) {
-                nextMessage.append("$it\r")
+                if (it.length + nextMessage.length <= nextMessage.capacity()) {
+                    if (it.length > 4000) {
+                        logger.warn("Message segment ${it.substring(0,3)} longer than 4000 characters, truncating")
+                        nextMessage.append("${it.substring(0,4000)}\r")
```
jimmyfagan, line 199: Do we want to truncate this at `nextMessage.capacity - nextMessage.length` rather than 4000? If the message is say 3000 chars long, and we're appending a 5000 char `it`, even if we truncate that to 4000, we'll end up with 7000 chars.

jimmyfagan (commented):
Looks good, left a comment for a small potential improvement, let me know when you check that out, I can re-review anytime.

prime-router/src/main/kotlin/serializers/Hl7Serializer.kt
arnejduranovic, line 199: same as above, magic number? If you change this 4000 in the future we should only have to change it in one spot.

arnejduranovic (changes requested):

prime-router/src/main/kotlin/serializers/Hl7Serializer.kt
JessicaWNava: I think there is some confusion here. This is just truncating a singular line of the message to 4000 characters because we have deemed it unreasonable that a single message segment would be longer than that. If it is, we are limiting it to 4000 characters here. Happy to go over this with you if that isn't making sense.

prime-router/src/main/kotlin/serializers/Hl7Serializer.kt
```diff
@@ -193,7 +195,17 @@ class Hl7Serializer(
...
+                        logger.warn("Message segment ${it.substring(0,3)} longer than 4000 characters, truncating")
+                        nextMessage.append("${it.substring(0, MESSAGE_SEGMENT_LENGTH)}\r")
+                    } else {
+                        nextMessage.append("$it\r")
+                    }
+                } else {
+                    logger.error("Message too long, returning.")
+                    error("Message too long")
+                }
             }
```
victor-chaparro, line 209: Do we need to add Unit tests to test these changes?

prime-router/src/main/kotlin/serializers/Hl7Serializer.kt
jimmyfagan: Ah okay, I was equating the `nextMessage` capacity and this 4000 without realizing they were different numbers, so that did confuse me a little bit. I'm still thinking that maybe the check at 196 should use min(it.length, 4000) instead of just it.length, since there could be cases where the full segment would exceed the capacity, but the truncated segment would still fit. Maybe an edge case that's not worth worrying about since these boundaries were set beyond what we realistically expect. Will leave that up to you, not going to block approval on that.

jimmyfagan (approved):

arnejduranovic (approved):
Thank you for taking on this work and getting a PR up so quickly!

prime-router/src/main/kotlin/serializers/Hl7Serializer.kt
JessicaWNava: The first if statement checks for the capacity and will throw an error if the two together exceed it. Does that make sense?

prime-router/src/main/kotlin/serializers/Hl7Serializer.kt
JessicaWNava: I started to but couldn't figure out how to actually call the method directly. No matter what I did I just couldn't call it in the test. Do you have an idea of how to do this?

prime-router/src/main/kotlin/serializers/Hl7Serializer.kt
jimmyfagan: I follow that, I'm just saying that it might not actually exceed the capacity after the new segment is truncated, so we _could_ let those cases through. For example, say we have 30000 of 40000 used in the `nextMessage` buffer and a 15000 length `it` comes in, 15000 + 30000 is above 40000 so we'd throw that error, but if we instead let it through, we'd truncate the 15000 down to 4000, and appending that, we'd still only be at 34000 of 40000 used in the overall buffer. Might be easier to discuss over a call, but it's an unimportant edge case anyway, so we could also disregard this 🤷

victor-chaparro (approved):
Code coverage looks good!

Issue #7253, opened 2022-11-04
UP Sender Transform schema can define constants to use for an element


carlosfelix2:
### User Story:
As a administrator, I want to be able to define constants in a UP sender transform schema, so I don't have to duplicate data.

### Description/Use Case
Similar to the FHIR to HL7 library, we want to be able to define string constants for the UP sender transform schemas to be used at the schema or element level.

### Risks/Impacts/Considerations
* 

### Dev Notes:
* Follow the same implementation used for the FHIR to HL7 library

### Acceptance Criteria
* The UP sender transform schema can have constants defined at the schema level.
* The UP sender transform schema can have constants defined at the element level.
* Element level constants are inherited by child schemas defines at the same element.
* Schema level constants are inherited by all the elements in the schema.
* Constants are applied to the sender transform

PR #8652, opened 2023-03-08
Jimmy/7253 up sender transform constants

jimmyfagan:
This PR involves adding tests to verify/demonstrate that constants are appropriately handled in the `FhirTransformer`. This was pretty straightforward with the exception of the `bundleProperty` field. `bundleProperty` is parsed on our end before passing it over to `FhirPathUtils`; we need to find the path to the property being updated (including each step along that path in case we can create missing properties), as well as the property name itself, and therefore, we can't depend on the `FhirPathUtils` to handle constants as we can for the other fields which may make use of the constants. Instead, I re-introduced the `ConstantSubstitutor` so we can do some pre-processing string substitution on constants that are wrapped in `%{` and `}`.

I also stumbled upon some error cases in the process of writing these tests, so included some changes intended to make certain error cases a bit more graceful, and more clearly logged.

Test Steps:
1. Fhir Sender Transforms have still not been integrated in the pipeline, so for now, running the `FhirTransformerTests` and checking code coverage is the best way to test these changes.

## Changes
- New test cases to cover the acceptance criteria listed in #7253 in `FhirTransformerTests.kt`
- Add ConstantSubsitutor to `FhirTransformer.kt` (and update comment about ConstantSubstitutor in `FhirToHl7Converter.kt`), remove currently unneeded property `strict` from `FhirTransformer.kt`
- Adjust error handling and logging in `FhirTransformer.kt` and `FhirBundleUtils.kt`
- Add logging checks and adjust existing error testing in `FhirTransformerTests.kt`

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- Fixes #8652


prime-router/src/main/kotlin/fhirengine/translation/hl7/FhirTransformer.kt
```diff
@@ -172,15 +161,23 @@ class FhirTransformer(
...
     internal fun setBundleProperty(
-        bundleProperty: String?,
+        rawBundleProperty: String?,
         value: Base,
         context: CustomContext,
         bundle: Bundle,
         focusResource: Base
     ) {
-        if (bundleProperty == null) return
+        if (rawBundleProperty == null || rawBundleProperty == "") {
```
JessicaWNava, line 170: You can replace this with a check for .isNullOrBlank()

JessicaWNava (approved):
One minor suggestion, other than that, looks good!

prime-router/src/main/kotlin/fhirengine/translation/hl7/FhirTransformer.kt
jimmyfagan: Oo good call, thanks for the reminder!

jimmyfagan:
Updated this with another commit to remove the support for `%{constant}` and to add some error handling to enforce some of the restrictions around the use of constants within `bundleProperty`. Ready for review again!

prime-router/src/main/kotlin/fhirengine/translation/hl7/FhirTransformer.kt
```diff
@@ -178,22 +164,42 @@ class FhirTransformer(
...
     ) {
-        if (bundleProperty == null) return
+        if (bundleProperty.isNullOrBlank()) {
+            logger.warn("bundleProperty was not set.")
+            return
+        }
 
         val pathParts = bundleProperty.split(".")
+        if (pathParts.size < 2) {
+            logger.warn("Expected at least 2 parts in bundle property '$bundleProperty'.")
```
carlosfelix2, line 174: Not sure this limitation is needed. Are you sure?

prime-router/src/main/kotlin/fhirengine/translation/hl7/FhirTransformer.kt
```diff
@@ -213,7 +219,11 @@ class FhirTransformer(
         }
         // Finally set the value
         val property = childResource.getNamedProperty(pathParts.last())
-        val newValue = FhirBundleUtils.convertFhirType(value, value.fhirType(), property.typeCode)
+        if (property == null) {
+            logger.warn("Could not find property '${pathParts.last()}'.")
```
carlosfelix2, line 223: This may  be an error

prime-router/src/main/kotlin/fhirengine/translation/hl7/FhirTransformer.kt
```diff
@@ -213,7 +219,11 @@ class FhirTransformer(
...
+            return
```
carlosfelix2, line 224: Too are many returns in this function. Is there a way to refactor this?

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirBundleUtils.kt
```diff
@@ -40,7 +41,7 @@ object FhirBundleUtils : Logging {
      * Converts a [value] of type [sourceType] into a compatible Base of type [targetType]. Returns the original value
      * and logs an error if the conversion is not supported.
      */
-    fun convertFhirType(value: Base, sourceType: String, targetType: String): Base {
+    fun convertFhirType(value: Base, sourceType: String, targetType: String, logger: KotlinLogger = this.logger): Base {
```
carlosfelix2, line 44: logger should not be needed as an argument. It is already provided by the class

prime-router/src/test/kotlin/fhirengine/translation/hl7/FhirTransformerTests.kt
```diff
@@ -332,42 +352,222 @@ class FhirTransformerTests {
...
             )
         }.isFailure()
 
-        // Can't currently create new resources on the fly
+        // Improper extension format
+        transformer.setBundleProperty(
+            "Bundle.entry.resource.ofType(Patient).extension(regexNonMatch).value[x]", IdType("newId"),
+            CustomContext(bundle, bundle), bundle, bundle
+        )
+        verify(exactly = 1) { logger.warn(match<String> { it.contains("Could not find property") }) }
```
carlosfelix2, line 385: Brittle tests. Remove all the tests for the specific message in the log. This also makes it sound like it should be an error.

prime-router/src/test/kotlin/fhirengine/translation/hl7/FhirTransformerTests.kt
```diff
@@ -332,42 +352,222 @@ class FhirTransformerTests {
...
+            "", IdType("newId"),
+            CustomContext(bundle, bundle), bundle, bundle
+        )
+        verify(exactly = 1) { logger.warn(match<String> { it.contains("bundleProperty was not set") }) }
+
+        transformer.setBundleProperty(
+            "id", IdType("newId"),
+            CustomContext(bundle, bundle), bundle, bundle
+        )
+        verify(exactly = 1) { logger.warn(match<String> { it.contains("Expected at least 2 parts") }) }
```
carlosfelix2, line 398: Brittle tests. Remove all the tests for the specific message in the log. This also makes it sound like it should be an error.

prime-router/src/test/kotlin/fhirengine/translation/hl7/FhirTransformerTests.kt
```diff
@@ -332,42 +352,222 @@ class FhirTransformerTests {
...
+            "%key.text", StringType("SomeName"),
+            CustomContext(
+                bundle,
+                bundle,
+                constants = mutableMapOf(Pair("key", "Bundle.entry.resource.ofType(Patient).name"))
+            ),
+            bundle, bundle
+        )
+        verify(exactly = 1) {
+            logger.warn(match<String> { it.contains("cannot dynamically create components relying on constants") })
```
carlosfelix2, line 410: Brittle tests. Remove all the tests for the specific message in the log. This also makes it sound like it should be an error.

prime-router/src/test/kotlin/fhirengine/translation/hl7/FhirTransformerTests.kt
```diff
@@ -332,42 +352,222 @@ class FhirTransformerTests {
...
+        }
+
+        transformer.setBundleProperty(
+            "Bundle.entry.resource.ofType(Patient).name.%key", StringType("SomeName"),
+            CustomContext(bundle, bundle, constants = mutableMapOf(Pair("key", "text"))), bundle, bundle
+        )
+        verify(exactly = 1) { logger.warn(match<String> { it.contains("Constants not supported in lowest level") }) }
```
carlosfelix2, line 417: Your test are verifying logs and not results. It makes it seem there should be some return to this functions and perhaps needs a refactor.

prime-router/src/test/kotlin/fhirengine/translation/hl7/FhirTransformerTests.kt
```diff
@@ -332,42 +352,222 @@ class FhirTransformerTests {
...
+
+        // Incompatible value types
         assertThat {
             transformer.setBundleProperty(
-                "Bundle.entry.resource.ofType(DiagnosticReport).status", CodeType("final"),
+                "Bundle.entry.resource.ofType(Patient).name.text", CodeableConcept(),
                 CustomContext(bundle, bundle), bundle, bundle
             )
         }.isFailure()
+        verify(exactly = 1) { logger.error(match<String> { it.contains("between CodeableConcept and string") }) }
```
carlosfelix2, line 426: Remove

prime-router/src/test/kotlin/fhirengine/translation/hl7/FhirTransformerTests.kt
```diff
@@ -332,42 +352,222 @@ class FhirTransformerTests {
...
 
-        // Improper extension format
         assertThat {
             transformer.setBundleProperty(
-                "Bundle.entry.resource.ofType(Patient).extension(regexNonMatch).value[x]", IdType("newId"),
+                "Bundle.entry.resource.ofType(Patient).active", StringType("nonBoolean"),
                 CustomContext(bundle, bundle), bundle, bundle
             )
         }.isFailure()
+        verify(exactly = 1) { logger.error(match<String> { it.contains("between string and boolean") }) }
```
carlosfelix2, line 434: Remove

prime-router/src/main/kotlin/fhirengine/translation/hl7/FhirTransformer.kt
jimmyfagan: Yes, without it we can't use `setProperty` since it expects a property name and doesn't use fhir path.

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirBundleUtils.kt
jimmyfagan: It's an optional argument, by default uses the one on the class, but accepts a different logger to assist in testing the logging with mocks. From your other comments it sounds like you're not a fan of testing the logging, so let's discuss this as a 17th and we can decide how to handle this.

prime-router/src/main/kotlin/fhirengine/translation/hl7/FhirTransformer.kt
jimmyfagan: I extracted some of the validation at the beginning of the function into a helper so there is one return there instead of three, and then reordered some logic to remove another one. Hopefully that helps. Will check that in with a new commit shortly.

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirBundleUtils.kt
jimmyfagan: I do think this is needed here to help with the logging mocking, even if the specific messages still aren't being verified, it's nice to be able to verify that an error or warning is logged.

prime-router/src/main/kotlin/fhirengine/translation/hl7/FhirTransformer.kt
jimmyfagan: Changed to an error

prime-router/src/test/kotlin/fhirengine/translation/hl7/FhirTransformerTests.kt
jimmyfagan: Changed to check for an error without checking for specific message.

prime-router/src/test/kotlin/fhirengine/translation/hl7/FhirTransformerTests.kt
jimmyfagan: Changed to check for an error without checking for specific message

prime-router/src/test/kotlin/fhirengine/translation/hl7/FhirTransformerTests.kt
jimmyfagan: Changed to check for an error without checking for specific message, also moved some of these cases into a new test.

prime-router/src/test/kotlin/fhirengine/translation/hl7/FhirTransformerTests.kt
jimmyfagan: Done

jimmyfagan:
In my latest two commits, I made some refactors to hopefully make functions easier to read. Part of this involved extracting a function for validation of inputs. I also removed the parts of tests that looked for specific logging messages and instead will just verify that errors did happen. I also changed most/all of the warnings in the function to errors. I think as part of the integration into the pipeline, there will be more error handling work to be done, but I don't want to get more into that as part of this PR, since it's not really related to this story.

carlosfelix2 (approved):

PR #8656, opened 2023-03-08
Move api keys to organization

mkalish:
This PR ...

Test Steps:

## Changes
- Adds a keys field to the organization for storing JwkSet
- Updates the authentication process to look for keys on the organization
- Updates the gradle step for adding key to create it on the organization rather than the sender

### Testing
- [ ] Tested locally?
- [ ] Added tests?

### Process

## Linked Issues
- Fixes #issue

## To Be Done
- #issue

    - new endpoint A
    - new endpoint B

### DIFF
--- a/prime-router/src/main/kotlin/Organization.kt
+++ b/prime-router/src/main/kotlin/Organization.kt
@@ -1,6 +1,7 @@
 package gov.cdc.prime.router
 
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties
+import gov.cdc.prime.router.tokens.JwkSet
 
 /**
  * Organization represents a partner organization of the hub. It has a jurisdiction.
@@ -14,9 +15,11 @@ open class Organization(
     val filters: List<ReportStreamFilters>? = emptyList(), // one ReportStreamFilters obj per topic.
     // enabled features for organization. Features defined in lookup table rs_feature_flags
     val featureFlags: List<String>? = emptyList(),
+    val keys: List<JwkSet>?
 ) {
     constructor(org: Organization) : this(
-        org.name, org.description, org.jurisdiction, org.stateCode, org.countyName, org.filters, org.featureFlags
+        org.name, org.description, org.jurisdiction, org.stateCode, org.countyName, org.filters, org.featureFlags,
+        org.keys
     )
 
     enum class Jurisdiction {
@@ -63,12 +66,13 @@ class DeepOrganization(
     countyName: String? = null,
     filters: List<ReportStreamFilters>? = emptyList(),
     featureFlags: List<String>? = emptyList(),
+    keys: List<JwkSet>? = emptyList(),
     val senders: List<Sender> = emptyList(),
     val receivers: List<Receiver> = emptyList(),
-) : Organization(name, description, jurisdiction, stateCode, countyName, filters, featureFlags) {
+) : Organization(name, description, jurisdiction, stateCode, countyName, filters, featureFlags, keys) {
     constructor(org: Organization, senders: List<Sender>, receivers: List<Receiver>) :
         this(
             org.name, org.description, org.jurisdiction, org.stateCode, org.countyName, org.filters, org.featureFlags,
-            senders, receivers
+            org.keys, senders, receivers
         )
 }
--- a/prime-router/src/main/kotlin/Organization.kt
+++ b/prime-router/src/main/kotlin/Organization.kt
@@ -1,6 +1,7 @@
 package gov.cdc.prime.router
 
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties
+import gov.cdc.prime.router.tokens.Jwk
 import gov.cdc.prime.router.tokens.JwkSet
 
 /**
@@ -20,6 +21,17 @@ open class Organization(
     constructor(org: Organization) : this(
         org.name, org.description, org.jurisdiction, org.stateCode, org.countyName, org.filters, org.featureFlags,
         org.keys
+    )
+
+    constructor(copy: Organization, newScope: String, newJwk: Jwk) : this(
+        copy.name,
+        copy.description,
+        copy.jurisdiction,
+        copy.stateCode,
+        copy.countyName,
+        copy.filters,
+        copy.featureFlags,
+        JwkSet.addJwkSet(copy.keys, newScope, newJwk)
     )
 
     enum class Jurisdiction {
@@ -49,6 +61,10 @@ open class Organization(
                 else null
             }
         }
+    }
+
+    fun makeCopyWithNewScopeAndJwk(scope: String, jwk: Jwk): Organization {
+        return Organization(this, scope, jwk)
     }
 }
 
--- a/prime-router/src/main/kotlin/Sender.kt
+++ b/prime-router/src/main/kotlin/Sender.kt
@@ -166,40 +166,6 @@ abstract class Sender(
         const val fullNameSeparator = "."
 
         /**
-         * Copy an old set of authorizations to a new set, and add one to it, if needed.
-         * This whole list of lists thing is confusing:
-         * The 'orig' obj, and the return val, are list of JwkSets.   And each JwkSet has a list of Jwks.
-         */
-        fun addJwkSet(orig: List<JwkSet>?, newScope: String, newJwk: Jwk): List<JwkSet> {
-            if (orig == null) {
-                return listOf(JwkSet(newScope, listOf(newJwk))) // create brand new
-            }
-            val newJwkSetList = mutableListOf<JwkSet>()
-            var done = false
-            orig.forEach {
-                if (it.scope == newScope) {
-                    if (it.keys.contains(newJwk)) {
-                        // The orig already has this key with this scope.  Just use it.
-                        newJwkSetList.add(it)
-                    } else {
-                        // Don't create a whole new JwkSet.   Instead, add this Jwk to existing JwkSet.
-                        val newJwkList = it.keys.toMutableList()
-                        newJwkList.add(newJwk)
-                        newJwkSetList.add(JwkSet(newScope, newJwkList))
-                    }
-                    done = true
-                } else {
-                    newJwkSetList.add(it) // existing different scope, make sure we keep it.
-                }
-            }
-            // If the old/new scopes didn't match, then the new scope was never added.  Add a new JwkSet now.
-            if (!done) {
-                newJwkSetList.add(JwkSet(newScope, listOf(newJwk)))
-            }
-            return newJwkSetList
-        }
-
-        /**
          * returns a tuple of the sender name split into parts, with the
          * second part defaulted to 'default' if not provided
          */
@@ -271,7 +237,7 @@ class FullELRSender : Sender {
         copy.organizationName,
         copy.format,
         copy.customerStatus,
-        addJwkSet(copy.keys, newScope, newJwk)
+        JwkSet.addJwkSet(copy.keys, newScope, newJwk)
     )
 
     /**
@@ -346,7 +312,7 @@ open class TopicSender : Sender, HasSchema {
         copy.customerStatus,
         copy.schemaName,
         copy.topic,
-        addJwkSet(copy.keys, newScope, newJwk)
+        JwkSet.addJwkSet(copy.keys, newScope, newJwk)
     )
 
     /**
@@ -421,7 +387,7 @@ class CovidSender : TopicSender, HasSchema {
         copy.format,
         copy.customerStatus,
         copy.schemaName,
-        addJwkSet(copy.keys, newScope, newJwk)
+        JwkSet.addJwkSet(copy.keys, newScope, newJwk)
     )
 
     /**
@@ -486,7 +452,7 @@ class MonkeypoxSender : TopicSender, HasSchema {
         copy.format,
         copy.customerStatus,
         copy.schemaName,
-        addJwkSet(copy.keys, newScope, newJwk)
+        JwkSet.addJwkSet(copy.keys, newScope, newJwk)
     )
 
     /**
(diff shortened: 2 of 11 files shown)


prime-router/src/main/kotlin/Organization.kt
```diff
@@ -14,9 +16,22 @@ open class Organization(
...
     val featureFlags: List<String>? = emptyList(),
+    val keys: List<JwkSet>?
 ) {
     constructor(org: Organization) : this(
-        org.name, org.description, org.jurisdiction, org.stateCode, org.countyName, org.filters, org.featureFlags
+        org.name, org.description, org.jurisdiction, org.stateCode, org.countyName, org.filters, org.featureFlags,
+        org.keys
+    )
+
+    constructor(copy: Organization, newScope: String, newJwk: Jwk) : this(
```
mkalish, line 26: This is the same pattern that `Sender` used

prime-router/src/main/kotlin/Sender.kt
```diff
@@ -165,40 +165,6 @@ abstract class Sender(
     companion object {
         const val fullNameSeparator = "."
 
-        /**
-         * Copy an old set of authorizations to a new set, and add one to it, if needed.
-         * This whole list of lists thing is confusing:
-         * The 'orig' obj, and the return val, are list of JwkSets.   And each JwkSet has a list of Jwks.
-         */
-        fun addJwkSet(orig: List<JwkSet>?, newScope: String, newJwk: Jwk): List<JwkSet> {
```
mkalish, line 173: This is moved to the JwkSet companion object

prime-router/src/test/kotlin/azure/TokenFunctionTests.kt
```diff
@@ -188,6 +207,7 @@ class TokenFunctionTests {
     @Test
     fun `Test expired key`() {
         settings.senderStore.put(sender.fullName, CovidSender(sender, validScope, jwk))
+        settings.organizationStore.put(organization.name, organization)
```
mkalish, line 210: If the sender doesn't have an associated organization the server2server code will now raise an error.

prime-router/src/test/kotlin/azure/TokenFunctionTests.kt
```diff
@@ -278,6 +301,91 @@ class TokenFunctionTests {
...
+        var httpRequestMessage = MockHttpRequestMessage()
+        httpRequestMessage.parameters.put("client_assertion", token)
+        httpRequestMessage.parameters.put("scope", validScope)
+        // Invoke
+        var response = TokenFunction(UnitTestUtils.simpleMetadata).token(httpRequestMessage)
+        // Verify
+        // This fails because the current logic short circuits if the JWK scope matches the requested scope; the
+        // organization JwkSet has the correct scope, but the key is not valid, so not key is returned.
+        // See: Server2ServerAuthentication.kt#225
+        assertThat(response.getStatus()).isEqualTo(HttpStatus.UNAUTHORIZED)
```
mkalish, line 386: It was unclear whether the logic in the server2server code needs to get updated, but this assertion is accurate based on current implementation, since the code will stop processing the remaining keys after one matches the requested scope regardless if the key can ultimately be used.

arnejduranovic:
Did we create a new ticket for this epic to go back and remove the temporary backwards compatibility code? We also need a new ticket to manually update production and staging environments? If those already exist, please link them in to "TODO" section.

arnejduranovic (approved):

prime-router/src/main/kotlin/Organization.kt
```diff
@@ -14,9 +16,22 @@ open class Organization(
     val filters: List<ReportStreamFilters>? = emptyList(), // one ReportStreamFilters obj per topic.
     // enabled features for organization. Features defined in lookup table rs_feature_flags
     val featureFlags: List<String>? = emptyList(),
+    val keys: List<JwkSet>?
```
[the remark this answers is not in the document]
mkalish: So on sender, it defaults to `null` which did seem strange to me.  `emptyList` seems like the more logical default, so I'll go ahead and change that

prime-router/src/main/kotlin/tokens/Server2ServerAuthentication.kt
```diff
@@ -216,9 +216,12 @@ class FindSenderKeyInSettings(val scope: String, val metadata: Metadata) :
         val workflowEngine = WorkflowEngine.Builder().metadata(metadata).build()
```
jimduff-usds, line 216: Should we change the way we use `iss`?  For old keys stored in the Sender, the `iss` would be `orgName.senderName`, but for new keys stored in the organization, the `iss` would be just `orgName`?   Needs discussion on what the purpose of `issuer` is.

OTOH, this will likely require the Sender to change their code.   I'm hoping we can do this refactor without having any customer changes.  But once we've fully moved over, then it would be weird to use `orgName.senderName` as the issue, true?

prime-router/src/test/kotlin/azure/TokenFunctionTests.kt
jimduff-usds: Yes, my vague memory is that may be a flaw in the server2server auth code - it only tries the first key, not all the possible keys.

prime-router/src/main/kotlin/SettingsProvider.kt
```diff
@@ -22,6 +23,17 @@ interface SettingsProvider {
     fun findSender(fullName: String): Sender?
 
     fun findOrganizationAndReceiver(fullName: String): Pair<Organization, Receiver>?
+
+    // TODO: https://github.com/CDCgov/prime-reportstream/issues/8659
+    // this should be removed after all setting keys have been migrated to the organization
+    // it is just a temporary helper function for consolidating the keys between organizations and senders
+    fun getKeys(fullName: String): List<JwkSet>? {
```
jimduff-usds, line 30: Can you add kdoc here?
   It looks like currently this works for:
 `getKeys("simple_report.csvuploader")`
but not for:
`getKeys("simple_report")`

But I'm thinking it would make more sense if we allow the user to pass in an `issuer` that is just the `orgName` only.   (See my other comments on `issuer` in this review)

prime-router/src/main/kotlin/cli/SenderUtilsCommand.kt
```diff
@@ -67,27 +72,39 @@ class AddPublicKey : SettingCommand(
             return
         }
         val jwk = SenderUtils.readPublicKeyPemFile(publicKeyFile)
-        jwk.kid = if (kid.isNullOrEmpty()) settingName else kid
+        jwk.kid = if (kid.isNullOrEmpty()) orgName else kid
```
jimduff-usds, line 75: Yes, this looks right - from now on the default `kid` is just the orgName.    So, our various examples should just use the `orgName`, and not `orgName.senderName` any more.

jimduff-usds (changes requested):
Excellent work - thanks @mkalish !

Two main questions before we merge:   what do we do with `kid`, and with `iss`.

## kid

With the keys stored at the organization level, I think the `kid` the user sends us should normally just be the organization name now?   For example in this CLI call:

`  ./gradlew primeCLI --args='sender addkey --public-key "./my-rsa-public-key.pem" --scope "simple_report.*.report" --orgName "simple_report" --kid "simple_report.default" --doit'`

... the kid should just be "simple_report" not "simple_report.default" ?

Needs discussion on what the meaning and purpose of the `'kid` is.

## iss

Comments on `iss` in the code.

## Other

just for completeness, we should run the oktaauth not normally park of the Smoke:

```
./prime login --env staging
./prime test --run oktaauth
```

prime-router/src/main/kotlin/Organization.kt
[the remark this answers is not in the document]
mkalish: @jimduff-usds What would you expect the correct value to be `keys:null` vs `keys:[]`?

prime-router/src/main/kotlin/Sender.kt
```diff
@@ -486,7 +452,7 @@ class MonkeypoxSender : TopicSender, HasSchema {
         copy.format,
         copy.customerStatus,
         copy.schemaName,
-        addJwkSet(copy.keys, newScope, newJwk)
+        JwkSet.addJwkSet(copy.keys, newScope, newJwk)
```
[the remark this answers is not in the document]
mkalish: That makes sense to me!

prime-router/src/main/kotlin/tokens/Server2ServerAuthentication.kt
arnejduranovic: I think that would be weird BUT I think we have code to take a string and get either the org from it or the org and sender if the string has the "." character in it? If we use this code, then the sender can still send us `org.sender` and we can just use our special function to just pull out the org name?

prime-router/src/test/kotlin/azure/TokenFunctionTests.kt
mkalish: Ok... I'll be creating a bug ticket for this

prime-router/src/main/kotlin/tokens/Server2ServerAuthentication.kt
mkalish: One option could be in the follow up work after the keys have been migrated is to change this code to treat issuer first as an organization name and then treat it as a sender if one isn't found.  I think that would keep support for both and allow customers to switch over gracefully.

prime-router/src/main/kotlin/cli/SenderUtilsCommand.kt
mkalish: Ok, I'll find the spots where that's mentioned.

prime-router/src/main/kotlin/SettingsProvider.kt
mkalish: Yes, I can add doc as fullName in this case corresponds to a sender.  I was viewing this as a temporary function just to support grabbing keys from the two spots and not changing how issuer is treated.

mkalish:
Thanks @jimduff-usds!

Follow ups on your review:

> With the keys stored at the organization level, I think the kid the user sends us should normally just be the organization name now? For example in this CLI call:

I guess we could consider dropping the `kid` param  to the CLI if there is no reason for it to be overridden to something different; it only gets used while trying to find a key to use.  I had included that in the command for the testing because `reqtoken` cli command uses the sender name as the `kid` it uses to generate the jws for the request.  If that is a tool only used internally to check that the key works as expected, I can just update it to use the organization name as the `kid`.   Might also make sense to drop that param entirely from the `addkey` command.

From what I saw it looks like `kid` was [used](https://github.com/CDCgov/prime-reportstream/blob/3b3067513b86557e2437b55ed8afa3452296c265/prime-router/src/main/kotlin/tokens/Server2ServerAuthentication.kt#L228) in `find` condition for determining if one of JwkSet found with correct scope.  I think changing how that value is used/generated might create backwards compatibility issues.


> iss

I left my thoughts in the comment, but overall one option might be to defer changes to it for now until the keys are migrated.

prime-router/src/main/kotlin/Organization.kt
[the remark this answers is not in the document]
mkalish: For reference:
```yaml
name: "waters"
description: "Test Sender from Waters"
jurisdiction: "FEDERAL"
filters: []
featureFlags: []
version: 0
createdBy: "local@test.com"
createdAt: "2023-02-06T18:45:14.578Z"
```

prime-router/src/main/kotlin/cli/SenderUtilsCommand.kt
mkalish: Updated the readme for creating keys.

prime-router/src/test/kotlin/azure/TokenFunctionTests.kt
mkalish: https://github.com/CDCgov/prime-reportstream/issues/8670

prime-router/src/main/kotlin/cli/SenderUtilsCommand.kt
```diff
@@ -67,27 +72,39 @@ class AddPublicKey : SettingCommand(
...
 
-        val origSenderJson = get(environment, oktaAccessToken, SettingType.SENDER, settingName)
-        val origSender = jsonMapper.readValue(origSenderJson, Sender::class.java)
+        val origOrganizationJson =
+            get(environment, oktaAccessToken, SettingType.ORGANIZATION, orgName)
+        val origOrganization = jsonMapper.readValue(origOrganizationJson, Organization::class.java)
 
-        if (!Scope.isValidScope(scope, origSender)) {
-            echo("Sender full name in scope must match $settingName.  Instead got: $scope")
+        if (!Scope.isValidScope(scope, origOrganization)) {
```
JessicaWNava, line 81: I am a bit confused, if you are ripping out all of the sender to replace it with organization, how are you keeping backwards compatibility?

mkalish: So the backwards compatibility is maintained for where the keys are read from, i.e. a key on a sender can still be used for authenticating. 

The ability to add keys to a sender though is not getting maintained.

jimduff-usds (approved):
My concerns are addressed in the follow-on ticket #8659 

Thanks, @mkalish !

prime-router/src/main/kotlin/cli/SenderUtilsCommand.kt
mkalish: @JessicaWNava Small bump if there was anything else I could clarify/update here

JessicaWNava: Nope, sorry, just got sucked into other things. Thank you for the clarification!

JessicaWNava (approved):
Looks good to me and worked when I tested it!

PR #8714, opened 2023-03-14
Experience/8531/file validator redesign

etanb:
## What's in this PR

This PR covers the redesign and content audit of the File Validation work and is the first step before making the tool public so an end-user doesn't need to be authenticated to check a file.

**To test:**

1. Log in
2. Click "Admin" -> "Validate"
3. Choose a file format
4. Choose a valid file
5. Try a variety of files, perfect ones, some with a few issues and completely invalid ones.

Validation tool Figma: [Link here](https://www.figma.com/file/6mwI5ac6rprACKDzDo4Ady/ReportStream-Workspace-%7C-2023?node-id=2173-22160&t=YvTK54vuIRwufTpm-0)

## The Four Steps of File Validation

screencapture-localhost-3000-file-handler-validate-2023-03-17-17_13_16

_Step One: Choose a schema for validation_

screencapture-localhost-3000-file-handler-validate-2023-03-17-17_17_51

_Step Two: Choose a valid file for validation. An error box will appear upon submittal if there are issues with the file._

screencapture-localhost-3000-file-handler-validate-2023-03-17-17_18_04

_Step Three (Optional): If there are errors or warnings, they will be displayed here along with the ability to download the recommendations in a .csv. If there are ONLY warnings, then a user may proceed to the final step._

screencapture-localhost-3000-file-handler-validate-2023-03-17-17_18_24

_Step Four: Success screen!_

## Video Examples

https://user-images.githubusercontent.com/516877/226073705-8145154d-3193-47ca-be2e-5605da6a012f.mp4

_HL7 Schema Chosen -> Valid file uploaded -> Success screen_

https://user-images.githubusercontent.com/516877/226073708-3860420e-8bba-49ac-b5a6-4b9f6dd64ee2.mp4

_HL7 Schema Chosen -> Valid file with warnings uploaded -> Warning rendered, ignored -> Success screen_


frontend-react/src/hooks/UseFileHandler.ts
penny-lischer, line 1: Can we move the FileType and ContentType enums to TemporarySettingsAPITypes.ts ?
These are properties that I can add to for .pem file upload

frontend-react/src/components/FileHandlers/FileHandler.tsx
```diff
@@ -13,25 +13,26 @@ import { parseCsvForError } from "../../utils/FileUtils";
...
+import { USExtLink } from "../USLink";
 
-import { FileHandlerForm } from "./FileHandlerForm";
-import {
-    RequestLevel,
-    FileQualityFilterDisplay,
-    FileSuccessDisplay,
-    RequestedChangesDisplay,
-} from "./FileHandlerMessaging";
+import { FileHandlerStepTwo } from "./FileHandlerStepTwo";
```
stephenkao, line 25: It'd be better to name these steps something a little more meaningful about their functionality.  `One`, `Two`, `Three`, `Four` don't mean anything without knowing the order as-is right now, and it'd require cascading renames if we, say, added a step in between `Two` and `Three`.

frontend-react/src/components/FileHandlers/FileHandler.tsx
```diff
@@ -91,25 +106,54 @@ export function getClientHeader(
...
-        cancellable,
-        errorType,
         warnings,
         localError,
-        overallStatus,
         reportItems,
         selectedSchemaOption,
     } = state;
+    const [fileContent, setFileContent] = useState("");
+    const [fileHandlerStepIndex, setFileHandlerStepIndex] = useState(0);
```
stephenkao, line 125: The index is something that should be computed rather than set explicitly; explicitly setting the numerical value for the index technically allows for a user to get to a point where we're making incorrect assumptions about their current state.

Basically the index equals (building on assumptions each step of the way):
- 0 if there's no schema set 
- 1 if they have not submitted anew (and they have no file, but that's implied)
- 2 if they have submitted and there are warnings/errors
- 3 if they have submitted and 1) there are warnings and they've confirmed going to the next step OR 2) there are nor warnings/errors

Technically, we could break this out into a function on its own with the props/reducer state passed in as the parameters.  That's kind of what the [reducer in UseFilerHandler.ts](https://github.com/CDCgov/prime-reportstream/blob/master/frontend-react/src/hooks/UseFileHandler.ts#L214) is getting at: setting programmatic steps based on state, and we can just make an inferred value (the index in this case) from the all the state we have in it.  Happy to chat more about this, though!

frontend-react/src/components/FileHandlers/FileHandler.tsx
```diff
@@ -91,25 +106,54 @@ export function getClientHeader(
...
+    // Schema Select, File Select, [optional]Show Errors, Success Page
+    // The stages can be seen here: https://figma.fun/fGCeo4
+    const fileHandlerStep = fileHandlerStepsArray[fileHandlerStepIndex];
+    const resetState = () => {
+        setFileContent("");
+        dispatch({ type: FileHandlerActionType.RESET });
+    };
+
+    const handlePrevFileHandlerStep = () => {
+        // We have to clear different parts of state depending
```
stephenkao, line 136: I'd argue that these should be discrete actions that can be dispatched rather than handling this entirely in an `if-else`.

frontend-react/src/components/FileHandlers/FileHandler.tsx
```diff
@@ -91,25 +106,54 @@ export function getClientHeader(
...
+            resetState();
+            dispatch({
+                type: FileHandlerActionType.SCHEMA_SELECTED,
+                payload: selectedSchemaOption,
+            });
+            setFileHandlerStepIndex(fileHandlerStepIndex - 1);
+        }
+    };
+
+    const handleNextFileHandlerStep = (nextStep = fileHandlerStepIndex + 1) => {
```
stephenkao, line 154: I don't think we should default to proceeding to the next step because each step is based on the assumptions of each prior step.  This is related to [my comment above](https://github.com/CDCgov/prime-reportstream/pull/8714/files#r1137187764) about computing the index rather than setting it explicitly, though.

frontend-react/src/components/FileHandlers/FileHandler.tsx
```diff
@@ -222,49 +261,62 @@ function FileHandler() {
                 },
             });
         }
+        const isFileSuccess =
+            eventData?.overallStatus === OverallStatus.VALID &&
+            !(eventData?.errorCount || eventData?.warningCount);
+        if (isFileSuccess) {
+            handleNextFileHandlerStep(3);
```
stephenkao, line 268: Gonna be a broken record about this, but `3` means nothing to me without reading all of the code here.  I still think we should be inferring the index rather than setting it explicitly, but at the very least, this should be a human-readable/-understandable enum like `FileHandlerStep.WarningsErrorStep` for more clarity.

frontend-react/src/components/FileHandlers/FileHandler.tsx
```diff
@@ -222,49 +261,62 @@ function FileHandler() {
...
+        } else {
+            handleNextFileHandlerStep();
```
stephenkao, line 270: Small/subjective point, but it feels very arbitrary to call this with no argument.

frontend-react/src/components/FileHandlers/FileHandler.tsx
```diff
@@ -13,25 +13,26 @@ import { parseCsvForError } from "../../utils/FileUtils";
...
-    RequestedChangesDisplay,
-} from "./FileHandlerMessaging";
+import { FileHandlerStepTwo } from "./FileHandlerStepTwo";
+import { FileHandlerStepOne } from "./FileHandlerStepOne";
+import { FileHandlerStepThree } from "./FileHandlerStepThree";
+import { FileHandlerStepFour } from "./FileHandlerStepFour";
 
-const FileHandlerSpinner = ({ message }: { message: string }) => (
-    <div className="grid-col flex-1 display-flex flex-column flex-align-center margin-top-4">
+export const FileHandlerSpinner = ({ message }: { message: ReactNode }) => (
```
stephenkao, line 30: Should we just change this to `children` instead?

frontend-react/src/components/FileHandlers/FileHandlerMessaging.tsx
```diff
@@ -136,44 +128,48 @@ export const RequestedChangesDisplay = ({
...
+            <StaticAlert type={alertType} heading={heading} message={message} />
             {showTable && (
                 <>
-                    <h3>{title}</h3>
+                    <div className="display-flex flex-justify">
+                        <h3>{title}</h3>
+                        <Button
+                            className="usa-button flex-align-self-start height-5 margin-top-4 usa-button--outline"
+                            type={"button"}
+                            onClick={() => {
```
stephenkao, line 141: Let's break out the CSV exporting logic into a generic function that can be used anywhere.  We also have custom code [in OrgsTable.tsx](https://github.com/CDCgov/prime-reportstream/blob/master/frontend-react/src/components/Admin/OrgsTable.tsx#L56-L84) that could use this function in the future -- we can add a ticket for that once we have a generic function, though.

frontend-react/src/components/FileHandlers/FileHandlerMessaging.tsx
```diff
@@ -136,44 +128,48 @@ export const RequestedChangesDisplay = ({
...
+                                const options = {
+                                    fieldSeparator: ",",
+                                    quoteStrings: '"',
+                                    decimalSeparator: ".",
+                                    showLabels: true,
+                                    showTitle: true,
+                                    filename: `recommended_${schemaColumnHeader}_edits`,
+                                    title: `Recommended ${schemaColumnHeader} edits`,
+                                    useTextFile: false,
+                                    useBom: true,
```
stephenkao, line 151: 💣

frontend-react/src/components/FileHandlers/FileHandler.tsx
```diff
@@ -276,65 +328,52 @@ function FileHandler() {
...
-                    heading={errorMessaging.heading}
-                />
-            )}
-            {hasQualityFilterMessages && (
-                <FileQualityFilterDisplay
-                    destinations={qualityFilterMessages}
-                    heading=""
-                    message={`The file does not meet the jurisdiction's schema. Please resolve the errors below.`}
+
+            {fileHandlerStep === FileHandlerSteps.STEP_ONE && (
```
stephenkao, line 332: I wonder if we could simplify toggling the currently shown component like (`CurrentStep = FileHandlerStepOne|FileHandlerStepTwo|...`).  Then we could always pass the total set of usable props:

```tsx
<CurrentStep
    fileType={ ... }
    handleNextFileHandlerStep={ ... }
    onSchemaChange={ ... }
    ...otherStuff
/>
```

These steps are mutually exclusive so we'd only have to render one at a time, and it'd be easier to update one place for all that in my opinion 🤷 

OR we could even do something like come up with a common approach to multistep flows like this:
```tsx
function getOrderedSteps(state) {
    return [{
        Component: FileHandlerStepOne,
        // whether or not the user is able to proceed to the next step
        isValid: Boolean(state.selectedSchemaOption)
        // whether or not the step should be rendered
        isShown: !state.selectedSchemaOption,
        // additional props passed to this step's component
        additionalProps: { ... }
    }, 
    ...
    ]
}
```

frontend-react/src/components/FileHandlers/FileHandler.tsx
```diff
@@ -13,25 +13,26 @@ import { parseCsvForError } from "../../utils/FileUtils";
...
-} from "./FileHandlerMessaging";
+import { FileHandlerStepTwo } from "./FileHandlerStepTwo";
+import { FileHandlerStepOne } from "./FileHandlerStepOne";
+import { FileHandlerStepThree } from "./FileHandlerStepThree";
+import { FileHandlerStepFour } from "./FileHandlerStepFour";
 
-const FileHandlerSpinner = ({ message }: { message: string }) => (
-    <div className="grid-col flex-1 display-flex flex-column flex-align-center margin-top-4">
+export const FileHandlerSpinner = ({ message }: { message: ReactNode }) => (
+    <div className="grid-col flex-1 display-flex flex-column flex-align-center margin-top-10">
```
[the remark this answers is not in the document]
etanb: Removed many of the classes 🙏

frontend-react/src/components/DownloadCSVButton.tsx
```diff
@@ -0,0 +1,47 @@
+import { Button, Icon } from "@trussworks/react-uswds";
```
stephenkao, line 1: I don't think this is a useful abstraction: it's not generic enough for any data (data is typed as `ResponseError[]`), and it's passing in the `csvOptions` anyway.  What'd be more useful in my opinion would be to pull the CSV download into its own function that would could hook up to other buttons as needed.  Will update!

frontend-react/src/components/DownloadCSVButton.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+export const DownloadCSVButton = ({
+    secondaryButton,
+    csvOptions,
+    data,
+    buttonText,
+}: DownloadCSVButtonProps) => {
+    return (
+        <Button
+            className={`usa-button flex-align-self-start height-5 margin-top-4 ${
```
stephenkao, line 32: Should use `classnames` since it's available and easier to maintain.

frontend-react/src/components/FileHandlers/FileHandler.tsx
```diff
@@ -223,50 +262,79 @@ function FileHandler() {
...
     }
 
+    const handleOnSchemaChange = (schemaOption: SchemaOption) => {
+        dispatch({
+            type: FileHandlerActionType.SCHEMA_SELECTED,
+            payload: schemaOption,
+        });
+    };
+
+    const siteTooltip = () => (
```
stephenkao, line 299: It's an anti-pattern to define a component inside another component: it creates a new definition _every_ render, and it bypasses any potential memoization and prop type checking.  Will update.

stephenkao:
@penny-lischer @jpandersen87 Mind taking a look at this when you get a chance?  There are still two or three failing tests that I'll look at tomorrow, but it should be just about ready for review.  The most substantial commit is [here](https://github.com/CDCgov/prime-reportstream/pull/8714/commits/696a3ac701a39c1c62b4f24a85962fab2aff3989), where the step logic is made a bit more programmatic and the step components have more encapsulation.  There's still plenty of cleanup that can be done with the existing file handler stuff (including cleaning up/removing `useFileHandler`), but this pull request is already pretty big so I've split that out into a separate ticket.

frontend-react/src/components/FileHandlers/FileHandlerErrorsWarningsStep.tsx
```diff
@@ -68,23 +88,27 @@ export const FileHandlerErrorsWarningsStep = ({
                     message={`The file does not meet the jurisdiction's schema. Please resolve the errors below.`}
```
penny-lischer, line 91: This can be a string.
"The file does not meet the jurisdiction's schema. Please resolve the errors below."

frontend-react/src/components/FileHandlers/FileHandlerMessaging.test.tsx
```diff
@@ -20,48 +19,6 @@ import {
 // Note: following a pattern of finding elements by text (often text passed as props)
```
penny-lischer, line 19: Can we remove this comment?

stephenkao:
Note: I've been keeping this up-to-date with merge commits since there are multiple contributors to this branch so we don't accidentally drop commits in the conflict resolution.  We can squash when we're merging back into master, though!

stephenkao:
For transparency:

@etanb and I just caught up, and he's going to take this pull request over again.  Other than current (and future?) review comments, I don't thiiiiiink there's anything else that needs to be worked on 🙏

frontend-react/src/components/FileHandlers/FileHandlerMessaging.test.tsx
etanb: Yes!

etanb:
Hmmmm I'm noticing if a CSV has an `error` the entire file doesn't get validated and kind of blows up:

Screenshot 2023-03-28 at 5 01 53 PM

You can see above the response in the console has valid errors, but it never gets to that page. Will investigate.

etanb:
I remember us talking in some conversation that the `"httpStatus" : 400,` for `errors` should still be `200`. Let me know if I'm missing something @stephenkao !

stephenkao:
> I remember us talking in some conversation that the `"httpStatus" : 400,` for `errors` should still be `200` since currently in my local dev I see it as `400` which is why it's hitting that `catch` block. Let me know if I'm missing something @stephenkao !

@etanb The [endpoint update](https://github.com/CDCgov/prime-reportstream/pull/8893) has been merged!

stephenkao:
Just resolved merge conflicts, and it's looking good to me :) 

Screenshot 2023-03-30 at 14 55 31

Screenshot 2023-03-30 at 15 03 37

stephenkao (approved):

penny-lischer:
@etanb Scenario - 
1. Select a file(submit button becomes enabled)
2. Change my mind
3. Open the file modal
4. Click cancel
Should the Submit button go from enabled to disabled?

penny-lischer (approved):
One comment/scenario that can be tackled in another ticket, otherwise LGTM!

etanb:
Thanks for finding that @penny-lischer . I've checked the issue, and it's a USWDS level problem, where upon cancel, they clear the file input for some reason. That should either be removed behavior or optional. Filed a github issue with them here: https://github.com/uswds/uswds/issues/5212

My interim solution is just to reset state on cancel.

Issue #8723, opened 2023-03-15
bug: Merging of ConfigSchemas only handles merging of elements, but not top level properties in the schemas.


jimmyfagan:
## Describe the bug
When a schema extends another schema, the expectation is that constants from both schemas should exist in the results. Instead the schema being extended covers up the constants on the child schemas. Additionally, the name of the resulting merged schema is that of the parent schema instead of that of the child.

This issue will involve writing unit tests to test this case, and to fix the code in the ConfigSchema/ConfigSchemaReader to allow the case to work as expected.

## Impact on ReportStream
- When working on converter schemas for California or for parity with COVID pipeline, it is difficult to utilize constants without adding them to the ORU R01 base schema.

## Steps to reproduce
1. Run ```fhirdata --input-file src/testIntegration/resources/datatests/FHIR_to_HL7/sample_SR_1_20230302-0001.fhir -s <some schema>.yml --output-format HL7 --output-file src/testIntegration/resources/datatests/FHIR_to_HL7/sample_SR_1_20230302-0001.hl7``` using a yml file that extends another yml file with constants defined at the top level of both and using those constants as part of the conversion.
2. Observe the results to see that the constant values from the child schema were not used.

## Expected behavior
Constants from both the base schema and the child schema are used.

PR #8730, opened 2023-03-15
8723: Replace schema/element merge with extend and add tests

jimmyfagan:
This PR changes the approach to merging schemas (combining a schema with the schema specified by the `extends` property). Whereas before, we had a `merge` method for a parent schema to be combined with its child schema, we now have a `extend` method, which allows a child schema to extend a parent schema. This allows for more seamless preservation of a child schema's contents like it's name, and also makes it conceptually clearer how this merge happens. The parent schema is unaffected by the `extend`, but the child schema will have parent schema elements and constants added to it.

As part of this, I added some testing which exercises the new code, and I adjusted existing tests which relied on the `merge` method.

Test Steps:

- Run fhirdata --input-file src/testIntegration/resources/datatests/FHIR_to_HL7/sample_SR_1_20230302-0001.fhir -s <some schema>.yml --output-format HL7 --output-file src/testIntegration/resources/datatests/FHIR_to_HL7/sample_SR_1_20230302-0001.hl7 using a yml file that extends another yml file with constants defined at the top level of both and using those constants as part of the conversion.
- Observe the results to see that the constant values from both the child schema and the parent schema were used.

or

- Run test `test read from file with extends`

## Changes
- Replace `merge` schema with `extend` schema in `ConfigSchema.kt` `ConverterSchema.kt` and `FhirTransformSchema.kt`
- Replace `merge` element with `extend` element in `ConfigSchema.kt` `ConverterSchema.kt` and `FhirTransformSchema.kt`
- Adjust `ConfigSchemaReader.kt` to use `extend` instead of `merge`
- Add to `ConfigSchemaReaderTests.kt` `ConverterSchemaTests.kt` and `FhirTransformSchemaTests.kt` to exercise new code paths and to more closely verify behavior of inherited constants.

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- Addresses #8723


jimmyfagan:
Closing this in favor of #8741

Issue #8398, opened 2023-02-17
[Manage Public Key] Upload Key and Fail/Success message


stephenkao:
## Problem statement
The sender emails us two separate public keys. These keys are stored in the database under sender settings. Then we send them an email that they can start testing (authentication) and includes the API endpoint to point to and link them to API Programmer's guide.
This happens on almost every new sender and is a manual process.

## What you need to know

https://www.figma.com/file/6mwI5ac6rprACKDzDo4Ady/ReportStream-Workspace-%7C-2023?node-id=2167-21742&t=hrkPPWw0VtqV0v9V-0
API - https://github.com/CDCgov/prime-reportstream/pull/8678

## Acceptance criteria
- [x]  Senders key saves on Submit 
- [x]  Display success message when successful submit of a senders public key
- [x]  Display fail message when error of a senders public key
- [x]  Try again button returns the user back to the Manage Public Key page
- [ ] QA with Design/Content before merging

## To do

- [x]  Add endpoint hook and save Senders key
- [x]  Display success message when successful submit of a senders public key
- [x]  Display fail message and 'Try again' button when error of a senders public key
- [x]  Loading step while request is pending
- [x] Unit tests

Andrey-Glazkv:
@stephenkao and @jpandersen87 to review it EoD 4/18

laurenwins1:
ghp_ X8wYT2ZsNhemb9D4FyDyV5hlmxZUw72FtrGi

PR #8731, opened 2023-03-15
8396- Public key tool select sender and upload key component

### LINKED ISSUE #8396: [Manage Public Key] Select a Sender and upload key
## Problem statement
The sender emails us two separate public keys. These keys are stored in the database under sender settings. Then we send them an email that they can start testing (authentication) and includes the API endpoint to point to and link them to API Programmer's guide.
This happens on almost every new sender and is a manual process.

## What you need to know
https://www.figma.com/file/6mwI5ac6rprACKDzDo4Ady/ReportStream-Workspace-%7C-2023?node-id=2167-21742&t=hrkPPWw0VtqV0v9V-0
API - https://github.com/cdcgov/prime-reportstream/issues/8472
Assumes complete - [https://app.zenhub.com/workspaces/web-experience-607d9d5e68b95200150fec37/issues/gh/cdcgov/prime-reportstream/8395]()

## Acceptance criteria

Select a Sender step:
- [x] There's helper text about the API Programmer's Guide
- [x] There's a dropdown that lists Sender
- [x] When Sender is selected, display upload step
- [x] If < 2 Senders, skip this step and go directly to upload step

--- 

Upload step:
- [x] There's helper text about the API Programmer's Guide
- [x] There's a file input that allows a user to upload a .pem (NOTE: may want to support .pub file)
- [x] Submit button is disabled until file is selected

## To do

- [x] Add endpoint hook to fetch sender(s)
- [x] Unit tests

penny-lischer:
This PR ...

Selects a Sender and displays the upload key component.

Test Steps:
1. Select an organization with one or more senders.If your org only has 1 sender you will see the upload key component. If your org has >1  sender you will have the option to select a sender before seeing the upload key component.

## Changes

- There's helper text about the API Programmer's Guide
-  There's a dropdown that lists Sender
-  When Sender is selected, display upload step
-  If < 2 Senders, skip this step and go directly to upload step
- There's helper text about the API Programmer's Guide
-  There's a file input that allows a user to upload a .pem
-  Submit button is disabled until file is selected

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- Fixes #8396


frontend-react/src/components/ManagePublicKey/ManagePublicKey.tsx
```diff
@@ -0,0 +1,207 @@
...
+    useEffect(() => {
+        if (localError) {
+            showError(localError);
+        }
+    }, [localError]);
+
+    return (
+        <div className="grid-col-12">
+            {senders?.length > 1 && !showUploadDisplay && (
+                <Form name="senderSelect" onSubmit={() => handleSenderSubmit()}>
```
penny-lischer, line 129: Is using Form overkill here?

stephenkao: I think it's the correct use of Form since this is used for data submission :)

frontend-react/src/hooks/network/ManagePublicKey/ManagePublicKeyHooks.ts
```diff
@@ -0,0 +1,183 @@
...
+    state: ManagePublicKeyState,
+    payload: FileSelectedPayload
+): Partial<ManagePublicKeyState> {
+    const { file } = payload;
+    try {
+        // look at the filename extension.
+        const fileNameArray = file.name.split(".");
+        const uploadFileExtension = fileNameArray[fileNameArray.length - 1];
+
+        if (uploadFileExtension.toUpperCase() !== Format.PEM) {
```
stephenkao, line 89: Can this not be done with `accept`?

frontend-react/src/hooks/network/ManagePublicKey/ManagePublicKeyHooks.ts
```diff
@@ -0,0 +1,183 @@
...
+            return {
+                ...state,
+                localError: "The file type must be of type .pem",
+            };
+        }
+
+        if (file.size > PAYLOAD_MAX_BYTES) {
```
stephenkao, line 96: Might be worth breaking out the logic to check the file size into a separate function.  OR we could create a custom Form component that would do this automatically on file change.

frontend-react/src/components/ManagePublicKey/ManagePublicKey.tsx
```diff
@@ -1,18 +1,202 @@
...
+    ManagePublicKeyActionType,
+} from "../../hooks/network/ManagePublicKey/ManagePublicKeyHooks";
+import { ContentType } from "../../utils/TemporarySettingsAPITypes";
+
+import { ManagePublicKeyUpload } from "./ManagePublicKeyUpload";
+
+const LightbulbIcon = Icon.Lightbulb;
+
+// TODO: create common component???
+const ManagePublicKeySpinner = ({ message }: { message: ReactNode }) => (
```
jpandersen87, line 31: Wouldn't hurt to break this out into its own module. I have a ticket to replace our usage of the third-party library spinner with our own (based off the spinner svg in one of USWD's modules) so having a local spinner module in place and in-use would help with the eventual swap. A local Spinner component that can take custom className and an optional message prop.

frontend-react/src/hooks/network/ManagePublicKey/ManagePublicKeyHooks.ts
```diff
@@ -0,0 +1,183 @@
+import { useReducer } from "react";
```
stephenkao, line 1: Hmmm I don't think we need to follow the pattern in UseFileHandler.ts, and I'd argue that the logic there is a bit more complicated than it needs to be.  We're essentially working in a state machine in moving between the four "steps":
- Choose Sender
- Choose file
- Loading
- Result
    - Success (END)
    - Error (can reset to 'Choose file' step)

Each of these steps can be defined by inferred state:
- No Sender chosen --> render 'Choose Sender' step
- No file chosen + not yet submitted --> render 'Choose file' step
- File submitted and pending response --> render 'Loading' step
- Result
    - File submitted + successful response --> render 'Success' step
    - File submitted + errored response --> render 'Error' step

Rather than dispatching individual actions, we can derive this data from the actual form state + query state itself.

frontend-react/src/hooks/network/ManagePublicKey/ManagePublicKeyHooks.ts
```diff
@@ -0,0 +1,183 @@
...
+function calculateRequestCompleteState(
+    state: ManagePublicKeyState,
+    payload: RequestCompletePayload
+): Partial<ManagePublicKeyState> {
+    const {
+        response: { id, timestamp },
+    } = payload;
+
+    return {
+        fileInputResetValue: state.fileInputResetValue + 1,
```
stephenkao, line 134: I understand the reasoning for this in the existing FileHandler, but my feeling is that it's an anti-pattern since we shouldn't have to force a re-rendering by keeping track of a separate key.  If anything, it should just be a derived value from the data we'd actually want to submit.

frontend-react/src/hooks/network/ManagePublicKey/ManagePublicKeyHooks.ts
jpandersen87: I would greatly advise not using UseFileHandler as a template for derived work. I feel that hook is in need of review in the future.

frontend-react/src/components/ManagePublicKey/ManagePublicKey.tsx
stephenkao: Yeah absolutely agree -- no reason not to have a single, generalized component for all spinners!

penny-lischer: Will do. Validator can also use this common component.

frontend-react/src/components/ManagePublicKey/ManagePublicKey.tsx
```diff
@@ -1,18 +1,202 @@
...
+
+        // initializes necessary state and sets `isSubmitting`
+        dispatch({ type: ManagePublicKeyActionType.PREPARE_FOR_REQUEST });
+
+        try {
+            sendFile({
+                contentType: ContentType.PEM,
+                fileContent: fileContent,
+                fileName: fileName,
+                selectedSender: selectedSender,
```
penny-lischer, line 105: Just learned that we will not need to send the selectedSender to the POST for this iteration.

frontend-react/src/styles/_components.scss
```diff
@@ -79,4 +79,11 @@ hr {
         border-color: color('primary');
         border-width: medium;
     }
+}
+
+.manage-public-key {
+    .usa-icon {
+        height: 1.25em;
```
stephenkao, line 86: I recommend that we use a consistent unit for sizes/spacings.  USWDS uses 8px as [the atomic size for spacing](https://designsystem.digital.gov/design-tokens/spacing-units/), so we could use pixels for more explicit styling OR use rems if needed.

frontend-react/src/hooks/UseValueSets.ts
```diff
@@ -45,7 +45,7 @@ const findTableMetaByName = (
 
   useValueSetsTable
 
-  a useQuery based custom hook used to get value sets and value set rows (defined by passsed dataTableName)
+  a useQuery based custom hook used to get value sets and value set rows (defined by passed dataTableName)
```
penny-lischer, line 48: Spelling errors kill me softly!

frontend-react/src/utils/TemporarySettingsAPITypes.ts
```diff
@@ -12,6 +12,13 @@ enum Format {
     CSV = "CSV",
     HL7 = "HL7",
     FHIR = "FHIR",
+    PEM = "PEM",
```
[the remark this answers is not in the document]
penny-lischer: Will fix

frontend-react/src/components/ManagePublicKey/ManagePublicKey.test.tsx
```diff
@@ -0,0 +1,87 @@
...
+            useOrganizationSendersExports,
+            "UseOrganizationSenders"
+        ).mockReturnValue({
+            isLoading: false,
+            senders: DEFAULT_SENDERS,
+            ...result,
+        });
+    }
+
+    describe("when the sender options are loading", () => {
```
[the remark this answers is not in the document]
stephenkao: It's also not entirely important to check each individual step's contents.  Since ManagePublicKey is the "manager" component of all its children, you could even do something like:

```tsx
jest.mock("./ManagePublicKeyUpload", () => {
    ManagePublicKeyUpload: () => <div data-testid="ManagePublicKeyUpload" />
});

// then just verify that the test id exists in the dom
expect(screen.getByTestId("ManagePublicKeyUpload")).toBeVisible();
```

Each individual step component should have its own accompanying tests so it's irrelevant for the manager component to test that as well specifically for the state --> step checks.  However, mocking individual components like this would prevent testing the entire flow through an integration test so that's something to consider.

frontend-react/src/components/ManagePublicKey/ManagePublicKeyChooseSender.tsx
```diff
@@ -0,0 +1,59 @@
+import React, { useState } from "react";
+import { Form, Dropdown, Button, FormGroup } from "@trussworks/react-uswds";
+
+import { UseOrganizationSenders } from "../../hooks/UseOrganizationSenders";
+import Spinner from "../Spinner";
+
+type Props = {
```
stephenkao, line 7: Let's name prop types after their components, which helps to clarify if we need to import them elsewhere: `ManagePublicKeyChooseSenderProps`.

frontend-react/src/components/ManagePublicKey/ManagePublicKeyChooseSender.tsx
```diff
@@ -0,0 +1,59 @@
...
+    onSenderSelect: any; //What type should this be???
```
stephenkao, line 8: Likely something like `(sender: string) => void` unless we want to explicitly type the Sender value.

frontend-react/src/components/ManagePublicKey/ManagePublicKeyChooseSender.tsx
```diff
@@ -0,0 +1,59 @@
...
+};
+
+export function ManagePublicKeyChooseSender({ onSenderSelect }: Props) {
```
stephenkao, line 11: One pattern I'd like us to use is exporting default components when it makes sense:

```tsx
export default function ManagePublicKeyChooseSender(...) { ... }
```

This pattern encourages that we orient each file around a _single_ component to locate them more easily and cut down on number of lines in a file (which gets unwieldy after a certain threshold).

frontend-react/src/components/ManagePublicKey/ManagePublicKey.test.tsx
```diff
@@ -0,0 +1,87 @@
...
+        beforeEach(() => {
+            mockUseOrganizationSenders({ isLoading: true });
+
+            renderApp(<ManagePublicKey />);
+        });
+
+        test("renders a spinner", () => {
```
stephenkao, line 64: These tests seem like they're testing ManagePublicKeyChooseSender rather than ManagePublicKey, no?

frontend-react/src/components/ManagePublicKey/ManagePublicKey.tsx
```diff
@@ -1,18 +1,148 @@
...
+    ManagePublicKeyAction,
+    CONTENT_TYPE,
+    FORMAT,
+    validateFileSize,
+} from "../../hooks/network/ManagePublicKey/ManagePublicKeyHooks";
+
+import { ManagePublicKeyChooseSender } from "./ManagePublicKeyChooseSender";
+import { ManagePublicKeyUpload } from "./ManagePublicKeyUpload";
+
+const LightbulbIcon = Icon.Lightbulb;
```
stephenkao, line 22: Why is this alias needed?

frontend-react/src/components/ManagePublicKey/ManagePublicKey.tsx
```diff
@@ -1,18 +1,148 @@
...
+
+const ManagePublicKeySwitchDisplay = () => {
+    const navigate = useNavigate();
+    const [action, setAction] = useState(ManagePublicKeyAction.SELECT_SENDER);
```
stephenkao, line 26: I think it's fine to differentiate between steps with explicit step types like this, but I don't think we should be setting them directly in state.  Why not just infer that from whatever state we have stored already?

frontend-react/src/components/ManagePublicKey/ManagePublicKey.tsx
```diff
@@ -1,18 +1,148 @@
...
+    const [fileContent, setFileContent] = useState("");
+    const [fileName, setFileName] = useState("");
+
+    // TODO: mocked for now - make the call you need when sending the file
```
stephenkao, line 30: Can we add a noop/mocked query hook file in the meantime?

frontend-react/src/components/ManagePublicKey/ManagePublicKey.tsx
```diff
@@ -1,18 +1,148 @@
...
+    const { sendFile } = {
+        sendFile: (data: {
```
stephenkao, line 32: I realize this is a mock, but this should return a Promise.

frontend-react/src/components/ManagePublicKey/ManagePublicKey.tsx
```diff
@@ -1,18 +1,148 @@
...
+            fileContent?: string;
+            fileName?: string;
+        }) => {
+            data = { fileName: "filename" };
+            return data;
+        },
+    };
+
+    const onSenderSelect = (selectedSender: string) => {
+        if (selectedSender !== "") {
```
stephenkao, line 43: Can this just check truthiness?  `if (selectedSender)`?

frontend-react/src/components/ManagePublicKey/ManagePublicKey.tsx
```diff
@@ -1,18 +1,148 @@
...
+            setAction(ManagePublicKeyAction.SELECT_PUBLIC_KEY);
```
stephenkao, line 44: Shouldn't we be setting the selected Sender here?  Additionally, this is kind of what I mean with avoiding explicitly setting the action in favor of inferring from state -- we can infer that we're past the Sender selection step if there's already a selected Sender so we don't need to set an action like this.  Setting an action explicitly makes it easier to violate state assumptions, which could lead to errors in the flow.

frontend-react/src/components/ManagePublicKey/ManagePublicKey.tsx
```diff
@@ -1,18 +1,148 @@
...
+        }
+    };
+
+    const onBack = () => {
+        navigate("/resources");
```
stephenkao, line 49: I don't think we'll always want to go back to /resources from this page, right?  Shouldn't this just be literally going back in navigation history?

frontend-react/src/components/ManagePublicKey/ManagePublicKey.tsx
```diff
@@ -1,18 +1,148 @@
...
+    ) => {
+        event.preventDefault();
+
+        if (fileContent.length === 0) {
+            showError("No file contents to validate.");
+            return;
+        }
+
+        try {
+            setAction(ManagePublicKeyAction.SAVE_PUBLIC_KEY);
```
stephenkao, line 63: Why do we need this differentiated step?

frontend-react/src/components/ManagePublicKey/ManagePublicKey.tsx
```diff
@@ -1,18 +1,148 @@
...
+    };
+
+    const onBack = () => {
+        navigate("/resources");
+    };
+
+    const onPublicKeySubmit = async (
+        event: React.FormEvent<HTMLFormElement>
+    ) => {
+        event.preventDefault();
```
stephenkao, line 55: Do we need to prevent default here?

frontend-react/src/components/ManagePublicKey/ManagePublicKey.tsx
```diff
@@ -1,18 +1,148 @@
...
+    FORMAT,
+    validateFileSize,
+} from "../../hooks/network/ManagePublicKey/ManagePublicKeyHooks";
+
+import { ManagePublicKeyChooseSender } from "./ManagePublicKeyChooseSender";
+import { ManagePublicKeyUpload } from "./ManagePublicKeyUpload";
+
+const LightbulbIcon = Icon.Lightbulb;
+
+const ManagePublicKeySwitchDisplay = () => {
```
stephenkao, line 24: Can we prefer to use [function declarations](https://github.com/CDCgov/prime-reportstream/blob/master/frontend-react/docs/best-practices.md#prefer-using-function-declarations-over-function-expressions-for-first-class-functions-and-components) at least for components?

stephenkao, line 24: Minor point, but I don't see much of a benefit of having ManagePublicKeySwitchDisplay being broken out of the ManagePublicKey component.  I feel like it can all be inside ManagePublicKey 🤷

frontend-react/src/components/ManagePublicKey/ManagePublicKey.tsx
```diff
@@ -1,18 +1,148 @@
...
+        } catch (e: any) {
+            console.trace(e);
+            showError(`Uploading public key failed. ${e.toString()}`);
+            setAction(ManagePublicKeyAction.SELECT_PUBLIC_KEY);
+        }
+    };
+
+    const onFileChange = async (event: React.ChangeEvent<HTMLInputElement>) => {
+        // No file selected
+        if (!event?.target?.files?.length) return;
```
stephenkao, line 80: Do we need to check against `files.length` and `files.item` for file falsiness?

frontend-react/src/components/ManagePublicKey/ManagePublicKeyChooseSender.tsx
```diff
@@ -0,0 +1,59 @@
+import React, { useState } from "react";
+import { Form, Dropdown, Button, FormGroup } from "@trussworks/react-uswds";
+
+import { UseOrganizationSenders } from "../../hooks/UseOrganizationSenders";
```
stephenkao, line 4: SMALL point, but let's use lowercase letters when referring to non-component functions: `import { useOrganizationSenders }`.

frontend-react/src/components/ManagePublicKey/ManagePublicKeyChooseSender.tsx
```diff
@@ -0,0 +1,59 @@
...
+        onSenderSelect(selectedSender);
+    }
+
+    function handleSubmit() {
+        onSenderSelect(selectedSender);
+    }
+
+    return (
+        <>
+            {isLoading && <Spinner message="Processing file..." />}
```
stephenkao, line 27: This message isn't correct, right?  This is tied to loading Senders, not processing the file.

frontend-react/src/components/ManagePublicKey/ManagePublicKeyChooseSender.tsx
```diff
@@ -0,0 +1,59 @@
...
+            {!isLoading && senders && (
+                <Form name="senderSelect" onSubmit={handleSubmit}>
```
stephenkao, line 29: Let's be consistent with our names.  We're using a pattern of lowercase and hyphenated names already: `sender-select`

frontend-react/src/hooks/network/ManagePublicKey/ManagePublicKeyHooks.ts
```diff
@@ -0,0 +1,64 @@
+import {
```
stephenkao, line 1: There are no hooks in the file so the name is misleading.  Let's update that!

frontend-react/src/components/ManagePublicKey/ManagePublicKeyChooseSender.tsx
```diff
@@ -0,0 +1,59 @@
...
+                                <option key={name} value={name}>
+                                    {name}
+                                </option>
+                            ))}
+                        </Dropdown>
+                        <Button
+                            key="submit-sender"
+                            type="submit"
+                            outline
+                            className="padding-bottom-1 padding-top-1"
```
stephenkao, line 49: `padding-y-1`?

frontend-react/src/components/ManagePublicKey/ManagePublicKeyChooseSender.tsx
```diff
@@ -0,0 +1,59 @@
...
+    onSenderSelect: any; //What type should this be???
+};
+
+export function ManagePublicKeyChooseSender({ onSenderSelect }: Props) {
+    const { isLoading, senders } = UseOrganizationSenders();
+    const [selectedSender, setSelectedSender] = useState(
+        senders?.length === 1 ? senders[0].name : ""
+    );
+
+    if (senders?.length === 1) {
```
stephenkao, line 17: This means that `onSenderSelect` will fire every time this component renders, which can happen A LOT based on state updates.  At the very least, it should be moved into an effect:

```ts
useEffect(() => {
    if (senders.length === 1) {
        onSenderSelect(selectedSender);
    }
}, [senders.length]); // just using senders.length as a primitive comparison value, not very intelligent
```

However, this should probably be done earlier on so this step can be skipped over entirely in the first place 🤔

frontend-react/src/components/ManagePublicKey/ManagePublicKeyChooseSender.tsx
stephenkao, line 29: Should we wrap _all_ steps in a single Form?

frontend-react/src/components/ManagePublicKey/ManagePublicKeyUpload.tsx
```diff
@@ -0,0 +1,73 @@
...
+} from "@trussworks/react-uswds";
+
+export interface ManagePublicKeyFormProps {
+    onPublicKeySubmit: (e: React.FormEvent<HTMLFormElement>) => void;
+    onFileChange: (e: React.ChangeEvent<HTMLInputElement>) => void;
+    onBack: () => void;
+    fileName: string;
+}
+
+const BASE_ACCEPT_VALUE = ".pem";
```
stephenkao, line 18: I think this can just be inlined below since it's such a simple value.

frontend-react/src/components/ManagePublicKey/ManagePublicKeyUpload.tsx
```diff
@@ -0,0 +1,73 @@
...
+import {
+    Button,
+    FileInput,
+    FileInputRef,
+    Form,
+    FormGroup,
+    Label,
+} from "@trussworks/react-uswds";
+
+export interface ManagePublicKeyFormProps {
```
stephenkao, line 11: It looks like we're vacillating between types and interfaces for props.  Let's choose one (I vote interfaces for props) and stick with that structure.

frontend-react/src/components/ManagePublicKey/ManagePublicKeyUpload.tsx
```diff
@@ -0,0 +1,73 @@
...
+    onPublicKeySubmit: (e: React.FormEvent<HTMLFormElement>) => void;
+    onFileChange: (e: React.ChangeEvent<HTMLInputElement>) => void;
+    onBack: () => void;
+    fileName: string;
+}
+
+const BASE_ACCEPT_VALUE = ".pem";
+
+export const ManagePublicKeyUpload = ({
```
stephenkao, line 20: `default`?

frontend-react/src/components/ManagePublicKey/ManagePublicKeyUpload.tsx
```diff
@@ -0,0 +1,73 @@
...
+                    data-testid="upload-pem-input"
+                    onChange={(e) => onFileChange(e)}
+                    required
+                    ref={fileInputRef}
+                    accept={BASE_ACCEPT_VALUE}
+                />
+            </FormGroup>
+            <div className="grid-row">
+                <div className="grid-col flex-1 display-flex flex-column flex-align-start">
+                    <div className="grid-col flex-1 display-flex flex-row flex-align-start">
```
stephenkao, line 61: Duplicate div?

frontend-react/src/components/ManagePublicKey/ManagePublicKeyUpload.tsx
```diff
@@ -0,0 +1,73 @@
...
+                        <Button onClick={onBack} type="button" outline>
+                            Back
+                        </Button>
+                        <Button disabled={isDisabled} type={"submit"}>
```
stephenkao, line 65: Don't need {} for string props

frontend-react/src/components/ManagePublicKey/ManagePublicKeyUpload.tsx
```diff
@@ -0,0 +1,73 @@
...
+    onFileChange,
+    onBack,
+    fileName,
+}: ManagePublicKeyFormProps) => {
+    const isDisabled = fileName.length === 0;
+    const fileInputRef = useRef<FileInputRef>(null);
+
+    return (
+        <Form
+            name="managePublicKey"
```
stephenkao, line 31: Let's use hyphenated names.

frontend-react/src/styles/_components.scss
```diff
@@ -79,4 +79,11 @@ hr {
         border-color: color('primary');
         border-width: medium;
     }
+}
+
+.manage-public-key {
+    .usa-icon {
```
stephenkao, line 85: Can we use [size](https://github.com/trussworks/react-uswds/blob/main/src/components/Icon/Icon.tsx#L7) instead of explicitly setting the viewbox dimensions?

frontend-react/src/hooks/UseOrganizationSenders.ts
```diff
@@ -0,0 +1,37 @@
...
+import { useSessionContext } from "../contexts/SessionContext";
+
+const { senders } = servicesEndpoints;
+
+export type UseOrganizationSendersResult = {
+    senders: RSSender[];
+    isLoading: boolean;
+};
+
+export const UseOrganizationSenders = () => {
```
stephenkao, line 14: `useOrganizationSenders`

frontend-react/src/components/Spinner.tsx
```diff
@@ -1,32 +1,41 @@
...
+                    height={60}
+                    width={60}
+                    color="#2378c3"
+                    secondaryColor="#aacdec"
+                    strokeWidth={8}
+                    strokeWidthSecondary={8}
+                    ariaLabel="loading-indicator"
+                />
+            </div>
+            <span className="text-center">{message}</span>
```
stephenkao, line 38: This is optional so we should make it conditionally rendered:

```tsx
{ message && <span className="text-center">{message}</span> }
```

stephenkao: Would also be worth adding a test for this?

frontend-react/src/components/ManagePublicKey/ManagePublicKeyUpload.tsx
```diff
@@ -0,0 +1,73 @@
...
+            className="rs-full-width-form"
+        >
+            <FormGroup className="margin-bottom-3">
+                <Label
+                    className="font-sans-xs"
+                    id="upload-pem-input-label"
+                    htmlFor="upload-pem-input"
+                >
+                    Upload public key
+                    <br />
```
stephenkao, line 42: Let's not use `<br>`s because they're literally only used for line breaks, which we can get with block-level elements.  Since this is wrapped in a `<label>`, though, we need to use `<span>`s instead of, say, `<p>`s:

```tsx
<Label>
    <span className="display-block">Upload public key</span>
    <span className="display-block text-gray-50">...</span>
</Label>
```

frontend-react/src/components/ManagePublicKey/ManagePublicKeyUpload.tsx
```diff
@@ -0,0 +1,73 @@
...
+                    name="upload-pem-input"
+                    aria-describedby="upload-pem-input-label"
+                    data-testid="upload-pem-input"
+                    onChange={(e) => onFileChange(e)}
+                    required
+                    ref={fileInputRef}
+                    accept={BASE_ACCEPT_VALUE}
+                />
+            </FormGroup>
+            <div className="grid-row">
```
stephenkao, line 59: Looks like we can use `<Grid row`> and `<Grid col="auto">`?  https://trussworks.github.io/react-uswds/?path=/docs/components-grid--column-spans

frontend-react/src/components/ManagePublicKey/ManagePublicKeyUpload.tsx
```diff
@@ -0,0 +1,73 @@
...
+    fileName: string;
+}
+
+const BASE_ACCEPT_VALUE = ".pem";
+
+export const ManagePublicKeyUpload = ({
+    onPublicKeySubmit,
+    onFileChange,
+    onBack,
+    fileName,
```
stephenkao, line 24: Shouldn't this receive a File instance instead?

frontend-react/src/components/ManagePublicKey/ManagePublicKey.test.tsx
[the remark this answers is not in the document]
penny-lischer: Definitely was intended to add more test cases. Was waiting on feedback before proceeding.

frontend-react/src/components/ManagePublicKey/ManagePublicKeyChooseSender.tsx
penny-lischer: Agree, my bad.

frontend-react/src/components/ManagePublicKey/ManagePublicKey.tsx
penny-lischer: Not needed just seemed cleaner.

frontend-react/src/components/ManagePublicKey/ManagePublicKey.tsx
penny-lischer: Yup, I was not putting effort into this since the next story will cover this.

frontend-react/src/components/ManagePublicKey/ManagePublicKeyChooseSender.tsx
penny-lischer: I originally had the logic in ManagePublicKey prior to breaking this out. Should I move it back there and pass the senders to ManagePublicKeyChooseSender?

frontend-react/src/components/ManagePublicKey/ManagePublicKeyUpload.tsx
penny-lischer: Merge!!!

frontend-react/src/components/ManagePublicKey/ManagePublicKey.tsx
penny-lischer: No, but if we don't have this `if (!file) return; ` check then typescript complains that file maybe null.

frontend-react/src/components/ManagePublicKey/ManagePublicKey.test.tsx
penny-lischer: Will fix

frontend-react/src/components/ManagePublicKey/ManagePublicKeyUpload.tsx
penny-lischer: Will update

frontend-react/src/components/ManagePublicKey/ManagePublicKeyChooseSender.tsx
penny-lischer: Actually this component/logic will only get called initially if org has one sender and then onSenderSelect is a vallback to the parent to set the selectedSender.
```
if (senders.length === 1) {
        onSenderSelect(selectedSender);
    }
```

frontend-react/src/hooks/network/ManagePublicKey/ManagePublicKeyHooks.ts
penny-lischer: Refactored per comment

frontend-react/src/components/Spinner.tsx
penny-lischer: Done

frontend-react/src/components/ManagePublicKey/ManagePublicKey.test.tsx
```diff
@@ -0,0 +1,126 @@
...
+                    senders: DEFAULT_SENDERS,
+                });
+
+                renderApp(<ManagePublicKey />);
+            });
+
+            test("renders ManagePublicKeyChooseSender", () => {
+                expect(
+                    screen.getByText(/Manage Public Key/)
+                ).toBeInTheDocument();
```
stephenkao, line 65: Let's stick to `toBeVisible` over `toBeInTheDocument` since the former is stricter about actual visibility (well, as much as style-less, in-memory rendering can be).

frontend-react/src/components/ManagePublicKey/ManagePublicKey.test.tsx
```diff
@@ -0,0 +1,126 @@
...
+                    ).toBeVisible();
+                });
+            });
+        });
+
+        describe("when only one sender", () => {
+            beforeEach(() => {
+                mockUseOrganizationSenders({
+                    isLoading: false,
+                    senders: DEFAULT_SENDERS.splice(0, 1),
```
stephenkao, line 107: Minor minor point, but we could just do `[DEFAULT_SENDERS[0]]`.

frontend-react/src/components/ManagePublicKey/ManagePublicKey.tsx
```diff
@@ -1,18 +1,136 @@
...
+            contentType?: string;
+            fileContent?: string;
+            file?: File | null;
+        }) => {
+            data = { file: null };
+            return data;
+        },
+    };
+
+    const onSenderSelect = (selectedSender: string) => {
```
stephenkao, line 37: Minor point, but unless we need to reference callbacks in multiple places, they can just be defined inline like:

```tsx
onSenderSelect={ (selectedSender: string) => setSender(selectedSender) }
```

stephenkao, line 37: Also, I think it helps to differentiate the naming convention to know whether or not a function is defined within the current component or if it's coming down as a prop.  I tend to use `handle<object><action>` for functions defined within components and `on<object><action>` for names of props:

```tsx
// in ManagePublicKey

// defining in component: use handle*
function handleSenderSelect(selectedSender: string) {
    setSender(selectedSender);
}

// passing as a prop: use on*
onSenderSelect={handleSenderSelect}
```

This helps make it clearer when we're invoking callbacks passed from above as props too.  Like if there were _another_ callback we had to call:

```tsx
// if we want to trigger higher-level side-effects when the Sender is selected,
// we might want to pass in an onSenderSelect into ManagePublicKey

function ManagePublicKey({ onSenderSelect }) {
    function handleSenderSelect(selectedSender: string) {
        // manage local state
        setSender(selectedSender);
   
        // trigger the higher-level side-effects
        onSenderSelect(selectedSender);
    }
}

// easier differentiation if we need to pass this down as well
onSenderSelect={ handleSenderSelect }
```

frontend-react/src/components/ManagePublicKey/ManagePublicKey.tsx
```diff
@@ -1,18 +1,136 @@
...
+
+        try {
+            sendFile({
+                contentType: CONTENT_TYPE,
+                fileContent: fileContent,
+                file: file,
+            });
+            setFileSubmitted(true);
+        } catch (e: any) {
+            console.trace(e);
```
stephenkao, line 63: Do we need the `console.trace` here?  There's already one invoked in `showError`: https://github.com/CDCgov/prime-reportstream/blob/master/frontend-react/src/components/AlertNotifications.tsx#L40

frontend-react/src/components/ManagePublicKey/ManagePublicKey.tsx
```diff
@@ -1,18 +1,136 @@
...
+            showError(`Uploading public key failed. ${e.toString()}`);
+        }
+    };
+
+    const onFileChange = async (event: React.ChangeEvent<HTMLInputElement>) => {
+        // No file selected
+        if (!event?.target?.files?.length) {
+            setFile(null);
```
stephenkao, line 71: Is it possible to unselect files?

frontend-react/src/utils/FileUtils.ts
```diff
@@ -34,3 +34,55 @@ export const parseCsvForError = (
...
+        }
+
+        if (file.type !== mimeType) {
+            return {
+                fileTypeError: `The file MIME type must be of type .'${mimeType}'`,
+            };
+        }
+
+        return {
+            fileName: file.name,
```
stephenkao, line 61: Why would we return the file name in a validation function?

frontend-react/src/utils/FileUtils.ts
```diff
@@ -34,3 +34,55 @@ export const parseCsvForError = (
...
+export const validateFileSize = (file: File) => {
+    try {
+        if (file.size > PAYLOAD_MAX_BYTES) {
+            return {
+                fileSizeError: `The file '${file.name}' is too large. The maximum file size is ${PAYLOAD_MAX_KBYTES}k`,
+            };
+        }
+
+        return {
+            fileName: file.name,
```
stephenkao, line 80: Why are we returning filename here too?

frontend-react/src/utils/FileUtils.ts
```diff
@@ -34,3 +34,55 @@ export const parseCsvForError = (
...
+        if (file.type !== mimeType) {
+            return {
+                fileTypeError: `The file MIME type must be of type .'${mimeType}'`,
+            };
+        }
+
+        return {
+            fileName: file.name,
+        };
+    } catch (err: any) {
```
stephenkao, line 63: I don't think we need a `try-catch` around this; if there are exceptional cases, they should be escalated accordingly 🤔

frontend-react/src/components/ManagePublicKey/ManagePublicKeyChooseSender.tsx
```diff
@@ -0,0 +1,64 @@
...
+    }, [senders]);
+
+    function handleSubmit() {
+        onSenderSelect(selectedSender);
+    }
+
+    return (
+        <div data-testid="ManagePublicKeyChooseSender">
+            {isLoading && <Spinner message="Loading..." />}
+            {!isLoading && senders && senders?.length > 1 && (
```
stephenkao, line 33: Very minor point, but `senders && senders?.length > 1` is redundant; we can just simplify to `senders?.length > 1` since that'll always be false if `senders` is falsy.

frontend-react/src/components/ManagePublicKey/ManagePublicKeyChooseSender.tsx
```diff
@@ -0,0 +1,64 @@
...
+export interface ManagePublicKeyChooseSenderProps {
+    onSenderSelect: (sender: string) => void;
+}
+
+export default function ManagePublicKeyChooseSender({
+    onSenderSelect,
+}: ManagePublicKeyChooseSenderProps) {
+    const { isLoading, senders } = useOrganizationSenders();
+    const [selectedSender, setSelectedSender] = useState(
+        senders?.length === 1 ? senders[0].name : ""
```
stephenkao, line 16: I realize that `useOrganizationSenders` is called above this component, but we shouldn't build an assumption on that.  If this component were rendered anywhere else, `senders` wouldn't be loaded at this point so I think we can just default to `undefined`/`null`, and it'll be chosen automatically in [the effect below](https://github.com/CDCgov/prime-reportstream/pull/8731/files#diff-8e46f388cb0b202962a67f71e21169bf7b34cd92b20920f6e6579c274dda97b8R19).

frontend-react/src/components/ManagePublicKey/ManagePublicKey.tsx
```diff
@@ -1,18 +1,136 @@
...
+        if (!fileTypeError && !fileSizeError) {
+            setFile(file);
+            setFileContent(content);
+        }
+    };
+
+    return (
+        <>
+            {sender.length === 0 && (
+                <ManagePublicKeyChooseSender onSenderSelect={onSenderSelect} />
```
stephenkao, line 98: Can we omit this step for now since we're not going to need Sender granularity in the key?  (We've discussed this already, but just wanted to follow up on it!)

frontend-react/src/components/ManagePublicKey/ManagePublicKeyChooseSender.test.tsx
```diff
@@ -0,0 +1,106 @@
...
+                mockUseOrganizationSenders({
+                    isLoading: false,
+                    senders: DEFAULT_SENDERS,
+                });
+
+                renderApp(<ManagePublicKeyChooseSender {...DEFAULT_PROPS} />);
+            });
+
+            test("renders the sender options", () => {
+                expect(screen.getByText("-Select-")).toBeVisible();
```
stephenkao, line 81: Small point, but if we're checking for the options below, I don't think we need to check for this.

frontend-react/src/components/ManagePublicKey/ManagePublicKeyChooseSender.test.tsx
```diff
@@ -0,0 +1,106 @@
...
+                expect(screen.getByText("default")).toBeVisible();
+                expect(screen.getByText("ignore-full-elr")).toBeVisible();
+            });
+
+            test("renders the submit button", () => {
+                expect(screen.getByText("Submit")).toBeVisible();
+            });
+        });
+
+        describe("and Organizations have no senders", () => {
```
stephenkao, line 91: Should we ever get to this point?  They should be redirected if they have no Senders [from here](https://github.com/CDCgov/prime-reportstream/blob/master/frontend-react/src/components/ManagePublicKey/ManagePublicKey.tsx#L23), no?

stephenkao: Otherwise, it's not great UX to render a blank page for users if they don't have certain privileges 😬

frontend-react/src/components/ManagePublicKey/ManagePublicKeyChooseSender.tsx
```diff
@@ -0,0 +1,64 @@
...
+                                    {name}
+                                </option>
+                            ))}
+                        </Dropdown>
+                        <Button
+                            key="submit-sender"
+                            type="submit"
+                            outline
+                            className="padding-y-1"
+                            disabled={selectedSender === ""}
```
stephenkao, line 55: This can be simplified to just `disabled={!selectedSender}`

frontend-react/src/components/ManagePublicKey/ManagePublicKeyChooseSender.tsx
```diff
@@ -0,0 +1,64 @@
...
+                <Form name="sender-select" onSubmit={handleSubmit}>
+                    <FormGroup>
+                        <Dropdown
+                            id="senders-dropdown"
+                            name="senders-dropdown"
+                            onChange={(e) => {
+                                setSelectedSender(e.target.value);
+                            }}
+                        >
+                            <option value="">-Select-</option>
```
stephenkao, line 43: Should this be disabled?

frontend-react/src/components/ManagePublicKey/ManagePublicKeyChooseSender.tsx
```diff
@@ -0,0 +1,64 @@
...
+                                <option key={name} value={name}>
+                                    {name}
+                                </option>
+                            ))}
+                        </Dropdown>
+                        <Button
+                            key="submit-sender"
+                            type="submit"
+                            outline
+                            className="padding-y-1"
```
stephenkao, line 54: This'd be @jillian-hammer's call, but I think we should align with the default styles of out-of-the-box components like Button rather than apply custom styling.  If we need to create a "slim" button, we should create a new variant for that component so we don't end up with ad hoc styles all over the site.

frontend-react/src/components/ManagePublicKey/ManagePublicKey.tsx
```diff
@@ -1,18 +1,136 @@
...
+            return data;
+        },
+    };
+
+    const onSenderSelect = (selectedSender: string) => {
+        setSender(selectedSender);
+    };
+
+    const onBack = () => {
+        navigate(-1);
```
stephenkao, line 42: Since this is all on one page, I think the `navigate(-1)` would be incorrect since we're not pushing to history state so each `onBack` should really be unsetting the state so we end up at the proper step instead.  Sorry, this is my fault for suggesting `navigate` usage here.

frontend-react/src/components/Spinner.tsx
```diff
@@ -1,32 +1,45 @@
...
-                height={60}
-                width={60}
-                color="#2378c3"
-                secondaryColor="#aacdec"
-                strokeWidth={8}
-                strokeWidthSecondary={8}
-                ariaLabel="loading-indicator"
-            />
+        <div>
+            <div
```
stephenkao, line 23: The styles of this are off:
Screenshot 2023-03-24 at 14 33 04

Two things are causing this:
- `grid-container` is applied to the spinner container, which means that it'll take up the full width
- The message is wrapped in a `span`, which is inline and _won't_ take up the full width

stephenkao: (There should also probably be some spacing between the tip and the spinner)

frontend-react/src/styles/_components.scss
stephenkao: What I mean is the `size` prop on the Icon component, not a CSS value :)

frontend-react/src/components/ManagePublicKey/ManagePublicKey.test.tsx
penny-lischer: no, typescript complains

frontend-react/src/components/ManagePublicKey/ManagePublicKey.tsx
penny-lischer: Yes, thats why I added this.

frontend-react/src/components/ManagePublicKey/ManagePublicKeyChooseSender.tsx
penny-lischer: Needed to add since 'senders.length' is possibly 'undefined'.

frontend-react/src/utils/FileUtils.ts
penny-lischer: Will remove.

frontend-react/src/utils/FileUtils.ts
penny-lischer: Will remove

frontend-react/src/components/ManagePublicKey/ManagePublicKeyChooseSender.test.tsx
penny-lischer: true

frontend-react/src/components/ManagePublicKey/ManagePublicKey.tsx
penny-lischer: So should we redirect to the Resources page? Otherwise back may just display `ManagePublicKeyUpload` if the org only has one sender.

frontend-react/src/components/ManagePublicKey/ManagePublicKeyChooseSender.tsx
penny-lischer: Not sure I understand?

frontend-react/src/components/ManagePublicKey/ManagePublicKey.tsx
penny-lischer: Sure, should I add a comment and comment it out or remove entirely?

frontend-react/src/styles/_components.scss
penny-lischer: Whoops

frontend-react/src/components/ManagePublicKey/ManagePublicKeyChooseSender.tsx
stephenkao: But wouldn't `senders?.length` (with the safe navigation/optional chaining via `?.`) already handle that?  Like:

```ts
const senders = undefined; // just for demonstration purposes
senders?.length > 1; // evaluates to false
```

frontend-react/src/components/ManagePublicKey/ManagePublicKey.tsx
stephenkao: I think you can comment it out and add an inline comment for what it's not currently being used yet -- that'd make the re-introduction of this logic much simpler whenever we have that scope granularity :)

frontend-react/src/components/ManagePublicKey/ManagePublicKey.tsx
stephenkao: We can break this out into a subsequent ticket/pull request of course!  We don't have to hold up this pull request to get everything perfect since the flow is intended to be incomplete at this stage anyway.  BUT this should be fixed prior to when we go live.

penny-lischer: Will remove and create a new ticket for future sprint.

frontend-react/src/components/ManagePublicKey/ManagePublicKeyChooseSender.tsx
penny-lischer: ok I will change the return value in `UseOrganizationSenders.ts` to return empty [] if no senders and that way I can apply `senders?.length > 1 `here.

frontend-react/src/components/ManagePublicKey/ManagePublicKey.tsx
```diff
@@ -1,18 +1,131 @@
...
+
+        if (!fileTypeError && !fileSizeError) {
+            setFile(file);
+            setFileContent(content);
+        }
+    };
+
+    return (
+        <>
+            {/*Waiting on backend to support this*/}
```
stephenkao, line 87: You can just comment out the entire region:

```tsx
{ /* Waiting on backend to support this 
    { sender.length === 0 && (
        ...
    ) }
*/ }
```

frontend-react/src/components/ManagePublicKey/ManagePublicKeyUpload.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+    });
+
+    describe("when file selected", () => {
+        beforeEach(() => {
+            doRender({
+                file: fakeFile,
+            });
+        });
+
+        test("enables the submit", async () => {
```
stephenkao, line 45: I don't think this needs to be `async`

frontend-react/src/components/Spinner.tsx
```diff
@@ -1,32 +1,52 @@
...
+                    />
+                </div>
+            </div>
+            {message && (
+                <div className="grid-row">
+                    <div
+                        className="margin-x-auto tablet:grid-col-5 line-height-sans-6"
+                        data-testid="spinner-message"
+                    >
+                        <p>{message}</p>
```
stephenkao, line 46: Let's only wrap this in a `<p>` if `message` is a string and render verbatim otherwise.  If we pass in markup, there's a possibility that wrapping in a `<p>` would be invalid HTML.

stephenkao (approved):
Couple of minor points, but this looks good to me!

frontend-react/src/components/ManagePublicKey/ManagePublicKey.tsx
```diff
@@ -1,18 +1,131 @@
...
 import { MemberType } from "../../hooks/UseOktaMemberships";
+import { validateFileType, validateFileSize } from "../../utils/FileUtils";
+
+import ManagePublicKeyUpload from "./ManagePublicKeyUpload";
+
+export const CONTENT_TYPE = "application/x-x509-ca-cert";
+export const FORMAT = "PEM";
+
+function ManagePublicKeySwitchDisplay() {
+    // const [sender, setSender] = useState("");
```
etanb, line 17: Remove?

frontend-react/src/components/ManagePublicKey/ManagePublicKey.tsx
```diff
@@ -1,18 +1,131 @@
...
+    const [fileContent, setFileContent] = useState("");
+    const [file, setFile] = useState<File | null>(null);
+    const [fileSubmitted, setFileSubmitted] = useState(false);
+
+    // TODO: mocked for now - make the call you need when sending the file
```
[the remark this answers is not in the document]
penny-lischer: Not necessary since I'm already working on the ticket #8398

frontend-react/src/components/ManagePublicKey/ManagePublicKey.tsx
penny-lischer: @stephenkao I thought you wanted me to comment out???

stephenkao: Oh, duh -- you're right.  @etanb for context, we decided to keep this code commented out for now since we're not going to support this functionality as part of this push.  If/when we can support this, we can uncomment to enable the functionality again.

@penny-lischer you have a comment for the JSX below, but would you mind adding an inline comment here explaining the situation here as well?

frontend-react/src/styles/_components.scss
stephenkao: I don't think we should be adding custom styles for each of our icons, which is becoming a pattern with our recent work.  It'd be worth checking in with Jillian to see if we can just use the `size` prop as noted above and align with the [variants provided by the USWDS](https://designsystem.digital.gov/components/icon/).

frontend-react/src/components/ManagePublicKey/ManagePublicKey.tsx
etanb: I think we should remove for now and reference this thread, or have a comment above (hopefully with a ticket) that explains this? 

We've had commented out code with explanation across the codebase before and it makes it hard to maintain. Who knows, we might not work on this for a while and then we'll lose context.

etanb (approved):

frontend-react/src/utils/FileUtils.ts
```diff
@@ -9,28 +9,56 @@ export const PAYLOAD_MAX_KBYTES = (PAYLOAD_MAX_BYTES / 1024).toLocaleString(
...
 };
+
+export const validateFileType = (
+    file: File,
+    fileExt: string,
+    mimeType: string
+) => {
+    // look at the filename extension.
+    const fileNameArray = file.name.split(".");
+    const uploadFileExtension = fileNameArray[fileNameArray.length - 1];
```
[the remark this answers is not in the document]
penny-lischer: Will update in next iteration

PR #8735, opened 2023-03-15
Remove scripts/test.js in favor of config

### LINKED ISSUE #8734: Remove use of scripts/test.js in favor of jest.config.ts
Now that we are ejected from CRA we have a jest.config.ts file that automatically gets used by jest that we can use to declaratively make configuration changes. The reliance currently on calling the scripts/test.js file to initiate tests prevents built-in jest support in IDEs such as IntelliJ products from working without extreme effort. The test.js file should be removed in favor of doing necessary work in the jest.config.ts file instead (ex: loading env variables). This would complete some work related to #8469.

jpandersen87:
Closes #8734 

This PR removes the use of the scripts/test.js script to initiate jest testing in favor of doing necessary work inside of jest.config.ts so that custom setup is not required to use jest (ex: IDEs with built-in UI features for jest testing). NPM scripts have been updated to call jest directly instead of the test.js file.


stephenkao (approved):
Nice catch!

penny-lischer (approved):
LGTM

PR #8741, opened 2023-03-15
Add tests to schema config reader and adjust tests for merging schema…

jimmyfagan:
This PR adds a missing piece of the merge between two schemas. Previously, the merge would pull child schema elements into the parent schema, either by adding them straight in, or by merging them into an element within the parent schema, but the name of the resulting merged schema would still reflect the base schema and the constants from the child schema would be ignored. A couple lines of code was enough to fix that behavior.

As part of this, I adjust the unit tests to make sure we're checking the constants and the name of a schema post-merge.

Test Steps:

- Run fhirdata --input-file src/testIntegration/resources/datatests/FHIR_to_HL7/sample_SR_1_20230302-0001.fhir -s <some schema>.yml --output-format HL7 --output-file src/testIntegration/resources/datatests/FHIR_to_HL7/sample_SR_1_20230302-0001.hl7 using a yml file that extends another yml file with constants defined at the top level of both and using those constants as part of the conversion.
- Observe the results to see that the constant values from both the child schema and the parent schema were used.

or

- Run test `test read from file with extends`

## Changes
- Edit the `merge` function in `ConfigSchema.kt` to allow merging of two schemas' `constants` and `name` values
- Add to `ConfigSchemaReaderTests.kt` `ConverterSchemaTests.kt` and `FhirTransformSchemaTests.kt` to exercise new code and to more closely verify behavior of inherited constants/names.

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- Addresses #8723

### DIFF
--- a/prime-router/src/main/kotlin/fhirengine/translation/hl7/schema/ConfigSchema.kt
+++ b/prime-router/src/main/kotlin/fhirengine/translation/hl7/schema/ConfigSchema.kt
@@ -105,6 +105,8 @@ abstract class ConfigSchema<T : ConfigSchemaElement>(
                 this.elements.add(childElement)
             }
         }
+        this.constants.putAll(childSchema.constants)
+        this.name = childSchema.name
     }
 
     /**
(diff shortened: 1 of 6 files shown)


victor-chaparro (approved):
I tested this with my branch and constants are working, nice! By the way something we tried to do to suppress AOEs was to override a constant in the extension schema. I tested that and it works too, but it may be worth adding a unit test for that, in case we need it in the future.

jimmyfagan:
Covered Victor's suggestion in [this commit](https://github.com/CDCgov/prime-reportstream/pull/8741/commits/73c5d05e7172c377455425e4c721aacc2de66bbf), just waiting on another approval to merge.

carlosfelix2 (approved):

Issue #8737, opened 2023-03-15
Enable receivers to receive data with the dates in a specified timezone.


JessicaWNava:
### User Story:
As a receiver, I want to receive data with the dates in my local time, as I could in the Covid Pipeline by calling convertDateTimesToReceiverLocalTime.

### Description/Use Case
This will be used to convert the times sent to a receiver into their local date time, or the timezone they request. 

### Risks/Impacts/Considerations
This is needed for CA.

### Dev Notes:
Suggesting that we create a FHIR function that will handle converting date times to receiver local times.

Acceptance Criteria
Dates can have their timezone changed and the time itself changes accordingly.

bishoyayoub:
Adding Jimmy as supplemental resource while Jessica is away.

PR #8751, opened 2023-03-16
Custom FHIR function to change the timezone a date is in

JessicaWNava:
This PR creates a FHIR function to change the timezone of a date.

Test Steps:
1. Using the terminal, find a date to test on and make it the resource. For example: resource=Bundle.entry.resource.ofType(Organization).meta.lastUpdated[0]
2. Run the method ex. %resource.changeTimezone("Asia/Tokyo")
3. Make sure it outputs the date you expect.
4. Should fail with bad date data
5. Should fail with incorrect timezone

## Changes
- Added a FHIR function that is run on a date and takes a timezone

### Testing
- [x] Tested locally?
- [x] Added tests?

### Process

## Linked Issues
- Addresses #8737 

## To Be Done
- #issue

    - new endpoint A
    - new endpoint B


jimmyfagan (commented):
Looks pretty good, just left a question and a couple minor suggestions.

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/CustomFHIRFunctions.kt
```diff
@@ -431,4 +450,35 @@ object CustomFHIRFunctions {
             mutableListOf(StringType(result))
         }
     }
+
+    /**
+     * Using the [focus] as the date to be changed and the [parameters] as the timezone to change it to
```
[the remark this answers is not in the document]
carlosfelix2: You always want to return the changed value. This allows the user to decide if they want to replace the value in config or use it somewhere else. Agreed it needs a reword.

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/CustomFHIRFunctions.kt
```diff
@@ -431,4 +450,35 @@ object CustomFHIRFunctions {
...
+     * @return a date in the new timezone
+     */
+    fun changeTimezone(
+        focus: MutableList<Base>,
+        parameters: MutableList<MutableList<Base>>?
+    ): MutableList<Base> {
+        if (focus.size != 1) {
+            throw SchemaException("Must call changeTimezone on a Date")
```
carlosfelix2, line 463: Change error message to be more descriptive of the check. You also do not check to see if this is a date time.

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/CustomFHIRFunctions.kt
```diff
@@ -431,4 +450,35 @@ object CustomFHIRFunctions {
...
+        val timezonePassed = parameters!!.first().first().primitiveValue()
+
+        val isoFormat = SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssXXX")
+
+        try {
+            val date: Date = isoFormat.parse(focus[0].primitiveValue())
+            val changedDate = date.toInstant().atZone(ZoneId.of(timezonePassed))
+            return mutableListOf(StringType(changedDate.toString()))
+        } catch (zoneRulesException: ZoneRulesException) {
+            throw SchemaException("Invalid timezone passed. See ZoneId for available timezone values.")
```
carlosfelix2, line 479: Change to "Invalid timezone $timezone..."

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/CustomFHIRFunctions.kt
```diff
@@ -431,4 +450,35 @@ object CustomFHIRFunctions {
...
+        } catch (parseException: ParseException) {
+            throw SchemaException("Date does not match expected format.")
```
carlosfelix2, line 481: You are not dealing with dates. Change to date time

prime-router/src/test/kotlin/fhirengine/translation/hl7/utils/CustomFHIRFunctionsTests.kt
```diff
@@ -68,7 +70,9 @@ class CustomFHIRFunctionsTests {
         // Individual function results are tested on their own unit tests.
         CustomFHIRFunctions.CustomFHIRFunctionNames.values().forEach {
             // todo: this is temporary until this code is moved
-            if (it != CustomFHIRFunctions.CustomFHIRFunctionNames.LivdTableLookup) {
+            if (it != CustomFHIRFunctions.CustomFHIRFunctionNames.LivdTableLookup &&
+                it != CustomFHIRFunctions.CustomFHIRFunctionNames.ChangeTimezone
```
carlosfelix2, line 74: Remove ChangeTimezone from this if. We removed Livd due to its dependency on metadata

carlosfelix2 (changes requested):

jimmyfagan (approved):

prime-router/src/test/kotlin/fhirengine/translation/hl7/utils/CustomFHIRFunctionsTests.kt
jimmyfagan: Still needed to handle this specially because the generic inputs used in this `forEach` would cause `changeTimezone` to fail, but instead of skipping it altogether, I just added a specific `if` block for that case.

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/CustomFHIRFunctions.kt
jimmyfagan: Adjusted the error message, but still did not check for a date time specifically since the incoming type can be either a string, an instant, a datetime, etc, and the exception checking later on will cover cases where it was not a compatible type.

jimmyfagan:
Latest commit covers additional testing to show that the cases we discussed this morning are covered. It also makes some adjustments to the `changeTimezone` function itself to utilize existing fhir type conversion and the constructor of `DateTimeType` which takes a date, precision, and timezone, rather than doing that conversion more manually. Currently, this just uses the precision of the input, but if we need another function to adjust precision, we can do that in a later ticket.

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/CustomFHIRFunctions.kt
```diff
@@ -102,6 +107,15 @@ object CustomFHIRFunctions {
                 )
             }
 
+            CustomFHIRFunctionNames.ChangeTimezone -> {
+                FHIRPathEngine.IEvaluationContext.FunctionDetails(
+                    "changes the timezome of the date resource to the timezome passed in as the " +
```
carlosfelix2, line 112: It says date here which is wrong, but maybe this should say what types it can handle.
Type: time zone

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirBundleUtils.kt
```diff
@@ -44,13 +44,27 @@ object FhirBundleUtils : Logging {
...
         return if (sourceType == targetType || targetType == "*") {
             value
+        } else if (targetType == StringCompatibleType.DateTime.typeAsString) {
+            when (sourceType) {
+                StringCompatibleType.DateTime.typeAsString -> value
+                StringCompatibleType.Instant.typeAsString -> DateTimeType(value.primitiveValue())
+                StringCompatibleType.Date.typeAsString -> DateTimeType(value.primitiveValue())
+                StringCompatibleType.String.typeAsString -> {
+                    var strValue = value.primitiveValue()
+                    strValue = strValue.replace("""\[[a-zA-Z0-9/_+\-"']+]""".toRegex(), "")
```
carlosfelix2, line 54: What is this for? Add comment and explain.

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirBundleUtils.kt
```diff
@@ -44,13 +44,27 @@ object FhirBundleUtils : Logging {
...
+                    DateTimeType(strValue)
```
carlosfelix2, line 55: What happens when the string does not have a date time?

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/CustomFHIRFunctions.kt
```diff
@@ -429,4 +447,36 @@ object CustomFHIRFunctions {
...
+    ): MutableList<Base> {
+        if (focus.size != 1) {
+            throw SchemaException("Must call changeTimezone on a single element")
+        }
+
+        val inputDate = FhirBundleUtils.convertFhirType(focus[0], focus[0].fhirType(), "dateTime") as? DateTimeType
+            ?: throw SchemaException("Must call changeTimezone on a dateTime compatible value")
+
+        if (parameters == null || parameters[0].size != 1) {
+            throw SchemaException("Must pass a timezone as the parameter")
```
carlosfelix2, line 467: Change timezone to time zone (add a space)

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/CustomFHIRFunctions.kt
```diff
@@ -429,4 +447,36 @@ object CustomFHIRFunctions {
...
+    fun changeTimezone(
+        focus: MutableList<Base>,
+        parameters: MutableList<MutableList<Base>>?
+    ): MutableList<Base> {
+        if (focus.size != 1) {
+            throw SchemaException("Must call changeTimezone on a single element")
+        }
+
+        val inputDate = FhirBundleUtils.convertFhirType(focus[0], focus[0].fhirType(), "dateTime") as? DateTimeType
+            ?: throw SchemaException("Must call changeTimezone on a dateTime compatible value")
```
carlosfelix2, line 464: This message may be confusing. What is a dateTime compatible value? Is that defined in FHIR or FHIR Path? Simplify

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/CustomFHIRFunctions.kt
```diff
@@ -429,4 +447,36 @@ object CustomFHIRFunctions {
...
+        if (parameters == null || parameters[0].size != 1) {
+            throw SchemaException("Must pass a timezone as the parameter")
+        }
+
+        val inputTimeZone = parameters.first().first().primitiveValue()
+        val timezonePassed = try {
+            TimeZone.getTimeZone(ZoneId.of(inputTimeZone))
+        } catch (zoneRulesException: ZoneRulesException) {
+            throw SchemaException(
+                "Invalid timezone $inputTimeZone passed. See FHIR timezone valueset " +
```
carlosfelix2, line 475: Maybe add a link to the valueset? Also, you are using a Java ZoneId and not a FHIR value here. Perhaps change to some IEEE list or something more specific to this implementation. Do not say what ZoneId needs as that will not make sense to a user either.

carlosfelix2 (changes requested):

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/CustomFHIRFunctions.kt
jimmyfagan: Will update the description to say dateTime, and to talk about acceptable types.

As for timezone vs time zone, I'm not seeing anything conclusive, I think in English in general, `time zone` is more common, but http://hl7.org/fhir/R4B/valueset-timezones.html and http://build.fhir.org/valueset-timezones.html both say `timezone`, so I was thinking we should match that. What do you think?

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirBundleUtils.kt
```diff
@@ -44,13 +44,27 @@ object FhirBundleUtils : Logging {
     fun convertFhirType(value: Base, sourceType: String, targetType: String, logger: KotlinLogger = this.logger): Base {
         return if (sourceType == targetType || targetType == "*") {
             value
+        } else if (targetType == StringCompatibleType.DateTime.typeAsString) {
+            when (sourceType) {
+                StringCompatibleType.DateTime.typeAsString -> value
+                StringCompatibleType.Instant.typeAsString -> DateTimeType(value.primitiveValue())
+                StringCompatibleType.Date.typeAsString -> DateTimeType(value.primitiveValue())
```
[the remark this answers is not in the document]
jimmyfagan: Dates don't have timezones, but a date can be converted to a dateTime, because the dateTime can have a precision level of YEAR or MONTH or DAY. At those precisions, the timezone will be ignored, but either way, I think the conversion should be allowed at this level.

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/CustomFHIRFunctions.kt
jimmyfagan: Wasn't sure about whether or not the link would be stable enough, but I suppose it probably won't change from https://hl7.org/fhir/valueset-timezones.html, will add that.

[Java ZoneId](https://docs.oracle.com/javase/8/docs/api/java/time/ZoneId.html#of-java.lang.String-) and [FHIR](https://docs.oracle.com/javase/8/docs/api/java/time/ZoneId.html#of-java.lang.String-) both use the IANA Time Zone Database for their list of time zones, and `DateTimeType`'s constuctor takes a Java TimeZone, so it should be safe to utilize Java's TimeZone and ZoneId in this implementation.

jimmyfagan:
At this point I believe I've addressed all concerns except for two:

1. I've kept `timezone` instead of `time zone` in most comments and variable names. In English in general, `time zone` seems to be more common, but http://hl7.org/fhir/R4B/valueset-timezones.html and http://build.fhir.org/valueset-timezones.html both say `timezone`, so I was thinking we should match that.
2. I've kept `date` as an acceptable input to `changeTimezone`. With dates, it becomes a no-op basically, but it still is an acceptable form of `dateTime`, so felt that it shouldn't result in an error.

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirBundleUtils.kt
```diff
@@ -44,13 +44,27 @@ object FhirBundleUtils : Logging {
...
+                StringCompatibleType.String.typeAsString -> {
```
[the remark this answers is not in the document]
jimmyfagan: Turned out we didn't have a good use case for handling strings at this point, so I ended up reworking a lot of this to not even deal with a conversion. It makes some of the other comments irrelevant in the latest version of this branch.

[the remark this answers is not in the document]
jimmyfagan: Turned out we didn't have a good use case for handling strings at this point, so I ended up reworking a lot of this to not even deal with a conversion. It makes some of the other comments irrelevant in the latest version of this branch.

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/CustomFHIRFunctions.kt
```diff
@@ -429,4 +447,38 @@ object CustomFHIRFunctions {
...
+     */
+    fun changeTimezone(
+        focus: MutableList<Base>,
+        parameters: MutableList<MutableList<Base>>?
+    ): MutableList<Base> {
+        if (focus.size != 1) {
+            throw SchemaException("Must call changeTimezone on a single element")
+        }
+
+        val inputDate = focus[0] as? BaseDateTimeType ?: throw SchemaException(
```
carlosfelix2, line 463: I like this check with the superclass!

prime-router/src/test/kotlin/fhirengine/translation/hl7/FhirTransformerTests.kt
```diff
@@ -605,4 +607,37 @@ class FhirTransformerTests {
...
+        FhirTransformer(schema).transform(bundle)
+        val newValue =
+            FhirPathUtils.evaluate(
+                CustomContext(bundle, bundle),
+                bundle,
+                bundle,
+                "Bundle.entry.resource.ofType(Patient).meta.lastUpdated"
+            )
+        assertThat(resource.meta.lastUpdated).isEqualTo(bundle.timestamp)
+        assertThat(newValue[0].primitiveValue()).isEqualTo("2021-08-09T05:52:00.000-07:00")
```
carlosfelix2, line 641: Add tests with partial date times:
1. What happens if there is only a year?
2. What happens if there is only a date (1/1/2023)?
Then figure out if that behavior is correct.

prime-router/src/test/kotlin/fhirengine/translation/hl7/utils/CustomFHIRFunctionsTests.kt
```diff
@@ -68,7 +75,13 @@ class CustomFHIRFunctionsTests {
         // Individual function results are tested on their own unit tests.
         CustomFHIRFunctions.CustomFHIRFunctionNames.values().forEach {
             // todo: this is temporary until this code is moved
```
carlosfelix2, line 77: this comment needs to be moved down to the LIVD one.

prime-router/src/test/kotlin/fhirengine/translation/hl7/utils/CustomFHIRFunctionsTests.kt
```diff
@@ -68,7 +75,13 @@ class CustomFHIRFunctionsTests {
...
-            if (it != CustomFHIRFunctions.CustomFHIRFunctionNames.LivdTableLookup) {
+            if (it == CustomFHIRFunctions.CustomFHIRFunctionNames.ChangeTimezone) {
```
carlosfelix2, line 78: Consider in the future generically checking how many parameters are needed and doing the test.

prime-router/src/test/kotlin/fhirengine/translation/hl7/utils/CustomFHIRFunctionsTests.kt
```diff
@@ -323,4 +336,137 @@ class CustomFHIRFunctionsTests {
             )
         }.isFailure()
     }
+
+    @Test
+    fun `test changeTimezone`() {
+        val isoFormat = SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssXXX")
```
carlosfelix2, line 342: Add tests with partial date times:

1. What happens if there is only a year?
2. What happens if there is only a date (1/1/2023)?
3. Make a test for Instant type too.
Then figure out if that behavior is correct.

carlosfelix2 (approved):

prime-router/src/test/kotlin/fhirengine/translation/hl7/FhirTransformerTests.kt
```diff
@@ -605,4 +607,37 @@ class FhirTransformerTests {
         assertThat(servRequest1.id).isEqualTo("0")
         assertThat(servRequest2.id).isEqualTo("1")
     }
+
+    @Test
+    fun `test transform dateTime property with changeTimezone`() {
```
carlosfelix2, line 612: I do not understand which unit this is testing. Does this need to be removed?

prime-router/src/test/kotlin/fhirengine/translation/hl7/utils/CustomFHIRFunctionsTests.kt
jimmyfagan: I like the idea, might be tricky with the number and type of parameters mattering, but worth thinking about.

PR #8760, opened 2023-03-17
Experience 8616: Relax Organization/service name validation

### LINKED ISSUE #8616: Remove reserved word, hyphen restriction for Organization/Sender/Receiver names
## Problem statement

In [#7139](https://github.com/CDCgov/prime-reportstream/issues/7139), we added a restriction against special characters and reserved words for Organization/Sender/Receiver names just in case this were to cause issues in the back end.  However, reserved words aren't an issue at the moment so we can remove that restriction for now.

Additionally, we should allow hyphens since those are already part of many of our service names.

## What you need to know

- [isProhibitedName](https://github.com/CDCgov/prime-reportstream/blob/master/frontend-react/src/utils/misc.ts#L53-L79) is the function that checks against special characters and reserved words.

## Acceptance criteria

- [ ] Organization/Sender/Receiver names can contain reserved words and hyphens again
- [ ] Organization/Sender/Receiver names still cannot contain disallowed characters

stephenkao:
This changeset relaxes the Organization/service name validation to allow letters, numbers, hyphens, and underscores, and it removes the reserved word restrictions we had previously.  This was causing issues with current Sender/Receiver settings, and as far as I know, we haven't run into issues with them in production anyway.  I think it'd probably be best to run this validation on the back end instead of the front end since the user can make the PUT themselves with invalid names, but this is just a helpful reminder to admins to stay in line with our naming conventions.  Tests have been updated as well.

Test Steps:
1. Log in as an admin
2. Go to /admin/settings and choose an Organization
3. Add a new service with only letters/numbers/hyphens/underscores --> verify success + saved
4. Add a new service with disallowed characters --> verify error + not saved
5. Edit an existing service whose name only contains letters/numbers/hyphens/underscores --> verify success + saved
6. Edit a new service with disallowed characters (may need to manually update database) --> verify error + not saved

## Changes
Screenshot 2023-03-17 at 15 47 37
Screenshot 2023-03-17 at 15 48 31
Screenshot 2023-03-17 at 15 48 52

### Testing
- [x] Tested locally?
- ]x] Added tests?

### Process

## Linked Issues
- Fixes #8616

### DIFF
--- a/frontend-react/src/components/Admin/EditSenderSettings.tsx
+++ b/frontend-react/src/components/Admin/EditSenderSettings.tsx
@@ -14,7 +14,7 @@ import { jsonSortReplacer } from "../../utils/JsonSortReplacer";
 import {
     getErrorDetailFromResponse,
     getVersionWarning,
-    isProhibitedName,
+    isValidServiceName,
     VersionWarningType,
 } from "../../utils/misc";
 import { ObjectTooltip } from "../tooltips/ObjectTooltip";
@@ -119,12 +119,10 @@ const EditSenderSettingsForm: React.FC<EditSenderSettingsFormProps> = ({
 
     const ShowCompareConfirm = async () => {
         try {
-            const { prohibited, errorMsg } = isProhibitedName(
-                orgSenderSettings.name
-            );
+            const { name } = orgSenderSettings;
 
-            if (prohibited) {
-                showError(errorMsg);
+            if (!isValidServiceName(name)) {
+                showError(`${name} cannot contain special characters.`);
                 return false;
             }
 
--- a/frontend-react/src/utils/misc.ts
+++ b/frontend-react/src/utils/misc.ts
@@ -46,37 +46,9 @@ export const checkJson = (
     }
 };
 
-export function includesSpecialChars(text: string): boolean {
-    return /[^a-z\d\s]/i.test(text);
+export function isValidServiceName(text: string): boolean {
+    return /^[a-z\d-_]+$/i.test(text);
 }
-
-export const isProhibitedName = (
-    name: string
-): { prohibited: boolean; errorMsg: string } => {
-    const prohibitedNames = [
-        "sender",
-        "receiver",
-        "org",
-        "organization",
-        "list",
-        "all", //api/check issue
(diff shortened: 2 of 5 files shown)


frontend-react/src/components/Admin/EditSenderSettings.tsx
```diff
@@ -119,12 +119,10 @@ const EditSenderSettingsForm: React.FC<EditSenderSettingsFormProps> = ({
 
     const ShowCompareConfirm = async () => {
         try {
-            const { prohibited, errorMsg } = isProhibitedName(
```
stephenkao, line 122: Some of the logic here is duplicated [in NewSetting.tsx](https://github.com/CDCgov/prime-reportstream/pull/8760/files#diff-ba4adccf9523006412c5595106a2679a0f83f9537a90816cf1f6c15929951bd7R40-R43).  I'm okay with this for now because my feeling is that we'll want to spend more time consolidating these components in the long run; the flow should be similar for adding and editing service settings in the future.

frontend-react/src/utils/misc.ts
```diff
@@ -46,38 +46,10 @@ export const checkJson = (
     }
 };
 
-export function includesSpecialChars(text: string): boolean {
-    return /[^a-z\d\s]/i.test(text);
+export function isValidServiceName(text: string): boolean {
+    return /^[a-z\d-_]+$/i.test(text);
```
stephenkao, line 50: Just glancing at the Organization/Sender/Receivers we have in the production database, this regex should cover all of them.  Important to add underscores because a few of our Senders already them in their names.

jpandersen87 (approved):
The validation should happen on BOTH ends. Frontend for best-effort prevention of sending incorrect data to backend (save a round-trip we know will fail), with the backend validating as well.

stephenkao:
> The validation should happen on BOTH ends. Frontend for best-effort prevention of sending incorrect data to backend (save a round-trip we know will fail), with the backend validating as well.

Fair point!  I just think it's much more important for the back end to validate because that's the final part before a database commit, at which point invalid names would be able to wreak havoc thereafter 😆

arnejduranovic (approved):

PR #8815, opened 2023-03-22
Mocking library

### LINKED ISSUE #8191: Add proposal for Faker and Fishery
## Problem statement

[Faker](https://github.com/faker-js/faker) and [Fisher](https://github.com/thoughtbot/fishery) are two proposed dependencies that would assist us in mocking test data.  This has been raised in team discussions already, so we just need to solidify the reasoning in a proposal and decide whether or not incorporate these dependencies.

## What you need to know

- The format of the proposal should follow [the proposal template](https://github.com/CDCgov/prime-reportstream/blob/master/frontend-react/docs/proposals/xxxx-template.md) in the codebase.

## Acceptance criteria

- [ ] A proposal is written and presented to the team with reasoning and adoption possibilities for Faker and Fishery.

jpandersen87:
Closes #8191 

This PR introduces a mocking library and a data factory library to aid in generating data for tests. Included are customizations on top of what they provide to allow for cleaner usage patterns.

Faker.js allows us to generate context-specific random data (numbers, addresses, specialized/generic words and sentences, etc.). MSW/data allows us to create a "factory" of model entities (it mimics being a database where we specify our models [ex: OrganizationSettings]) that we can call to create and access this mocked data on demand. A helper function has been added to aid in creating the factory, faker instance (by default faker is a singleton which can be dangerous when testing), and a reset function (clears the factory and restarts the faker sequence) in tests . Faker is initialized with a seed and a date reference that can be used in faker date functions. This allows the data to be random but deterministic (note: this cannot be relied on across faker versions as it relies on the structure of their mocking data that can change at any time). The factory can create enhanced rest handlers for the msw library that allow for customizing the paths and allowing for custom resolvers (response from original resolver provided in context).


frontend-react/jest.config.ts
```diff
@@ -59,6 +61,7 @@ const config: Config.InitialOptions = {
         "jest-watch-typeahead/testname",
     ],
     resetMocks: true,
+    testTimeout: IS_DEBUG ? 1000 * 60 * 5 : undefined, // Relax test timeout when debugging
```
stephenkao, line 64: Might be dumb, but I'm wondering if we should just increase it to `Infinity` when we have the inspector running since it's most likely that we have breakpoints sets for debugging.  Then again, this could just cause frustration with hanging tests 😬

frontend-react/package.json
```diff
@@ -7,6 +7,7 @@
   "dependencies": {
     "@microsoft/applicationinsights-react-js": "^3.4.1",
     "@microsoft/applicationinsights-web": "^2.8.11",
+    "@mswjs/data": "^0.11.2",
```
stephenkao, line 10: Should this be in `devDependencies`?

frontend-react/package.json
```diff
@@ -182,13 +184,21 @@
...
-    "typescript": "^4.9.5",
+    "typescript": "^5.0.2",
     "webpack": "^5.76.0",
     "webpack-bundle-analyzer": "^4.7.0",
     "webpack-dev-server": "^4.11.1",
     "webpack-manifest-plugin": "^5.0.0",
     "workbox-webpack-plugin": "^6.5.4"
   },
+  "resolutions": {
+    "fork-ts-checker-webpack-plugin": "^8.0.0",
```
stephenkao, line 195: Oh, interesting!  What else uses this dependency?

jpandersen87: react-dev-utils, a sub-package of CRA that is targeted towards a 6.X.X version of fork-ts-checker-webpack-plugin. we will need to do more work to drop all CRA attachments

frontend-react/jest.config.ts
jpandersen87: Yeah I initially was going to do Infinity but figured 5 minutes was a better default and devs can tweak it locally as needed.

frontend-react/src/__mocks__/DBMock.test.ts
```diff
@@ -0,0 +1,239 @@
...
+        reset();
+    });
+
+    describe("createMany", () => {
+        const createManyScenarios: {
+            label: string;
+            args: [number, Partial<RSOrganizationSettings>[] | undefined];
+        }[] = [
+            {
+                label: "predictably random",
```
stephenkao, line 27: This is subjective, but I think these sorts of tests would benefit from being broken out into their own blocks rather than using `test.each`.  I realize that this is just checking to make sure the records are created according to the specified numbers and initial data, but it's not entirely clear at first glance in my opinion -- especially with the `initial.length === 0` check (which really should be contained in its own test case).

stephenkao: Overall, my opinion is that a small amount of redundancy in tests for the sake of clarity is a good thing.  DRY can be used if there's a significant benefit to it, but it feels like we're adding extra logic (nested loops + `initial.length` check) in the test cases to get around DRY rather than benefitting entirely from it.

frontend-react/src/__mocks__/DBMock.test.ts
```diff
@@ -0,0 +1,239 @@
...
+                expect(all[0]).not.toStrictEqual(all[1]);
+                reset();
+                db.organizationSettings.createMany(2);
+                // check is predictably random
+                expect(db.organizationSettings.getAll()).toStrictEqual(all);
+            }
+        });
+    });
+
+    describe("json serialized return methods", () => {
```
stephenkao, line 64: I don't know if there's a better way of doing this since we're randomizing the records, but it kiiiiiiind of feels like we're reimplementing the source code logic in these tests.  I wonder if it'd be better to initialize the database with records we can reference instead like:

```tsx
const records = [{
    name: "first organization"
}, {
    name: "second organization"
}];

db.organizationCreateMany(records.length, records);
```

Then we could run expectations accordingly on `records` instead of needing to call `getAll` directly 🤔 Though I know that the randomized properties could cause some issues there so this isn't a hard blocker in my mind.

frontend-react/src/utils/Faker.ts
```diff
@@ -0,0 +1,161 @@
...
+ * naturally due to using prototype constructor. Remove this type and
+ * use HelpersModuleBase directly when we figure out how or reach v8.0.
+ */
+export type HelpersModule = OrigHelpersModule & {
+    rangeToNumber: HelpersModuleBase["rangeToNumber"];
+    multiple: HelpersModuleBase["multiple"];
+};
+
+/**
+ * Hybrid Faker with some safe features from 8.0 alpha. Allows storing
```
stephenkao, line 111: Can we just specify the v8 alpha the version for Faker in our codebase?  I don't think we should maintain this logic internally, and if there are any changes before v8 goes live, we should adjust anyway so we don't have long-lived deviations from the library.

frontend-react/src/__mocks__/DBMock.test.ts
```diff
@@ -0,0 +1,239 @@
...
+            mockServer.listen();
+
+            const records = db.organizationSettings.getAllJson();
+            const methods = Object.keys(relativePathMap ?? parentResolverMap ?? {}) as RestHandlerMethods[];
+            // If no custom paths/resolvers given, we're just testing a default handler
+            if (methods.length === 0) {
+                methods.push("getList");
+            }
+
+            for (const method of methods) {
```
stephenkao, line 196: This is another place where breaking up into dedicated tests rather than using `test.each` could help with overall clarity.  I don't think we should have a lot of `if-else`s (or any, really) when we're running expectations -- each logical fork should be its own `describe` or `test` in my opinion.

frontend-react/src/__mocks__/DBMock.ts
```diff
@@ -0,0 +1,20 @@
+import { Faker } from "../utils/Faker";
+import { createDbMock } from "../utils/MSWData/CreateDbMock";
+
+import { createSettingsModels } from "./SettingsMock";
+
+export function createModels(faker: Faker) {
+    return {
+        ...createSettingsModels(faker),
```
stephenkao, line 8: I realize we've already talked about this, but I do think it's worth aligning with the singleton pattern that Faker recommends for normal use cases.  After configuration/initialization, we shouldn't be mutating the instance in test cases, and it'd benefit us by not having to instantiate new Faker instances with extended options each time we want to invoke one of these functions.  If we want the ability to pass in a separate Faker instance to each of these functions, I'd at least recommend defaulting the `faker` value to be the default Faker instance created [here](https://github.com/CDCgov/prime-reportstream/pull/8815/files#diff-6ffc1dec4e79ad72a998f64e278625d0e52cd1b2fd098a0498daf00c04afddafR161).  What are your thoughts about this?

PR #8831, opened 2023-03-23
Catch translation errors in UP and record in action log

arnejduranovic:
Handles and records errors that can happen in the translation step of the UP.

Test Steps:
1. See/Run new test case that was added.

## Changes
- If a translation error occurs, record it and go on to next sender instead of just giving up alltogether

### Testing
- [ ] Tested locally?
- [ ] Added tests?

### Process

## Linked Issues
- Fixes 8830
- Fixes 8810

### DIFF
--- a/prime-router/src/main/kotlin/fhirengine/engine/FHIRTranslator.kt
+++ b/prime-router/src/main/kotlin/fhirengine/engine/FHIRTranslator.kt
@@ -18,9 +18,6 @@ import gov.cdc.prime.router.azure.QueueAccess
 import gov.cdc.prime.router.azure.db.Tables
 import gov.cdc.prime.router.azure.db.enums.TaskAction
 import gov.cdc.prime.router.fhirengine.translation.hl7.FhirToHl7Converter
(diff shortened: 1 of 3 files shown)


jack-h-wang:
I am a little concerned that if the error causes the system to move on to the next sender, that will make it more complicated to re-run the unit of data that caused the error after the error is corrected.

arnejduranovic:
> I am a little concerned that if the error causes the system to move on to the next sender, that will make it more complicated to re-run the unit of data that caused the error after the error is corrected.

Hmm I'm not sure about this. If there is an error, you'll fix whatever it is (like let's say a bad filter on one of the senders or a bad schema path) and then you're just going to run the original report again, right? So how does stopping vs continuing after an error occurs matter? Because the first sender in the list isn't always going to be the one that throws the error, it could be a middle one or the last one, right?

victor-chaparro (approved):
Nice catch! Looks good - tested it locally.

victor-chaparro (approved):
Nice catch! Looks good - tested it locally.

Issue #7252, opened 2022-11-04
UP "Convert step" transforms sender data using specified schema


carlosfelix2:
### User Story:
As an administrator, I want to specify which sender transform schema to use for a sender in the sender settings, so I can control the transform process on a per sender basis.

### Description/Use Case
We want the UP to do the sender transform after the data is converted to FHIR, but before it is routed.  The settings allow us to specify a schema we can use in the UP to specify which schema to use for the sender transform.

### Dev Notes:
* Note that the schema name may be checked at some point in the code by the COVID pipeline.  You would need to make sure the schema name is processed correctly for the full-elr topic.

### Acceptance Criteria
* The Universal Pipeline performs sender transformation after converting the submitted data to FHIR.
* The Universal Pipeline performs sender transformation before routing the data to receivers.
* The Universal Pipeline uses a specified schema in a sender's setting to perform the sender transformation.

PR #8833, opened 2023-03-23
Initial integration of fhir transforms into universal pipeline

jimmyfagan:
This PR adds code to FHIRConverter.kt to perform sender transforms. Part of enabling this involved adding a schema name both to the sender configuration and the `RawSubmission` object that gets passed to the message queue. Other than that, things here worked pretty seamlessly.

Test Steps:
1. Pull the branch and add ```      schemaName: metadata/fhir_transforms/fhir-transform-sample``` to a full-elr sender. Then reload settings
2. Run ReportStream locally and send a report to that sender.
3. Either on the receiver side, or in blob storage, look for some representation of the report from after the convert step. Search for `TestValue` to see that the patient name was converted to the value from the transform.

## Changes
- Added `schemaName` to `ELRSender` to track a Fhir transform schema name in sender settings
- In the convert step, before looping through fhir bundles, load the Fhir transformer, and transform the bundles
- Removed `HasSchema` interface (didn't seem to provide any value since it just had a simple property), was formerly used to identify `TopicSender` so updated some of those checks.
- Added tests that verify that `transform` is called during the convert step.

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- Addresses #7252


prime-router/src/main/kotlin/transport/GAENTransport.kt
```diff
@@ -247,7 +247,7 @@ class GAENTransport : ITransport, Logging {
                 // STLTs will handle the content error and request for resubmission.
                 // Then, the new GAEN message from regenerate and resend accordingly.
                 400, 409, 412 -> PostResult.SUCCESS // Bad parameters, Unsupported test type, and UUID already present
-                                                    // can not be rectified by re (consider this a success)
+                // can not be rectified by re (consider this a success)
```
jimmyfagan, line 250:
```suggestion
                                                    // can not be rectified by re (consider this a success)
```

JessicaWNava (approved):
Works and LGTM!

prime-router/src/main/kotlin/SubmissionReceiver.kt
```diff
@@ -256,6 +256,7 @@ class ELRReceiver : SubmissionReceiver {
         payloadName: String?,
         metadata: Metadata?
     ) {
+        check(sender is FullELRSender)
```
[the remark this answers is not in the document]
jimmyfagan: Basically this should only be used by `FullELRSender`s, but we didn't previously need to smart cast it to a `FullELRSender` because everything we needed is in the top level `Sender` definition. With the usage of `schemaName` at line 346, we do need to cast `sender`, and figured rather than do that locally there, we might as well add this check at the entry point to the function and be able to use `sender` as a `FullELRSender` throughout.

prime-router/src/main/kotlin/Sender.kt
```diff
@@ -199,12 +192,15 @@ abstract class Sender(
  *  as the base Sender abstract class, although may be extended / modified in the future.
  */
 class FullELRSender : Sender {
+    var schemaName: String
```
arnejduranovic, line 195: Why not add this to the base `Sender` class? It seems like both FullELR and Topic senders use it?

prime-router/src/main/kotlin/SubmissionReceiver.kt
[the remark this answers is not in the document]
arnejduranovic: See my comment regarding `schemaName`

prime-router/src/test/kotlin/fhirengine/azure/FhirFunctionTests.kt
```diff
@@ -154,8 +154,8 @@ class FhirFunctionTests {
             "\"blobURL\":\"http://azurite:10000/devstoreaccount1/reports/receive%2Fignore.ignore-full-elr%2F" +
             "None-011bb9ab-15c7-4ecd-8fae-0dd21e04d353-20220729171318.hl7\",\"digest\":\"58ffffffaaffffffc22ffffff" +
             "f044ffffff85ffffffd4ffffffc9ffffffceffffff9bffffffe3ffffff8fffffff86ffffff9a5966fffffff6ffffff87fffff" +
-            "fff5bffffffae6015fffffffbffffffdd363037ffffffed51ffffffd3\",\"sender\":\"ignore.ignore-full-elr\"," +
-            "\"blobSubFolderName\":\"ignore.ignore-full-elr\"}"
+            "fff5bffffffae6015fffffffbffffffdd363037ffffffed51ffffffd3\",\"blobSubFolderName\":" +
+            "\"ignore.ignore-full-elr\",\"schemaName\":\"metadata/fhir_transforms/fhir-transform-sample\"}"
```
arnejduranovic, line 158: should `fhir-transform-sample` be in a test resources folder instead?

arnejduranovic (commented):
Do you have an automated test(s) for the "test steps" you outline in the PR description?

prime-router/src/main/kotlin/SubmissionReceiver.kt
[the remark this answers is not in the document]
jimmyfagan: I think that's a good call to pull `schemaName` into the top level `Sender`. I'm inclined to keep this check here either way, so let me know if you think otherwise.

prime-router/src/main/kotlin/Sender.kt
jimmyfagan: Yep, makes sense, and we can just make it nullable or use `""` later if a different child class doesn't use it (though I don't expect additional child classes anyway), will be in an upcoming commit

prime-router/src/test/kotlin/fhirengine/azure/FhirFunctionTests.kt
jimmyfagan: It's actually not even used here because of the mocked functions, I just wanted the queue message to more closely resemble a real thing. I'll update the schemaName here to be `someSchema` to make it more clear that it doesn't need to point to a real thing for this test.

jimmyfagan: But yes, conceptually, I agree that if the test relies on a file, that file should be in a `test` folder.

prime-router/src/test/kotlin/fhirengine/engine/FhirConverterTests.kt
```diff
@@ -223,4 +242,14 @@ class FhirConverterTests {
...
     }
+
+    @Test
+    fun `test getTransformerFromSchema`() {
+        val engine = spyk(makeFhirEngine(metadata, settings, TaskAction.process) as FHIRConverter)
+
+        assertThat(engine.getTransformerFromSchema("")).isNull()
+
+        // Expect a failure since the schema doesn't exist, but this verifies that we tried
+        assertThat { engine.getTransformerFromSchema("some-schema") }.isFailure()
```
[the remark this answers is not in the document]
jimmyfagan: As for the positive case, this test case was meant to just exercise `getTransformerFromSchema`, which in one case returns a clean null, and in the other case calls a constructor for `FhirTransformer` (which leads to the exception). I don't know that a non-failing constructor call needs to be covered here, since all I wanted to verify here is that the constructor does get called. I'll think about this though, I suppose if we tested that on a schema file that does exist, we could avoid the issue of the `.isFailure` specificity altogether.

prime-router/src/main/kotlin/fhirengine/engine/FHIRConverter.kt
```diff
@@ -143,6 +148,18 @@ class FHIRConverter(
...
     }
 
+    /**
+     * Loads a transformer schema with [schemaName] and returns it.
+     * Returns null if [schemaName] is the empty string.
+     * Using this function instead of calling the constructor directly simplifies the process of mocking the
+     * transformer in tests.
+     */
+    internal fun getTransformerFromSchema(schemaName: String): FhirTransformer? {
+        return if (schemaName != "") {
```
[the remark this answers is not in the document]
jimmyfagan: Yea I suppose if for some reason it was set to whitespace other than `""`, we would still want to treat that as being blank rather than fail to read a file called `       `... will update this in an upcoming commit.

jimmyfagan:
@arnejduranovic I think I addressed all of the comments in the code. As for the question of automation of the test steps: that's not currently automated because it would rely on the organization settings and need to be in the smoke/end2end tests, which I don't have much experience editing. I think my preference would be to open a separate ticket to tie-in sender transforms into the smoke tests, but I'm not opposed to working on adding that here if you think that should be part of this ticket. Let me know what you think!

arnejduranovic (approved):

PR #8856, opened 2023-03-27
Engagement/bgreen/check default time before conversion

brick-green:
This PR adds a check to the `TemporalAccessor.toZonedDateTime` function. This checks if the time is set to 00:00:00. If true, then the date is not converted to the receiver timezone. Currently, we're defaulting the time to midnight if a time is not provided in a datetime field. In that case we cannot convert the timezone as the orginial time and zone were not provided.

Test Steps:
1. Log in to prod: `./prime login --env prod`
2. pull settings for CA: `./prime multiple-settings get --filter ca-dph`
3. load settings for CA
    - copy output from above command and copy to `organizations.yml`
    - change transport to: 
    ```transport:
        type: SFTP
        host: sftp
        port: 22
        filePath: ./upload
        credentialName: DEFAULT-SFTP
    ```
    - run `./gradlew reloadSettings`
4. Download file for testing:  
     - `./prime sender-files --message-id 80b38d14-0a4e-4ae9-b --o <PATH/TO/DIR> --env prod`
6. run test file: 
    - `./prime data --input <PATH/TO/FILE> --input-schema upload-covid-19 --output-dir <PATH/TO/DIR> --route`
7. Check that SPM-17 = "20230301000000+0000"

### Testing
- [X] Tested locally?


GilmoreA6 (commented):
Tested and it does what it's supposed to. One thing to consider is that this does potentially present an issue if we receive a future date that will then not be converted. (e.g. we receive a specimen collection date of 3/28 and no time so it gets set to 202303280000+0000. This will cause an issue where CA receives a future date/time.

GilmoreA6 (approved):

oslynn (commented):
Looking good.

Issue #7249, opened 2022-11-04
Support inline value sets in UP Sender Transform schema


carlosfelix2:
### User Story:
As a developer, I want to be able to specify a set of key/value pairs, so data that matches the key is converted to data specified by the value of the pair.

### Description/Use Case
This feature is similar to the existing FHIR to HL7 value sets.  The idea is that we will take FHIR values and convert them to another value.  For example, we may get strings like "PROD" and "DEMO" that are not valid HL7 values and convert them to proper values using the value sets, like to "production" or "test". 

### Dev Notes:
* This feature is the same as in the FHIR to HL7 library, so reuse is possible, but maintain the library code separate from the rest of the code.

### Acceptance Criteria
* The UP Sender Transform can use a valueSet list of key/value pairs to match a FHIR value to a key and set a FHIR property to the matching value of the key/value pair.
* The UP Sender Transform can have one or more key/value pairs in its valueSet list.
* The UP Sender Transforms valueSet can have an empty string for the value in a key/value pair.
* The UP Sender Transforms valueSet does not set the value of a field if there is no matching key/value pair (no op).

jimmyfagan:
Design doc to reference: https://app.zenhub.com/files/304423150/d666073a-0d70-42dc-9779-82f163ae5593/download

PR #8867, opened 2023-03-28
7249 Support inline value sets in UP Sender Transform schema

jack-h-wang:
This PR makes changes to the initial valueset implementation and adds a unit test to verify the valuesets function as expected.

Test Steps:
1. Run the `test transform with value set` unit test and verify the result passes.

## Changes
- Adjust the conditions in which valuesets are evaluated.
- Change the valueset value to be parsed as a string rather than a FHIRPath expression.
- Ensure the original untransformed value is preserved if a valueset key is not matched.
- Added unit testing for valueset.
- Reorganized unit tests and adjusted existing tests to match expected behavior.
- ktLint formatting updates unrelated to this story.

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- #7249


arnejduranovic:
I think it would be appropriate to create a user document for the valueset feature in general. For example, in your PR description, you have: `Adjust the conditions in which valuesets are evaluated.` Wouldn't it be nice if engagement team knew what these were without looking at code? We should document all configuration options, but as part of this ticket we can document "valuesets". 

I propose you add the documentation to this PR for now, and then we can decide where to put it. Thoughts?

jack-h-wang:
> I think it would be appropriate to create user documentation for the valueset feature in general. For example, in your PR description, you have: `Adjust the conditions in which valuesets are evaluated.` Wouldn't it be nice if engagement team knew what these were without looking at code? We should document all configuration options, but as part of this ticket we can document "valuesets".
> 
> I propose you add the documentation to this PR for now, and then we can decide where to put it. Thoughts?

There is already some documentation I was working off of linked in the original issue. The link is here:
https://app.zenhub.com/files/304423150/d666073a-0d70-42dc-9779-82f163ae5593/download

I'll try to edit the PR description to be more clear.

arnejduranovic:
To automatically close and issue when this PR is merged, you should update the description to this syntax:

`- Fixes #7249`

https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

prime-router/src/test/kotlin/fhirengine/translation/hl7/FhirToHl7ConverterTests.kt
```diff
@@ -18,10 +24,217 @@ import io.mockk.mockk
...
+
+        element = ConverterSchemaElement("name", value = listOf("Bundle.timestamp", "Bundle.id"))
+        assertThat(converter.getValue(element, bundle, bundle, customContext)).isEqualTo(IdType(bundle.id))
+
+        element = ConverterSchemaElement("name", value = listOf("Bundle.timestamp", "Bundle.timestamp"))
+        assertThat(converter.getValue(element, bundle, bundle, customContext)).isNull()
+    }
+
+    @Test
+    fun `test get value from value set`() {
```
arnejduranovic, line 182: Given your changes and the existence of `test get value as string from set`, what does this test add? Seems to me we have identical/duplicate tests?

prime-router/src/test/kotlin/fhirengine/translation/hl7/FhirTransformerTests.kt
```diff
@@ -323,6 +324,48 @@ class FhirTransformerTests {
...
+        bundle.addEntry().resource = resource3
+
+        val patientElement = FhirTransformSchemaElement(
+            "patientElement",
+            value = listOf("%resource.id"),
+            resource = "%resource",
+            bundleProperty = "%resource.id",
+            valueSet = sortedMapOf(
+                Pair("abc123", "ghi789"),
+                Pair("def456", "")
```
arnejduranovic, line 348: What happens if there is a pair with `null`. Is this an impossible state? If so, where do we test/verify that?

prime-router/src/test/kotlin/fhirengine/translation/hl7/FhirTransformerTests.kt
```diff
@@ -323,6 +324,48 @@ class FhirTransformerTests {
...
+            schemaRef = childSchema,
+            resourceIndex = "patientIndex",
+        )
+
+        val schema = FhirTransformSchema(elements = mutableListOf(elemA))
+
+        FhirTransformer(schema).transform(bundle)
+
+        assertThat(resource.id).isEqualTo("ghi789")
+        assertThat(resource2.id).isNullOrEmpty()
```
arnejduranovic, line 365: Is this right? if the pair value is "", should the resource be null?

arnejduranovic (commented):
Where do we test that we can actually read and load the valueset from the FHIR transform file?

prime-router/src/test/kotlin/fhirengine/translation/hl7/FhirToHl7ConverterTests.kt
jimmyfagan: Since `getValueAsString` and `getValue` are similar, yet independent, these look identical/duplicate, but are both necessary. However as part of the reorganization of tests out of `ConfigSchemaProcessorTests.kt`, I would suggest moving the few tests using `getValue` to `FhirTransformerTests.kt` (and changing any `FhirToHl7Converter` to `FhirTransformer`, `ConverterSchemaElement` to `FhirTransformSchemaElement`, and `ConverterSchema` to `FhirTransformSchema`) since in practice, `getValue` is geared towards the FHIR transformer and `getValueAsString` is geared towards the FHIR to HL7 Converter.

prime-router/src/test/kotlin/fhirengine/translation/hl7/FhirTransformerTests.kt
jimmyfagan: I think this is just compiler enforced since `valueSet` is of type `SortedMap<String, String>`

prime-router/src/test/kotlin/fhirengine/translation/hl7/FhirTransformerTests.kt
jimmyfagan: The property should be set to the empty string currently, so this test is less restrictive than we might want it to be. I think we should specifically check for the empty string (which might also boost readability when seen next to the other two assertions here).

```suggestion
        assertThat(resource2.id).isEqualTo("")
```

jimmyfagan (commented):
Looks good, left some comments, many of which were replies to conversations Arnej started. Nothing too critical, let me know when this is ready for another look and I should be able to give it a quick approval!

jack-h-wang:
> To automatically close and issue when this PR is merged, you should update the description to this syntax:
> 
> `- Fixes #7249`
> 
> https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

We have been intentionally not doing this because we need the issue to go to Product Review rather than be closed automatically.

jack-h-wang:
> Where do we test that we can actually read and load the valueset from the FHIR transform file?

By FHIR transform file do you mean a sender schema? I think that piece relies on #8833.

prime-router/src/main/kotlin/fhirengine/translation/hl7/ConfigSchemaProcessor.kt
jimmyfagan, line 1: > > Where do we test that we can actually read and load the valueset from the FHIR transform file?
> 
> By FHIR transform file do you mean a sender schema? I think that piece relies on #8833.

Sorry missed this comment earlier, making a comment on a file to make it easier to thread discussion of this.. It doesn't rely on 8833 necessarily, you could probably adjust tests in `ConfigSchemaReaderTests.kt` to make sure that one of those tests includes verification that valueSets can be read from a FhirTransformer schema. I think that would be important to include here just to make sure that when we get to the point of merging in both this and 8833, we'll know that all of the pieces of this process work independently.

prime-router/src/test/kotlin/fhirengine/translation/hl7/FhirTransformerTests.kt
jack-h-wang: Correct, this is immediately flagged as a type mismatch in the IDE if you attempt to define either the key or the value as anything other than `String`.

prime-router/src/main/kotlin/fhirengine/translation/hl7/ConfigSchemaProcessor.kt
jack-h-wang: The existing test already parsed the value set in `sample_schema.yml` but I found that it was possible to import null instead of a string, which would result in a null reference exception if any attempt was made to access the null value. I added a check for null values so that this would cause schema validation to fail if it occurs, and added an additional test schema to exercise this.

jimmyfagan (approved):
Looks great!

arnejduranovic (approved):

PR #8893, opened 2023-03-30
Experience-8170: Update validate endpoint to return 200 for strictly validation errors

### LINKED ISSUE #8170: Update API endpoint to still return 200 for validation errors
## Is your feature request related to a problem? Please describe.
Currently, the `/validate` endpoint returns a 400 if there are validation errors in addition to real errors.  We should fix this so only real errors are handled as such.  This should be done before the validation page goes public so we don't see an unnecessary spike in request errors.

## Describe the solution you'd like
The `/validate` endpoint should return 200s if all of the "errors" are only specifically validation errors -- meaning there's something off about the sent file.  In this case, the validation logic is still working as intended so a 200 should be expected then.

## Acceptance Criteria
- [ ] /validate endpoint returns 200s if all of the "errors" are only specifically validation errors
- [ ] /validate endpoint continues to return 400s if there are real, non-validation errors
- [ ] Platform team has reviewed the code and approved

stephenkao:
This changeset updates the validation endpoint to still return 200 if there are only validation errors.  We should still continue to return 4xx/5xx for _real_ errors, but catching validation errors is unexceptional logic for the validator -- that's working as expected.

Test Steps:
1. Submit a file with only errors to the /validate endpoint, expect:
    - 200 response code is received
    - `overallStatus` to equal `Error`
    - `errors` to be populated with expected errors
    - `warnings` to be empty

2. Submit a file with only warnings to the /validate endpoint, expect:
    - 200 response code is received
    - `overallStatus` to equal `Valid`
    - `errors` to be empty
    - `warnings` to be populated with expected warnings

3. Submit a file with both errors and warnings to the /validate endpoint, expect:
    - 200 response code is received
    - `overallStatus` to equal `Error`
    - `errors` to be populated with expected errors
    - `warnings` to be populated with expected warnings

Example Postman request with only errors (1 above):
Screenshot 2023-03-30 at 10 30 00

### Testing
- [x] Tested locally?
<strike>- [ ] (For Changes to /frontend-react/...) Ran `npm run lint:write`? </strike>
- [x] Added tests?

## Linked Issues
- Fixes #8170

### DIFF
--- a/prime-router/docs/api/validate.yml
+++ b/prime-router/docs/api/validate.yml
@@ -17,7 +17,7 @@ paths:
     post:
       tags:
         - validate
-      summary: Validate a message using client information
+      summary: Validate a message using client information. Validation errors and warnings are reported in the response json.
       parameters:
         - in: header
           name: client
--- a/prime-router/src/main/kotlin/azure/ValidateFunction.kt
+++ b/prime-router/src/main/kotlin/azure/ValidateFunction.kt
@@ -108,11 +108,10 @@ class ValidateFunction(
                     validatedRequest.routeTo,
                     allowDuplicates,
                 )
-                // return OK status, report validation was successful
                 HttpStatus.OK
             } catch (e: ActionError) {
                 actionHistory.trackLogs(e.details)
-                HttpStatus.BAD_REQUEST
+                HttpStatus.OK
             } catch (e: IllegalArgumentException) {
                 actionHistory.trackLogs(
                     ActionLog(InvalidReportMessage(e.message ?: "Invalid request."), type = ActionLogLevel.error)
@@ -162,7 +161,7 @@ class ValidateFunction(
         }
 
         // set status for validation response
-        submission.overallStatus = if (httpStatus == HttpStatus.BAD_REQUEST)
+        submission.overallStatus = if (httpStatus == HttpStatus.BAD_REQUEST || submission.errorCount > 0)
             DetailedSubmissionHistory.Status.ERROR
         else
             DetailedSubmissionHistory.Status.VALID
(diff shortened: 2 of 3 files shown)


snesm (changes requested):
200 appears to be what the [spec](https://github.com/CDCgov/prime-reportstream/blob/master/prime-router/docs/api/validate.yml) implies already.

A little more usage guidance in the spec `summary:` might be appropriate.

stephenkao:
> 200 appears to be what the [spec](https://github.com/CDCgov/prime-reportstream/blob/master/prime-router/docs/api/validate.yml) implies already.
> 
> A little more usage guidance in the spec `summary:` might be appropriate.

The issue is that we're currently returning 400s whenever we have validation errors, which isn't correct because the validation logic is working as expected -- i.e., we _should_ return a 200 since this is an expected/unexceptional case, but we're not.  I'm happy to update the summary, but I'd argue our logic right now needs to be corrected as proposed in this changeset.

arnejduranovic (approved):

stephenkao:
@snesm Is this good to go from your end as well?  Just checking!

snesm:
> @snesm Is this good to go from your end as well? Just checking!

@stephenkao the logic fix is great.  I am suggesting [this](https://github.com/CDCgov/prime-reportstream/pull/8897) minor addition to your branch.

stephenkao:
> > @snesm Is this good to go from your end as well? Just checking!
> 
> @stephenkao the logic fix is great. I am suggesting [this](https://github.com/CDCgov/prime-reportstream/pull/8897) minor addition to your branch.

Looks good to me!  Thank you!

snesm (approved):

PR #8954, opened 2023-04-03
Experience/8943/mdx general layout

### LINKED ISSUE #8943: Add generalizable sidenav/content MDX layout
## Problem statement

With MDX support, we'll be able to generalize our documentation pages through reusable layouts rather than the existing IATemplate logic.

## What you need to know

- This is intended to replace the existing [IATemplate logic](https://github.com/CDCgov/prime-reportstream/blob/master/frontend-react/src/components/Content/Templates/IATemplate.tsx#L24-L63) in the future

## Acceptance criteria

- [ ] There are MDX templates for sidebar navigation and page content
    - [ ] Nice-to-have: The templates are generalizable for any content page

## Follow-ups
- Convert existing Markdown pages to MDX
    - [#8952](https://app.zenhub.com/workspaces/web-experience-607d9d5e68b95200150fec37/issues/gh/cdcgov/prime-reportstream/8952)
    - [#8953](https://app.zenhub.com/workspaces/web-experience-607d9d5e68b95200150fec37/issues/gh/cdcgov/prime-reportstream/8953)

jpandersen87:
Closes #8943

This PR:
- adds a MarkdownLayout component for use with MDX components
  - It accepts the main MDX component article by natural component composition and an optional sidenav article via the sidenav prop
  - The layout uses a uswds grid-row to align the two articles side-by-side
- moves the internal/external link logic from the MarkdownRSLink component into a USSmartLink component that can be used by both the older markdown implementation and the new MarkdownLayout component
- adds a SideNavItem component
  - Provides the parent navlink and children sidenav element as siblings for use in the items prop of SideNav
  - Allows the children subnav to be visible when that path is active by default
- Updates USNavLink logic to include path hash in determining being active (/foo#bar and /foo will be considered active but not /foo#other vs current behavior)


frontend-react/src/components/Content/MarkdownLayout.tsx
```diff
@@ -0,0 +1,34 @@
...
+export interface MarkdownLayoutProps {
+    sidenav?: JSX.Element;
+    children?: React.ReactNode;
+}
+
+export function MarkdownLayout({ sidenav, children }: MarkdownLayoutProps) {
+    return (
+        <MDXProvider
+            components={{
+                a: USSmartLink,
```
stephenkao, line 16: 🙌 🙌 🙌

frontend-react/src/components/Content/MarkdownLayout.tsx
```diff
@@ -0,0 +1,34 @@
...
+            }}
+        >
+            <GridContainer>
+                <GridRow>
+                    {sidenav ? (
+                        <nav
+                            aria-label="side-navigation"
+                            className="grid-col-auto"
```
stephenkao, line 24: Should this take up a set number of columns rather than auto?  For example, the Product pages use four columns for the side nav and eight columns for the main content on tablet+ and just stack on smaller viewports.

frontend-react/src/components/Content/MarkdownLayout.tsx
```diff
@@ -0,0 +1,34 @@
...
+                        >
+                            {sidenav}
+                        </nav>
+                    ) : undefined}
+                    <main className="grid-col">{children}</main>
```
stephenkao, line 29: Same point here about taking up a set number of columns.

frontend-react/src/components/Content/MarkdownRSLink.tsx
```diff
@@ -21,39 +21,8 @@ export type MarkdownComponentProps<
...
-
-    if (props.href !== undefined) {
-        isExternal = isExternalUrl(props.href);
-    }
-
-    if (isExternal) {
-        return <USExtLink {...props}>{children}</USExtLink>;
-    }
-    return <USLink {...props}>{children}</USLink>;
+export const MarkdownRSLink = (props: MarkdownRSLinkProps) => {
```
stephenkao, line 26: Can we just remove this component entirely since it's just an alias to USSmartLink now?

frontend-react/src/components/Grid.tsx
```diff
@@ -0,0 +1,10 @@
+import classNames from "classnames";
+
+export interface GridRowProps {
+    children: React.ReactNode;
+    className?: string;
+}
+
+export function GridRow({ children, className }: GridRowProps) {
+    return <div className={classNames("grid-row", className)}>{children}</div>;
```
stephenkao, line 9: Isn't this functionality already provided by [the Grid component](https://github.com/trussworks/react-uswds/blob/main/src/components/grid/Grid/Grid.tsx#L61)?

```tsx
<Grid className={ className } row>
    { children }
</Grid>
```

frontend-react/src/__mocks__/mdxjsReactMock.tsx
```diff
@@ -0,0 +1,7 @@
+interface MDXProviderProps {
+    children: React.ReactNode;
+}
+
+export function MDXProvider({ children }: MDXProviderProps) {
+    return <>{children}</>;
```
stephenkao, line 6: Simple simple 🙏

frontend-react/src/components/SideNav.test.tsx
```diff
@@ -0,0 +1,118 @@
+import { screen } from "@testing-library/react";
+
+import { renderApp } from "../utils/CustomRenderUtils";
+
+import { SideNavItem } from "./SideNav";
+import { USNavLink } from "./USLink";
+
+// link -> li -> sidenav (ul) -> div
+function getIsVisible(e: HTMLElement) {
+    return !e.parentElement?.parentElement?.parentElement?.classList.contains(
```
stephenkao, line 10: This feels fragile since it's so tightly coupled to an exact element hierarchy.  I'd recommend one of the following:

- Omit the entire [subnav](https://github.com/CDCgov/prime-reportstream/pull/8954/files#diff-0b6f9b990aac3ec98d4b3517114cacc1b3fff66b869546a5b049046ad5947be0R50-R54) OR add an `aria-hidden`/`hidden` if `isSubnavVisible` is `true`.  This would result in the query returning nothing for this link so we could run an expectation on that instead.
- Iterate upward through `parentElement`s until we find an ancestor element with a `display-none` class (since that should affect all descendants).  The other option is more preferable in my opinion since it's less reliant on a modifier class, though.

frontend-react/src/components/SideNav.tsx
```diff
@@ -0,0 +1,57 @@
...
+    const isSubnavVisible =
+        isActive !== undefined
+            ? isActive
+            : href === undefined ||
+              href === "" ||
+              pathname === href ||
+              pathname.startsWith(href);
+    const subnavClassname = !isSubnavVisible ? "display-none" : "";
+
+    // SideNav doesn't allow custom classes so we have to wrap in a div 😡
```
stephenkao, line 44: Honestly, I don't think it's a huuuuuge deal if we just want to build our own since [their SideNav](https://github.com/trussworks/react-uswds/blob/main/src/components/SideNav/SideNav.tsx) is so simplistic anyway.  Let's hold off on that and reconsider if it becomes an issue going forward, though!

frontend-react/src/components/SideNav.tsx
```diff
@@ -0,0 +1,57 @@
...
+    isActive?: boolean;
+}
+
+/**
+ * Wrapper component that provides the NavLink and SideNav sibling elements (if items
+ * provided). The parent navlink can be customized by passing a functional component
+ * or html element string as the "to" prop (defaults to USNavLink). The SideNav sibling
+ * can be forced visible via the "isActive" prop (otherwise defaults to checking path).
+ */
+export function SideNavItem({
```
stephenkao, line 26: Can we `export default` here?  I'm thinking there should always be a default export in component files to encourage us to have a 1:1 between files and components.  There'll be times that we'll have "sub-components" of course, but that's something we should try to avoid whenever possible.

frontend-react/src/components/SideNav.tsx
```diff
@@ -0,0 +1,57 @@
+import { SideNav } from "@trussworks/react-uswds";
```
stephenkao, line 1: Can we add this file to a new `src/shared` directory for common components?  cc @etanb who's doing a similar thing with the new Table component.

frontend-react/src/components/USLink.tsx
```diff
@@ -167,18 +167,64 @@ export const USNavLink = ({
...
+        );
+    } catch (e: any) {
+        return false;
+    }
+}
+
+export interface USSmartLinkProps
+    extends React.AnchorHTMLAttributes<HTMLAnchorElement> {}
+
+export function USSmartLink({ children, ...props }: USSmartLinkProps) {
```
stephenkao, line 219: 🤓

stephenkao (changes requested):
Added some points that I think should be addressed before merging, but this is looking great overall!

frontend-react/src/components/Grid.tsx
jpandersen87: Could be! If so, will drop this in favor of it. Thanks for the heads up!

frontend-react/src/components/Content/MarkdownLayout.tsx
```diff
@@ -0,0 +1,34 @@
...
+}
+
+export function MarkdownLayout({ sidenav, children }: MarkdownLayoutProps) {
+    return (
+        <MDXProvider
+            components={{
+                a: USSmartLink,
+            }}
+        >
+            <GridContainer>
```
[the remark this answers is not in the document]
jpandersen87: I was expecting that the usage of GridContainer here to be temporary until it's implemented further up the render tree.

frontend-react/src/components/Content/MarkdownLayout.tsx
```diff
@@ -0,0 +1,74 @@
...
+        to?:
+            | string
+            | React.FunctionComponent<React.HTMLAttributes<HTMLElement>>;
+    };
+}
+
+/**
+ * Default markdown layout.
+ * With sidenav:
+ *  ________________________
```
stephenkao, line 24: 🧑‍🎨

frontend-react/src/components/Content/MarkdownLayout.tsx
```diff
@@ -0,0 +1,74 @@
...
+ * |    |     MAIN    |     |
+ * |    |             |     |
+ * |    |             |     |
+ * --------------------------
+ * |________________________|
+ */
+export function MarkdownLayout({
+    sidenav,
+    children,
+    mainProps: { to: Main = "main", ...mainProps } = {},
```
stephenkao, line 47: 🙏

frontend-react/src/components/Content/MarkdownLayout.tsx
```diff
@@ -0,0 +1,74 @@
...
+            components={{
+                a: USSmartLink,
+            }}
+        >
+            <GridContainer>
+                <Grid row>
+                    {sidenav ? (
+                        <Nav
+                            aria-label="side-navigation"
+                            className="tablet:grid-col-4"
```
stephenkao, line 61: 🙏

stephenkao (approved):
Looks great to me!  Thanks for setting this up!

penny-lischer (approved):
LGTM

PR #8993, opened 2023-04-05
Experience-8202: Add proposal for website-API communication

### LINKED ISSUE #8202: Proposal for consolidated API response payload structure
## Problem statement

We're sending back top-level arrays for collection-oriented responses (`/deliveries` and `/submissions` endpoints in particular), and we can update this to return an object so that we can include collection metadata like entity type and pagination information. 

## What you need to know

- See [this Slack conversation](https://www.slack.com/archives/C01HFHPNJEL/p1670348915918859) for more information.

## Acceptance criteria

- [ ] A proposal doc is reviewed and approved by Platform team

stephenkao:
This changeset adds a doc that proposes communication patterns between the website and the API layer.

## Linked Issues
- Fixes #8202

### DIFF
--- a/prime-router/docs/proposals/0019-website-api-payload-structure.md
+++ b/prime-router/docs/proposals/0019-website-api-payload-structure.md
@@ -8,8 +8,9 @@ As we continue to develop more features for the ReportStream website, we're inev
 
 The purpose of this proposal is to align on communication between ReportStream API and the website, including:
 - HTTP response codes
-- HTTP methods
 - Req/res structures
+    - Resource structures
+    - Collection structures
 
 Having a consistent approach to the above would provide some predictability in implementation as we continue to build out the website, and it would help to simplify logic in common UX patterns like search filters and pagination.  (For example, UsePagination.ts contains [visual slot logic](https://github.com/CDCgov/prime-reportstream/blob/master/frontend-react/src/hooks/UsePagination.ts#L26-L84) that could be drastically simplified if there were, say, a `totalCount` included in the response payload.)
 
@@ -81,10 +82,6 @@ The following is a non-exhaustive list of HTTP response codes that should common
 | 500  | Internal Server Error | Unhandled exception                                                                              | Catch-all for all unhandled exceptions                             |
 
 NOTE: This table does not include common HTTP response codes commonly used as navigation signals, such as `301 Moved Permanently` or `308 Permanent Redirect`, which are not currently applicable to the website due to its single-page application (SPA) approach.  This may change in the future if we migrate to a more traditional website architecture.
-
-### HTTP methods
-
-There are no proposed changes to how we're using HTTP methods at the moment.  However, for all mutating methods (`POST`, `PUT`, `PATCH`, `DELETE`), we should return the entire modified record with all data through a `200 OK` rather than subsets of data or a `204 No Content`.  This is slightly atypical behavior, but it helps simplify request bookkeeping on the website.
 
 ### Req/res structures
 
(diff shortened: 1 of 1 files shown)


stephenkao:
@arnejduranovic No rush on this at all!  It's just what we were discussing re: establishing communication patterns between the ReportStream website + API.  Let me know if this is missing anything!

prime-router/docs/proposals/0019-website-api-payload-structure.md
```diff
@@ -0,0 +1,332 @@
...
+| 404  | Not Found             | Error; requested resource does not exist                                                         | Admin requests Receivers for a non-existent Organization           |
+| 500  | Internal Server Error | Unhandled exception                                                                              | Catch-all for all unhandled exceptions                             |
+
+NOTE: This table does not include common HTTP response codes commonly used as navigation signals, such as `301 Moved Permanently` or `308 Permanent Redirect`, which are not currently applicable to the website due to its single-page application (SPA) approach.  This may change in the future if we migrate to a more traditional website architecture.
+
+### HTTP methods
+
+There are no proposed changes to how we're using HTTP methods at the moment.  However, for all mutating methods (`POST`, `PUT`, `PATCH`, `DELETE`), we should return the entire modified record with all data through a `200 OK` rather than subsets of data or a `204 No Content`.  This is slightly atypical behavior, but it helps simplify request bookkeeping on the website.
```
arnejduranovic, line 87: I think we should follow the spec wherever possible, and I don't think "request bookkeeping" is a strong enough reason to deviate from well-established practices? For example, the [latest spec](https://www.rfc-editor.org/rfc/rfc9110#name-post) I can find explains what a POST should return:


> If one or more resources has been created on the origin server as a result of successfully processing a POST request, the origin server SHOULD send a [201 (Created)](https://www.rfc-editor.org/rfc/rfc9110#status.201) response containing a [Location](https://www.rfc-editor.org/rfc/rfc9110#field.location) header field that provides an identifier for the primary resource created ([Section 10.2.2](https://www.rfc-editor.org/rfc/rfc9110#field.location)) and a representation that describes the status of the request while referring to the new resource(s).


Let me know if we should get together to discuss this further

prime-router/docs/proposals/0019-website-api-payload-structure.md
```diff
@@ -0,0 +1,332 @@
...
+    "property2": ...,
+    "property3": ...,
+    // other resource data     
+    ...
+}
+```
+
+##### Example
+
+`GET /api/settings/organizations/ignore`
```
arnejduranovic, line 113: > `resources` is the entity view, facade, or database table

This definition is not quite correct, right? If we look at the example of `GET /api/settings/organizations/ignore`, which you say is okay, "setting" is the table in the database and "organizations" is not anything, its just a logical concept. So we need to either clarify this definition (which I am in favor of) or we say this example is not correct (which I am not in favor of). I think what we should improve in this example is singularize "settings" and "organizations"

arnejduranovic (changes requested):
Looks good to me overall from a Platform perspective, added a couple comments.

prime-router/docs/proposals/0019-website-api-payload-structure.md
stephenkao: Ah, you're right; `201 Created` is correct for resource creation via `POST`, and I don't think we should deviate from this so I'll fix that.  My emphasis here was more about returning the entire record rather than subset of data specifically for `PUT`/`PATCH`/`DELETE`.  If it's a strong preference to not do this from the Platform team, that's fine -- it's only a small benefit to the WebX team and we can work around this.

prime-router/docs/proposals/0019-website-api-payload-structure.md
stephenkao: Fair point!  This is based on the current [SettingsFunction.kt](https://github.com/CDCgov/prime-reportstream/blob/master/prime-router/src/main/kotlin/azure/SettingsFunctions.kt), which reads into an OrganizationAPI from `getOne` [here](https://github.com/CDCgov/prime-reportstream/blob/master/prime-router/src/main/kotlin/azure/SettingsFunctions.kt#L59).  BUT you raise a good point -- since settings are overloaded for different types (Organization, Sender, Receiver), I'm thinking it'll just be better to choose a different entity type altogether for clarity in this doc.

Regarding the singularization, are you proposing that we do something like `GET /api/setting/organization/ignore`?  If so, I disagree -- my feeling is that we should be aligning with standard REST practices regarding resources/collections so we can have a predictable pattern:

- `GET /resources`: fetch a list of resources
- `GET /resources/<resource_id>`: fetch a single resource
- `POST /resources`: create a new resource
- `(PUT|PATCH) /resources/<resource_id>`: update a single resource
- `DELETE/resources/<resource_id>`: delete a single resource

arnejduranovic: re singularization: never mind, you are correct :)

prime-router/docs/proposals/0019-website-api-payload-structure.md
```diff
@@ -0,0 +1,332 @@
...
+    ]
+}
+```
+
+#### Sorted collections
+
+Sorted collection request paths should take the form of `*/<resources>?sortDir=<sortDir>&sortColumn=<sortColumn>`, where:
+
+- `sortDir` is the order of sorting (`ASC` or `DESC`)
+- `sortColumn` is the name of the sorted column (e.g., `id`, `name`, `expiresAt`)
```
jpandersen87, line 292: Is there potential for a need to accept multiple columns for sortColumn in the future? Or any property for that matter. Should we establish the the syntax rules for this scenario (ex: comma separated)?

penny-lischer (approved):
Nice!

prime-router/docs/proposals/0019-website-api-payload-structure.md
stephenkao: Great point, I think we should use comma-separated values for this with descending priority -- will update the proposal!  Thanks for calling this out!

penny-lischer (approved):
LGTM

arnejduranovic (approved):
This looks good to me at the moment. Once we start referencing this in the future, it's possible we would propose changes, but I don't suspect anything major.

PR #9001, opened 2023-04-05
Update MDX loader to ignore md files

### LINKED ISSUE #9000: MDX loader webpack config includes .md files, breaking existing pages
The mdx loader is configured to target both .md and .mdx files. This is causing some pages based on those md files to break. The configuration needs to be updated to target only mdx.

jpandersen87:
Fixes #9000

This PR updates the mdx loader config for webpack to target only mdx files and not md so they can be consumed by react-markdown as normal.


stephenkao (approved):
Ah, good catch!

PR #9003, opened 2023-04-05
Engagement/angela/8644 soap filename null

### LINKED ISSUE #8644: SOAP recording filename "null" on success
## Describe the bug
When updating the ActionHistory upon a successful send in `SoapTransport.kt` the filename is recorded as "null". This can be seen in Metabase by filtering the Action table by Action Name 'send' and Action Result contains "SOAP".

## Impact on ReportStream
Success of the filename is not being properly logged, which makes debugging and troubleshooting more difficult.

## Steps to reproduce
I'm unsure how to manually reproduce this locally, but you should be able to use the `SoapTransportIntegrationTests.kt` to debug the error.

## Expected behavior
The filename that is sent is properly added into the ActionHistory

adegolier:
This PR ...

Test Steps:
I'm not sure how to properly test this, but I ran the SoapTransportIntegrationTests on debug and was able to see that a report id is accessible at that point whereas filename was null. These don't seem to actually go to the database, though.

## Changes
- Changed where it uses "filename" to use "report id" when logging a successful SOAP transport to be consistent with the logging of failed transports and make it easier to track what happened with a report.

### Testing
- [ ] Tested locally? -Sort of, as stated in Test Steps
- [ ] Added tests?

### Process

## Linked Issues
- Fixes #8644


prime-router/src/main/kotlin/transport/SoapTransport.kt
```diff
@@ -129,8 +129,10 @@ class SoapTransport(private val httpClient: HttpClient? = null) : ITransport {
         val receiver = header.receiver ?: error("No receiver defined for report ${header.reportFile.reportId}")
         // get the external file name to send to the client, if we need it
         val fileName = header.reportFile.externalName
+        // get the report id from the header
+        val headerReportId = header.reportFile.reportId
```
brick-green, line 133: Looks like the reportID is passed to the send function as property: **sentReportId**. Could you use that instead of creating a new variable?

adegolier: I could, and in theory it _should_ be the same, but there's no way for me to know if someone is passing in a different report id there for whatever reason. Part of the reason I went with creating the new variable is because on (formerly line 133) the existing code was pulling from the header.

brick-green: Oh yeah, it looks like there's a mixture of uses. I see the property used in actionHistory.trackSentReport() and actionHistory.trackItemLineages(). But then the header is used again on line 186. Seems like this is an opportunity to align on usage throughout the function. Then if you decide to only use the header version then property should just be removed.

adegolier: Can't remove the property because it's an overridden function.

jalbinson (approved):

PR #8690, opened 2023-03-13
Create Settings API area hooks

jpandersen87:
Closes #8366 

This PR renames existing hooks and creates the remaining hooks for the settings API area. The following settings entities have corresponding list/create/read/update/delete hooks:
- Organization
- Sender
- Receiver

PR #9007, opened 2023-04-05
8398 Upload public key

### LINKED ISSUE #8398: [Manage Public Key] Upload Key and Fail/Success message

penny-lischer:
This PR ...

Allows a sender/organization to upload a public key (.pem). If the sender already has a public key saved,  they will be prompted to upload a new one. 

**Test Steps**:
1. Log into RS as a sender or admin
2. Navigate to /resources/manage-public-key
3. Create two keys on your machine in order to test error and success scenarios (see below)
4. Navigate back to /resources/manage-public-key after you have successfully uploaded a key and you should see the screen that your key has already been configured.

**Create a Key on your machine**:
1. In your Terminal run the following command to create a key that will display an error
`openssl genrsa -out my-rsa-keypair.pem 2048`
2. In your Terminal run the following command to create a key that will display a success
`openssl rsa -in my-rsa-keypair.pem -outform PEM -pubout -out my-rsa-public-key.pem`

## Changes
- All changes should reflect Figma here - https://www.figma.com/file/6mwI5ac6rprACKDzDo4Ady/ReportStream-Workspace-%7C-2023?node-id=2167-21811&t=VJ0iDBOEGssSQHvN-0

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- Fixes #8398 

## To Be Done
- #issue


frontend-react/src/components/ManagePublicKey/ManagePublicKey.test.tsx
```diff
@@ -73,8 +68,8 @@ describe("ManagePublicKey", () => {
                 renderApp(<ManagePublicKey />);
             });
 
-            test("renders ManagePublicKeyChooseSender", () => {
-                expect(screen.getByText(/Manage Public Key/)).toBeVisible();
+            test.skip("renders ManagePublicKeyChooseSender", () => {
```
penny-lischer, line 71: Skipped for now since there is an issue with re-render of ManagePublicKeyChooseSender

frontend-react/src/components/ManagePublicKey/ManagePublicKeyChooseSender.tsx
jpandersen87, line 1: This component should be dumber and accept the list of senders as a prop.

frontend-react/src/components/ManagePublicKey/ManagePublicKeyUpload.tsx
jpandersen87, line 1: This component should be dumber and accept the public key info and kid as props.

frontend-react/src/components/ManagePublicKey/ManagePublicKey.tsx
```diff
@@ -82,50 +91,77 @@ function ManagePublicKeySwitchDisplay() {
...
                 />
             )}
-            {sender && !fileSubmitted && (*/}
-            {!fileSubmitted && (
+            {showPublicKeyConfigured && (
+                <ManagePublicKeyConfigured
+                    onUploadNewPublicKey={() => setUploadNewPublicKey(true)}
+                />
+            )}
+            {uploadNewPublicKey && (
```
jpandersen87, line 142: It should be possible to rely on the ManagePublicKeyUpload component entirely for the "already configured" messaging once ManagePublicKeyUpload is made dumber and takes the public key info as a prop.

frontend-react/src/components/ManagePublicKey/ManagePublicKey.tsx
```diff
@@ -82,50 +91,77 @@ function ManagePublicKeySwitchDisplay() {
...
+                            If you need more information on generating your
+                            public key, reference page 7 in the{" "}
+                            <USLink href="/resources/programmers-guide">
+                                API Programmer’s Guide.
+                            </USLink>
+                        </span>
+                    </SiteAlert>
+                </>
+            )}
             {sender.length === 0 && (
```
jpandersen87, line 132: Once the ManagePublicKeyChooseSender is made dumber to accept senders, we could check to see if there are more than 1 senders for the org and show this component. If there is only one, we could technically automatically set it as the upload sender.

frontend-react/src/components/ManagePublicKey/ManagePublicKey.tsx
```diff
@@ -82,50 +91,77 @@ function ManagePublicKeySwitchDisplay() {
...
+        hasPublicKey &&
+        !uploadNewPublicKey &&
+        !fileSubmitted;
+    const showUploadMsg =
+        (sender.length > 0 && !fileSubmitted && !hasPublicKey) ||
+        (!uploadNewPublicKey && sender.length === 0);
+    const showUpload =
+        (sender.length > 0 && !fileSubmitted && !hasPublicKey) ||
+        uploadNewPublicKey;
+    const showUploadError = fileSubmitted && !isUploading && !isSuccess;
```
jpandersen87, line 105: You should format the boolean names here to be is/hasX instead of showX. Also, an empty string is considered falsey and can be used directly in boolean rules.

frontend-react/src/components/ManagePublicKey/ManagePublicKey.tsx
```diff
@@ -1,34 +1,42 @@
...
 export const CONTENT_TYPE = "application/x-x509-ca-cert";
 export const FORMAT = "PEM";
 
-function ManagePublicKeySwitchDisplay() {
-    // const [sender, setSender] = useState("");
+export function ManagePublicKey() {
+    const [hasPublicKey, setHasPublicKey] = useState(false);
+    const [uploadNewPublicKey, setUploadNewPublicKey] = useState(false);
+    const [sender, setSender] = useState("");
+    const [showBack, setShowBack] = useState(false);
```
jpandersen87, line 26: You should use is/hasX naming instead of showX.

frontend-react/src/components/ManagePublicKey/ManagePublicKeyUpload.tsx
```diff
@@ -9,25 +9,54 @@ import {
...
+import { useOrganizationPublicKeys } from "../../hooks/UseOrganizationPublicKeys";
+import { useSessionContext } from "../../contexts/SessionContext";
+import { ApiKey } from "../../config/endpoints/settings";
+
 export interface ManagePublicKeyUploadProps {
+    onFetchPublicKey: (hasPublicKey: boolean) => void;
     onPublicKeySubmit: (e: React.FormEvent<HTMLFormElement>) => void;
     onFileChange: (e: React.ChangeEvent<HTMLInputElement>) => void;
+    onBack: () => void;
+    showBack: boolean;
```
jpandersen87, line 21: is/hasX vs showX

frontend-react/src/components/ManagePublicKey/ManagePublicKeyChooseSender.tsx
```diff
@@ -5,7 +5,7 @@ import { useOrganizationSenders } from "../../hooks/UseOrganizationSenders";
 import Spinner from "../Spinner";
 
 export interface ManagePublicKeyChooseSenderProps {
-    onSenderSelect: (sender: string) => void;
+    onSenderSelect: (sender: string, showBack: boolean) => void;
```
jpandersen87, line 8: is/hasX vs showX

frontend-react/src/components/ManagePublicKey/ManagePublicKey.test.tsx
jpandersen87: Can you elaborate here?

frontend-react/src/hooks/UseOrganizationPublicKeys.ts
```diff
@@ -0,0 +1,36 @@
...
+    const memoizedDataFetch = useCallback(
+        () =>
+            authorizedFetch(publicKeys, {
+                segments: {
+                    organizationName: activeMembership?.parsedName!!,
+                },
+            }),
+        [activeMembership?.parsedName, authorizedFetch]
+    );
+    const { data, isLoading } = rsUseQuery(
```
jpandersen87, line 26: Return the whole object, don't pick out of it.

frontend-react/src/config/endpoints/settings.ts
```diff
@@ -5,6 +5,7 @@ export enum ServicesUrls {
     SENDERS = "/settings/organizations/:orgName/senders",
     SENDER_DETAIL = "/settings/organizations/:orgName/senders/:sender",
     RECEIVERS = "/settings/organizations/:orgName/receivers",
+    PUBLIC_KEYS = "/settings/organizations/:organizationName/public-keys",
```
jpandersen87, line 8: we should stick to the :orgName pattern

jpandersen87 (commented):
Looking good. Just some concerns on component smartness, boolean variable naming, and the skipped tests.

frontend-react/src/components/ManagePublicKey/ManagePublicKey.test.tsx
stephenkao: It looks like two instances of ManagePublicKey are getting rendered in the test so `getByText`/`getByTestId` are returning two duplicate elements.  This is likely due to the two renders:
  - https://github.com/CDCgov/prime-reportstream/pull/9007/files#diff-7cb838f0917dababbf6a240e9428c5683ccda98ebb9d01a2a423754a37e99a58R58
  - https://github.com/CDCgov/prime-reportstream/pull/9007/files#diff-7cb838f0917dababbf6a240e9428c5683ccda98ebb9d01a2a423754a37e99a58R68

We should only have one render call per example OR we should be cleaning up between test runs.  (We should be cleaning up between test runs anyway, but I mean doing it explicitly if we're going to be rendering multiple times in an example.  This is much less preferable than just having one render call per example.)

frontend-react/src/components/ManagePublicKey/ManagePublicKey.tsx
```diff
@@ -82,50 +91,77 @@ function ManagePublicKeySwitchDisplay() {
         }
     };
 
+    const showPublicKeyConfigured =
+        sender.length > 0 &&
```
stephenkao, line 95: Could be variable overload, but I feel like it'd help to add some variable aliases for some of these pieces.  Like `sender.length > 0` and `(sender.length > 0 && !fileSubmitted && !hasPublicKey)` 🤷 

Might also be a good idea to infer an index in the steps based on the current state.  I don't think we should be passing arbitrary numbers to set the index, but we could do something like [what we're doing in FileHandler](https://github.com/CDCgov/prime-reportstream/blob/master/frontend-react/src/components/FileHandlers/FileHandler.tsx#L27), i.e., setting up an array of steps and inferring the index from the state like we discussed in a previous pull request.  Doesn't have to be part of this change, but I can imagine that we'll want to reuse this stepwise logic in the future so it'd be good to get a sense of what its capabilities should be.

frontend-react/src/components/ManagePublicKey/ManagePublicKey.tsx
stephenkao: Agreed -- I don't think that needs to be a separate component on its own.

frontend-react/src/components/ManagePublicKey/ManagePublicKey.tsx
```diff
@@ -82,50 +91,77 @@ function ManagePublicKeySwitchDisplay() {
...
+                    Your public key is already configured.
+                </p>
+            )}
+            {showUpload && (
                 <ManagePublicKeyUpload
                     onPublicKeySubmit={handlePublicKeySubmit}
                     onFileChange={handleFileChange}
+                    onBack={() => setSender("")}
+                    onFetchPublicKey={setHasPublicKey}
+                    showBack={showBack}
```
stephenkao, line 153: Rather than having a `showBack` AND an `onBack`, could we just pass a function or null for `onBack`?  Then we could just check for that in the component:

```tsx
{ onBack && (
    <Button type="button" onClick={ onBack }>Back</Button>
) }
```

stephenkao:
As we're working on these multi-step flows, I think it'd benefit us to have better management across components -- especially for form state.  @jpandersen87 brought this up before in an eng sync, and I've used [React Hook Form](https://react-hook-form.com/) before, and it's pretty straightforward to use/hook up so I'd be in down to get a proof-of-concept up for this.  It heavily emphasizes the parent-as-a-manager/mediator approach we've been tending toward too.  (Not a bad thing, just something that we need to keep in mind as we build more complex flows like this.) What do y'all think?

HOWEVER, this feature doesn't have to wait for our decision on form state management.  Just wanted to get the discussion started earlier rather than later!

frontend-react/src/config/endpoints/settings.ts
penny-lischer: Agree, unfortunately the backend API is `organizationName`. @stephenkao should I update to `orgName` in backend to stay consistent throughout?

stephenkao: The segments here like `:organizationName` are arbitrary names so we could change this to `:orgName` (or anything that's a valid variable name) and just make sure we pass in an `orgName` in `segments` [here](https://github.com/CDCgov/prime-reportstream/pull/9007/files/07620cb92bd7e08fed5d6f67199eca3b45bbb160#diff-624d4d47c219d2c816d15fb78b90941179412aa9149d8a5a7ce637195a856c8cR21):

```ts
segments: {
    orgName: activeMembership?.parsedName!!,
}
```

frontend-react/src/components/ManagePublicKey/ManagePublicKeyChooseSender.tsx
stephenkao: I agree that we should be striving for dumber components, but in my mind it also depends on how we can balance the complexity of the parent manager component.  Dumber components would work for this use case, but it might get a little messy if we had a ten-step flow with requested resources in each step:

```
ManagePublicKey (parent manager): uses the following hooks:
    - useCollection1
    - useCollection2
    - useCollection3
    - useCollection4
    ...

ManagePublicKeyResource1: takes in collection1, calls callback with selected resource1
ManagePublicKeyResource2: takes in collection2, calls callback with selected resource2
ManagePublicKeyResource3: takes in collection3, calls callback with selected resource3
ManagePublicKeyResource4: takes in collection4, calls callback with selected resource4
...
```

I admit the above is very far-fetched, and we'd likely want to create a consolidate logic elsewhere -- maybe in a central hook or something like that -- but I can imagine that we'd benefit from the components being a little more self-sufficient in making the requests themselves.  (This is also a bit tied to our bigger form state discussion, though, since that'd dictate how we'd manage the flow state and how that'd be interwoven with requested resources.)

frontend-react/src/hooks/UseOrganizationPublicKeys.ts
```diff
@@ -0,0 +1,41 @@
...
+import { useSessionContext } from "../contexts/SessionContext";
+
+const { publicKeys } = servicesEndpoints;
+
+export type UseOrganizationPublicKeysResult = {
+    orgPublicKeys: RSApiKeysResponse;
+    isLoading: boolean;
+};
+
+export const useOrganizationPublicKeys = () => {
```
stephenkao, line 17: Can we make this the default export?

```tsx
export default function useOrganizationPublicKeys() {
    ...
}
```

frontend-react/src/hooks/UseOrganizationPublicKeys.ts
```diff
@@ -0,0 +1,41 @@
...
+    const { data, isLoading } = rsUseQuery(
+        [publicKeys.queryKey, activeMembership],
+        memoizedDataFetch,
+        {
+            enabled:
+                !!activeMembership?.parsedName && !!activeMembership.service,
+        }
+    );
+
+    return { orgPublicKeys: data, isLoading };
```
stephenkao, line 40: Our custom react-query hooks should just be a loose wrapper of configuration, and they should just pass back what `useQuery`/`useMutation` returns verbatim:

```ts
const query = rsUseQuery(...);

// or more simply
return rsUseQuery(...);
```

Destructuring beforehand limits the utility of the return value since we have to explicitly pass back `isLoading`, `isInitialLoading`, `status`, et cetera if we ever need that from the query/mutation.

frontend-react/src/hooks/UseCreateOrganizationPublicKey.ts
```diff
@@ -0,0 +1,53 @@
...
+    sender: string;
+}
+
+export type UseCreateOrganizationPublicKeyResult = UseMutationResult<
+    RSApiKeysResponse,
+    unknown,
+    OrganizationPublicKeyPostArgs
+>;
+
+export const useCreateOrganizationPublicKey =
```
stephenkao, line 22: Can this be the default?

```ts
export default function useCreateOrganizationPublicKey() {
    ...
}
```

frontend-react/src/hooks/UseCreateOrganizationPublicKey.ts
```diff
@@ -0,0 +1,53 @@
+import { useCallback } from "react";
```
stephenkao, line 1: I still think we should take a more intentional approach to our hooks/components rather than have a catch-all hooks/ or components/ directory, but this is fine for now.  In the future, I'd recommend we take the domain-driven approach as described [in the proposal](https://github.com/CDCgov/prime-reportstream/blob/master/frontend-react/docs/proposals/0002-domain-driven-directory-structure.md).

frontend-react/src/hooks/UseOrganizationPublicKeys.ts
penny-lischer: Is this the pattern we want to use in the hooks going forward?

frontend-react/src/hooks/UseCreateOrganizationPublicKey.ts
penny-lischer: Agree. I can create a /hooks/PublicKeys folder or  /hooks/Organizations/PublicKeys and move. Thoughts?

frontend-react/src/components/ManagePublicKey/ManagePublicKeyUpload.tsx
```diff
@@ -12,55 +12,78 @@ import {
...
     file: File | null;
 }
 
 export default function ManagePublicKeyUpload({
     onPublicKeySubmit,
     onFileChange,
+    onBack,
+    hasBack,
+    hasPublicKey,
     file,
```
jpandersen87, line 27: I feel like we can simplify this to a "publicKey" prop that can be `boolean | File`. if publicKey == true then show already configured, else show the form. Parent component would pass something like `publicKey={hasPublicKey ?? file}` in component initialization. What do you think @stephenkao ?

frontend-react/src/components/ManagePublicKey/ManagePublicKey.test.tsx
```diff
@@ -1,136 +1,280 @@
-import { screen, fireEvent } from "@testing-library/react";
+/* eslint-disable testing-library/no-unnecessary-act */
```
stephenkao, line 1: Let's only disable when it's necessary rather than doing it at the top of the file:

```ts
/* eslint-disable testing-library/no-unnecessary-act */
act(() => { ... });
/* eslint-enable testing-library/no-unnecessary-act */
```

It's a little obnoxious to do this, but it's better for visibility into workarounds rather than disabling the rule wholesale.

operations/docker-compose.yml
```diff
@@ -7,8 +7,8 @@ services:
     build:
       context: .
       args:
-        AZURE_CLI_VERSION: 2.27.2
```
stephenkao, line 10: Can you unstage this from the commit?

prime-router/docs/schema_documentation/direct-anavasidx-covid-19.md
```diff
@@ -1583,7 +1583,7 @@ Custom field. Note, value matched LIVD column "F", "Test Performed LOINC Code"
 
 **Format**: use value found in the Code column
 
-**Default Value**: T
+**Default Value**: P
```
stephenkao, line 1586: Revert?

prime-router/metadata/schemas/direct/anavasidx-covid-19.schema
```diff
@@ -11,4 +11,36 @@ elements:
     default: AnavasiDx
 
   - name: processing_mode_code
-    default: T
```
stephenkao, line 14: Revert?

frontend-react/src/styles/_components.scss
```diff
@@ -96,4 +96,10 @@ hr {
     .usa-icon {
         font-size: 20px;
     }
+}
+.manage-public-key-configured {
+    .usa-icon {
```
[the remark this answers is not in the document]
stephenkao: I absolutely agree, and I've previously suggested using the [`size`](https://github.com/trussworks/react-uswds/blob/main/src/components/Icon/Icon.tsx#L7) prop for iconography dimensions.  If we're intent on deviating from USWDS defaults, then we need to align with design on making them _universal_ deviations.

frontend-react/src/hooks/network/Organizations/PublicKeys/UseCreateOrganizationPublicKey.ts
```diff
@@ -0,0 +1,52 @@
...
+export default function useCreateOrganizationPublicKey(): UseCreateOrganizationPublicKeyResult {
+    const { activeMembership } = useSessionContext();
+    const parsedName = activeMembership?.parsedName;
+
+    const { authorizedFetch, rsUseMutation } = useAuthorizedMutationFetch<
+        RSApiKeysResponse,
+        unknown,
+        OrganizationPublicKeyPostArgs
+    >();
+    const mutationFn = useCallback(
```
stephenkao, line 31: I really don't think we need to be using `useCallback` and `useMemo` so frequently; if the references for `authorizedFetch` or `parsedName` change, we should trigger a re-render anyway, no?

Not a huge deal, but I just don't think we need to be so focused on memoization when re-renders are expected behaviors anyhow.

frontend-react/src/components/ManagePublicKey/ManagePublicKey.tsx
```diff
@@ -82,50 +99,85 @@ function ManagePublicKeySwitchDisplay() {
...
+                            If you need more information on generating your
+                            public key, reference page 7 in the{" "}
+                            <USLink href="/resources/programmers-guide">
+                                API Programmer’s Guide.
+                            </USLink>
+                        </span>
+                    </SiteAlert>
+                </>
+            )}
             {sender.length === 0 && (
```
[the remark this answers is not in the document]
penny-lischer: @jpandersen87 if I change from `sender.length === 0` to `sender` the logic changes since sender is an empty string. I can change to from `sender.length === 0` to `!sender`

frontend-react/src/components/ManagePublicKey/ManagePublicKeyUpload.tsx
stephenkao: I agree, and I feel like something similar could be done with `onBack`/`hasBack` -- we could make it just not pass `onBack` if the component shouldn't show a back button 🤷

frontend-react/src/hooks/network/Organizations/PublicKeys/UseCreateOrganizationPublicKey.ts
jpandersen87: I'm likely guilty of this as well 😆. It's true that the re-creation of simple functions is low performance lift.

frontend-react/src/components/ManagePublicKey/ManagePublicKey.tsx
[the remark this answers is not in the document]
jpandersen87: doh i'm dumb, that should've been `!sender && (`. sorry

frontend-react/src/components/ManagePublicKey/ManagePublicKeyUpload.tsx
jpandersen87: Oh yeah, good catch!

stephenkao:
@penny-lischer I think there's an issue with the spinner show logic:

stephenkao (approved):
A few bugs (one we chatted about already and one I left a comment about) and some unrelated changes that leaked into this (schema + Terraform changes) that should be addressed before merging, but this looks good to me overall!  Thank you!

Just a note for the future, I'd recommend breaking similar amounts of work down into smaller pull requests, especially since we have the advantage of using Storybook.  Each "step" of the public key process could be made into a dumb component (like Joseph's recommendation), and they'd just boil down into simple input-output.

jpandersen87 (approved):
Since my comments were minor nitpicks I will give overall approval to help move things along 🙂

frontend-react/src/styles/_components.scss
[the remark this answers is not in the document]
penny-lischer: Agree

frontend-react/src/hooks/network/Organizations/PublicKeys/UseCreateOrganizationPublicKey.ts
penny-lischer: @jpandersen87 any thoughts on Stephens comment? This was following the new pattern from https://github.com/CDCgov/prime-reportstream/pull/8690

jpandersen87: Completely OK to drop the callback. I need to do the same.

PR #9033, opened 2023-04-07
Experience/8939/remove-react-dev-utils-dependency

### LINKED ISSUE #8939: Drop usage of react-dev-utils dependency
While we are now free from direct dependency with react-scripts (Create-React-App), we still have remaining ties with sub-dependencies of the CRA project. One of those is react-dev-utils. This package has outdated dependencies that require resolution overrides to prevent errors with our building libraries. We need to either integrate this library directly into the project, if needed, or drop usage of what it provides so we can remove it as a dependency.

jpandersen87:
Closes #8939 

This PR makes numerous changes to dev dependencies and local dev related files to progress moving away from depending on CRA packages directly. CRA package files for eslint, babel, and react-dev-utils have been copied directly to the project and their dependencies moved to our package.json. This will help reduce errors with transient package version splits as we continue to evaluate and simplify/remove the now-local code. Updates were also made to the storybook configuration to enable v7 specific features to help prepare for future migration (ex: babel configuration loading is now completely dependant on the project's babelrc file instead of using a storybook default config). The storybook npm script command was updated to run the scss compiler in parallel so that scss changes can be recompiled automatically.


frontend-react/.storybook/main.js
```diff
@@ -1,3 +1,6 @@
+"use strict;";
```
stephenkao, line 1: Extra semicolon here :)

stephenkao:
@jpandersen87 This looks good to me overall, but I have two points:

- Would you mind marking (either through GitHub comments or inline comments) areas that are custom code as opposed to what's been extracted out of the remains of CRA?  I know the Storybook and Babel updates are custom to us, but was there anything else?
- Would it be useful to add a quick README in the react-dev-utils directory explaining why we need these pieces + TODOs for future cleanup (if any)?

penny-lischer (approved):
LGTM

jpandersen87:
> @jpandersen87 This looks good to me overall, but I have two points:
> 
> * Would you mind marking (either through GitHub comments or inline comments) areas that are custom code as opposed to what's been extracted out of the remains of CRA?  I know the Storybook and Babel updates are custom to us, but was there anything else?
> * Would it be useful to add a quick README in the react-dev-utils directory explaining why we need these pieces + TODOs for future cleanup (if any)?

Everything in the react-dev-utils should be mostly as-is aside from formatting and import references. There were files (ex: chalk.js) that were just direct import/export of the chalk library that were removed and affected import references changed to point directly to the library. The CRA eslint code was simple enough to embed all directly into the .eslintrc.js file. Babel config files were copied as-is and references updated. Pieces of webpack config were simplified (detection of typescript, detection of tailwindcss, etc.) by removing the conditionals and applying the needed configurations. We're not using tailwindcss so those conditional pieces of configuration were removed entirely and the tailwindcss dependency dropped.

stephenkao (approved):
Looks good to me!

PR #9043, opened 2023-04-07
Universal Pipeline filter logging improvements

### LINKED ISSUE #8915: bug: Default filters not recorded in submission history filteredReportRows/filteredReportItems
## Describe the bug
When a quality filter is not specified by a receiver, a default is used. However, when the default quality filter does filter out reports, the logged message reflecting that is not helpful; for example:

```
For flexion.simulated-lab, filter [][] filtered out item 969bcbb3-cd34-49be-ac4f-e1b8479b8219"
```

This makes it hard to know which filter led to the item being filtered out and where that filter came from. This issue should allow that logging message to reveal more useful information in these cases.

## Impact on ReportStream
- When testing the sending of reports to receivers for ETOR, since lab orders don't pass the default quality filter, it is difficult to see why reports don't get through if a quality filter on the receiver is not set.

## Steps to reproduce
1. Send a report that is missing some of the content required by the default quality filter to a receiver that does not have a quality filter specified.
2. Query the submission history and look in the results for filteredReportRows or filteredReportItems and find the logged message.

## Expected behavior
Filtered results print useful information about what filter caused the item to be filtered out.

jimmyfagan:
This PR makes adjustments to the way that filter results are logged in the Universal Pipeline. There were previously several cases where the logged filter message would not have sufficient information to identify why a result was filtered out, and those should be a bit more clear with these changes. For details on the updated behavior, see the Markdown file in the changeset.

Test Steps:
There are several different cases that can/should be tested here, I'll list a few scenarios.

Default filter logging:
1. Delete a receiver's quality filter and reload the credentials.
2. Send a report that matches the receiver's jurisdictional filter, but which fails the default quality filter. (Removing the new lines from [this file](https://github.com/CDCgov/trusted-intermediary/blob/main/examples/fhir/lab_order.json) is an example of a report which does not pass the quality filter)
3. Query the submission history for that report id.
4. Verify that the filter logs in the response include an indication that the default filter was used, and that only the predicate(s) from the default quality filter that evaluate to false are listed.

Reversed filter logging

1. Delete a receiver's quality filter, set `reverseTheQualityFilter` to true, and reload the credentials.
2. Send a report that matches the receiver's jurisdictional filter, and which passes the default quality filter (meaning it will fail a reversed default quality filter).
3. Query the submission history for that report id.
4. Verify that the filter logs in the response include an indication that the default filter was used and the filter was reversed, and all predicates from the default quality filter are listed.

Exception filter logging

1. Set a receiver's quality filter to include some predicate(s) which will cause a schema exception.
2. Send a report that matches the receiver's jurisdictional filter, and which will exercise the the exception-causing filter.
3. Query the submission history for that report id.
4. Verify that the filter logs in the response include an indication that the filter led to an exception, and all predicates which lead to exceptions are listed.

## Changes
- Edit `evaluateFilterCondition` to return a Boolean and a String so that calling functions have access to a String description of the failing filters (which allows for the printed list of predicates to be different than the entire list for the filter, also allows for other indicators for reversed filters or filters causing exceptions)
- Update tests to reflect `evaluateFilterCondition` return type changes, and to remove kotlin warnings with minor changes
- Remove the application of default filters "by lazy" in a simple Boolean form, and instead allow returning of default filters in `getQualityFilters` and `getProcessingModeFilter` (simplifies `evaluateFilterCondition` having access to predicates within the default filter)
- Adjust logging in `evaluateFilterAndLogResult` to identify whether or not the filter applies was a default filter and add to the description of the failing filters to indicate as such

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- Fixes #8915


prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
```diff
@@ -437,60 +418,88 @@ class FHIRRouter(
...
         )
         if (!passes) {
-            logFilterResults(filters, bundle, report, receiver, filterType)
+            val filterToLog = "${
+            if ((filterType == ReportStreamFilterType.QUALITY_FILTER && filters === qualityFilterDefault) ||
+                (
+                    filterType == ReportStreamFilterType.PROCESSING_MODE_FILTER &&
+                        filters === processingModeFilterDefault
+                    )
+            )
```
jimmyfagan, line 435: The usage of `===` (referential equality operator) is intentional and necessary here because we only want to tag a filter as a default filter if it references the default filter, and not in cases where a filter is equivalent to the default filter but references a filter that was derived from a receiver's settings.

JessicaWNava (approved):
LGTM!

prime-router/docs/proposals/0019-universal-pipeline-filter-reporting.md
```diff
@@ -0,0 +1,167 @@
...
+Currently only jurisdictional filters have default results that would filter out reports, and jurisdictional filter
+results are not logged. If that were to change, we would have to log that we filtered out an item, but without any
+filter to reference, so this case might have an entirely different format, or the message might look like:
+"For someOrg.someReceiver, filter default response[] filtered out item someItemId"
+
+## Future Considerations
+
+Ticket [#8837](https://github.com/CDCgov/prime-reportstream/issues/8837)
+involves changing the way that condition filters are evaluated so that instead of evaluating condition
+filters with the usual , we would instead use . We’ll want to make sure during that process that the logging behavior
```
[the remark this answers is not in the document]
jimmyfagan: Good catch! In the Google Doc, it was an equation, so it didn't copy over cleanly, checking in a commit to fix this now.

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
TurkeyFried: This may be worth adding as a comment in the code itself

prime-router/docs/proposals/0019-universal-pipeline-filter-reporting.md
```diff
@@ -0,0 +1,168 @@
...
+If a report is filtered out due to a quality filter along with a setting of `reverseTheQualityFilter: true`, the logged
+message will include the text “(reversed)”. Imagine a reversed filter of [A, B, C, D, E], which can be thought of as or
+equivalently . The only way for this to yield a negative result is if each and every predicate A, B, C, D, and E
+evaluate to true; therefore each of those predicates is relevant in logging why the filter yielded a negative result. So
+while in non-reversed cases, we only include the individual predicates that evaluated to false, in reversed cases, we
+include all predicates, and the resulting message might look like this:
+"For someOrg.someReceiver, filter (reversed) [A, B, C, D, E][] filtered out item someItemId"
+As it is today, the only filters that can be reversed are quality filters. If the default quality filter is reversed,
+the resulting message might look like this:
+"For someOrg.someReceiver, filter (default filter) (reversed) [A, B, C, D, E][] filtered out item someItemId"
```
TurkeyFried, line 149: The responses in this part of the code are missing the nice little code block formatting that they have earlier in the file

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
```diff
@@ -437,60 +418,88 @@ class FHIRRouter(
...
-        } catch (e: SchemaException) {
-            actionLogger?.warn(
-                EvaluateFilterConditionErrorMessage(e.message)
-            )
-            return false
         }
-        return if (reverseFilter) !result else result
+
+        return if (exceptionFilters.isNotEmpty()) {
+            Pair(false, "(exception found) $exceptionFilters")
```
arnejduranovic, line 487: You're not actually testing these paths anywhere, are you? Should test that the returned pair is exactly equal to the expected pair for each one of these if/else if branches?

prime-router/src/test/kotlin/fhirengine/engine/fhirRouterTests/DefaultFilterTests.kt
```diff
@@ -320,7 +320,7 @@ class DefaultFilterTests {
             engine.qualityFilterDefault,
             bundle,
             false
-        )
+        ).first
```
arnejduranovic, line 323: See my other comment, but it seems like instead of just testing that the first part of the pair is what we expect, we need to update all these tests to test the whole pair?

arnejduranovic (changes requested):

prime-router/docs/proposals/0019-universal-pipeline-filter-reporting.md
```diff
@@ -0,0 +1,167 @@
+# Summary
```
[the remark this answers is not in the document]
jimmyfagan: I like that idea, on it!

prime-router/docs/proposals/0019-universal-pipeline-filter-reporting.md
```diff
@@ -0,0 +1,168 @@
...
+
+When a report satisfies a receiver’s jurisdictional filter requirements, the report still need to satisfy other filters
+before it actually gets sent to that receiver. If any of those filters return negative results, logs summarizing the
+filtered report are saved to the submission history with a description of which filters resulted in the filtered items.
+Because there are a few different situations in which a report would be filtered out at this step, this document intends
+to summarize what information is included in the filtered report logs, and why that information is included.
+
+# Details
+
+## Filter Types
```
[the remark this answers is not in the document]
jimmyfagan: I was thinking that might eventually be better suited for a document focused on the filters rather than on the logging, but I suppose it doesn't hurt to include that here too.

TurkeyFried:
Cannot officially Approve just yet (permission issue) but wanted to confirm that the 3 use cases pass manual testing.

prime-router/docs/proposals/0019-universal-pipeline-filter-reporting.md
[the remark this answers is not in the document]
arnejduranovic: My point is, I don't think we need a "logging" document. I think much more valuable is to have a "UP Filters" document, with a subsection on logging as it relates to the filters. That way everything related to filters is in one place. Thoughts?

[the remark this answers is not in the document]
jimmyfagan: That may be true, but feels like a reach for this ticket, so I think for now, I'll just add some filter details to this page, and if this ends up being becoming the general filter documentation down the line, then that works fine.

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
jimmyfagan: Beefed up the tests in an incoming commit.

prime-router/docs/features/0001-universal-pipeline-filter-reporting.md
```diff
@@ -52,19 +54,22 @@ val qualityFilterDefault: ReportStreamFilter = listOf(
 
 Additionally, quality filters can be reversed using the setting `reverseTheQualityFilter` on a receiver. When that
 property is set to `true`, anything that would normally pass the quality filter will be filtered out, and anything that
-would normally be filtered by the quality filter will be allowed through.
+would normally be filtered by the quality filter will be allowed through. In practice, this is usually used when an
```
arnejduranovic, line 57: Great info!

PR #9085, opened 2023-04-11
Experience/8280/one table to rule

etanb:
⚠️ PR VERY IN PROGRESS ⚠️

### DIFF
--- a/frontend-react/package.json
+++ b/frontend-react/package.json
@@ -41,6 +41,7 @@
     "remark-gfm": "^3.0.1",
     "remark-toc": "^8.0.1",
     "rest-hooks": "^6.1.7",
+    "sanitize-html": "^2.10.0",
     "use-deep-compare-effect": "^1.8.1",
     "web-vitals": "^3.1.1"
   },
--- a/frontend-react/src/components/shared/Table/Table.stories.tsx
+++ b/frontend-react/src/components/shared/Table/Table.stories.tsx
@@ -8,114 +8,70 @@ export default {
     component: Table,
 } as ComponentMeta<typeof Table>;
 
-function DefaultTableContent() {
-    return (
-        <>
-            <thead>
-                <tr>
-                    <th>Shape</th>
-                    <th>Sides</th>
-                    <th>Note</th>
-                </tr>
-            </thead>
-            <tbody>
-                <tr>
-                    <td>Triangle</td>
-                    <td>3</td>
-                    <td>
-                        Lorem ipsum dolor sit amet, consectetur adipiscing elit,
-                        sed do eiusmod tempor incididunt ut labore et dolore
-                        magna aliqua.
-                    </td>
-                </tr>
-                <tr>
-                    <td>Square</td>
-                    <td>4</td>
-                    <td>
-                        Sed ut perspiciatis unde omnis iste natus error sit
-                        voluptatem accusantium doloremque laudantium, totam rem
-                        aperiam, eaque ipsa quae ab illo inventore veritatis et
-                        quasi architecto beatae vitae dicta sunt explicabo.
-                    </td>
-                </tr>
-                <tr>
-                    <td>Pentagon</td>
-                    <td>5</td>
-                    <td>
-                        Ut enim ad minima veniam, quis nostrum exercitationem
-                        ullam corporis suscipit laboriosam, nisi ut aliquid ex
-                        ea commodi consequatur?
-                    </td>
-                </tr>
-            </tbody>
-        </>
-    );
-}
-
-const defaultHeaderContent = ["Shape", "Sides", "Notes"];
-const defaultRowContent = [
+const defaultData = [
     [
-        "Triangle",
-        "3",
-        "Lorem ipsum dolor sit amet, consectetur adipiscing elit sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.",
+        {
+            columnKey: "Shape",
+            content: "Triangle",
+        },
+        {
+            columnKey: "Sides",
+            content: "3",
+        },
+        {
+            columnKey: "Note",
+            content:
+                "Lorem ipsum dolor sit amet, consectetur adipiscing elit sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.",
+        },
     ],
     [
-        "Square",
-        "4",
-        "Sed ut perspiciatis unde omnis iste natus error sit voluptatem accusantium doloremque laudantium, totam rem aperiam, eaque ipsa quae ab illo inventore veritatis et quasi architecto beatae vitae dicta sunt explicabo.",
+        {
+            columnKey: "Shape",
+            content: "Square",
+        },
+        {
+            columnKey: "Sides",
+            content: "4",
+        },
+        {
+            columnKey: "Note",
+            content:
+                "Sed ut perspiciatis unde omnis iste natus error sit voluptatem accusantium doloremque laudantium, totam rem aperiam, eaque ipsa quae ab illo inventore veritatis et quasi architecto beatae vitae dicta sunt explicabo.",
+        },
     ],
     [
-        "Pentagon",
-        "5",
-        "Ut enim ad minima veniam, quis nostrum exercitationem ullam corporis suscipit laboriosam, nisi ut aliquid ex ea commodi consequatur?",
+        {
+            columnKey: "Shape",
+            content: "Pentagon",
+        },
+        {
+            columnKey: "Sides",
+            content: "5",
+        },
+        {
+            columnKey: "Note",
+            content:
+                "Ut enim ad minima veniam, quis nostrum exercitationem ullam corporis suscipit laboriosam, nisi ut aliquid ex ea commodi consequatur?",
+        },
     ],
 ];
 
 export const Borderless: ComponentStory<typeof Table> = () => {
-    return (
-        <Table
-            borderless
-            columnHeaders={defaultHeaderContent}
-            rowData={defaultRowContent}
-        />
-    );
+    return <Table borderless rowData={defaultData} />;
 };
 
 export const Bordered: ComponentStory<typeof Table> = () => {
-    return (
-        <Table
-            columnHeaders={defaultHeaderContent}
-            rowData={defaultRowContent}
-        />
-    );
+    return <Table rowData={defaultData} />;
 };
 
 export const Striped: ComponentStory<typeof Table> = () => {
-    return (
-        <Table
-            striped
-            columnHeaders={defaultHeaderContent}
-            rowData={defaultRowContent}
-        />
-    );
+    return <Table striped rowData={defaultData} />;
 };
 
 export const Scrollable: ComponentStory<typeof Table> = () => {
-    return (
-        <Table
-            scrollable
-            columnHeaders={defaultHeaderContent}
-            rowData={defaultRowContent}
-        />
-    );
+    return <Table scrollable rowData={defaultData} />;
 };
 
 export const Sortable: ComponentStory<typeof Table> = () => {
-    return (
-        <Table
-            sortable
-            columnHeaders={defaultHeaderContent}
-            rowData={defaultRowContent}
-        />
-    );
+    return <Table sortable rowData={defaultData} />;
 };
--- a/frontend-react/src/components/shared/Table/Table.stories.tsx
+++ b/frontend-react/src/components/shared/Table/Table.stories.tsx
@@ -1,0 +1,16 @@
+import React from "react";
+import { ComponentMeta, ComponentStory } from "@storybook/react";
+
+import { Table } from "./Table";
+
+export default {
+    title: "components/Table",
+    component: Table,
+} as ComponentMeta<typeof Table>;
+
+export const Default: ComponentStory<typeof Table> = (args) => (
+    <Table {...args} />
+);
+Default.args = {
+    children: "Children are optional! (Try clearing me out)",
+};
--- a/frontend-react/src/components/shared/Table/Table.stories.tsx
+++ b/frontend-react/src/components/shared/Table/Table.stories.tsx
@@ -32,18 +32,19 @@ function DefaultTableContent() {
                     <td>Square</td>
                     <td>4</td>
                     <td>
-                        Lorem ipsum dolor sit amet, consectetur adipiscing elit,
-                        sed do eiusmod tempor incididunt ut labore et dolore
-                        magna aliqua.
+                        Sed ut perspiciatis unde omnis iste natus error sit
+                        voluptatem accusantium doloremque laudantium, totam rem
+                        aperiam, eaque ipsa quae ab illo inventore veritatis et
+                        quasi architecto beatae vitae dicta sunt explicabo.
                     </td>
                 </tr>
                 <tr>
                     <td>Pentagon</td>
                     <td>5</td>
                     <td>
-                        Lorem ipsum dolor sit amet, consectetur adipiscing elit,
-                        sed do eiusmod tempor incididunt ut labore et dolore
-                        magna aliqua.
+                        Ut enim ad minima veniam, quis nostrum exercitationem
+                        ullam corporis suscipit laboriosam, nisi ut aliquid ex
+                        ea commodi consequatur?
                     </td>
                 </tr>
             </tbody>
@@ -51,34 +52,70 @@ function DefaultTableContent() {
     );
 }
 
-export const Borderless: ComponentStory<typeof Table> = (args) => {
+const defaultHeaderContent = ["Shape", "Sides", "Notes"];
+const defaultRowContent = [
+    [
+        "Triangle",
+        "3",
+        "Lorem ipsum dolor sit amet, consectetur adipiscing elit sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.",
+    ],
+    [
+        "Square",
+        "4",
+        "Sed ut perspiciatis unde omnis iste natus error sit voluptatem accusantium doloremque laudantium, totam rem aperiam, eaque ipsa quae ab illo inventore veritatis et quasi architecto beatae vitae dicta sunt explicabo.",
+    ],
+    [
+        "Pentagon",
+        "5",
+        "Ut enim ad minima veniam, quis nostrum exercitationem ullam corporis suscipit laboriosam, nisi ut aliquid ex ea commodi consequatur?",
+    ],
+];
+
+export const Borderless: ComponentStory<typeof Table> = () => {
     return (
-        <Table borderless>
-            <DefaultTableContent />
-        </Table>
+        <Table
+            borderless
+            columnHeaders={defaultHeaderContent}
+            rowData={defaultRowContent}
+        />
     );
 };
 
-export const Bordered: ComponentStory<typeof Table> = (args) => {
+export const Bordered: ComponentStory<typeof Table> = () => {
     return (
-        <Table>
-            <DefaultTableContent />
-        </Table>
+        <Table
+            columnHeaders={defaultHeaderContent}
+            rowData={defaultRowContent}
+        />
     );
 };
 
-export const Striped: ComponentStory<typeof Table> = (args) => {
+export const Striped: ComponentStory<typeof Table> = () => {
     return (
-        <Table striped>
-            <DefaultTableContent />
-        </Table>
+        <Table
+            striped
+            columnHeaders={defaultHeaderContent}
+            rowData={defaultRowContent}
+        />
     );
 };
 
-export const Scrollable: ComponentStory<typeof Table> = (args) => {
+export const Scrollable: ComponentStory<typeof Table> = () => {
     return (
-        <Table scrollable>
-            <DefaultTableContent />
-        </Table>
+        <Table
+            scrollable
+            columnHeaders={defaultHeaderContent}
+            rowData={defaultRowContent}
+        />
+    );
+};
+
+export const Sortable: ComponentStory<typeof Table> = () => {
+    return (
+        <Table
+            sortable
+            columnHeaders={defaultHeaderContent}
+            rowData={defaultRowContent}
+        />
     );
 };
--- a/frontend-react/src/components/shared/Table/Table.stories.tsx
+++ b/frontend-react/src/components/shared/Table/Table.stories.tsx
@@ -9,8 +9,8 @@ export default {
 } as ComponentMeta<typeof Table>;
 
 export const Default: ComponentStory<typeof Table> = (args) => (
-    <Table {...args} />
+    <Table>
+        <thead></thead>
+        <tbody></tbody>
+    </Table>
 );
-Default.args = {
-    children: "Children are optional! (Try clearing me out)",
-};
--- a/frontend-react/src/components/shared/Table/Table.stories.tsx
+++ b/frontend-react/src/components/shared/Table/Table.stories.tsx
@@ -8,9 +8,77 @@ export default {
     component: Table,
 } as ComponentMeta<typeof Table>;
 
-export const Default: ComponentStory<typeof Table> = (args) => (
-    <Table>
-        <thead></thead>
-        <tbody></tbody>
-    </Table>
-);
+function DefaultTableContent() {
+    return (
+        <>
+            <thead>
+                <tr>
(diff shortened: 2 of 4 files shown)


frontend-react/src/components/shared/Table/Table.module.scss
```diff
@@ -0,0 +1,53 @@
+@use 'uswds-core' as *;
```
stephenkao, line 1: I think the directory structure should be reversed here: `frontend-react/src/shared/components`.  This'll allow for other `shared` directories like `frontend-react/src/shared/utils`, `frontend-react/src/shared/hooks`, et cetera.

frontend-react/package.json
```diff
@@ -150,7 +152,7 @@
     "camelcase": "^6.2.1",
     "case-sensitive-paths-webpack-plugin": "^2.4.0",
     "cross-env": "^7.0.3",
-    "css-loader": "^6.7.3",
+    "css-loader": "5.2.6",
```
stephenkao, line 155: Hmmmm I don't think we should downgrade major versions specifically for @storybook/preset-scss -- these are loaders that we're using to build production code.

frontend-react/src/components/shared/Table/Table.module.scss
```diff
@@ -0,0 +1,53 @@
...
+
+.reportstreamTable {
+  :global {
+    // Since these are USDS overrides, we use the :global property so we can
```
stephenkao, line 5: Why wouldn't being nested within `.reportstreamTable` provide enough specificity to override the USWDS rules?

frontend-react/src/components/shared/Table/Table.module.scss
```diff
@@ -0,0 +1,53 @@
...
+
+    .usa-table thead .column-header-title-container p {
+        font-size: 14px;
+        margin: 0;
+    }
+
+    .usa-table thead .column-header-title-container svg {
+        flex: none;
+        margin-left: 2px;
+        height: 28px;
```
stephenkao, line 35: Can we not use `size` for the Icon component?  I think we should align with the given sizes we have rather than introducing custom ones all the time.

frontend-react/src/components/shared/Table/Table.module.scss
```diff
@@ -0,0 +1,53 @@
...
+        width: 28px;
+    }
+
+    .usa-table td.message-receiver-break-word {
```
stephenkao, line 39: This shouldn't be here, right?  It's a specific rule for the Message Tracker feature that should exist separately, not part of the generic table styles.

frontend-react/src/components/shared/Table/Table.module.scss
stephenkao: Additionally, can we just apply modifier classes instead of custom CSS rules below?  I think it'd be better if we reuse modifier classes whenever possible rather than introduce new CSS.

frontend-react/src/components/shared/Table/Table.tsx
```diff
@@ -0,0 +1,225 @@
...
+    stackedStyle?: "default" | "headers";
+    striped?: boolean;
+    rowData: RowData[][];
+}
+
+const SortableTableHeader = ({
+    columnTitle,
+    activeColumn,
+    sortOrder,
+    setSortOrder,
```
stephenkao, line 46: Let's use the naming pattern of `on<entity><action>` for callback props -- something like `onSortOrderChange`, `onActiveColumnChange`?

frontend-react/src/components/shared/Table/Table.tsx
```diff
@@ -0,0 +1,225 @@
...
+    const isActive = columnTitle === activeColumn;
+    if (isActive && sortOrder === FilterOptionsEnum.ASC) {
+        SortIcon = Icon.ArrowUpward;
+    } else if (isActive && sortOrder === FilterOptionsEnum.DESC) {
+        SortIcon = Icon.ArrowDownward;
+    }
+    const isCurrentlyActiveColumn = columnTitle === activeColumn;
+
+    const handleHeaderClick = () => {
+        if (!isCurrentlyActiveColumn) {
```
stephenkao, line 59: I wonder if we can just maintain active column and sort order in a single state value:

```
{ <active_column_key>: <sort_order> }
```

It'll only be populated if there's an active column and current sort order; otherwise, `null` means that there's no sort order?

frontend-react/src/components/shared/Table/Table.tsx
```diff
@@ -0,0 +1,225 @@
...
+        <>
+            <thead>
+                <tr>
+                    {columnHeaders.map((header) => {
+                        return (
+                            <SortableTableHeader
+                                columnTitle={header}
+                                activeColumn={activeColumn}
+                                sortOrder={sortOrder}
+                                setActiveColumn={setActiveColumn}
```
stephenkao, line 126: These setters also need to account for custom callbacks in case of API sorting.

frontend-react/src/components/shared/Table/Table.tsx
```diff
@@ -0,0 +1,225 @@
...
+}: {
+    rowData: RowData[][];
+    columnHeaders: string[];
+}) => {
+    const [activeColumn, setActiveColumn] = useState("");
+    const [sortOrder, setSortOrder] = useState(FilterOptionsEnum.NONE);
+    const sortedData = useMemo(() => {
+        return sortOrder !== FilterOptionsEnum.NONE && activeColumn
+            ? rowData.sort((a, b): number => {
+                  let contentColA = convertStringToSortableRepresentation(
```
stephenkao, line 101: These can be `const`s.

frontend-react/src/components/shared/Table/Table.tsx
```diff
@@ -0,0 +1,225 @@
...
+const SortableTable = ({
+    rowData,
+    columnHeaders,
+}: {
+    rowData: RowData[][];
+    columnHeaders: string[];
+}) => {
+    const [activeColumn, setActiveColumn] = useState("");
+    const [sortOrder, setSortOrder] = useState(FilterOptionsEnum.NONE);
+    const sortedData = useMemo(() => {
```
stephenkao, line 98: Minor point, but I don't think memoization would really help out too much here; if anything related to the table content or sorting changes, a full re-render would be expected anyway.

frontend-react/src/components/shared/Table/Table.tsx
```diff
@@ -0,0 +1,225 @@
...
+    activeColumn: string;
+    sortOrder: string;
+    setSortOrder: (order: string) => void;
+    setActiveColumn: (column: string) => void;
+}
+
+interface RowData {
+    columnKey: string;
+    content: string;
+    modalContent?: string;
```
stephenkao, line 27: We chatted about this before, but I don't think `modalContent` is necessarily something we want to build into the table.  I think it might be better to break this down into different pieces:
- `displayValue`: the value that's displayed on the cell prior to any interaction
- `rawValue`: the value we get back verbatim (could be equivalent to `displayValue` in many cases)
- `CellComponent`: custom component that would render the values
    - By default, we can make a component that renders `rawValue` in a modal

The idea here is to ensure that we're as flexible as we need to be with the different values so the table can easily scale to different display capabilities.

frontend-react/src/components/shared/Table/Table.tsx
```diff
@@ -0,0 +1,225 @@
+import classnames from "classnames";
+import sanitizeHtml from "sanitize-html";
+import { useMemo, useState } from "react";
+import { Icon } from "@trussworks/react-uswds";
+
+import { convertStringToSortableRepresentation } from "../../../utils/misc";
+
+import styles from "./Table.module.scss";
+
+const FilterOptionsEnum = {
```
stephenkao, line 10: Can we not use an actual `enum` here?  I think the `Enum` suffix is redundant in any case.

frontend-react/src/components/shared/Table/Table.tsx
stephenkao, line 98: Can the sorting function here be pull out into its own top-level function?

```ts
function sortTableData(rowData: RowData[][], sortOrder?: FilterOptionsEnum) {
    if (sortOrder === FilterOptionsEnum.None) {
        return rowData;
    }

   return rowData.sort(...);
}
```

frontend-react/src/components/shared/Table/Table.tsx
```diff
@@ -0,0 +1,225 @@
...
+        "usa-table--striped": striped,
+    });
+
+    const columnHeaders = [
+        ...new Set(rowData.flat().map((data: RowData) => data.columnKey)),
+    ];
+
+    return (
+        <div
+            className={styles.reportstreamTable}
```
stephenkao, line 180: Let's use `classnames` for variable class names.

frontend-react/src/components/shared/Table/Table.tsx
```diff
@@ -0,0 +1,225 @@
...
+            {...(scrollable && {
+                className: `usa-table-container--scrollable ${styles.reportstreamTable}`,
+                tabIndex: 0,
```
stephenkao, line 183: Why does this need to be tabbable if it's scrollable?  In any case, it's not great practice to add a tab index to non-interactive elements like `<div>`s.

frontend-react/src/components/shared/Table/Table.tsx
```diff
@@ -0,0 +1,225 @@
...
+                ) : (
+                    <>
+                        <thead>
+                            <tr>
+                                {columnHeaders.map((header) => {
+                                    return <th>{header}</th>;
+                                })}
+                            </tr>
+                        </thead>
+                        <tbody>
```
stephenkao, line 201: There's duplicate code between the SortableTable and non-SortableTable with the same markup.  Can we make a TableBody component that share that functionality?

frontend-react/src/utils/misc.ts
```diff
@@ -175,3 +175,27 @@ export const parseFileLocation = (
...
+export const convertStringToSortableRepresentation = (
+    input: string | undefined
+): string | number => {
+    let convertedContent = "";
+    if (!input) return convertedContent;
+
+    // Check if input is a date, then we could leave it as is,
+    // Or convert to 100% safe JS Date String
+    if (!!Date.parse(input)) {
+        convertedContent = new Date(input).toString();
```
stephenkao, line 192: If it's a date, why couldn't we just compare the number of milliseconds via `getTime()`?

frontend-react/src/utils/misc.ts
```diff
@@ -175,3 +175,27 @@ export const parseFileLocation = (
         fileName,
     };
 };
+
+export const removeHTMLFromString = (input: string) => {
+    return input.replace(/<(.|\n)*?>/g, "");
+};
+
+export const convertStringToSortableRepresentation = (
```
stephenkao, line 183: Hmmm I'm not sure if this function is really useful on its own or if it should be built into the sorting function with type checks along the way.  One flag is that the content may not always be a string (as it's assumed `content` would be [here](https://github.com/CDCgov/prime-reportstream/pull/9085/files#diff-4a8691b6c15e1d775d398ac71c10833729efd0af84f55106b12ff9677a836f60R102)); it could be any type of primitive values -- most commonly `null`, a number, or a boolean --  all of which I think should be acceptable values in sorting, no?

frontend-react/src/components/shared/Table/Table.module.scss
etanb: Hmmm I feel like that's a little redundant (having two have two `/components` folders) and a little confusing.  I could see shared utility functions, hooks etc being in a root folder `/shared` folder, but these are components, and I think they should all live together in a singular `/components` folder.

frontend-react/.storybook/main.js
```diff
@@ -8,6 +8,16 @@ module.exports = {
         "@storybook/addon-essentials",
         "@storybook/addon-interactions",
         "@storybook/addon-a11y",
+        {
```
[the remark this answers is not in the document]
etanb: I spent WAY too long trying to get our webpack config for the production app working with storybook using: https://storybook.js.org/docs/react/builders/webpack#import-a-custom-webpack-configuration. There's a ton of relative paths utilized in `paths.js` and simply adding one for Storybook created a slew of errors. I think it's very much worth combining them, so I agree with making a ticket! Let me know if you do and if you can link it here 🙏

frontend-react/package.json
etanb: agreed, I can revert this and see if it messes with anything.

frontend-react/src/components/shared/Table/Table.module.scss
etanb: Since this is sass + modules, doing:

```
.reportstreamTable {
  :global {}
 }
```

Scopes anything underneath the class `.reportstreamTable` with the css module name spacing protection BUT allows any classes within the `:global` space to not be affected by css module pattern. For example, if we nested a class without `:global`, like in this example:

```
.testClass {
  .subTestClass {
  }
}
```

The way we'd have to access that in our component would be: `className={styles.testClass.subTestClass}`. You couldn't do: `className={styles.testClass + ".subTestClass"}` for example. It needs to be escaped.

frontend-react/src/components/shared/Table/Table.module.scss
```diff
@@ -0,0 +1,53 @@
+@use 'uswds-core' as *;
+
+.reportstreamTable {
```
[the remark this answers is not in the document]
etanb: Using camel case is pretty conventional in css modules because we'd go from `className={styles.reportstreamTable}` to `className={styles['reportstreamTable']}`. Just want to confirm, is that what we want to do?

frontend-react/src/components/shared/Table/Table.module.scss
```diff
@@ -0,0 +1,53 @@
...
+    .usa-table td.active-col-td {
+        background-color: color('blue-10');
+    }
+
+    .usa-table thead .column-header-title-container {
+        display: flex;
+        align-items: center;
+    }
+
+    .usa-table thead .column-header-title-container p {
```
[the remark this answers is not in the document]
etanb: Yeah absolutely, we can just have `.column-header-title-container` here.

frontend-react/src/components/shared/Table/Table.module.scss
etanb: Fixed all the CSS classes to be more specific / tailored to a generic component.

frontend-react/src/components/shared/Table/Table.module.scss
etanb: The size only goes down to `size={3}`, when Jillian's design is really `2.5` or so. See here: https://designsystem.digital.gov/components/icon/#component-variants-icon

We can definitely talk about changing the icons to be inline with USWDS standards! I'll make that change for now.

frontend-react/src/components/shared/Table/Table.tsx
```diff
@@ -0,0 +1,225 @@
...
+}
+
+interface RowData {
+    columnKey: string;
+    content: string;
+    modalContent?: string;
+}
+
+export interface TableProps {
+    borderless?: boolean;
```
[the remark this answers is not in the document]
etanb: They are not! You can combine them however you see fit.

frontend-react/src/components/shared/Table/Table.tsx
etanb: Sounds good! Making that change.

etanb: Hmmmm....can you tell me the full `useState` you expect? The first value is just a static state variable so `onSortOrderChange` sounds more like the callback itself, but the second parameter for `useState` must start with `set`, so `const [onSortOrderChange, setOnSortOrderChange] = useState("")` looks a little strange to me? As if both parameters are callback functions.

frontend-react/src/components/shared/Table/Table.module.scss
etanb: I had an epiphany: we don't need the whole escape hatch with `:global` if we are just giving every element their own custom class, so we can remove the containing `.reportstreamTable` and just make custom classes per element. Will update the PR with this idea.

frontend-react/src/components/shared/Table/Table.tsx
```diff
@@ -0,0 +1,225 @@
...
+
+const SortableTableHeader = ({
+    columnTitle,
+    activeColumn,
+    sortOrder,
+    setSortOrder,
+    setActiveColumn,
+}: SortableTableHeaderProps) => {
+    let SortIcon = Icon.SortArrow;
+    const isActive = columnTitle === activeColumn;
```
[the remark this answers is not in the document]
etanb: I will be using `columnKey` for the rows, but the `columnHeaders` are generated dynamically. Should I include the `columnKey` in the `columnHeaders` output?

https://github.com/CDCgov/prime-reportstream/pull/9085/files#diff-4a8691b6c15e1d775d398ac71c10833729efd0af84f55106b12ff9677a836f60R174

frontend-react/src/components/shared/Table/Table.tsx
```diff
@@ -0,0 +1,225 @@
...
+            setSortOrder(FilterOptionsEnum.ASC);
+        } else if (sortOrder === FilterOptionsEnum.ASC) {
+            setSortOrder(FilterOptionsEnum.DESC);
+        } else if (sortOrder === FilterOptionsEnum.DESC) {
+            setSortOrder(FilterOptionsEnum.NONE);
+        }
+    };
+    return (
+        <th
+            role="button"
```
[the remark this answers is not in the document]
etanb: We talked about this a little bit in the previous Receivers Table PR, but using `<button/>` affects the native styling of tables. Is there a reason we can't just use `role="button"`? I believe that's what we settled on in the last PR.

frontend-react/src/components/shared/Table/Table.tsx
etanb: Let's talk about this feature synchronously! My gut feeling is that this should be in a separate PR / when we replace one of the tables that require API sorting.

frontend-react/src/components/shared/Table/Table.tsx
etanb: So if we want to utilize the modal in the future, each cell would have to have a instance of a modal?

frontend-react/src/components/shared/Table/Table.tsx
```diff
@@ -0,0 +1,225 @@
...
+            </thead>
+            <tbody>
+                {sortedData.map((row) => {
+                    return (
+                        <tr>
+                            {row.map((data) => {
+                                return (
+                                    <td
+                                        dangerouslySetInnerHTML={{
+                                            __html: sanitizeHtml(data.content),
```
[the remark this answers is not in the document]
etanb: So add a boolean, that defaults to a `string`, that will sanitize it for HTML usage or not? I can update that.

frontend-react/src/components/shared/Table/Table.tsx
etanb: Can do!

frontend-react/src/components/shared/Table/Table.tsx
```diff
@@ -0,0 +1,225 @@
...
+        </th>
+    );
+};
+
+const SortableTable = ({
+    rowData,
+    columnHeaders,
+}: {
+    rowData: RowData[][];
+    columnHeaders: string[];
```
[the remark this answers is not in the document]
etanb: Ah seeing this comment later down, but this answers my question here: https://github.com/CDCgov/prime-reportstream/pull/9085#discussion_r1169034662

Will make this change!

frontend-react/src/components/shared/Table/Table.tsx
etanb: Can you tell me more? This is a CSS Module reference.

frontend-react/src/components/shared/Table/Table.tsx
etanb: It's in the USWDS docs: https://designsystem.digital.gov/components/table/#accessibility-table

Check the section on scrollable tables.

frontend-react/src/components/shared/Table/Table.tsx
etanb: With the optional `isHTML` boolean + the highlighting of cells for sortable tables, there isn't much markup that's exactly the same and we'd then have to pass extra booleans for `isSortable` since the contents of both the `<th>` and `<td>` (ie the majority of content of a table) are different.

frontend-react/src/utils/misc.ts
```diff
@@ -175,3 +175,27 @@ export const parseFileLocation = (
...
+): string | number => {
+    let convertedContent = "";
+    if (!input) return convertedContent;
+
+    // Check if input is a date, then we could leave it as is,
+    // Or convert to 100% safe JS Date String
+    if (!!Date.parse(input)) {
+        convertedContent = new Date(input).toString();
+        // Check if contains HTML tags then strip them out
+    } else if (input.match(/<(.|\n)*?>/g)) {
```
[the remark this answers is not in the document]
etanb: Love it, done.

frontend-react/src/utils/misc.ts
etanb: The expected return value should be a `string`.

frontend-react/src/utils/misc.ts
etanb: Happy to chat about this! I thought it would be nice to use this function to sort of show what inputs we "support" and that we could keep expanding the list. We could make an `enum` and make this more explicit.

frontend-react/src/components/shared/Table/Table.tsx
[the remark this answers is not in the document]
etanb: Screenshot 2023-04-17 at 1 44 40 PM

frontend-react/src/utils/misc.ts
etanb: I'm going to move this out of `utils`, I agree it's too specific, especially with some of the other changes we've made.

frontend-react/.storybook/main.js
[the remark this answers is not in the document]
etanb: Zendoc link for reference: https://app.zenhub.com/workspaces/web-experience-607d9d5e68b95200150fec37/issues/gh/cdcgov/prime-reportstream/9109

frontend-react/src/components/shared/Table/Table.module.scss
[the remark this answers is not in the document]
etanb: Will change to use name of component as container class, ie `styles.Table`

frontend-react/src/components/shared/Table/Table.tsx
etanb: Updated based on our conversation today!

frontend-react/src/components/shared/Table/Table.stories.tsx
```diff
@@ -0,0 +1,103 @@
+import React from "react";
+import { ComponentMeta, ComponentStory } from "@storybook/react";
+
+import { Table } from "./Table";
+
+export default {
+    title: "components/Table",
```
penny-lischer, line 7: nit: should this be components/shared/Table ?

frontend-react/src/components/shared/Table/Table.tsx
```diff
@@ -0,0 +1,291 @@
...
+) {
+    return sortOrder !== FilterOptions.NONE && activeColumn
+        ? rowData.sort((a, b): number => {
+              const contentColA = convertToSortable(
+                  a.find((item) => item.columnKey === activeColumn)
+              );
+              const contentColB = convertToSortable(
+                  b.find((item) => item.columnKey === activeColumn)
+              );
+              console.log("contentColA = ", contentColA);
```
penny-lischer, line 116: Was this left intentionally?

penny-lischer (approved):
Few comments but overall LGTM

frontend-react/src/components/shared/Table/Table.stories.tsx
etanb: Absolutely need to move that! Thank you.

PR #9090, opened 2023-04-12
Bump org.flywaydb.flyway from 8.5.13 to 9.21.1 in /prime-router

snesm:
Reverts CDCgov/prime-reportstream#9086 and re-applies https://github.com/CDCgov/prime-reportstream/pull/7587

Bumps org.flywaydb.flyway from 8.5.13 to 9.20.0.

[Dependabot compatibility score](https://docs.github.com/en/github/managing-security-vulnerabilities/about-dependabot-security-updates#about-compatibility-scores)

## Updates required

Flyway [release notes](https://documentation.red-gate.com/fd/release-notes-for-flyway-engine-179732572.html)


oslynn:
Need to discuss with team regarding the new version of flyway.  Is it still having issue with staging deployment?

jack-h-wang:
I made some experimental changes to this PR but I think we need a branch that does not rename any existing migrations so they don't get flagged as an error; I cannot revert the migrations to their original state in this PR due to ktlint.

jack-h-wang (changes requested):
I suggest closing in favor of #10029 as I think the solution to this issue shouldn't involve renaming or modifying existing migrations - unless replacing all migrations with a new base migration. The changes to migrations cannot be rolled back in this branch without rolling back all changes.

luis-pabon-tf:
Blocked as it needs refactoring work

jack-h-wang:
Closing in favor of #10864.

PR #9128, opened 2023-04-13
change-kv-cert-exp-gha-runtime

lucasdze:
This PR is just to change the cron job timing


snesm (changes requested):
@jeremy-page suggested 7:07 (even though it was typo'd) to avoid all the other jobs that people drop in on the hour, which should improve performance a tiny bit.

Also, what timezone?  GitHub defaults to UTC.  You may need a conversion.

jeremy-page (approved):

snesm (approved):

PR #9158, opened 2023-04-14
Experience/9145/tooltip-component

### LINKED ISSUE #9145: Add tooltip components
The trussworks tooltip component uses logic created by its developers to handle the placement of the tooltip. Special elements such as dialogs, popovers, modals, and tooltips require constant updating to keep up with the latest css and accessibility standards while we wait for HTML specifications for those elements to materialize and become stable (ex: Dialog is now an official element but is still being iterated on, and a spec for popovers is in progress). We should opt to use a library to handle the complex placement logic for our tooltips (to handle cases such as screen-edge cutoff, etc.).

jpandersen87:
Closes #9145

This PR:
- adds @floating-ui/react library
- adds Tooltip, TooltipTrigger, and TooltipContent components that use @floating-ui/react
- adds resize-observer-polyfill library to support jest testing with @floating-ui/react


frontend-react/src/shared/Tooltip.tsx
```diff
@@ -0,0 +1,215 @@
...
+    open: controlledOpen,
+    onOpenChange,
+    offsetBy,
+}: TooltipOptions = {}) {
+    const [uncontrolledOpen, setUncontrolledOpen] = React.useState(initialOpen);
+    const isControlled = controlledOpen != null;
+    const open = controlledOpen ?? uncontrolledOpen;
+    const setOpen = React.useCallback(
+        (isOpen: boolean) => {
+            console.log("ding");
```
penny-lischer, line 42: Was this intentional?

penny-lischer (approved):
LGTM

frontend-react/src/shared/Tooltip.tsx
```diff
@@ -0,0 +1,215 @@
...
+import React from "react";
+
+export const TOOLTIP_POSITIONS = ["top", "bottom", "right", "left"];
+
+export interface TooltipOptions {
+    initialOpen?: boolean;
+    placement?: Placement;
+    open?: boolean;
+    onOpenChange?: (open: boolean) => void;
+    offsetBy?: OffsetOptions;
```
etanb, line 27: We should confirm with Jillian / Audrey about what the default should be here. As the default is way too closely cropped to the element it's stemming from:

Screenshot 2023-04-18 at 12 08 07 PM

We should also extrapolate for the Developer "approved" offsets in whole numbers like they do in [Trussworks library for Icons](https://github.com/trussworks/react-uswds/blob/main/src/components/Icon/Icon.tsx#L7) (and other custom sized components). So that a developer could pass something like `offset={1}`, `offset={2}`, `offset={3}`, etc. instead of having to pass in a [slew of options](https://floating-ui.com/docs/offset) (since they provide so much like crossAxis, mainAxis etc). I'm hoping our components can be "approved out of the box" instead of having to go to Design with what offsets should be.

frontend-react/package.json
```diff
@@ -5,6 +5,7 @@
     "npmClient": "yarn",
     "proxy": "http://127.0.0.1:7071",
     "dependencies": {
+        "@floating-ui/react": "^0.23.0",
```
stephenkao, line 8: Out of curiosity, how did you land on this library to use?

jpandersen87: I have previous experience with tippyjs and when looking at their docs about website support they recommended @floating-ui/react. The library used is completely up for debate, as long as it is in active development.

frontend-react/src/shared/Tooltip.tsx
```diff
@@ -0,0 +1,215 @@
...
+    useInteractions,
+    useMergeRefs,
+    FloatingPortal,
+    offset,
+    OffsetOptions,
+} from "@floating-ui/react";
+import classNames from "classnames";
+import React from "react";
+
+export const TOOLTIP_POSITIONS = ["top", "bottom", "right", "left"];
```
stephenkao, line 20: Could these be from the [`Side`](https://github.com/floating-ui/floating-ui/blob/master/packages/core/src/types.ts#L2) enum instead?

jpandersen87: Absolutely, thanks for pointing out an existing type.

frontend-react/src/shared/Tooltip.test.tsx
```diff
@@ -0,0 +1,113 @@
...
+describe("Tooltip", () => {
+    describe("useTooltipContext", () => {
+        test("no context - error", () => {
+            expect(() =>
+                renderApp(<TooltipContextReceiverComponent />)
+            ).toThrowError("Tooltip components must be wrapped in <Tooltip />");
+        });
+
+        test("context", () => {
+            renderApp(
```
stephenkao, line 46: Along the same lines of the previous point, I think `renderHook` also takes a `wrapper`.  Either way works, though!  Definitely not a big deal!

frontend-react/src/shared/Tooltip.test.tsx
```diff
@@ -0,0 +1,113 @@
...
+                </TooltipContextCreatorComponent>
+            );
+            expect(screen.getByText("Test")).toBeInTheDocument();
+        });
+    });
+
+    describe("Tooltip", () => {
+        test("creates tooltip context", () => {
+            renderApp(
+                <Tooltip placement={"Test" as any}>
```
stephenkao, line 58: Should this be allowed or should it be limited to the four possible options?

jpandersen87: I should've just followed the typing and passed a valid string like bottom but it's obvious I was brain fried at this point haha. Will fix.

stephenkao: No problem!  I definitely understand 😆

frontend-react/src/shared/Tooltip.tsx
```diff
@@ -0,0 +1,215 @@
...
+        move: false,
+        enabled: !isControlled,
+    });
+    const focus = useFocus(context, {
+        enabled: !isControlled,
+    });
+    const dismiss = useDismiss(context);
+    const role = useRole(context, { role: "tooltip" });
+
+    const interactions = useInteractions([hover, focus, dismiss, role]);
```
stephenkao, line 77: Whoaaaa this feels like such an odd signature 😆

frontend-react/src/shared/Tooltip.tsx
```diff
@@ -0,0 +1,215 @@
...
+export interface TooltipOptions {
+    initialOpen?: boolean;
+    placement?: Placement;
+    open?: boolean;
+    onOpenChange?: (open: boolean) => void;
+    offsetBy?: OffsetOptions;
+}
+
+export function useTooltip({
+    initialOpen = false,
```
stephenkao, line 31: Minor, minor point, but I wonder if we could just infer that this is controlled/uncontrolled.  Like if `open` is a boolean, it's controlled; otherwise, it's uncontrolled?

frontend-react/src/shared/Tooltip.tsx
```diff
@@ -0,0 +1,215 @@
...
+    const tooltip = useTooltip(options);
+    return (
+        <TooltipContext.Provider value={tooltip}>
+            {children}
+        </TooltipContext.Provider>
+    );
+}
+
+export type TooltipTriggerPropsBase = React.HTMLProps<HTMLElement> & {
+    asChild?: boolean;
```
stephenkao, line 129: Could we just pass in the forward ref as a prop instead of specifying a boolean and then children?  Not sure if that'd be better or worse, honestly -- just a thought!  😬

frontend-react/src/shared/Tooltip.tsx
```diff
@@ -0,0 +1,215 @@
...
+    HTMLElement,
+    TooltipTriggerPropsBase
+>(function TooltipTrigger({ children, asChild = false, ...props }, propRef) {
+    const context = useTooltipContext();
+    const childrenRef = (children as any).ref;
+    const ref = useMergeRefs([context.refs.setReference, propRef, childrenRef]);
+
+    // `asChild` allows the user to pass any element as the anchor
+    if (asChild && React.isValidElement(children)) {
+        // Not ideal to have to resort to manipulating the children data,
```
stephenkao, line 150: 👍

PR #8966, opened 2023-04-03
Experience/8944/apiguide reportstream api page

jpandersen87:
Closes #8944 

This PR adds the reportstream-api index page of the api guide. This includes the main mdx article, the page component, and the app router entry.

PR #9199, opened 2023-04-18
Sidenav route integration

### LINKED ISSUE #9198: Sidenav route integration
In the API Programmer's Guide work, there is want of a sidenav that can be dictated by a data structure. Since we already now represent our routes as data structures as well, it feels like these two areas overlap and there is a possibility of simply updating how our routes are laid out and exposed so that we can use routes as our source of truth when creating UI navigation structures.

jpandersen87:
Fixes #9198

This PR:
- Creates route inheritance inside the pages/resources directory
- Adds a new lazy helper to allow for lazy-loading components that are not the default export
- Uses lazy loading on some resources routes
- Enhances the RouteObjects to RSRouteObjects to allow for custom properties
- Defines api programmer's guide routes with title properties
- Adds new helpers for sidenav item creation
- Creates a Sidenav component for use in api programmer's guide pages that uses routes

### DIFF
--- a/frontend-react/src/utils/misc.ts
+++ b/frontend-react/src/utils/misc.ts
@@ -1,3 +1,9 @@
+import React from "react";
+import { ConditionalPick, ReadonlyDeep } from "type-fest";
+import { RouteObject } from "react-router-dom";
+
+import { RSRouteObject } from "./UsefulTypes";
+
 /**
  * splitOn('foo', 1);
  * // ["f", "oo"]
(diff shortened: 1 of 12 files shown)


frontend-react/src/utils/misc.ts
```diff
@@ -175,3 +181,55 @@ export const parseFileLocation = (
...
+        return React.lazy(fn);
+    }
+
+    return React.lazy(async () => {
+        const module = await fn();
+        return { default: module[namedExport] };
+    });
+}
+
+export function matchRoute(
```
penny-lischer, line 206: Can we add a unit test for this?

penny-lischer:
Besides adding some unit tests LGTM

jpandersen87:
punting this due questions regarding our app's router's future. will use the statically built sidenav in #8966

PR #9019, opened 2023-04-06
Experience-8928: Add Chromatic GHA

stephenkao:
This changeset adds a Chromatic GHA workflow to deploy our Storybook to Chromatic.  This'll run on any `push` event on any branch, and it'll automatically approve any changes on `master.  Hopefully this'll help during design planning and review since we'll have a central component styleguide!

## Linked Issues
- Fixes #8928

PR #9236, opened 2023-04-21
Experience-9235: Fix storybook command

stephenkao:
In https://github.com/CDCgov/prime-reportstream/pull/9019, we removed the `storybook:start` command, but it was being used for the base `storybook` command.  This changeset just restores the command so everything works again.


penny-lischer (approved):
LGTM

PR #9250, opened 2023-04-21
BUG: hash anchor links are using react-router

### LINKED ISSUE #9249: BUG: hash anchor links are using react-router
The getHrefRoute function for our USLinks needs to check if the href is going to a hash anchor route and return undefined so that it becomes a normal link rather than a react-router link.

jpandersen87:
Fixes #9249

This PR fixes links to hash anchors (ex: a heading on a local page) to bypass react-router so that the browser can properly scroll to the hash reference. Research will need to be done on how best to handle this within react so that we can go back to using react-router to prevent full page refreshes.

### DIFF
--- a/frontend-react/.storybook/main.js
+++ b/frontend-react/.storybook/main.js
@@ -1,5 +1,14 @@
 "use strict";
 
+const remarkPluginFiles = ["remark-gfm"];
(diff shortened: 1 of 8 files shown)


stephenkao (approved):
Looks good to me!

penny-lischer (approved):
LGTM

PR #9254, opened 2023-04-24
change job runtime from 10PM to 3 AM

lucasdze:
This PR is just to change the runtime from 10 to to 3 AM


jeremy-page:
Please change it to run at "7 13 * * *" it’s a best practice not to run jobs on the hour. (this will make it run at 8:07 AM ET)

Issue #9144, opened 2023-04-14
UP Should support sending FHIR to receivers


arnejduranovic:
### User Story
As a receiver of the Universal Pipeline, I want to be able to receive FHIR.

### Description/Use Case
The UP translate step only supports translating to HL7. We want the ability to skip transforming the internal FHIR message to HL7 if the receiver indicates, via their settings, they want to receive FHIR. 

### Dev Notes:
UP currently only supports HL7 (See `FHIRTranslator.doWork`). Inside that function, instead of converting bundle to HL7, we want to do something like this, probably: `FhirContext.forR4().newJsonParser().encodeResourceToString(updatedBundle)`

Whatever solution we come up with, we need to make sure it is [open-closed](https://stackify.com/solid-design-open-closed-principle/), meaning we can add future formats without directly modifying the doWork function. That probably means some refactoring will need to occur.

We also need to implement support for FHIR->FHIR transforms

### Acceptance Criteria
- [ ] UP can send FHIR to receivers
- [ ] Receiver FHIR->FHIR transforms supported
- [ ] Verified to check if Provenance resource shows that the bundle is coming from ReportStream (and other programs if they modified the provenance as well). If this is not the case, created a ticket for this work. Linked here: TODO

arnejduranovic:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/9144?planning-poker) with Zenhub @TurkeyFried @victor-chaparro @JessicaWNava

jimmyfagan:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/9144?planning-poker) with Zenhub @jimmyfagan

jack-h-wang:
At the routing step, all intended receivers are added to Provenance as Endpoint resources. We may need to consider removing the Endpoint resources before the bundle is sent to a receiver.

PR #9261, opened 2023-04-24
Support FHIR receivers and receiver transforms

jimmyfagan:
This PR adds all of the necessary functionality to allow for receivers to be created in the FULL_ELR topic with a FHIR format to receive FHIR messages. This involved a lot of pieces to make things work seamlessly, which I will break down below in the `Changes` section.

Test Steps:
1. Reload the settings with `./gradlew reloadSettings`
2. Send a report with a state of `IG` to trigger the ignore receivers
3. Check the sftp folder for resulting batched messages; you should see one in HL7 format and one in FHIR format

## Changes
- Added a placeholder receiver transform to verify that receiver fhir transforms are applied
- Added a new FHIR receiver to the ignore organization and update smoke tests to account for that (including addition of `TestReportStream.getAllChildrenReportId`
- Added `useBatching` getter to `Receiver` to determine whether or not to batch messages for a receiver
- Moved `HL7MessageHelpers.takeHL7GetReport` to `Report.generateReportAndUploadBlob` (with basically no changes)
- Added `FHIRConfiguration` subtype of `TranslatorConfiguration` and add `useBatching` getter to `TranslatorConfiguration` (used in conjunction with the same getter on `Receiver`)
- Added `FHIRBundleHelpers.batchMessages` to batch FHIR messages into a single `fhir+ndjson` file
- Updated the default processing mode filter to prevent exceptions
- Refactored `FHIRTranslator` to allow for
  - flexibility to translate to HL7v2 or FHIR
  - clean pruning of endpoints/conditions that shouldn't be sent to a given receiver
- Added/adjusted tests

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- Addressed #9144


oslynn:
When it is ready.  Would you please include me for review?  I just got the REST Transport to function with ETOR under the integration test.

prime-router/src/main/kotlin/Receiver.kt
```diff
@@ -141,6 +141,9 @@ open class Receiver(
     @get:JsonIgnore
     val format: Report.Format get() = translation.format
 
+    @get:JsonIgnore
+    val useBatching: Boolean get() = translation.useBatching
```
arnejduranovic, line 145: What is the purpose of this change?

prime-router/src/main/kotlin/TranslatorConfiguration.kt
```diff
@@ -160,38 +166,59 @@ data class Hl7Configuration
...
-            "receiving_application" to receivingApplication,
-            "receiving_facility" to receivingFacility,
-            "message_profile_id" to (messageProfileId ?: ""),
-            "reporting_facility" to reportingFacility
-        )
-    }
+}
+
+/**
+ * Standard HL7 report configuration
```
arnejduranovic, line 208: Standard `FHIR` report configuration?

arnejduranovic (changes requested):

prime-router/src/main/kotlin/fhirengine/engine/FHIRTranslator.kt
```diff
@@ -109,6 +110,30 @@ class FHIRTranslator(
         }
     }
 
+    /**
+     * Returns a byteArray representation of the bundle in a format [receiver] expects, or throws an exception if the
```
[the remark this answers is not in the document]
jimmyfagan:
```suggestion
     * Returns a byteArray representation of the [bundle] in a format [receiver] expects, or throws an exception if the
```

prime-router/src/main/kotlin/TranslatorConfiguration.kt
jimmyfagan:
```suggestion
 * Standard FHIR report configuration
```

prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
```diff
@@ -254,4 +259,13 @@ object FHIRBundleHelpers {
         }
         return observationsToKeep
     }
+
+    fun batchMessages(messages: List<String>): String {
```
victor-chaparro, line 263: Missing kdoc

victor-chaparro (commented):
Looks good so far. Great code coverage just have so minor feedback.

prime-router/src/main/kotlin/TranslatorConfiguration.kt
```diff
@@ -160,38 +166,59 @@ data class Hl7Configuration
...
-        val reportingFacility = when {
-            reportingFacilityName != null && reportingFacilityId != null && reportingFacilityIdType == null ->
-                "$reportingFacilityName^$reportingFacilityId^CLIA"
-            reportingFacilityName != null && reportingFacilityId != null && reportingFacilityIdType != null ->
-                "$reportingFacilityName^$reportingFacilityId^$reportingFacilityIdType"
-            reportingFacilityName != null && reportingFacilityId == null ->
-                reportingFacilityName
-            else -> ""
+    override val useBatching: Boolean get() = useBatchHeaders
+
```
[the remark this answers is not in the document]
jimmyfagan: I think when that has come up in the past, we have said that we're not expecting to need to support that, but if we did, this framework allows for that decently well; we would probably just introduce a `useBatching` json property for hl7 configs and then use that property instead of overriding the getter to return `useBatchHeaders` like we're doing here. Then in the code, we'd just have to handle `useBatching` and `useBatchHeaders` separately. So I don't think we're boxing ourselves in here, and we can handle that separation in it's own ticket if/when that requirement shows up.

arnejduranovic (commented):
Looks good just need to test locally

victor-chaparro (approved):

prime-router/src/main/kotlin/TranslatorConfiguration.kt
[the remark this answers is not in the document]
victor-chaparro: Something else I just thought about is if all FHIR receivers would want to receive batch FHIR in ndjson or if we would need to support another format like sending the bundles in a JSON array. But we can probably worry about that later. Do we know what's needed for the connectathon?

arnejduranovic:
I ran the RadxMars FHIR test message through and it got delivered as FHIR! I compared the input message vs the output and found these differences:

1. output added the following under `messageHeader.meta`:
```
"tag": [
            {
              "code": "P"
            }
          ]
```
2. output changed `messageHeader.source.name `from "ACME Application Server" to "CDC PRIME - Atlanta: 2.16.840.1.114222.4.1.237821" but not other info in the source object like `software` or `contact.value` (phone number).
Also changed `source.endpoint` from http://at-home-in-vitro-test.org/fhir/endpoint/1 to urn:oid:2.16.840.1.114222.4.1.237821

3. Output had following provenance resource:

```
    {
      "resource": {
        "resourceType": "Provenance",
        "target": [
          {
            "reference": "https://fhir.org/fhir/DiagnosticReport/diagnosticreport-at-home-in-vitro-results-covid-p-noelle"
          },
          {
            "reference": "Endpoint/1683128407232487000.dec50ddb-77dd-46d4-8311-72cdcbca69db"
          }
        ]
      }
    },
```

4. output had following Endpoint resource:

```
    {
      "fullUrl": "Endpoint/1683128407232487000.dec50ddb-77dd-46d4-8311-72cdcbca69db",
      "resource": {
        "resourceType": "Endpoint",
        "id": "1683128407232487000.dec50ddb-77dd-46d4-8311-72cdcbca69db",
        "identifier": [
          {
            "system": "https://reportstream.cdc.gov/prime-router",
            "value": "development.DEV_FULL_ELR_FHIR"
          }
        ],
        "status": "active",
        "name": "Development FULL_ELR_FHIR"
      }
    }
```

jimmyfagan:
Just to record the findings from our quick huddle on those differences:

1. This is the processing mode and signifies production data. If we need this to be different or to not be set, that can hopefully be handled through receiver transforms or by adding a new setting similar to `useTestProcessingMode` in Hl7 receivers. No action required at this time.
2. This is the result of the default sender transform. We need to find out more about what is generally expected of the message header and the provenance to determine what makes sense as a default. We may want to adjust the default transform to only set a new source if one isn't set, and we may want to adjust it to overwrite all fields of the source instead of just the name and the url. Either way, that can hold off until we are working with specific senders, so no action required at this time.
3. This is expected, so no action required at this time.
4. This is expected, so no action required at this time.

PR #9305, opened 2023-04-25
add linting for console.log

etanb:
This PR adds `"no-console": ["error", { allow: ["warn", "error", "info", "trace"] }],` rule to the `eslint` file so that we can't accidentally commit `console.log` specifically anymore.

Here's the output showing that all other `console` statements get through besides `console.log`:

Screenshot 2023-04-25 at 2 23 36 PM


etanb:
@penny-lischer @jpandersen87 not sure if the right move is to swap some of these to `console.error` but that seemed more accurate than just logging them. I'm also pro just removing the statement entirely.

frontend-react/src/hooks/UseValueSets.ts
```diff
@@ -32,7 +32,7 @@ const findTableMetaByName = (
     );
 
     if (!filteredBody.length) {
-        console.log("Unable to find metadata for lookup table: ", tableName);
+        console.error("Unable to find metadata for lookup table: ", tableName);
```
penny-lischer, line 35: maybe console.info here instead of error.

etanb: So change all the more informal `console.log` we have around the production codebase to `console.info`?

penny-lischer: No, i'm just suggesting for this specific console to be console.info since it doesn't seem like an error.

etanb: Sounds good.

penny-lischer (approved):
lgtm

PR #9306, opened 2023-04-25
Injecting the provenance resource when it is missing

### LINKED ISSUE #9125: Inject Provenance resource when missing
### User Story:
As a Connectathon attender, I want to ensure that people can send me messages that do not contain a provenance without erroring out. In the future, as a Reportstream sender, I want to be able to successfully send messages without a provenance. 

### Description/Use Case
We currently produce an error when a message is sent without a provenance. Instead of throwing an error, we want to inject the provenance message so that the error is not produced. 

### Acceptance Criteria
- [x] A message can be sent without the provenance and it will not fail.

JessicaWNava:
This PR adds a provenance resource when one is missing as well as sets the processing mode filter to evaluate to true so that the sample message for Connectathon can be processed as is. 

Test Steps:
1. Navigate to http://build.fhir.org/ig/HL7/home-lab-report/Bundle-bundle-at-home-in-vitro-test-results-p-noelle.json.html to get the test message
2. Update the state in the message to be IG
3. Send the message through /reports
4. Check that the message is sent to an organization via /waters/report/<insert_id_here>/history

## Changes
- Add the provenance resource when it is missing from the message
- Set the processing mode filter to something that will make it evaluate to true. 

### Testing
- [x] Tested locally?
- [ ] Added tests?

### Process

## Linked Issues
- Fixes #9125 

## To Be Done
- #issue

    - new endpoint A
    - new endpoint B


prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
```diff
@@ -80,7 +80,8 @@ object FHIRBundleHelpers {
         val provenanceResource = try {
             fhirBundle.entry.first { it.resource.resourceType.name == "Provenance" }.resource as Provenance
         } catch (e: NoSuchElementException) {
-            throw IllegalStateException("The FHIR bundle does not contain a Provenance resource")
+            fhirBundle.addEntry().resource = Provenance()
+            fhirBundle.entry.first { it.resource.resourceType.name == "Provenance" }.resource as Provenance
```
jimmyfagan, line 84: Does this new behavior need to be tested in unit tests? Or is that not necessary since this is just for connectathon?

prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
```diff
@@ -80,7 +80,8 @@ object FHIRBundleHelpers {
         val provenanceResource = try {
```
jimmyfagan, line 80: Again, depends on if this is just for connectathon or if it'll eventually be merged into master, but this function is kind of doing two things, adding a provenance and adding references to reports within the provenance, it might be nice to split out the piece that does the try/catch since that's currently happening twice in this path. Not essential, especially if this is just for connectathon and won't be merged into master.

jimmyfagan (commented):
Looks basically good, still had some comments, mostly questions. A lot of the needs here depend on what long term plans are, don't necessarily need things to be optimized or thoroughly unit-tested for a temporary feature branch. Also not that this comes up a lot, but for PRs into feature branches, it can be helpful to highlight that in the PR description; I was confused comparing this to `master` initially 😆

arnejduranovic:
Hmm, I think we should aim to merge this with master. The reason this isn't on master yet is because its not a problem, for now, since we don't support sending FHIR. However, with the completion of #9261, master branch will support sending FHIR,  and so this should go to master as well?

JessicaWNava:
> Hmm, I think we should aim to merge this with master. The reason this isn't on master yet is because its not a problem, for now, since we don't support sending FHIR. However, with the completion of #9261, master branch will support sending FHIR, and so this should go to master as well?

Should I maybe split this into two PRs then? One that will add the provenance and unit testing. The other that will go to the connectathon branch and update the organizations.yml file since it seems like we are using the ignore org? Or are we not planning to use the ignore org for the connectathon @arnejduranovic ?

JessicaWNava:
Moving this PR to draft until I get it ready for a master branch change rather than a quick connectathon change.

prime-router/settings/organizations.yml
```diff
@@ -190,13 +190,15 @@
...
+      processingModeFilter:
+            - "(%performerState.exists() and %performerState = 'IG') or (%patientState.exists() and %patientState = 'IG')"
       timing:
         operation: MERGE
         numberPerDay: 1440 # Every minute
         initialTime: 00:00
       translation: !<HL7>
         schemaName: "metadata/hl7_mapping/STLTs/CO/CO"
-        useTestProcessingMode: false
+#        useTestProcessingMode: false
```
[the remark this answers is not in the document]
JessicaWNava: I will revert all of the changes to the organizations.yml file since this is going into main now.

prime-router/settings/organizations.yml
```diff
@@ -190,13 +190,15 @@
       jurisdictionalFilter: [ "(%performerState.exists() and %performerState = 'IG') or (%patientState.exists() and %patientState = 'IG')" ]
       qualityFilter:
         - "(%performerState.exists() and %performerState = 'IG') or (%patientState.exists() and %patientState = 'IG')"
+      processingModeFilter:
+            - "(%performerState.exists() and %performerState = 'IG') or (%patientState.exists() and %patientState = 'IG')"
```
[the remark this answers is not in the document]
JessicaWNava: Good callout, I am sorry I wasn't clearer. I will make sure to specify next time. Asking Arnej what org we will be using, but since the changes are going to master, I will revert this.

prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
JessicaWNava: The catch part of the try/catch is different though. In the addProvenanceReference method it is creating the provenance. In the addReceivers method it calls the addProvenanceReference because we not only want to create the provenance, we want to populate it. However, we do not want to call that method if the provenance already exists. Does that make sense?

jimmyfagan: Yea it does, do we only want/need to populate it if we're creating it for the first time, or should we maybe just be calling `addProvenanceReference` either way?

prime-router/settings/organizations.yml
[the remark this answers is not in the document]
arnejduranovic: For the connectathon, we can make new sender/receivers under the development ORG.

prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
JessicaWNava: The second call is in the addProvenanceReference method, so it would not make sense for it to call itself.

jimmyfagan: Sorry confusing comment, I'm going to resolve this thread and comment at the piece of code I'm talking about.

prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
```diff
@@ -38,8 +38,8 @@ object FHIRBundleHelpers {
...
             fhirBundle.entry.first { it.resource.resourceType.name == "Provenance" }.resource as Provenance
         } catch (e: NoSuchElementException) {
-            // TODO: Create Provenance here
-            throw IllegalStateException("The FHIR bundle does not contain a Provenance resource")
+            addProvenanceReference(fhirBundle)
+            fhirBundle.entry.first { it.resource.resourceType.name == "Provenance" }.resource as Provenance
         }
```
jimmyfagan, line 43: Do we need this try/catch or can we just always call `addProvenanceReference`? Could look something like this:

```
addProvenanceReference(fhirBundle)
val provenanceResource = fhirBundle.entry.first { it.resource.resourceType.name == "Provenance" }.resource as Provenance
```

Or if `addProvenanceReference` is updated to return a reference to that provenance, it could be just 
```
val provenanceResource = addProvenanceReference(fhirBundle)
```

This is optional, but has two benefits:

- (minor) eliminates one of the two near-identical try/catch blocks in this path
- ensures that all provenances after this point, whether they existed already or not, have diagnostic report targets populated

JessicaWNava: I don't know. I think I was assuming that we didn't want to change it if it already existed and was populated. I am not sure the code is set up to work with ones that are already populated, but I could be wrong. Maybe @arnejduranovic could weigh in on this?

jimmyfagan (commented):
Looks good, just one optional comment at this point. Will keep an eye out for a response on that and then give my approval!

jimmyfagan (approved):
Going ahead and approving, changes are good either way, and Arnej can weigh in on the open thread when he does his pass.

prime-router/src/test/kotlin/fhirengine/engine/fhirRouterTests/FHIRBundleHelpersTests.kt
```diff
@@ -152,6 +173,24 @@ class FHIRBundleHelpersTests {
         assertThat(receiversOut[0]).isEqualTo("co-phd.full-elr-hl7")
     }
 
+    @Test
+    fun `test adding receivers to bundle without provenance`() {
+        // set up
+        val bundle = FhirTranscoder.decode(validFhirWithoutProvenance)
```
arnejduranovic, line 179: If I was doing this, I would probably just reuse the existing file in resources? something like:
```
val fhirBundle = File("src/test/resources/fhirengine/engine/valid_data_no_provenance.fhir").readText()
val messages = FhirTranscoder.getBundles(fhirBundle, actionLogger)
assertThat(messages).isNotEmpty()
val bundle = messages[0]
```

prime-router/src/test/kotlin/fhirengine/engine/fhirRouterTests/FHIRBundleHelpersTests.kt
```diff
@@ -174,6 +213,28 @@ class FHIRBundleHelpersTests {
         assertThat(references).isNotEmpty()
     }
 
+    @Test
+    fun `test adding diagnosticreport references to bundle when no original provenance`() {
```
arnejduranovic, line 217: If we do Jimmy's change, let's add a test similar to this but that makes sure that if the diagnostic report references already exist that they don't get doubled.

arnejduranovic (changes requested):
Looks good, thank you!! I added a question in the try/catch thread that we should look into before approving.

prime-router/src/test/kotlin/fhirengine/engine/fhirRouterTests/FHIRBundleHelpersTests.kt
JessicaWNava: Yes, the diagnostic reports will get duplicated currently. It does not check if it already exists, just blindly adds them, so I can create a tech debt ticket for this.

prime-router/src/test/kotlin/fhirengine/engine/fhirRouterTests/FHIRBundleHelpersTests.kt
JessicaWNava: Updated

arnejduranovic (approved):

JessicaWNava:
Just as a follow up, discussed with Arnej and decided we were not creating a tech debt ticket. Tests are updated.

Issue #9224, opened 2023-04-20
Bug: Incorrect data is written to the database when UP routing has no receivers


jack-h-wang:
### User Story

As a ReportStream developer, I want the action table to indicate there are no items and not upload to Azure if there are no receivers to route to, so that I can properly track routing in the UP.

### Description/Use Case

If there is no data that passes the filters within the UP routing function, then by design the data proceeds no further. The information written to the database in this circumstance needs to reflect this more clearly.

### Risks/Impacts/Considerations

The work done here may inform the work to be done for #8451

### Dev Notes

Currently, the Universal Pipeline routing function has alternate logic that skips sending a report to the translate function if it determines there are no receivers for the data it is routing. This logic needs to be updated further to set the `item_count` in the action table to 0, and to not upload an Azure blob.

For previous work and instructions to reproduce the scenario, see the previous issue: #7545

### Acceptance Criteria
- [ ] If UP routing has no receivers, `item_count` must be set to 0.
- [ ] If UP routing has no receivers, an Azure blob should not be uploaded.

bishoyayoub:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/9224?planning-poker) with Zenhub @TurkeyFried

bishoyayoub:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/9224?planning-poker) with Zenhub @PatriciaPerozoUSDS

Issue #8451, opened 2023-02-22
BUG: UP Submission History reports "receiving" when data goes to no receivers


jack-h-wang:
## User Story:

As an administrator, I want the Submission History overall status to be able to indicate that a report was not delivered to any receivers so that the correct status is reported.

## Description/Use Case

The submission history currently reports an overall status of "receiving" in the instance a report has no receivers. It does this in two distinct cases:

- Data hasn't been processed yet (common in async submissions)
- Very rare: No data matches any geographical location.

At the moment the function does not distinguish between these cases. This needs to be changed so that the overall status returned in the Submission History reflects what actually occurred.

## Risks/Impacts/Considerations

- The existing behavior for when data is not processed yet needs to be preserved.

## Dev Notes

- As of #7545 the UP routing function adds a "none" event in the case of the data having no receivers. This is also reflected in the "next action" of the "routing" event.
- As of #9224 some additional changes were made, including setting the item count to 0 in the "none" event.

## Acceptance criteria

- The Submission History reports an overall status of "not delivering" when UP routing has determined there are no receivers.
- The Submission History reports an overall status of "receiving" when data is still waiting for routing to start.

bishoyayoub:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/8451?planning-poker) with Zenhub @arnejduranovic @jack-h-wang @JessicaWNava @jimmyfagan @thetaurean @victor-chaparro

mkalish:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/8451?planning-poker) with Zenhub @mkalish

PR #9307, opened 2023-04-26
9224 correct database entries when UP routing sends to no receivers

jack-h-wang:
This PR makes changes that correct issues with the database entries written when the UP routing function sends to no receivers.

Test Steps:
1. Use the `Single HL7 Full ELR` test to submit data to be routed. Ensure the data sent does not have a valid receiver.
2. Retrieve the report `id` from the data returned from the API call.
3. Allow time for the data to be routed, then confirm the entry written for the `none` event on the `report_file` table contains an `item_count` of `0` and a `body_url` of `null`.

## Changes
- UP routing function no longer attempts to upload an Azure blob attached to the `none` event when there are no receivers.
- Update ActionHistory functions to allow creating reports without Azure blobs.
- Add additional tests for ActionHistory functions.
- Updated the test for UP routing no receivers.

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- #9224

## To Be Done
- `trackCreatedReport`'s overloads can probably be refactored into a single function. This function is used heavily, however, and we would want to consider how much of this function should remain when the legacy pipeline is retired. Ticket created: #9345
- The changes in this PR may inform the implementation of #8451


prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -510,10 +510,15 @@ class ActionHistory(
...
-        reportFile.blobDigest = blobInfo.digest
-        reportFile.itemCount = report.itemCount
+        if (blobInfo != null) {
+            reportFile.bodyUrl = blobInfo.blobUrl
+            reportFile.bodyFormat = blobInfo.format.toString()
+            reportFile.blobDigest = blobInfo.digest
+            reportFile.itemCount = report.itemCount
+        } else {
+            reportFile.bodyFormat = "FHIR" // currently only the UP sends null blobs
+            reportFile.itemCount = 0
```
victor-chaparro, line 520: Are we not allowed to set to bodyFormat to Null, if there's no report? If we need to set it, it might better to use the `Report.Format` enum.

victor-chaparro (approved):
Tested it locally sending reports to receivers and no receivers and data looks good. Just had one small comment.

prime-router/src/main/kotlin/azure/ActionHistory.kt
jack-h-wang: Correct, bodyFormat cannot be null. I've changed the code to use the `Report.Format` enum for this value.

arnejduranovic:
Can we please add a tech debt ticket for trackCreatedReport refactor and link here? Can we also update https://github.com/CDCgov/prime-reportstream/issues/8451 to mention this ticket?

arnejduranovic (changes requested):
Look good! Please see comment regarding creating new ticket and updating existing ticket.

arnejduranovic (approved):

jack-h-wang:
> Can we please add a tech debt ticket for trackCreatedReport refactor and link here? Can we also update #8451 to mention this ticket?

#8451 is already linked to the issue this PR is linked to. I've created #9345 to cover the trackCreatedReport refactor.

PR #9337, opened 2023-04-28
8451: return a receiving status if there are items, but no destinations yet

mkalish:
This PR ...

Test Steps:

## Changes

### Testing
- [ ] Tested locally?
- [ ] Added tests?

### Process

## Linked Issues
- Fixes #issue

## To Be Done
- #issue

    - new endpoint A
    - new endpoint B

### DIFF
--- a/prime-router/src/main/kotlin/history/SubmissionHistory.kt
+++ b/prime-router/src/main/kotlin/history/SubmissionHistory.kt
@@ -163,6 +163,7 @@ class DetailedSubmissionHistory(
         ERROR("Error"),
         RECEIVED("Received"),
         NOT_DELIVERING("Not Delivering"),
+        RECEIVING("Receiving"),
         WAITING_TO_DELIVER("Waiting to Deliver"),
         PARTIALLY_DELIVERED("Partially Delivered"),
         DELIVERED("Delivered");
@@ -512,15 +513,13 @@ class DetailedSubmissionHistory(
         }
 
         if (destinations.size == 0) {
-            /**
-             * Cases where this may hit:
-             *     1) Data hasn't been processed yet (common in async submissions)
-             *     2) Very rare: No data matches any geographical location.
-             *         e.g. If both the testing tab and patient data were foreign addresses.
-             * At the moment we have NO easy way to distinguish the latter rare case,
-             * so it will be treated as status RECEIVED as well.
-             */
-            return Status.RECEIVED
+            return if (reportItemCount != 0) {
+                // Data hasn't been processed yet (common in async submissions)
+                Status.RECEIVING
+            } else {
+                // Very rare: No data matches any geographical location.
+                Status.NOT_DELIVERING
+            }
         } else if (realDestinations.isEmpty()) {
             return Status.NOT_DELIVERING
         }
--- a/prime-router/src/test/kotlin/history/SubmissionHistoryTests.kt
+++ b/prime-router/src/test/kotlin/history/SubmissionHistoryTests.kt
@@ -531,7 +531,7 @@ class SubmissionHistoryTests {
         )
         testReceived.enrichWithSummary()
         testReceived.run {
-            assertThat(overallStatus).isEqualTo(DetailedSubmissionHistory.Status.RECEIVED)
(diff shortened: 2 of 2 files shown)


prime-router/src/main/kotlin/history/SubmissionHistory.kt
```diff
@@ -512,15 +512,13 @@ class DetailedSubmissionHistory(
...
-             *         e.g. If both the testing tab and patient data were foreign addresses.
-             * At the moment we have NO easy way to distinguish the latter rare case,
-             * so it will be treated as status RECEIVED as well.
-             */
-            return Status.RECEIVED
+            return if (reportItemCount != 0) {
+                // Data hasn't been processed yet (common in async submissions)
+                Status.RECEIVED
+            } else {
+                // Very rare: No data matches any geographical location.
```
arnejduranovic, line 519: Is this comment accurate? A report can end up passing the jurisdictional filter (geographical location) but fail another one, like the quality filter, and go nowhere, right?

mkalish: I can look into this some more, but I was just maintaining the existing comment.

mkalish: So I think the under case is handled by the conditional starting on 522L.  `realDestinations` are the original destinations but filtered down to the ones that they actually had an item delivered.

jimmyfagan:
I haven't been able to hit the `NOT_DELIVERED` case you added; I can verify that the request response and the submission history both show "Received", but I think it was doing that before too. Maybe I'm missing something, but I might just need a bit more detailed test steps to see how to test the changes here.

prime-router/src/main/kotlin/history/SubmissionHistory.kt
```diff
@@ -512,15 +512,13 @@ class DetailedSubmissionHistory(
...
-             *     2) Very rare: No data matches any geographical location.
-             *         e.g. If both the testing tab and patient data were foreign addresses.
-             * At the moment we have NO easy way to distinguish the latter rare case,
-             * so it will be treated as status RECEIVED as well.
-             */
-            return Status.RECEIVED
+            return if (reportItemCount != 0) {
+                // Data hasn't been processed yet (common in async submissions)
+                Status.RECEIVED
+            } else {
```
jimmyfagan, line 518: In testing, I still can't hit this else case to hit this new `NOT_DELIVERED`, seems like the `reportItemCount` is a count of how many items are in the report, not of how many receivers get the report, so don't think this works quite right.

mkalish: I'm going to circle back with Luis to see what he was using to test this when he wrote the original comment as I'm having a hard time hitting this condition as well outside of tests.

jack-h-wang: I think the logic behind testing `reportItemCount` is the change made for #9224 in that the item count is now explicitly set to zero when writing the "none" report data during routing. I'm not 100% sure where the item count being checked here is sourced from though.

I was also having difficulty reproducing the original error. Another change made in the time since the bug was written might have changed the originally observed behavior.

prime-router/src/main/kotlin/history/SubmissionHistory.kt
```diff
@@ -513,14 +520,19 @@ class DetailedSubmissionHistory(
...
+             * been processed yet vs. a report that has been processed, but did not have any eligible receivers.
+             *
+             * The most likely scenario for that is when the item does not pass the jurisdictional filter for any of
+             * the receivers.
+             *
+             * Note: This method only works for the universal pipeline as the covid pipeline does the filtering and
+             * routing in one step.
              */
-            return Status.RECEIVED
+            if (actionsPerformed == setOf(TaskAction.receive)) {
```
mkalish, line 532: If you put a breakpoint here:
- a report going through the UP will initially have a receive action and then a route action even if there is ultimately no receivers
- the Covid pipeline will only ever have a receive action since there is no separate route step

mkalish:
Back to you @arnejduranovic

mkalish:
Follow up ticket is here #9378

arnejduranovic (approved):

PR #9233, opened 2023-04-20
8434 Extract LIVD Lookup FHIR custom function outside the conversion library

victor-chaparro:
This PR extracts the LIVD lookup FHIR custom function outside the conversion library. It also adds the ability to pass an additional set of custom FHIR functions to the conversion library. The LIVD lookup table needed to be extracted because it makes calls to the ReportStream database and can't be part of the conversion library once it's moved to its own repo.

**If you are suggesting a fix for a currently exploitable issue, please disclose the issue to the prime-reportstream team directly outside of GitHub instead of filing a PR, so we may immediately patch the affected systems before a disclosure. See [SECURITY.md/Reporting a Vulnerability](https://github.com/CDCgov/prime-reportstream/blob/master/SECURITY.md#reporting-a-vulnerability) for more information.**

Test Steps:
1. The LIVD lookup function isn't being used in our mappings currently. The best way to test this is by running unit tests and checking for code coverage.
2. The test  `test convert with context` in `FhirToHl7ConverterTests.kt` shows how this function can be used in the FHIR to HL7 transforms and shows how we can pass additional FHIR custom functions to the `FhirToHl7Converter`
3. Test the LIVD lookup function using the Prime CLI.
4. Make sure your DB environment variables point to your local postgres DB
    `export POSTGRES_URL=jdbc:postgresql://localhost:5432/prime_data_hub`
    `export POSTGRES_USER=prime`
     `export POSTGRES_PASSWORD=changeIT!`
5. Start the Prime CLI with this command: `./prime fhirpath --input-file src/testIntegration/resources/datatests/FHIR_to_HL7/sample_SR_1_20230302-0001.fhir`
6. Execute this FHIR path command: `Bundle.entry.resource.ofType(Observation)[0].livdTableLookup("Test Performed LOINC Code")
` and make sure it returns results.

## Changes
- Extracts the LIVD Lookup FHIR custom function outside the conversion library
- Adds the ability to additional FHIR custom functions to the FHIR to HL7 conversion library
- Added unit tests

### Testing
- [x] Tested locally?
- [x] Ran `./prime test` or `./gradlew testSmoke` against local Docker ReportStream container?
- [ ] (For Changes to /frontend-react/...) Ran `npm run lint:write`? 
- [x] Added tests?

### Process
- [ ] Are there licensing issues with any new dependencies introduced?
- [ ] Includes a summary of what a code reviewer should test/verify?
- [ ] Updated the release notes?
- [ ] Database changes are submitted as a separate PR?
[...]

PR #9393, opened 2023-05-04
Update tests to unmockk objects to fix inconsistent test failures

jack-h-wang:
This PR adjusts two unit tests that mockk objects to unmockk them at the conclusion of the test in order to resolve an inconsistent failure in other unit tests that use those objects.

Test Steps:
1. Run Gradle tasks `clean` and `package`.
2. Confirm `package` is successful (unit tests are run in the process; errors were previously observed in this process on M2 Mac environments)

## Changes
- Adjust two unit tests to unmockk objects at the end.

### Testing
- [x] Tested locally?
- [ ] Added tests?

## Linked Issues
Addresses a change made in #9233.


adegolier (approved):
LGTM, we all saw this pass the issues we were having.

jimmyfagan (approved):
Looks good, hopefully this helps us keep an eye out for this kind of issue in other tests too, guessing there are probably other spots where we are missing `unmockkObject` calls that happen to not negatively affect anything.

prime-router/src/test/kotlin/fhirengine/engine/CustomFhirPathFunctionTest.kt
```diff
@@ -140,5 +140,6 @@ class CustomFhirPathFunctionTest {
         assertThat(
             (result[0] as StringType).value
         ).isEqualTo(loincCode)
+        unmockkObject(LivdLookup, Metadata)
```
jalbinson, line 143: Tip: using @AfterEach is generally cleaner than cleaning up directly in your test

jalbinson (approved):

prime-router/src/test/kotlin/fhirengine/engine/CustomFhirPathFunctionTest.kt
adegolier: I didn't think it was normal to put an @AfterEach when only half your tests use a thing.

jalbinson: I assume it's a noop when the object isn't currently mocked. It looks like there is also the option to call `unmockkAll` in the `@AfterEach` block to really clean everything up. If some tests are using the unmocked versions of those objects then the current way is probably the right way to do it.

https://mockk.io/#object-mocks

jimmyfagan: Some combination of `unmockkAll` and `clearAllMocks` seems to be used in some combination of `@BeforeEach` and `@AfterEach` in some test files, but it's all inconsistent. I hope we can discuss this next eng sync and establish a best practice for this moving forward. There may be some cases where we don't want to unmock/clear everything between tests, so just need to talk through the options, and then document what we decide so we can reference that when making these kinds of decisions and hopefully get to some consistency eventually. Either way, it seems like we're on the same page that this is fine for this kind of quick fix, so all good for now 👍

adegolier:
So I'm not super sure why the change I made worked, but, since we discovered yesterday that a mockObject seems to just live on forever doing its thing if it's not unmocked, I can only assume that the other tests in the suite were using the mock(s), so when the unmock was added, they started failing. I put the mock in the beforeEach and unmock in the afterEach (as opposed to before/after all) just in case there's some weirdness with threading/concurrently running tests. The best part is that it seems to require the "every" sections as well because just putting the mock/unmock in the before/afters was not enough. It didn't work until I moved the everys up there as well.

jimmyfagan:
I think Angela's explanation makes sense, @jack-h-wang , do you want to sign off on that and get this merged in? It hasn't been affecting me, but I figure the sooner the better for those of you who are seeing this fail.

PR #9418, opened 2023-05-05
Experience/8959/replace organizations table

etanb:
This PR replaces the table component with our custom table in `/shared/Table`. This will be the table component use across ALL tables going forwards in ReportStream. We're also adding sticky header functionality. 

Been working closely with @jillian-hammer for the sticky header functionality and overall design. This is a 1:1 replacement -- any other edits across all tables will come in a later PR.

Example: 

https://github.com/CDCgov/prime-reportstream/assets/516877/c8e0e831-a006-4db9-9feb-784150520288


jillian-hammer:
thanks @etanb ! two design notes:

- can you take a look at why the headers jump in position once you start scrolling? ideally, they'd stay in place. the scrollbox looks like it has extra inner padding at the top? 
- question: is it possible to keep the header row underline sticky, too?

etanb:
After chatting with @jillian-hammer we're going to move forward _without_ the internal scrollbar model for now. Here's the latest-and-greatest:


https://github.com/CDCgov/prime-reportstream/assets/516877/1c498b64-954e-4bdf-917f-58072c681250

penny-lischer (approved):
LGTM

Issue #9464, opened 2023-05-10
Set up default filters for the ETOR topic


jimmyfagan:
### User Story
As an ETOR organization, I want to have reasonable default filters defined in the code so that I will be able to filter out obviously undesirable data without needing to define my own filters.

### Description/Use Case
In the UP, for full-elr, we have default filters that can be used when an organization doesn't define their own.

Those defaults are as follows:
<details>
  <summary>Quality Filter</summary>
  
  ```kotlin
    /**
     * Default Rules:
     *   Must have message ID, patient last name, patient first name, DOB, specimen type
     *   At least one of patient street, patient zip code, patient phone number, patient email
     *   At least one of order test date, specimen collection date/time, test result date
     */
    val qualityFilterDefault: ReportStreamFilter = listOf(
        "%messageId.exists()",
        "%patient.name.family.exists()",
        "%patient.name.given.count() > 0",
        "%patient.birthDate.exists()",
        "%specimen.type.exists()",
        "(%patient.address.line.exists() or " +
            "%patient.address.postalCode.exists() or " +
            "%patient.telecom.exists())",
        "(" +
            "(%specimen.collection.collectedPeriod.exists() or " +
            "%specimen.collection.collected.exists()" +
            ") or " +
            "%serviceRequest.occurrence.exists() or " +
            "%observation.effective.exists())"
    )
  ```
</details>

<details>
  <summary>Processing Mode Filter</summary>
  
  ```kotlin
    /**
     * Default Rule:
     *  Must have a processing mode id of 'P'
     */
    val processingModeFilterDefault: ReportStreamFilter = listOf(
        "%processingId.exists() and %processingId = 'P'"
    )
  ```
</details>

With the introduction of the `etor` topic in #8592, we have an opportunity to introduce default filter behavior for a topic other than `full-elr`, and since `etor` is a very different use case than `full-elr`, the default filters may look different.

This story will involve determining a reasonable default quality filter for the `etor` topic (perhaps just verifying the message type as a starting point, but maybe doing some research or collaboration with Flexion to decide with more detail). The processing mode filter can use the existing default for `full-elr` most likely. This story should also cover the code that determines default behavior based on topic values.

### Risks/Impacts/Considerations

Any existing documentation about filters and especially default filters should be updated.

### Acceptance Criteria

- [x] A new default quality filter for `etor` is added
- [x] The `etor` default filter doesn't affect `full-elr` receivers
[...]

Issue #8661, opened 2023-03-09
tech debt:  can't create a new Sender child every time we have a new datatype


jimduff-usds:
### User Story
As ReportStream, I should not associate datatypes with specific Sender objects.

### Description/Use Case
Currently, senders in the code are represented by a few different types. The overall parent class is called `Sender`. `Sender` is an abstract class which is extended by `FullELRSender` (used for UP) and `TopicSender` (used for the legacy pipeline). `TopicSender` has subtypes `CovidSender` and `MonkeypoxSender`.

We should avoid creating a new Sender class every time we have a new type of data or new topic - that won't scale, especially with ETOR likely needing a wide variety of data types and/or topics.

### Dev Notes:
The structure of the Sender configuration data needed is dependent on the Pipeline (Old PIpeline, New Pipeline), not on the datatype, nor on the topic.

Based on the proposal in https://github.com/CDCgov/prime-reportstream/blob/master/prime-router/docs/proposals/0020-univeral-pipeline-topics.md, `FullELRSender` will be renamed to `UniversalPipelineSender`, and `TopicSender` will be renamed to `LegacyPipelineSender` (still with subclasses `CovidSender` and `MonkeypoxSender`). Topic will still be used to determine which subclass of `Sender` to use; `full-elr` and `etor-ti` will correspond to `UniversalPipelineSender`, while `covid-19` will correspond to `CovidSender` and `monkeypox` will correspond to `MonkeypoxSender`.

See also:
#8586 ticket design for `topic` usage in the UP.
#8592 ticket for separation of `etor` from existing `full-elr` topic.

### Acceptance Criteria
- [ ] `Sender` subtypes match the proposal documented [here](https://github.com/CDCgov/prime-reportstream/blob/master/prime-router/docs/proposals/0020-univeral-pipeline-topics.md)

bishoyayoub:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/8661?planning-poker) with Zenhub @arnejduranovic @jack-h-wang @JessicaWNava @jimmyfagan @luis-pabon-tf @mkalish @victor-chaparro

Issue #8592, opened 2023-03-03
Split out universal pipeline topics from `full-elr` to allow for `etor` topic to point to UP.


jimduff-usds:
### User Story:

As ReportStream, I should stop requiring that senders and receivers using the Universal Pipeline have a topic of `full-elr` because the Universal Pipeline should support different types of healthcare data.

### Description/Use Case

Currently, the UP is tied to a 'topic' called `full-elr`.  This is inaccurate, because it handles much more than ELR (eg, ETOR).    Basically, currently we're using 'topic' as a synonym for 'which pipeline to use'. See `SettingsProvider.kt`:

```
/**
 * A submission with topic FULL_ELR will be processed using the full ELR pipeline (fhir engine), submissions
 * from a sender with topic COVID_19 will be processed using the covid-19 pipeline.
 */
enum class Topic(val json_val: String) {
    @JsonProperty("full-elr")
    FULL_ELR("full-elr"),

    @JsonProperty("covid-19")
    COVID_19("covid-19"),

    @JsonProperty("monkeypox")
    MONKEYPOX("monkeypox"),

    @JsonProperty("CsvFileTests-topic")
    CSV_TESTS("CsvFileTests-topic"),

    @JsonProperty("test")
    TEST("test"),
}
```

### Dev Notes

The proposal for this is outlined [here](https://github.com/CDCgov/prime-reportstream/blob/master/prime-router/docs/proposals/0020-universal-pipeline-topics.md). The proposal suggests retaining the `full-elr` topic, but also adding an `etor` topic. Those topics will both signify usage of the UP, however senders for one topic will only be able to send data to receivers of the same topic. That routing aspect of these topics will be handled in a separate ticket (#9495).

The proposed design does acknowledge the coupling between topics and pipelines. Even though the topic won't directly specify a pipeline, the topic will still be used to determine which pipeline is being used. This is okay, since we only test/implement a topic with a specific pipeline in mind, and shouldn't allow for topics and pipelines to be decoupled as they are currently structured.

For reference, for now, I'll leave Arnej's other original proposal. We determined that it shouldn't be necessary for now, but it still may have other benefits that make it worth doing in the future.

[...]

Issue #8586, opened 2023-03-03
Design way to constrain ETOR data to ETOR receivers only.


jimduff-usds:
### User Story:

As someone outside the ReportStream team, I need a simple, foolproof way, independent of the data itself, to configure my Senders and Receivers so my data is partitioned from everyone else's data.

Detailed descriptions of **simple**, **foolproof**, and **independent**, are below.

### Description/Use Case/Further Details

A couple analogies might be helpful
- I need something that acts a like a VPN.  My data flows in my ReportStream 'private virtual network', and your data flows in yours, and RS guarantees the wires never cross.
- Another analogy might be observer/observable or 'topics' in subscription models. My senders submit messages within a particular topic, and my receivers subscribe to those topics.   Its leak-proof - anyone not subscribed never gets my data.

Here's why it needs to be simple, foolproof, and independent:

1. **Simple**:  The Requirement is best described with an example: It is likely that  ETOR onboarding and operations will be managed by Technical Assistants ("TA") hired directly by CDC, not by people on the ReportStream team.   Those non-ReportStream teams need a very _simple foolproof_ way to stovepipe their data, to avoid confusion and errors.
2. **Foolproof**:  The implementation should default to not send -- it can't depend on a human remembering to put in an extra clause in one of the FHIR path filters.    (Although I think that solution is fine for a UP MVP).
3.  **Independent of the data itself**:  _You can easily conceive of situations where the data itself might not contain enough info to determine if it belongs within a particular topic._   Or, situations where the customers repeatedly change datatypes as they setup their system.  That is, the private network is tied to senders and receivers, and may not be in the data itself.   Topics should be an abstraction.

### Risks/Impacts/Considerations

Its possible this the current filters will become a performance issue.  ETOR may end up having thousands of receivers.   RS's powerful filter mechanism tries them all right now.   'topic', or something like it, acting basically as an "initial filter" will help you constrain the number of comparisons occurring.  Why bother trying to send full-elr data through ETOR and vice versa?

### Dev Notes:

The problem of mixing data has already happened.  The first time I submitted ETOR data, I immediately found it flowing to bizarre places, like the 'ignore' full-elr test receivers even though its not full-elr data. This is confusing, and further, it breaks easily. (See #8583 for an example - this is not a one-off problem -- filters necessarily make assumptions about the _type_ of data coming through their receivers, so this will recur)

[...]

PR #9465, opened 2023-05-10
Etor topic design

jimmyfagan:
This PR creates a proposal document for the proposed handling of partitioning of etor/elr senders/receivers.

Test Steps: n/a

## Changes
- Added a proposal document

## Linked Issues
- Addresses #8586 

## To Be Done
All of these will require some updates based on the proposed design before they are ready for grooming.
- #8661 (this will cover the refactoring of the existing `Sender` subclasses)
- #8592 (this ticket would be updated to be more specific to the splitting out of `full-elr` and `etor` topics both pointing to the Universal Pipeline)
- #9464 (this will cover default filters for the `etor` topic)


prime-router/docs/proposals/0020-univeral-pipeline-topics.md
```diff
@@ -0,0 +1,63 @@
...
+could set a reasonable starting set of quality expectations that are reasonable for ETOR requests.
+
+### Implications on internal types
+
+Currently, senders in the code are represented by a few different types. The overall parent class is called `Sender`.
+`Sender` is an abstract class which is extended by `FullELRSender` (used for UP) and `TopicSender` (used for the legacy
+pipeline). `TopicSender` has subtypes `CovidSender` and `MonkeypoxSender`.
+
+Under this proposal, `FullELRSender` will be renamed to `UniversalPipelineSender`, and `TopicSender` will be renamed
+to `LegacyPipelineSender` (still with subclasses `CovidSender` and `MonkeypoxSender`). Topic will still be used to
```
TurkeyFried, line 44: The refactoring of `TopicSender` may be better happening after all other parts of this proposal since there's no overlap with that name.

prime-router/docs/proposals/0020-univeral-pipeline-topics.md
```diff
@@ -0,0 +1,63 @@
...
+### Implications on organization settings
+
+After the changes here, the existing settings for ETOR senders and receivers will largely still hold, with the exception
+of the value of `topic`, which will be `etor` instead of `full-elr`.
+
+Since ELR senders and receivers have a topic of `full-elr` already, they will not require any changes.
+
+Monkeypox and Covid-19 are special cases initially, but eventually, those will be phased out, and condition-specific
+filtering will only occur in the Universal Pipeline by using condition filters. However, to minimize impact in the
+short-term, we will continue to support `covid-19` and `monkeypox` as topics, so those settings will not require
```
TurkeyFried, line 62: Might be fine, but wanted to bring up that if covid and monkeypox end up being moved to UP down the road, we will need to either use a transitionary topic name or change its implementation for everybody at once.

TurkeyFried:
Proposal is pretty straightforward and overall I'm very much pro-it. Had to get nitpicky to give feedback...

prime-router/docs/proposals/0020-univeral-pipeline-topics.md
jimmyfagan: Sure, a lot of that refactoring would be for clarity/readability only, so I think that would come after the introduction of an `etor` topic and any default filters that would exist for that topic.

prime-router/docs/proposals/0020-univeral-pipeline-topics.md
jimmyfagan: Yea we would probably want some kind of rollout plan based on what the current state of senders/receivers is, but especially with the covid emergency wrapped up, it might not be too disruptive to just move these topics to the UP all at once.

prime-router/docs/proposals/0020-univeral-pipeline-topics.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Goals
+
+Since ELR and ETOR have different purposes and their data is held to different expectations, we want organization
+settings to prevent the two data types (as well as future other data types) from getting mixed up between
+organizations (senders/receivers) that are intended for the other data type. The solution for this should be simple,
+foolproof, and independent as laid out here:
+
+- Simple: The Requirement is best described with an example: It is likely that ETOR onboarding and operations will be
+  managed by Technical Assistants (TA) hired directly by CDC, not by people on the ReportStream team. Those
```
arnejduranovic, line 18: We should follow-up with ETOR on this to confirm this is the plan. RS does NOT want to be doing ETOR onboarding.

prime-router/docs/proposals/0020-univeral-pipeline-topics.md
```diff
@@ -0,0 +1,63 @@
...
+  info to determine if it belongs within a particular topic. Or, situations where the customers repeatedly change
+  datatypes as they set up their system. Therefore, the information that tells us where certain data goes should be tied
+  to senders and receivers, and may not be solely determinable from the data itself. Topics should be an abstraction.
+
+## Proposal
+
+Currently, the Universal Pipeline (UP) is used for both use cases, and organization settings specify to use the UP by
+setting the `topic` field to `full-elr`. This proposal suggests retaining the `full-elr` topic, but also adding a `etor`
+topic. Those topics will both signify usage of the UP, however senders for one topic will only be able to send data to
+receivers of the same topic. Additionally, each topic may have a different set of default receivers; for instance, the
```
arnejduranovic, line 32: each topic may have a different set of default `filters`?

arnejduranovic: `Those topics will both signify usage of the UP`. Regarding this, presently topics are used to determine what pipeline is used, and I think we should work to DECOUPLE the pipeline from the topic. Have we thought about this? See this ticket for reference: https://github.com/CDCgov/prime-reportstream/issues/8592 mainly the following point: `If you decide to resurrect 'topic' and use it as originally structured in the settings, then you'll need some other technique to decide when to route data to new vs old pipelines.`

prime-router/docs/proposals/0020-univeral-pipeline-topics.md
```diff
@@ -0,0 +1,63 @@
...
+`Sender` is an abstract class which is extended by `FullELRSender` (used for UP) and `TopicSender` (used for the legacy
+pipeline). `TopicSender` has subtypes `CovidSender` and `MonkeypoxSender`.
+
+Under this proposal, `FullELRSender` will be renamed to `UniversalPipelineSender`, and `TopicSender` will be renamed
+to `LegacyPipelineSender` (still with subclasses `CovidSender` and `MonkeypoxSender`). Topic will still be used to
+determine which subclass of `Sender` to use; `full-elr` and `etor` will correspond to `UniversalPipelineSender`,
+while `covid-19` will correspond to `CovidSender` and `monkeypox` will correspond to `MonkeypoxSender`.
+
+Receivers don't have the same kind of structure as senders, so the `Receiver` class and subclasses will not have the
+same kinds of changes. Instead, during routing in the UP, the topic of the receiver will be considered so that reports
```
arnejduranovic, line 49: I like this section a lot

arnejduranovic (changes requested):

prime-router/docs/proposals/0020-univeral-pipeline-topics.md
jimmyfagan: Ok, yea I'll bring it up on a standup next week, if it's not the plan, or if it's not clear enough to speculate, I can come back and edit that bit out.

prime-router/docs/proposals/0020-univeral-pipeline-topics.md
jimmyfagan: Yea good catch, will make that update.

As for the decoupling, the reality is that it is tightly coupled, topics are implemented/tested/intended for one pipeline and should not be used for the other pipeline, so I don't see much value in decoupling the logic in this case. In the longer term, there won't really be any logic for which pipeline to use since everything will just go to the UP, but for now, we can kind of think of `covid-19` and `monkeypox` as special cases, and the default assumption logically is that the UP is used. Does that make sense? I don't have a specific issue with your proposal of separate api endpoints in that ticket, but it does feel a bit more complicated/disruptive than just having the topic enum include pipeline logic until the legacy pipeline is phased out.

jimmyfagan, line 32:
```suggestion
receivers of the same topic. Additionally, each topic may have a different set of default filters; for instance, the
```

arnejduranovic: I think this is fine for now and can get tackled separately later down the road if needed. Thank you for looking into this!

arnejduranovic (approved):

arnejduranovic (approved):

Issue #8837, opened 2023-03-24
Modify UP condition filter to OR filter statements instead ANDing


thetaurean:
### User Story:
As a receiver, I want to receive data matching any of the condition filters I have configured (rather than all of them as currently implemented) so that I can receive data I am interested in.

### Description/Use Case
The Universal Pipeline utilizes a filter on health condition (i.e. LOINC code). This filter ensures the UP only sends data to receivers that are interested. This filter, configured per receiver in `organizations.yml`, supports multiple condition filter "statements" in the configuration file.

Currently these statements are ANDed together such that all must evaluate to TRUE. It would make more sense for them to be OR'd together, so that if any evaluates to true, the data is sent.

Modify the condition filter evaluation logic such that data passes if it matches any of the configured condition filter statements.

### Risks/Impacts/Considerations
- Further documentation is needed for this and similar filters in the pipeline.

### Dev Notes:
- Original issue prompting need for condition filter: #8095
- Issue tracking the work to implement condition filter: #8315
- Example of condition filters in action: #8820 (WIP reflecting current implementation of condition filter)

Key persons: @PatriciaPerozoUSDS and @JessicaWNava 

### Acceptance Criteria
- [ ] Condition filter statements are evaluated on OR basis
- [ ] Documentation in code updated to reflect change
- [ ] Updated Docs (universal-pipeline-filter-reporting.md)

bishoyayoub:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/8837?planning-poker) with Zenhub @arnejduranovic @thetaurean @JessicaWNava @jimmyfagan @jack-h-wang @victor-chaparro

PR #9472, opened 2023-05-11
Update Condition Filter to use OR instead of AND

TurkeyFried:
This PR updates the Condition filters to pass items if any of the filters matches the report's data.

Note - This PR has attached a CURL call that I used on Postman to test. Feel free to use it or make your own test.

Test Steps:
1. In organizations.yml, update the `Ignore FULL_ELR` receiver to have the following condition filter, replacing 600-7 with whatever LOINC code shows up in your Postman test:
`conditionFilter: ["%obsPerformedCodes.intersect('600-7').exists()"]`
2. Run ReportStream, then run`./gradlew reloadTables` and `./gradlew reloadSettings` to update the settings.
3. Submit a report with data that would go to the `Ignore FULL_ELR` Receiver.
4. Using the History API, verify that the report's items were successfully delivered to `Ignore FULL_ELR`.
5. Repeat steps 1-4, but change the `conditionFilter`'s LOINC code to something that is not in the submitted report (e.g.`999-9`). The item should now be filtered out.
6. Add multiple values to `conditionFilter`, with at least one that matches the submission data, and at least one that does not matches. for example,
```
conditionFilter:
        - "%obsPerformedCodes.intersect('999-9').exists()"
        - "%obsPerformedCodes.intersect('600-7').exists()"
        - "%obsPerformedCodes.intersect('123-4').exists()"
```

## Changes
- Update logic on the FHIRRouter to check condition filters in an OR manner instead of AND.

## Linked Issues
- #8837 

[local call curl.txt](https://github.com/CDCgov/prime-reportstream/files/11466673/local.call.curl.txt)


jimmyfagan:
Oops accidentally got my tabs mixed up and updated this branch with the latest from master, should hopefully be okay, just pointing that out in case.

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
```diff
@@ -479,7 +498,9 @@ class FHIRRouter(
             }
         }
 
-        return if (exceptionFilters.isNotEmpty()) {
+        return if (singlePass && successFilters.isNotEmpty()) {
+            Pair(true, null)
+        } else if (exceptionFilters.isNotEmpty()) {
```
jimmyfagan, line 503: I'd prefer to check for exceptions first, just so we don't keep filters that cause exceptions without realizing it. It would result in the condition filter failing if we find an exception even if it does have a passing condition, but I think that'd be for the best.

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
```diff
@@ -440,22 +452,29 @@ class FHIRRouter(
...
         focusResource: Base = bundle,
+        singlePass: Boolean = false
     ): Pair<Boolean, String?> {
         // the filter needs to check all expressions passed in, or if the filter is null or empty it will return the
         // default response
         if (filter.isNullOrEmpty()) {
             return Pair(defaultResponse, "defaultResponse")
         }
-
+        val successFilters = mutableListOf<String>()
```
jimmyfagan, line 477: It seems like we're using this list as more of a boolean, it'll either be empty or it won't, but we're never looking at it's contents beyond that. I'd suggest replacing this with a boolean `passedOne` initialized to `false`, and later where you add to this list, instead set the boolean to `true`. Then when evaluating the overall result, just check for `singlePass && passedOne`.

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
```diff
@@ -479,7 +498,9 @@ class FHIRRouter(
             }
         }
 
-        return if (exceptionFilters.isNotEmpty()) {
+        return if (singlePass && successFilters.isNotEmpty()) {
+            Pair(true, null)
+        } else if (exceptionFilters.isNotEmpty()) {
             Pair(false, "(exception found) $exceptionFilters")
         } else if (reverseFilter) {
```
jimmyfagan, line 505: How does `reverseFilter` work with `singlePass`? For condition filters it's not possible for them both to be true currently, but hypothetically if that were to happen, the logic here gets kind of confusing: `reverseFilter` reverses the result, but not if we find an exception, and not if `singlePass` is true and we find a passing condition.

jimmyfagan (commented):
Left a handful of comments, some things you can probably update pretty easily, but overall I think `evaluateFilterCondition` is becoming a bit too complicated and doing too many different things depending on the parameters. I do like the general approach to just set a boolean to indicate whether to use and or or, so it might be best to just try to pair on this on Monday to see if we can figure out a clean way to break down that functionality.

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
```diff
@@ -377,7 +377,8 @@ class FHIRRouter(
                         ReportStreamFilterType.CONDITION_FILTER,
                         defaultResponse = true,
                         false,
-                        observation
+                        observation,
+                        true
```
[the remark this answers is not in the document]
TurkeyFried: Yeah that makes sense. Updated!

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
```diff
@@ -395,9 +396,18 @@ class FHIRRouter(
...
      * that specific [filterType]
+     * @param filters Filters that will be evaluated
+     * @param bundle FHIR Bundle that will be evaluated
+     * @param report Report object passed for logging purposes
+     * @param receiver Receiver of the report passed for logging purposes
+     * @param filterType Type of filter passed for logging purposes
+     * @param defaultResponse Response returned if the filter is null or empty
+     * @param reverseFilter Optional flag used to reverse the result of the filter evaluation
+     * @param focusResource Starting point for the evaluation, can be [bundle] if checking from root
+     * @param singlePass Optional flag used to allow elements that pass any of the filter expressions
```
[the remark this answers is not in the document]
TurkeyFried: Been getting conflicting feedback about documentation, both for using `@param` and for using `[paramName]`. I think this needs to be tackled at some point via an official standard for the project

prime-router/src/test/kotlin/fhirengine/engine/fhirRouterTests/RoutingTests.kt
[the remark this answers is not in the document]
TurkeyFried: Yeah that's true! Added

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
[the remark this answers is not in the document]
jimmyfagan: Sounds like an eng sync topic!

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
TurkeyFried: I was also thinking of that one. I would think that the reverse of singlePass is `if any fail, it's false`

prime-router/src/test/kotlin/fhirengine/engine/fhirRouterTests/RoutingTests.kt
```diff
@@ -249,6 +278,35 @@ class RoutingTests {
...
+        val fhirData = File("src/test/resources/fhirengine/engine/routerDefaults/qual_test_0.fhir").readText()
+        val bundle = FhirTranscoder.decode(fhirData)
+        val settings = FileSettings().loadOrganizations(oneOrganization)
+        val jurisFilter = listOf("Bundle.entry.resource.ofType(Provenance).count() > 0")
+        val qualFilter = listOf("Bundle.entry.resource.ofType(Provenance).count() > 0")
+        val routingFilter = listOf("Bundle.entry.resource.ofType(Provenance).count() > 0")
+        val procModeFilter = listOf("Bundle.entry.resource.ofType(Provenance).count() > 0")
+        // with multiple condition filters, we are looking for any of them passing
+        val conditionFilter = listOf(
+            "Bundle.entry.resource.ofType(Provenance).count() > 50", // true
```
jimmyfagan, line 292:
```suggestion
            "Bundle.entry.resource.ofType(Provenance).count() > 50", // not true
```

jimmyfagan (commented):
Thanks for the additions, it looks better! I'm still worried about some edge cases in `evaluateFilterCondition`, like when `reversed` is true or when the filter list is empty. Will try to huddle with you between meetings today, and hopefully we can work it out there!

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
TurkeyFried: I'm gonna go with the latter of the above, and then we can go into more detail while discussing the overall complexity of the method

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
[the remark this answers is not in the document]
jimmyfagan: I think these functions might be a good example of why we shouldn't be too rigid with this, it's sometimes nice to explain params in-line, and it's sometimes better to explain generally what a function does, and then break down each param individually. We can optionally still chat about this in a future eng sync, but I'm good to resolve this conversation.

jimmyfagan (commented):
We had a working session where we reworked some of this so that the evaluate condition function is not handling all combinations of "and/or" and "reversed/not-reversed", and instead we have two separate functions. I think that was a big improvement to readability and debugability, even if it is a little bit longer code overall given the separate functions. Holding off on approval til Luis checks out the code coverage of the split functions, and then will give my final review.

jimmyfagan (approved):
Might be good to add another test case for `reverseFilter=true` on the "or" path, but since that is currently an impossible case in the UP, I won't let that block my approval. If you add anything that warrants another look, let me know!

victor-chaparro (commented):
Tested it locally and conditions were not getting filtered when I updated the conditionFilter for the development organization to use the new format.
The current filter is: 
```
      conditionFilter:
        # Accept all positive flu A and B results or negative flu A and B results for cities in LA and Orange counties
        - >-
          %resource.where(code.coding.code in ('85478-6' | '76080-1' | '92141-1' | '80383-3' | '82170-2' | '85477-8' | '76078-5' | '77026-3' | '77028-9' | '77027-1' | '92142-9' | '80382-5' | '82166-0' | '82167-8' | '82169-4' | '82168-6' | '92809-3' | '92808-5' | '95423-0')).value.coding.code.intersect('260373001' | '407479009' | '407480007').exists()
          or (%resource.where(code.coding.code in ('85478-6' | '76080-1' | '92141-1' | '80383-3' | '82170-2' | '85477-8' | '76078-5' | '77026-3' | '77028-9' | '77027-1' | '92142-9' | '80382-5' | '82166-0' | '82167-8' | '82169-4' | '82168-6' | '92809-3' | '92808-5')).value.coding.code.intersect('260415000').exists() and (%patient.address.city.intersect(%LACountyCities | %OrangeCountyCities).exists() or %performer.address.city.intersect(%LACountyCities | %OrangeCountyCities).exists()))
```
After updating it to use the new format the Covid conditions don't get filtered out:
 ```
     conditionFilter:
        # Accept all positive flu A and B results or negative flu A and B results for cities in LA and Orange counties
        - "%resource.where(code.coding.code in ('85478-6' | '76080-1' | '92141-1' | '80383-3' | '82170-2' | '85477-8' | '76078-5' | '77026-3' | '77028-9' | '77027-1' | '92142-9' | '80382-5' | '82166-0' | '82167-8' | '82169-4' | '82168-6' | '92809-3' | '92808-5' | '95423-0')).value.coding.code.intersect('260373001' | '407479009' | '407480007').exists()"
        - "(%resource.where(code.coding.code in ('85478-6' | '76080-1' | '92141-1' | '80383-3' | '82170-2' | '85477-8' | '76078-5' | '77026-3' | '77028-9' | '77027-1' | '92142-9' | '80382-5' | '82166-0' | '82167-8' | '82169-4' | '82168-6' | '92809-3' | '92808-5')).value.coding.code.intersect('260415000').exists() and (%patient.address.city.intersect(%LACountyCities | %OrangeCountyCities).exists() or %performer.address.city.intersect(%LACountyCities | %OrangeCountyCities).exists()))"
```

Attached is the postman collection I used to test this.

[CA condition test.txt](https://github.com/CDCgov/prime-reportstream/files/11502555/CA.condition.test.txt)

**Comment by jimmyfagan** [comment]
Checked in a new commit that should fix what Victor saw. There's still some questions I have about that section of code (I'll leave a comment and tag Jessica I think), and we may want to add some testing that would have caught that issue.

File: prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
Comment (jimmyfagan) [inline prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt:1]: @JessicaWNava Why does `getObservationExtensions` return an empty list when `observationsToKeep.size == allObservations.size` and also when `observationsToKeep` is empty? They seem like very different scenarios that use the same behavior, which is a little bit confusing to me.

File: prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt [hunk 7]
```diff
@@ -438,27 +451,30 @@ class FHIRRouter(
...
         focusResource: Base = bundle,
     ): Pair<Boolean, String?> {
-        // the filter needs to check all expressions passed in, or if the filter is null or empty it will return the
-        // default response
         if (filter.isNullOrEmpty()) {
             return Pair(defaultResponse, "defaultResponse")
         }
-
+        check(filter.isNotEmpty())
         val failingFilters = mutableListOf<String>()
```
victor-chaparro, line 476: Is this check necessary since the if before is checking if filter is null or empty?

jimmyfagan: No it's not, that's leftover from an earlier iteration, we can remove it, good catch.

victor-chaparro (approved):
That was a quick fix! Re-tested it and I'm seeing the condition getting removed. I had written a ticket a while ago [here](https://app.zenhub.com/workspaces/platform-6182b02547c1130010f459db/issues/gh/cdcgov/prime-reportstream/8910) to add the integration test for removing conditions, but hasn't been groomed. I'll bring it up during our next grooming session.

jimmyfagan:
@TurkeyFried Arnej said that he briefly looked through this, and without a formal approval, with me and Victor already approving it, we're okay to merge.

PR #9526, opened 2023-05-18
9471 - Remove hyperlink on file download where date has expired.

### LINKED ISSUE #9471: [Daily Data] Remove hyperlink on files that have expired
## Problem statement

If a file has expired and a user clicks on the file to download, the file does not download and a console error is thrown which notifies Dev Ops.

## Acceptance criteria

- [x] disable the link in the `Daily Data` table if the expiration date is <= now
- [x] hide the `HLT(BATCH)` button in the `Report Details` page if the expiration date is <= now

## To do

- [ ] ...

penny-lischer:
This PR ...

If a file has expired and a user clicks on the file to download, the file does not download and a console error is thrown which notifies Dev Ops.
Do not allow the user to download the file when the date has expired.

Test Steps:
1. Log into RS
2. Navigate to Daily Data
3. If the date has expired in the row, the file should not be a hyperlink.

## Changes

Screenshot 2023-05-19 at 9 14 41 AM

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- Fixes #9471


jpandersen87 (changes requested):

jpandersen87 (approved):

frontend-react/src/components/Table/ColumnData.tsx
```diff
@@ -101,15 +102,29 @@ export const ColumnData = ({
...
-            <Button
-                className="font-mono-2xs line-height-alt-4"
-                type="button"
-                unstyled
-                onClick={doAction}
-            >
-                {displayValue}
-            </Button>
+            <>
+                {showActionButton() ? (
```
etanb, line 115: can `actionButtonHandler` or `actionButtonParam` ever actually be `undefined`? `actionButtonHandler` is just a simple `handleExpirationDate` function, and `actionButtonParam` is just a string that says `expires` to get the date. So we can just remove the `showActionButton()` and put `actionButtonHandler(rowData[actionButtonParam])` here?

or if it's dubious, just `actionButtonHandler?.(rowData[actionButtonParam])`

frontend-react/src/components/Table/ColumnData.tsx
```diff
@@ -101,15 +102,29 @@ export const ColumnData = ({
...
+                    <Button
+                        className="font-mono-2xs line-height-alt-4"
+                        type="button"
+                        unstyled
+                        onClick={doAction}
```
etanb, line 120: can we just have this be `() => action(rowData[param])`?

frontend-react/src/pages/deliveries/Table/DeliveriesTable.tsx
```diff
@@ -84,6 +85,9 @@ const DeliveriesTableContent: React.FC<DeliveriesTableContentProps> = ({
     const transformDate = (s: string) => {
         return new Date(s).toLocaleString();
     };
+    const handleExpirationDate = (expiresDate: string) => {
```
etanb, line 88: Just curious, why are we putting this functionality here as opposed to within `ColumnData.tsx` directly?

penny-lischer: This will allow for components to handle whether to show the action button based on a function. In this case, I need to check for expiration date but other tables may want to show/hide based off another parameter.

frontend-react/src/components/Table/ColumnData.tsx
penny-lischer: actionButtonHandler and actionButtonParam are currently only being used in the DeliveriesTable. The other tables in the RS do not use these params and therefore can be undefined.

etanb: but this exists within the `hasFeature("action")` block which should ensure that there is an action function. if that's not the case, then we should refactor what `hasFeature("action")` actually does.

frontend-react/src/components/Table/ColumnData.tsx
etanb: I believe these lines should cover that case: 

https://github.com/CDCgov/prime-reportstream/blob/6221b00d63a07953d504388373255ce90248e23b/frontend-react/src/components/Table/ColumnData.tsx#L97-L99

frontend-react/src/pages/deliveries/Table/DeliveriesTable.tsx
etanb: I think instead of passing down a chain of optional functions, and then checking if they're `undefined` or not, it would be more semantic just to have an explicit feature like `checkForExpiry` that would then implement all that logic if it's `true`.

penny-lischer:
Discussed Etans concerns and regarding existing functionality/poor design. DeliveriesTable will be replaced in the near future with an improved version which will cover the above mentioned comments.

etanb (approved):

PR #9535, opened 2023-05-19
Introduce `etor-ti` topic to Universal Pipeline

jimmyfagan:
This PR introduces a new topic for ETOR. This topic will be used for senders and receivers that are parts of our ETOR solution in collaboration with Flexion's Trusted Intermediary.

These changes do require some attention for deployment to staging/prod:
- Since messages now include a topic, any existing queued messages may lead to deserialization failures. I'll mention this to the team to strategize on our rollout, and maybe temporarily having a default setting for the topic would be enough to get through the short-term
- The existing flexion organization settings in staging will need to be updated

Test Steps:
1. Reload updated etor settings with this command `./prime multiple-settings set --input ./settings/staging/0149-etor.yml`
2. Send a lab order report using this postman collection: 
[Temp Collection.postman_collection_json.txt](https://github.com/CDCgov/prime-reportstream/files/11548178/Temp.Collection.postman_collection_json.txt)

3. Query the submission history to verify that the report is sent and the topic is properly recorded, and check the sftp folder to look for the resulting HL7V2 file.

## Changes
- Updated org settings in `0149-etor.yml` to use new topic, and to shorten the organization's schema names (see #9576)
- In `SettingsProvider.kt` added a new topic for ETOR_TI, and added a property to the enum to indicate usage of the universal pipeline.
- In several files, updated existing checks like `topic == Topic.FULL_ELR` to use `topic.isUniversalPipeline` instead.
- In order for the submission history to accurately reflect the correct topic, updated Report.kt to take a topic value
- Updated `FullELRSender` to take a topic value instead of defaulting to FULL_ELR, also adjusted deserialization of the sender
- Updated tests to match new parameter requirements (explicitly using  `full-elr` for now)
- Added a Topic Binding and Converter so jooq can retrieve a topic `VARCHAR` from the database directly as a `Topic`

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- Addresses #8592 

## To Be Done
- In the course of this ticket, I discovered a bug and opened #9576


prime-router/src/main/kotlin/Receiver.kt
```diff
@@ -236,25 +236,25 @@ open class Receiver(
      */
     fun consistencyErrorMessage(metadata: Metadata): String? {
```
jimmyfagan, line 237: This function needs some rework, currently has some dead code, not sure how/why it is necessary. Will investigate briefly, maybe open a tech debt ticket if it doesn't make sense to fix it in this PR.

prime-router/src/main/kotlin/Sender.kt
```diff
@@ -36,10 +36,12 @@ import java.time.OffsetDateTime
 @JsonTypeInfo(
     use = JsonTypeInfo.Id.NAME,
     include = JsonTypeInfo.As.EXISTING_PROPERTY,
-    property = "topic"
+    property = "topic",
+    visible = true
```
jimmyfagan, line 40: `visible = true` is necessary to allow us to deserialize the `topic` value and not _just_ use it to determine the sender type.

prime-router/src/main/kotlin/fhirengine/engine/Message.kt
```diff
@@ -58,6 +58,7 @@ data class RawSubmission(
     val blobURL: String,
     val digest: String,
     val blobSubFolderName: String,
+    val topic: String,
```
jimmyfagan, line 61: Considering making this change at least a temporary way to account for the transition period where messages in the queue might not have a topic.
```suggestion
    val topic: String = "",
```

mkalish (commented):
Still need to test locally, but main comment is around serializing/deserializing `Topic`

jimmyfagan:
Good points, I checked in a new commit to address the nits and now I'll do some testing with your other ideas to see what can be done here and/or addressed in a follow-up tech debt ticket.

prime-router/src/main/kotlin/SettingsProvider.kt
```diff
@@ -44,19 +44,28 @@ enum class CustomerStatus {
  * A submission with topic FULL_ELR will be processed using the full ELR pipeline (fhir engine), submissions
  * from a sender with topic COVID_19 will be processed using the covid-19 pipeline.
  */
-enum class Topic(val json_val: String) {
+enum class Topic(val jsonVal: String, val isUniversalPipeline: Boolean) {
```
mkalish, line 47: Maybe default this to true since we likely won't be adding more topics for the the covid pipeline?

prime-router/src/main/kotlin/fhirengine/engine/FHIREngine.kt
```diff
@@ -129,4 +130,16 @@ abstract class FHIREngine(
...
     )
+
+    /**
+     * Given a [message], returns the topic that is embedded in the message. Defaults to full-elr since that was the
+     * first topic supported in the universal pipeline.
+     */
+    fun getTopicFromMessage(message: RawSubmission): Topic {
+        return Topic.fromJsonValue(message.topic) ?: let {
+            // This shouldn't happen, but if it does, log the error and reset to full-elr
+            logger.warn("Invalid topic ${message.topic} found in message, defaulting to 'full-elr'")
```
mkalish, line 141: This ties into my comment on deserializing into `Topic` rather having to go from the string, but I would we would want to blow up here?

prime-router/src/main/kotlin/fhirengine/engine/Message.kt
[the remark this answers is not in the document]
mkalish: I think we might also want a follow on ticket that captures cleaning up after we're confident there are no messages left on the queue without a topic?

prime-router/src/main/kotlin/fhirengine/engine/FHIREngine.kt
jimmyfagan: Updated RawSubmission to use `Topic` directly, so was able to remove this function.

prime-router/src/main/kotlin/fhirengine/engine/Message.kt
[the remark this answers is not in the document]
jimmyfagan: Good call! Given our discussion of not needing to worry about the transition as much, I went with `Topic` instead of `Topic?`, let me know if that seems okay.

prime-router/src/main/kotlin/SettingsProvider.kt
```diff
@@ -44,19 +44,28 @@ enum class CustomerStatus {
...
+    CSV_TESTS("CsvFileTests-topic", false),
 
     @JsonProperty("test")
-    TEST("test"),
+    TEST("test", false),
+    ;
+
+    companion object {
+        private val jsonMap = Topic.values().associateBy(Topic::jsonVal)
+        fun fromJsonValue(jsonVal: String?) = jsonMap[jsonVal]
```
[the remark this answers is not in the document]
jimmyfagan: Was going to make a tech debt to look into this, but found a way to allow for the `jsonVal` to be used for the deserializer wiithout needing a separate `@JsonProperty`, does that look okay? Ended up keeping the `fromJsonVal` function, just added a comment to make that a bit more clear.

jimmyfagan:
@mkalish added a few more commits, now Message has a `Topic` (not `Topic?` because we should be able to make sure we don't have messages in the queue when we deploy this) instead of a `String`, also found a way to remove the `@JsonProperty` annotations from `Topic`. I think I covered all nits too, so this is ready for another pass.

mkalish (commented):
I'll pull this again and test it one more time

prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -378,7 +378,7 @@ class ActionHistory(
         reportFile.sendingOrg = source.organization
         reportFile.sendingOrgClient = source.client
         reportFile.schemaName = report.schema.name
-        reportFile.schemaTopic = report.schema.topic.json_val
+        reportFile.schemaTopic = report.schema.topic.jsonVal
```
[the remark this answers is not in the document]
jimmyfagan: This conversion wan't needed anymore since the binding allowed for the `reportFile.schemaTopic` to be a `Topic`

prime-router/src/main/kotlin/history/SubmissionHistory.kt
```diff
@@ -298,7 +301,7 @@ class DetailedSubmissionHistory(
         actionsPerformed.addAll(descendants.map { submission -> submission.actionName }.distinct())
 
         // Enforce an order on the enrichment:  process/translate, send, download
-        if (topic == Topic.FULL_ELR.json_val) {
+        if (Topic.fromJsonValue(topic)?.isUniversalPipeline == true) {
```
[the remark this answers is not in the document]
jimmyfagan: Changed the class to use a `Topic`

mkalish (commented):
Changes look good to me and everything worked locally.  I left one thought on a way to potentially clean up `TopicBinding`

prime-router/src/main/kotlin/JOOQBindings.kt
```diff
@@ -74,4 +74,51 @@ abstract class JsonBinding<T> (val klass: Class<T>) : AbstractBinding<JSONB, T>(
...
+/**
+ * Provides a converter for Topics into Strings and back using the jackson library
+ * and it's kotlin extensions.
+ */
+class TopicConverter : Converter<String, Topic> {
+    private val mapper = JacksonMapperUtilities.defaultMapper
+    override fun from(dbObject: String): Topic {
+        class TopicWrapper(val topic: Topic)
+
+        val topicWrapper = mapper.readValue("{\"topic\":\"$dbObject\"}", TopicWrapper::class.java)
```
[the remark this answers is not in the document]
jimmyfagan: Amazing, looks like this works and eliminates the need for the wrapper, thank you thank you 👏

mkalish (approved):
Nice work!

prime-router/src/main/kotlin/SettingsProvider.kt
```diff
@@ -44,19 +45,11 @@ enum class CustomerStatus {
...
-
-    @JsonProperty("monkeypox")
-    MONKEYPOX("monkeypox"),
-
-    @JsonProperty("CsvFileTests-topic")
-    CSV_TESTS("CsvFileTests-topic"),
-
-    @JsonProperty("test")
-    TEST("test"),
+enum class Topic(@JsonValue val jsonVal: String, val isUniversalPipeline: Boolean = true) {
```
arnejduranovic, line 48: Not sure where to put this comment, but with this implementation we are limiting ourselves to two pipelines... universal and not universal. Have we thought about the implications here? This design is not open-closed to new pipelines. Does it need to be? Not sure...

arnejduranovic (commented):
Just one main comment I think we need to have a discussion around... let me know what you think! Otherwise, this looks to be on par with what we discussed, great work!

prime-router/src/main/kotlin/SettingsProvider.kt
arnejduranovic: For example, what about when we start processing case data? I don't think we know for sure yet, and so it's better we err on the side of good practices, in my opinion, and not box ourselves in.

jimmyfagan: Fair question! My take is that at least in the short-term future, we can foresee two (one legacy, one universal) pipelines with the just the pre-existing topics on legacy, and at least two (potentially many more) topics on universal. So I think this setup serves us well for that short-term. I think it'd be easy enough to remove the `isUniversalPipeline` if we ever go down to just one pipeline, or if we need to extend to other pipelines, we can fairly simply replace `isUniversalPipeline` with some kind of `pipeline` variable (either on `topic` or separately from `topic` if those do get decoupled). Good to think about the implications, but I don't think this implementation locks us too hard into anything, and without knowing what that might look like in the far future, I'm not sure what we would want to do differently.

arnejduranovic (approved):

Issue #8689, opened 2023-03-13
Identify which observation has been filtered out by the condition filter


JessicaWNava:
### User Story:
As a sender, I want to be able to query the submission history and have it tell me which observations have been filtered out by the condition filter.

### Description/Use Case
When I query the submission history and get an error that observations were filtered out, that error should return which observations(s) were filtered out and, ideally, the observation id will be included in the error message. Alternatively, the code could be included

### Acceptance Criteria
- [ ] Some identifying information, preferably the id, is returned when an observation does not pass the condition filter.

bishoyayoub:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/8689?planning-poker) with Zenhub @arnejduranovic @jack-h-wang @mkalish @thetaurean @victor-chaparro

PR #9559, opened 2023-05-22
Log the element type and UUID for FHIR elements filtered out by Condition Filters

TurkeyFried:
This PR updates the logging of filter results on submitted reports. When filters are run on specific parts of the report's Bundle, the type and UUID for the filtered element is included on the log.

Note - This PR has attached a CURL call that I used on Postman to test. Feel free to use it or make your own test.

Test Steps:
1. In organizations.yml, update the `Ignore FULL_ELR` receiver to have a conditionFilter that will cause at least one `OBX` segment to be filtered out. e.g. `conditionFilter: ["%obsPerformedCodes.intersect('123-4').exists()"]`
2. Run ReportStream, then run `./gradlew reloadSettings` to update the settings.
3. Submit a report with data that would go to the `Ignore FULL_ELR_FHIR` Receiver.
4. Using the History API, verify that at least one row is inside `filteredReportRows` for `Ignore FULL_ELR_FHIR` in the `destinations` list.
5. Verify that the output on the Report Rows matches the format: `For ignore.FULL_ELR_FHIR, filter [FILTERDETAILS] filtered out item [ITEMID] at Observation/[UUID]"`, with relevant values replacing the bracketed contents in the example.

## Changes
- Update `FHIRRouter` to include the part of the item that's been filtered out.

## Linked Issues
- #8689  

[local call curl.txt](https://github.com/CDCgov/prime-reportstream/files/11534718/local.call.curl.txt)


prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
```diff
@@ -555,21 +559,32 @@ class FHIRRouter(
...
         filterName: String,
         bundle: Bundle,
         report: Report,
         receiver: Receiver,
-        filterType: ReportStreamFilterType
+        filterType: ReportStreamFilterType,
+        focusResource: Base? = null
     ) {
+        var filteredTrackingElement = bundle.identifier.value ?: ""
+        if (focusResource != null) {
```
jimmyfagan, line 578: If we instead just check for `focusResource != bundle` here, we can always pass `focusResource` through; we wouldn't need `focusResource` to be nullable or have a default value here, and we wouldn't need the if/else at L436. Works either way, just an idea that I think simplifies this a little bit.

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
```diff
@@ -555,21 +559,32 @@ class FHIRRouter(
     /**
      * Log the results of running filters (referenced by the given [filterName]) on items out of a [report] during the
      * "route" step for a [receiver], tracking the [filterType] and tying the results to a [receiver] and [bundle].
+     * @param filterName Name of evaluated filter
+     * @param bundle FHIR Bundle that was evaluated
+     * @param report Report object passed for logging purposes
+     * @param receiver Receiver of the report
+     * @param filterType Type of filter used
+     * @param focusResource Starting point for the evaluation, can add additional data to the log
```
jimmyfagan, line 567: If you take my suggestion my other comment, then I'd suggest a slight rewording here.
```suggestion
     * @param focusResource Starting point for the evaluation, logged if different than [bundle]
```

jimmyfagan (commented):
Just a small optional suggestion, otherwise looks good! Still need to test locally, will approve after that.

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
```diff
@@ -555,21 +559,32 @@ class FHIRRouter(
...
         bundle: Bundle,
         report: Report,
         receiver: Receiver,
-        filterType: ReportStreamFilterType
+        filterType: ReportStreamFilterType,
+        focusResource: Base? = null
     ) {
+        var filteredTrackingElement = bundle.identifier.value ?: ""
+        if (focusResource != null) {
+            filteredTrackingElement += " at " + focusResource.idBase
```
jimmyfagan, line 579: @JessicaWNava the result of this is a filter log message like this: `For ignore.FULL_ELR, filter [%obsPerformedCodes.intersect('123-4').exists()][] filtered out item MT_IGCAA_ORU_AAPHELR.1.6214638 at Observation/1685109567290401000.51a5d5cd-f839-4161-af20-315a2e526a1d`. Is that the kind of ID you were referring to in the description on [the ticket](https://github.com/CDCgov/prime-reportstream/issues/8689)? I'm a little wary of using that, since it only really carries meaning in the context of our internal FHIR bundle (or if the report was FHIR to begin with). Still helps if you pull down the FHIR bundle from object storage to look up that observation, but I'm wondering if instead using the LOINC code would be more useful. Any thoughts?

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
TurkeyFried: Yeah, it's a lot more readable with the changes

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
TurkeyFried: I'm fine waiting to merge until we're 100% sure of what will be the most useful log message content

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
```diff
@@ -433,7 +433,7 @@ class FHIRRouter(
             if (isDefaultFilter(filterType, filters)) "(default filter) "
             else ""
             }${failingFilterName ?: "unknown"}"
-            logFilterResults(filterToLog, bundle, report, receiver, filterType)
+            logFilterResults(filterToLog, bundle, report, receiver, filterType, focusResource)
```
jack-h-wang, line 436: Existing unit tests for `logFilterResults` are no longer passing with this pattern change.

jack-h-wang (changes requested):

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
JessicaWNava: How difficult would it be to use the observation id and loinc code? Honestly, this was a ticket Carlos had me write up. I am not sure what he had in mind, so it's really kind of a "use our best judgement" sort of thing. The goal is for the sender to be able to tell which one didn't make it through

TurkeyFried: Using the LOINC code is not too bad when we know it's an Observation. Made updates to use it. Additionally, it shows the codesystem being used. e.g. `"For ignore.FULL_ELR, filter [%obsPerformedCodes.intersect('123-4').exists()][] filtered out item MT_IGCAA_ORU_AAPHELR.1.6214638 with http://loinc.org code:19000-9",`

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
```diff
@@ -565,19 +562,23 @@ class FHIRRouter(
...
     ) {
         var filteredTrackingElement = bundle.identifier.value ?: ""
-        if (focusResource != null) {
-            filteredTrackingElement += " at " + focusResource.idBase
+        if (focusResource is Observation) {
+            // for Observation-type elements, we use the code if property when available
+            // if more elements need specific logic, consider extending the FHIR libraries
+            // instead of adding more if/else statements
+            val coding = focusResource.code.coding.firstOrNull()
+            if (coding != null)  filteredTrackingElement += " with " + coding.system + " code:" + coding.code
```
jimmyfagan, line 581: This works fine, although it takes out the idBase piece of it. I think the latest proposal was to use both the code and the id, would that still make sense?

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
```diff
@@ -565,19 +562,23 @@ class FHIRRouter(
...
         receiver: Receiver,
         filterType: ReportStreamFilterType,
-        focusResource: Base? = null
+        focusResource: Base
     ) {
         var filteredTrackingElement = bundle.identifier.value ?: ""
-        if (focusResource != null) {
-            filteredTrackingElement += " at " + focusResource.idBase
+        if (focusResource is Observation) {
+            // for Observation-type elements, we use the code if property when available
```
jimmyfagan, line 577: Typo maybe?
```suggestion
            // for Observation-type elements, we use the code property when available
```

jimmyfagan (commented):
A few hopefully quick/minor things, but this looks promising! Let me know when it's ready for another look!

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
jimmyfagan: I think it makes sense either way actually, the id is mostly only useful for us to debug using the internal fhir, so I will defer to whatever you decide.

jimmyfagan (approved):
Since my comments are all either optional or limited to comments, I'll mark this approved, if anything warrants another look, let me know, otherwise, you can let Arnej know this is ready for his pass. Thanks!

prime-router/src/test/kotlin/fhirengine/engine/fhirRouterTests/RoutingTests.kt
```diff
@@ -898,18 +906,50 @@ class RoutingTests {
...
+            bundle,
+            report,
+            receiver,
+            ReportStreamFilterType.CONDITION_FILTER,
+            defaultResponse = true,
+            reverseFilter = false,
+            focusResource = observation,
+            useOr = true
+        )
+        verify(exactly = 1) {
```
arnejduranovic, line 951: This isn't actually testing your new functionality? We need unit tests that test that different paths within `logFilterResults`

arnejduranovic (changes requested):
Please update ticket with migration plan for staging and production. You are changing the settings for the Ignore org which can potentially impact smoke tests.

prime-router/src/test/kotlin/fhirengine/engine/fhirRouterTests/RoutingTests.kt
TurkeyFried: Yeah that's true. I just added a new test to cover the new logging functionality

arnejduranovic (approved):

jack-h-wang (approved):

Issue #9378, opened 2023-05-03
bug: In the UP, a report that has been routed but not translated has a status of "NOT DELIVERED" when it should be "RECEIVED"


mkalish:
### User Story
As as user, I would like the submission API to always return the correct status.

### Description/Use Case
There is an edge case in the submission API where if a report has had the route step run but not the translate step, the API will return "NOT DELIVERING" even if ultimately items do make it to a receiver. 

### Risks/Impacts/Considerations
- This is a relatively small edge case that would likely only ever be seen if the system was under a lot of load and would fix itself as soon as the translate step is run.

### Dev Notes
- Adding a timeout can help to reliably recreate this issue
  -  Add `runBlocking { delay(10000) }` to `doWork` in FHIRTranslator.kt
- The code will likely need to be changed to refactor [this](https://github.com/CDCgov/prime-reportstream/blob/64d4044e7afd82e713e51b61cd49f7465609bd4d/prime-router/src/main/kotlin/history/SubmissionHistory.kt#L361) or to adjust how the filter logs get written to the db
  - Of note, a filter log is not written for a jurisdictional filter

[This](https://app.zenhub.com/workspaces/platform-6182b02547c1130010f459db/issues/gh/cdcgov/prime-reportstream/8451) is some previous work that was done that helped discover this bug.

### Acceptance Criteria
- [ ] The status for a report that will be routed to receivers after the route step has run but before translate is "RECEIVED"

luis-pabon-tf:
When demoing this ticket, submitting [this body](https://app.zenhub.com/files/304423150/84a00019-2191-4dbc-be90-f3d868fe773b/download) and then immediately checking the Submission History showed the `Not Received` value still appearing. Checking why this happened.

PR #9621, opened 2023-05-25
Correct Submission History to display "Received" and "Not Received" correctly

TurkeyFried:
This PR corrects a display bug where the status for Submission History displays `Not Delivering` on reports that will be delivered but do not have Destinations calculated yet.

Test Steps:
1. Submit a report with data that you know will be delivered to at least 1 Receiver
2. Immediately after receiving the `SubmissionId`, use the `History API` to check the status of the report. It should display `Received`
3. Now submit a report with data that you know will not be delivered to any Receivers
4. Immediately after receiving the `SubmissionId`, use the `History API` to check the status of the report. It should display `Not Delivering`

## Changes
- Added a new var to track the total number of items that passed filtering on a submission
- Update `calculateStatus` to use the new var in calculation
- Added a few documentation fixes

## Linked Issues
- #9378


prime-router/src/main/kotlin/history/SubmissionHistory.kt
```diff
@@ -204,6 +204,12 @@ class DetailedSubmissionHistory(
         return destinations.filter { it.itemCount != 0 }.size
     }
 
+    /**
+     * Container for the sum of items in total that have passed any filters
```
jimmyfagan, line 208: Found this wording a little confusing, maybe this instead?
```suggestion
     * Number of items in total that have passed any filters
```

prime-router/src/main/kotlin/history/SubmissionHistory.kt
```diff
@@ -529,10 +539,10 @@ class DetailedSubmissionHistory(
              * Note: This method only works for the universal pipeline as the covid pipeline does the filtering and
              * routing in one step.
              */
-            if (actionsPerformed == setOf(TaskAction.receive)) {
-                return Status.RECEIVED
+            if (actionsPerformed.contains(TaskAction.route) && itemCountAfterFiltering == 0) {
+                return Status.NOT_DELIVERING
```
jimmyfagan, line 543: Maybe this should be "waiting to deliver" if `itemCountAfterFiltering` is greater than 0?

prime-router/src/main/kotlin/history/SubmissionHistory.kt
jimmyfagan, line 1: Just curious, did you do some testing of the legacy pipeline on this? It looks like it would be okay for legacy too, but  I don't have as much experience with that, so I didn't know for sure.

jimmyfagan (commented):
Left a few comments/questions, let me know when this is ready for another look!

prime-router/src/main/kotlin/history/SubmissionHistory.kt
jimmyfagan: The acceptance criteria actually does say "received" so I'm fine with this either way!

prime-router/src/main/kotlin/history/SubmissionHistory.kt
TurkeyFried: Yeah, running both sync and async on the legacy pipeline gives the expected status with this change

prime-router/src/main/kotlin/history/SubmissionHistory.kt
```diff
@@ -417,6 +423,10 @@ class DetailedSubmissionHistory(
                 }
             }
         }
+
+        descendant.reports?.forEach {
```
arnejduranovic, line 427: NIT: Why do you need a separate loop for this, why can we not increment the counter inside the same block where we do destinations.add?

arnejduranovic (changes requested):
We should have some tests. The reason we had this bug in the first place is because we didn't have tests?

arnejduranovic (approved):

TurkeyFried:
Came up with a test case that fails with the old logic, then passes with the changes made. Added it on the latest commit

jimmyfagan (approved):

Issue #9460, opened 2023-05-10
Design: Standardize styling for a table variant


jillian-hammer:
## User story

As a ReportStream designer/engineer, I need our tables to be standardized, so that we can reduce the amount of technical & design debt and present our tables with visual consistency across the website. 

## Background & context

This work is part of the custom table component MVP. The custom component will also allow for customized future features that will not only meet our users needs, but also standardize tables across our website and make implementing new tables more efficient.

As part of this work, there are three variants of tables, this work covers the first one. 

## Open questions

-   Will the standardized style(s) still provide enough flexibility in meeting the needs of tabular data across our site?
-   How should we handle text fields with long strings?

## Working links
-   [Table Design Discovery Mural](https://app.mural.co/t/primedatainput6410/m/primedatainput6410/1677789052306/05f1e11a2448c3349735fe808833da01c6c0283b?sender=u0c3cdba5126734aceb613229)

## Acceptance criteria

-   [ ] Design spec document created for the Bordered table style

yudonoodles:
Just want to make sure that I understand what the AC correctly, not sure what a "complex" table is, there is only three types I see on the linked [mural](https://app.mural.co/t/primedatainput6410/m/primedatainput6410/1677789052306/05f1e11a2448c3349735fe808833da01c6c0283b?sender=u0c3cdba5126734aceb613229) - borderless, bordered and striped. I spoke with @jillian-hammer about this and she asked me to sync with @etanb, which I did and he wasn't sure what that's referring to either. 

Can I go ahead and update the ticket to "striped" instead of complex @jillian-hammer?

jillian-hammer:
@yudonoodles my bad, typo – you're right, I meant striped!

yudonoodles:
**Need alignment on what specs we want defined for the three tables @jillian-hammer** 
- Color values 
- Font and font-weight definition 
- Sticky header across three types
- Sorting across across three types

**Unsure about:**
Based on what's in the mural "Table Component Layers Priority"Screenshot 2023-06-02 at 3.14.01 PM.png](https://images.zenhubusercontent.com/6435ca831530a05687c6a34a/a47348db-360b-4f86-a443-bcfc9f6d0816), are we also looking to add a feature to the table where it can:
- Expand/Collapse on click to reveal secondary data? This currently is not being offered by USWDS table component. What's the specific use case for this? Where would we need this? 
[...]

PR #9624, opened 2023-05-25
Experience/8962+8963/replace tables within message tracker

etanb:
This fix replaces the Message Search and the Message Tracker tables with our shared component.

Current state demo:

https://github.com/CDCgov/prime-reportstream/assets/516877/8046db1d-6b37-4dfb-b6fb-c7fa10287e88

Most importantly, we figured out a path forwards of how to use USWDS with CSS modules.


jillian-hammer:
@etanb  yay looks great!

two notes:
- [nitpick, can be punted] can the scrollbar start BELOW the sticky header?
- the content of the cell looks like it's vertically center-aligned, can you adjust to being aligned to the top to make visual scanning easier?

etanb:
> [nitpick, can be punted] can the scrollbar start BELOW the sticky header?

I'm going to go ahead and say this would introduce some complications as a "sticky" header is out of the flow of the document and exists on top of the content it's scrolling on top of. The only way we could potentially do this is add some empty / negative space at the top that would push the content down, but that might add a blank area within the scrolling we'd want to avoid.

If you want to make a ticket, I can make a mini "spike" into this, and see how possible this is within a reasonable timebox. cc @jillian-hammer

jillian-hammer:
@etanb thank you! this looks good to me. I will hold off on any style comments as Tshering will provide those specs for you as she wraps #9460. however there is one inconsistency I'm noticing between tables that scroll horizontally & vertically versus vertically only.

To replicate:

**Horizontal & Vertical Scrolling**
Message ID `6cf1a59ad17b4f` first result:
image

**Vertical Scrolling only**
Message ID `d697b700-6b32-4bb6-954f-3dbb74f0cf99` first result:
image

**Notes**

- [ ] Vertical scrolling only table has extra white padding in the table headers (should be removed)
- [ ] Vertical scrolling table: the last column (currently `Transport Results`) is centered (should be left aligned)
- [ ] Adjust max height to 80vh (currently looks like 50vh but I may be using Inspect wrong)

jillian-hammer:
> > [nitpick, can be punted] can the scrollbar start BELOW the sticky header?
> 
> I'm going to go ahead and say this would introduce some complications as a "sticky" header is out of the flow of the document and exists on top of the content it's scrolling on top of. The only way we could potentially do this is add some empty / negative space at the top that would push the content down, but that might add a blank area within the scrolling we'd want to avoid.
> 
> If you want to make a ticket, I can make a mini "spike" into this, and see how possible this is within a reasonable timebox. cc @jillian-hammer

all good! thanks for thinking about it – not a high priority. @etanb

frontend-react/src/uswds-loader.scss
jpandersen87, line 1: lets call this global-modules.scss

jpandersen87 (approved):
A comment on a file name, but otherwise g2g.

frontend-react/src/config/endpoints/messageTracker.ts
```diff
@@ -41,6 +41,7 @@ export interface RSMessageDetail {
 }
 
 export interface MessageListResource {
+    id?: number;
```
penny-lischer, line 44: nit, would this ever be null?

penny-lischer (approved):
LGTM

frontend-react/src/config/endpoints/messageTracker.ts
etanb: This piece of code got around typing that: 

https://github.com/CDCgov/prime-reportstream/blob/master/frontend-react/src/components/Table/ColumnData.tsx#L80-L82

etanb: Going to update the MessageTrackerMockServer:

https://github.com/CDCgov/prime-reportstream/blob/fe8425f4b8897181204d47402133843c5738d688/frontend-react/src/__mocks__/MessageTrackerMockServer.ts#L28-L47

Let me know if you see any problems @penny-lischer

PR #9660, opened 2023-05-31
Experience/9212/file validator 404

etanb:
This is the simplest fix for the issue of calling `useSenderResource` with an unknown `sender` in the  `"/settings/organizations/:orgName/senders/:sender"` endpoint. By reconfiguring `useSenderResource` to accept an optional `initialData` parameter, we can make sure the endpoint doesn't error out BUT does return an empty `{}` so we can handle it properly. This isn't the _ideal_ solution, but it is by far the lowest impact. 

**Thought Process:**

- Best solution / initial idea was to `useSessionContext` and manipulate the `isUserSender` and the `isUserReceiver` booleans depending on the org a user chose.
- On a non-admin / regular login, either `isUserSender` or `isUserReceiver` is set to `true` for their entire session. This is how the app makes most of its "decisions".
- The third pathway, for RS people, is `isUserAdmin` being true which PRECLUDES either `isUserSender` or `isUserReceiver` being `true`.
- The reason why this matters is that upon calling on the API, specifically `useSenderResource` in this case, what sort of call we should make (imitating a receiver or sender).
- Next issue: for RS admins, upon choosing which org to imitate via the Org Table, we need to switch the user state to be `isUserSender` or `isUserReceiver` to be `true` / `false`.
- On a per-org level, we don't have a clean mechanism for determining if it is a sender or receiver, unless we: rejigger the API that serves the orgs OR make an API call per org determining their status. Neither of these are efficient or near-term solutions.
- Since we can't determine the receiver / sender status of the org we're imitating, we just have to handle the error at the API-being-called level.
- Current solution is lowest impact since this is all within the admin pathway only.


frontend-react/src/components/FileHandlers/FileHandlerSchemaSelectionStep.tsx
```diff
@@ -62,7 +62,7 @@ export default function FileHandlerSchemaSelectionStep({
                         - Select -
                     </option>
                     {schemaOptions.map(({ title, value }) => (
-                        <option key={value} value={value}>
+                        <option key={`key-${value}`} value={value}>
```
jpandersen87, line 65: Any particular reason for this change? Being nitpicky but `key` is purely a react technical thing so unless we're adding some randomness to it manually, dressing it up with a static string template functionally changes nothing.

frontend-react/src/network/QueryClients.ts
```diff
@@ -9,6 +9,8 @@ export const appQueryClient = new QueryClient({
             staleTime: Infinity,
             cacheTime: Infinity,
             refetchOnWindowFocus: false,
+            onError: (error: any) =>
+                console.error(`Something went wrong: ${error.message}`),
```
jpandersen87, line 13: We may want to to create a new error with this message (setting the `cause` property to the original error) and console.error that to keep an intact stack trace.

jpandersen87 (changes requested):
A blocker concern with error logging but otherwise will be g2g.

frontend-react/src/components/FileHandlers/FileHandlerSchemaSelectionStep.tsx
etanb: That is true, this is because React complains that the `key` property _must_ be unique, so without that prefix, it'll give us that warning endlessly.

jpandersen87: It stops warning with a string that doesn't change the uniqueness? Weird. I would add the index from the map iteration to the string just to be safe.

penny-lischer: Agree, key should be index of the map

frontend-react/src/hooks/UseSenderResource.ts
```diff
@@ -12,7 +12,9 @@ export type UseSenderResourceHookResult = {
     isInitialLoading: boolean;
 };
 
-export const useSenderResource = (): UseSenderResourceHookResult => {
+export const useSenderResource = (
+    initialData?: any
+): UseSenderResourceHookResult => {
     const { authorizedFetch, rsUseQuery } = useAuthorizedFetch<RSSender>();
```
penny-lischer, line 18: Just a thought, instead of calling `rsUseQuery` could we call `wrapUseQuery` and pass true to check if initialized?
Then you wouldn't need the initialData prop.

frontend-react/src/network/QueryClients.ts
etanb: Will make this change.

frontend-react/src/components/FileHandlers/FileHandlerSchemaSelectionStep.tsx
etanb: Yup, totally makes sense.

frontend-react/src/styles.scss
```diff
@@ -0,0 +1,12 @@
+// override (change some of the defaults loaded above)
+@forward "styles/uswds_settings.scss"; // _uswds_settings.scss
+@forward "./node_modules/@uswds/uswds/packages/uswds/_index.scss";
+
+// @import "~@trussworks/react-uswds/lib/uswds.css";
```
penny-lischer, line 5: Is this on purpose?

frontend-react/src/network/QueryClients.ts
```diff
@@ -9,6 +11,11 @@ export const appQueryClient = new QueryClient({
             staleTime: Infinity,
             cacheTime: Infinity,
             refetchOnWindowFocus: false,
+            onError: (error: any) => {
+                const errorString = `Something went wrong: ${error.message}`;
+                showError(errorString);
+                console.error(errorString);
+            },
```
jpandersen87, line 18:
```suggestion
            onError: (error: any) => {
                const errorString = `Something went wrong: ${error.message}`
                const e = new Error(errorString, { cause: error });
                showError(errorString);
                console.error(e);
            },
```

jpandersen87 (changes requested):

jpandersen87 (approved):

Issue #7415, opened 2022-11-18
Tech Debt: Investigate need for "CsvFileTests-topic" Topic


jimmyfagan:
### User Story:
As a RS developer, I want to confirm whether or not the "CsvFileTests-topic" Topic is necessary so I can remove it from the repo if it is not necessary.

### Description/Use Case
In the process of updating objects to use the `Topic` enum instead of hardcoded strings, we added a few enum values to allow all existing tests to continue to run. One of these was for "CsvFileTests-topic", but at the time of [PR 7384](https://github.com/CDCgov/prime-reportstream/pull/7384) we weren't confident whether or not it was necessary to keep that. This issue involves investigating that need, and if it is deemed unnecessary, removing that `Topic` value and any associated tests to clean up our code base.

### Risks/Impacts/Considerations
* If cleanup of the Topic is done, we should be sure that all tests pass and that coverage isn't affected.

### Dev Notes:
* See [this comment](https://github.com/CDCgov/prime-reportstream/pull/7384#discussion_r1026491808) for context

### Acceptance Criteria
- [ ] The necessity of "CsvFileTests-topic" has been confirmed.
- [ ] If "CsvFileTests-topic" is deemed unnecessary, all references to that topic in tests, functions, and other files has been removed.

bishoyayoub:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/7415?planning-poker) with Zenhub @arnejduranovic @jack-h-wang @mkalish @thetaurean @victor-chaparro

PR #9693, opened 2023-06-02
Remove references to CsvFileTests

luis-pabon-tf:
This PR handles some tech debt as a follow-up to PR #7384. We remove one of the enum values that is no longer needed, and references related to them.

The `CsvFileTests` topic usage was a bit of a "closed circle". The only non-test code touching it consisted of private vals, code only used by very specific tests, and an enum that has no text matches in the entire codebase for the removed value.

Upon making the changes, running various manual submissions into the pipeline and running all automated tets, I determined the removal was safe.

Test Steps:
1. Verify building and running the app works
2. Runs `test`, `testSmoke`, `testIntegration`

## Changes
- Deleted the `CsvFileTests` value from its enum
- Removed functions that exclusively tested `CsvFileTests`
- Remove schemas and test files only used with csv file tests

## Linked Issues
- #7415


jimmyfagan (commented):
These changes look fine, but I think we're losing some valuable test coverage by deleting `CsvFileTests.kt`. Instead of deleting those files referencing the `CsvFileTests` topic, can we just change the topic used by those tests to be `test` or `covid-19`? Let me know if you want to talk about it.

luis-pabon-tf:
> These changes look fine, but I think we're losing some valuable test coverage by deleting `CsvFileTests.kt`. Instead of deleting those files referencing the `CsvFileTests` topic, can we just change the topic used by those tests to be `test` or `covid-19`? Let me know if you want to talk about it.

I can do a checkup on coverage loss and move tests elsewhere while changing topics. Will do so as soon as I hit a good pausing point on my current task

jimmyfagan:
Just to clarify, I don't necessarily mean code coverage, just that with these tests removed, we might not have any remaining tests using CSV input in our smoke tests. I did look at this a little bit on Friday, so when the time comes, let me know if you'd be interested in pairing!

prime-router/src/test/kotlin/CsvFileTests.kt
```diff
@@ -108,7 +106,7 @@ class CsvFileTests {
         assertThat(settings.receivers).hasSize(2)
         assertThat(
             settings.findReceiver("federal-test.receiver")?.jurisdictionalFilter
-        ).isNotNull().hasSize(2)
+        ).isNotNull().hasSize(1)
```
jimmyfagan, line 109: Why did this change?

luis-pabon-tf: The tests on this file used the removed CSV defaults, so to keep it working I moved those defaults to be on federal test instead. The non-default juris filter for federal-test only has one criteria, which is why it's smaller.

You can see the relevant change on `test-organizations.yml`

arnejduranovic (approved):

PR #9736, opened 2023-06-07
Replace alerts with Tip Callout Component

### LINKED ISSUE #9689: Replace alerts with Tip Callout Component
## Problem statement

A new Tip Callout component was created and needs to be added into the existing pages that use it.
This ticket is to standardize that component and add it to the API Documentation Pages, File Validator, and Public Key tool.

## What you need to know

Tip callout components to update to the new standard component (alert-tip):

-  API Documentation Pages
-  File Validator
-  Public Key Tool

## Acceptance criteria

- [ ] Create a Tip Callout component that can be re-used around the site
- [ ] API Page: Replace alerts with Tip Callout component
- [ ] File Validator Tool: Replace alerts with Tip Callout component
- [ ] Public Key Tool: Replace alerts with Tip Callout component

## To do

- [ ] ...

jpandersen87:
Fixes #9689 #9988

This PR:
- Creates a new shared `Icon` component
  - Allows for using icons by their string name instead of picking specific icons from trussworks' `Icon class`
- Creates a new shared `Alert` component
  - Acts as a proxy to trussworks' Alert component with some differences
  - `validation` will always be enabled in order for its children to be prevented from being wrapped in a `p`, which could be erroneous depending on its content
  - the `usa-alert--validation` class name currently has no appreciable visible change when applied, so this is safe
  - accepts an `icon` prop to customize the alert's icon
  - the icon is positioned and sized using a copy of the same rules currently applied to the `:before` version of the icon (our component module style overrides the :before rule so as not to apply to our alert)
  - allows `tip` as `type` to apply a `usa-alert--tip` class name that styles the alert as a "Tip" component
- Updates usages of Alert in api guide to use our shared Alert with the tip prop
- Updates `FileHandler` to use our shared Alert with the tip prop instead of a one-off component
- Fixes some MDX syntax errors that created erroneous HTML


KeithNava (approved):

frontend-react/src/shared/Alert/Alert.module.scss
```diff
@@ -0,0 +1,34 @@
...
+    @include u-border("primary-lighter");
+    @include u-border(1px);
+    @include u-border("solid");
+    @include u-radius("md");
+}
+
+.usa-alert__icon {
+    @include u-width(4);
+    height: (2 * units($theme-alert-padding-y)) + units(3);
+    // padding - optical spacing value
```
etanb, line 23: nit: comment needed?

frontend-react/src/shared/Alert/AlertIcon.tsx
```diff
@@ -0,0 +1,54 @@
+import classnames from "classnames";
```
etanb, line 1: I'm thinking that just like the new `<Table />` component, we should just allow the component to be passed in directly for the `<Icon />. IE instead of:

```
<Alert
    headingLevel="h3"
    type="tip"
    icon={{ name: "Lightbulb", size: 3 }}
>
```

we would just pass in the icon component with the size to the component:

```
<Alert
    headingLevel="h3"
    type="tip"
    icon={<Icon.Lightbulb size={3} />}
>
```

Otherwise we'll have to keep maintaining state on this file with what icons we support. Also, keeps the components more stateless.

penny-lischer (approved):
LGTM

jillian-hammer (commented):
thank you @jpandersen87! a few comments:

- [x] remove the ability to decide the HeadingLevel, it should be fixed to the body text size `1rem`
- [ ] when I toggle `noIcon` in storybook to `True`, the icon doesn't disappear – is that expected?
image

jpandersen87:
> thank you @jpandersen87! a few comments:
> 
> * [ ]  remove the ability to decide the HeadingLevel, it should be fixed to the body text size `1rem`
> * [ ]  when I toggle `noIcon` in storybook to `True`, the icon doesn't disappear – is that expected?
> 
> image

@jillian-hammer 
Definitely an oversight on taking into account the noIcon option, thanks for the callout! HeadingLevel is an inherent feature of Alert, so I have created style rules to enforce `sm` (1rem) font-size to heading level when type is `tip`.

jillian-hammer (approved):
reviewed, this looks great! thank you @jpandersen87 and sorry for the delay!

Issue #9395, opened 2023-05-04
Update API Endpoint: Fetch Reports per Receiver 


TurkeyFried:
### User Story
As a ReportStream user, I want existing endpoints to gain new functionality without harming the existing integration

### Description/Use Case
The existing Delivery Reports endpoint (/waters/org/{organization}/deliveries) needs to be updated to support 

### Dev Notes
This is a perfect chance to add a version marker to the new version of the endpoint.

Expected data on the mockup is: 
- Order provider
- Performing facility
- Submitter
- Report id

### Acceptance Criteria
- [ ] Delivery Reports Endpoint supports Date filters, sorting, pagination
- [ ] Delivery Reports Endpoint includes all expected fields
- [ ] Existing endpoint is not affected by the changes

bishoyayoub:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/9395?planning-poker) with Zenhub @arnejduranovic @jack-h-wang @JessicaWNava @jimmyfagan @luis-pabon-tf @mkalish @victor-chaparro

PR #9935, opened 2023-06-14
9395: deliveries v1 api

mkalish:
This PR adds the V1 API for fetching delivered reports to receivers

Test Steps:
1. Make sure that your DB has some reports recorded that were sent to a receiver
2. Use the following to test the API
```sh
curl --location 'http://localhost:7071/api/v1/receivers/ignore.HL7/deliveries' \
--header 'Content-Type: application/json' \
--data '{
    "sort": {
        "direction": "ASC",
        "property": "number_of_tests"
    },
    "pagination": {
        "page": 1,
        "limit": 10
    },
    "filters": [
        {
            "value": "2023-05-21T00:00:00+00:00",
            "filterName": "UNTIL"
        }
    ]
}'
```
Refer to the `delivery.yml` file for the parameters that you can use with the API
3. Verify that the returned report IDs were delivered to the receiver
4. Verify that the data on provider, facility, sender and test numbers are all correct for the delivered reports (I tested this by sending data checking the results, resetting the db and sending data with different provider, facility, etc.)

## Changes
- Adds a V1 API for fetching reports delivered to receivers

### Testing
- [x] Tested locally?
- [x] Added tests?

### Process

## Linked Issues
- #9395


prime-router/src/main/kotlin/history/db/ReportGraph.kt
```diff
@@ -133,10 +136,16 @@ class ReportGraph(
...
+                        .concat(",")
+                        .concat(ITEM_LINEAGE.CHILD_INDEX)
+                        .concat(")")
+                        .concat("->(")
                         .concat(ITEM_LINEAGE.PARENT_REPORT_ID.cast(SQLDataType.VARCHAR))
                         .concat(",")
                         .concat(ITEM_LINEAGE.PARENT_INDEX)
-                        .concat(")").`as`(PATH_FIELD)
+                        .concat(")").`as`(PATH_FIELD),
+                    ITEM_LINEAGE.CHILD_REPORT_ID.`as`(STARTING_REPORT_ID_FIELD)
```
mkalish, line 148: Need to keep track of the starting report so that I can return it as part of the delivery API

mkalish:
The conflict will get resolved after the stacked PR is merged

jack-h-wang (approved):
Looks good to me! Was able to confirm returned results and pagination works as expected.

arnejduranovic (approved):
Looks good minus one small comment, but let's wait for approval from Experience team before merging

prime-router/src/main/kotlin/history/azure/DeliveryFunction.kt
```diff
@@ -109,6 +112,37 @@ class DeliveryFunction(
...
+            name = "getDeliveriesV1",
+            methods = [HttpMethod.POST],
+            authLevel = AuthorizationLevel.ANONYMOUS,
+            route = "v1/receivers/{receiverName}/deliveries"
+        ) request: HttpRequestMessage<String?>,
+        @BindingName("receiverName") receiverName: String
+    ): HttpResponseMessage {
+        val claims = AuthenticatedClaims.authenticate(request)
+        val receiver =
+            BaseEngine.settingsProviderSingleton.findReceiver(receiverName) ?: return HttpUtilities.notFoundResponse(
```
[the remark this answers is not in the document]
mkalish: Yea... I'm had been thinking the same thing, but needed to put some thought into if there was an easy way to generalize returning the error message. 

I'll try to include that in this PR

etanb:
@mkalish is it possible to include an `all` option or some equivalent in the Pagination _just in case_ we want to get all the results at once?

mkalish:
@etanb I think from a performance standpoint that would be tough.  Curious what situation you be covering with that?

etanb:
@mkalish From my understanding, a lot of times the number of results aren't significant enough to slow down rendering on the client-side, so just wanted to give us options in how we rendered the data in the future. Is it a big lift to add this? Would doing something like `limit: 99999999` effectively do this?

mkalish:
@etanb Yea, my recommendation would be to adjust the limit to the amount of the data the front end wants to display which can be arbitrarily high.

Overall though, I think we would want to have a conversation around this and use case since there are also performance implications for loading that much data in the backend.

etanb (approved):

etanb:
@mkalish is it possible to tell what the biggest possible dataset we have that would be called? the FE could handle many rows of data, so just wanted to see if this is a problem that doesn't quite exist yet. Would love to eventually have an `all` option but LGTM.

prime-router/docs/api/delivery.yml
```diff
@@ -8,6 +8,39 @@ info:
     email: reportstream@cdc.gov
   version: 0.2.0-oas3
 paths:
+  /v1/receivers/{receiverFullName}/deliveries:
```
snesm, line 11: POST parameters do not occur in the path.

This is the first API to use a versioning based path.  Is there an agreement that we will use this format moving forward?  Is there a plan on how to handle the existing API's (add a /v1/ alias)?

prime-router/docs/api/delivery.yml
```diff
@@ -8,6 +8,39 @@ info:
...
+    post:
+      description: Returns a list of reports that have been delivered to the receiver
+      security:
+        - OAuth2: [ user ]
+        - ApiKeyAuth: [ ]
```
snesm, line 16: securitySchemes is missing

```
  securitySchemes:
    OAuth2:
      $ref: 'https://raw.githubusercontent.com/CDCgov/prime-reportstream/master/prime-router/docs/api/components/security_schemes.yml#/OAuth2'
```

prime-router/docs/api/delivery.yml
```diff
@@ -8,6 +8,39 @@ info:
...
+            application/json:
+              schema:
+                $ref: '#/components/schemas/DeliveryApiSearchResults'
+        '401':
+          description: Unauthorized
+        '400':
+          description: The search request could not be parsed
+        '404':
+          description: The receiver could not be found
   /v1/receivers/{receiverFullName}/deliveries/submitters/search:
```
snesm, line 44: POST parameters do not occur in the path.

snesm (changes requested):

luis-pabon-tf:
Screwed up the diff. Gonna fix that up real quick.....

luis-pabon-tf:
Didn't mean to merge it before responding to @snesm comments. Will respond to them now!

prime-router/docs/api/delivery.yml
luis-pabon-tf: We've been using `security` and `securitySchemes` somewhat interchangeably on our API docs. I think it's worth discussing the structure of those components. 
I noticed there's different use cases for each: https://swagger.io/docs/specification/authentication/

Issue #9675, opened 2023-06-02
Set up connection to send to Trusted Intermediary using oauth v2


jimmyfagan:
### User Story
As a ReportStream developer, I need ReportStream to be able to securely connect to Flexion's Trusted Intermediary so I can send them FHIR bundles as part of the TI ETOR architecture.

### Description/Use Case
As part of the [ETOR architecture](https://lucid.app/lucidchart/623b1288-8d02-49bf-8b1a-7266fae73425/edit?invitationId=inv_7d50bada-e845-41a0-9f10-c39f657f4e04&referringApp=slack&page=0_0#), step two involves sending a FHIR bundle to Trusted Intermediary. This requires the ability to connect to the application using oauth v2. This is supported by the existing `REST` transport type within ReportStream, so this story will involve assessing how that connection can be set up, and testing to verify that we're able to connect to them properly.

### Risks/Impacts/Considerations
If for some reason the auth implementation of TI differs from that of other `REST` transport receivers that ReportStream has sent to in the past, then we may need to either rework the `REST` transport, or create a new transport type altogether.

### Dev Notes
The core functionality for this is likely already implemented in `RESTTransport.kt`, but for reference, Trusted Intermediary had to do something pretty similar to send reports to ReportStream, and that code can be referenced here: [ReportStreamLabOrderSender](https://github.com/CDCgov/trusted-intermediary/blob/story/148/authEndpoint/app/src[…]ermediary/external/reportstream/ReportStreamLabOrderSender.java) => getRsToken() => requestToken() => [JjwtEngine.java](https://github.com/CDCgov/trusted-intermediary/blob/story/148/authEndpoint/app/src[…]a/gov/hhs/cdc/trustedintermediary/external/jjwt/JjwtEngine.java) => generateToken()

Collaboration with the Engagement team on ReportStream, and the Flexion team will likely be necessary for some of the finer details of setting up the connection.

### Acceptance Criteria
- [ ] We have a reference for what a connection with TI will look like in organization settings
- [ ] We have a receiver in staging that has a REST transport connection with Flexion (whether active or inactive)
- [ ] We have verified with Flexion that we are able to send FHIR bundles to TI using their new auth method

bishoyayoub:
[...]

PR #9948, opened 2023-06-14
Add ETOR support to RESTTransport

oslynn:
This PR adds ETOR support to RESTTransport for Jimmy's latest code.  The transport allows the sender/Lab/Hospital to report Test Result to ETOR/Trusted-Intermediary.

Test Steps:
1. ./gradlew clean package
2. ./prime test

## Changes

### Testing
- [x] Tested locally?
- [ ] Added tests?

### Process

## Linked Issues
- Fixes #issue
- https://app.zenhub.com/workspaces/engagement-6166edbd409257001e09f1b2/issues/gh/cdcgov/prime-reportstream/9547

## To Be Done
- #issue

    - new endpoint A
    - new endpoint B


prime-router/src/testIntegration/kotlin/transport/RESTTransportIntegrationTests.kt
```diff
@@ -373,4 +394,19 @@ class RESTTransportIntegrationTests : TransportIntegrationTests() {
         val retryItems = mockRestTransport.send(transportType, header, reportId, null, context, actionHistory)
         assertThat(retryItems).isNotNull()
     }
+    @Test
+    fun `test flexion rest transport`() {
```
jimmyfagan, line 392: This test fails for me, not sure if I'm missing some setup or something, but I think generally integration tests should work without any setup.

prime-router/src/main/kotlin/transport/RESTTransport.kt
jimmyfagan, line 1: A lot of the logic in this file is depending on class-types instead of variables, and using hard coded logic for parameters and similar things. This is functionally okay, but I'm wondering if there's a better, more maintainable way of organizing this. For instance, maybe instead of having `UserApiKeyCredential`/`UserEtorCredential` and `getAuthTokenWithUserApiKey`/`getAuthTokenWithUserEtor`, could we just make the existing classes/functions  more flexible? Maybe `UserApiKeyCredential` has something like `var paramMap: Map<String, String>` and we build the form parameters from that map in `getAuthTokenWithUserApiKey`? Then we wouldn't need any of the etor specific classes or functions, we could just populate that map as part of the deserialization.

jimmyfagan (commented):
I have some comments, questions, and suggestions. You should also fill out the description in the PR to give more context to reviewers. Some basic test steps, a description of the changes you made, and some light context around why you made some of the changes you did would be very helpful.

prime-router/src/main/kotlin/credentials/Credential.kt
```diff
@@ -79,6 +79,17 @@ data class UserApiKeyCredential(
     val apiKey: String
 ) : Credential(), RestCredential
 
+data class UserEtorCredential(
```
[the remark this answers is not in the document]
oslynn: Yes, we can.

prime-router/src/main/kotlin/transport/RESTTransport.kt
```diff
@@ -409,6 +443,10 @@ class RESTTransport(private val httpClient: HttpClient? = null) : ITransport {
                         "hl7" -> {
                             TextContent(message.toString(Charsets.UTF_8), ContentType.Text.Plain)
                         }
+                        // Flexion
+                        "demographics" -> {
```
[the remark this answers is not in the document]
oslynn: Yes, I totally agreed with you.  I will create a ticket to do so.

prime-router/settings/staging/0153-add-etor-service-receiver.yml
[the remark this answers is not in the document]
oslynn: Ok, if that is what you want.  I will edit the existing 0149-etor.yml.  In the past when we added a new setting, other people showed me to add a new updated file and keep the old one as it is.

prime-router/src/testIntegration/kotlin/transport/RESTTransportIntegrationTests.kt
oslynn: It is fixed, now.

prime-router/src/main/kotlin/transport/RESTTransport.kt
oslynn: Yes, I will remove the UserEtorCredential class and only UserApiKeyCredential.  Adding paramMap though, I will have to update all the existing credentials in the Vault.  I will look into this as well.

prime-router/settings/staging/0153-add-etor-service-receiver.yml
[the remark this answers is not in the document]
jimmyfagan: Ah I see, yea I could see that being the right approach in some situations, but I think in our case, I'd appreciate having all the ETOR settings in that file, I've been keeping that updated with things like the `etor-ti` topic already. I could see having separate receivers for REST and for SFTP in 0149-etor.yml for now if we want to preserve the SFTP until the REST transport is more stable, but keeping it in the same file makes it easier to find and easier to keep track of what the most up-to-date settings are.

prime-router/src/testIntegration/kotlin/transport/RESTTransportIntegrationTests.kt
```diff
@@ -373,4 +379,20 @@ class RESTTransportIntegrationTests : TransportIntegrationTests() {
...
+        val header = makeHeader()
+        val mockRestTransport = spyk(RESTTransport(mockClientPostOk()))
+        every { mockRestTransport.lookupDefaultCredential(any()) }.returns(
+            UserApiKeyCredential("flexion", "123")
+        )
+        every { runBlocking { mockRestTransport.getAuthTokenWithUserEtor(any(), any(), any(), any()) } }.returns(
+            TokenInfoEtor("878a0e79-0e97-4f06-8c44-d756b74e8134", "reportStream",
+                "036231e0-5a1a-4ded-b784-2c4212e311bb")
```
jimmyfagan, line 392: What do these values correspond to? Or are they just random?

jimmyfagan (commented):
Thanks for meeting with me to talk about this more! Left some comments based on our conversation and also a few more questions. Thanks for the adjustments so far, looking pretty good!

oslynn:
I have new PR for this.

prime-router/src/main/kotlin/TransportType.kt
```diff
@@ -116,6 +116,8 @@ data class RESTTransportType
     val authTokenUrl: String,
     /** [tlsKeystore]The name for the credential manager to get the JKS used in TLS/SSL */
     val tlsKeystore: String? = null,
+    /** [parameters] The map of parameters to be sent in the message */
+    val parameters: Map<String, String>? = emptyMap(),
```
jimmyfagan, line 120: If the intention is that this is never null, the `?` should be removed so that'll be enforced by the compiler.
```suggestion
    val parameters: Map<String, String> = emptyMap(),
```

prime-router/src/main/kotlin/transport/RESTTransport.kt
```diff
@@ -252,25 +252,14 @@ class RESTTransport(private val httpClient: HttpClient? = null) : ITransport {
...
-                    BearerTokens(tokenInfo.accessToken, tokenInfo.refreshToken ?: "")
-                }
+                tokenInfo = getAuthTokenWithUserApiKey(
+                    restTransportInfo,
+                    credential,
+                    logger,
+                    tokenClient
+                )
+                // if successful, add the token returned to the token storage
+                bearerTokens =BearerTokens(tokenInfo.accessToken, tokenInfo.refreshToken ?: "")
```
jimmyfagan, line 262: Did you run `./gradlew ktlintFormat`? Do you have your precommit checks set up? I wouldn't have been able to check this in because it doesn't conform to our code format restrictions, I _think_ those checks can be installed with `.environment/githooks.sh install`.
```suggestion
                bearerTokens = BearerTokens(tokenInfo.accessToken, tokenInfo.refreshToken ?: "")
```

prime-router/src/main/kotlin/transport/RESTTransport.kt
```diff
@@ -338,20 +327,31 @@ class RESTTransport(private val httpClient: HttpClient? = null) : ITransport {
...
     ): TokenInfo {
         httpClient.use { client ->
             val tokenInfo: TokenInfo = client.submitForm(
-                restUrl,
+                restTransportInfo.authTokenUrl,
                 formParameters = Parameters.build {
-                    append("grant_type", "client_credentials")
-                    append("client_id", credential.user)
-                    append("client_secret", credential.apiKey)
+                    if (restTransportInfo.parameters!!.isEmpty()) {
```
jimmyfagan, line 339: If you change `parameters` to be non-nullable, then you can remove the `!!`
```suggestion
                    if (restTransportInfo.parameters.isEmpty()) {
```
If you don't make `parameters` non-nullable, then maybe something like `restTransportInfo.parameters?.isEmpty() != false` would be safer.

prime-router/src/main/kotlin/transport/TokenInfo.kt
jimmyfagan, line 1: This looks good, but you can delete `TokenInfoEtor` now!

prime-router/src/testIntegration/kotlin/transport/RESTTransportIntegrationTests.kt
```diff
@@ -19,13 +19,8 @@ import io.ktor.client.engine.mock.MockEngine
 import io.ktor.client.engine.mock.respond
 import io.ktor.client.request.HttpRequest
 import io.ktor.client.statement.HttpResponse
-import io.ktor.http.Headers
-import io.ktor.http.HttpMethod
-import io.ktor.http.HttpProtocolVersion
-import io.ktor.http.HttpStatusCode
-import io.ktor.http.Url
+import io.ktor.http.*
```
jimmyfagan, line 22: This is another thing that our pre-commit checks don't allow, be sure to run `./gradlew ktlintFormat` before you commit!

prime-router/src/testIntegration/kotlin/transport/RESTTransportIntegrationTests.kt
```diff
@@ -387,9 +383,8 @@ class RESTTransportIntegrationTests : TransportIntegrationTests() {
         every { mockRestTransport.lookupDefaultCredential(any()) }.returns(
             UserApiKeyCredential("flexion", "123")
         )
-        every { runBlocking { mockRestTransport.getAuthTokenWithUserEtor(any(), any(), any(), any()) } }.returns(
-            TokenInfoEtor("878a0e79-0e97-4f06-8c44-d756b74e8134", "reportStream",
-                "036231e0-5a1a-4ded-b784-2c4212e311bb")
+        every { runBlocking { mockRestTransport.getAuthTokenWithUserApiKey(any(), any(), any(), any()) } }.returns(
+                TokenInfo("MockToken", null, null, null, "bearer")
```
jimmyfagan, line 387: We can use the default null values for the middle 3 parameters
```suggestion
                TokenInfo(accessToken = "MockToken", tokenType = "bearer")
```

jimmyfagan (commented):
Thanks for making the changes, things look pretty good! Left a few comments for some minor suggestions and formatting issues, but I think we're almost ready to merge.

prime-router/src/main/kotlin/TransportType.kt
oslynn: Old REST Transports (i.e NY, OK, WA, .. etc) have null as Parameters.  I intentionally have it for backward compatibility with the old RESTransport.  Currently, only REST Transport for ETOR has parameters.

prime-router/src/main/kotlin/transport/RESTTransport.kt
oslynn: I will run the command, now.

prime-router/src/main/kotlin/transport/RESTTransport.kt
oslynn: As I explained above, parameters can be null.

prime-router/src/main/kotlin/transport/TokenInfo.kt
oslynn: Ah, I must forget to remove, it.

prime-router/src/main/kotlin/TransportType.kt
jimmyfagan: Using the default value `emptyMap()` here means that even for those old transport, the Map should be an empty map, but still not `null`. We should test that to make sure, but given that there is a non-null default value, the only way for this to be `null` would be to have it explicitly set.

prime-router/src/main/kotlin/transport/RESTTransport.kt
jimmyfagan: If `parameters` can be `null`, this check will fail due a null pointer exception. `!!` asserts that the variable is _not_ null, and will throw an exception if that is untrue. Based on the other conversation, since the default for `parameters` is `emptyMap()` I don't think this will actually be `null`, but if you think it might, then we should still use `restTransportInfo.parameters?.isEmpty() != false` to prevent execptions.

prime-router/src/main/kotlin/transport/RESTTransport.kt
oslynn: Done.

prime-router/src/main/kotlin/transport/TokenInfo.kt
oslynn: Removed.

prime-router/src/testIntegration/kotlin/transport/RESTTransportIntegrationTests.kt
oslynn: Done.  IntelliJ got me, again!

prime-router/src/main/kotlin/transport/RESTTransport.kt
```diff
@@ -350,6 +361,39 @@ class RESTTransport(private val httpClient: HttpClient? = null) : ITransport {
...
+    /**
+     * Get the OAuth token by submitting Assertion credential
+     * as OAuth 2.0 Client Credentials, used by ETOR
+     *
+     * @param restUrl The URL to post to get the OAuth token
+     * @param credential The Assertion credential
+     * @param context Really just here to get logging injected
+     * @param httpClient the HTTP client to make the call
+     */
+    suspend fun getAuthTokenWithUserEtor(
```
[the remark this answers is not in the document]
oslynn: Done

prime-router/src/testIntegration/kotlin/transport/RESTTransportIntegrationTests.kt
oslynn: Random.

prime-router/src/main/kotlin/TransportType.kt
oslynn: I set this to default as emptyMap().  If transport doesn't provide the parameters, it will default as an emptyMap().  I am sorry in the earlier response I used null instead of emptyMap().  In the code, it checks for .isEmpty() or not.  I tested a transport with parameters and without parameters and it works.

prime-router/src/main/kotlin/transport/RESTTransport.kt
oslynn: Yes, you are correct.  As I mentioned above, I used null wrongly.

prime-router/src/testIntegration/kotlin/transport/RESTTransportIntegrationTests.kt
oslynn: Ok, I will change it.  I wonder why since they are defaulting to null already in the TokenInfo object?

prime-router/src/main/kotlin/TransportType.kt
jimmyfagan: I understand, since it is not intended to ever be null, the `?` should be removed so that will be enforced by the compiler.

prime-router/src/main/kotlin/transport/RESTTransport.kt
jimmyfagan: Got it, in that case, we should make the variable non-nullable and remove the `!!` here.

prime-router/src/testIntegration/kotlin/transport/RESTTransportIntegrationTests.kt
jimmyfagan: Functionally it doesn't change anything, but I find it a lot more readable and less error-prone to use the parameter names and omit variables that use the defaults. But it's not critical.

prime-router/src/main/kotlin/transport/RESTTransport.kt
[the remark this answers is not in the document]
jimmyfagan: Is there a ticket out there for this yet?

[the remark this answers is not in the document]
oslynn: I just create ticket #10119 to move this hardcode out to the setting file instead.

prime-router/src/testIntegration/kotlin/transport/RESTTransportIntegrationTests.kt
oslynn: Done

prime-router/src/testIntegration/kotlin/transport/RESTTransportIntegrationTests.kt
```diff
@@ -373,4 +380,22 @@ class RESTTransportIntegrationTests : TransportIntegrationTests() {
...
     }
+
+    @Test
+    fun `test flexion rest transport`() {
+        val header = makeHeader()
+        val mockRestTransport = spyk(RESTTransport(mockClientPostOk()))
+        every { mockRestTransport.lookupDefaultCredential(any()) }.returns(
+            UserApiKeyCredential("flexion", "123")
+        )
+        every { runBlocking { mockRestTransport.getAuthTokenWithUserApiKey(any(), any(), any(), any()) } }.returns(
```
victor-chaparro, line 391: Is there a way we can test getAuthTokenWithUserApiKey without mocking it? RESTTransport.kt doesn't have any code coverage, because most of it is mocked. Maybe we can create an additional ticket to add that. But, I'm concerned that we don't have enough testing to make sure this doesn't break other receivers that use REST. Is there a way we can test this doesn't break that before merging?

victor-chaparro (commented):
I spent some time trying to understand this code and it looks good to me, but I'm just wondering how we can test this locally and make sure it doesn't break existing functionality.

prime-router/src/testIntegration/kotlin/transport/RESTTransportIntegrationTests.kt
oslynn: @victor-chaparro, thank you for reviewing the PR.  To verify the Transport working with ETOR/Trusted-Intermediary, I run the Trusted-Intermediary and ReportStream server locally with the setting provided in 0149-etor.yml (uncommented out the localhost).  Then, I manually send patient resource to the reportStream's localhost /api/report endpoint and check for a successful response.  However, this is not what I really want for testing the ETOR use case.  I agreed with you that the getAuthTokenWithUserApiKey needs to be tested without mock.  I will create a ticket to do so. 

For code coverage, I use IntelliJ run the test with the Coverage command and check for path coverage.

victor-chaparro: Can we also test that it works for other receivers that use REST locally?

oslynn: Yes, I can and will test with NY and OK to make certain that it works with them as well.

oslynn:
Need to conduct a test with NY and OK to make sure it is working with them as well as ETOR.

victor-chaparro (approved):
I'm good with approving this as long as we've made sure it doesn't break other receivers, and tickets for improving code coverage have been created. Thanks Ott for all the work you've put into this.

prime-router/src/testIntegration/kotlin/transport/RESTTransportIntegrationTests.kt
oslynn: Tested with NY and it work well.
image

jimmyfagan (approved):
Changes look good to me too, would just like to also confirm that we've tested this against Trusted Intermediary in staging to make sure we can send to them now that they enforce the auth tokens on the `demographics` endpoint.

oslynn:
Tested with Trusted-Intermediary staging:

jimmyfagan:
Thanks Ott, good work! I think we'll be good to close #9675 whenever this gets merged in!

oslynn:
@jimmyfagan Before I merge, I would like to add two Unit Test Cases to cover "getAuthTokenWithUserApiKey" that I added for when Transport with parameters filled and without.

prime-router/src/testIntegration/kotlin/transport/RESTTransportIntegrationTests.kt
oslynn: Test cases to cover getAuthTokenWithUserApiKey are added to RESTTransportIntegrationTest.kt.

PR #9951, opened 2023-06-14
Feature/65 add check fa for monitoring tasktable

lucasdze:
This PR is to Add check to function app for monitoring task table and look for any data in the table older than 10 min, then send a slack notification


jeremy-page (approved):
Thank you Lucas

Issue #9778, opened 2023-06-07
Bug: batched UP submissions can cause routing action attempts before convert process is committed to database


jack-h-wang:
## Describe the bug
In staging, while processing a batched submission consisting of nine items, the UP was observed to attempt to process routing before the convert step had committed information to the database, causing a foreign key violation on database insert because the parent report_id had not yet been saved which in turn caused an exception to be reported in Azure.

## Dev notes
The convert step currently sends queue messages to the routing step per bundle before iterating to the next bundle in the submission. This can cause the routing function to attempt to act on the message before the convert step has committed information to the database necessary to complete the action successfully. Queue messages to the routing step should be held until the convert step has completed.

## Logs
An image of the Azure appinsights log demonstrating the timing issue (the attempt to commit the routing action to the database is the cause of the exception): 

![Screenshot 2023-06-07 at 6 08 31 PM](https://github.com/CDCgov/prime-reportstream/assets/123022331/9761bc1d-18d7-4e50-bc93-c0839261e0a9)

The log was filtered as follows:
```
union isfuzzy=true
    availabilityResults,
    requests,
    exceptions,
    pageViews,
    traces,
    customEvents,
    dependencies
| where timestamp > datetime("2023-06-07T15:13:04.000Z") and timestamp < datetime("2023-06-07T15:14:04.000Z") and (message contains "2ee1aaab-a344-46ea-a491-ab3e186958db" or innermostMessage contains "2ee1aaab-a344-46ea-a491-ab3e186958db" or message contains "f38ca562-f02c-4cc7-ac4b-958b4e6ef544" or innermostMessage contains "f38ca562-f02c-4cc7-ac4b-958b4e6ef544") 
| order by timestamp desc
| take 100
```

PR #9964, opened 2023-06-15
9778: add route message after all conversion is done

mkalish:
This PR moves sending the messages for the routing step to after all the FHIR bundles have been included

Test Steps:
This is a bug that occurred in the azure environment, so it's likely hard to manually verify.  I would recommend sending some test reports through the UP and there is a test that specifically covers the change

## Changes
- Sends the route messages after all the FHIR bundles have been converted

### Testing
- [x] Tested locally?
- [x] Added tests?

### Process

## Linked Issues
- #9778


prime-router/src/test/kotlin/fhirengine/engine/FhirConverterTests.kt
```diff
@@ -279,4 +280,54 @@ class FhirConverterTests {
             engine.getTransformerFromSchema("src/test/resources/fhir_sender_transforms/sample_schema")
         ).isNotNull()
     }
+
+    @Test
```
jimmyfagan, line 284: If I'm reading this correctly, the test verifies that we made a report for the first bundle, and then started processing the second bundle before any queue messages were sent. I think that makes sense, it just took me a minute to follow, I think largely because the `RuntimeException` confused me. Maybe a comment to help explain that the `RuntimeException` is being used to return control at a specific point in `doWork`? Optional, might just be me, and either way, it made sense after looking at the function a bit more closely.

jimmyfagan (approved):
Looks good to me!

prime-router/src/test/kotlin/fhirengine/engine/FhirConverterTests.kt
mkalish: Adding a comment makes sense to me!  But yes, you are correct the exception is arbitrary and just makes us exit early to confirm we don't send messages.

jack-h-wang (approved):
Seems good to me, had one comment above regarding the test but seems functionally sound!

prime-router/src/test/kotlin/fhirengine/engine/FhirConverterTests.kt
```diff
@@ -279,4 +280,54 @@ class FhirConverterTests {
...
+        val bodyFormat = Report.Format.FHIR
+        val bodyUrl = "http://anyblob.com"
+
+        every { actionLogger.hasErrors() } returns false
+        every { message.downloadContent() }
+            .returns(File("src/test/resources/fhirengine/engine/bundle_multiple_bundles.fhir").readText())
+        every { Report.getFormatFromBlobURL(message.blobURL) } returns Report.Format.FHIR
+        every { BlobAccess.Companion.uploadBlob(any(), any()) } returns "test"
+        every { accessSpy.insertTask(any(), bodyFormat.toString(), bodyUrl, any()) }.returns(Unit)
+        every { actionHistory.trackCreatedReport(any(), any(), any()) }.returns(Unit) andThenThrows(RuntimeException())
```
mkalish, line 311:
```suggestion
        // Throw an exception the second time trackCreatedReport is called to exit processing early and demonstrate sendMessage is not called
        every { actionHistory.trackCreatedReport(any(), any(), any()) }.returns(Unit) andThenThrows(RuntimeException())
```

mkalish:
@arnejduranovic This is ready for review

Issue #7250, opened 2022-11-04
Support the use of lookup tables for value sets in UP Sender Transform schema


carlosfelix2:
### User Story:
As a developer, I want to be able to specify a ReportStream lookup table to be used as a value set, so data that matches the key is converted to data specified by the value of the pair.

### Description/Use Case
There are multiple lookup tables in the database that are used for normalizing values from senders.  We want to reuse these tables as well as we want to store future value sets in tables as well.

### Risks/Impacts/Considerations
* Think about when to load the tables to improve performance.  You should not load ALL the tables when only one or a few are needed.

### Dev Notes:
* Care must be taken to not introduce the concept of lookup tables and database to the FHIR to HL7 library code.

### Acceptance Criteria
* The UP Sender Transform can use a value set defined as a lookup table in the database.
* The UP Sender Transform schema specifies which columns in the lookup table to use as the key and value.

PR #10061, opened 2023-06-22
7250 load sender transform valuesets from lookup tables

jack-h-wang:
This PR makes changes to the handling of sender transform valuesets to allow extensibility and extends valuesets to support retrieving data from RS lookup tables.

Test Steps:
1. Add a sender transform that uses a lookup table to `default-sender-transform.yml`:
```
  - name: test-transform
    resource: 'Bundle.entry.resource.ofType(Patient).name'
    condition: 'Bundle.entry.resource.ofType(Patient).exists()'
    bundleProperty: '%resource.family'
    value: [ '%resource.family' ]
    valueSet:
      lookupTable:
        tableName: npi-lookup
        keyColumn: first_name
        valueColumn: last_name
```
2. Send a report to the `full-elr` pipeline that will have data that should be transformed, e.g. patient's last name appears in the list of first names in the `npi-lookup` table.
3. Check the sftp output of the report and verify the transform occurred.

## Changes
- Created `ValueSetCollection` interface, which replaces `SortedMap<String, String>` in `ConfigSchemaElement`. The interface describes function `getMappedValue(String)`, which returns the corresponding value for a given key.
- Created `inlineValueSet` class which acts as the default implementation of `ValueSetCollection`, duplicating the previous behavior (with a slight change to the config schema).
- Created `lookupValueSet` class which is an extended implementation of `ValueSetCollection` specific to RS that retrieves valueset data from a given lookup table.
- Updated `ConfigSchemaElement` and its child objects to utilize the `ValueSetCollection` interface. 
- Updated `ConfigSchemaReader` to attach the `lookupValueSet` class as a possible subtype when deserializing transform schemas.
- Updated all existing HL7 mapping files to match changes to the config schema.
- Moved logic to retrieve valueset value for a given key from `ConfigSchemaProcessor` to `ValueSetCollection`.
- Refactored schemas so that `value` and `valueSet` are nullable instead of defaulting to empty, and updated usage accordingly.

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- #7250


prime-router/src/main/kotlin/fhirengine/engine/CustomFhirPathFunctions.kt
```diff
@@ -144,4 +145,50 @@ class CustomFhirPathFunctions : FhirPathFunctions {
             mutableListOf(StringType(result))
         }
     }
+
+    /**
```
mkalish, line 149: FYI, some commented code that can get reverted

mkalish (commented):
Well done!  This is looking reat

prime-router/src/main/kotlin/fhirengine/engine/CustomFhirPathFunctions.kt
jack-h-wang: All changes to this file should have been reverted

prime-router/src/test/resources/fhir_sender_transforms/circular_schema.yml
```diff
@@ -14,9 +14,6 @@ elements:
     condition: '%resource.issued.exists()'
     bundleProperty: '%resource.issued'
     value: [ '"CO"' ]
-    valueSet:
```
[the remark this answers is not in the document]
jack-h-wang: It can go either way, but these files were built to test specific test cases unrelated to valuesets and it seemed unnecessary to me.

prime-router/src/main/kotlin/fhirengine/engine/LookupTableValueSet.kt
```diff
@@ -0,0 +1,60 @@
...
+    ValueSetMap<LookupTableValueSetConfig> {
+    private val metadata = Metadata.getInstance()
+    private var mapValues: SortedMap<String, String>? = null
+
+    /**
+     * @return a SortedMap<String, String> representation of the valueSet
+     */
+    override fun getMapValues(): SortedMap<String, String> {
+        if (mapValues == null) {
+            val lookupTable = metadata.findLookupTable(name = configData.tableName)
```
jack-h-wang, line 39: Most of this logic could live in Metadata instead...

prime-router/src/main/kotlin/fhirengine/translation/hl7/schema/ConfigSchemaReader.kt
```diff
@@ -118,7 +120,8 @@ object ConfigSchemaReader : Logging {
         inputStream: InputStream,
         schemaClass: Class<out ConfigSchema<out ConfigSchemaElement>> = ConverterSchema::class.java
     ): ConfigSchema<*> {
-        val mapper = ObjectMapper(YAMLFactory())
+        val mapper = ObjectMapper(YAMLFactory()).registerKotlinModule()
```
jack-h-wang, line 123: Some aspects of the Jackson deserialization didn't function as expected without an excessive number of annotations. Adding registerKotlinModule() resolved the issues.

prime-router/src/main/kotlin/fhirengine/translation/hl7/ValueSetMap.kt
```diff
@@ -0,0 +1,33 @@
...
+    use = JsonTypeInfo.Id.DEDUCTION,
+    include = JsonTypeInfo.As.PROPERTY,
+    defaultImpl = InlineValueSet::class
+)
+/**
+ * Interface for valueSet. Implement this interface to create a data provider
+ * that can return a SortedMap<String, String> at runtime.
+ */
+interface ValueSetMap<T> {
+    val configData: T
```
jimmyfagan, line 17: `configData` seems to be being used for very different purposes in each of the concrete implementations of this interface, and I'd have trouble saying at the interface level what the property does. Maybe we can instead remove this and store whatever is needed for each class in the class itself. For instance, in `LookupTableValueSet` each of the three required fields can be their own properties, and the `LookupTableValueSetConfig` class can be removed. In `InlineValueSet`, instead of overriding this `configData`, the Json Property `values` can go directly into a property called `values`. Thoughts? Would be happy to talk through it further or pair on this tomorrow.

prime-router/src/main/kotlin/fhirengine/translation/hl7/ValueSetMap.kt
```diff
@@ -0,0 +1,33 @@
...
+    fun getMapValues(): SortedMap<String, String>
```
jimmyfagan, line 18: Would it be useful to add `fun getMappedValue(String): String` here? The caller often (or maybe always?) just gets the map just to look up a particular value, so maybe we can simplify usage of the interface with that call.

prime-router/src/main/kotlin/fhirengine/translation/hl7/schema/ConfigSchema.kt
```diff
@@ -188,19 +190,19 @@ abstract class ConfigSchemaElement(
         when {
             !schema.isNullOrBlank() && value.isNotEmpty() ->
                 addError("Schema property cannot be used with the value property")
-            !schema.isNullOrBlank() && valueSet.isNotEmpty() ->
+            !schema.isNullOrBlank() && valueSet.getMapValues().isNotEmpty() ->
```
jimmyfagan, line 193: Might be good to adjust the default to `null` or something to simplify some of these checks. We shouldn't need to build the map to be able to determine whether or not the schema has a `valueSet`.

prime-router/src/main/kotlin/fhirengine/translation/hl7/schema/ConfigSchema.kt
```diff
@@ -188,19 +190,19 @@ abstract class ConfigSchemaElement(
...
 
         // value sets need a value to be...set
-        if (valueSet.isNotEmpty() && value.isEmpty()) {
+        if (valueSet.getMapValues().isNotEmpty() && value.isEmpty()) {
             addError("Value property is required when using a value set")
         }
 
         // value set keys and values cannot be null
-        if (valueSet.keys.any { it == null } || valueSet.values.any { it == null }) {
+        if (valueSet.getMapValues().keys.any { it == null } || valueSet.getMapValues().values.any { it == null }) {
```
jimmyfagan, line 205: This condition probably only applies to `InlineValueSet`s, if the table is empty, that wouldn't be a reason to call the schema invalid. On the other hand, if the table or columns given are invalid, that might be reason to call the schema invalid.

jimmyfagan (commented):
This works great! I did have some comments related to the design of the interface that I hope we can address before merging, let me know if you'd like to huddle on this at some point tomorrow to discuss!

JFU-NAVA-PBC (approved):
Good enhancement, although, yaml has anchor / reference mechanism that could be leveraged to achieve one yaml element references another, but that does not go beyond the containing yaml file.

This lookup table backed mechanism provides a good alternative, especially for frequently referenced large sets.

Tested locally

prime-router/src/main/kotlin/fhirengine/translation/hl7/schema/ConfigSchema.kt
jack-h-wang: `valueSet` itself is not defined as a nullable type in `ConfigSchemaElement` so `null` wouldn't be a good check, but also the default value of `valueSet` for a schema that didn't provide one is an empty `InlineValueSet`, in which case `getMapValues` is working with the SortedMap directly. The only case a `LookupTableValueSet` would be empty is if the lookup table itself is empty or doesn't contain any values for the key and value columns, and we do need to build the map in order to detect that - though the case could be made that this should throw an exception instead.

prime-router/src/main/kotlin/fhirengine/translation/hl7/schema/ConfigSchema.kt
jack-h-wang: This check isn't testing if the valueset is empty but testing whether any of the keys or values in the map are `null`, which isn't allowed. This is the case the `invalid_value_set` test schema was built for.

jimmyfagan: Ah right, my bad, disregard this one.

prime-router/src/main/kotlin/fhirengine/translation/hl7/schema/ConfigSchema.kt
jimmyfagan: I don't mind making it a nullable type at this point. I would say an empty lookup table is a valid case, so if we're trying to enforce that `schema` and `valueSet` can't be specified in the same element, this misses that edge case.

prime-router/src/main/kotlin/fhirengine/translation/hl7/ValueSetMap.kt
jack-h-wang: I removed `configData` from the interface entirely in favor of each implementation storing the deserialized object directly. I'd prefer to keep `LookupTableValueSetConfig` as that is offloading input validation to Jackson.

jimmyfagan: Ah yea that's fine, looks good, but with `configData` removed, you can also remove the generic type `T`!

prime-router/docs/features/0003-universal-pipeline-sender-transforms.md
```diff
@@ -100,12 +101,11 @@ Element properties (in order of execution):
...
-  valueSetTable.
-- `valueSetTable` (dependent on ticket [7250](https://github.com/CDCgov/prime-reportstream/issues/7250)) - configuration
-  to use a lookup table (in the database) for the value set.
-  The configuration includes the name of the lookup table to use as well as the name of
-  the columns to use as key and value. Cannot be used with schema or valueSet.
+  property to another value that matches the key. Key value pairs can be listed directly in a configuration schema 
+  under the `values` property; an interface is provided to be able to define a custom class that returns the key value 
+  pairs programmatically using a unique property name to contain its configuration values. The UP implements these
+  extended properties:
+  - `lookupTable` - provide `tableName`, `keyColumn`, `valueColumn` to retrieve key value pairs from a lookup table.
```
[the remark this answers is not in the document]
jack-h-wang: I tried rewording the documentation to be a little more clear. Some nuance about `lookupTable` being specific to RS is lost, but given this is RS documentation that explanation isn't really needed here.

prime-router/src/main/kotlin/fhirengine/translation/hl7/ValueSetMap.kt
jack-h-wang: I further removed the generic type.

prime-router/src/main/kotlin/fhirengine/translation/hl7/ValueSetMap.kt
jack-h-wang: I added this function, moving the logic out of `ConfigSchemaProcessor`.

prime-router/src/main/kotlin/fhirengine/translation/hl7/schema/ConfigSchema.kt
jack-h-wang: I don't think we have to worry about that edge case, since `valueSet` requires `value` and there is a check to prevent `schema` and `value` from being used together. I'm not necessarily against making it nullable, but I don't think it's necessary.

jimmyfagan: Thats true, but the check for `valueSet` requiring `value` has the same problem. Take this case:
```
  - name: test-transform
    resource: 'Bundle.entry.resource.ofType(Patient).name'
    condition: 'Bundle.entry.resource.ofType(Patient).exists()'
    bundleProperty: '%resource.family'
    schema: someSchema.txt
    valueSet:
      lookupTable:
        tableName: some_empty_table
        keyColumn: first_name
        valueColumn: last_name
```
That schema is invalid for two reasons, 1) `schema` and `valueSet` are both present. 2) `valueSet` is present but `value` is not. Yet I think that case would be treated as valid because it doesn't look like we have a `valueSet` as far as these checks are concerned.
So we just need some way of knowing whether or not a `valueSet` is present, and just checking the resulting map isn't enough. Defaulting to null is one way, or you could add another function to the interface for `isConfigured` or something.

jack-h-wang: I think setting a default value of null (as is the case for the majority of `ConfigSchema`) is the appropriate solution for this problem. I spent a little time on this but it's not a trivial effort to refactor the existing code and tests to handle this; I may spin this off into a separate issue since this isn't really in scope for this issue.

jimmyfagan: I would say it is in scope mostly because it's an issue that is being introduced by these changes. And it's not just in this validation code, it's also in the element `merge` function, a valueSet won't be overwritten if the overwriting set it empty. I don't think it should be too complicated, should just be changing some default parameter values and then going through some of the valueSet logic and determining when `isEmpty` is appropriate and when `== null` is appropriate, but I wouldn't have expected any big refactor. If there's something I'm missing, let me know, maybe we can do some pairing to get through it.

jack-h-wang: I took a stab at it today after my brief look at it last night. Most of the lift was in updating tests, and I happened to notice the specific tests weren't testing `valueSet` when they should have been.

prime-router/docs/features/0003-universal-pipeline-sender-transforms.md
```diff
@@ -98,15 +98,17 @@ Element properties (in order of execution):
...
-- `valueSet` - a list of key value pairs used to convert the value generated by the value
-  property to another value that matches the key. Key value pairs can be listed directly in a configuration schema 
-  under the `values` property; an interface is provided to be able to define a custom class that returns the key value 
-  pairs programmatically using a unique property name to contain its configuration values. The UP implements these
-  extended properties:
+  assigned to the property specified in the `bundleProperty` element property. The first expression to
+  have a value wins. This allows you to set defaults at the end of the list. Cannot be used with `schema`.
+- `valueSet` - a list of key value pairs used to convert the value generated by the value property (the key)
+  to another value that matches the key. If a `valueSet` is defined and a match is not found, the element in 
+  `bundleProperty` will not be transformed. Can only be used with `value`. The property name within `valueSet`
```
jimmyfagan, line 105: This is great 👍

jimmyfagan (approved):
Looks good, thanks for dealing with my comments!

prime-router/src/main/kotlin/fhirengine/engine/LookupTableValueSet.kt
```diff
@@ -0,0 +1,69 @@
...
+)
+
+/**
+ * Implementation of [ValueSetCollection] to allow valueSet to be retrieved from a lookup table.
+ * Provide [LookupTableValueSetConfig] to configure the lookup table source.
+ */
+class LookupTableValueSet
+(@JsonProperty("lookupTable") private val configData: LookupTableValueSetConfig) : ValueSetCollection {
+    private val metadata = Metadata.getInstance()
+    private var mapVal: SortedMap<String, String>? = null
```
mkalish, line 31: Nit: this  could also be
```suggestion
    private val mapVal: SortedMap<String, String> by lazy {
      buildMap()
    }
```

prime-router/src/main/kotlin/fhirengine/translation/hl7/ConfigSchemaProcessor.kt
```diff
@@ -40,10 +40,8 @@ abstract class ConfigSchemaProcessor : Logging {
         }
 
         // when valueSet is available, use the matching value else just pass the value as is
-        // does a lowerCase comparison
-        if (retVal.isNotBlank() && element.valueSet.isNotEmpty()) {
-            val lowerSet = element.valueSet.mapKeys { it.key.lowercase() }
-            retVal = lowerSet.getOrDefault(retVal.lowercase(), retVal)
+        if (retVal.isNotBlank() && element.valueSet != null) {
+            retVal = element.valueSet?.getMappedValue(retVal) ?: retVal
```
mkalish, line 44: Why is this `?` required since we check for null?

prime-router/src/main/kotlin/fhirengine/engine/LookupTableValueSet.kt
```diff
@@ -0,0 +1,69 @@
...
+        return lowerSet[keyValue.lowercase()]
+    }
+
+    override fun isNotEmpty(): Boolean {
+        return toSortedMap().isNotEmpty()
+    }
+
+    private fun buildMap() {
+        val lookupTable = metadata.findLookupTable(name = configData.tableName)
+            ?: throw SchemaException("Specified lookup table not found")
```
mkalish, line 51: Nit: This could potentially get moved into the constructor rather than waiting until this actually gets used

prime-router/src/main/kotlin/fhirengine/translation/hl7/ConfigSchemaProcessor.kt
```diff
@@ -73,10 +71,8 @@ abstract class ConfigSchemaProcessor : Logging {
         }
 
         // when valueSet is available, return mapped value or null if match isn't found
-        // does a lowerCase comparison
-        if (retVal != null && element.valueSet.isNotEmpty()) {
-            val lowerSet = element.valueSet.mapKeys { it.key.lowercase() }
-            val valStr = lowerSet[retVal?.primitiveValue()?.lowercase() ?: ""]
+        if (retVal != null && element.valueSet != null) {
+            val valStr = element.valueSet?.getMappedValue(retVal?.primitiveValue() ?: "")
```
mkalish, line 75: Same question here

prime-router/src/main/kotlin/fhirengine/translation/hl7/FhirToHl7Converter.kt
```diff
@@ -164,7 +164,7 @@ class FhirToHl7Converter(
                     }
 
                     // A value
-                    element.value.isNotEmpty() && element.hl7Spec.isNotEmpty() -> {
+                    !element.value.isNullOrEmpty() && element.hl7Spec.isNotEmpty() -> {
```
mkalish, line 167: Maybe this should go in a follow up, but I think the water is still a little muddy here on `null` vs an empty map.  The new state going forward is that every where now needs to check for both?

Fine by be if you drop this in the tech debt ticket.

I would think that we would want a new test though that asserts this branch when its null and when its an empty map

prime-router/src/main/kotlin/fhirengine/translation/hl7/FhirTransformer.kt
```diff
@@ -124,7 +124,7 @@ class FhirTransformer(
                     }
 
                     // A value
-                    element.value.isNotEmpty() -> {
+                    !element.value.isNullOrEmpty() -> {
```
mkalish, line 127: Same thought here, but at a minimum I think we need a test?

prime-router/src/main/kotlin/fhirengine/translation/hl7/ValueSetCollection.kt
```diff
@@ -0,0 +1,49 @@
+package gov.cdc.prime.router.fhirengine.translation.hl7
+
+import com.fasterxml.jackson.annotation.JsonProperty
+import com.fasterxml.jackson.annotation.JsonTypeInfo
+import java.util.SortedMap
+
+@JsonTypeInfo(
+    use = JsonTypeInfo.Id.DEDUCTION,
```
mkalish, line 8: 🔥

prime-router/src/main/kotlin/fhirengine/translation/hl7/schema/ConfigSchemaReader.kt
mkalish: Just as a reference moving forward there is https://github.com/CDCgov/prime-reportstream/blob/bc40d2e04d422eb8519786ae4d0617524b67f213/prime-router/src/main/kotlin/common/JacksonMapperUtilities.kt#L56 that has a mapper set up with all the goodies.

Might be worth switching it out here or adding a similar utility class that lives in this package

prime-router/src/test/kotlin/fhirengine/engine/LookupTableValueSetTests.kt
```diff
@@ -0,0 +1,113 @@
...
+        val testTable = Table.create(
+            "lookuptable",
+            StringColumn.create("key", "abc123", "def456"),
+            StringColumn.create("value", "ghi789", "")
+        )
+        val testLookupTable = LookupTable(name = "lookuptable", table = testTable)
+
+        mockkConstructor(Metadata::class)
+        every { anyConstructed<Metadata>().findLookupTable("lookuptable") } returns testLookupTable
+        mockkObject(Metadata)
```
mkalish, line 38: I think you can drop this and line 39

prime-router/src/test/kotlin/fhirengine/engine/LookupTableValueSetTests.kt
```diff
@@ -0,0 +1,113 @@
...
+        val testTable = Table.create(
+            "table",
+            StringColumn.create("key", "abc123", "def456"),
+            StringColumn.create("value", "ghi789", "")
+        )
+        val testLookupTable = LookupTable(name = "table", table = testTable)
+
+        mockkConstructor(Metadata::class)
+        every { anyConstructed<Metadata>().findLookupTable("table") } returns testLookupTable
+        mockkObject(Metadata)
```
mkalish, line 68: Same here

prime-router/src/test/resources/fhir_sender_transforms/circular_schema.yml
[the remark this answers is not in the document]
mkalish: Works for me!

mkalish (changes requested):
Great stuff!  This is some really complicated work and the changes are great.

I had a couple of small things, but the overarching comment is making sure that the difference between `null` vs `emptyMap` is clear and making sure that we have coverage everywhere for both cases

prime-router/src/main/kotlin/fhirengine/translation/hl7/ConfigSchemaProcessor.kt
jimmyfagan: Since it's a `var` and not a `val`, it could theoretically (though nearly impossibly) be non-null when we check for it in L43, but then null when it is used in L44. So the compiler enforces this, we could either use `!!` to tell the compiler it's okay, or we could get rid of the check for null in L43. When we were pairing, we decided that it reads to match the comment in L42 more easily if we leave it like this, and using `?` instead of `!!` is still safer. On another look, I think we should just get rid of the null check in L43 and maybe reword the comment.

mkalish: Ah right.  I think in this particular case it could only be null if a line was added to set it to `null` so `!!` is safe, but yes I agree that just removing the check might make the most sense since it's redundant with the `?` operator

prime-router/src/main/kotlin/fhirengine/translation/hl7/ConfigSchemaProcessor.kt
jack-h-wang: This one's a little more difficult to unpack, as the logic differs from the other use of `valueSet` above in that it returns `null` if `valueSet` is present but has no match, but if `valueSet` is not present than it returns `retVal`. The difference is down to the two methods having different consumers and use cases, though perhaps the behavior of both should be explored further.

prime-router/src/main/kotlin/fhirengine/translation/hl7/FhirToHl7Converter.kt
jack-h-wang: Worth pointing out this is `value` and not `valueSet`, but agree there's some room to explore the handling of `value` further. We could either explore the open questions from this ticket in a new ticket, or perhaps add it to #10531.

mkalish: Right, but we changed the default value for it?  Fine deferring thinking through this more, but I think the change warrants a new test

prime-router/src/main/kotlin/fhirengine/translation/hl7/ConfigSchemaProcessor.kt
mkalish: I think though this is the same situation though where the the null check on line 74 isn't required

jack-h-wang: It's materially different because `getMappedValue` also returns `null` if there is no match within `valueSet` for a given key. This boils down to the following possible results for `getValue`:

1. No match for the `value` parameter is found, so the state of `valueSet` is moot; return `null`.
2. There is a match for `value` but `valueSet` is `null`; return matching `value` string.
3. There is a match for `value`, `valueSet` exists, but `getMappedValue` returns null; return `null`.
4. There is a match for `value`, `valueSet` exists, and `getMappedValue` returns a value; return matching `valueSet` string.

Removing the separate null check for `valueSet` would conflate 2 and 3.

prime-router/src/test/kotlin/fhirengine/engine/LookupTableValueSetTests.kt
jack-h-wang: The test should function either way but I wanted to short circuit this so that the test could run quicker. `Metadata` has its own suite of tests and we don't need it to go to the database to load lookup tables when we won't be using the actual lookup table information.

prime-router/src/main/kotlin/fhirengine/engine/LookupTableValueSet.kt
jack-h-wang: I moved this all into the constructor now, thanks for the suggestion!

mkalish (approved):
LGTM!  Just add those few tests before merging

Great stuff!

PR #10067, opened 2023-06-22
BUG: Custom environment variables not loaded

### LINKED ISSUE #10066: BUG: Custom environment variables not loaded
Vite uses dotenv for .env loading which by default does not support custom environments outside of production/development. Configure dotenv to use the NODE_ENV env variable for this.

jpandersen87:
Fixes #10066

This PR forces dotenv to use custom environments for .env files via NODE_ENV.


frontend-react/vite.config.ts
```diff
@@ -8,6 +8,10 @@ import remarkGfm from "remark-gfm";
 import rehypeSlug from "rehype-slug";
 import remarkToc from "remark-toc";
 import { checker } from "vite-plugin-checker";
+import dotenv from "dotenv";
+
+// Force usage of custom environment files
+dotenv.config({ path: `.env.${process.env.NODE_ENV}` });
```
KeithNava, line 14: Interesting 🤔 ... I thought this would be the default behavior

jpandersen87: I thought so too! It seems it's just prod/dev by default though, so you gotta force it.

PR #10068, opened 2023-06-22
Updated the convertDateToAge function to take an another optional par…

JessicaWNava:
…am that specifies the date to compare it to

This PR adds another optional function to the convertDateToAge function that allows you to specify which date to compare it to. This will make the zero day tests less brittle and allow for greater functionality.

Test Steps:
1. Make sure the unit tests pass
2. In your terminal run something like ./prime fhirpath -i src/testIntegration/resources/datatests/HL7_to_FHIR/bundle1.txt on a bundle that has a birthdate then run the function on the birth date ex. Bundle.entry.resource.ofType(Patient).birthDate.convertDateToAge() Make sure it returns the years or months or days you expect. Pass it params for just the unit of time, just the date to compare it to as well as those params in any order.

## Changes
- Updated the convertDateToAge function to take an another optional param
- Updated and added tests

### Testing
- [ ] Tested locally?
- [x] Added tests?

### Process

## Linked Issues
- Fixes #issue

## To Be Done
- #issue

    - new endpoint A
    - new endpoint B


prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/CustomFHIRFunctions.kt
```diff
@@ -203,21 +203,26 @@ object CustomFHIRFunctions : FhirPathFunctions {
         val date = focus.first()
```
jimmyfagan, line 203: Probably ought to have some description of the order and type of parameters in the function comment.

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/CustomFHIRFunctions.kt
```diff
@@ -121,7 +121,7 @@ object CustomFHIRFunctions : FhirPathFunctions {
                     "returns the number of years, months if less than a year, weeks if less" +
                         " than a month, and days if less than a week that a person is old for a date resource",
```
jimmyfagan, line 122: Not sure if it'd be too verbose, or if we just want to keep the parameters hidden for now, but we could consider adding to the description to describe the parameters.

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/CustomFHIRFunctions.kt
```diff
@@ -248,15 +253,37 @@ object CustomFHIRFunctions : FhirPathFunctions {
...
+        } else {
+            pair
         }
     }
 
-    private fun calculateAgeWithAssumption(date: DateType): MutableList<Base> {
+    private fun calculateAgeWithAssumption(date: DateType, comparisonDate: LocalDate): MutableList<Base> {
         val period = Period.between(
             // have to do plus one for the month because it is expecting 1 based and we get zero based
             LocalDate.of(date.year, date.month + 1, date.day),
```
jimmyfagan, line 285: We do this `LocalDate.of(x.year, x.month + 1, x.day)` a lot in this file, it might be helpful to extract that into a one-line helper to make sure the `+ 1` is never missed on the month, and to only have to explain that in one place.

prime-router/src/test/kotlin/fhirengine/engine/FhirConverterTests.kt
jimmyfagan, line 1: Was this maybe from the other PR?

prime-router/src/test/kotlin/fhirengine/translation/hl7/utils/CustomFHIRFunctionsTests.kt
```diff
@@ -662,12 +662,52 @@ class CustomFHIRFunctionsTests {
...
+        )
+
+        val age = ageList[0]
+        assertThat(age is Age).isEqualTo(true)
+        if (age is Age) {
+            assertThat(age.unit).isEqualTo("day")
+            assertThat(age.value.toInt()).isEqualTo(0)
+            assertThat(age.code).isEqualTo("d")
+        }
+    }
```
jimmyfagan, line 701: These two tests are super similar, and I think could use some tiny edits:

- The calendar seems unnecessary since `DateType(date)` seems to work
- `randomDate` implies randomness in the test, I'd maybe recommend `someDate` to avoid misunderstandings
- Since the result should be the same for each case, it might make sense to combine these into one test

Took a stab at how that might look:

```suggestion
    @Test
    fun `test convertDateToAge 0 days`() {
        val someDate = Date(1687464192857)

        val results = listOf(
            // Explicitly look for "day"
            CustomFHIRFunctions.convertDateToAge(
                mutableListOf(DateType(someDate)),
                mutableListOf(mutableListOf(DateType(someDate), StringType("day")))
            ),
            // Get default response without specifying type
            CustomFHIRFunctions.convertDateToAge(
                mutableListOf(DateType(someDate)),
                mutableListOf(mutableListOf(DateType(someDate)))
            )
        )

        results.forEach { ageList ->
            val age = ageList[0]
            assertThat(age is Age).isEqualTo(true)
            if (age is Age) {
                assertThat(age.unit).isEqualTo("day")
                assertThat(age.value.toInt()).isEqualTo(0)
                assertThat(age.code).isEqualTo("d")
            }
        }
    }
```

prime-router/src/test/kotlin/fhirengine/translation/hl7/utils/CustomFHIRFunctionsTests.kt
```diff
@@ -678,19 +718,19 @@ class CustomFHIRFunctionsTests {
...
             assertThat(age.unit).isEqualTo("day")
-            assertThat(age.value.toInt()).isEqualTo(0)
+            assertThat(age.value.toInt()).isEqualTo(4)
             assertThat(age.code).isEqualTo("d")
         }
     }
 
     @Test
-    fun `test convertDateToAge 0 days - default`() {
+    fun `test convertDateToAge 4 days - default`() {
```
jimmyfagan, line 727: I think we can similarly combine some of these other tests with the same kind of `forEach` block as my suggestion above. Consider it optional!

prime-router/src/test/kotlin/fhirengine/translation/hl7/utils/CustomFHIRFunctionsTests.kt
```diff
@@ -730,7 +770,7 @@ class CustomFHIRFunctionsTests {
     }
 
     @Test
-    fun `test convertDateToAge 4 days - default`() {
+    fun `test convertDateToAge 4 days - Add value for both but switched order`() {
```
jimmyfagan, line 773: This might explain why you had those helper functions, did not realize you would have to account for parameters in either position. Seems like there's gotta be a better way, but I don't know what that is 🤔

jimmyfagan (commented):
Left some comments, some are just ideas to make things a little cleaner and more readable, some of it might need some follow-up (basically the handling of multiple parameters questions). Going to leave it at that for now, feel free to go through and respond with your thoughts and make edits as you see fit, I'll think more about the parameters thing and hopefully come back with some idea of how to handle that.

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/CustomFHIRFunctions.kt
```diff
@@ -248,15 +253,37 @@ object CustomFHIRFunctions : FhirPathFunctions {
                 else -> throw SchemaException("Call with day, month, or year")
             }
         } else {
-            calculateAgeWithAssumption(date)
+            calculateAgeWithAssumption(date, comparisonDate)
+        }
+    }
+
+    private fun getParamValue(params: List<Base>): Pair<LocalDate, String?> {
```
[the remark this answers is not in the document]
JessicaWNava: Definitely. Both the unit of time and the date to compare are optional params. It just seemed easiest to allow them in either position since they may only pass one or both of them and I will already have the logic for checking the first param for both. I will update the method names to make it include the actual function name as well as an explanation. I will also add kdocs for them all

prime-router/src/test/kotlin/fhirengine/engine/FhirConverterTests.kt
JessicaWNava: Probably? I will revert

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/CustomFHIRFunctions.kt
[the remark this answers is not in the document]
jimmyfagan: Gotcha, yea I think this might be a good 17th, or eng sync topic. This `parameters` setup has always been confusing and I can't quite tell whether it's something we're in control of or having to conform to. But having a list of lists has never made sense to me, and it might be nice to be able to have named parameters or something. Kind of a larger discussion.. As for this PR:

I still think that if we want to support these params in either position based on the value type, this is still a pretty hard-to-read way of doing that, just following the various pairs and having somewhat generic variable names gets confusing. What about doing something like this (pseudocode), either in a helper function or in the main function?

```
if params.size > 2 fail
initialize timePeriodIndicator, comparisonDate
params.forEach{
  when(it.type) {
    DateType -> set comparisonDate, fail if already set
    StringType -> set timePeriodIndicator, fail if already set or invalid value
    else -> fail
  }
}
```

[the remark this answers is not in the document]
JessicaWNava: Changes made, ready for you again. Thank you for the cleanup, it is really awesome. I think I was too far into the weeds on this one.

prime-router/src/test/kotlin/fhirengine/translation/hl7/utils/CustomFHIRFunctionsTests.kt
```diff
@@ -646,103 +612,51 @@ class CustomFHIRFunctionsTests {
...
+        results.forEach { ageList ->
+            val age = ageList[0]
+            assertThat(age is Age).isEqualTo(true)
+            if (age is Age) {
+                assertThat(age.unit).isEqualTo("day")
+                assertThat(age.value.toInt()).isEqualTo(0)
+                assertThat(age.code).isEqualTo("d")
+            }
         }
     }
```
jimmyfagan, line 662: We should probably update the name of the below test.

jimmyfagan: This also applies for some of the combined tests above, don't need things like `default` in the test names anymore for the most part.

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/CustomFHIRFunctions.kt
```diff
@@ -253,36 +271,23 @@ object CustomFHIRFunctions : FhirPathFunctions {
...
+     * @return an age in years, months, weeks, or days.
+     */
+    private fun convertDateToAgeCalculateAgeWithAssumption(
+        date: DateType,
+        comparisonDate: LocalDate?
+    ): MutableList<Base> {
         val period = Period.between(
-            // have to do plus one for the month because it is expecting 1 based and we get zero based
-            LocalDate.of(date.year, date.month + 1, date.day),
+            // have to do plus one for the month because it is expecting 1 based, and we get zero based
```
jimmyfagan, line 289: We can move this comment to the helper function!

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/CustomFHIRFunctions.kt
```diff
@@ -207,11 +212,24 @@ object CustomFHIRFunctions : FhirPathFunctions {
...
-            timePeriodSpecified = params.second
-            comparisonDate = params.first
+        var comparisonDate: LocalDate? = LocalDate.now()
+        if (!parameters.isNullOrEmpty()) {
+            if (parameters.size > 2) {
+                throw SchemaException("Must call the convertDateToAge function with no more than two parameters")
+            }
+
+            parameters.forEach {
+                it.forEach { actualParam ->
```
jimmyfagan, line 222: Instead of looping through the sublist with another `forEach`, we should probably instead check that each sublist has exactly 1 element and just throw another SchemaException if that's not the case.

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/CustomFHIRFunctions.kt
```diff
@@ -207,11 +212,24 @@ object CustomFHIRFunctions : FhirPathFunctions {
...
-        var comparisonDate: LocalDate = LocalDate.now()
-        if (!parameters.isNullOrEmpty() && !parameters[0].isNullOrEmpty()) {
-            val params = getParamValue(parameters[0])
-            timePeriodSpecified = params.second
-            comparisonDate = params.first
+        var comparisonDate: LocalDate? = LocalDate.now()
+        if (!parameters.isNullOrEmpty()) {
+            if (parameters.size > 2) {
+                throw SchemaException("Must call the convertDateToAge function with no more than two parameters")
+            }
```
jimmyfagan, line 219: Maybe something like this too? Might make these checks long enough to justify a helper function again, idk 😬 
```suggestion
if (parameters.size > 2) {
                throw SchemaException("Must call the convertDateToAge function with no more than two parameters")
            } else if ((parameters.size > 0 && parameters[0].size != 1) ||
                (parameters.size > 1 && parameters[1].size != 1)
            ) {
                throw SchemaException("Must call the convertDateToAge function without empty parameters")
            } else if (parameters.size == 2) {
                if ((parameters[0][0] is DateType && parameters[1][0] is DateType) ||
                    (parameters[0][0] is StringType && parameters[1][0] is StringType)
                ) {
                    throw SchemaException("Cannot call the convertDateToAge function with two parameters of the same type")
                }
            }
```

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/CustomFHIRFunctions.kt
jimmyfagan: If you take the suggestion from the other comment, this can actually just become something like `val actualParam = it[0]` since it will be validated above.

jimmyfagan (commented):
Looks great, just a few more comments for minor cleanup things and some additional validation on `parameters`.

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/CustomFHIRFunctions.kt
JessicaWNava: So, after changing this and having the tests fail, I remembered that it is just a list of a list that holds the params so listOf(listOf(param1, param2)) so we actually need that second foreach because that is where the actual params are. I guess I could do the check for the first list, but given that is just how it works, it seems odd

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/CustomFHIRFunctions.kt
JessicaWNava: Got it. Added checks for that as well as tests.

jimmyfagan (approved):
Looks good, I think it's ready for Arnej's review.

JessicaWNava:
@arnejduranovic ready for yah

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/CustomFHIRFunctions.kt
```diff
@@ -195,29 +196,58 @@ object CustomFHIRFunctions : FhirPathFunctions {
...
+        var timePeriodSpecified: String? = null
+        var comparisonDate: LocalDate? = LocalDate.now()
+        if (!parameters.isNullOrEmpty()) {
+            if (parameters[0].size > 2) {
+                throw SchemaException("Must call the convertDateToAge function with no more than two parameters.")
+            }
+
+            if (parameters[0].size == 2 && parameters[0][0]::class == parameters[0][1]::class) {
+                throw SchemaException(
+                    "Must call the convertDateToAge function no more than one " +
```
arnejduranovic, line 223: `with` no more than one?

arnejduranovic (changes requested):

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/CustomFHIRFunctions.kt
```diff
@@ -195,29 +196,58 @@ object CustomFHIRFunctions : FhirPathFunctions {
...
         if (date !is DateType) {
-            throw SchemaException("Must call the convertDateToAge function on a single dateTimeType")
+            throw SchemaException("Must call the convertDateToAge function on a DateType.")
         }
 
-        val periodsOfTime = listOf("day", "month", "year")
-        return if (parameters != null && parameters.size > 0 && parameters[0].size > 0 &&
-            parameters[0][0].toString().lowercase() in periodsOfTime
-        ) {
+        var timePeriodSpecified: String? = null
```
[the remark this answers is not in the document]
JessicaWNava: Ready for your re review @arnejduranovic Thank you for all of the feedback!

prime-router/src/test/kotlin/fhirengine/translation/hl7/utils/CustomFHIRFunctionsTests.kt
```diff
@@ -562,7 +563,7 @@ class CustomFHIRFunctionsTests {
...
 
     @Test
-    fun `test convertDateToAge in future`() {
+    fun `test convertDateToAge - add a day`() {
```
arnejduranovic, line 566: Anything testing convertDateToAge specifically should probably get moved to a new test class specific to the converter class. This helps with organization and next time we have to extend/modify the worker class we don't have to touch code that does other stuff.

arnejduranovic (changes requested):

arnejduranovic (approved):
Thank you both for this great effort!

JessicaWNava:
Added one more commit based on failing manual test if you want to re-review @arnejduranovic

PR #10071, opened 2023-06-22
Optimize build-frontend workflow

### LINKED ISSUE #14324: Improve frontend PR checks overall time to complete
Frontend PR checks should ideally take around 5 minutes to complete as the longer it takes the longer it takes to react to those results (such as merging the PR).

# Acceptance Criteria
- Frontend PR checks take ~5 minutes to complete

jpandersen87:
Fixes #14324

This PR creates a dedicated `frontend_ci` workflow for pull requests within the `frontend-react` subdirectory. Both this new workflow and `build_frontend` have been updated with proper conditionals so that this new workflow runs only for non-deployment PRs (with the older workflow still continuing to conditionally run for anything else).

Also:
- `chromatic_pr` removed as it is built into `frontend_ci` (and fixed so that TurboSnap works within PR context)
- `chromatic_master` renamed to `frontend_chromatic_main`
- chromatic comments in PRs updated to give more details and to remove prior comment before posting new one (to match general bot behavior of analysis runs as it keeps the comment within the relevant timeframe and will always send notifications)
- new output variable from `build-var` for deployment PR detection
- concurrency added to `sonarcloud`, `build_hub`, and `build_frontend` workflows


.github/actions/build-backend/action.yml
```diff
@@ -30,11 +44,32 @@ runs:
...
+      if: inputs.GITHUB_TOKEN != '' && inputs.SONAR_TOKEN != ''
+      uses: sonarsource/sonarcloud-github-action@db501078e936e4b4c8773d1bb949ba9ddb7b6b6a
+      env:
+        GITHUB_TOKEN: ${{ inputs.GITHUB_TOKEN }}
+        SONAR_TOKEN: ${{ inputs.SONAR_TOKEN }}
+      with:
+        args: >
+          -Dsonar.cpd.exclusions=prime-router/src/testIntegration/**,prime-router/src/main/kotlin/cli/tests/**
+          -Dsonar.coverage.exclusions=prime-router/src/test/**,prime-router/src/testIntegration/**,prime-router/src/main/kotlin/cli/tests/**
+          -Dsonar.sources=prime-router/src
```
snesm, line 67: I've found that if we don't include both the frontend and backend source location in the sonarcloud call, the missing source results get deleted from sonarclouds results data.

.github/actions/build-backend/action.yml
```diff
@@ -30,11 +44,32 @@ runs:
...
+      working-directory: prime-router
+      run: ./gradlew jacocoTestReport -x fatjar
+      shell: bash
+
     - name: Tar router files
       working-directory: prime-router
       shell: bash
       run: tar -czvf prime-router-build.tar.gz build
 
+    - name: SonarCloud Scan
```
snesm, line 57: I don't believe we want to add a sonarcloud scan to this.  It's redundant with build_hub.yml and would slow things down.

snesm (changes requested):

.github/workflows/build_hub.yml
```diff
@@ -4,6 +4,11 @@ on:
   pull_request:
     branches:
       - master
+    paths:
```
snesm, line 7: unnecessary: this is short-circuted based on changes detected via build vars
or should we move away from the using the build vars logic

.github/workflows/build_hub.yml
```diff
@@ -4,6 +4,11 @@ on:
   pull_request:
     branches:
       - master
+    paths: 
+      - prime-router/**
+    types:
```
snesm, line 9: Default is `opened`, `synchronize`, and `reopened`.  Why remove reopened?
(same throughout PR)

.github/workflows/frontend_ci.yml
```diff
@@ -0,0 +1,298 @@
...
+
+  e2e_tests:
+    name: E2E tests
+    needs: lint
+    runs-on: ubuntu-latest
+    # Keep shards under 5 minutes
+    timeout-minutes: 5
+    strategy:
+      fail-fast: false
+      matrix:
```
snesm, line 76: nice parallelization

.github/workflows/frontend_ci.yml
```diff
@@ -0,0 +1,298 @@
+name: Frontend CI
+
+on:
+  pull_request:
```
snesm, line 4: all branches (that match path)?

.github/workflows/build_frontend.yaml
```diff
@@ -3,7 +3,6 @@ name: Frontend
 on:
   pull_request:
     branches:
```
snesm, line 5: Does this obsolete this workflow and `.github/actions/build-frontend` (used below)

.github/workflows/build_hub.yml
jpandersen87: Ignorance on my part on what the default set of types are if not specified. Removed.

.github/workflows/build_hub.yml
jpandersen87: We should probably move away from doing path detection if it's simple enough to be covered using the `paths` property in a github action doc.

.github/workflows/frontend_ci.yml
jpandersen87: Except for `production`, I've just made an update to fix that.

.github/workflows/build_frontend.yaml
jpandersen87: Currently, it does not. This only changes `pull_request`-based workflows (except `production`) as a starter. Push-based actions will still trigger the original flows as well as PRs for `production`. This can be applied wider after an observation period if dev-ops desires.

.github/workflows/build_frontend.yaml
```diff
@@ -3,7 +3,6 @@ name: Frontend
...
-      - master
```
snesm, line 6: removing `master` means this workflow will not run.  `required` workflows must run (even if there is nothing to do).

.github/workflows/build_hub.yml
snesm: limiting this to run only when there are changes to `prime-router` means this workflow will not run.  `required` workflows must run (even if there is nothing to do).

snesm (dismissed):

.github/workflows/build_frontend.yaml
jpandersen87: Ah, I see what you were getting at now. Yeah there's a conundrum here with required workflows...

jpandersen87: build_frontend has been reverted back to conditional-based behavior

frontend-react/dummy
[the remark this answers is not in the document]
jpandersen87: it will be deleted prior to merge, it's only there to force PR workflows to run prior to approval

.github/workflows/frontend_ci.yml
```diff
@@ -0,0 +1,310 @@
...
+          body: |
+            Branch deployed to Chromatic 🚀. 
+            - ${{ steps.chromatic.outputs.changeCount && format('⚠️ Detected {0} tests with visual changes.', steps.chromatic.outputs.changeCount) }}
+            - ${{ steps.chromatic.outputs.errorCount > 0 && format('❌ {0} tests failed!', steps.chromatic.outputs.errorCount) || '✅ All tests passed.' }}
+
+            View via: 
+            - Chromatic: ${{ steps.chromatic.outputs.url }}
+            - Storybook: ${{ steps.chromatic.outputs.storybookUrl }}
+
+  # codeql:
```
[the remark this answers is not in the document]
jpandersen87: I left it there for future reference in case it becomes relevant

penny-lischer (approved):
lgtm

snesm (approved):

etanb (approved):
honestly (and sadly), we could probably remove Chromatic for now until we integrate ourselves into the Design process again.

PR #10102, opened 2023-06-26
Feature/update functionapp tasktable check

lucasdze:
This PR is just to re-word message title on the slack notification workflow


jeremy-page (approved):
Thank you

Issue #9735, opened 2023-06-06
Test Forcing TLSv1.2 for REST Transport


GilmoreA6:
Stephen Nesman is asking for our review on PR#9515 to force TLSv1.3 with fallback to 1.2 (only) for REST transport. We currently have three REST receivers OK, NY & WA. We need to validate that each can use TLS1.3 or 1.2 and test sending a message to their environment

https://github.com/CDCgov/prime-reportstream/pull/9515

jalbinson:
I ran the following command to test whether TLS1.2 was supported by NY
```
openssl s_client -connect api2.health.ny.gov:443 -tls1_2
```

Although the session was not opened because NY requires a client certificate, it does appear that TLS 1.2 version is supported by NY.

PR #9515, opened 2023-05-17
Force strong TLS

snesm:
This PR forces TLS v 1.3 (with fallback to 1.2 only) for the REST transport.  Use of the non-specific "TLS" may allow negotiating (or downgrading to) obsolete, dangerous and disallowed SSL protocols.

**If you are suggesting a fix for a currently exploitable issue, please disclose the issue to the prime-reportstream team directly outside of GitHub instead of filing a PR, so we may immediately patch the affected systems before a disclosure. See [SECURITY.md/Reporting a Vulnerability](https://github.com/CDCgov/prime-reportstream/blob/master/SECURITY.md#reporting-a-vulnerability) for more information.**

Test Steps:
1. Verify that all REST transport recipients are unaffected

## Changes
- Change "TLS" to "TLSv1.3". If this fails, attempt "TLSv1.2".

### Testing
- [ ] Tested locally?
- [ ] Ran `./prime test` or `./gradlew testSmoke` against local Docker ReportStream container?
- [ ] (For Changes to /frontend-react/...) Ran `npm run lint:write`? 
- [ ] Added tests?

### Process
- [ ] Are there licensing issues with any new dependencies introduced?
- [x] Includes a summary of what a code reviewer should test/verify?
- [ ] Updated the release notes?
- [ ] Database changes are submitted as a separate PR?
- [ ] DevOps team has been notified if PR requires ops support?

## Linked Issues
- Fixes #issue

## To Be Done
*Create GitHub issues to track the work remaining, if any*
- #issue

## Specific Security-related subjects a reviewer should pay specific attention to
- Does this PR introduce new endpoints?
    - new endpoint A
    - new endpoint B
- Does this PR include changes in authentication and/or authorization of existing endpoints?
- Does this change introduce new dependencies that need vetting?
- Does this change require changes to our infrastructure?
- Does logging contain sensitive data?
- Does this PR include or remove any sensitive information itself?

If you answered '_yes_' to any of the questions above, conduct a detailed Review that addresses at least:

- What are the potential security threats and mitigations? Please list the _STRIDE_ threats and how they are mitigated
    - **S**poofing (faking authenticity)
        - Threat _T_, which could be achieved by _A_, is mitigated by _M_
    - **T**ampering (influence or sabotage the integrity of information, data, or system)
    - **R**epudiation (the ability to dispute the origin or originator of an action)
    - **I**nformation disclosure (data made available to entities who should not have it)
    - **D**enial of service (make a resource unavailable)
[...]

PR #10123, opened 2023-06-27
9735 force using TLS 1.3 for NY-PHD

jalbinson:
This PR forces TLS v1.2 for Rest Transport that uses a client authentication certificate.  We do not want receivers to be able to negotiate to an outdated TLS version which was possible in the previous implementation.  This is currently only used with NY-PHD but could be others in the future.

Test Steps:
1. Pull this branch to your local machine
2. Update your locally running ReportStream with NY-PHD production settings and credentials
3. Use the check functionality in the org UI to test connecting to the NY-PHD receiver
4. The dialog will let you know if everything works

## Changes
- Update SSL version in Rest Transport

### Testing
- [x] Tested locally?
- [ ] Added tests?

### Process

## Linked Issues
- Fixes https://github.com/CDCgov/prime-reportstream/issues/9735


oslynn (approved):

prime-router/src/main/kotlin/transport/RESTTransport.kt
```diff
@@ -500,7 +500,7 @@ class RESTTransport(private val httpClient: HttpClient? = null) : ITransport {
             keyManagerFactory.init(keyStore, jksPasscode)
             // create the sslContext, type TLS, with the keyManager
             // note the ktor sample code uses a TrustManger instead, but that fails with the NY cert file
-            val sslContext = SSLContext.getInstance("TLS")
+            val sslContext = SSLContext.getInstance("TLSv1.2")
```
snesm, line 503: 1.3 would be a better long term choice. https://github.com/CDCgov/prime-reportstream/pull/9515/files

I'm told that this code affects all REST partners.

snesm (changes requested):

prime-router/src/main/kotlin/transport/RESTTransport.kt
snesm: and NY is capable of 1.3
https://www.ssllabs.com/ssltest/analyze.html?d=api2.health.ny.gov&hideResults=on

jalbinson: It only affects NY as the `getSslContext` function only runs given the receiver has a configured JKS for client certificate auth. Our other REST partners (OK, WA) don't use that so this code does not run.

snesm: In that case, this should be"TLSv1.3".

jalbinson: Sounds good but let me double check that this works with JDK 11.

jalbinson: worked fine

Screenshot 2023-06-27 at 2 40 51 PM

snesm (approved):

PR #10361, opened 2023-07-11
update GH workflows + update table filename

etanb:
From the updated README:

## Chromatic and CI Autobuilds

[Chromatic](https://www.chromatic.com/) is a tool for hosting and publishing different versions of a given repository's Storybook. We use Chromatic to host an up-to-date version of all of our Storybook components (any file that ends with `**.stories.tsx` syntax) so that our non-technical folks can see all of our components on the web. All of our CI Autobuild Github workflows can be found in both `.github/workflows/chromatic-master.yml` and `.github/workflows/chromatic-pr.yml`.

`.github/workflows/chromatic-master.yml` triggers a Chromatic build anytime a PR gets merged into our `master` branch.

`.github/workflows/chromatic-pr.yml` triggers a Chromatic build anytime a file with `// AutoUpdateFileChromatic` comment on its FIRST LINE is checked in to a PR. The goal here is to automatically update our Chromatic anytime a file that has an associated Storybook is modified.


etanb:
Once we're happy with this method, I'll make the future PRs with `__chromatic-sync` appending to corresponding components.

.github/workflows/chromatic-pr.yml
```diff
@@ -1,13 +1,13 @@
 # Build a branch-specific Storybook instance on Chromatic
-# To trigger, add the 'chromatic' label to the pull request
-# To re-trigger, remove and add the 'chromatic' label to the pull request
+# Any file that has "__chromatic-sync" appended to the end of the filename
```
KeithNava, line 2: Awesome work @etanb! I have a question and this might be a silly question since I don't know a lot about Chromatic but why is the `__chromatic-sync` appending required? I would imagine there would be a way to have Chromatic automatically sync any component that's pushed in a PR. Or maybe not? I'm mostly just curious

etanb: I changed the "magic updating" to be if a file with the first line comment being `// AutoUpdateFileChromatic` is checked into a PR, then update Chromatic.

We can easily make Chromatic launch on EVERY PR, but the huge majority of our file changes don't involve component updates. This way, whenever specific files with the special `AutoUpdateFileChromatic` comment are updated, Chromatic is automatically triggered.

penny-lischer (approved):
LGTM

.github/workflows/chromatic-pr.yml
```diff
@@ -19,15 +19,40 @@ jobs:
     name: Chromatic
     runs-on: ubuntu-latest
     steps:
-      - name: Set vars
+      - uses: actions/checkout@v3
```
snesm, line 22: Please use [sha hash](https://devopsjournal.io/blog/2021/12/11/GitHub-Actions-Maturity-Levels) based versioning

.github/workflows/chromatic-pr.yml
```diff
@@ -19,15 +19,40 @@ jobs:
...
+        with:
+          fetch-depth: 0
+
+      - name: Get changed files
+        id: changed_files
+        uses: tj-actions/changed-files@v37
```
snesm, line 28: Please use [sha hash](https://devopsjournal.io/blog/2021/12/11/GitHub-Actions-Maturity-Levels) based versioning

snesm (changes requested):

jeremy-page:
@etanb Will you please update the actions to use a SHA instead of version #? This makes sure they can't be replaced maliciously.

snesm (approved):

Issue #10367, opened 2023-07-11
end2end UP smoketest is flaky


mkalish:
## Describe the bug
The  end2end_up can fail on the CI server even where there no changes.

The failure is `***async end2end_up FAILED***: Batch report id null`

## Impact on ReportStream
Causes devs to have to re-run the smoke tests on PRs in order to merge

## Steps to reproduce
It does not happen every time, but there are various examples of this occurring:

- https://github.com/CDCgov/prime-reportstream/actions/runs/5517339169/jobs/10059825346
- https://github.com/CDCgov/prime-reportstream/actions/runs/5498677992/jobs/10043870926

## Expected behavior
The tests should not be flaky

## Additional context
This is likely due to the resource constraints of the CI infrastructure as this cannot be recreated locally.

PR #10369, opened 2023-07-11
10367: increase timeout for step polling and confirm batch results were found

mkalish:
This PR fixes the flaky end2end_up smoke test by increasing the timeout

Test Steps:
n/a

I'll run the CI several times to verify

## Changes
- Increases the step poll timeout to 600 secs
- Errors early if there were not batch results found

### Testing
- [x] Tested locally?
- [ ] Added tests?

### Process

## Linked Issues
- #10367


prime-router/src/main/kotlin/cli/tests/TestReportStream.kt
```diff
@@ -474,7 +474,7 @@ abstract class CoolTest {
     suspend fun pollForStepResult(
         reportId: ReportId,
         taskAction: TaskAction,
-        maxPollSecs: Int = 480,
+        maxPollSecs: Int = 600,
```
mkalish, line 477: The end2end_up smoke test for the ELR pipeline topic actually has two translate steps (and batch, send).  The flakiness was the result of some time that batch for the second translate report had not run yet and sometimes it had.

prime-router/src/main/kotlin/cli/tests/BasicTests.kt
```diff
@@ -175,6 +175,11 @@ class End2EndUniversalPipeline : CoolTest() {
             }
             translateReportIds.forEach { translateReportId ->
                 val batchResults = pollForStepResult(translateReportId, TaskAction.batch)
+                if (batchResults.isEmpty()) {
```
mkalish, line 178: Since the size of `batchResults` is never checked, the for loop on 184L, will happily pass even if there are no results which should be an error.

snesm (approved):

luis-pabon-tf (approved):
All tests pass, changes make sense. Glad to see these tests fixed!

Issue #10200, opened 2023-06-30
Create receiver for CDC-ELIMS-HL7 org in STAGING environment


arnejduranovic:
### User Story

As the CDC-ELIMS-HL7 organization, I need the ability to receive all HL7v2 ELR (ORU-R01) messages flowing through RS that have the "elims-elr" topic assigned.

### Description/Use Case

We need to configure RS:
1. Create a receiver under the CDC-ELIMS-HL7 org in staging that can receive any HL7v2 ELR message with the "elims-elr" topic
2. Setup transport protocol with CDC-ELIMS in staging and test they can receive a message. **ELIMS would like to use SFTP WITH our Azure Blob Store**, so we will need to contact them for the info.

### Risks/Impacts/Considerations

- CDC ELIMS would like to receive the exact same message that goes to a particular STLT. This means that we CANNOT use any kind of transforms until a solution is figured out that lets us send an exact copy of the message a STLT received. This is because if we set up a special transform for a STLT, then the cdc receiver will need that same exact transform configured otherwise the message they and the STLT receive will be different and this is untenable due to having to support more than one STLT.

### Dev Notes

Request for connectivity from ELIMS:
1. We can connect Rhapsody to Azure Storage BLOBS using sFTP client comm points
  - You can configure Azure Storage BLOBS to make their data available through sFTP interfaces
  - Each Azure Storage BLOB could be considered sFTP storage and Rhapsody could interact with it the same as it interacts with any sFTP server host
  - This approach would only work if RS has turned on sFTP services for the Storage BLOB

### Acceptance Criteria

- [x] receiver created and configured
- [x] transport protocol configured (SFTP with Azure Blob)
- [x] Tested receiver can receive any ORU-R01 message under the "elims-elr" topic

bishoyayoub:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/10200?planning-poker) with Zenhub @arnejduranovic @JessicaWNava @JFU-NAVA-PBC @jimmyfagan @luis-pabon-tf @thetaurean

avnieldravid:
we plan on discussing this today at 3pm

avnieldravid:
Email sent to ELIMS team on 7/12

avnieldravid:
We will setup ELIMS to receive messages with the REST protocol .

avnieldravid:
[...]

PR #10114, opened 2023-06-26
Adjust convert/translate schemas to handle ELIMS test message 40 (HL7->FHIR->HL7)

jimmyfagan:
This PR involves a variety of updates to convert/translate schemas in support of lossless conversion of a sample rhapsody message from HL7v2 to FHIR and back to HL7v2. Note that the diff for some files may be much easier to view in an IDE than in github.

Test Steps:
1. Update the dev-elims receiver in `organizations.yml` to be active and reload the settings.
2. Send a report to ReportStream using message 40 in [this folder](https://cdc.sharepoint.com/teams/ReportStream/Shared%20Documents/Forms/AllItems.aspx?RootFolder=%2Fteams%2FReportStream%2FShared%20Documents%2FELIMS%2DELR) as the message body.
3. Compare the resulting HL7 to the input HL7 and observe that any differences are appropriately documented in the [notes spreadsheet](https://cdc.sharepoint.com/:x:/r/teams/ReportStream/Shared%20Documents/ELIMS-ELR/[40_Message_RS_processing_notes.xlsx](https://cdc.sharepoint.com/:x:/r/teams/ReportStream/Shared%20Documents/ELIMS-ELR/40_Message_RS_processing_notes.xlsx?d=w0e69763787ea46ffa7187cee9417b41d&csf=1&web=1&e=8bpFIa)?d=w0e69763787ea46ffa7187cee9417b41d&csf=1&web=1&e=8bpFIa)

## Changes
- Added custom extension for MSH.15 and MSH.16
- Removed CE.6 from the `coding_1.version` in `CodeableConcept.yml` (seemed incorrect and could not find justification for this)
- Add FHIR mapping using custom extension for MSH.21
- Add CLIA universal id in `MessageSender.yml`
- Use `CodeableConcept` instead of `Reference/Organization` when mapping OBX.15 to FHIR (FHIR->HL7v2 can still handle either, this was done because Organization is incomplete in getting all 6 fields from OBX.15)
- Fix reading of `CodeableConcept` to populate OBX.15
- Add FHIR->HL7v2 translation of OBX.16
- Add condition to only populate OBX.25 if the practitioner has a MDIR code
- Fix bad constant in `ce-coded-with-exceptions` to allow CWE.9 to be properly populated
- Add conditions for PID.10.3, CX.4, and CX.6 to avoid unnecessarily populating
- Update XAD mapping to more closely match spec
- Add condition for XTN.12 to not use for email
[...]

PR #10426, opened 2023-07-13
Add elr-elims topic for UP

brick-green:
This PR creates a new topic(`elr-elims`) for use on the universal pipeline and adds a sample sender to the org file.

Test Steps:

1. Check out and build the branch
2. add a receiver:
```
receivers:
    - name: MN_ELIMS
      organizationName: CDC-ELIMS-HL7
      topic: elr-elims
      customerStatus: active
      jurisdictionalFilter: [ "(%performerState.exists() and %performerState = 'MN') or (%patientState.exists() and %patientState = 'MN')" ]
      qualityFilter: []
      conditionFilter: []
      timing:
        operation: MERGE
        numberPerDay: 1440 # Every minute
        initialTime: 00:00
      translation: !<HL7>
        schemaName: "metadata/hl7_mapping/ORU_R01/ORU_R01-base"
        useTestProcessingMode: false
        useBatchHeaders: true
        type: "HL7"
        receivingApplicationName: ""
        receivingFacilityName: ""
      transport:
        type: SFTP
        host: sftp
        port: 22
        filePath: ./upload
        credentialName: DEFAULT-SFTP
```
4. run `./gradlew reloadSettings`
5. Send a test report using this collection:
[elims test.postman_collection.txt](https://github.com/CDCgov/prime-reportstream/files/12044120/elims.test.postman_collection.txt)
6. Verify that the request was accepted
7. Open the generated HL7 file in the local sftp folder and verify that it is HL7

## Changes
- added elr-elims topic
- add one test and expanded others
- added sender to org file

### Testing
- [X] Tested locally?


prime-router/settings/organizations.yml
```diff
@@ -987,4 +987,28 @@
...
+    schemaName: "metadata/fhir_transforms/senders/default-sender-transform"
+    processingType: "sync"
+    allowDuplicates: true
+    senderType: null
+    primarySubmissionMethod: null
+    topic: "elr-elims"
+    version: null
+    createdBy: null
+    createdAt: null
+  receivers: []
```
jimmyfagan, line 1014: Totally optional, but we can keep things a little bit more concise by getting rid of some of the `null` and `[]` entries here.

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
```diff
@@ -115,7 +124,7 @@ class FHIRRouter(
      */
     val processingModeDefaults = mapOf(
         Pair(Topic.FULL_ELR, processingModeFilterDefault),
-        Pair(Topic.ETOR_TI, processingModeFilterDefault),
+        Pair(Topic.ETOR_TI, processingModeFilterDefault)
```
jimmyfagan, line 127: I want to make a pitch in favor of changing your formatter settings to allow trailing commas, but not important 😄 Also want to mention that I'm happy with not having a default proc mode filter for the new topic here for now.

jimmyfagan (approved):
Looks good, left a few comments, but nothing that blocks me from approving. Nice work!

prime-router/settings/organizations.yml
brick-green: good call here!

prime-router/settings/organizations.yml
```diff
@@ -987,4 +987,15 @@
         host: sftp
         port: 22
         filePath: ./upload
-        credentialName: DEFAULT-SFTP
\ No newline at end of file
+        credentialName: DEFAULT-SFTP
+- name: "CDC-ELIMS-HL7"
```
mkalish, line 991: Optional, but could add a testing receiver here as well and add an additional assertion in [this](https://github.com/CDCgov/prime-reportstream/blob/251110fb97cca7f5662a9e3b21af7f0c01b18428/prime-router/src/main/kotlin/cli/tests/BasicTests.kt#L91) smoke test

mkalish (commented):
Looks great!

prime-router/settings/organizations.yml
jimmyfagan: A receiver here overlaps a bit with #10200, but I think that would be fine (especially since I'll need it in [my PR](https://github.com/CDCgov/prime-reportstream/pull/10114) once this is merged), and then that ticket can be more focused on the transport side of things.

prime-router/src/main/kotlin/SettingsProvider.kt
```diff
@@ -48,6 +48,7 @@ enum class CustomerStatus {
 enum class Topic(@JsonValue val jsonVal: String, val isUniversalPipeline: Boolean = true) {
     FULL_ELR("full-elr", true),
     ETOR_TI("etor-ti", true),
+    ELR_ELIMS("elr-elims", true),
```
arnejduranovic, line 51: I think the idea here is slightly off. The topic should be "CDC-ELIMS" because this topic is to be specific for the CDC ELIMS use case/work flow. This use case, at this time, has the following special requirements that a topic allows us to meet easier:
1. Observability: We need to write a query to answer this question: For a particular STLT (RS receiver), is there a connection set up for that STLT to receive both test messages and production messages from CDC ELIMS? This is much more doable with a topic and something like a "elr-elims" topic that is used by other senders would not allow us to do
2. CDC ELIMS as system of record: For the pilot effort, the CDC ELIMS team wants their system to be the "system of record", meaning all messages that we deliver to a STLT should be sent to the configured CDC receiver so the CDC ELIMS team can run automatic comparisons on the data they sent us vs the data we sent to the STLT. This is probably doable without a topic (we can filter on the system name in the MSH) 
3. Partitioning data streams: For this pilot, we want to ensure that existing receivers DO NOT receive this data since we will have to work with each STLT that is to be onboarded to verify they accept the messages CDC ELIMS is sending (although the pilot is currently set to be staging only, we want to have this guard rail in place in case CLIA compliance clears and we can go to production).

I'm hoping we can find a better solution to these issues during the pilot (like ReportStream being the system of record and not having to loopback messages to a CDC ELIMS receiver), but these are the needs at the moment. Indeed, only item one above really actually depends on the topic concept and we should look to figure out a better long term solution.

I understand wanting to create a generic topic like "ELR_ELIMS" that can be used more broadly, but I'm not sure I see the value of such a topic. "ELR_ELIMS" is just "FULL_ELR" and I think we can just use the FULL_ELR topic for all those cases. If we find that we can't, then perhaps we can consider a "ELR_ELIMS" topic.

arnejduranovic (changes requested):

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
```diff
@@ -94,12 +94,21 @@ class FHIRRouter(
         "%messageId.exists()",
     )
 
+    /**
+     * Default Rules for quality filter on ELR_ELIMS topic:
+     *   no rules; completely open
+     */
+    private val elrElimsQualityFilterDefault: ReportStreamFilter = listOf(
+        "true",
```
arnejduranovic, line 102: I can see this getting to be quite juicy, eventually. But good for the pilot for now

arnejduranovic (approved):

prime-router/settings/organizations.yml
brick-green: I added to the end2end_up test which required adding a sender and receiver to the ignore org. So with that I deleted the CDC ELIMS org and sender that I had previously added

jimmyfagan (approved):
New commits look good to me!

prime-router/src/main/kotlin/SettingsProvider.kt
jimmyfagan: Just remembered this conversation, @brick-green was this resolved offline, or does this need to be talked through this again?

brick-green: We're good here @jimmyfagan. Appreciate the follow up!!

PR #10578, opened 2023-07-21
Feature/9117 automate reverting deploys

lucasdze:
This PR has been created to rollback the deployment process to the previous containerID before the latest
This workflow is manual and should only be run when there is a deployment failure to prod


jeremy-page (approved):
This is safe to merge even if it does not work since it's not triggered automatically. Just make sure you test it in `test` 😄

PR #10614, opened 2023-07-25
HL7 ORM to FHIR Translation (OBX and OBR segments, and some Patient modifications)

jorg3lopez:
This PR add additional logic to translate ORM HL7 data to FHIR.

Specifically,

1. Added `OBX` to `Observation` mapping.
   1. Added a new `observation-status-original-text` extension that persists the original `OBX.11` value because they do not map cleanly.
2. Mapped `OBR` to a `Specimen` and a bit to `ServiceRequest`.
3. Stop setting the `type` for the `value` in a `Quantity` when the value is a decimal.  If we kept it as a `FLOAT`, then the decimal is incorrectly truncated.  Not setting the `type` results in a complete decimal number showing up in the end FHIR.  This occurs if you have enough decimal places beyond what a float can support.
4. Remove the extra `birthTime` extension from `Patient.extension` by editing `Patient.yml`.
5. Edited the custom code in `HL7toFhirTranslator.kt` to only create the extension if the size of `PID.7` is over 8, make the extension value of type `DateTime`, and add the extension under the `birthDate` field which is serialized as `_birthDate`.
Added tests to `HL7toFhirTranslatorTests.kt` to make assertions on whether the extension exists or not depending on the size of `PID.7`.

### Test Steps

1. `/gradlew primeCLI --args='fhirdata --output-format FHIR --input-file ./test/test.hl7 --output-file ./test/test.fhir'`
3. check the expected.fhir file under the test folder and compare it to the test.fhir file. Some information will be different as the FHIR library adds more too it; we are still working on this. For the most part, look for the information from the expected file to be present in the test.fhir file.

### Testing
- [x] Tested locally?


prime-router/src/main/kotlin/fhirengine/translation/HL7toFhirTranslator.kt
```diff
@@ -122,11 +130,17 @@ class HL7toFhirTranslator internal constructor(
                 bundle.addEntry().resource = Patient()
                 bundle.entry.first { it.resource.resourceType.name == "Patient" }.resource as Patient
             }
+
+            if ((birthTime?.length ?: 0) <= 8) {
+                // this field is just a date with no time, no need to create an extension to track the time
+                return
```
JessicaWNava, line 136: The date needs to be added in here since the mapping for the date was removed.

prime-router/src/test/kotlin/fhirengine/translation/HL7toFhirTranslatorTests.kt
```diff
@@ -81,17 +92,38 @@ DG1|1||F11.129^Opioid abuse with intoxication,unspecified^I10C|||W|||||||||1
...
         val patient = bundle.entry.first { it.resource.resourceType.name == "Patient" }.resource as Patient
+
         assertThat(
-            patient.getExtensionByUrl(
+            patient.birthDateElement.getExtensionByUrl(
                 "http://hl7.org/fhir/StructureDefinition/patient-birthTime"
-            ).valueAsPrimitive.toString()
-        ).isEqualTo(birthDate)
+            )
+        ).isNull()
```
JessicaWNava, line 127: This line currently causes a compilation error.

prime-router/src/test/kotlin/fhirengine/translation/HL7toFhirTranslatorTests.kt
```diff
@@ -81,17 +92,38 @@ DG1|1||F11.129^Opioid abuse with intoxication,unspecified^I10C|||W|||||||||1
...
+            "http://hl7.org/fhir/StructureDefinition/patient-birthTime"
+        ).value as DateTimeType
+
+        assertThat(
+            extensionValue.valueAsString
+        ).isEqualTo(Hl7RelatedGeneralUtils.dateTimeWithZoneId(birthDateTime, ""))
+    }
+
+    @Test
+    fun `test birth date extension is missing when birthdate is only date`() {
```
JessicaWNava, line 113: We need to ensure the date is still there as well.

prime-router/metadata/fhir_mapping/hl7/message/ORM_O01.yml
```diff
@@ -10,21 +10,41 @@ resources:
...
+      - .ORDER_DETAIL.OBR
+      - .ORDER_DETAIL.NTE
+      - PATIENT.PATIENT_VISIT.PV1
+      - MSH
+      - PATIENT.PID
 
-  - resourceName: ServiceRequest
-    segment: .ORC
+  - resourceName: Observation
+    segment: .ORDER_DETAIL.OBSERVATION.OBX
```
JessicaWNava, line 30: Does this work? I thought the segment should be `.OBSERVATION.OBX`

JessicaWNava, line 30: Does this work? I thought the segment should be .OBSERVATION.OBX

JessicaWNava (changes requested):
Love that y'all figured out the birthDate extension, awesome job! Got some changes to request and then I will do another, more thorough review. Please lmk when it is ready. Thanks again!

prime-router/src/main/kotlin/fhirengine/translation/HL7toFhirTranslator.kt
halprin: This just handles the birth time extension.  The base `birthDate` still works through the `Patient.yml` file.  That part was not removed.

https://github.com/CDCgov/prime-reportstream/blob/35366ec302a4df4d579ceab5bc8922dff1c36ada/prime-router/metadata/fhir_mapping/hl7/resource/Patient.yml#L260-L263

prime-router/src/test/kotlin/fhirengine/translation/HL7toFhirTranslatorTests.kt
halprin: Oof, yep.  Import statements are hard.  Fixed.

prime-router/src/test/kotlin/fhirengine/translation/HL7toFhirTranslatorTests.kt
halprin: Great idea!  Done.

prime-router/metadata/fhir_mapping/hl7/message/ORM_O01.yml
halprin: It does work.  `Observation.yml` supports extracting information from `ORC` and `OBR` in addition to `OBX`.  So, in order to pass in the `ORC` and `OBR`, we need to change the `group` to the higher level `ORDER` group which then requires the `segment` to have the longer qualified path.  It took us a while to figure out the correct combination of `group` and `segment` since some information in Caristix didn't align with the HAPI HL7v2 library.

prime-router/src/main/kotlin/fhirengine/translation/HL7toFhirTranslator.kt
JessicaWNava: Yep, you are right, sorry about that.

JessicaWNava:
I think this looks good! Let's merge it and then I will do some more thorough testing in the main branch on Monday rather than doing the more thorough testing twice since this is just a side branch anyways. Thanks again for all of your hard work!

JessicaWNava (approved):

Issue #10204, opened 2023-06-30
Test CDC-ELIMS messages: Part 2


arnejduranovic:
### User Story
As CDC ELIMS Rhapsody, I want to be able to send a message through ReportStream that is lossless so that a STLT can receive a CLIA compliant message.

### Description/Use Case
We want to run sample messages sent by the CDC ELIMS team through RS and validate that no data is lost once the message undergoes the translation to internal FHIR and back to HL7v2. If data is lost, we want to update the HL7->FHIR or FHIR->HL7 mappings to fix the issue.

### Risks/Impacts/Considerations
- Some differences between the input HL7v2 and output HL7v2 messages may be okay. We need to meet with the CDC ELIMS team to approve any discrepancies between the two messages. 

### Dev Notes
If fixes are needed, we will update the HL7->FHIR or FHIR->HL7 mappings to address the issues. We may also create a sender transform if needed.

For this story, the following messages will need to be tested and verified:

- [52_04204582_5128_Test.HL7](https://cdc.sharepoint.com/:u:/r/teams/ReportStream/Shared%20Documents/ELIMS-ELR/ELIMS%20artifacts/52_04204582_5128_Test.HL7?csf=1&web=1&e=19Ge9a)
- [52_24095_04204575_5120.HL7](https://cdc.sharepoint.com/:u:/r/teams/ReportStream/Shared%20Documents/ELIMS-ELR/ELIMS%20artifacts/52_24095_04204575_5120.HL7?csf=1&web=1&e=LfQ0ao)

Each HL7v2 message that is being tested should have its own notes document. See the following document for example on how it should be formatted and where it should live: 
- [40_Message_RS_processing_notes.xlsx](https://cdc.sharepoint.com/:x:/r/teams/ReportStream/Shared%20Documents/ELIMS-ELR/40_Message_RS_processing_notes.xlsx?d=w0e69763787ea46ffa7187cee9417b41d&csf=1&web=1&e=oPWYkc)

### Acceptance Criteria
- [ ] The aforementioned HL7v2 messages were run through the universal pipeline and CDC ELIMS team confirmed the output message is good

bishoyayoub:
[...]

PR #10655, opened 2023-07-28
Fix MSH-2, MSH-3 for ELIMS message 52

jimmyfagan:
This PR makes changes that allow MSH-2 and MSH-3 to be preserved when processing ELIMS messages. MSH-2 is typically either `^~\&` or `^~\$#`, but ReportStream has always reduced that to `^~\&`. These changes allow for a custom extension on the FHIR `MessageHeader` to store the value of the field in cases where it is something other than `^~\&`. Then on conversion from FHIR to HL7v2, that extension can also be used to preserve the value as well. As for MSH-3, that was only being changed due to the default sender transform; overriding the relevant schema elements allowed for that to be resolved very easily.

Test Steps:
1. Update the dev-elims receiver in `organizations.yml` to be active and reload the settings.
2. Send a report to ReportStream from the `ignore-elr-elims` sender using message 52 in [this folder](https://cdc.sharepoint.com/teams/ReportStream/Shared%20Documents/Forms/AllItems.aspx?RootFolder=%2Fteams%2FReportStream%2FShared%20Documents%2FELIMS%2DELR) as the message body.
3. Compare the resulting HL7 to the input HL7 and observe that any differences are appropriately documented in the [notes spreadsheet](https://cdc.sharepoint.com/:x:/r/teams/ReportStream/Shared%20Documents/ELIMS-ELR/52_Message_RS_processing_notes.xlsx)

## Changes
- Adjusted HL7v2->FHIR mapping for MessageHeader to store MSH-2 as a custom extension if not `^~\&`
- Added elements in `elims-sender-transform` to override default transforms for MSH-3
- Adjusted currently unused schemas to use the new format for `valueSet`s (unrelated, but discovered while testing and is an easy fix)
- Added FHIR->HL7v2 mapping to read custom extension for MSH-2
- Renamed `ELRReceiver` to `UniversalPipelineReceiver` (again unrelated, but relatively small changeset to do this while in the area)
- Moved or deleted code that blindly replaces `^~\&#` with `^~\&`. In some cases, this is now only done for legacy pipelines.
- Added code to temporarily replace `^~\&#` with `^~\&` before calling library functions which don't handle `^~\&#` in v2.5.1, and then change back to the original value afterward.

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- Addresses #10204

## To Be Done (before submitting PR)
- [x] Add integration tests
- [x] Fix integration tests to expect MSH-2 to be preserved
- [x] Check code coverage

## To Be Done (generally)
- Get updates from ELIMS team about remaining items from [the spreadsheet](https://cdc.sharepoint.com/:x:/r/teams/ReportStream/_layouts/15/Doc.aspx?sourcedoc=%7BFAED6C19-BA41-4F03-A148-6C8490228A1C%7D&file=HL7%20IN%20vs%20HL7%20OUT%20_%2052_24095_04204575_5120_DIFFERENCES.xlsx&action=default&mobileredirect=true)
- #10733


prime-router/src/main/kotlin/cli/ProcessFhirCommands.kt
```diff
@@ -127,7 +128,11 @@ class ProcessFhirCommands : CliktCommand(
      * @return a FHIR bundle that represents the data in the one HL7 message
      */
     private fun convertToFhir(hl7String: String, actionLogger: ActionLogger): Bundle {
-        val messages = HL7Reader(actionLogger).getMessages(hl7String)
+        val hasFiveEncodingChars = hl7String.contains("MSH|^~\\&#|")
```
jimmyfagan, line 131: This logic makes an assumption that if the batch message contains 1 `MSH|^`\&#`, then all of the messages in the batch have that. In an edge case where we have a batch with a mixture of `MSH|^`\&#` and `MSH|^`\&` and we're wanting to translate one of the `MSH|^`\&` messages, that message will show `MSH|^`\&#` before getting to `translate` and the resulting fhir will mistakenly include a `encoding-characters` extension with value `^`\&#` on it's MessageHeader. This is low impact since it's only in the CLI, and it's likely a valid assumption in most cases. Still in the process of writing this comment, I thought of a different way to do this that might not add too much complexity. Stay tuned for a follow-up commit.

prime-router/src/main/kotlin/azure/ReportFunction.kt
```diff
@@ -150,21 +151,22 @@ class ReportFunction(
...
+                        sender.allowDuplicates
 
                 // Only process the report if we are not checking for connection or validation.
                 if (option != Options.CheckConnections && option != Options.ValidatePayload) {
                     val receiver = SubmissionReceiver.getSubmissionReceiver(sender, workflowEngine, actionHistory)
-
+                    val content =
+                        if (receiver is UniversalPipelineReceiver) validatedRequest.content
+                        // removes incoming '#' if included in separation characters
+                        else validatedRequest.content.replace("|^~\\&#", "|^~\\&")
```
mkalish, line 168: I'm not 100% understanding this change?  Is this being added out of an abundance of caution?

prime-router/src/main/kotlin/cli/ProcessFhirCommands.kt
jimmyfagan: Resolved this in my latest commit.

prime-router/src/main/kotlin/cli/ProcessFhirCommands.kt
```diff
@@ -139,9 +142,28 @@ class ProcessFhirCommands : CliktCommand(
...
 
+    /**
+     * @return true if a message header (either the one at hl7ItemIndex or the first one if hl7ItemIndex is null) in the
+     * given string contains MSH-2 of `^~\&#`, false otherwise
+     */
+    private fun hl7MessageHasFiveEncodingChars(hl7String: String): Boolean {
+        // This regex should match `MSH|^~\&|` or `MSH|^~\&#`
+        val mshStarts = "MSH\\|\\^~\\\\\\&[#|]".toRegex().findAll(hl7String)
+        mshStarts.forEachIndexed { index, matchResult ->
+            if (hl7ItemIndex == null || hl7ItemIndex == index) {
```
mkalish, line 160: Should this short circuit if index is null?

prime-router/src/main/kotlin/fhirengine/engine/FHIRTranslator.kt
```diff
@@ -237,4 +239,23 @@ class FHIRTranslator(
...
+fun Message.encodePreserveEncodingChars(): String {
+    // get encoding characters ...
+    val msh = this.get("MSH") as Segment
+    val encCharString = Terser.get(msh, 2, 0, 1, 1)
+    val hasFiveEncodingChars = encCharString == "^~\\&#"
+    if (hasFiveEncodingChars) Terser.set(msh, 2, 0, 1, 1, "^~\\&")
+    var encodedMsg = encode()
+    if (hasFiveEncodingChars) {
+        encodedMsg = encodedMsg.replace("^~\\&", "^~\\&#")
+        Terser.set(msh, 2, 0, 1, 1, "^~\\&#")
```
mkalish, line 258: Is this line required?  Or just being careful incase the caller doesn't use the returned `encodedMsg`?

mkalish (commented):
Couple of question, but great work!  I'll get to testing it locally.

prime-router/metadata/fhir_mapping/hl7/resource/MessageHeader.yml
```diff
@@ -78,3 +78,18 @@ extension_3:
         system: "http://terminology.hl7.org/CodeSystem/v3-AcknowledgementCondition"
       vars:
         code: MSH.16
+
+extension_4:
+  condition: $msh2 NOT_NULL && $msh2 NOT_EQUALS_STRING ^~\&
```
arnejduranovic, line 83: Why not just set this always if NOT_NULL? My concern is that the spec allows you to pick any characters here, and the ones you are checking against are just the recommended ones? `Recommended values are ^~\& (ASCII 94, 126, 92, and 38, respectively). See Section 2.5.4, "Message delimiters'.`

prime-router/metadata/hl7_mapping/OML_O21/base/service-request-order.yml
```diff
@@ -9,11 +9,12 @@ elements:
   - name: order-status
     value: [ '%resource.status' ]
     valueSet:
-      revoked: CA
-      completed: CM
-      entered-in-error: ER
-      on-hold: HD
-      active: IP
+      values:
```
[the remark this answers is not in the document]
jimmyfagan: If it's used anywhere at the moment, it's in staging, and I don't believe the sender it would be tied to is actively sending anything.

prime-router/src/main/kotlin/azure/ReportFunction.kt
arnejduranovic: I'm also curious about this, but maybe for a different reason. If we take my suggestion above and always copy the value of MSH to our FHIR extension, then we can just always right the MSH info from our FHIR extension to the output HL7v2 automatically and not need any extra logic like this?

prime-router/src/main/kotlin/cli/ProcessFhirCommands.kt
```diff
@@ -139,9 +142,28 @@ class ProcessFhirCommands : CliktCommand(
         // if a hl7 parsing failure happens, throw error and show the message
         if (message.toString().lowercase().contains("failed"))
             throw CliktError("HL7 parser failure. $message")
+        if (hasFiveEncodingChars) {
```
arnejduranovic, line 145: See other comment questioning need for extra logic

arnejduranovic (changes requested):

arnejduranovic (approved):

prime-router/src/main/kotlin/fhirengine/engine/FHIRTranslator.kt
```diff
@@ -237,4 +239,23 @@ class FHIRTranslator(
...
         return this
     }
+}
+
+/**
+ * Encodes a message while avoiding an error when MSH-2 is five characters long
+ *
+ * @return the encoded message as a string
+ */
+fun Message.encodePreserveEncodingChars(): String {
```
arnejduranovic, line 249: We discussed making this function more generic and having it take in params, but agreed there does not seem to be another clear use case for this. We can revisit this idea in the future if it comes up.

arnejduranovic (approved):

prime-router/src/main/kotlin/fhirengine/engine/FHIRTranslator.kt
```diff
@@ -237,4 +239,23 @@ class FHIRTranslator(
...
+    // get encoding characters ...
+    val msh = this.get("MSH") as Segment
+    val encCharString = Terser.get(msh, 2, 0, 1, 1)
+    val hasFiveEncodingChars = encCharString == "^~\\&#"
```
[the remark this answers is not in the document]
jimmyfagan: In my next commit, I'll use a const for these, at least in this file. Might stop short of using the consts everywhere in the code they could be used (for instance in the CLI processing).

prime-router/metadata/fhir_transforms/senders/CDC-ELIMS/elims-sender-transform.yml
```diff
@@ -1,5 +1,11 @@
 extends: ../default-sender-transform
 elements:
+  - name: sending-application-namespace-id
```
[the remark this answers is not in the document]
jimmyfagan: @arnejduranovic any thoughts on this based on your conversation with ELIMS? Is this basically an exception we're only making for them?

prime-router/src/main/kotlin/azure/ReportFunction.kt
jimmyfagan: This line intends to preserve existing behavior in this path for the legacy pipeline. The code to handle the five-char encoding is only in the Universal Pipeline, so without this, any five-char encoding strings sent to the legacy pipeline will lead to failures.

prime-router/src/main/kotlin/cli/ProcessFhirCommands.kt
jimmyfagan: Tweaked this to make it more readable.

prime-router/src/main/kotlin/fhirengine/engine/FHIRTranslator.kt
jimmyfagan: Added a comment for clarification.

prime-router/metadata/fhir_transforms/senders/CDC-ELIMS/elims-sender-transform.yml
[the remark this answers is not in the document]
victor-chaparro: I think we should change the default sender transform to set this field only if `MessageHeader.source.name` is not present. That's the way it works in the Covid pipeline.

prime-router/metadata/fhir_mapping/hl7/resource/MessageHeader.yml
victor-chaparro, line 83: Will these characters always appear in this order?

prime-router/src/main/kotlin/cli/ProcessFhirCommands.kt
```diff
@@ -127,7 +128,9 @@ class ProcessFhirCommands : CliktCommand(
      * @return a FHIR bundle that represents the data in the one HL7 message
      */
     private fun convertToFhir(hl7String: String, actionLogger: ActionLogger): Bundle {
-        val messages = HL7Reader(actionLogger).getMessages(hl7String)
+        val hasFiveEncodingChars = hl7MessageHasFiveEncodingChars(hl7String)
+        val stringToEncode = hl7String.replace("MSH|^~\\&#|", "MSH|^~\\&|")
+        val messages = HL7Reader(actionLogger).getMessages(stringToEncode)
         if (messages.isEmpty()) throw CliktError("No HL7 messages were read.")
```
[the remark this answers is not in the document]
mkalish: Will add a comment

prime-router/metadata/fhir_mapping/hl7/resource/MessageHeader.yml
mkalish: It looks like this is the recommended set of characters from the [spec ](https://hl7-definition.caristix.com/v2/HL7v2.5.1/Fields/MSH.2), but it does not necessarily have to be those set of characters. 

The second part of this check is not strictly necessary as when msh2 is not populated, the FHIR->HL7 library will automatically default to `^~\&`

mkalish:
@victor-chaparro From my read it looks like the FHIR -> HL7 library will use the default encoding characters to initially create the message, but that will get overridden by the schema processing.  Is there something else I'm missing?

mkalish:
From a slack convo with @victor-chaparro:

The encoding characters  `^~\&` are recommended, but there is no requirement that those are the characters used.  I'll add a follow up ticket to research what happens when different encoding characters are used

prime-router/metadata/fhir_mapping/hl7/resource/MessageHeader.yml
mkalish: Updated to specifically only look for `^~\&` when adding this to the bundle

mkalish (approved):
Ran message 52 through and did a comparison, the only discrepancies I saw were already noted in the spreadsheet which looks like we'll handle as some follow up work.

mkalish:
@victor-chaparro This is ready to go if you wanted to take a final look.

victor-chaparro (approved):
LGTM! Thanks for addressing all the feedback.

prime-router/metadata/fhir_transforms/senders/CDC-ELIMS/elims-sender-transform.yml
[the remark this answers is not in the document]
mkalish: I'll create a follow up ticket for this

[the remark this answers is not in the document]
mkalish: #10770 created

PR #10694, opened 2023-07-31
Jessica/10370 orm mapping

JessicaWNava:
This PR ...

Adds/updates the necessary resource files to get the information from an ORM_O01 HL7 V2 message to a FHIR message. 

Test Steps:
1. Reach out to me (Jessica) to get the sample ORM.hl7 message
2. Run something like ./prime fhirdata -i prime-router/src/testIntegration/resources/datatests/HL7_to_FHIR/orm.hl7 -o prime-router/src/testIntegration/resources/datatests/HL7_to_FHIR/orm.fhir  --output-format FHIR to convert it to FHIR
3. Check that the HL7 fields all make it over to the FHIR message.

## Changes
Updated resources to have the changes needed to transfer all ORM data to FHIR

### Testing
- [x] Tested locally?
- [ ] Added tests?


prime-router/metadata/fhir_mapping/hl7/resource/MessageHeader.yml
```diff
@@ -78,3 +76,17 @@ extension_3:
...
+  expressionsMap:
+    url:
+      type: SYSTEM_URL
+      value: message-sent
+    valueDateTime:
+      type: STRING
+      valueOf: "GeneralUtils.dateTimeWithZoneId(dateTimeIn,ZONEID)"
+      expressionType: JEXL
+      vars:
+        dateTimeIn: MSH.7
```
victor-chaparro, line 92: We are already storing the MSH.7 inside Bundle.timestamp. That logic is inside `HL7toFhirTranslator.enhanceBundleMetadata`. We need to decide if we need to keep that logic or keep this extension and then update the FHIR to HL7 mapping to grab the value from either the extension or  Bundle.timestamp. Also, I don't see why we can't have a mapping to  set Bundle.timestamp? Maybe because it's not under a resource

prime-router/metadata/fhir_mapping/hl7/resource/ServiceRequest.yml
```diff
@@ -104,6 +104,13 @@ identifier_4:
     code: "PGN"
     display: "Placer Group Number"
 
+
+identifier_5:
+  valueOf: datatype/Identifier_Gen
+  generateList: true
+  expressionType: resource
+  specs: ORC.21
+
```
victor-chaparro, line 113: Do we need both the identifier and the extension under PractitionerRole for ORC.21?

prime-router/src/main/kotlin/SubmissionReceiver.kt
```diff
@@ -369,7 +371,7 @@ class ELRReceiver : SubmissionReceiver {
 
         // TODO: This may need to be a configurable value in the future, if we ever support message types other
         //  than ORU_RO1. As of 6/15/2022 multiple message type support is out of scope
-        if (messageType != "ORU_R01") {
+        if (messageType != "ORU_R01" && messageType != "ORM_O01") {
```
victor-chaparro, line 374: Can we turn this into an Enum?

prime-router/src/main/kotlin/fhirengine/utils/HL7Reader.kt
```diff
@@ -91,5 +92,40 @@ class HL7Reader(private val actionLogger: ActionLogger) : Logging {
...
         }
+
+        fun getMessageType(message: Message): String? {
+            return (message["MSH"] as MSH).msh9_MessageType.msg1_MessageCode.toString()
+        }
+
+        fun isBirthTime(message: Message): Boolean {
+            val patient = getORMPatient(message) ?: return false
+            val dateTimeOfBirth = patient.pid.dateTimeOfBirth ?: return false
+            return dateTimeOfBirth.ts1_Time.toString().length > 8
```
victor-chaparro, line 103: It might be easier to get the birth time using the Terser and avoid having to get the patient and have all these checks in place.
I think this should work:
```
val pid = message.get("PID") as Segment
val dateTimeOfBirth = Terser.get(pid, 7, 0, 1, 1)
```

victor-chaparro (changes requested):
This is breaking some of the integration tests we have in place already. I think it's related to the CodeableConcept change. We don't have integration test for the ELIMS test messages Jimmy has been using, but we should probably check these changes don't break what he has been working on. Is there a follow up ticket to update the FHIR to HL7 mappings based on these changes?

prime-router/metadata/fhir_mapping/hl7/resource/MessageHeader.yml
JessicaWNava: I added this because I was not seeing it as a dateTime in the message. It looks like it is for some reason being put as a string for a visit number in the identifier for the ServiceRequest? That doesn't seem right to me. 

{
        "type" : {
          "coding" : [ {
            "system" : "http://terminology.hl7.org/CodeSystem/v2-0203",
            "code" : "VN",
            "display" : "Visit number"
          } ]
        },
        "value" : "20230506052916-0500"
      }

prime-router/metadata/fhir_mapping/hl7/resource/Observation.yml
```diff
@@ -342,3 +342,69 @@ extension_4:
...
+extension_5:
+  condition: $value NOT_NULL
+  generateList: true
+  valueOf: extension/Extension
+  expressionType: resource
+  vars:
+    value: String, OBX.11
+  constants:
+    KEY_NAME_SUFFIX: String
+    urlValue: observation-status-original-text
```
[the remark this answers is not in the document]
halprin: I believe we do.  There are a number of possible values of `OBX.11` other than `X` or `N` that [do not have mappings](https://docs.google.com/spreadsheets/d/1RsA6SEYsjfrR6TbwY1JHBetrPyRED2RDzTa1jvHRvX0/edit#gid=0).  For example, `O` doesn't map to anything.  If we don't save this value in this extension (or some other way), we can't convert the FHIR back to HL7v2 and know what goes in `OBX.11`.

prime-router/metadata/fhir_mapping/hl7/resource/ServiceRequest.yml
JessicaWNava: No, I can remove the extension, it did not seem to be working anyways. Good catch!

prime-router/src/main/kotlin/fhirengine/utils/HL7Reader.kt
JessicaWNava: Cool that cleans things up quite a bit, thank you!

JessicaWNava: So, with ORM it is not always PID which is why it is necessary to check for either the PATIENT or the PATIENT_RESULT.

JessicaWNava:
> This is breaking some of the integration tests we have in place already. I think it's related to the CodeableConcept change. We don't have integration test for the ELIMS test messages Jimmy has been using, but we should probably check these changes don't break what he has been working on. Is there a follow up ticket to update the FHIR to HL7 mappings based on these changes?

Yes, I was running into merge issues with some of the front end changes when I tried to pull the changes in from master to this branch. Flexion resolved these issues now. 

Yes, I believe that the OML ticket is intended to do that part: 
https://app.zenhub.com/workspaces/platform-6182b02547c1130010f459db/issues/gh/cdcgov/prime-reportstream/10310 although, now that I am reading this ticket, maybe I am misunderstanding @arnejduranovic ?

JessicaWNava:
Back to you @victor-chaparro

JessicaWNava:
I see what you mean about the integration tests @victor-chaparro, I will address those next

prime-router/metadata/fhir_mapping/hl7/resource/ServiceRequest.yml
```diff
@@ -313,18 +322,6 @@ extension_7:
       valueOf: OBR.20
       expressionType: HL7Spec
 
-extension_8:
```
JessicaWNava, line 316: @victor-chaparro removing this extension breaks the integration tests. I can easily replace the data in OBR.21 is lost.

prime-router/metadata/fhir_mapping/hl7/resource/Observation.yml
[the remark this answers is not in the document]
victor-chaparro: I see why we need the extension now. It's going to be confusing to whoever is ingesting this data to know from where to get the value from, but I don't know if there's anything we can do.

prime-router/metadata/fhir_mapping/hl7/resource/ServiceRequest.yml
JessicaWNava: Wrong extension removed, will replace.

prime-router/metadata/fhir_mapping/hl7/resource/MessageHeader.yml
JessicaWNava: Discussed and will remove. This is getting sent over it just did not look like it via search because it was being converted into another timezone.

prime-router/src/main/kotlin/fhirengine/utils/HL7Reader.kt
JessicaWNava: So I tried what we discussed this morning and `val pid = message.get("PATIENT") as Segment` results in `class ca.uhn.hl7v2.model.v251.group.ORM_O01_PATIENT cannot be cast to class ca.uhn.hl7v2.model.Segment ` so basically it doesn't eliminate the problem you were trying to solve because it will still need to differentiate between the message types.

victor-chaparro (changes requested):
Just noticed those 3 extensions should go under Specimen, it would also be good to add an integration test for HL7->FHIR with the sample file you've been using.

prime-router/metadata/fhir_mapping/hl7/resource/Observation.yml
```diff
@@ -342,3 +342,69 @@ extension_4:
...
+  condition: $value NOT_NULL
+  generateList: true
+  valueOf: extension/Extension
+  expressionType: resource
+  vars:
+    value: String, OBR.15.3
+  constants:
+    KEY_NAME_SUFFIX: String
+    urlValue: specimen-collection-method
+
```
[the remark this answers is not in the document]
JessicaWNava: Done! Also added an integration test. Back to you @victor-chaparro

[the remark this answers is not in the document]
basiliskus: We overlooked this, but we're actually already mapping `OBR.15.1` in the specimen resource under `type` (line 27 of Specimen.yml). Our mapping was done based on the expected fhir format that we have given there is no OBR specimen mapping for [ORM_O01](https://docs.google.com/spreadsheets/d/19C2jjzXZ8D7Um1iB-e6NmLfZVPvyZNkgwU8IxCJiwGs/edit#gid=0). Others in my team may be able to provide more context

victor-chaparro (approved):
Looks good! Thanks for addressing all the feedback!

JessicaWNava:
Thank you so much for your super thorough review and all of your help @victor-chaparro! Your turn @arnejduranovic :)

prime-router/metadata/fhir_mapping/hl7/datatype/Quantity.yml
```diff
@@ -1,5 +1,11 @@
 value_1 :
   condition: $value CONTAINS_STRING '.'
+  #  Not specifying a type results in the full "value" being reported correctly and no rounding/float conversion occurs.
+  #  However, it also breaks the integration tests, so we are going to have to duplicate this value into an extension
+  #  until we LinuxForHealth provides a permanent solution.
```
arnejduranovic, line 5:
```suggestion
  #  until LinuxForHealth provides a permanent solution.
```

prime-router/metadata/fhir_mapping/hl7/datatype/Quantity.yml
```diff
@@ -1,5 +1,11 @@
 value_1 :
   condition: $value CONTAINS_STRING '.'
+  #  Not specifying a type results in the full "value" being reported correctly and no rounding/float conversion occurs.
+  #  However, it also breaks the integration tests, so we are going to have to duplicate this value into an extension
```
arnejduranovic, line 4: How does it break the integration tests? Why can't we just change the integration tests?

arnejduranovic (changes requested):

prime-router/src/main/kotlin/SubmissionReceiver.kt
```diff
@@ -369,7 +376,7 @@ class UniversalPipelineReceiver : SubmissionReceiver {
 
         // TODO: This may need to be a configurable value in the future, if we ever support message types other
         //  than ORU_RO1. As of 6/15/2022 multiple message type support is out of scope
-        if (messageType != "ORU_R01") {
+        if (messageType != MessageType.ORU_R01.toString() && messageType != MessageType.ORM_O01.toString()) {
```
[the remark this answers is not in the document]
JessicaWNava: Ticket added here: https://app.zenhub.com/workspaces/platform-6182b02547c1130010f459db/issues/gh/cdcgov/prime-reportstream/10848

prime-router/src/main/kotlin/fhirengine/translation/HL7toFhirTranslator.kt
```diff
@@ -105,4 +111,37 @@ class HL7toFhirTranslator internal constructor(
...
+     *  fhir documentation https://hl7.org/fhir/json.html#primitive, if a value has an id attribute or extension,
+     *  it is represented with an underscore before the name. Currently, it seems hl7v2-fhir-converter library does not
+     *  support this, so this method is a workaround to add an extension to birthDate. There is also no support for
+     *  getting the length of the field, for which this issue was created:
+     *  https://github.com/LinuxForHealth/hl7v2-fhir-converter/issues/499
+     *  This method looks in the [hl7Message] for the birthdate and add an extension to the [bundle] if it includes
+     *  the time
+     */
+
+    private fun handleBirthTime(bundle: Bundle, hl7Message: Message) {
```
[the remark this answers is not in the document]
arnejduranovic: I think `BundleEnhancer` or something like that is a better name for the class I suggested fwiw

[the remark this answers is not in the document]
arnejduranovic: Also, we are currently looking into forking the LinuxForHealth HL7->FHIR conversion library, and this change may be more appropriate to make there.

prime-router/metadata/fhir_mapping/hl7/datatype/Quantity.yml
basiliskus: @JessicaWNava Agreed, it is an intrusive workaround that may have unintended side effects and we're Ok with finding another solution for what we need. We have an open PR in the `LinuxForHealth/hl7v2-fhir-converter` repo for a permanent fix, but in the meantime I think we're Ok with leaving it as is

prime-router/src/main/kotlin/fhirengine/translation/HL7toFhirTranslator.kt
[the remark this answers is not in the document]
JessicaWNava: Ticket created here: https://app.zenhub.com/workspaces/platform-6182b02547c1130010f459db/issues/gh/cdcgov/prime-reportstream/10849

JessicaWNava:
Feedback addressed, back to you @arnejduranovic

arnejduranovic (approved):

PR #10716, opened 2023-08-01
Experience: vite preview

### LINKED ISSUE #10715: Experience: vite preview
The project needs to implement a `preview` script for vite's `preview` command. The best action to take when executed would be to set `NODE_ENV` to `preview`, run the build script, and then run the vite preview command. A new `.env.preview` file will need to be made using the appropriate ports in URLs (4173). The preview server should also serve the same `Content-Security-Policy` header as the live site.

jpandersen87:
Fixes #10715

This PR:
- adds a new `preview` script that builds and runs `vite preview` under the `preview` environment
- adds a new `.env.preview` file
- adds the current CSP header given on the live site to the preview server
  - hhs-prime.okta.com swapped to hhs-prime.oktapreview.com


penny-lischer (approved):
LGTM

PR #10718, opened 2023-08-01
Missing autoprefixer

### LINKED ISSUE #10717: Missing autoprefixer
With the vite migration, autoprefixing on our CSS files was lost. The `autoprefixer` dev dependency needs to be added, and a `postcss.config.cjs` configuration file created using it.

jpandersen87:
Fixes #10717

This PR adds the missing `autoprefixer` dev dependency that was lost during vite migration (for browser autoprefixing on CSS).

### DIFF
--- a/frontend-react/postcss.config.cjs
+++ b/frontend-react/postcss.config.cjs
@@ -1,0 +1,5 @@
+module.exports = {
+    plugins: {
+        autoprefixer: {},
+    },
+};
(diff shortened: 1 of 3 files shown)


frontend-react/postcss.config.cjs
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+    plugins: {
+        autoprefixer: {},
```
KeithNava, line 3: do you have to add to the vite config as well?

KeithNava (approved):

PR #10723, opened 2023-08-01
Update USWDS to 3.4.1

### LINKED ISSUE #10721: Update USWDS to 3.4.1
Technically we should be able to use any USWDS version alongside react-uswds. USWDS styles being bundled with react-uswds' index.css file is unintentional (official fix pending). Utilize `patch-package` to fix it from our end for now so that we can freely change USWDS versions. The latest version is 3.5, but our designers are working off of 3.4.1 currently.

jpandersen87:
Fixes #10721

This PR:
- introduces `patch-package` dev dependency
- patches `react-uswds`@5.1.1 index.css to remove the unintended USWDS style bundling
- removes errant flex settings from uswds_settings
- updates shared Alert component styling


penny-lischer (approved):
lgtm

PR #10732, opened 2023-08-02
10774 Refactor and increase coverage for BlobAccess

jack-h-wang:
This PR increases code coverage for BlobAccess and implements reuse of Azure blob container clients.

Test Steps:
1. Confirm [code coverage report](https://sonarcloud.io/code?id=CDCgov_prime-data-hub&pullRequest=10732&selected=CDCgov_prime-data-hub%3Aprime-router%2Fsrc%2Fmain%2Fkotlin%2Fazure%2FBlobAccess.kt) is over 90%.
2. Ensure all tests run successfully.

## Changes
- Store Azure container client after creation so that subsequent calls can reuse it.
- Isolated function to return Blob connection string (needed for mocking)
- Added unit tests.

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- #10774


prime-router/src/test/kotlin/azure/BlobAccessTests.kt
jimmyfagan, line 1: I didn't realize how mixed we are about using `test some action` vs `some action` as the names of these unit test functions in the repo 😮 I guess it's just been mostly `test some action` in UP unit tests, so I found these surprising. Since it's so mixed, not recommending we change anything here, just thought it was interesting.

prime-router/src/test/kotlin/azure/BlobAccessTests.kt
```diff
@@ -52,4 +73,237 @@ class BlobAccessTests {
...
+            )
+        } returns
+            BlobAccess.BlobInfo(report1.bodyFormat, testUrl, BlobAccess.sha256Digest(testBytes))
+
+        val testBlob = BlobAccess()
+        val result = testBlob.uploadReport(report1, testBytes)
+
+        assertThat(result.format).isEqualTo(testFormat)
+        assertThat(result.blobUrl).isEqualTo(testUrl)
+        assertThat(result.digest).isEqualTo(BlobAccess.sha256Digest(testBytes))
```
jimmyfagan, line 105: Might be able to make this a little more concise by making a BlobInfo object in this test function, using that in the `every {} returns blobInfo`, and then comparing the full result object instead of comparing each field individually.

prime-router/src/test/kotlin/azure/BlobAccessTests.kt
```diff
@@ -52,4 +73,237 @@ class BlobAccessTests {
...
+        mockkClass(BlobAccess::class)
+        mockkObject(BlobAccess.Companion)
+        every { BlobAccess.Companion.getBlobConnection(testEnv) } returns "testconnection"
+        every { BlobAccess.Companion.uploadBlob(any(), testBytes) } returns testUrl
+
+        testEvents.forEach {
+            val result = when (it) {
+                null -> BlobAccess.uploadBody(testFormat, testBytes, testName, "")
+                Event.EventAction.CONVERT -> BlobAccess.uploadBody(testFormat, testBytes, testName, action = it)
+                else -> BlobAccess.uploadBody(testFormat, testBytes, testName, testFolder, it)
```
jimmyfagan, line 137: Is this arbitrarily to just cover both null and non-null `testFolder` values? If so, a one-line comment to mention that would be helpful for folks reading this test without the context of code coverage in mind.

prime-router/src/test/kotlin/azure/BlobAccessTests.kt
```diff
@@ -52,4 +73,237 @@ class BlobAccessTests {
...
+            Event.EventAction.TRANSLATE,
+            Event.EventAction.NONE,
+            Event.EventAction.CONVERT,
+            null
+        )
+
+        mockkClass(BlobAccess::class)
+        mockkObject(BlobAccess.Companion)
+        every { BlobAccess.Companion.getBlobConnection(testEnv) } returns "testconnection"
+        every { BlobAccess.Companion.uploadBlob(any(), testBytes) } returns testUrl
```
jimmyfagan, line 131: Instead of using `any()` here and having them all return the same url, it might be nice to verify that each action actually results in the expected bloburl, by having a different `every { uploadBlob(<expected param for action>, testBytes) } returns <action-specific-url>` for each action. Makes this a bit more verbose, but I think it would help this feel like it's a productive test instead of a test purely for code coverage.

prime-router/src/test/kotlin/azure/BlobAccessTests.kt
```diff
@@ -52,4 +73,237 @@ class BlobAccessTests {
...
+        every { anyConstructed<BlobClientBuilder>().buildClient() } returns mockedBlobClient
+
+        val result = BlobAccess.uploadBlob(testName, testBytes, testContainer, testEnv)
+        val result2 = BlobAccess.uploadBlob(testName, testBytes, testContainer, testEnv)
+
+        verify(exactly = 2) { mockedBlobClient.upload(any<ByteArrayInputStream>(), testBytes.size.toLong()) }
+        assertThat(result).isEqualTo("testurlname")
+        assertThat(result2).isEqualTo("testurlname")
+
+        unmockkAll()
```
jimmyfagan, line 182: It might be best to add a `@BeforeAll` or `@AfterAll` function to this class that calls this, it would help make sure future test cases in this file don't make the mistake of forgetting to unmock.

jimmyfagan (commented):
Some of the tests feel so mock-heavy that they're basically pointless aside from increasing code coverage. It's kind of like we're testing that the function calls all these other functions instead of testing any of the functionality. I wonder if 90% was too high of a bar for this file, but a 90% mark does necessitate some of that, so with that in mind this looks pretty good, just left a few comments for optional improvements.

prime-router/src/test/kotlin/azure/BlobAccessTests.kt
jack-h-wang: For part of the test I wanted to simulate the actual data flow as much as possible, thus the mocked BlobInfo object being returned being constructed using the format from the input report instead of a prebuilt BlobInfo. I'm not really sure if anything is actually being tested if I did otherwise. I tried adjusting the comparison to a BlobInfo object built directly from the test variables, but apparently Kotlin tests equality by reference this way so the assertion fails; I could test the hashcodes of each instance of the data class instead, but testing the elements directly seemed more immediately readable.

prime-router/src/test/kotlin/azure/BlobAccessTests.kt
jack-h-wang: I switched this test to capture the blobName and had the mock return it as the URL so that we could test the value is as expected.

prime-router/src/test/kotlin/azure/BlobAccessTests.kt
jack-h-wang: I added an `@AfterEach` function to be executed in lieu of explicitly calling `unmockkAll()` after each test.

prime-router/src/test/kotlin/azure/BlobAccessTests.kt
jack-h-wang: I added a comment to clarify this.

prime-router/src/test/kotlin/azure/BlobAccessTests.kt
```diff
@@ -128,17 +134,29 @@ class BlobAccessTests {
         mockkClass(BlobAccess::class)
         mockkObject(BlobAccess.Companion)
         every { BlobAccess.Companion.getBlobConnection(testEnv) } returns "testconnection"
-        every { BlobAccess.Companion.uploadBlob(any(), testBytes) } returns testUrl
+        every { BlobAccess.Companion.uploadBlob(capture(blobSlot), testBytes) } answers
+            { "http://" + blobSlot.captured }
```
jimmyfagan, line 138: Didn't realize this was possible, very nice!

prime-router/src/test/kotlin/azure/BlobAccessTests.kt
```diff
@@ -173,13 +191,12 @@ class BlobAccessTests {
         every { anyConstructed<BlobClientBuilder>().buildClient() } returns mockedBlobClient
 
         val result = BlobAccess.uploadBlob(testName, testBytes, testContainer, testEnv)
+        // upload a second blob to the same container to test container client reuse
```
jimmyfagan, line 194: It'd be nice if this test verified that `create` is only called once. As it stands, it may cover those lines, but the test isn't actually verifying the client reuse.

prime-router/src/test/kotlin/azure/BlobAccessTests.kt
jimmyfagan: Ah right, I got the name of that wrong, glad you figured it out!

prime-router/src/test/kotlin/azure/BlobAccessTests.kt
jack-h-wang: That works, but it doesn't work if I keep the original return value of the mock. From reading up on it I think it's because it's a data class where one of the elements of the class is an Array. All the same, if you're OK with the test in its current form I think it'd be best to just leave it as is.

prime-router/src/test/kotlin/azure/BlobAccessTests.kt
jack-h-wang: Good call, added that verification.

prime-router/src/test/kotlin/azure/BlobAccessTests.kt
jimmyfagan: Yep, that's fine, and that makes sense about the Array, didn't think about that! Closing this convo.

jimmyfagan (approved):
I think all of my comments that needed to be addressed were, approved!

prime-router/src/test/kotlin/azure/BlobAccessTests.kt
```diff
@@ -52,4 +79,241 @@ class BlobAccessTests {
             assertThat(e).isNotNull()
         }
     }
+
+    @Test
+    fun `upload report`() {
```
arnejduranovic, line 84: I am unclear how this is testing anything in BlobAccess?

arnejduranovic (approved):

PR #10478, opened 2023-07-17
Settings API - Use all authentication options

luis-pabon-tf:
This PR updates the Settings API to both use server2server auth in addition to Okta auth.

Test Steps:
* Run the smoke test
* Verify the code makes sense

## Changes
- Updated `SettingsFunctions` to use both Okta and Server2Server during the authentication process
- Made all tests using copied Okta auth functions pull from a new location, with a new name to stop a function name conflict

## Linked Issues
- #1946

Issue #1946, opened 2021-08-24
Settings commands use server2server auth as well as Okta auth, following our new pattern


jimduff-usds:
Right now, the settings API functions use Okta auth only - that is, they are designed for humans to use them.

However, we now have a set of methods/functions that combine server2server and Okta auth pretty cleanly.

This story is to use those combined methods, for the settings endpoints, instead of Okta-auth-only.

Combined Authentication is done by this call:
```
val claims = AuthenticatedClaims.authenticate(request)
            ?: return HttpUtilities.unauthorizedResponse(request, authenticationFailure)
```
This will first try Okta.  If that fails, then try server2server.

Then authorization is a bit different - its more specific to the specific function - you authorize based on knowledge of what Function is called vs what `claims` you have.  An example is `claims.authorizedForSendOrReceive(sender, request))` in the first link below.

Examples of both authz and authn methods can be found in

https://github.com/CDCgov/prime-reportstream/blob/2870fe1086b5960f323c1add720f45bf85fa3e24/prime-router/src/main/kotlin/azure/ReportFunction.kt#L83

https://github.com/CDCgov/prime-reportstream/blob/2870fe1086b5960f323c1add720f45bf85fa3e24/prime-router/src/main/kotlin/history/azure/ReportFileFunction.kt#L155

There are other examples too!

So the task here is to mimic the existing calls to the combination okta/server2server in the links above, now for the Settings API endpoints.

Then add Smoke tests for server2server.    See an insane number of examples here: https://github.com/CDCgov/prime-reportstream/blob/master/prime-router/src/main/kotlin/cli/tests/AuthTests.kt

Note this very ironic twist:  to test server2server in Staging, you first need todo an authorized Okta login, so that you can use that to upload the server2server public key!

I think you'll need a new "scope" for this, maybe?

Your PR must fill out the detailed security review section.  In addition your PR must be reviewed by those on the team with security expertise, per our process.

### Acceptance Criteria

- [ ] All Settings endpoints can be accessed via either Okta auth or Server2server
- [ ] Adequate tests have been added to positively confirm that those who should not get access do not get access, and those who should get access do get access.
- [ ] The security process has been followed

jimduff-usds:
To get a sense of the usefulness of this work, see the follow-on task #1942

PR #10799, opened 2023-08-07
Allow normal users to read settings

arnejduranovic:
This PR updates the Settings API to both use server2server auth in addition to Okta auth. This is an addition to #10478 and fixes a bug where a normal org user could not read settings, breaking the daily data page.

Test Steps:
* Run the smoke test

Manual Test:
1. Log in to ReportStream as a normal user (Create a new account in Okta if needed)
2. Verify Daily Data page loads

## Changes
- Updated `SettingsFunctions` `getOne` to support getting an organization better (no longer need to pass in null for orgName, this logic was moved lower down into SettingsFacade)
- Updated getOne and other settings and api key functions to allow for user access (all the ones that are read)

## Linked Issues
- #1946

## Future Work
1. #10810


prime-router/src/main/kotlin/azure/SettingsFunctions.kt
```diff
@@ -330,9 +339,14 @@ open class BaseFunction(
         clazz: Class<T>,
         organizationName: String? = null
     ): HttpResponseMessage {
-        authenticateAdmin(request)
-            ?: return HttpUtilities.unauthorizedResponse(request, authenticationFailure)
-
+        val claims = AuthenticatedClaims.authenticate(request)
+        if (claims == null || !claims.authorized(
+                setOf(PRIME_ADMIN_PATTERN, "$settingName.*.admin", "$settingName.*.user")
```
mkalish, line 344:
```suggestion
                setOf(PRIME_ADMIN_PATTERN, "$organizationName.*.admin", "$organizationName.*.user")
```

prime-router/src/test/kotlin/azure/ApiKeysFunctionsTest.kt
```diff
@@ -766,26 +766,6 @@ class ApiKeysFunctionsTest {
             assertThat(response.getStatus()).isEqualTo(HttpStatus.UNAUTHORIZED)
         }
 
-        @Test
```
mkalish, line 769: Nit: could just modify this to allow a `user` in

prime-router/src/main/kotlin/azure/SettingsFunctions.kt
```diff
@@ -274,9 +275,13 @@ open class BaseFunction(
         clazz: Class<T>,
         organizationName: String? = null
     ): HttpResponseMessage {
-        authenticateAdmin(request)
-            ?: return HttpUtilities.unauthorizedResponse(request, authenticationFailure)
-
+        val claims = AuthenticatedClaims.authenticate(request)
+        if (claims == null ||
+            !claims.authorized(setOf(PRIME_ADMIN_PATTERN, "$organizationName.*.admin", "$organizationName.*.user"))
```
mkalish, line 280: Just noting that `organizationName` is nullable,  I don't think it will have a negative impact since the scope will be `null.*.admin` until we have an org name null.

Could consider a follow up ticket, that splits this function up between fetching all settings vs just fetching an orgs.

mkalish (commented):
Looks good, just need to test it locally.

mkalish (approved):
LGTM!

Ran the smoke tests and logged in to staging and ran the oktaauth tests too.

prime-router/src/main/kotlin/azure/SettingsFunctions.kt
arnejduranovic: This actually does not work, functionality would fail and tests would fail with this suggested change

mkalish: So I get how this would work if the setting name was an org, but how does this work when fetching a receiver or sender?

mkalish (approved):
Updates make sense and I'm cool deferring that other comment as tech debt

Ran the tests locally again and LGTM!

prime-router/src/main/kotlin/azure/SettingsFunctions.kt
arnejduranovic: Will create a new ticket

luis-pabon-tf (approved):
Changes are good

Issue #10839, opened 2023-08-10
Test CDC-ELIMS messages: Part 5 - MEGA MESSAGES!!!


mkalish:
### User Story
As ELIMS, I would like the HL7 in to match the HL7 out exactly.

### Description/Use Case
The messages currently match with the one major exception where the 2.5.1 spec differs the NIST 2.5.1 spec.  The difference is that there are many elements (example OBR.4) which are `CE` in 2.5.1 standard and `CWE` in the NIST 2.5.1 spec which means that we lose data during the process.

### Risks/Impacts/Considerations
- Potentially some receivers get additional data than they had before, but this could likely be mitigated by a receiver transform

### Dev Notes
- The HL7 specs are suppose to be 100% backwards compatible; new versions append new fields
- The NIST 2.5.1 spec matches the 2.7.1 spec in the cases where we are dropping data, so a potential solution would be to parse the message as a 2.7.1 message 

### Acceptance Criteria
- [x] Running [47_04608717_11184_Mega_Specimen.HL7](https://cdc.sharepoint.com/:u:/r/teams/ReportStream/Shared%20Documents/ELIMS-ELR/ELIMS%20artifacts/47_04608717_11184_Mega_Specimen.HL7?csf=1&web=1&e=L9ynNz) through the pipeline results in no data loss
  - There are known places where additional data is added (i.e. more precise timestamps) and those are ok

PR #10840, opened 2023-08-10
10838: elims message part 4

mkalish:
This PR enables support for mapping AOE OBX segments in the SPECIMEN group into ServiceRequest.supportingInfo and then back to HL7

Test Steps:
Using both sample messages in the linked issue
1. Update the `dev.DEV_ELIMS` receiver to be active
2. Send the message in the linked issue to  `ignore.ignore-elr-elims` sender
3. Verify that the only discrepancies between the two are noted in the spreadsheet
https://cdc.sharepoint.com/:x:/r/teams/ReportStream/Shared%20Documents/ELIMS-ELR/[47_04608717_11184_mega_specimen.xlsx](https://cdc.sharepoint.com/:x:/r/teams/ReportStream/Shared%20Documents/ELIMS-ELR/47_04608717_11184_mega_specimen.xlsx?d=w7651af5fc4944146a679a0a961f0c526&csf=1&web=1&e=D97aMQ)?d=w7651af5fc4944146a679a0a961f0c526&csf=1&web=1&e=D97aMQ

## Changes
- Adds mappings for SPECIMEN.OBX -> ServiceRequest.supportingInfo
- Adds mapping for ServiceRequest.supportingInfo -> SPECIMEN.OBX
- Fixes an issue in the FHIR -> HL7 library that didn't support multiple elements appearing in the schema twice

### Testing
- [x] Tested locally?
- [x] Added tests?

### Process

## Linked Issues
- #10838 

## To Be Done
- #10839


prime-router/metadata/fhir_mapping/hl7/message/ORU_R01.yml
```diff
@@ -61,6 +61,21 @@ resources:
       - .OBSERVATION.NTE
       - MSH
 
+  - resourceName: ServiceRequest
```
mkalish, line 64: This had to be pulled out so that all the OBX segments from the specimen were passed in

prime-router/metadata/hl7_mapping/ORU_R01/base/observation-result.yml
```diff
@@ -1,7 +1,3 @@
-constants:
```
mkalish, line 1: These are now vars that need to be passed in

prime-router/src/main/kotlin/fhirengine/translation/hl7/schema/ConfigSchema.kt
```diff
@@ -126,12 +126,10 @@ abstract class ConfigSchema<T : ConfigSchemaElement>(
             // Why the distinct? A schema can make references to the same schema multiple times, so you could get
             // a list of elements that are identical, so we make sure to get only those that at different.
             elementsInSchema = elements.filter { it.schemaRef != null }.mapNotNull {
-                it.schemaRef?.findElement(elementName)
-            }.distinct()
+                it.schemaRef?.findElements(elementName)
+            }.flatten().distinct()
         }
-        // Sanity check
-        check(elementsInSchema.size <= 1)
```
mkalish, line 133: The need for this restriction was unclear to me and now that an observation-result can appear in two places was actually causing problems.  The California transform overrides `observation-note` so it is always false.  Now that `observation-result` is used in `order-observation` and `specimen` this code failed as `observation-note` now appeared twice in the schema.

The change was to return all the found elements and merge the override into all of them.  This made sense in the example above, where the assumption would be that California wants `observation-note` suppressed everywhere.

prime-router/src/main/kotlin/fhirengine/translation/hl7/schema/converter/ConverterSchema.kt
```diff
@@ -22,6 +22,11 @@ class ConverterSchema(
     constants: SortedMap<String, String> = sortedMapOf(),
     extends: String? = null
 ) : ConfigSchema<ConverterSchemaElement>(elements = elements, constants = constants, extends = extends) {
+
+    override fun toString(): String {
```
mkalish, line 26: While debugging I got stick of seeing `...ConverterSchema@194fa1`, so I had a toString to make it easier to understand what element you're looking at.

prime-router/src/testIntegration/resources/datatests/HL7_to_FHIR_to_HL7/elims_47_04608646_11024_mega_specimen-expected.hl7
mkalish, line 1: File rename, I prefixed the filename with `elims_`

prime-router/src/main/kotlin/fhirengine/translation/hl7/schema/ConfigSchema.kt
jimmyfagan: What you're describing makes sense, but it concerns me a little bit when  I think about what happens when the element that appears twice isn't overridden. There are checks for duplicate elements in the convert and translate steps (search for usages of `duplicateElements`), would those fail if `observation-result` for example isn't overridden? Coming off of vacation, I'm having a hard time wrapping my mind around the details of schema extension vs schema reference within an element, so it might not be a real concern, and I can probably test it on my own too.

jimmyfagan: Also, if it is a problem, we might be able to fix that by just not checking for "duplicate elements" in those steps (duplicates are really just elements with the same name). It almost seems like this was enforced in schema merging and in the other steps to meet an arbitrary assumption that names would be unique in a schema tree. You point out a good reason to allow named to be repeated, and I can't think of a great reason not to allow that.

prime-router/metadata/fhir_mapping/hl7/message/ORU_R01.yml
jimmyfagan: All of the reference documents from Carlos (like [this one](https://lucid.app/lucidchart/dc67748e-b67e-4224-80e6-27e26d6359c1/edit?invitationId=inv_5f0d1789-cbf2-43b5-b0b1-adc762fecd08&page=0_0#)) show that ServiceRequests only exist within a Diagnostic Report. Don't know if those documents are more of a reflection of the way we currently lay things out, or if it was tied to the FHIR specs. It may just be that we've never had input messages that necessitated this kind of change and that this is all good, but I haven't done my research really, so was wondering if you had any thoughts on that chart based on what you've looked into.

jimmyfagan (commented):
I think this probably makes sense, but I'm slow wrapping my head around all the aspects of this, so I'll need to finish my review next week. Left a few comments with some questions/concerns in the meantime, open to huddling on this next week to make sure I understand more of the context/needs for these changes too.

prime-router/src/main/kotlin/fhirengine/translation/hl7/schema/ConfigSchema.kt
mkalish: I think error wise this actually works as expected as only the California translation overrides `observation-result` but the base one does not.  I can take at those usages and try to see why that isn't causing an error.

prime-router/metadata/fhir_mapping/hl7/message/ORU_R01.yml
mkalish: So even before this change the ServiceRequest was still an object in resources rather than something nested in DiagnosticReport and it was referenced by DiagnosticReport, this change just means that we make it before rather than having it be created when we make the diagnostic report.

I can leave a comment though that makes that more obvious

jimmyfagan: Ah okay that makes sense, I don't think a comment is needed, I think I just need to look more at the actual resulting bundle instead of just looking at changed files.

prime-router/src/testIntegration/resources/datatests/HL7_to_FHIR_to_HL7/elims_47_04608717_11184_mega_specimen.hl7
```diff
@@ -0,0 +1,20 @@
+MSH|^~\&#|STARLIMS.CDC.Stag^2.16.840.1.114222.4.3.3.2.1.2^ISO|CDC Atlanta^11D0668319^CLIA|CDPH CA CALREDIE^2.16.840.1.114222.4.3.3.10.1.1^ISO|CDPH_CID^2.16.840.1.114222.4.1.2.14104^ISO|20230802180802-0400||ORU^R01^ORU_R01|3015894743_04608717_11184|T|2.5.1|||NE|NE|USA||||PHLabReport-NoAck^PHIN^2.16.840.1.113883.9.11^ISO
+SFT|CDC^^^^^CDC&2.16.840.1.114222.4&ISO^XX^^^CDC CLIA|ELIMS V11|STARLIMS|Binary ID unknown
+PID|1||PID03707032^^^STARLIMS.CDC.Stag&2.16.840.1.114222.4.3.3.2.1.2&ISO^PI~PID123^^^SPHL-000048&2.16.840.1.114222.4.1.10765&ISO^PI||Mega^HL7^MI^^^^L||0000|F|||^^^^^USA^H
```
jimmyfagan, line 3: PID-7 here is `0000` and is not preserved in the output, seems like an invalid value we're getting, but just to check, has that been either discussed with ELIMS or added to the spreadsheet?

prime-router/src/testIntegration/resources/datatests/translation-test-config.csv
```diff
@@ -14,8 +14,9 @@ HL7_to_FHIR/sample_co_2_20220803-0001.hl7,,HL7_to_FHIR/sample_co_2_20220803-0001
...
-HL7_to_FHIR_to_HL7/47_32361_04608646_11034_mega_case.hl7,,HL7_to_FHIR_to_HL7/47_32361_04608646_11034_mega_case-expected.hl7,metadata/hl7_mapping/ORU_R01/ORU_R01-base,HL7,PASS,"","","metadata/fhir_transforms/senders/CDC-ELIMS/elims-sender-transform"
+HL7_to_FHIR_to_HL7/elims_47_04608646_11024_mega_specimen.hl7,,HL7_to_FHIR_to_HL7/elims_47_04608646_11024_mega_specimen-expected.hl7,metadata/hl7_mapping/ORU_R01/ORU_R01-base,HL7,PASS,"","","metadata/fhir_transforms/senders/CDC-ELIMS/elims-sender-transform"
+HL7_to_FHIR_to_HL7/elims_47_32361_04608646_11034_mega_case.hl7,,HL7_to_FHIR_to_HL7/elims_47_32361_04608646_11034_mega_case-expected.hl7,metadata/hl7_mapping/ORU_R01/ORU_R01-base,HL7,PASS,"","","metadata/fhir_transforms/senders/CDC-ELIMS/elims-sender-transform"
+HL7_to_FHIR_to_HL7/elims_47_04608717_11184_mega_specimen.hl7,,HL7_to_FHIR_to_HL7/elims_47_04608717_11184_mega_specimen-expected.hl7,metadata/hl7_mapping/ORU_R01/ORU_R01-base,HL7,PASS,"","","metadata/fhir_transforms/senders/CDC-ELIMS/elims-sender-transform"
```
jimmyfagan, line 19: Took me a minute to recognize that we don't have any duplicates here, the names are just similar until pretty far down the string. Completely optional, but how bout something like this? Would require the files to be renamed as well.
```suggestion
HL7_to_FHIR_to_HL7/elims_47_1_04608646_11024_mega_specimen.hl7,,HL7_to_FHIR_to_HL7/elims_47_1_04608646_11024_mega_specimen-expected.hl7,metadata/hl7_mapping/ORU_R01/ORU_R01-base,HL7,PASS,"","","metadata/fhir_transforms/senders/CDC-ELIMS/elims-sender-transform"
HL7_to_FHIR_to_HL7/elims_47_2_32361_04608646_11034_mega_case.hl7,,HL7_to_FHIR_to_HL7/elims_47_2_32361_04608646_11034_mega_case-expected.hl7,metadata/hl7_mapping/ORU_R01/ORU_R01-base,HL7,PASS,"","","metadata/fhir_transforms/senders/CDC-ELIMS/elims-sender-transform"
HL7_to_FHIR_to_HL7/elims_47_3_04608717_11184_mega_specimen.hl7,,HL7_to_FHIR_to_HL7/elims_47_3_04608717_11184_mega_specimen-expected.hl7,metadata/hl7_mapping/ORU_R01/ORU_R01-base,HL7,PASS,"","","metadata/fhir_transforms/senders/CDC-ELIMS/elims-sender-transform"
```

jimmyfagan (approved):
Thanks for your patience, I'll look into that duplicate element question as I have time today, left another optional comment, otherwise looks great, approved!

prime-router/src/testIntegration/resources/datatests/HL7_to_FHIR_to_HL7/elims_47_04608717_11184_mega_specimen.hl7
mkalish: I'll double check but I believe I noted that down (and I think was an issue in an earlier message)

arnejduranovic (approved):

prime-router/src/testIntegration/resources/datatests/HL7_to_FHIR_to_HL7/elims_47_04608717_11184_mega_specimen.hl7
jimmyfagan: Ah yea that's right, in message 40 it was this same issue. Not sure if we ever discussed that with ELIMS or not, but it's preserved when the value is valid like it is in message 52, and I think we assumed we don't need to preserve invalid/meaningless inputs there, which still seems reasonable.

Issue #10706, opened 2023-08-01
Update Swagger Authentication


arnejduranovic:
### User Story

As the Swagger documentation page, I need to provide users a simple way to authenticate so the API calls will actually work on staging and production.

### Description/Use Case

Presently, the [API programmers guide](https://staging.reportstream.cdc.gov/resources/api/getting-started) covers TWO ways authentication can happen:
1. Method 1: Token-based authentication with a public/private key pair (Note: This method is the recommended best practice.)
2. Method 2: Shared secret API key (This is the Azure Function authentication scheme where you can generate a static API KEY in azure for a specific azure function and use that to authenticate. This is less secure than Method1 and is not encouraged)

The Swagger page has an "authorize" button. When the user clicks this, they should be able to at least use the second part of Method 1 to authenticate. Then, when they execute an endpoint in Swagger, the bearer token they entered will be sent in the header. Note, it would be nice to be able to do part one of Method 1 as well, requesting the token that is used in part 2 of Method 1, but they can separately use the swagger page for the `api/token` endpoint to do that part if needed. So, this means the user flow could potentially be:

1. User navigates to the `/api/token` swagger page and executes endpoint to get bearer token
2. User navigates to the swagger page containing the endpoint they want to use
3. User clicks "authorize" and enters the bearer token returned in step 1
4. User executes the desired endpoint and the bearer token is sent in the header

The downsides of this is that its kind of cumbersome on the user since the token will expire (every 5 min?). It would be really nice if the /api/token call was able to be configured in the same "authorize" window and can happen automatically. This seems to be how the default swagger Okta auth0 works?

### Risks/Impacts/Considerations

- Should be as user friendly as possible

This should also be tested locally. By default, the auth step will get skipped on a local env if no key is provided. If you provide the key in the header, the auth step WILL run. See flow below for more detail:

1. `AuthenticateClaims.authenticate` gets called at the top of the azure function
2. `getAccessToken` gets called at the top of `AuthenticateClaims.authenticate`
3. `isLocal` gets called right after `getAccessToken` and if the returned value of `getAccessToken` is not null auth will be executed. See `isLocal` for more details.

### Acceptance Criteria

- [X] User auth flow for Swagger designed and implemented (should be documented in swagger doc, separate story for this)
- [X] "Authorize" button works with server-server auth and tested locally (on the api keys page)
- [ ] Optionally, "Authorize" button works with x-functions key (on the api keys page)

[...]

Issue #10621, opened 2023-07-26
Deployment: API Docs Swagger UI hosting and configuration


JFU-NAVA-PBC:
### User Story

As a Report Stream end user / application developer, I want the API Docs Swagger UI hosted from Azure Storage Blob Container (like a static site)

### Description/Use Case

the API Docs can be accessed from external (internet) and API calls can be made to the backend

### Risks/Impacts/Considerations

1. Proper access level has to be enabled for the swagger ui contents hosted on the blob container
2. CORS settings on Azure Functions policies configured so that API calls originated from swagger ui page allowed
3. API Docs swagger UI contents can be hosted on a dedicated Azure storage server / container - separate from RS Azure Functions' persistent storage - Queues / Tables etc.
4. 

### Dev Notes

platform has completed the app dev work of the API Docs generation and hosting on a local dev setup
see #9302

Need DevOp involved to move on to deployment to LLE, and PROD

### Acceptance Criteria

API Docs swagger UI deployed and become available to external / internal users and developers
On deployment ENV, the security settings / policies are compliant with CDC requirement

PR #10842, opened 2023-08-10
10706 Enable Authorization Methods on RS API Docs Swagger UI

JFU-NAVA-PBC:
This PR:

Enabled Authorization on API Docs swagger UI, the supported authorizations:

OAUTH2
Server2Server
Azure API Key

Test Steps:
1. Checkout the PR to local machine and start report stream server as per README
2. At the completion of the RS server start up, the swagger UI of API docs should be provisioned
3. Go to Azure Storage Explorer and copy the URL of swaggerUI index.html, on local setup, it is:
http://127.0.0.1:10000/devstoreaccount1/apidocs/index.html
4. Paste it into a browser, and you will see:
5. On top of the page: there is external Docs link - currently I put Report Stream API Programmer's Guide link there ( can also use the new UP documents "Swagger UI" mark down as the entry point )
6. Click "Authorize" button to bring up authorization panel where the following auth methods are available:
7. "primeSecurityServerToServer  (http, Bearer)" Method 1 - the process of obtaining the token using primeCLI commands or the python sample program remain as the recommended practice: just keep the legacy process since it worked well and mostly is only used by internal developers
8. "primeSecurityAPIKey (apiKey)" Method 2 - Only available on Azure
9. "primeSecurityOAUTH (OAuth2, authorizationCode)" - associated with OKTA Staging (hhs-prime.oktapreview.com)
10. For OAUTH2 to work, the swagger UI needs to be in a browser started with CORS disabled
e.g on Mac, start chrome as below:
`open -n -a /Applications/Google\ Chrome.app/Contents/MacOS/Google\ Chrome --args --user-data-dir="/tmp/chrome_dev_test" --disable-web-security`
Note, this PR mostly deal with swagger UI API Docs on a local setup, on PROD and LLEs, there is a ticket #10621 

On auth server side add the redirect url to the redirect URI list.

For local dev test:

`http://127.0.0.1:10000/devstoreaccount1/apidocs/oauth2-redirect.html`

## Changes
- *Added swagger annotations to enable authorizations supported: server 2 server, azure API key*

### Testing
- [X] Tested locally?
- [X] Added tests: manual tests: see the setup notes above, after Okta login close the swagger UI auth dialogue and call API  end points, e.g. GET public key for an org, to verify access token is checked, either turn on DEBUG logging and check the logging showing Oauth authentication passed, or put a break point at the authentication point in the end point:
e.g. AuthenticatedClaims.authenticate(request) <=== verify the header has "Authorization bearer: ....." with token value

### Process

## Linked Issues
- #10706

## To Be Done
- #issue
#10862 
#10621

    - new endpoint A
    - new endpoint B


prime-router/src/main/kotlin/azure/ApiKeysFunctions.kt
```diff
@@ -146,15 +147,24 @@ SwIDAQAB
 
 @OpenAPIDefinition(
     info = Info(
-        title = "Prime ReportStream",
-        description = "A router of public health data from senders to receivers",
+        title = "OpenAPI 3.0 API Docs for Prime ReportStream",
+        description = "Prime ReportStream: A router of public health data from senders to receivers",
```
arnejduranovic, line 151:
```suggestion
        description = "Prime ReportStream: A router of public health data",
```

prime-router/src/main/kotlin/azure/ApiKeysFunctions.kt
```diff
@@ -174,36 +184,34 @@ SwIDAQAB
...
             description = "OAUTH2 Authorization for Report Stream API Access."
         ),
-        SecurityScheme(
-            name = "primeSecurityAPIKey",
-            type = SecuritySchemeType.APIKEY,
-            paramName = "x-functions-key",
-            description = "Azure Function Key Authorization for Report Stream API Access."
-        ),
         SecurityScheme(
             name = "primeSecurityServerToServer",
```
arnejduranovic, line 204: This worked locally, but I had to use the python script in the repo (or some other means) to get the bearer token. Can we build this into swagger so users can do everything in one place? Server-server is based on OAuth2.0, which swagger supports so I'm thinking it should be possible?

prime-router/src/main/kotlin/azure/ApiKeysFunctions.kt
```diff
@@ -174,36 +184,34 @@ SwIDAQAB
...
             type = SecuritySchemeType.HTTP,
             scheme = "Bearer",
             bearerFormat = "JWT",
             description = "HTTP Bearer Token Authorization for Report Stream API Access."
+        ),
+        SecurityScheme(
+            name = "primeSecurityAPIKey",
```
arnejduranovic, line 211: When I tested this locally, I did not see "x-functions-key" in the header of swagger's HTTP request? Can you confirm this works?

prime-router/src/main/kotlin/azure/ApiKeysFunctions.kt
```diff
@@ -174,36 +184,34 @@ SwIDAQAB
 @SecuritySchemes(
     value = [
         SecurityScheme(
-            name = "primeSecurity",
+            name = "primeSecurityOAUTH",
```
arnejduranovic, line 187: Did not test this locally since you did a demo on this. Can you explain why the CORS trouble? Is that because we are running locally and this will not be an issue on staging/production?

arnejduranovic (changes requested):

prime-router/src/main/kotlin/azure/ApiKeysFunctions.kt
JFU-NAVA-PBC: chatted in slack convo

prime-router/src/main/kotlin/azure/ApiKeysFunctions.kt
JFU-NAVA-PBC: I did not try it either, thought to try it out on Azure when got a chance, however, good catch, the x-functions-key - it could be in header or qparam, let me check if it's in qparam, if so, need to change the authteniticate(...) a bit

prime-router/src/main/kotlin/azure/ApiKeysFunctions.kt
JFU-NAVA-PBC: On staging / prod, CORS still needs to be properly configured, CORS configuration should be a common practice for DevOpSec team, in this case, the browser expects a request with Origin header present but there is not:

OKTA doc on the error :

https://support.okta.com/help/s/article/Browser-requests-to-the-token-endpoint-must-use-Proof-Key-for-Code-Exchange?language=en_US

prime-router/src/main/kotlin/azure/ApiKeysFunctions.kt
JFU-NAVA-PBC: > When I tested this locally, I did not see "x-functions-key" in the header of swagger's HTTP request? Can you confirm this works?

Forgot to indicate where the value is: there is an "In" attribute needs to show up in the request, should add it with "header" value..., will fix

JFU-NAVA-PBC: Now, it has the x-functions-key in header:
x-functions-key -> fake-api-key-00001

JFU-NAVA-PBC: the authenticate check on each API methods is left as is, so on local you do APIKEY auth or not, it does not matter - it fall back to the default - grant a claim.

On staging (assume CORS policies etc. are properly set), the authenticate will behave as is.

And if APIKEY auth is selected, the swagger UI API docs should work as expected - assuming Azure API management security policies are properly set - this will be verified when swagger ui got deployed  on staging

JFU-NAVA-PBC:
@thetaurean @arnejduranovic removed the azure api key (x-functions-key) auth from swagger ui auth security scheme due to its deprecation from RS API, also removed prod server from target servers list.

arnejduranovic (approved):

arnejduranovic (changes requested):
Removing approval while I figure out why I'm not seeing updates locally

arnejduranovic (approved):

Issue #8876, opened 2023-03-28
Design how to truncate HL7 v2 fields for messages being sent to receivers


victor-chaparro:
### User Story:
As ReportStream, I want HL7 v2 messages being sent to be truncated to a configurable length. 

### Description/Use Case
Currently the COVID pipeline has the ability to truncate HL7 fields as part of the receiver settings using the `truncateHl7Fields` and `truncateHDNamespaceIds` settings. We need the ability to support this as part of the Universal Pipeline. 

### Risks/Impacts/Considerations
* Do we want to follow what the Covid pipeline is doing now, or provide the ability to configure the max length for different fields.

### Dev Notes:
* We need to find out whether we want every receiver to follow the HL7 v2 spec or allow receivers to configure the max length for fields
* Find out if this needs to be in code or part of receiver transforms.
* Look into how HL7 v2 spec expects truncated fields to be represented.
* Should we have a default configuration and allow that to be overridden?
* See #8752 for the work around we did to support CA pilot

### Acceptance Criteria

- [ ] Design the HL7 v2 truncate component and document the design.
- [ ] Present the design to the Pipeline and Engagement team to get buy-in.

arnejduranovic:
My proposal for the user story: 

`As ReportStream, I must support the ability to truncate HL7v2 Field values to a configurable length. `

To me, the user story, as written, gives the impression you want to truncate the HL7 message as a whole, and not the individual fields? Is my understanding correct?

PR #10777, opened 2023-08-04
Enable DateTime Conversion Translation Setting in the UPn

oslynn:
This PR enables the ability to convert dates in the UP according to a receiver's specification, so that I can send messages in format the receiver expects.

Test Steps:
1. clone the repos
2. ./gradlew clean package
3. ./gradlew :test --tests "gov.cdc.prime.router.fhirengine.engine.CustomTranslationFunctionsTest"
4. ./gradlew :test --tests "gov.cdc.prime.router.fhirengine.translation.hl7.FhirToHl7ConverterTests"
5. For integration test using Postman collection below:
[Simple.Report.Bug.Fix.postman_collection.txt](https://github.com/CDCgov/prime-reportstream/files/12305168/Simple.Report.Bug.Fix.postman_collection.txt)

Output HL7 file should contain high precision datetime as checkmark
image

## Changes
- *Include a comprehensive list of changes in this PR*
- *(For web UI changes) Include screenshots/video of changes*

### Testing
- [x] Tested locally?
- [x] Ran `./prime test` or `./gradlew testSmoke` against local Docker ReportStream container?
- [ ] (For Changes to /frontend-react/...) Ran `npm run lint:write`? 
- [x] Added tests?

### Process
- [ ] Are there licensing issues with any new dependencies introduced?
- [ ] Includes a summary of what a code reviewer should test/verify?
- [ ] Updated the release notes?
- [ ] Database changes are submitted as a separate PR?
- [ ] DevOps team has been notified if PR requires ops support?

## Linked Issues
- Fixes #issue
- https://app.zenhub.com/workspaces/engagement-6166edbd409257001e09f1b2/issues/gh/cdcgov/prime-reportstream/10545

## To Be Done
*Create GitHub issues to track the work remaining, if any*
- #issue

## Specific Security-related subjects a reviewer should pay specific attention to
- Does this PR introduce new endpoints?
    - new endpoint A
    - new endpoint B
- Does this PR include changes in authentication and/or authorization of existing endpoints?
- Does this change introduce new dependencies that need vetting?
- Does this change require changes to our infrastructure?
- Does logging contain sensitive data?
- Does this PR include or remove any sensitive information itself?

If you answered '_yes_' to any of the questions above, conduct a detailed Review that addresses at least:

- What are the potential security threats and mitigations? Please list the _STRIDE_ threats and how they are mitigated
    - **S**poofing (faking authenticity)
        - Threat _T_, which could be achieved by _A_, is mitigated by _M_
[...]

PR #10853, opened 2023-08-10
8876 add Hl7 truncation design document

jalbinson:
This PR adds Hl7 truncation design document.

## Linked Issues
- [#8876](https://github.com/CDCgov/prime-reportstream/issues/8876)


victor-chaparro (commented):
This is good and shows how we're going to extract the functionality out of the Covid pipeline, but it should explain how we're going to use that functionality in the UP. You can look at Ott's PR [here](https://github.com/CDCgov/prime-reportstream/pull/10777). I made some changes there to introduce an interface that allows you to pass `TranslationFunctions`. Maybe we can add this function to that interface.

victor-chaparro (approved):
LGTM. I think there will be some implementation details we haven't accounted for when calling this function from the UP, but will find out once the work starts.

PR #10861, opened 2023-08-11
Adding logging to track the duration of downloading a blob in the bat…

JessicaWNava:
…ch function

This PR adds logging around the blob store in the batch function. A second part of this ticket will be actually adding the query to Azure once this is deployed so that I have logs to work with. 

Test Steps:
1. I guess you can send a message through the pipeline and make sure you see the logs?

## Changes
- Added logging around downloading the blob in the batch function

### Testing
- [ ] Tested locally?
- [ ] Added tests?

### DIFF
--- a/prime-router/src/main/kotlin/azure/BatchFunction.kt
+++ b/prime-router/src/main/kotlin/azure/BatchFunction.kt
@@ -12,6 +12,7 @@ import gov.cdc.prime.router.fhirengine.utils.HL7MessageHelpers
 import org.apache.logging.log4j.kotlin.Logging
 import org.jooq.Configuration
 import java.time.OffsetDateTime
+import java.util.Calendar
 
 const val batch = "batch"
 const val defaultBatchSize = 100
@@ -138,6 +139,7 @@ class BatchFunction(
                             receiver.format.isSingleItemFormat -> inReports // don't merge, when we are about to split
                             receiver.timing?.operation == Receiver.BatchOperation.MERGE ->
                                 listOf(Report.merge(inReports))
+
                             else -> inReports
                         }
                         val outReports = if (receiver.format.isSingleItemFormat)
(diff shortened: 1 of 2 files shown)


luis-pabon-tf (approved):
With this change, the logs are appearing and look accurate. Approved!

prime-router/src/main/kotlin/azure/BatchFunction.kt
```diff
@@ -192,7 +194,7 @@ class BatchFunction(
                 actionHistory.trackExistingInputReport(it.task.reportId)
 
                 // download message
-                val bodyBytes = BlobAccess.downloadBlob(it.task.bodyUrl)
+                val bodyBytes = downloadBlobWithMetrics(it)
```
arnejduranovic, line 197: This change will measure the time it takes to download a single report, right? What we want to measure is the time it takes to download all reports, correct? For that is what we are aiming to speed up? So, where it says "Batch Download Time: " it should actually say "Blob Download Time" (which we can't speed up and probably shouldn't measure)?

prime-router/src/main/kotlin/azure/WorkflowEngine.kt
```diff
@@ -770,7 +776,11 @@ class WorkflowEngine(
 
         val downloadContent = (reportFile.bodyUrl != null && fetchBlobBody)
         val content = if (downloadContent && BlobAccess.exists(reportFile.bodyUrl)) {
-            BlobAccess.downloadBlob(reportFile.bodyUrl)
+            val startDownloadTime = Calendar.getInstance().timeInMillis
+            val contentToReturn = BlobAccess.downloadBlob(reportFile.bodyUrl)
```
arnejduranovic, line 780: Downloading data in a factory method is a weird thing to do imo, have we tried to see what it would take to refactor this?

prime-router/src/main/kotlin/azure/BatchFunction.kt
arnejduranovic: Also why are we downloading here again anyways? It seems like validHeaders would already have the bytes in `content` field by now? Populated from `createHeader`?

arnejduranovic (changes requested):
I may be missing something, let's discuss on call if needed

PR #10879, opened 2023-08-14
10839: elims parse as 27

mkalish:
This PR completes the required ELIMS updates getting the messages to match exactly

Test Steps:
1. Update the DEV_ELIMS receiver to be active
2. Send any of the elims_* messages from the integration tests from the ignore.ignore-elr-elims sender
3. Verify that the messages match exactly (barring the changes the ELIMS team will make)

## Changes
- Changes the HL7 reader to first try parsing an ORU message as v27
- Updates the base ORU FHIR->HL7 schema to encode fields as CWE where they are CWE in the the NIST 2.5.1 spec

### Testing
- [x] Tested locally?
- [x] Added tests?

### Process

## Linked Issues
- #10839


prime-router/metadata/fhir_mapping/hl7/datatype/CodeableConcept.yml
```diff
@@ -37,6 +37,20 @@ coding_4:
     nameOfSystem: CWE.3
 
 coding_5:
+  generateList: true
```
mkalish, line 40: Bad diff, but these just got arranged so that the coding array is ordered correctly.

prime-router/src/test/kotlin/fhirengine/engine/FhirTranslatorTests.kt
```diff
@@ -48,7 +48,7 @@ import kotlin.test.Test
 
 private const val ORGANIZATION_NAME = "co-phd"
 private const val RECEIVER_NAME = "full-elr-hl7"
-private const val ORU_R01_SCHEMA = "metadata/hl7_mapping/ORU_R01/ORU_R01-base"
+private const val ORU_R01_SCHEMA = "metadata/hl7_mapping/receivers/STLTs/CA/CA"
```
mkalish, line 51: This change is to get the test that verifies we can handle generating 2.5.1 messages when the message encoding contains 5 characters.  The default ORU_R01 schema now encodes to 27

prime-router/metadata/fhir_mapping/hl7/datatype/CodeableConcept.yml
```diff
@@ -57,14 +73,15 @@ coding_6:
     display: String, CWE.2 | CNE.2 | CE.2
     system: $overrideSystem | CWE.3 | CNE.3 | CE.3
     version: String, CWE.7 | CNE.7
+    codingSystemOID: String, CWE.14
```
jimmyfagan, line 76: (optional) Thoughts on adding the `codeIndex: "identifier"` here? Even though it's the default, it lets all the CWE fields here follow the same pattern and makes the `codeIndex` obvious here without looking at the other file.

prime-router/metadata/hl7_mapping/ORU_R01/base/specimen.yml
```diff
@@ -143,6 +143,6 @@ elements:
     schema: observation-result
     resourceIndex: resultIndex
     constants:
-      hl7ObservationPath: '/PATIENT_RESULT/ORDER_OBSERVATION(%{orderIndex})/SPECIMEN(%{resultIndex})'
-      hl7OBXField: '/PATIENT_RESULT/ORDER_OBSERVATION(%{orderIndex})/SPECIMEN(%{resultIndex})/OBX'
+      hl7ObservationPath: '/PATIENT_RESULT/ORDER_OBSERVATION(%{orderIndex})/SPECIMEN(%{resultIndex})/SPECIMEN_OBSERVATION'
+      hl7OBXField: '/PATIENT_RESULT/ORDER_OBSERVATION(%{orderIndex})/SPECIMEN(%{resultIndex})/SPECIMEN_OBSERVATION(%{resultIndex})/OBX'
```
jimmyfagan, line 147: Is `resultIndex` supposed to be in `hl7OBXField` twice? If so, can you explain that?

prime-router/metadata/hl7_mapping/common/datatype/ce-coded-with-exceptions.yml
```diff
@@ -1,47 +1,156 @@
 elements:
   - name: identifier-code
-    value: [ '%resource.coding[0].code' ]
+    value: [ '%resource.coding.where(extension(%`rsext-code-index-name`).value = "identifier").code' ]
```
jimmyfagan, line 3: You should be able to combine the two cases into one element with something like
```suggestion
    value: [ '%resource.coding.where(extension(%` rsext-code-index-name` ).value = "identifier").code', '%resource.coding[0].code' ]
```
You might need to inspect the code or test to make sure that ordering is correct, but if one expression evaluates to null, it'll try the next in the list until it finds one that works.

jimmyfagan (commented):
Leaving some early comments since I took a preemptive look at this!

prime-router/metadata/fhir_mapping/hl7/datatype/CodeableConcept.yml
mkalish: Will add!

prime-router/metadata/hl7_mapping/ORU_R01/base/specimen.yml
mkalish: Good catch...I need to look at this again.  I think this only works for our case of there only being a single observation under the specimen

prime-router/metadata/hl7_mapping/common/datatype/ce-coded-with-exceptions.yml
mkalish: Interesting... let me play around with this.  There are a few places where I don't think that will work since the condition needs to look at the resource (specifically system), but maybe can clean up the others

prime-router/metadata/hl7_mapping/ORU_R01/base/specimen.yml
mkalish: Updated

mkalish:
@jimmyfagan FYI, this PR now means we are 99% matching the current messages we have from the ELIMS team with the known differences being items they'll be changing.  Noting the remaining differences:
- timestamp precision across many fields
- Message 40 has bad data
  - name data is invalid
  - birthdate is invalid
- fixing ORC.23 (WPN -> NET) when it's an email rather than a phone number.

I've updated all of the expected outputs, but I think just checking a few would suffice as we'll be writing that tool to do is programatically.

prime-router/metadata/hl7_mapping/common/datatype/ce-coded-with-exceptions.yml
mkalish: Tried a couple of things, but I'm leaning towards leaving it the same as this can't be done for every case because there are situations where the coding array is `['identifier', 'second alternate identifier']`

prime-router/src/main/kotlin/fhirengine/utils/HL7Reader.kt
```diff
@@ -1,13 +1,18 @@
 package gov.cdc.prime.router.fhirengine.utils
 
+import ca.uhn.hl7v2.AbstractHL7Exception
+import ca.uhn.hl7v2.DefaultHapiContext
 import ca.uhn.hl7v2.ErrorCode
 import ca.uhn.hl7v2.HL7Exception
+import ca.uhn.hl7v2.model.AbstractMessage
 import ca.uhn.hl7v2.model.Message
-import ca.uhn.hl7v2.model.v251.segment.MSH
+import ca.uhn.hl7v2.model.v27.message.ORU_R01
```
jimmyfagan, line 9: Could we import this with the version number included? Makes it a bit more readable later on. Something like `import ca.uhn.hl7v2.model.v27.message.ORU_R01 as v27-ORU_R01`. We could also import the 2.5.1 one similarly and use that below instead of `ca.uhn.hl7v2.model.v251.message.ORU_R01`.

prime-router/src/main/kotlin/fhirengine/utils/HL7Reader.kt
```diff
@@ -18,33 +23,122 @@ import java.util.Date
...
+     * try to parse the messages.
+     *
+     * This function assumes all the message types will be the same if this is a HL7 batch.
+     */
+    private fun getMessageModelClasses(rawMessage: String): List<Class<out AbstractMessage>> {
+        val iterator = Hl7InputStreamMessageIterator(rawMessage.byteInputStream())
+        if (iterator.hasNext()) {
+            try {
+                val firstMessage = iterator.next()
+                return when (val messageType = getMessageType(firstMessage)) {
```
jimmyfagan, line 120: Thoughts on adding a function `getMessageStructure` to use here in case we ever need an `ORU` which isn't `R01`? Could also be a backlog item, this seems safe enough for now.

prime-router/src/main/kotlin/fhirengine/utils/HL7Reader.kt
```diff
@@ -18,33 +23,122 @@ import java.util.Date
...
+                messageModelsToTry.forEach { model ->
+                    val context = DefaultHapiContext(CanonicalModelClassFactory(model))
+                    context.validationContext = validationContext
+                    try {
+                        val iterator = Hl7InputStreamMessageIterator(rawMessage.byteInputStream(), context)
+                        while (iterator.hasNext()) {
+                            messages.add(iterator.next())
+                        }
+                    } catch (e: Hl7InputStreamMessageStringIterator.ParseFailureError) {
+                        parseError = e
```
jimmyfagan, line 82: Just thinking about an error case here, what if we get here and `messages` isn't empty? Would happen if we could parse some messages with the given model, but not all of them (not sure if that's even possible, but I think each message is parsed by `iterator.next()`). Seems like maybe we'd want to clear out `messages` so that we can continue onto the next model in the list or appropriately log the failure if there are no more models to try.

prime-router/src/main/kotlin/fhirengine/utils/HL7Reader.kt
```diff
@@ -87,18 +195,23 @@ class HL7Reader(private val actionLogger: ActionLogger) : Logging {
          * @return the timestamp or null if not specified
          */
         fun getMessageTimestamp(message: Message): Date? {
-            val timestamp = (message["MSH"] as MSH).msh7_DateTimeOfMessage
-            return if (!timestamp.isEmpty && !timestamp.ts1_Time.isEmpty) {
-                timestamp.ts1_Time.valueAsDate
-            } else null
+            return when (val structure = message["MSH"]) {
+                is ca.uhn.hl7v2.model.v27.segment.MSH -> structure.msh7_DateTimeOfMessage.valueAsDate
+                is ca.uhn.hl7v2.model.v251.segment.MSH -> structure.msh7_DateTimeOfMessage.ts1_Time.valueAsDate
```
jimmyfagan, line 200: Optional but we could import these segments with things like `import ca.uhn.hl7v2.model.v251.segment.MSH as v251-MSH` similarly to my suggestion for the message models.

jimmyfagan (commented):
This looks awesome, and it works great! Left some comments, all pretty minor stuff.

prime-router/src/main/kotlin/fhirengine/utils/HL7Reader.kt
mkalish: Yep, makes sense.  I'll clean up these imports across the board

prime-router/src/main/kotlin/fhirengine/utils/HL7Reader.kt
mkalish: I

prime-router/src/main/kotlin/fhirengine/utils/HL7Reader.kt
mkalish: Clearing out the messages makes a ton of sense to me.

prime-router/src/testIntegration/resources/datatests/translation-test-config.csv
```diff
@@ -16,9 +16,6 @@ HL7_to_FHIR_to_HL7/elims_40_4988249_33033-with-truncation.hl7,,HL7_to_FHIR_to_HL
 HL7_to_FHIR_to_HL7/elims_52_24095_04204575_5120.hl7,,HL7_to_FHIR_to_HL7/elims_52_24095_04204575_5120-expected.hl7,metadata/hl7_mapping/ORU_R01/ORU_R01-base,HL7,PASS,"","","metadata/fhir_transforms/senders/CDC-ELIMS/elims-sender-transform"
 HL7_to_FHIR_to_HL7/elims_47_04608646_11024_mega_specimen.hl7,,HL7_to_FHIR_to_HL7/elims_47_04608646_11024_mega_specimen-expected.hl7,metadata/hl7_mapping/ORU_R01/ORU_R01-base,HL7,PASS,"","","metadata/fhir_transforms/senders/CDC-ELIMS/elims-sender-transform"
 HL7_to_FHIR_to_HL7/elims_47_32361_04608646_11034_mega_case.hl7,,HL7_to_FHIR_to_HL7/elims_47_32361_04608646_11034_mega_case-expected.hl7,metadata/hl7_mapping/ORU_R01/ORU_R01-base,HL7,PASS,"","","metadata/fhir_transforms/senders/CDC-ELIMS/elims-sender-transform"
-HL7_to_FHIR_to_HL7/elims_47_04608717_11184_mega_specimen.hl7,,HL7_to_FHIR_to_HL7/elims_47_04608717_11184_mega_specimen-expected.hl7,metadata/hl7_mapping/ORU_R01/ORU_R01-base,HL7,PASS,"","","metadata/fhir_transforms/senders/CDC-ELIMS/elims-sender-transform"
```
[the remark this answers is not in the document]
mkalish: Yep `co` for the bad files they contained v.27/NIST 2.5.1 fields for OBX and OBR fields, but also set some fields as `CE` when they need to be `CWE`. 

I'm actually not sure that we could actually support that case even with more changes.

prime-router/src/testIntegration/resources/datatests/HL7_to_FHIR/sample_co_1_20220518-0006.fhir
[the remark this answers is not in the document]
mkalish: Will do!

mkalish:
BTY @jimmyfagan

prime-router/metadata/hl7_mapping/common/datatype/ce-coded-with-exceptions.yml
jimmyfagan: Talked on slack about adding this for the primary identifier too in case we end up with an array like `["alternate identifier", "second alternate identifier"]`, Michael is planning on adding it in just to be safe.

jimmyfagan (approved):
We talked about adding conditions for `identifier` too, other than that, everything looks good, thanks for the quick turnaround on the follow-ups!

prime-router/src/main/kotlin/fhirengine/utils/HL7Reader.kt
```diff
@@ -87,18 +199,23 @@ class HL7Reader(private val actionLogger: ActionLogger) : Logging {
          * @return the timestamp or null if not specified
          */
         fun getMessageTimestamp(message: Message): Date? {
-            val timestamp = (message["MSH"] as MSH).msh7_DateTimeOfMessage
-            return if (!timestamp.isEmpty && !timestamp.ts1_Time.isEmpty) {
-                timestamp.ts1_Time.valueAsDate
-            } else null
+            return when (val structure = message["MSH"]) {
```
arnejduranovic, line 202: NIT: Where we use these strings, can we get this value from some enum in an existing library instead?

arnejduranovic (approved):

prime-router/metadata/hl7_mapping/common/message-header.yml
```diff
@@ -110,6 +110,11 @@ elements:
     value: [ '%resource.code' ]
     hl7Spec: [ MSH-11-2 ]
 
+  - name: version-id
```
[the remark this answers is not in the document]
mkalish: The only thing that I feel confident in saying is that we'll set MSH-12 to `2.5.1` by default and the receiver can interpret that how they would like.   I specifically added this because now that the HL7 message class used is the 27 one, that is what will be put in MSH-12 without this change.  Since, the goal for the ELIMS message was  `in == out` I made this update.

However, I do think this should lead to a larger conversation around what is meant by these different versions and what that means for report stream

prime-router/src/main/kotlin/fhirengine/utils/HL7Reader.kt
mkalish: Didn't see any enum values for this so I just pulled this up as a constant

jimmyfagan:
Tried to keep this up-to-date with `master` so the automerge would work, but it seems like the smoke tests are failing, we can look into it next week.

Issue #10422, opened 2023-07-13
Support automating prime login --env staging command


arnejduranovic:
### User Story

As a devops engineer (or anyone else for that matter),
I want to be able to log in to staging and run smoketest with out requiring an interactive user 
so I can automate the part of the deployment process where an engineer logs into staging and runs the smokeTests.

### Description/Use Case

Presently, the `prime login --env staging` command requires a user to enter Okta account credentials into the browser. We need to investigate and implement a method for providing these credentials non-interactively. We should see if there is a way to pass the Okta credentials from the command line instead of browser. If we do it this way, we need to ensure the credentials won't show up in any logs

### Risks/Impacts/Considerations

- Ensure this does not lead to passwords getting leaked

### Acceptance Criteria

- [ ] `prime login --env staging` process can be fully automated without requiring input from a person

bishoyayoub:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/10422?planning-poker) with Zenhub @arnejduranovic @brick-green @jack-h-wang @JessicaWNava @JFU-NAVA-PBC @jimmyfagan @luis-pabon-tf @mkalish @thetaurean

luis-pabon-tf:
Not assigning officially yet, but _dibs_

luis-pabon-tf:
Current findings:
- https://developer.okta.com/docs/guides/implement-grant-type/clientcreds/main/#request-for-token - We are able to send API calls directly via Client credentials authorization
- The one requirement we are missing is the authorization header, which is in the format of `authorization: Basic <Base64-encoded client ID and secret>`
- We have the client ID for staging available in the codebase, but not the secret

luis-pabon-tf:
We are able to make API calls using the encoded authorization value, as well as receive the access token.
The problem lies in how to store the secret:
* Given that this method bypasses Okta authentication while using the ReportStream account, we need to treat the Secret with some level of security
* Storing the secret value in the code itself is out of the question
* All our available secure Key Vaults require login to get data
* Passing the value via messages or some other method would be unsafe, and just adds additional steps

Based on these findings, it's my opinion that with our current setup we cannot do this ticket without sacrificing security.

PR #10933, opened 2023-08-16
Automate Okta login when the credentials are available

luis-pabon-tf:
This PR adds automates the Access Token fetching process so that Staging tests can be fully automated. It also includes commands to force refresh and do the old manual method.

Test Steps:
1. Go to `pdhstaging-functionapp` and the Configuration section of it
2. Get the `OKTA_authKey` value from `Application settings` and `export` it on your terminal
3. Run `./prime login --env staging` on the same terminal and confirm there is a log message starting with `Login valid until`
4. Repeat the previous step, and verify the log has a message starting with `Has a valid token until`
5. Run `./prime login --env staging --force` on the same terminal and confirm there once again is a log message starting with `Login valid until`
6. Run `./prime login --env staging --force --manual`, login to Okta via the browser, and verify that in the terminal there once again is a log message starting with `Login valid until`

## Changes
- Add automatic login to `OktaCommand` via the `Client Credentials` Api
- Replace hardcoded Api values from `OktaCommand` with Environmental Variable loading
- Add `manual` and `refresh` options for `OktaCommand`

## Linked Issues
- #10422
- #10964


prime-router/src/main/kotlin/cli/OktaCommands.kt
```diff
@@ -62,26 +68,53 @@ class LoginCommand : OktaCommand(
         .choice("local", "test", "staging", "prod")
         .default("local", "local")
 
+    private val manualLogin by option(
+        "--manual",
+        help = "Enable to open the login page and sign manually to Okta"
```
mkalish, line 73:
```suggestion
        help = "Enable to open the login page and sign in to Okta"
```

prime-router/src/main/kotlin/cli/OktaCommands.kt
```diff
@@ -62,26 +68,53 @@ class LoginCommand : OktaCommand(
         .choice("local", "test", "staging", "prod")
         .default("local", "local")
 
+    private val manualLogin by option(
```
mkalish, line 71: This name was a little confusing.  Just to confirm this flag changes whether to use an API key configured as an env var or to log in and get a temporary token?

prime-router/src/main/kotlin/cli/OktaCommands.kt
```diff
@@ -195,6 +244,23 @@ abstract class OktaCommand(name: String, help: String) : CliktCommand(name = nam
     }
 
     companion object {
+        /**
+         * Base endpoint for the Okta account
+         */
+        lateinit var oktaBaseUrl: String
```
mkalish, line 250: Nit: having these as a lateinit vars made this a little harder to read.  One other possible way to do this would be to create a wrapper like `OktaConfig` and then collapse the `oktaBaseUrls` and `clientIds` maps into one

prime-router/src/main/kotlin/cli/OktaCommands.kt
```diff
@@ -62,26 +68,53 @@ class LoginCommand : OktaCommand(
...
         val oktaApp = Environment.get(env).oktaApp ?: abort("No need to login in this environment")
-        val accessTokenFile = readAccessTokenFile()
-        if (accessTokenFile != null && isValidToken(oktaApp, accessTokenFile)) {
+
+        // Load environment variables for the Okta Api
+        // The prefix has to be added as this value is already set for other uses
+        oktaBaseUrl = oktaBaseUrls[oktaApp] ?: error("Invalid app - Okta url")
+        oktaClientId = clientIds[oktaApp] ?: error("Invalid app")
+
+        oktaAuthKey = System.getenv("OKTA_authKey")
```
mkalish, line 89: This would seem to break the current behavior?

mkalish (changes requested):
Maybe I'm misunderstanding the requirements, but this seems to now only work if you have grabbed the auth key where as before I could login without it

prime-router/src/main/kotlin/cli/OktaCommands.kt
arnejduranovic: Also, we should figure out how to get around using `System.getenv` and use azure to get the key because this env var cannot be stored in terraform since it is sensitive info?

prime-router/src/main/kotlin/cli/OktaCommands.kt
```diff
@@ -26,20 +26,26 @@ import java.time.LocalDateTime
...
+ * A local server is setup to handle the Oauth2 redirect and to capture
  * the authorization code.
  *
  * Based on Okta article https://developer.okta.com/blog/2018/12/13/oauth-2-for-native-and-mobile-apps
  */
+
 private const val oktaProdBaseUrl = "https://hhs-prime.okta.com"
-const val oktaPreviewBaseUrl = "https://hhs-prime.oktapreview.com"
+private const val oktaPreviewBaseUrl = "https://hhs-prime.oktapreview.com"
 private const val oktaProdClientId = "0oa6kt4j3tOFz5SH84h6"
```
arnejduranovic, line 43: These should all be env vars or something, but probably out of scope

prime-router/src/main/kotlin/cli/OktaCommands.kt
luis-pabon-tf: I'm thinking it might make more sense to have the default behavior be the current login page process, and instead have this new method run if specified

luis-pabon-tf: Updated the name to be clearer, and made the current behavior be the default

prime-router/src/main/kotlin/cli/OktaCommands.kt
luis-pabon-tf: I have moved the env loading to only happen when the key is required, so by default it's ignored

luis-pabon-tf: My understanding after discussing with DevOps is that terraform is able to be set to pull from the Azure KeyVault

prime-router/src/main/kotlin/cli/OktaCommands.kt
luis-pabon-tf: Turned it into a wrapper, all good now.

prime-router/src/main/kotlin/cli/OktaCommands.kt
```diff
@@ -293,4 +325,48 @@ abstract class OktaCommand(name: String, help: String) : CliktCommand(name = nam
...
+    var authKey: String = oktaDummyAccessToken
+) {
+    /**
+     * Load environment variables for the Okta Api
+     * @param oktaApp Configuration values
+     */
+    fun init(oktaApp: OktaCommand.OktaApp) {
+        baseUrl = oktaBaseUrls[oktaApp] ?: error("Invalid app - Okta url")
+        clientId = clientIds[oktaApp] ?: error("Invalid app")
+        authKey = System.getenv("OKTA_authKey")
```
mkalish, line 350: This is throwing an NPE for me when I don't have the this env var set.  I'm wondering if we should make this optional?

mkalish (commented):
Updates look good, still one question on regarding requiring the new env var.

prime-router/src/main/kotlin/cli/OktaCommands.kt
```diff
@@ -62,26 +69,47 @@ class LoginCommand : OktaCommand(
         .choice("local", "test", "staging", "prod")
         .default("local", "local")
 
+    private val useApiKey by option(
+        "--useApiKey",
+        help = "Enable to sign in to Okta via an Api request instead of a login page"
```
mkalish, line 74: Nit:
```suggestion
        help = "Enable to sign in to Okta via an API key instead of a login page"
```

prime-router/src/main/kotlin/cli/OktaCommands.kt
```diff
@@ -62,26 +69,47 @@ class LoginCommand : OktaCommand(
...
+    private val forceRefreshToken by option(
+        "--force",
+        help = "Enable to do a login regardless of any current access tokens"
+    ).flag(default = false)
+
     override fun run() {
         val oktaApp = Environment.get(env).oktaApp ?: abort("No need to login in this environment")
-        val accessTokenFile = readAccessTokenFile()
-        if (accessTokenFile != null && isValidToken(oktaApp, accessTokenFile)) {
+        oktaConfig.init(oktaApp)
```
mkalish, line 84: Nit: I'm fine leaving it as it is, but this could be potentially further cleaned up by not storing/mutating state in the companion object in favor of just passing an instance of `OktaConfig` to methods that need it

mkalish (changes requested):
Looking good, thanks for the changes!  Couple of other spots that still need to get cleaned up.  Let me know if I confused anything.

I confirmed I can login via the browser and will test now with the api key

prime-router/src/main/kotlin/cli/OktaCommands.kt
```diff
@@ -26,20 +26,27 @@ import java.time.LocalDateTime
...
-const val oktaPreviewBaseUrl = "https://hhs-prime.oktapreview.com"
+private const val oktaPreviewBaseUrl = "https://hhs-prime.oktapreview.com"
 private const val oktaProdClientId = "0oa6kt4j3tOFz5SH84h6"
 private const val oktaPreviewClientId = "0oa2fs6vp3W5MTzjh1d7"
 private const val oktaAuthorizePath = "/oauth2/default/v1/authorize" // Default authorization server
+private const val oktaIntrospectPath = "/oauth2/default/v1/introspect" // For checking if a token is active
 private const val oktaTokenPath = "/oauth2/default/v1/token"
-private const val oktaUserInfoPath = "/oauth2/default/v1/userinfo"
-private const val oktaScope = "openid"
+private const val oktaScope = "simple_report_dev"
```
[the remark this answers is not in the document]
luis-pabon-tf: The `openid` scope cannot be used with the client credentials Api, but `simple_report_dev` can be used for both that and the manual login logic. While I would like us to have our own scope, I have not been able to get a ReportStream scope added to the account...

prime-router/src/main/kotlin/cli/OktaCommands.kt
```diff
@@ -110,6 +138,25 @@ class LoginCommand : OktaCommand(
             "code_challenge=$codeChallenge"
     }
 
+    private fun clientCredentialsAuthorize(): JSONObject {
+        if (oktaConfig.authKey == "")  error("A valid OKTA_authKey environment variable is needed for this command")
```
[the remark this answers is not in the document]
luis-pabon-tf: Updated to make it nullable

mkalish (approved):
Changes look good to me, but I unfortunately don't seem to be able to see the configuration (or the staging function app), so I did not test that part.

@arnejduranovic recommend that you test that portion if you can see them.

prime-router/src/main/kotlin/cli/OktaCommands.kt
```diff
@@ -62,26 +70,55 @@ class LoginCommand : OktaCommand(
         .choice("local", "test", "staging", "prod")
         .default("local", "local")
 
+    private val useApiKey by option(
+        "--useApiKey",
+        help = "Enable to sign in to Okta via an API key instead of a login page"
```
arnejduranovic, line 75: This should point to a place where the user can learn how to set up their env vars. See main comment.

prime-router/src/main/kotlin/cli/OktaCommands.kt
```diff
@@ -62,26 +70,55 @@ class LoginCommand : OktaCommand(
...
+    ).flag(default = false)
+
+    private val forceRefreshToken by option(
+        "--force",
+        help = "Enable to do a login regardless of any current access tokens"
```
arnejduranovic, line 80: This sounds confusing, I'm not sure what this means?

prime-router/src/main/kotlin/cli/OktaCommands.kt
```diff
@@ -110,6 +154,32 @@ class LoginCommand : OktaCommand(
             "code_challenge=$codeChallenge"
     }
 
+    /**
+     * Make a call to Okta's Client Credentials Api for an access token.
+     * @return JSONObject of the Api response.
+     */
+    private fun clientCredentialsAuthorize(oktaConfig: OktaConfig): JSONObject {
+        if (oktaConfig.authKey == null) error("A valid OKTA_authKey environment variable is needed for this command")
```
arnejduranovic, line 162: Update with link to doc explaining how to set this?

arnejduranovic (changes requested):
The steps in your ticket need to be documented somewhere in the repo docs, probably should be a section in a document talking about deployments? You don't have to do the whole document, just your relevant sections.

arnejduranovic (approved):
Exciting work!

Issue #10216, opened 2023-06-30
Write documentation on the receive step of the UP


JessicaWNava:
### User Story
As senders, receivers, and Reportstream employees I want to understand the receive step of the UP. See the doc skelly for some idea on content: https://docs.google.com/document/d/1Xr2defgjUjarZgyROqOqM1Cezx_hp6S_Gd3Ue8bik64/edit?usp=sharing

### Description/Use Case
Create a document that describes the receive step of the UP.

### Risks/Impacts/Considerations
See the documentation skelly for some help on where to start

### Dev Notes
If https://app.zenhub.com/workspaces/platform-6182b02547c1130010f459db/issues/gh/cdcgov/prime-reportstream/10208 is completed, use the existing structure, otherwise put the doc in docs/features.

### Acceptance Criteria
- [ ] Define content that should be in documentation and review with team before writing documentation
- [ ] There is a document that describes the receive step of the UP
- [ ] The document includes a section on validation - This could potentially be a link to another document/location but should describe how validation applies to this step

bishoyayoub:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/10216?planning-poker) with Zenhub @arnejduranovic @brick-green @jack-h-wang @JessicaWNava @JFU-NAVA-PBC @jimmyfagan @luis-pabon-tf @mkalish @thetaurean

PR #10971, opened 2023-08-17
Jessica/10216 receive step documentation

JessicaWNava:
This PR adds documentation on the receive step.

Test Steps:
1. N/A it is documentation

## Changes
- Added documentation on the receive step

#10216


prime-router/docs/universal-pipeline/receive.md
```diff
@@ -1,11 +1,32 @@
...
+POST /api/reports
+POST /api/waters
+```
 
-The UP can receive data in either HL7 or FHIR. We are planning to support CSV in the future.
-## Validation
+`\reports` - old API, does not really have security other than checking that the sender exists. 
+Still widely used.
+
+`\waters` - new API, uses FHIR auth security
```
brick-green, line 14: Should info on FHIR auth be included here? Or will that be documented somewhere else?

prime-router/docs/universal-pipeline/receive.md
```diff
@@ -1,11 +1,32 @@
 # Receive
+`Receive` is the first step a message goes through when entering the Universal Pipeline (UP). Despite its name, this step
+actually occurs in the `ReportFunction.kt` file. 
 
+There are two calls that will kick off this function:
+```
+POST /api/reports
+POST /api/waters
+```
```
brick-green, line 9: Should there be info on expected parameters in the calls?

prime-router/docs/universal-pipeline/receive.md
```diff
@@ -1,11 +1,32 @@
...
+`\reports` - old API, does not really have security other than checking that the sender exists. 
+Still widely used.
+
+`\waters` - new API, uses FHIR auth security
 
-Profiles - define implementation guides, give examples
+Currently, the UP can receive data in either HL7 or FHIR. We are planning to support CSV in the future.
 
-## Code Mapping
+## Async/Sync
```
arnejduranovic, line 18: The Universal Pipeline is always asnyc. This function is shared with the legacy pipeline which has the option for sync.

arnejduranovic (changes requested):
I like the "How it works" section in batch and send.md, perhaps we can add that here as well?

prime-router/docs/universal-pipeline/receive.md
```diff
@@ -1,11 +1,32 @@
...
+One of the things decided in the receive step is how the message is going to be processed, synchronously or 
+asynchronously. This is determined by a `processing` arg passed to the request and, if not value is passed by the 
+default set for the sender. Details of what this means will be discussed later in the submission step where the 
+branching logic actually occurs.  
+
+## Validation
+The main thing the `receive` step is responsible for is validating the message.
+There are big plans for validation in the works (see the [design doc](../design/design/validation/design.md)). Currently 
+though, we do very basic validation, like ensuring that it is a type of message we can process. Moving forward, we 
+want to validate messages against what are called an implementation guide. These specify a standard for a message type and
```
arnejduranovic, line 28: We won't be validating against an implementation guide in the receive step though? Also, I wouldn't say validation is the main thing it is responsible for. I guess in a way it is, it "validates" the received report on a basic level (you should add detail here on how it does that - see my suggestion for a "How it works" section) and then it uploads it to the blob store and triggers the convert step via the Azure queue.

prime-router/docs/universal-pipeline/receive.md
arnejduranovic: Is this necessary? I feel like we should just point to the swagger doc (doesn't exist yet). If you add it here, then you have to keep it up to date in multiple places instead of having a single source of truth.

prime-router/docs/universal-pipeline/receive.md
```diff
@@ -1,11 +1,50 @@
...
-## Code Mapping
+Currently, the UP can receive data in either HL7 or FHIR. We are planning to support CSV in the future.
 
+### FHIR Auth Security
+This is used in the `\waters` endpoint. It retrieves the access token from the request then it will attempt to 
+authenticate with Okta. If that does not work then it will try Server2Server auth. If that does not work it will deem 
+you Unauthorized. 
+
+### Async/Sync
+UP pipeline is always asynchronous, but this branching logic is here since the code is shared with the legacy pipeline.
```
arnejduranovic, line 42: What is "this" branching logic? What/where are you referring to? (I know what you are referring to, but the reader may not)

If you explain somewhere that this step in the pipeline just accepts a report, does a little validation, and then puts it up so the next step can run then you don't even have to have a section that talks about sync vs async. Sync vs async is describing the pipeline as a whole and is not specific to any step in particular.

I would remove this whole section.

prime-router/docs/universal-pipeline/receive.md
```diff
@@ -1,11 +1,50 @@
...
+
+### Validation
+Currently, we do very basic validation in the `receive` step, like ensuring that it is a type of message we can process 
+and breaking each message out into a separate message if it is a batch message. We also check for duplicates and, if
```
arnejduranovic, line 46: `breaking each message out into a separate message if it is a batch message
`
I'm pretty sure this happens in the convert step? Can you double check?

arnejduranovic: Regarding duplicates, that seems to only apply to HL7 and not FHIR?

prime-router/docs/universal-pipeline/receive.md
```diff
@@ -1,11 +1,50 @@
...
+
+### Async/Sync
+UP pipeline is always asynchronous, but this branching logic is here since the code is shared with the legacy pipeline.
+
+### Validation
+Currently, we do very basic validation in the `receive` step, like ensuring that it is a type of message we can process 
+and breaking each message out into a separate message if it is a batch message. We also check for duplicates and, if 
+they are not allowed via the `allowDuplicate` param, reject the message. We check that the sender is valid and enabled.
+
+The last thing it does it upload the message to the blob store and triggers the convert step via the Azure queue.
```
arnejduranovic, line 49:
```suggestion
The last thing it does is upload the message to the blob store and trigger the convert step via the Azure queue. 
```

prime-router/docs/universal-pipeline/receive.md
arnejduranovic: Also, in case you're really passionate about discussing the synchronicity of the pipeline, I cover it in the main README in the UP docs folder.

arnejduranovic (changes requested):

prime-router/docs/universal-pipeline/receive.md
JessicaWNava: For duplicates, I am pretty sure this happens in the call on line 166 of the Submission receiver which is called from line 171 of the ReportFunction.

As far as breaking it out, I did not mean that they go on the queue separately or anything, but I will take it out.

arnejduranovic (approved):

brick-green (approved):

PR #10977, opened 2023-08-18
BUG: monorepo frontend precommit modification check always passes

### LINKED ISSUE #10976: BUG: monorepo frontend precommit modification check always passes
The logic (copied from other precommit code) to check for modifications to frontend-react is always passing. This needs to be fixed.

Acceptance Criteria:
- [ ] Modifications outside frontend fail the check

jpandersen87:
Fixes #10976

This PR fixes the git modification check code so that the grep output is piped through xargs to remove whitespacing and return the exact value so that checks properly fail.


snesm (approved):
This looks like a zsh macOS work around.

PR #11054, opened 2023-08-24
10119 remove hardcoded URL dependency in rest transport

jalbinson:
This PR removes the hardcoded URLs in rest transport and replaces them with something more generic. We still required 2 special cases for NY and WA as the structure they expect is out of the norm. We make selections on the organization name rather than the URL though to prevent having to make changes in the future if URLs change.

Test Steps:
1. `./gradlew test`

## Changes
- Added a configuration for the report content-type to be specified in the json configuration block (will default to `text/plain` if missing)
- Update the `setBody` when statement to key off of organization name rather than URL suffix
- set content type of body dynamically based on configuration rather than hardedcoded based on URL

### Testing
- [ ] Tested locally?
- [ ] Added tests?

## Linked Issues
- Fixes #issue

## To Be Done
- need to add ticket to add new configurations to all REST transport recievers to add their required content-type


jalbinson:
@GilmoreA6 @oslynn Can I get both of your thoughts on this approach? I think this would simplify onboarding for the more basic REST receivers who only need us to post the report in a standard way.

prime-router/src/main/kotlin/transport/RESTTransport.kt
```diff
@@ -442,54 +443,50 @@ class RESTTransport(private val httpClient: HttpClient? = null) : ITransport {
...
-                            TextContent(message.toString(Charsets.UTF_8), ContentType.Application.Json)
-                        }
-                        "orders" -> {
-                            TextContent(message.toString(Charsets.UTF_8), ContentType.Application.Json)
-                        }
-                        // WA
-                        "elr" -> {
+                    // most should use the else branch unless it's not possible on the reciever's end
+                    when (receiver.organizationName) {
+                        "wa-phd" -> {
```
adegolier, line 463: I don't like the idea of hard coding org names into our logic.

jalbinson: I agree that it's not great. It was the best I could come up with for NY and WA's very specific custom logic required. In essence we are already doing the same thing but with URLs instead.

Maybe it's possible to configure some kind of function in the configuration? Not sure off the top of my head.

oslynn:
We don't need this ticket any longer.  It will be include in ticket 10119.

oslynn:
We don't need this ticket any longer.  It will be include in ticket 10119.

PR #11059, opened 2023-08-24
dependencies and fixes for testing

jpandersen87:
This PR updates dependencies and fixes tests that broke due to mocking. In particular, the FileHandler component itself was updated to determine what step it should be at on first render using the state it has. The testing was updated to reflect this state-based testing (as interaction testing at the unit level is not possible due to mocks).


etanb (changes requested):
Talked to Joseph via Slack, going to take a deep dive into this so we don't have to completely change production code for our tests.

chris-kuryak:
Waiting on issues regarding tests

jpandersen87:
Fixes for tests being handled elsewhere.

Issue #11050, opened 2023-08-24
Test CDC-ELIMS messages: Part 6


arnejduranovic:
### User Story
As ELIMS, I would like the HL7 in to match the HL7 out exactly after the ELIMS team updates their messages.

### Description/Use Case
Presently, HL7 IN matches HL7 OUT for the messages we have tested with, with the exception of RS adding timezone/offset to datetimes when they are missing. There are also some differences that the ELIMS team will address on their side and then send us additional messages to test with that include these changes.

### Acceptance Criteria
- [x] Running test messages located in [Test Messages - 6](https://cdc.sharepoint.com/:f:/r/teams/ReportStream/Shared%20Documents/ELIMS-ELR/ELIMS%20artifacts/Test%20Messages%20-%206?csf=1&web=1&e=UPpudy) through the pipeline results in no data loss

PR #11093, opened 2023-08-25
11050: handle cancelled OBX segments

mkalish:
This PR updates the translations to support cancelled OBX segments

Test Steps:

You'll need to set the `DEV_ELIMS` in organizations.yml to active

For each of the new ELIM tests files
1. Run them through the UP
2. Verify that the outputted HL7 file matches the input exactly with the following exceptions
  - There are still instances where ELIMS is not sending a full timestamp
  - Some the messages, ELIMS is not including MSH.18 and it gets added 

## Changes
- Add mapping for cancelled (X status) and not asked (N status) for OBX segments
- Fix the FHIR -> HL7 mapping to set OBX.2 to `ST` only when the reason for no value is `unknown`

### Testing
- [x] Tested locally?
- [x] Added tests?

### Process

## Linked Issues
- #11050


prime-router/metadata/fhir_mapping/hl7/datatype/CodeableConcept.yml
```diff
@@ -63,7 +63,20 @@ coding_6:
   constants:
     codeIndex: "alternate identifier"
 
+# Allows alternate coding to be added from variables
 coding_7:
+  generateList: true
```
mkalish, line 68: Slight re-arrangement and updated the naming here to be less confusing.  This allows to directly set the the alternate identifier in a Codable concept

prime-router/metadata/hl7_mapping/ORU_R01/base/observation-result.yml
```diff
@@ -4,7 +4,7 @@ elements:
     hl7Spec: [ '%{hl7OBXField}-1' ]
 
   - name: result-value-type-st
-    condition: '%resource.value is String or %resource.value.exists().not()'
```
mkalish, line 7: This condition is incorrect as a cancelled observation will not have this set but can have value populate.

prime-router/metadata/hl7_mapping/ORU_R01/base/observation-result.yml
```diff
@@ -4,7 +4,7 @@ elements:
...
+    condition: '%resource.value is String or (%resource.dataAbsentReason.coding.where(extension(%`rsext-code-index-name`).value = "identifier").exists() and %resource.dataAbsentReason.coding.where(extension(%`rsext-code-index-name`).value = "identifier").code = "unknown")'
```
mkalish, line 7: I'm not 100% sure if this is actually correct, but this condition is setup to keep the existing CO tests working.  There are several instances where OBX.2 is `ST` in the expected output (and input), but OBX.5 is not valued.

prime-router/metadata/fhir_mapping/hl7/resource/Observation.yml
```diff
@@ -33,25 +33,55 @@ identifier_3:
...
+
 code:
   expressionType: nested
   required: true
   vars:
     obx3: OBX.3
     obx2: STRING, OBX.2
   expressions:
-    - condition: $obx3 NOT_NULL && $obx2 NOT_EQUALS_STRING TX
+    - condition: $obx3 NOT_NULL
```
mkalish, line 84: This got changed for two reasons

- This logic actually does not work as documented.  If OBX.2 is null, this condition evaluates to false and the observation is not created
- I did not see anywhere in the HL7 -> FHIR inventory that would indicate that this logic is correct

prime-router/metadata/fhir_mapping/hl7/datatype/Entity_Identifier_Pair_2.yml
```diff
@@ -5,7 +5,7 @@ type:
    expressionType: resource
    specs: $type
    vars:
-     valueIn: $code | $alternateCode | $type
+     valueIn: $code | $primaryCode | $type
```
jimmyfagan, line 8: Was this part of the change intentional?

jimmyfagan (approved):
Looks pretty good to me! I wasn't sure about the changes in `Entity_Identifier_Pair_2.yml` and `Specimen.yml`, they probably work fine, but the renaming doesn't quite make sense to me. In general, that `EIP` handling, doesn't quite look up to spec (SPM mapping [here](https://docs.google.com/spreadsheets/d/1RhRXhpavSHLckCYV2bJg_b0onlcjMTKLGbCb7ocsfCo/edit#gid=0), references [this](https://docs.google.com/spreadsheets/d/1Ez3350XkVw7cS5ufWsFVpVXV884WOtutsWFfEAZUUvo/edit#gid=0) and [this](https://docs.google.com/spreadsheets/d/1NQIzGuwmDkvB-OVjTlzW_bgkt19DXOgj3KnLl3JhmuE/edit#gid=0)), but that seems out of scope for this, and it seems to only matter for the intermediate FHIR, since it still correctly gets translated back to HL7v2.

prime-router/metadata/fhir_mapping/hl7/datatype/Entity_Identifier_Pair_2.yml
mkalish: Ah no, I think that needs to get put back

prime-router/metadata/hl7_mapping/ORU_R01/base/observation-result.yml
mkalish: I don't think so.  We have a ticket in the backlog that will clean up/remove some of these tests and we can likely update the logic then.  I've been trying to stick with not having to delete/break any of the outputs

jimmyfagan: Do we interpret `C(R/X)` to mean based on the Predicate, it's either required or excluded? Not sure where the key is for the `Usage` column... But anyway, yea that is interesting, the CO tests seem to be pretty arbitrary, they basically just confirm that the behavior matches what it was at a point in time. So I could see us considering changing this to match NIST specs, if that test was the only reason for this, but yea that test ticket might be a better place for that, we should probably note this there though, it'll be easy to lose track of. Will add a brief note on the ticket.

prime-router/metadata/fhir_mapping/hl7/datatype/Entity_Identifier_Pair_2.yml
mkalish: Actually I was wrong here, this change is correct as we want to set the first coding in teh codeable concept

jimmyfagan: I think I understand, basically the mapping for SPM.2 goes to two distinct identifiers, neither of which is exactly an "alternate-identifier", so we can/should call them both primaries? Looks like `code` in practice is never actually used here, maybe that's where some of my confusion comes from (`code` and `primaryCode` both sound like primary codes). Overall, the SPM.2 mapping seems out-of-sync with the spec and a bit confusing (maybe this shouldn't map to CodeableConcept at all), but it works for HL7->FHIR-HL7, and it's out of scope for this ticket, so I all good for now.

arnejduranovic (approved):

PR #11122, opened 2023-08-28
Oo/engagement/jgilmore/10151 code to condition tables

GilmoreA6:
This PR represents a proposal for an initial strategy for mapping code-to-condition


prime-router/docs/design/proposals/0023-condition-to-code-mapping/0023-condition-to-code-mapping.md
```diff
@@ -0,0 +1,229 @@
...
+      conditionFilter:
+          - "bundle.entry.ofType(Observation).where(meta.tag.code in (398102009' |'13906002' |'409498004' |'40610006' |'21061004' |'75702008' |'359761005' |'414015000' |'840539006' |'416925005' |'83436008')).exists()"
+      timing:
+        operation: MERGE
+        numberPerDay: 1440 # Every minute
+        initialTime: 00:00
+```
+Example output: [prime-reportstream/prime-router/docs/design/proposals/0023-condition-to-code-mapping/exampleoutput2.hl7]()
+
+This will potentially pass along results for conditions that are not desired by the receiver in the event the message/bundle represents a multiplex test where the receiver only desires some of the test results (e.g. if NY wanted to receive FLU A and FLU B results but not covid results and ReportStream received a messages where all three were represented, all three would be sent). This is likely an edge case that will not be commonly encountered and the risk of missing results by going with option #1 is much higher.
```
mkalish, line 225: Just want to call out that there is already some handling around this and might make sense to add that to the example here?  That logic works by adding an `Endpoint` resource to the bundle that indicates which observations the receiver would like to get.

Specifically, [this](https://github.com/CDCgov/prime-reportstream/blob/70e9cc5bd4260116cbaad253ddf8080cacf7701c/prime-router/src/main/kotlin/fhirengine/engine/FHIRTranslator.kt#L184) code will remove any observations for a particular receiver based on changes to the bundle that occur [here](https://github.com/CDCgov/prime-reportstream/blob/70e9cc5bd4260116cbaad253ddf8080cacf7701c/prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt#L65)

prime-router/docs/design/proposals/0023-condition-to-code-mapping/0023-condition-to-code-mapping.md
```diff
@@ -0,0 +1,229 @@
...
+
+1.) Utility to check sender test compendiums against current mapping table<br>
+2.) Utility to automatically update tables from Value Set Authority Center API (https://www.nlm.nih.gov/vsac/support/usingvsac/vsacfhirapi.html) <br>
+3.) Reporting mechanism to inform RS/sender when unmapped value was found.<br>
+4.) Mechanism to automate which conditions each state considers reportable.
+
+## Design
+
+
+### Condition Mapping Table
```
mkalish, line 31: Would there need to be new functionality to update these tables without deployments or does the existing lookup code logic cover that?

mkalish (commented):
Couple of thoughts, but this is 🔥

prime-router/docs/design/proposals/0023-condition-to-code-mapping/0023-condition-to-code-mapping.md
```diff
@@ -0,0 +1,229 @@
+## Proposal for Code-to-Condition Mapping
```
brick-green, line 1:
```suggestion
# Proposal for Code-to-Condition Mapping
```
Set this as as the H1

prime-router/docs/design/proposals/0023-condition-to-code-mapping/0023-condition-to-code-mapping.md
```diff
@@ -0,0 +1,229 @@
...
+
+## Background
+
+Public Health reporting in the United States is not uniform in nature with each public health agency (PHA) setting their
+own criteria for what conditions need to be reported to that jurisdiction. In the context of electronic lab reporting (ELR), the onus is typically on the reporting entity (lab, hospital, clinic etc.) to determine which result messages qualify for each PHA the entity reports to.
+
+Since ReportStream takes on the burden of identifying the appropriate destination for sender’s messages, it will be necessary to implement a feature to ensure only messages for the appropriate conditions/diseases are allowed to route to each connected receiver. This feature will be similar to the COVID-19 LIVD table in that it will be a table of LOINC codes and their associated conditions that is used to reference received LOINC/SNOMED codes and return their associated condition.
+
+## Assumptions
+1.) The CSTE ValueSets that make up the Reportable Conditions Trigger Codes (RCTC) will be sufficient to capture a majority of codes sent to ReportStream. If this turns out not to be true we can modify the design to include additional data sources.
```
brick-green, line 11:
```suggestion
1.) The CSTE ValueSets that make up the Reportable Conditions Trigger Codes (RCTC) will be sufficient to capture a majority of codes sent to ReportStream. If this turns out to be false, we can modify the design to include additional data sources.
```

prime-router/docs/design/proposals/0023-condition-to-code-mapping/0023-condition-to-code-mapping.md
```diff
@@ -0,0 +1,229 @@
...
+| Member OID                    | ValueSet Identifier                           | 2.16.840.1.113762.1.4.1146.239                                                 |
+| Name                          | ValueSet Name                                 | Chlamydia trachomatis Infection (Tests for Chlamydia trachomatis Nucleic Acid) |
+| Code                          | LOINC or SNOMED coded value                   | 16601-7                                                                        |
+| Descriptor                    | LOINC or SNOMED term descriptio               | Chlamydia trachomatis rRNA [Presence] in Urine by Probe                        |
+| Code System                   | Indicates whether code is LOINC or SNOMED     | LOINC                                                                          |
+| Version                       | LOINC or SNOMED release version               | 2.74                                                                           |
+| Status                        | Indicates if code is active or depracated     | Active                                                                         |
```
brick-green, line 43:
```suggestion
| Status                        | Indicates if code is active or deprecated     | Active                                                                         |
```

prime-router/docs/design/proposals/0023-condition-to-code-mapping/0023-condition-to-code-mapping.md
```diff
@@ -0,0 +1,229 @@
...
+| Condition Name                | Name of associated reportable condition       | Chlamydia trachomatis infection (disorder)                                     |
+| Condition Code                | SONMED value associated with condition        | 240589008                                                                      |
```
brick-green, line 45:
```suggestion
| Condition Code                | SNOMED value associated with condition        | 240589008                                                                      | 
```

prime-router/docs/design/proposals/0023-condition-to-code-mapping/0023-condition-to-code-mapping.md
```diff
@@ -0,0 +1,229 @@
...
+| Version                       | LOINC or SNOMED release version               | 2.74                                                                           |
+| Status                        | Indicates if code is active or depracated     | Active                                                                         |
+| Condition Name                | Name of associated reportable condition       | Chlamydia trachomatis infection (disorder)                                     |
+| Condition Code                | SONMED value associated with condition        | 240589008                                                                      | 
+| Condition Code System         | System used for condition code                | SNOMEDCT                                                                       |
+| Condition Code System Version | SNOMED version associated with condition code | 2023-03                                                                        |
+
+The column names are taken directly from the RCTC spreadsheet.Both LOINC and SNOMED codes are combined in this table and can be identified by column "Code System".
```
brick-green, line 49:
```suggestion
The column names are taken directly from the RCTC spreadsheet.Both LOINC and SNOMED codes are combined in this table and can be identified by column "Code System" or "Condition Code System".
```

prime-router/docs/design/proposals/0023-condition-to-code-mapping/0023-condition-to-code-mapping.md
```diff
@@ -0,0 +1,229 @@
...
+
+### Supplemental Condition Mapping Table
```
brick-green, line 51: Could you expand a bit more on the value of of the supplemental table vs adding custom mappings directly to the Condition Mapping Table?

brick-green (changes requested):
A couple little things then one question

prime-router/docs/design/proposals/0023-condition-to-code-mapping/0023-condition-to-code-mapping.md
GilmoreA6: The existing CLI tools allow for tables to be updated at any time

prime-router/docs/design/proposals/0023-condition-to-code-mapping/0023-condition-to-code-mapping.md
GilmoreA6: Yes, the idea is that the condition mapping table stays true to the RCTC and CSTE valuesets and we can create tooling to automatically update that table. That gets more complicated if we have to worry about maintaining other values as opposed to just updating the entire table, but I can also see the argument that says if we are creating something custom to update the table anyways why not just build in the functionality to keep any values we added.

prime-router/docs/design/proposals/0023-condition-to-code-mapping/0023-condition-to-code-mapping.md
```diff
@@ -0,0 +1,229 @@
...
+| Status                        | Indicates if code is active or depracated     | Active                                         |
+| Condition Name                | Name of associated reportable condition       | Carbapenem resistant Enterobacteriaceae        |
+| Condition Code                | SONMED value associated with condition        | 712662001                                      | 
+| Condition Code System         | System used for condition code                | SNOMEDCT                                       |
+| Condition Code System Version | SNOMED version associated with condition code | 2023-03                                        |
+
+### Appending condition information to FHIR Bundle
+
+The condition information will be most useful and easy to work with if it exists in the FHIR bundle itself. Making use of the fact that FHIR allows for metadata about a resource to include tags that include a code and display value we can provide a tag to each observation resource in a FHIR bundle that describes the condition associated with that observation
```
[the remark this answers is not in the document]
GilmoreA6: It is certainly an option to do it that way. The thought was that by adding the condition information onto the bundle you can then use that in other ways in the future without having to build something specifically for that use case. e.g. for reporting purposes if we create something similar to the Cov ID Result Metadata table for UP. The thinking is that the whole purpose of FHIR tags are to "identify and relate resources to process and workflow" so instead of creating something custom, if we can use existing methods why not?

prime-router/docs/design/proposals/0023-condition-to-code-mapping/0023-condition-to-code-mapping.md
GilmoreA6: Ultimately the hope is that as CSTE and RCKMS expand more into ELR space (which is on their roadmap) we have a common "source of truth"

prime-router/docs/design/proposals/0023-condition-to-code-mapping/0023-condition-to-code-mapping.md
GilmoreA6: Corrected

prime-router/docs/design/proposals/0023-condition-to-code-mapping/0023-condition-to-code-mapping.md
GilmoreA6: Corrected

prime-router/docs/design/proposals/0023-condition-to-code-mapping/0023-condition-to-code-mapping.md
GilmoreA6: "Condition Code System" will always be SNOMED. In the RCTC all LOINCS and SNOMED codes that appear as order or results codes are mapping to a condition that is represented with a SNOMED code.

prime-router/docs/design/proposals/0023-condition-to-code-mapping/0023-condition-to-code-mapping.md
brick-green: yeah I definitely think it's possible to add all to the same table. We'd just need a way to identify the source of the record(manual upload vs external import) which could just be a column. The only reason I can see to create a supplemental table is if additional/different columns are needed. I could see columns for created date, uploaded by, and maybe expiry date. I think most of those would be covered by the lookup table version for the imported table but could be useful for the manual upload records.

prime-router/docs/design/proposals/0023-condition-to-code-mapping/0023-condition-to-code-mapping.md
```diff
@@ -0,0 +1,229 @@
...
+
+## Background
+
+Public Health reporting in the United States is not uniform in nature with each public health agency (PHA) setting their
+own criteria for what conditions need to be reported to that jurisdiction. In the context of electronic lab reporting (ELR), the onus is typically on the reporting entity (lab, hospital, clinic etc.) to determine which result messages qualify for each PHA the entity reports to.
+
+Since ReportStream takes on the burden of identifying the appropriate destination for sender’s messages, it will be necessary to implement a feature to ensure only messages for the appropriate conditions/diseases are allowed to route to each connected receiver. This feature will be similar to the COVID-19 LIVD table in that it will be a table of LOINC codes and their associated conditions that is used to reference received LOINC/SNOMED codes and return their associated condition.
+
+## Assumptions
+1.) The CSTE ValueSets that make up the Reportable Conditions Trigger Codes (RCTC) will be sufficient to capture a majority of codes sent to ReportStream. If this turns out to be false, we can modify the design to include additional data sources.
```
arnejduranovic, line 11: Define CSTE?

prime-router/docs/design/proposals/0023-condition-to-code-mapping/0023-condition-to-code-mapping.md
```diff
@@ -0,0 +1,229 @@
...
+| Version                       | LOINC or SNOMED release version               | 2.74                                                                           |
+| Status                        | Indicates if code is active or deprecated     | Active                                                                         |
+| Condition Name                | Name of associated reportable condition       | Chlamydia trachomatis infection (disorder)                                     |
+| Condition Code                | SNOMED value associated with condition        | 240589008                                                                      | 
+| Condition Code System         | System used for condition code                | SNOMEDCT                                                                       |
+| Condition Code System Version | SNOMED version associated with condition code | 2023-03                                                                        |
+
+The column names are taken directly from the RCTC spreadsheet.Both LOINC and SNOMED codes are combined in this table and can be identified by column "Code System".
```
arnejduranovic, line 49:
```suggestion
The column names are taken directly from the RCTC spreadsheet. Both LOINC and SNOMED codes are combined in this table and can be identified by column "Code System".
```

arnejduranovic: Can we link the spreadsheet here?

prime-router/docs/design/proposals/0023-condition-to-code-mapping/0023-condition-to-code-mapping.md
arnejduranovic: We also need to ensure we are tracking changes to the table, right? I don't see anything in the doc for that?

prime-router/docs/design/proposals/0023-condition-to-code-mapping/0023-condition-to-code-mapping.md
```diff
@@ -0,0 +1,229 @@
...
+      lookupTable:
+        tableName: Supplemental-Condition
+        keyColumn: code
+        valueColumn: condition_code      
+        
+        
+```
+The first element will check against the RCTC values loaded into the Condition-Mapping tables, while the second will check against the supplemental table if a match was not found in the Condition_Mapping table.
+
+Not all reportable conditions can be identified from order and result LOINC codes. In the case of bacterial cultures. The condition is identified by the found organism which is populated in the result field/resource (OBX-5 or Bundle.entry.resource.ofType(Observation).value.valueCodeableConcept). In order to correctly tag these messages, an additional element will need to be added to the transform like below to check the result code.
```
arnejduranovic, line 151:
```suggestion
Not all reportable conditions can be identified from order and result LOINC codes. In the case of bacterial cultures, the condition is identified by the found organism which is populated in the result field/resource (OBX-5 or Bundle.entry.resource.ofType(Observation).value.valueCodeableConcept). In order to correctly tag these messages, an additional element will need to be added to the transform like below to check the result code.
```

prime-router/docs/design/proposals/0023-condition-to-code-mapping/0023-condition-to-code-mapping.md
```diff
@@ -0,0 +1,229 @@
...
+    - name: TEST-RECEIVER
+      externalName: TEST
+      organizationName: TEST
+      topic: full-elr
+      customerStatus: active
+      jurisdictionalFilter:
+        - "(%performerState.exists() and %performerState = 'TEST') or (%patientState.exists() and %patientState = 'TEST')"
+      qualityFilter: []
+      conditionFilter:
+          - "bundle.entry.ofType(Observation).where(meta.tag.code in (398102009' |'13906002' |'409498004' |'40610006' |'21061004' |'75702008' |'359761005' |'414015000' |'840539006' |'416925005' |'83436008')).exists()"
```
arnejduranovic, line 217: Is there another way we can achieve this without having to have the codes in our filters? Could we query RCKMS to see what conditions we need to let through for each jurisdiction?

arnejduranovic (commented):
This is really great work! I appreciate the detail and thought that went into this!

prime-router/docs/design/proposals/0023-condition-to-code-mapping/0023-condition-to-code-mapping.md
GilmoreA6: No unfortunately. Long term that's the hope, but currently RCKMS doesn't have this functionality.

prime-router/docs/design/proposals/0023-condition-to-code-mapping/0023-condition-to-code-mapping.md
GilmoreA6: Yes, the tables are versioned and we could also add a column for which version of the valueset we are referencing, however I'm not sure what value this gets us?.

prime-router/docs/design/proposals/0023-condition-to-code-mapping/0023-condition-to-code-mapping.md
GilmoreA6: Done.

prime-router/docs/design/proposals/0023-condition-to-code-mapping/0023-condition-to-code-mapping.md
GilmoreA6: I can, but it's behind a login. I'll see if I can add it to the google drive.

prime-router/docs/design/proposals/0023-condition-to-code-mapping/0023-condition-to-code-mapping.md
GilmoreA6: Also @brick-green I went back and forth on having things be in one table or two (or even more). I don't think there's a right or wrong answer, just whatever most people think is easier to work with is what I would vote for.

prime-router/docs/design/proposals/0023-condition-to-code-mapping/0023-condition-to-code-mapping.md
GilmoreA6: We could have a valueset/lookuptable for each jurisdiction and have a custom fhir function that checks to see if the value is in that valueset, although that is essentially the same thing, just hiding the codes from the filters and then we couldn't change the conditions without a PR

prime-router/docs/design/proposals/0023-condition-to-code-mapping/0023-condition-to-code-mapping.md
brick-green: @GilmoreA6 maybe this could be a eng sync topic?

prime-router/docs/design/proposals/0023-condition-to-code-mapping/0023-condition-to-code-mapping.md
```diff
@@ -0,0 +1,229 @@
...
+        initialTime: 00:00
+```
+Example output: [prime-reportstream/prime-router/docs/design/proposals/0023-condition-to-code-mapping/exampleoutput2.hl7]()
+
+This will potentially pass along results for conditions that are not desired by the receiver in the event the message/bundle represents a multiplex test where the receiver only desires some of the test results (e.g. if NY wanted to receive FLU A and FLU B results but not covid results and ReportStream received a messages where all three were represented, all three would be sent). This is likely an edge case that will not be commonly encountered and the risk of missing results by going with option #1 is much higher.
+
+## Monitoring Mapping
+
+A key component for long term success of mapping code-to-condition will be the ability to capture codes that are unmapped and subsequently map them and re-send any affected messages. This will require that processes are in place to review and map any codes that do not have a mapping at time of receipt. This is called out in the "out-of-scope" section but reiterated here for greater clarity as it is the single most important factor to long term viability of this method.
```
GilmoreA6, line 229: @arnejduranovic - added a request to add mapping monitoring strategy to this proposal.

GilmoreA6: Need to add specifics on having unmapped codes logged to queryable location (likely action_log).

prime-router/docs/design/proposals/0023-condition-to-code-mapping/0023-condition-to-code-mapping.md
GilmoreA6: Getting back to this. Took me awhile to parse how that was working. I'll add some language calling that out

prime-router/docs/design/proposals/0023-condition-to-code-mapping/0023-condition-to-code-mapping.md
GilmoreA6: Not a bad idea. Maybe see what people's thoughts are on table maintenance in general.

brick-green: @GilmoreA6 Was this discussed at an eng sync? I missed one or two so wasn't sure

bishoyayoub:
James will add details about a monitoring strategy to the proposal.

PR #11147, opened 2023-08-30
Only default xtn.2 to email when use is not set

mkalish:
This PR implements the test changes proposed in https://github.com/CDCgov/prime-reportstream/pull/11149 for XTN use field, line 2 of the spreadsheet. It also sets mobile numbers to have a PRS code, ensures that pager numbers and emails make it over to the other message if the use does not exist. 

Test Steps:
1. Run the tests to make sure that the fields are expected
2. Compare the spreadsheet to the tests and make sure that they align. 

## Changes
- This PR implements the test changes proposed in https://github.com/CDCgov/prime-reportstream/pull/11149 for XTN use field, line 2 of the spreadsheet. 
- Sets mobile numbers to have a PRS code
- Ensures that pager numbers and emails make it over to the other message if the use does not exist. 

### Testing
- [x] Tested locally?
- [ ] Added tests?

### Process

## Linked Issues
#11147


prime-router/src/testIntegration/resources/datatests/mapping-inventory/xtn/xtn_use_code_home.hl7
```diff
@@ -0,0 +1 @@
+MSH|^~\&#|STARLIMS.CDC.Stag^2.16.840.1.114222.4.3.3.2.1.2^ISO|CDC Atlanta^11D0668319^CLIA|MEDSS-ELR ^2.16.840.1.114222.4.3.3.6.2.1^ISO|MNDOH^2.16.840.1.114222.4.1.3661^ISO|20230501102531-0400||ORU^R01^ORU_R01|3003786103_4988249_33033|T|2.5.1|||NE|NE|USA|UNICODE UTF-8|||PHLabReport-NoAck^PHIN^2.16.840.1.113883.9.11^ISOPID|1||kc3oy^^^^PI^Any lab USA||Schaefer^Stephenie^Davis^^^^L||19930227|U||ASKU^Asked, but unknown^HL70005^^^^2.5.1|96633 Gilberto Course^^^IG||^PRN^BP^^1^260^7595016^^^^^+1 260 759 5016~^PRN^Internet^otto.daugherty@email.com|||||||||N^Non Hispanic or Latino^HL70189^^^^2.9||||||||N
```
mkalish, line 1: So one thing I'm struggling with looking at real tests, is I think it's going to be very difficult to remember what the functionality these files are specifically testing.  One way we could resolve that is adding anew column to config that is a description, but you could also move away from that config entirely and not use dynamic tests.

mkalish (commented):
Good catch in patient, only thought is a bit of structure change for the test.

prime-router/src/testIntegration/resources/datatests/mapping-inventory/xtn.hl7
```diff
@@ -0,0 +1,2 @@
+MSH|^~\&#|STARLIMS.CDC.Stag^2.16.840.1.114222.4.3.3.2.1.2^ISO|CDC Atlanta^11D0668319^CLIA|MEDSS-ELR ^2.16.840.1.114222.4.3.3.6.2.1^ISO|MNDOH^2.16.840.1.114222.4.1.3661^ISO|20230501102531-0400||ORU^R01^ORU_R01|3003786103_4988249_33033|T|2.5.1|||NE|NE|USA||||PHLabReport-NoAck^PHIN^2.16.840.1.113883.9.11^ISO
```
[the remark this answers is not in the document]
JessicaWNava: Oops this just needs to be deleted actually.

prime-router/src/testIntegration/kotlin/datatests/MappingInventoryTests.kt
```diff
@@ -0,0 +1,79 @@
...
+            inputFile, Report.Format.FHIR, "", outputFile,
+            Report.Format.HL7, outputSchemaPath, true, null, null, null
+        )
+        return TranslationTests().FileConversionTest(
+            testConfig
+        ).runTest()
+    }
+
+    @Test
+    fun `XTN use code home`() {
```
arnejduranovic, line 21: Does it make sense to tie these to the spreadsheet somehow? Should we at the very least have a comment at the top that points to the XTN page in the mapping inventory? Something like this?

```
/* Location of mapping inventory rules
XTN : PAGE HERE
PID: PAGE HERE
...
*/
```

Additionally, for each test, is it possible to tie each test, via a comment, to a row in the spreadsheet? This probably doesn't make sense since the spreadsheet can change? An idea to resolve this is we can maintain our own copy of the mapping inventory and manually sync with the official one once in a while. Part of the sync process could be writing new tests or updating the current tests? If this makes sense, this should probably be captured in the test plan. The test plan should discuss how we detect changes to the mapping inventory and then update our mappings and tests, if that's something we want to do (I think we probably should)? @mkalish any thoughts here?

prime-router/src/testIntegration/kotlin/datatests/TranslationTests.kt
mkalish, line 1: Changes might be revertable here?

prime-router/src/testIntegration/kotlin/datatests/MappingInventoryTests.kt
```diff
@@ -0,0 +1,79 @@
+package gov.cdc.prime.router.datatests
+
+import gov.cdc.prime.router.Report
+import gov.cdc.prime.router.cli.tests.CompareData
+import org.junit.jupiter.api.Test
+
+class MappingInventoryTests {
```
mkalish, line 7: So in the inverse of what we were talking about yesterday, I think this file is going to get huge if we put all the tests here?  Maybe just rename the file to XTN?

prime-router/src/testIntegration/kotlin/datatests/MappingInventoryTests.kt
mkalish: Storing them here makes sense to me and will make it easier when we need to fix something

prime-router/src/testIntegration/kotlin/datatests/MappingInventoryTests.kt
```diff
@@ -0,0 +1,79 @@
...
+            shouldTranslateLosslessly(
+                "mapping-inventory/xtn/xtn_use_code_home_2.fhir",
+                "mapping-inventory/xtn/xtn_use_code_home_2.hl7"
+            ).passed
+        )
+    }
+
+    @Test
+    fun `XTN use code mobile`() {
+        assert(
```
mkalish, line 32: @JessicaWNava I think Arnej was recommending that we comment the row in the inventory that we're asserting against?

mkalish (commented):
@JessicaWNava This looks great to me.  I had two thoughts:
- It might make sense to clean up the package structure and files to there is better linkage between everything
  - i.e `gov.cdc.prime.router.hl7FhirMapping.xtn` and then the test file, HL7/FHIR files and inventory sheet would all be in that directory
- Adding the line from the inventory sheet that the test case is verifying

prime-router/src/testIntegration/kotlin/datatests/MappingInventoryTests.kt
JessicaWNava: No, he was saying to put the row as part of the file name. I guess I can add a comment too.

prime-router/metadata/hl7_mapping/ORU_R01/RADx_MARS/RADx_MARS-base.yml
```diff
@@ -55,4 +55,10 @@ elements:
...
   - name: universal-service
-    condition: false
\ No newline at end of file
+    condition: false
+
+  # XTN
+  - name: xtn-use-code-from-system
+    condition: '%resource.system = "email"'
+    value: [ '"NET"' ]
+    hl7Spec: [ '%{hl7TelecomPath}-2' ]
```
jimmyfagan, line 64: This highlights a slight downside to having multiple elements with the same name in referenced schemas; we can't really replicate the behavior of having NET for email and BPN for pager without adding another element with a new name here. This is a good case to keep in mind when/if we do some schema extension/overriding redesign at some point.

We probably are fine without a BPN element here, but we can utilize the override to just barely shorten this, totally optional.

```suggestion
  - name: xtn-use-code-from-system
    condition: '%resource.system = "email"'
    value: [ '"NET"' ]
```

prime-router/metadata/hl7_mapping/common/patient.yml
```diff
@@ -112,7 +112,7 @@ elements:
 
   # We have to check for system equals "email" because we're not receiving use for email
   - name: phone-number-home-mobile-email
-    resource: '%resource.telecom.where(use != "work" or system = "email")'
+    resource: '%resource.telecom.where(system = "email" or use.empty() or use != "work")'
```
jimmyfagan, line 115: Would this be fair? Not sure if we're seeing that `use == "work"` and `system == "email"` at the same time, but the comment implies that we shouldn't see that. If this is a safe change, we could also update the comment at L113 to say "assume 'home' if use is not set", to not reference email at all.
```suggestion
    resource: '%resource.telecom.where(use.empty() or use != "work")'
```

prime-router/metadata/hl7_mapping/ORU_R01/RADx_MARS/RADx_MARS-base.yml
jimmyfagan: Same optional comment applies to the CA transforms.

prime-router/src/testIntegration/kotlin/datatests/TranslationTests.kt
jimmyfagan: Agreed!

jimmyfagan: Actually we do rely on the new `runTest` function in `XTNTests.kt`. So that either needs to stay here or be adjusted for in `XTNTests.kt`.

prime-router/src/testIntegration/kotlin/datatests/mappinginventory/xtn/XTNTests.kt
```diff
@@ -0,0 +1,86 @@
+package gov.cdc.prime.router.datatests.mappinginventory.xtn
+
+import gov.cdc.prime.router.Report
+import gov.cdc.prime.router.cli.tests.CompareData
+import gov.cdc.prime.router.datatests.TranslationTests
+import org.junit.jupiter.api.Test
+
+class XTNTests {
+    fun shouldTranslateLosslessly(inputFile: String, outputFile: String): CompareData.Result {
```
jimmyfagan, line 9: I like this approach of reusing functions from TranslationTests for "unit" tests, though I think we can improve the setup. Feel free to save this for future stories.

I'm not a big fan of the function name `shouldTranslateLosslessly`; it really has nothing to do with losslessness, it just compares a result to an expected output. Maybe `tranlsateAndCompareFile` instead? And since we could easily imagine the same test being used elsewhere, we should probably pull this function out of `XTNTests` and into a more general location.

As for parameters, instead of `outputFile`, maybe we call it `expectedOutputFile`? The same issue probably already exists everywhere in TranslationTests, so not too important. We could also include parameters for input/output formats (or automatically determine those formats based on the file names), but that can probably wait until we have tests that would be using different ones.

jimmyfagan (commented):
I like this approach overall, just left some comments on some of the implementation details, nothing too critical, but probably enough that I should take another look before approving.

prime-router/metadata/hl7_mapping/common/datatype/xtn-telecom.yml
```diff
@@ -14,12 +14,12 @@ elements:
         temp: ORN
```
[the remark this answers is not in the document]
JessicaWNava: That would be if we were setting field 2, but we are setting field 3 there.

prime-router/metadata/hl7_mapping/common/patient.yml
JessicaWNava: It is actually the only way ti works. It was previously causing a bug where when it wasn't equal to work but was missing it was dropping data. I can update the comment though.

prime-router/src/testIntegration/kotlin/datatests/TranslationTests.kt
JessicaWNava: It is necessary for it to work in XTNTests and some of the other changes are just lint changes.

prime-router/metadata/hl7_mapping/common/datatype/xtn-telecom.yml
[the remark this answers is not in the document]
jimmyfagan: Just to be clear, I'm talking about `xtn-use-code` which has `hl7Spec: [ '%{hl7TelecomPath}-2' ]`. I can't comment on the exact line because it wasn't changed in the PR, but looking at the element above this comment.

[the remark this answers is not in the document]
JessicaWNava: Got it, I can change that, but should we confer with Victor or something before just changing that?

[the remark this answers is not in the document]
jimmyfagan: Yea looks like it would break some integration tests, so probably best to check with Victor, but I'm guessing it would be fine if we just update the integration tests accordingly.

prime-router/metadata/hl7_mapping/common/patient.yml
jimmyfagan: Making this change doesn't drop data, but it does break one of the new XTN tests. But I think it's still a good change. I'll leave another comment on the breaking tests to demonstrate.

prime-router/metadata/hl7_mapping/common/datatype/xtn-telecom.yml
```diff
@@ -14,12 +14,12 @@ elements:
...
 
   - name: xtn-use-code-from-system
-    condition: '%resource.system = "pager"'
+    condition: '%resource.system = "pager" and %resource.use.exists().not()'
     value: [ '"BPN"' ]
     hl7Spec: [ '%{hl7TelecomPath}-2' ]
 
   - name: xtn-use-code-from-system
-    condition: '%resource.system = "email"'
```
jimmyfagan, line 22: On second look, I don't think we should be changing this condition. `system` of `email` should be enough to use `NET`, especially since `use` might only be set on the internal FHIR bundle to help distinguish between PID.13 and PID.14.

jimmyfagan (commented):
Responded to the open convos and left a new comment or two. Would be down to huddle on some of this sometime today if that'd be useful!

prime-router/metadata/hl7_mapping/common/datatype/xtn-telecom.yml
mkalish: It's been a while, but I think this was actually what was causing us to overwrite the field for ELIMS

JessicaWNava: I'll leave that up to @mkalish

prime-router/metadata/hl7_mapping/common/datatype/xtn-telecom.yml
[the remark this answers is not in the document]
JessicaWNava: @victor-chaparro What are your thoughts on this?

prime-router/src/testIntegration/resources/datatests/mappinginventory/xtn/xtn_use_code_work_2.hl7
```diff
@@ -0,0 +1 @@
+MSH|^~\&#|STARLIMS.CDC.Stag^2.16.840.1.114222.4.3.3.2.1.2^ISO|CDC Atlanta^11D0668319^CLIA|MEDSS-ELR ^2.16.840.1.114222.4.3.3.6.2.1^ISO|MNDOH^2.16.840.1.114222.4.1.3661^ISO|20230501102531-0400||ORU^R01^ORU_R01|3003786103_4988249_33033|T|2.5.1|||NE|NE|USA|UNICODE UTF-8|||PHLabReport-NoAck^PHIN^2.16.840.1.113883.9.11^ISOPID|1||kc3oy^^^^PI^Any lab USA||Schaefer^Stephenie^Davis^^^^L||19930227|U||ASKU^Asked, but unknown^HL70005^^^^2.5.1|96633 Gilberto Course^^^IG||^WPN^Internet^otto.daugherty@email.com|^WPN^BP^^1^260^7595016^^^^^+1 260 759 5016~^WPN^Internet^otto.daugherty@email.com||||||||N^Non Hispanic or Latino^HL70189^^^^2.9||||||||N
```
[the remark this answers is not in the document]
JessicaWNava: It should be in both because the system = email still. and it is an or statement.

prime-router/metadata/hl7_mapping/common/datatype/xtn-telecom.yml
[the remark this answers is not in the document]
victor-chaparro: Yeah, making that change makes sense to me. I had added a TODO on top of that element saying there were issues when doing that, but I can't remember why and can't think of why it would cause any issues.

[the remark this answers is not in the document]
JessicaWNava: Cool, I will change it then, thank you!

prime-router/src/testIntegration/kotlin/datatests/mappinginventory/TranslateAndCompareFile.kt
```diff
@@ -0,0 +1,21 @@
+package gov.cdc.prime.router.datatests.mappinginventory
+
+import gov.cdc.prime.router.Report
+import gov.cdc.prime.router.cli.tests.CompareData
+import gov.cdc.prime.router.datatests.TranslationTests
+
+class TranslateAndCompareFile {
```
jimmyfagan, line 7: In kotlin, we can just have a stand-alone function in this file, we don't need a class to contain it.

prime-router/metadata/hl7_mapping/common/datatype/xtn-telecom.yml
jimmyfagan: I'm okay leaving this as is, and we'll open a new story to establish desired behavior for XTNs on those kinds of cases.

prime-router/src/testIntegration/resources/datatests/mappinginventory/xtn/XTN.xlsx
jimmyfagan, line 1: We should probably rename this to reflect XTN->ContactPoint. Maybe something like `XTN-to-ContactPoint.xlsx`? That is the only mapping that exists in this case, but many datatypes in HL7 have mappings to multiple different resources in FHIR.

jimmyfagan (approved):
Left a few comments, but the remaining feedback is minor and pretty straightforward, so I'll go ahead an mark this approved. Let me know if anything changes that warrants a second look, thanks!

arnejduranovic (approved):

PR #11153, opened 2023-08-30
Inline report id filter for delivery APIs

mkalish:
This PR inlines fetching the report ids that have been sent.

Test Steps:
With some delivered reports locally, run

```
curl --location 'http://localhost:7071/api/v1/receivers/ignore.HL7/deliveries' \
--header 'Content-Type: application/json' \
--header 'Authorization: Bearer dummy' \
--data '{
    "sort": {
        "direction": "ASC",
        "property": "number_of_tests"
    },
    "pagination": {
        "page": 1,
        "limit": 10
    },
    "filters": [

    ]
}'
```
and
```
curl --location 'http://localhost:7071/api/v1/receivers/ignore.HL7/deliveries/submitters/search' \
--header 'Content-Type: application/json' \
--header 'Authorization: Bearer dummy' \
--data '{
    "sort": {
        "direction": "ASC",
        "property": "sort_id"
    },
    "pagination": {
        "page": 1,
        "limit": 10
    },
    "filters": [

    ]
}'
```
and verify that you get results.

## Changes
- Changes the base clause for the recursive query to select report ids rather than pass in an array

### Testing
- [x] Tested locally?
- [ ] Added tests?

### Process


prime-router/src/main/kotlin/history/db/ReportGraph.kt
```diff
@@ -164,7 +164,16 @@ class ReportGraph(
                     ITEM_LINEAGE.CHILD_REPORT_ID.`as`(STARTING_REPORT_ID_FIELD)
                 )
                     .from(ITEM_LINEAGE)
-                    .where(ITEM_LINEAGE.CHILD_REPORT_ID.`in`(reportIds))
```
mkalish, line 167: This appears to be the culprit for why it was running so much slower in the API as I was typically testing the queries this way naively assuming that the impact would be the same...

prime-router/src/main/kotlin/history/db/DeliveryDatabaseAccess.kt
```diff
@@ -172,7 +172,8 @@ class DeliveryDatabaseAccess(val db: DatabaseAccess = BaseEngine.databaseAccessS
         if (sentReportIdsForReceiver.isEmpty()) {
```
jack-h-wang, line 172: With this change it seems the only uses for the output of `fetchReportIdsForReceiverAndTask` is to check if the resulting dataset is empty. Could be a spot for further optimization?

jack-h-wang (approved):
Looks good to me! Just one optional comment on something I noticed

prime-router/src/main/kotlin/history/db/DeliveryDatabaseAccess.kt
mkalish: I think you're correct, but I left this is so that I could measure this in PROD.  It might be that finding that it's empty earlier might speed things up.

prime-router/src/main/kotlin/history/db/SubmitterDatabaseAccess.kt
```diff
@@ -204,7 +204,8 @@ class SubmitterDatabaseAccess(val db: DatabaseAccess = BaseEngine.databaseAccess
             return ApiSearchResult(0, 0, emptyList())
         } else {
```
jimmyfagan, line 206:
```suggestion
```

prime-router/src/main/kotlin/history/db/DeliveryDatabaseAccess.kt
jimmyfagan, line 172: (optional) With `sentReportIdsForReceiver` being used only to determine empty or not, would it be worth changing `fetchReportIdsForReceiverAndTask` to something that returns a boolean and use that instead? The underlying query can be changed to fetch a boolean if that simplification is beneficial from a performance perspective, or the "is empty or not" logic can just be offloaded to the function.

jimmyfagan (approved):
Looks good, left a comment for formatting, and an optional additional change, but nothing preventing me from approving.

prime-router/src/main/kotlin/history/db/DeliveryDatabaseAccess.kt
jimmyfagan: Oops, didn't see Jack's comment, seems like we're talking about the same thing, feel free to disregard.

arnejduranovic (approved):

PR #11213, opened 2023-09-05
react-uswds 5.3 patch fix

jpandersen87:
This PR fixes the react-uswds patch that is currently not working as intended.


penny-lischer (approved):
LGTM

PR #11216, opened 2023-09-05
10486: support multiple file schemes when loading schemas

mkalish:
This PR adds functionality to resolve schemas via different schemes, including the file system, the classpath and azure.

Test Steps:
Run the smoketests

There is also test coverage and this should be a NOOP since no schemas are getting changed.  It should be possible to update all of ORU_R01 to use URIs, but that is actually just the work of another ticket.

## Changes
- Adds support for reading inputs when resolving schemas from azure, filesystem and the classpath

### Testing
- [x] Tested locally?
- [x] Added tests?

### Process

## Linked Issues
- #10486


prime-router/src/main/kotlin/fhirengine/translation/hl7/schema/ConfigSchemaReader.kt
```diff
@@ -28,8 +30,27 @@ object ConfigSchemaReader : Logging {
         schemaClass: Class<out ConfigSchema<out ConfigSchemaElement>>
     ): ConfigSchema<*> {
         // Load a schema including any parent schemas.  Note that child schemas are loaded first and the parents last.
+        val schemaList = when (URI(schemaName).scheme) {
```
mkalish, line 33: Opted to fully fork here rather than attempting to support a mix of relative pathing and URIs since that proved to be very difficult to implement.

there is a ticket for switching everything over to URIs before any azure migration

prime-router/src/main/kotlin/fhirengine/translation/hl7/schema/ConfigSchemaReader.kt
```diff
@@ -74,12 +106,81 @@ object ConfigSchemaReader : Logging {
...
+        ancestry: List<String> = listOf(),
+        schemaClass: Class<out ConfigSchema<out ConfigSchemaElement>> = ConverterSchema::class.java
+    ): ConfigSchema<*> {
+        val rawSchema = when (schemaUri.scheme) {
+            "file" -> {
+                val file = File(schemaUri)
+                if (!file.canRead()) throw SchemaException("Cannot read ${file.absolutePath}")
+                readOneYamlSchema(file.inputStream(), schemaClass)
+            }
+            "classpath" -> {
```
mkalish, line 152: This is to load files that are part of the source code (in `resources`)

prime-router/src/main/kotlin/fhirengine/translation/hl7/schema/ConfigSchemaReader.kt
```diff
@@ -74,12 +106,81 @@ object ConfigSchemaReader : Logging {
...
+                val input = javaClass.classLoader.getResourceAsStream(schemaUri.path.substring(1))
```
mkalish, line 153: Leaving it this way for now, but there might be a cleaner way to do this.  The issue I hit is that 
`URI("classpath:fhirengine/translation/hl7/schema/schema-read-test-07/ORU_R01/patient-identification.yml").path` is `null`

but 
`javaClass.classLoader.getResourceAsStream("/fhirengine/translation/hl7/schema/schema-read-test-07/ORU_R01/patient-identification.yml")` does not return the resource.

brick-green: Tried a couple different options but none were an improvement on yours. Side note, chatGPT came up with the same solution as you :)

prime-router/src/main/kotlin/fhirengine/translation/hl7/schema/ConfigSchemaReader.kt
```diff
@@ -74,12 +106,82 @@ object ConfigSchemaReader : Logging {
         return parentSchema
     }
 
+    private fun processFileSchema(
```
brick-green, line 109: Is this function used anywhere? I'm not finding any usages

mkalish: Nope, good catch. Leftover from an initial approach will remove.

prime-router/src/main/kotlin/fhirengine/translation/hl7/schema/ConfigSchemaReader.kt
```diff
@@ -38,16 +59,27 @@ object ConfigSchemaReader : Logging {
...
         }
+        return schemaList
+    }
 
-        // Now merge the parent with all the child schemas
-        val mergedSchema = mergeSchemas(schemaList)
-
-        if (!mergedSchema.isValid()) {
-            throw SchemaException("Invalid schema $schemaName: \n${mergedSchema.errors.joinToString("\n")}")
+    private fun fromUri(
```
luis-pabon-tf, line 67: this guy could use kdoc

luis-pabon-tf (commented):
Tests are passing. My only nit was some kdoc requests

brick-green (approved):
lgtm

prime-router/src/main/kotlin/fhirengine/translation/hl7/schema/ConfigSchemaReader.kt
```diff
@@ -38,16 +63,31 @@ object ConfigSchemaReader : Logging {
...
+        return schemaList
+    }
 
-        // Now merge the parent with all the child schemas
-        val mergedSchema = mergeSchemas(schemaList)
-
-        if (!mergedSchema.isValid()) {
-            throw SchemaException("Invalid schema $schemaName: \n${mergedSchema.errors.joinToString("\n")}")
+    /**
+     * Read a schema from a [URI] using the scheme to determine who to read the schema
```
arnejduranovic, line 72: This sentence doesn't make sense? `who` -> `how`?

prime-router/src/main/kotlin/fhirengine/translation/hl7/schema/ConfigSchemaReader.kt
```diff
@@ -38,16 +63,31 @@ object ConfigSchemaReader : Logging {
...
+     *
+     */
+    private fun fromUri(
+        schemaUri: URI,
+        schemaClass: Class<out ConfigSchema<out ConfigSchemaElement>>
+    ): List<ConfigSchema<*>> {
+        val schemaList = mutableListOf<ConfigSchema<*>>()
+        schemaList.add(readSchemaTreeUri(schemaUri, schemaClass = schemaClass))
+        while (!schemaList.last().extends.isNullOrBlank()) {
```
arnejduranovic, line 81: Is any of this circular dependency checking needed if `readSchemaTreeUri` also does it? Can we consolidate it to one place?

arnejduranovic (approved):
Looks good, just a couple comments regarding things to consider before merging.

prime-router/src/main/kotlin/fhirengine/translation/hl7/schema/ConfigSchemaReader.kt
mkalish: Yes I think can likely clean that up

prime-router/src/main/kotlin/fhirengine/translation/hl7/schema/ConfigSchemaReader.kt
mkalish: Will update

prime-router/src/main/kotlin/fhirengine/translation/hl7/schema/ConfigSchemaReader.kt
mkalish: Ah ok, so this actually touches on what we were chatting about last week.  This is the special handling to make sure that the extending clauses don't create a circular dependency.  I think you're likely right that we could clean this up, but it might be out of scope here?

PR #11217, opened 2023-09-05
#11221: Create and upload a blob in the send step and support downloading it

mkalish:
This PR starts uploading reports as part of the send step as well as updating the fetch report function to download the report from either send step or the batch step

Test Steps:
Before going through this login into the okta staging and copy your token out

1. Send a report through the legacy pipeline (or run the smoketests on master)
2. Find the id of a report from the send step (it will have the transport result column set)
3. Run
```
curl --location 'http://localhost:7071/api/history/report/{{id}} \
--header 'organization: development' \
--header 'Authorization: Bearer {token}'
```
and verify that it does not download.
4. Switch to this branch
5. Run 
```
curl --location 'http://localhost:7071/api/history/report/{{id}} \
--header 'organization: development' \
--header 'Authorization: Bearer {token}'
```
and make sure it downloads again
6. Find the parent batch report and run
```
curl --location 'http://localhost:7071/api/history/report/{{id}} \
--header 'organization: development' \
--header 'Authorization: Bearer {token}'
```
and verify it downloads.
7. Run the smoketest and find the ReportFile for a sent report.  Verify the bodyUrl field is set and it can be downloaded

## Changes
- Uploads a report to the blob store as part of the send step
- Update the `getReport` function to support downloading a sent or batch report, regardless of whether a file was uploaded during the send step

### Testing
- [x] Tested locally?
- [x] Added tests?

### Process

## Linked Issues
- #11221


prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -517,14 +518,26 @@ class ActionHistory(
...
     ) {
         if (isReportAlreadyTracked(sentReportId)) {
             error(
                 "Bug:  attempt to track history of a report ($sentReportId) " +
                     "we've already associated with this action"
             )
         }
+
+        if (header.content == null) {
+            error("Bug: attempt to track sent report with no contents")
```
mkalish, line 531: This should not be possible as the contents here are just what was sent to the receiver, but `content` is nullable.

prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -517,14 +518,26 @@ class ActionHistory(
...
+        }
+
+        val blobInfo = BlobAccess.uploadBody(
```
mkalish, line 534: This is the forward looking change so that we can later remove the code that handles the ReportFile from the send step not having a `bodyurl`

JFU-NAVA-PBC (commented):
played with a report send HL7 Full ELR, and observed that a blob folder 'none' created in addition to 'batch', 'route', 'ready', 'receive', 'translate', see screen shots:

Note: getReportById not call yet.

image

none->send-> fhir and hl7 files seen:

image

is that intended? if yes, what's the purpose...

JFU-NAVA-PBC:
> played with a report send HL7 Full ELR, and observed that a blob folder 'none' created in addition to 'batch', 'route', 'ready', 'receive', 'translate', see screen shots:
> 
> Note: getReportById not call yet.
> 
> image
> 
> none->send-> fhir and hl7 files seen:
> 
> image
> 
> is that intended? if yes, what's the purpose...

Read the Title again, it seems by design...

JFU-NAVA-PBC:
Observation:

Per test steps - step 2:
==============
1. .......
2. Find the id of a report from the send step (it will have the transport result column set)
3. Run
curl --location 'http://localhost:7071/api/history/report/{{id}} \
--header 'organization: development' \
--header 'Authorization: Bearer {token}'
and verify that it does not download.
==========

On master, when call history with a report ID whose "transport result" column is set with value:

http://localhost:7071/api/history/report/709ed3a1-6303-49e1-8082-5ed43e0c0ff2

Got 404 not found response, 

Debugging showed that 

WorkflowEngine.kt:

    fun fetchHeader(
        reportId: ReportId,
        organization: Organization,
        fetchBlobBody: Boolean = true
    ): Header {
        val reportFile = db.fetchReportFile(reportId, organization)
        val task = db.fetchTask(reportId) <========== look up task with report ID throw 

java.lang.IllegalStateException: Could not find 709ed3a1-6303-49e1-8082-5ed43e0c0ff2 that matches a task

JFU-NAVA-PBC:
Thinking about use cases for this API:

http://localhost:7071/api/history/report/{{ report-id }}

Possible use cases:

1. client call this API to download (get report content as part of response payload)
2. client call this API to just exam the envelop info of the report - no report content needed

Wondering the content download addition will make the 2nd use case expensive.

mkalish:
@JFU-NAVA-PBC Yes `none` is intended as the folders line up with what will be processing the file next.  In this case, there is no additional processing so it goes into `none`

mkalish:
@JFU-NAVA-PBC Anything else needed here?  I see you commented, but didn't leave an approval.

prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -517,14 +518,26 @@ class ActionHistory(
...
+            receiver.format,
+            header.content,
+            filename ?: UUID.randomUUID().toString(),
+            "send",
+            Event.EventAction.NONE
```
arnejduranovic, line 539: In reference to Qian's comments: This is what is causing the file to get uploaded to a "folder" called none, right?

prime-router/src/main/kotlin/azure/HistoryFunctions.kt
```diff
@@ -334,74 +334,107 @@ open class BaseHistoryFunction : Logging {
         reportIdIn: String,
         context: ExecutionContext
     ): HttpResponseMessage {
-        val authClaims = checkAuthenticated(request, context)
-            ?: return request.createResponseBuilder(HttpStatus.UNAUTHORIZED).build()
+        val claims = AuthenticatedClaims.authenticate(request)
+        val requestOrgName: String = request.headers["organization"]
+            ?: return HttpUtilities.bad(request, "Missing organization ins header")
```
arnejduranovic, line 339: `ins` -> `in`

prime-router/src/main/kotlin/azure/HistoryFunctions.kt
```diff
@@ -334,74 +334,107 @@ open class BaseHistoryFunction : Logging {
         reportIdIn: String,
         context: ExecutionContext
     ): HttpResponseMessage {
-        val authClaims = checkAuthenticated(request, context)
```
arnejduranovic, line 337: Let's just double check this auth update will not break this endpoint for certain users. For example, was this working for the *.reports scope before?

arnejduranovic (approved):

prime-router/src/main/kotlin/azure/ActionHistory.kt
mkalish: Yep, that's correct

prime-router/src/main/kotlin/azure/HistoryFunctions.kt
mkalish: I'll confirm, but also might just add just to be safe.

mkalish: FYI, [this](https://github.com/CDCgov/prime-reportstream/blob/a019095ebf334ad67cdd860bffe8130fe236646b/prime-router/src/main/kotlin/azure/HistoryFunctions.kt#L541) is where the check was previously looking and actually seemed to only accept being part of the org exactly; i.e.
`DHSender_xyz` would be allowed in but `DHSender_xyzAdmins` would not

JFU-NAVA-PBC (approved):
tested per steps listed in PR, and it does what AC required

prime-router/src/main/kotlin/azure/HistoryFunctions.kt
mkalish: Added *.*reports just in case

PR #11227, opened 2023-09-06
platform/jimfuqian/11161-Bug: updating kotlin jvm plugin version to 1.9.10 causes a test to fail

JFU-NAVA-PBC:
This PR:

Fix test case TokenFunctionTest: 'Test missing issuer' assertion logic when kotlin jvm plugin updated from 1.8.22 to 1.9.10.

Test Steps:
1. *Passing all tests: unit tests, integration tests, smoke tests*

## Changes
- *Bumped kotlin jvm plugin version from 1.8.22 to 1.9.10 in build.gradle.kts*
- *Refactored the failing test case so that the validation is done in RS code instead of depending on kotlin jvm plugin intrinsics NPE enriched behavior - which is gone after the version bump.*

### Testing
- [x] Tested locally?
- [ ] Added tests?

### Process

## Linked Issues
- Fixes #issue

## To Be Done
- #issue

    - new endpoint A
    - new endpoint B


JFU-NAVA-PBC:
call stacks to be attached to show NPE processing kotlin jvm 1.8.22 vs 1.9.10

JFU-NAVA-PBC:
TL;DR

with 1.8.22, the asserted "message issuer must not be null" is actually enriched NPE from kotlin-stdlib, it is not generated by io.jsonwebtoken or server2server authentication, the kotlin stdlib NPE handling behavior change break the assertion, and the fix is:

to perform the issuer validation in server2server authentication logic, the throw validation exception.

the fix is also keeping "issuer must not be null" just to match the assert expectation, but it needs not be so - we could used whatever validation error message later on, and wrap it in a NPE so that it bubble up through the same exception handling logic...

prime-router/src/main/kotlin/tokens/Server2ServerAuthentication.kt
```diff
@@ -75,6 +75,9 @@ class Server2ServerAuthentication(val workflowEngine: WorkflowEngine) : Logging
         // http://hl7.org/fhir/uv/bulkdata/authorization/index.html#signature-verification:~:text=Upon%20registration%2C%20the%20client%20SHALL%20be%20assigned%20a%20client_id%2C%20which%20the%20client%20SHALL%20use%20when%20requesting%20an%20access%20token
         // However, in order to be backwards compatible, the issuer claim can be either the name of the sender
         // or the name of the organization.
+        if (issuer == null) {
+            throw NullPointerException("issuer must not be null")
```
mkalish, line 79: Can we adjust the type on 68L so that kotlin knows that `issuer` can be null.

I would also just throw  throw a `Server2ServerAuthenticationException` here to be consistent

mkalish (changes requested):
Can we find the part of the changelog that mentions this change?  I'm a little concern because our test coverage is only ok that this will introduce other bugs, so I think we should try to understand the change a little better.

snesm:
supersedes https://github.com/CDCgov/prime-reportstream/pull/11048

prime-router/build.gradle.kts
```diff
@@ -30,7 +30,7 @@ import java.time.format.DateTimeFormatter
 import java.util.Properties
 
 plugins {
-    kotlin("jvm") version "1.8.22"
+    kotlin("jvm") version "1.9.10"
```
snesm, line 33: Maybe line 43 should also stay in [sync](https://plugins.gradle.org/plugin/org.jetbrains.kotlin.plugin.serialization):
`id("org.jetbrains.kotlin.plugin.serialization") version "1.8.22"`

we also have this defined at line 109
`val kotlinVersion = "1.9.10"`
just to add confusions.

JFU-NAVA-PBC: good catch, that's logical, kotlin jvm plugin 1.9.10 should go with serializer 1.9.10, let me change it and see...

JFU-NAVA-PBC:
> Can we find the part of the changelog that mentions this change? I'm a little concern because our test coverage is only ok that this will introduce other bugs, so I think we should try to understand the change a little better.

Can dig more

JFU-NAVA-PBC:
@mkalish will debug to find out why the kotlin jvm plugin version bump affected NPE processing 

from this:

checkNotNullExpressionValue:93, Intrinsics (kotlin.jvm.internal)
parseJwt$prime_router:78, Server2ServerAuthentication (gov.cdc.prime.router.tokens)
invoke0:-1, NativeMethodAccessorImpl (jdk.internal.reflect)

to this:

throwJavaNpe:45, Intrinsics (kotlin.jvm.internal)
checkNotNull:25, Intrinsics (kotlin.jvm.internal)
parseJwt$prime_router:81, Server2ServerAuthentication (gov.cdc.prime.router.tokens)
invoke0:-1, NativeMethodAccessorImpl (jdk.internal.reflect)
invoke:62, NativeMethodAccessorImpl (jdk.internal.reflect)

snesm (approved):

JFU-NAVA-PBC:
fyi, the behavior change might be related to: 
https://youtrack.jetbrains.com/issue/KT-57784

mkalish:
@JFU-NAVA-PBC Looks like the bug you linked is targeting 1.9.20?

mkalish:
@JFU-NAVA-PBC Just to confirm, the remaining work here is to revert the server changes and the update the test confirmation

JFU-NAVA-PBC:
> @JFU-NAVA-PBC Just to confirm, the remaining work here is to revert the server changes and the update the test confirmation

I like the approach that we validate for missing issuer in report stream logic ;-), and the validation error was asserted in test case.

the test case now still verify asserts something coming out of kotlin nullability handling, and hence it could break again in future kotlin compiler bump...

that says, this is beyond the bug fix scope of this PR, just pushed changes per your feedback.

mkalish (approved):
LGTM!  Smoketests all pass locally

prime-router/src/test/kotlin/azure/TokenFunctionTests.kt
```diff
@@ -221,8 +221,9 @@ class TokenFunctionTests {
         assertThat(error.get("error_description").textValue()).isEqualTo("malformed_jwt")
         assertThat(error.get("error_uri").textValue()).isEqualTo("$OAUTH_ERROR_BASE_LOCATION#malformed-jwt")
         verify {
+            // per feedback, just fix the assertion based on kotlin compiler NPE handling behavior
```
arnejduranovic, line 224: This comment is not very helpful without the context of this ticket. Do we really need this comment? My vote is to just remove it.

arnejduranovic (changes requested):

prime-router/src/test/kotlin/azure/TokenFunctionTests.kt
JFU-NAVA-PBC: might be good to log some thinking in the zenhub for future improve..., will remove it

arnejduranovic (approved):

PR #11231, opened 2023-09-06
NY UP Onboarding

victor-chaparro:
This PR adds NY receiver transforms needed to onboard NY to the Universal Pipeline

Test Steps:
1. Check out and build the branch
2. run `./prime multiple-settings set -e local -i settings/staging/0155-NY-up-onboarding.yml`
3. Send a test report using the attached Postman collection
[NY.postman_collection.txt](https://github.com/CDCgov/prime-reportstream/files/12541742/NY.postman_collection.txt)
4. Verify that the request was accepted
5. Open the generated HL7 file in the local sftp folder.
6. Verify the file name contains NY receiver transforms specified in `NY-receiver-transforms.yml`

## Changes
- Added NY receiver transforms
- Enabled 6 second offset back to default sender transforms

### Testing
- [x] Tested locally?
- [ ] Added tests?


prime-router/settings/organizations.yml
```diff
@@ -975,6 +975,36 @@
         filePath: ./upload
         credentialName: DEFAULT-SFTP
     # THIS RECEIVER IS NOT USED IN ANY TESTS AND CAN BE UPDATED TO YOUR HEARTS CONTENT
+    - name: ny-development
```
GilmoreA6, line 978: Typically we leave these local and don't merge to master as it has a tendency to get really cluttered after awhile.

prime-router/metadata/hl7_mapping/receivers/STLTs/NY/NY-receiver-transform.yml
```diff
@@ -0,0 +1,66 @@
...
+    resource: 'Bundle'
+    value: [ '"ECLRS"' ]
+
+  # MSH-6
+  - name: receiving-facility-namespace-id
+    condition: true
+    resource: 'Bundle'
+    value: [ '"NYSDOH"' ]
+
+  # OBX CWE-CE
```
GilmoreA6, line 29: Can we be a little more detail here to clarify that this is changing all OBX segments from CE to CWE data type? Also nitpicky, but can we move this below the MSH and SFT elements. Ideally, it is easier to read if the order of operations is in line with the order the elements appear in the output message.

GilmoreA6: Also, unsure why this is the case but when I remove this I get an error "Caused by: ca.uhn.hl7v2.HL7Exception: 'CE' in record 1 is invalid for version 2.7 at OBX-2(0)" and I'm not sure why it's targeting v2.7?

GilmoreA6, line 29: Can we be a little more detail here to clarify that this is changing all OBX segments from CE to CWE data type? Also nitpicky, but can we move this below the MSH and SFT elements. Ideally, it is easier to read if the order of operations is in line with the order the elements appear in the output message.

GilmoreA6: also unsure why, but when this is removed I get an error: "Caused by: ca.uhn.hl7v2.HL7Exception: 'CE' in record 1 is invalid for version 2.7 at OBX-2(0)" which is true but I'm not sure why it's targeting v2.7?

GilmoreA6 (changes requested):

prime-router/settings/organizations.yml
victor-chaparro: Is there another way of doing it so that whoever is reviewing the PR can test this? Do you create a separate file specific for each receiver?

victor-chaparro: I pushed a separate file with only the NY receiver settings and updated the PR test steps to use it.

prime-router/metadata/hl7_mapping/receivers/STLTs/NY/NY-receiver-transform.yml
victor-chaparro: We actually don't need this and added a new change that we do need with a comment explaining why we need it.

prime-router/settings/organizations.yml
GilmoreA6: Yep, I' usually just include the yml in the PR test steps so people can copy paste.

GilmoreA6 (approved):

PR #11364, opened 2023-09-14
HL7 OBX Mapping for OML messages

basiliskus:
This PR adds mapping to HL7 `OBX` fields for the `OML` message type. It also extracts the common field definitions for `ORU` and `OML` into a common file to be reused (`metadata/hl7_mapping/common/observation-result.yml`)

Test Steps:
2. Check that the output HL7 message has the `OBX` field. Subfields to look for are `OBX-1`, `OBX-2`, `OBX-3`, `OBX-4`, `OBX-5`, `OBX-6`, `OBX-11` and `OBX-14`

## Changes
- Extracted common field definitions in `ORU` and `OML` to `metadata/hl7_mapping/common/observation-result.yml`
- Added condition to `OBX-2` and `OBX-5` to support `NM`/`Quantity` data type
- Removed duplicated definition for `observation-identifier-code` field

### Testing
- [x] Tested locally
- [x] Integration test

## Linked Issues
[Create Outbound OML OBX Segment](https://github.com/CDCgov/trusted-intermediary/issues/495)


prime-router/metadata/hl7_mapping/common/observation-result.yml
```diff
@@ -0,0 +1,195 @@
...
+    condition: '%resource.value is Ratio'
+    value: [ '"SN"' ]
+    hl7Spec: [ '%{hl7OBXField}-2' ]
+
+  - name: result-value-type-nm
+    condition: '%resource.value is Quantity'
+    value: [ '"NM"' ]
+    hl7Spec: [ '%{hl7OBXField}-2' ]
+
+  - name: observation-identifier-code
```
jorg3lopez, line 39: Before, there were two of the same name that had a condition for when the value existed and one for when the value did not. Both pointed to the same schema. I took out the condition and left only this one pointing to the schema. All tests pass in my environment.

oslynn (approved):

oslynn:
Looks good, tested and everything is working as it should.  Thanks.

jimmyfagan (commented):
Looks good overall, just wondering about the necessity of both observation-result files instead of putting everything in the common one.

PR #11369, opened 2023-09-15
Updated java version

### LINKED ISSUE #11361: Update java-version file
### User Story
As a developer I want the .java-version to match what we are using incase I have multiple JDKs installed.

### Description/Use Case
During the java upgrade we left behind one file: prime-router/.java-version. 

N/A
### Acceptance Criteria
- Version string updated

thetaurean:
This PR updates the version in the `.java-version` file to 17.

Test Steps:
N/A

## Changes
N/A

### Testing
- [x] Tested locally?

## Linked Issues
- Fixes #11361


prime-router/.java-version
```diff
@@ -1 +1 @@
-11
+17
```
jack-h-wang, line 1:
```suggestion
17.0
```

jack-h-wang (approved):
All local testing passed.

As far as I can tell, this file is only used by `jEnv`, a tool that allows automatic management of `JAVA_HOME`. At least on my environment, `17` is not a valid Java version; matching numbers were `17.0` and `17.0.8` (the latter may be specific to the particular version of openJDK installed). In any case, the value of this file doesn't seem to have any effect outside of using jEnv.

prime-router/.java-version
thetaurean: Did jEnv fail to resolve 11 in your environment? It resolved to the latest point release of 11 for me (jdk-11.0.20.1+1). This may be desirable as we'll get bugfixes from new point releases for free.

Also, should we consider removing this file altogether?
We could also consider sdkman instead, a similar tool that's more common/modern and can manage other SDKs as well.

jack-h-wang: I'm actually not certain about 11 as I don't have openJDK 11 installed on my machine, and did not install it in my testing. This was the output of running `jenv versions` on my machine:
```
jenv: version `17' is not installed
  system
  17.0
  17.0.8
  20.0
  20.0.1
  openjdk64-17.0.8
  openjdk64-20.0.1
```
I have no objections to removing the file altogether.

thetaurean:
Ready for final checks @arnejduranovic

PR #11403, opened 2023-09-19
increase integration test timeout

JosiahSiegel:
This PR updates the Build Hub "test-integration" timeout from 30 minutes to 60 minutes.

Test Steps:
1. Create a PR and watch "Build Hub"

## Changes
- Change test integration workflow action from 30 minutes to 60 minutes timeout with no retries

## Linked Issues
- Fixes https://github.com/CDCgov/prime-devops/issues/107

### DIFF
--- a/prime-router/src/main/kotlin/cli/tests/TestReportStream.kt
+++ b/prime-router/src/main/kotlin/cli/tests/TestReportStream.kt
@@ -475,7 +475,7 @@ abstract class CoolTest {
     suspend fun pollForStepResult(
         reportId: ReportId,
         taskAction: TaskAction,
-        maxPollSecs: Int = 120,
+        maxPollSecs: Int = 840,
         pollSleepSecs: Int = 20,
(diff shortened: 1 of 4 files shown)


prime-router/src/main/kotlin/cli/tests/TestReportStream.kt
```diff
@@ -475,7 +475,7 @@ abstract class CoolTest {
     suspend fun pollForStepResult(
         reportId: ReportId,
         taskAction: TaskAction,
-        maxPollSecs: Int = 840,
+        maxPollSecs: Int = 120,
```
mkalish, line 478: So we've bumped this up in order to get the tests pass more reliably, can we leave this as is?

mkalish: Actually with the automatic retries, reducing the timeout makes sense to me

PR #11174, opened 2023-08-31
gap analysis hci

jimmyfagan:
This PR covers many mapping updates and additions to help prepare the UP for HCI.

Test Steps:
1. Update organization settings to have a receiver without any receiver transforms receive data for patientState "NY"
2. Send a report with the HL7 body found in the HCI folder in sharepoint
3. Compare the output with the original HL7 body, cross-referencing the spreadsheet of noted differences in sharepoint
4. Run integration tests

## Changes

- Updated mappings that either do not match the latest HL7 official mappings, or that caused issues with the HCI data
- Added mappings that had not been used before
- Cleaned up mappings that used excessive subfield-specific elements instead of well-established datatype mappings
- Adjusted transforms to preserve current behavior for CA
- Adjusted integration tests to more closely match our expectations with the updated mappings

### Testing
- [x] Tested locally?
- [x] Ran `./prime test` or `./gradlew testSmoke` against local Docker ReportStream container?

## Linked Issues
- Addresses #10963

## To Be Done
- #11317

PR #11406, opened 2023-09-19
11345: upgrade hapi fhir library

mkalish:
This PR upgrades the HAPI structures library which has the side effect of upgrading the HAPI core FHIR library.

Test Steps:
1. Run some files through the UP
2. Run the smoke tests

## Changes
- Upgrades the HAPI FHIR structures
- Fixes how FhirPathEngine is instantiated in FhirPathUtils
- Cleans up the translations to work with the new FhirPathEngine

### Testing
- [x] Tested locally?
- [x] Added tests?

### Process

## Linked Issues
- Fixes #issue

## To Be Done
- #issue

    - new endpoint A
    - new endpoint B


prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/ConstantResolver.kt
```diff
@@ -195,7 +195,7 @@ class FhirPathCustomResolver(private val customFhirFunctions: FhirPathFunctions?
         }
     }
 
-    override fun resolveReference(appContext: Any?, url: String?): Base? {
+    override fun resolveReference(appContext: Any?, url: String?, refContext: Base?): Base? {
```
mkalish, line 198: The signature for this changed in the new version

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirPathUtils.kt
```diff
@@ -26,10 +27,12 @@ import java.time.format.DateTimeParseException
  * Utilities to handle FHIR Path parsing.
  */
 object FhirPathUtils : Logging {
+
+    private val fhirContext = FhirContext.forR4()
     /**
      * The FHIR path engine.
      */
-    val pathEngine = FHIRPathEngine(SimpleWorkerContext())
+    val pathEngine = FHIRPathEngine(HapiWorkerContext(fhirContext, fhirContext.validationSupport))
```
mkalish, line 35: The reason for this change is interesting.  [This](https://github.com/hapifhir/org.hl7.fhir.core/commit/26027fa3d6d777fb46abc4e0e17b1f4cd04d6068#diff-9412a76f9f92124fb0838f3589aba8264ce989c0c400701f82ba6d16f822fa11R1805) PR introduced the `isKnownType` function which calls `worker.fetchTypeDefintion` which looks in the `structures`; this value is empty for the SimpleWorkerContext.

I actually think this is technically the correct worker to use anyway

prime-router/build.gradle.kts
```diff
@@ -820,7 +820,9 @@ dependencies {
     }
     implementation("org.yaml:snakeyaml:2.0")
     implementation("io.github.linuxforhealth:hl7v2-fhir-converter:1.0.19")
-    implementation("ca.uhn.hapi.fhir:hapi-fhir-structures-r4:6.4.0")
+    implementation("ca.uhn.hapi.fhir:hapi-fhir-structures-r4:6.8.2")
+    // https://mvnrepository.com/artifact/ca.uhn.hapi.fhir/hapi-fhir-caching-caffeine
+    implementation("ca.uhn.hapi.fhir:hapi-fhir-caching-caffeine:6.8.2")
```
mkalish, line 825: This new dependency is required by the `HapiWorkerContext` now in use

prime-router/metadata/hl7_mapping/ORU_R01/base/observation-request.yml
```diff
@@ -13,17 +13,17 @@ elements:
     schema: ../../common/organization-identifier
 
   - name: observation-request-date-time
-    condition: '%resource.effective is dateTime'
+    condition: '%resource.effective.exists() and %resource.effective.is(dateTime)'
```
mkalish, line 16: Almost all the changes boil down to a change in how most FhirPathEngine functions get evaluated.  Previously, something like `is` would return a boolean when the focus resource is empty, now it will actually return an empty array which can't be used as a condition

prime-router/metadata/hl7_mapping/common/patient.yml
```diff
@@ -156,22 +156,22 @@ elements:
         false: N
 
   - name: patient-multiple-birth-order
-    condition: '%resource.multipleBirth.exists() and %resource.multipleBirth is Integer'
+    condition: '%resource.multipleBirth.exists() and %resource.multipleBirth is integer'
```
mkalish, line 159: across the board, all the primitive types are lower cased to start

prime-router/metadata/hl7_mapping/ORU_R01/base/observation-result.yml
```diff
@@ -155,29 +155,29 @@ elements:
     hl7Spec: [ '%{hl7OBXField}-19' ]
 
   - name: performing-organization-name-pracrole
-    condition: '%observation.performer.resolve() is PractitionerRole'
-    resource: '%resource.performer.resolve().organization.resolve()'
+    condition: '%observation.performer.resolve().ofType(PractitionerRole).exists()'
```
jimmyfagan, line 158: What determined whether you used `x.y.exists and x.y is Z` vs `x.y.ofType(Z).exists`? They seem similar, maybe the latter makes more sense for repeating components? Wasn't sure so wanted to check if you had more explanation.

prime-router/metadata/hl7_mapping/common/datatype/ce-coded-with-exceptions.yml
```diff
@@ -18,22 +18,22 @@ elements:
     hl7Spec: [ '%{cweFieldPath}-2' ]
   - name: identifier-system-url
     resource: '%resource.coding.where(extension(%`rsext-code-index-name`).value = "identifier").system'
-    condition: '%resource.value.startsWith("http")'
+    condition: '%resource.value.exists() and %resource.value.startsWith("http")'
```
jimmyfagan, line 21: I guess `startsWith` has the same issue now as `is`?

prime-router/metadata/hl7_mapping/receivers/STLTs/CA/ca-specimen.yml
```diff
@@ -53,32 +53,32 @@ elements:
...
 
   - name: specimen-received-time
-    condition: '%resource.receivedTime is dateTime'
+    condition: '%resource.receivedTime.exists() and %resource.receivedTime is dateTime'
     value: [ '%resource.receivedTime.changeTimezone(%timezone)' ]
     hl7Spec: [ '%{hl7SpecimenFieldPath}-18' ]
 
   - name: specimen-received-time-diagnostic
-    condition: '%resource.receivedTime.exists().not() and %resource.collection.collected is dateTime'
+    condition: '%resource.receivedTime.exists().not() and %resource.collection.collected.is(dateTime)'
```
jimmyfagan, line 82: Was this change meaningful?

jimmyfagan (commented):
Looks good! I just left a handful of comments, mostly questions for my own understanding.

prime-router/metadata/hl7_mapping/common/datatype/ce-coded-with-exceptions.yml
mkalish: Yes, anything that returns a boolean is impacted

prime-router/metadata/hl7_mapping/receivers/STLTs/CA/ca-specimen.yml
mkalish: It was not.  I'll switch it back.  I think this is a good topic to roll into formatting too  to write that consistently

prime-router/metadata/hl7_mapping/ORU_R01/base/observation-result.yml
mkalish: So this one was an interesting and maybe deserves a second look.  For some of the integration tests, `performer` was an array with two elements, one PractictionerRole and one Organization.  This check was to make sure that the condition only evaluates to true if we can get a PractitionerRole

jimmyfagan: Ah okay, guessing this used to work because the resource would only be valid if the condition was true anyway. As long as integration tests work, this seems fine, though yea I am curious about a few things here, might add another comment or two.

prime-router/metadata/hl7_mapping/ORU_R01/base/observation-result.yml
```diff
@@ -155,29 +155,29 @@ elements:
...
+    resource: '%resource.performer.resolve().ofType(PractitionerRole).organization.resolve()'
     schema: ../../common/datatype/xon-organization
     constants:
       hl7OrgField: '%{hl7OBXField}-23'
 
   - name: performing-organization-name-non-pracrole
-    condition: '%observation.performer.resolve().is(PractitionerRole).not()'
+    condition: '%observation.performer.resolve().ofType(PractitionerRole).exists().not()'
```
jimmyfagan, line 165: When performer has both a `PractitionerRole` and an `Organization`, this seems like it would cause issues. I'm thinking what we really want is more like this
```suggestion
    condition: '%resource.exists() and %resource.is(PractitionerRole).not()'
```
or
```suggestion
    condition: '%resource.exists() and %resource.is(Organization)'
```

prime-router/metadata/hl7_mapping/ORU_R01/base/observation-result.yml
```diff
@@ -155,29 +155,29 @@ elements:
...
+    condition: '%observation.performer.resolve().ofType(PractitionerRole).exists()'
+    resource: '%resource.performer.resolve().ofType(PractitionerRole).organization.resolve().address'
     constants:
       hl7AddressField: '%{hl7OBXField}-24'
     schema: ../../common/datatype/xad-extended-address
     resourceIndex: contactIndex
 
   - name: performing-organization-address-non-pracrole
-    condition: '%observation.performer.resolve().is(PractitionerRole).not()'
+    condition: '%observation.performer.resolve().ofType(PractitionerRole).exists().not()'
```
jimmyfagan, line 180: Similar issue to above looks like it would exist here, but it's less straightforward to resolve. Maybe the existence of a performer.address is enough to tell that the performer is an Organization, so this would work?
```suggestion
    condition: '%resource.exists and %observation.performer.resolve().ofType(Organization).exists()'
```

prime-router/metadata/hl7_mapping/ORU_R01/base/observation-result.yml
mkalish: Yes, I think you are correct and the current fact that it was working relied on the ordering of the elements and not the conditions.  Will update

jimmyfagan: Did some testing on this in your branch today and got some clarity. OBX.16/OBX.23/OBX.24 work fine on HL7->FHIR->HL7 in the branch, but it's a bit messed up on HL7>FHIR. Basically prior to the change you commented on [here](https://github.com/CDCgov/prime-reportstream/pull/11174/files#r1332117929) (which isn't merged into your branch yet), OBX.23/24 were being added to the `PractitionerRole` for OBX.16 so we were populating all three fields from the same `performer`, even though they should be pulled from one `PractitionerRole` `performer` and one `Organization` `performer` respectively. This will basically need to be adjusted once the latest from master gets pulled in here. The good news is that the hci Translation Tests do cover this case, so you won't have to concoct a test specially for this.

jimmyfagan (commented):
Left a comment to follow-up from some discussion last week. I think I'll need to re-review this once the latest from `master` gets merged in, and we might need to adjust some conditions for OBX; I'd be happy to huddle on that when the time comes.

mkalish:
@jimmyfagan Yep, you were right the HCI test needed to be resolved on this branch.  The fix was to check for PractictionerRole and an organization before falling back to the Organization

prime-router/metadata/hl7_mapping/common/note.yml
mkalish, line 1: I did not see anything in this file that would make it only usable in ORU_R01 messages so I've moved here and removed the duplicative observation-result

prime-router/metadata/hl7_mapping/common/observation-result.yml
```diff
@@ -165,30 +165,30 @@ elements:
     hl7Spec: [ '%{hl7OBXField}-19' ]
 
   - name: performing-organization-name-pracrole
-    condition: '%observation.performer.resolve() is PractitionerRole'
-    resource: '%resource.performer.resolve().organization.resolve()'
+    condition: '%observation.performer.resolve().ofType(PractitionerRole).exists() and %observation.performer.resolve().ofType(PractitionerRole).organization.exists()'
```
mkalish, line 168: @jimmyfagan Can you double check me on this?  The logic is not that we will attempt to populate OBX-23 using the PractionerRole from the performer array if there is a PractionerRole and it has an organization otherwise we'll use the Organization

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirPathUtils.kt
```diff
@@ -99,8 +102,14 @@ object FhirPathUtils : Logging {
             val expressionNode = parsePath(expression)
             val value = if (expressionNode == null) emptyList()
             else pathEngine.evaluate(appContext, focusResource, bundle, bundle, expressionNode)
-            if (value.size == 1 && value[0].isBooleanPrimitive) (value[0] as BooleanType).value
-            else {
+            if (value.size == 1 && value[0].isBooleanPrimitive) {
+                (value[0] as BooleanType).value
+            } else if (focusResource.isEmpty) {
```
mkalish, line 107: I can back this change out if folks don't agree.  I added this since even though I do think that the conditions should always check for existence there are a **lot** of places where they don't, but there is no issue from a test perspective because the resource is always present.

This serves as a catch all to make sure that those places don't become a problem in the future.

prime-router/metadata/hl7_mapping/common/observation-result.yml
jimmyfagan: These look okay, and clearly work based on the Translation Tests. They are a bit hard to understand without looking at the mappings to be honest though. Here and in `performing-organization-address-pracrole`, I might recommend something like the following, though optionally since in practice they should produce the same results:
```suggestion
    condition: '%observation.performer.resolve().ofType(PractitionerRole).exists() and %observation.performer.resolve().ofType(PractitionerRole).code.coding.code = "MDIR"'
```

prime-router/metadata/hl7_mapping/common/observation-result.yml
```diff
@@ -165,30 +165,30 @@ elements:
...
+    resource: '%resource.performer.resolve().ofType(PractitionerRole).organization.resolve()'
     schema: ./datatype/xon-organization
     constants:
       hl7OrgField: '%{hl7OBXField}-23'
 
   - name: performing-organization-name-non-pracrole
-    condition: '%observation.performer.resolve().is(PractitionerRole).not()'
-    resource: '%resource.performer.resolve()'
+    condition: '%observation.performer.resolve().ofType(PractitionerRole).exists().not() or %observation.performer.resolve().ofType(PractitionerRole).organization.exists().not()'
```
jimmyfagan, line 175: This again is fine and works, but it might be simpler to go with the following here and in `performing-organization-address-non-pracrole` (also optional):
```suggestion
    condition: '%observation.performer.resolve().ofType(Organization).exists()'
```

prime-router/metadata/hl7_mapping/common/observation-result.yml
```diff
@@ -165,30 +165,30 @@ elements:
...
+    resource: '%resource.performer.resolve().ofType(Organization)'
     schema: ./datatype/xon-organization
     constants:
       hl7OrgField: '%{hl7OBXField}-23'
 
   - name: performing-organization-address-pracrole
-    condition: '%observation.performer.resolve() is PractitionerRole'
+    condition: '%observation.performer.resolve().ofType(PractitionerRole).exists() and %observation.performer.resolve().ofType(PractitionerRole).organization.exists()'
     resource: '%resource.performer.resolve().organization.resolve().address'
```
jimmyfagan, line 183: Guessing you meant to change this too?
```suggestion
    resource: '%resource.performer.resolve().ofType(PractitionerRole).organization.resolve().address'
```

jimmyfagan (approved):
Looks like things work here, I left some suggestions for some improvements to conditions for `observation.performer`, but nothing critical, and nothing that should block us from getting this library update wrapped up. Marking approved, but available for follow-up discussion and re-review as needed.

prime-router/metadata/hl7_mapping/common/observation-result.yml
mkalish: This is condition came down to which one should have priority, the current check will only use `Organization` if nothing is set via practionerRole

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirPathUtils.kt
mkalish: Yes, I think that all sounds right, but we can actually further clean it up by checking if the evaluated value is empty or not.  I'll push that change

arnejduranovic (approved):

PR #11419, opened 2023-09-20
11288 enable truncation for any HL7 message type

jalbinson:
This PR updates how HL7 truncation is broken up between the Covid pipeline and the Universal Pipeline.

For the Covid pipeline we receive an HL7 element to be passed in and we use that to look up the path.
For the Universal pipeline we receive the path so no need to lookup. This allows us to support other message types without having to update the element to path table currently present in HL7Utils.

Test Steps:
1. `./gradlew test`
2. `./gradlew testIntegration`

E2E Test Steps

1. Update/add the following settings to organizations.yml
```
schemaName: "metadata/hl7_mapping/ORU_R01/ORU_R01-base"
truncateHDNamespaceIds: true
truncateHl7Fields: "OBX-23-1,ORC-21-1,OBR-16-1,OBR-16-3,ORC-12-1,ORC-12-3,PID-5-1,PID-5-2,PID-5-3"
```
2. Send the attached FHIR bundle to the Reports API:
[Simple Report.postman_collection.txt](https://github.com/CDCgov/prime-reportstream/files/11165604/Simple.Report.postman_collection.txt)
This sample contains all the fields that should be truncated and you can search for them by looking for "truncatethis"
3. Verify that the request was accepted
4. Open the generated HL7 file in the local sftp folder.
5. Verify all the fields were truncated by making sure the HL7 message doesn't include the text "truncatethis"
6. Run integration tests and make sure they pass

## Changes
- Change `HL7Truncator` to be an interface with shared behavior
- Added new `CovidPipelineHL7Truncator` and `UniversalPipelineHL7Truncator` classes which extend `HL7Truncator`
- updated many tests to check that different message types are handled

### Testing
- [x] Tested locally?

## Linked Issues
- Fixes [11288](https://app.zenhub.com/workspaces/engagement-6166edbd409257001e09f1b2/issues/gh/cdcgov/prime-reportstream/11288)


jalbinson:
E2E steps confirmed to work (although only tested for `ORU_R01`)

prime-router/src/test/kotlin/fhirengine/translation/hl7/FhirToHl7ConverterTests.kt
```diff
@@ -561,4 +562,54 @@ class FhirToHl7ConverterTests {
...
+
+        val element = ConverterSchemaElement(
+            "name",
+            required = true,
+            hl7Spec = listOf("/ORDER/OBSERVATION_REQUEST/OBSERVATION/OBX-18-1")
+        )
+        converter.setHl7Value(element, value, customContext)
+
+        val shouldBeTruncated = terser.get("/ORDER/OBSERVATION_REQUEST/OBSERVATION/OBX-18-1")
+        assertEquals(shouldBeTruncated.length, 199)
```
adegolier, line 613: 199? Comment says 194, though?

prime-router/src/main/kotlin/fhirengine/translation/hl7/HL7Truncator.kt
```diff
@@ -13,20 +14,22 @@ import kotlin.math.min
 /**
  * The shared HL7 truncation logic used in both the Covid Pipeline and the Universal Pipeline
  */
-class HL7Truncator {
+sealed interface HL7Truncator {
```
adegolier, line 17: Is there a reason we're splitting the truncators by pipeline instead of just adding the UP functionality? Because UP HL7 truncation should be able to handle the same truncation as the Covid pipeline, right?

prime-router/src/test/kotlin/fhirengine/translation/hl7/FhirToHl7ConverterTests.kt
jalbinson: whoops, I need to double check on some comments as much of the unit tests got copy/pasted with some values swapped

prime-router/src/main/kotlin/fhirengine/translation/hl7/HL7Truncator.kt
jalbinson: I split them out based on the types of parameters we were getting from each.

In the Covid pipeline we were getting an HL7 element like `PID-5-1` and then using a hardcoded map to find the path `/PATIENT_RESULT/PATIENT/PID` ([can see existing function here](https://github.com/CDCgov/prime-reportstream/blob/22c49438af032a57d8688427ac635cef29396571/prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/HL7Utils.kt#L101)).

For UP it directly hands us the full path `/PATIENT_RESULT/PATIENT/PID-5-1` so no need to look it up. This opens us up to use other message types as well by depending on the spec already handled the `Terser`.

prime-router/src/test/kotlin/fhirengine/translation/hl7/FhirToHl7ConverterTests.kt
jalbinson: 199 is right I'll change the comment

Screenshot 2023-09-20 at 1 26 21 PM

victor-chaparro (approved):
Tested it locally and changes look good!

jalbinson:
@victor-chaparro 

Were you able to try with another message type (besides `ORU_R01`)?

victor-chaparro:
@jalbinson I did I can show you how if you want, but all I did is change schemaName to `schemaName: "metadata/hl7_mapping/OML_O21/OML_O21-base"` and also changed the qualityFilter to `- "(%hl7MessageType.exists() and %hl7MessageType = 'O21')"`

jalbinson:
@victor-chaparro Great! Just wanted to make sure it worked end-to-end rather than just unit tests.

Issue #10998, opened 2023-08-21
Write proposal to allow the outer value of `%resource` to be used in an element, even when the element overrides that with a new value.


jimmyfagan:
### User Story
As a RS developer, I want to be able to use the outer value of `%resource` in an element, even when the element overrides that with a new value.

### Description/Use Case
In HL7 mapping files, there is often a focus resource that is used for the whole file. That recourse can be accessed with the constant `%resource`. However, within a given element, if a new resource is specified (with `resource: [some FHIR path expression]`), then in the rest of that element other components (like `condition`) will use the new resource value when they use that constant `%resource`. Since some conditions do depend on the original resource value, this has resulted in many schemas (observation-result.yml for example) having a separate constant (`observation` in the case of observation-result), which is intended to store the same value as `resource`. But as resource expressions become more complicated, populating a separate constant with the right value becomes increasingly difficult. If we can use a separate constant or custom function to facilitate the access to the original resource constant within an element, then there won't be a need for managing the separate constants in these cases.

This ticket should document a proposal and get feedback on that proposal to enable future tickets to implement this have a clear plan.

### Risks/Impacts/Considerations
Existing schemas should not break! This means we shouldn't change behavior for any existing functions/constants.

### Dev Notes
Some ideas that fave floated so far are:

- use `%outerResource` to capture the original resource coming into an element.
- use function `parent()` to walk backwards in the path to a given value (%resource.parent() could then be used to step backwards)

### Acceptance Criteria
- [x] A proposal for this work has been reviewed and merged to the repo
- [ ] Follow-up stories have been created

bishoyayoub:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/10998?planning-poker) with Zenhub @arnejduranovic @brick-green @jack-h-wang @JessicaWNava @JFU-NAVA-PBC @jimmyfagan @luis-pabon-tf @thetaurean

Andrey-Glazkv:
Follow up ticket was created and worked on, then also demoed (#11543)

Andrey-Glazkv:
Follow up ticket was created and worked on, then also demoed (#11543)

PR #11436, opened 2023-09-20
First draft of outer resource design

JessicaWNava:
This PR proposes a solution for being able to access the outer resource in hl7 mapping files.

Test Steps:
N/A

## Changes
-  Creates a proposal for being able to access the outer resource in hl7 mapping files.
- Created https://app.zenhub.com/workspaces/platform-6182b02547c1130010f459db/issues/gh/cdcgov/prime-reportstream/11543 as a result of the findings

#10998


prime-router/docs/design/design/outer-resource-design.md
```diff
@@ -0,0 +1,78 @@
...
+
+We want to have a way to access that outer resource, the one initially defined, so that in places like these we can
+reference the parent a
+nd some of its children without losing the value of the initial resource.
+
+There are two ways we can go about this, and they are not mutually exclusive. 
+
+## Proposals
+### Outer Resource
+This is, by far, the less difficult option. In `FhirToHl7Converter`, we would set a constant called `outerResource` to
```
mkalish, line 55: Smallest of nits, but curious what your thoughts on calling this `schemaResource` would be?  That would allow it to work for the root schema too (where schemaResource = Bundle)

mkalish (commented):
Awesome stuff!

prime-router/docs/design/design/outer-resource-design.md
JessicaWNava: Either way works for me, thoughts @jimmyfagan?

brick-green (approved):
The proposal makes sense to me and I think it will work. Nice call out on the .parent() and the variable solution that isn't sufficient.

arnejduranovic (commented):
This looks good, can we create a ticket to implement this and link it here?

arnejduranovic (approved):

PR #11463, opened 2023-09-21
HL7 NK1 Mapping for OML

basiliskus:
This PR adds mapping to HL7 `NK1` fields for the `OML` message type

Test Steps:
2. Check that the output HL7 message has the `NK1` field

## Changes
- Updated `NK1-3` to use `CWE` schema
- Updated OML input and expected output for integration test

### Testing
- [ ] Tested locally
- [ ] Integration test

## Linked Issues
[Create Outbound OML NK1 Segment](https://github.com/CDCgov/trusted-intermediary/issues/491)


prime-router/metadata/hl7_mapping/OML_O21/datatype/ce-coded-with-exceptions.yml
```diff
@@ -0,0 +1,141 @@
...
+    value: [ '%resource.getCodingSystemMapping()' ]
+    hl7Spec: [ '%{cweFieldPath}-3' ]
+  - name: identifier-system-url
+    resource: '%resource.coding[0]'
+    condition: '%resource.system.value.startsWith("http") and %resource.extension(%`rsext-code-index-name`).exists().not()'
+    value: [ '%resource.system.getCodingSystemMapping()' ]
+    hl7Spec: [ '%{cweFieldPath}-3' ]
+  - name: identifier-system-url
+    resource: '%resource.coding[0]'
+    condition: '%resource.system.startsWith("http") and %resource.extension(%`rsext-code-index-name`).exists().not()'
```
jorg3lopez, line 31: We were unable to reuse the original code under `common/datatype/ce-coded-with-exception.yml` because the code for `identifier-system-url` breaks the integration test. The `%resource.system` is used by other messages so if the original file is shared, this will produce unwanted behavior such that other messages other than OML, will have their nk1.3 affected. In order to maintain the changes to only the OML message, the changes were made at the OML package level.

mkalish: Just to confirm, this change is because you want to `%resource.system` and not %resource.system.getCodingSystemMapping`?

Could you maybe solve that by adding an entry [here](https://github.com/CDCgov/prime-reportstream/blob/11f98ae573e87eab3ad2dcc286cc4043cb8774e2/prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/CustomFHIRFunctions.kt#L267)?

prime-router/src/testIntegration/resources/datatests/FHIR_to_HL7/sample_OML_20230831-0001.fhir
mkalish, line 1: Just as an FYI, we try to keep a version of these files committed without the unique identifiers.  See [this](https://github.com/CDCgov/prime-reportstream/blob/11f98ae573e87eab3ad2dcc286cc4043cb8774e2/prime-router/src/testIntegration/resources/datatests/README.md#L56).  It helps to see what is happening in the diff

prime-router/metadata/hl7_mapping/OML_O21/datatype/ce-coded-with-exceptions.yml
jorg3lopez: We made changes to the coding mapping method, thanks for the input.

prime-router/src/testIntegration/resources/datatests/FHIR_to_HL7/sample_OML_20230831-0001.fhir
jorg3lopez: Thanks!

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/CustomFHIRFunctions.kt
```diff
@@ -273,6 +273,7 @@ object CustomFHIRFunctions : FhirPathFunctions {
         HL70006("http://terminology.hl7.org/CodeSystem/v2-0006", "HL70006"),
         HL70136("http://terminology.hl7.org/ValueSet/v2-0136", "HL70136"),
         HL70078("http://terminology.hl7.org/CodeSystem/v2-0078", "HL70078"),
+        HL70131("http://terminology.hl7.org/CodeSystem/v2-0131", "HL70131"),
```
mkalish, line 276: I think it's tangential here, but it looks like in FHIR the valueSet for `RelatedPerson.relationship` actually includes more values that what is in HL70131.

https://chat.fhir.org/#narrow/stream/179166-implementers/topic/relationship.20in.20Patient.2Econtact.20and.20RelatedPerson

PR #11475, opened 2023-09-22
Added function to download Blobs as binary data + retry mechanism to Blob downloads

### LINKED ISSUE #10487: Add a function to download a blob as BinaryData with retries
### User Story
As a report stream developer, I would like a function that downloads a blob as an `BinaryData` retrying if I/O errors are encountered

### Description/Use Case
Currently, the only download function generates a bytearray, but this can be simplified for downstream uses by returning `BinaryData`.  Additionally, since it's possible to encounter an I/O error we'll want to add some retry functionality

### Dev Notes
- [downloadContent](https://learn.microsoft.com/en-us/java/api/com.azure.storage.blob.specialized.blobclientbase?view=azure-java-stable#com-azure-storage-blob-specialized-blobclientbase-downloadcontent()) can be used
- The number of retries should be configurable 
- Check to see if there is a library that can implement the retry for us

### Acceptance Criteria
- [ ] A download method that returns `BinaryData`
- [ ] The method retries a configurable number of times when I/O errors are encountered

thetaurean:
This PR adds a retry mechanism to the existing blob download function and introduces a new download function that outputs BinaryData instead of a ByteArray. Number of retries is configured by environment variable `AzureBlobDownloadRetryCount`

Test Steps:
1. Running unit tests is sufficient

## Changes
- Introduce new env var `AzureBlobDownloadRetryCount` to control
- Rename existing function `BlobAccess.downloadBlob` -> `BlobAccess.downloadBlobAsByteArray`
- Add new function `BlobAccess.downloadBlobAsBinaryData`

### Testing
- [x] Tested locally?
- [x] ~~Added~~ Modified tests?

### Process

## Linked Issues
- Fixes #10487 

## To Be Done
- Coordinate with DevOps


prime-router/src/main/kotlin/azure/BlobAccess.kt
```diff
@@ -115,6 +118,13 @@ class BlobAccess() : Logging {
             return System.getenv(blobConnEnvVar)
         }
 
+        /**
+         * Obtain the download retry value from the given environment.
+         */
+        fun getBlobDownloadRetry(blobDownloadRetryVar: String = defaultBlobDownloadRetryVar): Int {
```
mkalish, line 124: I wonder if we need to provide a default when the env var is not set?

prime-router/src/main/kotlin/azure/BlobAccess.kt
```diff
@@ -149,21 +159,46 @@ class BlobAccess() : Logging {
...
+                null
+            ) }
             logger.debug("BlobAccess Finished download for blobUrl $blobUrl")
             return stream.toByteArray()
         }
 
+        /**
+         * Download the blob at the given [blobUrl] as BinaryData
+         */
+        fun downloadBlobAsBinaryData(blobUrl: String,  retries: Int = getBlobDownloadRetry()): BinaryData {
```
mkalish, line 184: Maybe add a small test for this?

mkalish (commented):
Changes are looking good, I'll take another pass after the testing steps are updated

prime-router/src/main/kotlin/azure/BlobAccess.kt
thetaurean: This should be covered [here](https://github.com/CDCgov/prime-reportstream/blob/bd7ac9c1b41209413b581b8649f92285011d401b/prime-router/src/test/kotlin/azure/BlobAccessTests.kt#L243C76-L243C76).

Let me know if you had something else in mind or I'm missing something.

mkalish: Ah totally missed that, that seems fine

prime-router/src/main/kotlin/azure/BlobAccess.kt
mkalish: Yep, that makes sense to me out of an abundance of caution

thetaurean: This has been resolved 👌 and will also handle cases where the env var is garbage.

mkalish (commented):
Changes look good just running the smoketests before approving

prime-router/build.gradle.kts
```diff
@@ -567,6 +567,7 @@ tasks.azureFunctionsRun {
 
     val env = mutableMapOf<String, Any>(
         "AzureWebJobsStorage" to devAzureConnectString,
+        "AzureBlobDownloadRetryCount" to 5,
```
mkalish, line 570: So I ran the smoketests with commented out and everything seemed to be fine (I assume because of the default in the code), so not strictly required, but seems fine to keep.  I would just maybe leave a comment here or in BlobAccess to keep these the same.

mkalish (approved):
LGTM!  All my testing passed

thetaurean:
Ready for final checks @arnejduranovic

PR #11087, opened 2023-08-25
Implemented parallel downloads in batch step

thetaurean:
This PR adds the use of coroutines in the batch step to facilitate parallel download and processing of report files.

Test Steps:

1. Set debug level logging in `src/main/resources/log4j2.xml` (line 31)
2. Run smoke tests on this feature branch
3. Observe logs for `Starting download for blobUrl` and corresponding `Finished` messages. You should see multiple downloads start in parallel.

## Changes
- BatchFunction.batchUniversalData() and WorkflowEngine.createHeader() now make use of `runBlocking`, `async`, and `awaitAll` to enable parallel download and processing of reportfiles.
- Added debug level logging for individual downloads; info level logging for aggregated duration of downloads

### Testing
- [x] Tested locally?
- [x] Ran `./prime test` or `./gradlew testSmoke` against local Docker ReportStream container?
- [ ] Added tests?

## Linked Issues
- Fixes #issue

## To Be Done
- Logging
- Test instructions (in this PR)

PR #11483, opened 2023-09-22
Fix repeated download in Batch step

### LINKED ISSUE #11370: Batch step downloads ReportFiles twice
### User Story
As any RS stakeholder, I want to ensure reportfiles are only downloaded once during batch so that the batch step executes more efficiently.

### Description/Use Case
During the batch step we are currently downloading the ReportFile two separate times. We need to refactor the 

- The code was recently updated to parallelize downloads, that may need to be refactored as parto f this.

### Dev Notes
- Follow-up to #10692

### Acceptance Criteria
- ReportFiles are only downloaded once during the batch step
- Code is refactored as necessary to maintain readability and DRYness

thetaurean:
This PR refactors downloads in the batch step, moving all download code into a single function that is called one time in the BatchFunction. As a result, superfluous downloads in batchUniversalData and uploadReport have been removed to avoid redownloading data already in memory.

It also:
- Adds a test checking both legacy and universal pipeline logical branches
- Cleans up results of #11087

Test Steps:
1. Running tests is sufficient

## Changes
- Added function `WorkflowEngine.createDeepHeaders()` to create headers, download content, and log errors
- Remove download in `WorkflowEngine.createReport()` and use `header.content!!` instead
- Remove download in `BatchFunction.batchUniversalData()` and use `header.content!!` instead
- Updated `BatchFunctionTests`
  - added 1 new test to check for repeated downloads in both pipelines
  - removed download verification from UP batch tests (since download no longer occurs there)

### Testing
- [x] Tested locally?
- [x] Added and updated tests?

## Linked Issues
- Fixes #11370 

## TODO
- #11636 
- #11639


prime-router/src/main/kotlin/azure/BatchFunction.kt
```diff
@@ -103,7 +103,9 @@ class BatchFunction(
                     workflowEngine.recordAction(actionHistory)
                 }
             } else {
-                workflowEngine.handleBatchEvent(event, maxBatchSize, backstopTime) { headers, txn ->
+                workflowEngine.handleBatchEvent(event, maxBatchSize, backstopTime, !receiver.topic.isUniversalPipeline)
```
arnejduranovic, line 106: I'm not too sure about this solution, it seems kind of unmaintainable to me, given the possibility for having to support other pipelines in the future. Generally, we should stay away sprinkling pipeline-based conditionals throughout the code. We have the` !receiver.topic.isUniversalPipeline` check here and then in the trailing lambda we do:
```
if (receiver.topic.isUniversalPipeline) {
    batchUniversalData(validHeaders, actionHistory, receiver, txn)
}
```

Why don't we just update `batchUniversalData` to NOT download the files again? Furthermore, why does handleBatchEvent only need to run for universal pipeline? There seem to be deeper issues here that we should look into/resolve.

arnejduranovic (changes requested):

mkalish (commented):
Changes look good to me!  I'll pull down and test with a final review after the conflicts are resolved.

mkalish (approved):
LGTM!  Ran the smoke tests and everything passed locally

arnejduranovic (approved):

Issue #10849, opened 2023-08-10
Make the code "enhancing" the bundle more maintainable and extensible


JessicaWNava:
### User Story
As a RS developer, I want a maintainable and extensible way of handling things like enhancing the bundle metadata and handling the birthtime. 

### Description/Use Case
There may be other cases where post translation to FHIR we need to parse the HL7 message, do some work on it to get a value, and then update the output FHIR message before completing the conversion. We want to figure out a more maintainable and extensible way to handle this.

### Dev Notes
Arnej's idea: Instead of updating translate every time we need to do something similar, I'm imagining a separate class that handles this. The class could be called something like `BundleEnhancer` and usage could be `BundleEnhancer.run()` and the run method would run all the methods like `handleBirthTime`. This way at least we don't have to update translate every time and then we can expand on this mechanism and make it even better, maybe. Also this way we are acknowledging that our business case requires us to do some sort of post conversion cleanup (I don't like cleanup hopefully we can think of a better word) and we can split this functionality out to be its own thing so everything is more cohesive.

Another note here is we may have a fork of the LinuxForHealth HL7 -> FHIR converter soon, and a lot of these solutions, like `handleBirthTime`, should be addressed there whenever possible and `BundleEnhancer` or its equivalent solution should be used as a last resort.

### Acceptance Criteria
- [ ] The code that "enhances" the bundle in moved out of the translate function
- [ ] The code is now extensible

bishoyayoub:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/10849?planning-poker) with Zenhub @arnejduranovic @jack-h-wang @JessicaWNava @JFU-NAVA-PBC @jimmyfagan @mkalish @thetaurean

bishoyayoub:
@arnejduranovic @avnieldravid Please revisit priority.

PR #11506, opened 2023-09-22
Cleaned up the bundle enhancing methods by moving them out into their…

JessicaWNava:
… own class

This PR moves the bundle enhancing methods into their own class so that they are separate and extensible. 

Test Steps:
1. Make sure the tests still run and pass. 

## Changes
- Class created and bundle enhancing methods moved. 

### Testing
- [x] Tested locally?
- [ ] Added tests?

## Linked Issues
- #10849


jack-h-wang:
Is this work ReportStream specific? If so, it probably shouldn't live in the section of code that will eventually become the separate FHIR-to-HL7v2 library.

jack-h-wang (approved):
Compiled and ran all tests (unit, integration, smoke) without issues.

Could think of other locations this code could live (perhaps with `FHIRBundleHelpers`?) but I'm OK with this being delegated to a later task.

prime-router/src/main/kotlin/fhirengine/translation/BundleEnhancer.kt
```diff
@@ -0,0 +1,74 @@
...
+
+import ca.uhn.hl7v2.model.Message
+import gov.cdc.prime.router.fhirengine.utils.HL7Reader
+import io.github.linuxforhealth.hl7.data.Hl7RelatedGeneralUtils
+import org.hl7.fhir.r4.model.Bundle
+import org.hl7.fhir.r4.model.DateTimeType
+import org.hl7.fhir.r4.model.Extension
+import org.hl7.fhir.r4.model.Patient
+
+class BundleEnhancer {
```
arnejduranovic, line 11: If you make this a non-static class and delegate the run task to the client, it could make this class more:

- testable: If you follow this approach, you could have unit tests for each method in here. But, the way it is now, in order to test these methods you have to test them via `HL7toFhirTranslator` integration tests.
- extensible/cohesive: we would delegate the run task to the client, that way different clients in the future can choose exactly what enhancements they run by easily chaining them together, instead of the enhancements being hardcoded in this class itself

Here is the patch that implements this suggestion (easier then explaining everything). Each enhancer returns `this` so you can chain the calls in the client

[OOP_suggestions.patch](https://github.com/CDCgov/prime-reportstream/files/12729019/OOP_suggestions.patch)

If this makes sense and you choose to do this, then let's also write unit tests for the enhancer methods.

jimmyfagan: I would recommend doing this differently, and not having a class for BundleEnhancer at all, instead using extension functions on Bundle. [This patch](https://github.com/CDCgov/prime-reportstream/files/12729752/Use_extension_functions_for_BundleEnhancer_kt.patch) demonstrates it.
It has the same benefits that Arnej highlights above, but instead of having this object `BundleEnhancer`, we can literally enhance `Bundle` to have these functions.

One slight issue with this approach is that we would have extension functions in both `BundleEnhancer.kt` and `FHIRBundleHelpers.kt`, with the later having a wrapping class. I would personally eliminate the class `FHIRBundleHelpers` and have all of the helping functions from both files in the same file without a class, but might be an overreach for this story, so I'd be okay holding off on that for now.

arnejduranovic: Jimmy's idea makes sense to me and sounds like a good way to better utilize Kotlin features for more maintainable code

arnejduranovic (changes requested):
Can we add unit tests for `enhanceBundleMetadata` and `handleBirthTime`, similar to how we have good unit tests for `addReceivers`

JessicaWNava:
> Can we add unit tests for `enhanceBundleMetadata` and `handleBirthTime`, similar to how we have good unit tests for `addReceivers`

Sure. There are tests for these already they just exist at a higher level in the HL7ToFHIRTranslatorTests

JessicaWNava:
> Can we add unit tests for `enhanceBundleMetadata` and `handleBirthTime`, similar to how we have good unit tests for `addReceivers`

Tests added, ready for your re review

prime-router/src/test/kotlin/fhirengine/engine/fhirRouterTests/FHIRBundleHelpersTests.kt
```diff
@@ -688,4 +693,136 @@ class FHIRBundleHelpersTests {
...
+
+        // create the hl7 reader
+        val hl7Reader = HL7Reader(actionLogger)
+        // get the hl7 from the blob store
+        val hl7Message = File("src/test/resources/fhirengine/engine/hl7_with_birth_time.hl7").readText()
+        val hl7messages = hl7Reader.getMessages(hl7Message)
+
+        bundle.enhanceBundleMetadata(hl7messages[0])
+
+        assertThat(bundle.timestamp.toString()).isEqualTo("Wed Feb 10 17:07:37 PST 2021")
```
jimmyfagan, line 779: We should avoid using toString for these kinds of comparisons; even if the dates are equivalent, it could result in failed tests depending on the timezone of the system it is run on. Either testing for equality to a second `Date`, or using `toGMTString` instead should work.

prime-router/src/test/kotlin/fhirengine/engine/fhirRouterTests/FHIRBundleHelpersTests.kt
```diff
@@ -688,4 +693,136 @@ class FHIRBundleHelpersTests {
...
+        val fhirBundle = File("src/test/resources/fhirengine/engine/fhir_without_patient.fhir").readText()
+        val messages = FhirTranscoder.getBundles(fhirBundle, actionLogger)
+        assertThat(messages).isNotEmpty()
+        val bundle = messages[0]
+        assertThat(bundle).isNotNull()
+
+        // create the hl7 reader
+        val hl7Reader = HL7Reader(actionLogger)
+        // get the hl7 from the blob store
+        val hl7Message = File("src/test/resources/fhirengine/engine/hl7_with_birth_time.hl7").readText()
```
arnejduranovic, line 774: Can we assert somehow this is indeed a 2.5.1?

prime-router/src/test/kotlin/fhirengine/engine/fhirRouterTests/FHIRBundleHelpersTests.kt
```diff
@@ -688,4 +693,136 @@ class FHIRBundleHelpersTests {
...
+        val fhirBundle = File("src/test/resources/fhirengine/engine/fhir_without_patient.fhir").readText()
+        val messages = FhirTranscoder.getBundles(fhirBundle, actionLogger)
+        assertThat(messages).isNotEmpty()
+        val bundle = messages[0]
+        assertThat(bundle).isNotNull()
+
+        // create the hl7 reader
+        val hl7Reader = HL7Reader(actionLogger)
+        // get the hl7 from the blob store
+        val hl7Message = File("src/test/resources/fhirengine/engine/hl7_2.7.hl7").readText()
```
arnejduranovic, line 797: Can we assert somehow this is indeed a 2.7?

prime-router/src/test/kotlin/fhirengine/engine/fhirRouterTests/FHIRBundleHelpersTests.kt
```diff
@@ -688,4 +693,136 @@ class FHIRBundleHelpersTests {
...
+
+        bundle.enhanceBundleMetadata(hl7messages[0])
+
+        assertThat(bundle.timestamp.toString()).isEqualTo("Wed Feb 10 17:07:37 PST 2021")
+        assertThat(bundle.identifier.value).isEqualTo("371785")
+        assertThat(bundle.identifier.system).isEqualTo("https://reportstream.cdc.gov/prime-router")
+    }
+
+    @Test
+    fun `Test enhance bundle metadata 2-6`() {
```
arnejduranovic, line 808: 2.6 is just a sample use case, the purpose of this test isn't to test 2.6 specifically, right? It's to test an "unsupported message ID", or whatever language you want to use, so I would rename this "Test enhance bundle unsupported message ID" or something.

arnejduranovic (changes requested):
Changes look good, I like the more focused unit tests quite a bit! I left a few tiny suggestions that are not blockers to merge, but I had issues running the tests. Two of them fail. It looks like the timestamp value is different from what you are expecting? When I run `Test enhance bundle metadata 2-5-1`, `bundle.timestamp.toString()` evals to `Wed Feb 10 17:07:37 CST 2021` but you are expecting `PST`. 

```
gov.cdc.prime.router.fhirengine.engine.fhirRouterTests.FHIRBundleHelpersTests Test enhance bundle metadata 2-5-1() FAILED

  org.opentest4j.AssertionFailedError: expected:<...Wed Feb 10 17:07:37 [P]ST 2021"> but was:<...Wed Feb 10 17:07:37 [C]ST 2021">
      at app//gov.cdc.prime.router.fhirengine.engine.fhirRouterTests.FHIRBundleHelpersTests.Test enhance bundle metadata 2-5-1(FHIRBundleHelpersTests.kt:779)

gov.cdc.prime.router.fhirengine.engine.fhirRouterTests.FHIRBundleHelpersTests Test enhance bundle metadata 2-7() FAILED

  org.opentest4j.AssertionFailedError: expected:<...Wed Feb 10 17:07:37 [P]ST 2021"> but was:<...Wed Feb 10 17:07:37 [C]ST 2021">
      at app//gov.cdc.prime.router.fhirengine.engine.fhirRouterTests.FHIRBundleHelpersTests.Test enhance bundle metadata 2-7(FHIRBundleHelpersTests.kt:802)
```

prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
```diff
@@ -27,317 +29,367 @@ import kotlin.reflect.jvm.isAccessible
...
-                    conditionFilter
-                )
-            }
+        patient.birthDateElement.addExtension(extension)
+    }
+}
 
-            if (passes) {
-                observationsToKeep.add(observation)
+class FHIRBundleHelpers {
```
jimmyfagan, line 357: Even these functions can be taken out of the `FHIRBundleHelpers` class, and that class and the companion object can be removed altogether.

jimmyfagan (commented):
The changes and the test coverage looks good! Testing with timezones is tough to get right, so had some comments about necessary changes to those, but overall, I think we're close to ready here.

prime-router/src/test/kotlin/fhirengine/engine/fhirRouterTests/FHIRBundleHelpersTests.kt
```diff
@@ -688,4 +693,136 @@ class FHIRBundleHelpersTests {
...
+        val actionLogger = ActionLogger()
+        val fhirBundle = File("src/test/resources/fhirengine/engine/fhir_without_birth_time.fhir").readText()
+        val messages = FhirTranscoder.getBundles(fhirBundle, actionLogger)
+        assertThat(messages).isNotEmpty()
+        val bundle = messages[0]
+        assertThat(bundle).isNotNull()
+
+        // create the hl7 reader
+        val hl7Reader = HL7Reader(actionLogger)
+        // get the hl7 from the blob store
```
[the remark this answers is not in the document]
jimmyfagan: Removed this and similar ones

prime-router/src/test/kotlin/fhirengine/engine/fhirRouterTests/FHIRBundleHelpersTests.kt
jimmyfagan: Yep, added.

prime-router/src/test/kotlin/fhirengine/engine/fhirRouterTests/FHIRBundleHelpersTests.kt
jimmyfagan: Yep, added.

prime-router/src/test/kotlin/fhirengine/engine/fhirRouterTests/FHIRBundleHelpersTests.kt
jimmyfagan: Renamed this test 👍

prime-router/src/test/resources/fhirengine/engine/hl7_2.7.hl7
```diff
@@ -0,0 +1,6 @@
+MSH|^~\&|CDC PRIME - Atlanta, Georgia (Dekalb)^2.16.840.1.114222.4.1.237821^ISO|Avante at Ormond Beach^10D0876999^CLIA|PRIME_DOH|Prime ReportStream|20210210170737||ORM^O01^ORM_O01|371785|P|2.7|||NE|NE|USA||||PHLabReportNoAck^ELR_Receiver^2.16.840.1.113883.9.11^ISO
```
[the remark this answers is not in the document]
jimmyfagan: Added a timezone

prime-router/src/test/resources/fhirengine/engine/hl7_with_birth_time.hl7
```diff
@@ -0,0 +1,6 @@
+MSH|^~\&|CDC PRIME - Atlanta, Georgia (Dekalb)^2.16.840.1.114222.4.1.237821^ISO|Avante at Ormond Beach^10D0876999^CLIA|PRIME_DOH|Prime ReportStream|20210210170737||ORM^O01^ORM_O01|371784|P|2.5.1|||NE|NE|USA||||PHLabReportNoAck^ELR_Receiver^2.16.840.1.113883.9.11^ISO
```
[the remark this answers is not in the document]
jimmyfagan: Added a timezone

prime-router/src/test/kotlin/fhirengine/engine/fhirRouterTests/FHIRBundleHelpersTests.kt
jimmyfagan: Changed this to compare to a `Date`

prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
jimmyfagan: For some reason when I make this change, I get out of memory exceptions in `test filterObservations`. I don't understand why but it consistently fails when I try to change this, and it consistently works when I back that change out. After spending most of my day trying to figure this out, I'm going to give up on it and just add a comment about why this util class exists instead of just having all top-level functions.

Issue #10484, opened 2023-07-18
Add support for BlobAccess to accept environment


mkalish:
### User Story
As a report stream developer, I would like to choose the environment I am accessing azure blob storage for

### Description/Use Case
To facilitate future work we need to add the ability for BlobAccess to access environments other than the current RS instance. Add a means to allow BlobAccess to target a configurable environment. 

### Dev Notes
The tech debt portion of this ticket was spun off to #10774 

This work could take several forms, such as:
BlobAccess should take `Environment` in its constructor, then be able to access blobs for that specific environment.
or
BlobAccess should take `Environment` as an optional parameter in function calls, then be able to access blobs for that specific environment.

### Acceptance Criteria
- [ ] BlobAccess is able to access blobs in other environments

bishoyayoub:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/10484?planning-poker) with Zenhub @arnejduranovic @jimmyfagan @JFU-NAVA-PBC @mkalish @luis-pabon-tf @jack-h-wang @JessicaWNava

jack-h-wang:
For discussion: BlobAccess is largely a collection of static methods, with `uploadReport` being the only exception. Does it make sense to instantiate an instance of BlobAccess per environment, or should environment be a parameter for each function, keeping BlobAccess itself agnostic?

arnejduranovic:
@jeremy-page is this blocked by CDC Security? @jack-h-wang

PR #11558, opened 2023-09-26
10484 BlobAccess can access alternate environments

jack-h-wang:
This PR refactors several public functions of `BlobAccess` to take `BlobContainerMetadata` as an optional parameter, in order to facilitate access to blob stores other than the default. `BlobContainerMetadata` is also made a public class and a builder function has been added so that it can be created when calling `BlobAccess`.

Currently the only environment that is accessed, other than the current environment the RS instance is running, are environments for receivers that use the blob store transport. This PR changes the way BlobAccess and the blob store transport interact but has no functional difference from current behavior. In future work it should be possible for `Environment` to return the information needed to access a particular environment's blobstore.

Test Steps:
1. Confirm via smoke test that blobstore transport functions correctly.

## Changes
- Recreated `BlobContainerMetadata` as a public data class and added builder functions.
- Refactored `BlobAccess` public functions to use `BlobContainerMetadata` to refer to the target blobstore.
- Refactored `BlobStoreTransport` to utilize `BlobContainerMetadata` when interfacing with `BlobAccess`.
- Moved storage of environment variable name of the default blobstore to `Environment`. For the time being, the variable name does not change between environments.
- Updated tests as necessary.

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- #10484

## To Be Done
- #11565


prime-router/src/main/kotlin/azure/BlobAccess.kt
```diff
@@ -55,6 +58,31 @@ class BlobAccess() : Logging {
         }
     }
 
+    /**
+     * Data structure for configuring reusable blob store container access.
+     */
+    data class BlobContainerMetadata(
```
jack-h-wang, line 64: This was formerly the class used to uniquely identify container clients for reusability, but the existing properties were exactly what we needed for the purpose of defining a blob store to access, so the class is now exposed to other methods in RS that may use BlobAccess. The build functions serve to provide a connection string resolver in cases where it needs to be resolved from environment variable.

prime-router/src/main/kotlin/azure/BlobAccess.kt
```diff
@@ -155,19 +180,23 @@ class BlobAccess() : Logging {
...
         }
 
         /**
          * Download the blob at the given [blobUrl] as a ByteArray
          */
-        fun downloadBlobAsByteArray(blobUrl: String, retries: Int = getBlobDownloadRetry()): ByteArray {
+        fun downloadBlobAsByteArray(
+            blobUrl: String,
+            blobConnInfo: BlobContainerMetadata = defaultBlobMetadata,
+            retries: Int = blobDownloadRetryCount
```
jack-h-wang, line 193: Specifying a function as the default value of `retries` caused some strange mocking behavior. As we don't expect the default number of retries to vary at runtime, I moved the retrieval of this value to private values in the companion object instead, so that a static value is available when this is not provided by the caller.

prime-router/src/main/kotlin/transport/BlobStoreTransport.kt
```diff
@@ -26,7 +26,7 @@ class BlobStoreTransport() : ITransport {
             val receiver = header.receiver ?: error("No receiver defined for report ${header.reportFile.reportId}")
             val bodyUrl = header.reportFile.bodyUrl ?: error("Report ${header.reportFile.reportId} has no blob to copy")
             context.logger.info("About to copy $bodyUrl to $envVar:$storageName")
-            var newUrl = BlobAccess.copyBlob(bodyUrl, envVar, storageName)
+            val newUrl = BlobAccess.copyBlob(bodyUrl, BlobAccess.BlobContainerMetadata.build(transportType))
```
jack-h-wang, line 29: This is currently the only example of alternate environment use - in this case, an environment external to RS when a receiver uses the blob store transport method.

thetaurean (approved):
Looks good to me 👍 all tests passed ✅

arnejduranovic (changes requested):
Changes look fine but the added test doesn't make sense to em

prime-router/src/test/kotlin/azure/BlobAccessTests.kt
```diff
@@ -327,10 +332,37 @@ class BlobAccessTests {
...
+        val testBlobMetadata = BlobAccess.BlobContainerMetadata.build(testContainer, testEnvVar)
+
+        every { BlobAccess.exists(any(), defaultBlobMetadata) } returns true
+        every { BlobAccess.exists(any(), testBlobMetadata) } returns false
+
+        assertThat(defaultBlobMetadata.connectionString).isEqualTo("defaultconnection")
+        assertThat(testBlobMetadata.connectionString).isEqualTo("testconnection")
+        val result = BlobAccess.exists(testUrl, defaultBlobMetadata)
+        val result2 = BlobAccess.exists(testUrl, testBlobMetadata)
+        assertThat(result).isTrue()
```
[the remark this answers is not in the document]
jack-h-wang: I think I wanted to test accessing a blob without seeming like a retread of existing tests, but upon re-reviewing the existing tests, I think they actually do already test the things I wanted to test. With that in mind, I've now limited this test's scope to just ensuring that the `BlobContainerMetadata.build` methods function as expected.

arnejduranovic (approved):
Looks good. Before merging, make sure you have a plan for how to **DEMONSTRATE** the following acceptance criteria at Demo Day:

`BlobAccess is able to access blobs in other environments`

Issue #10943, opened 2023-08-16
Clean up translation integration tests for UP


jimmyfagan:
### User Story
As a ReportStream developer, I want integration tests to clearly verify necessary behavior, and I don't want to maintain integration tests for abandoned pilots.

### Description/Use Case
Many integration tests in the UP have historically been sample data from Colorado. These tests are based on a pilot that never happened, so now that we have some other sample data and integration tests for the UP (from ELIMS and CA), the old Colorado cases don't serve much purpose, and have become difficult to maintain as we work through tech debt and improve the losslessness of the UP for new pilots. Additionally, while it may be useful to have tests that verify that things work as expected for a given use-case of the UP, it may also be useful to have integration tests that are more based on some feature of the data, rather than just a particular sender/receiver combination.

### Risks/Impacts/Considerations
We shouldn't remove so many integration tests as to lose a "safety net" of making sure we're handling a variety of cases appropriately. We couldn't possibly cover all the possibilities of segments and fields present/missing from a message, so some variety in tests helps our confidence that things are working.

### Dev Notes
Some CA tests (particularly FHIR->HL7) utilized the Colorado data. We should be sure to work with Victor or others in the loop with CA/Simple Report before removing any of those.

### Acceptance Criteria
- [ ] Translation tests for the UP in the integration tests each serve a clear purpose (either related to an in-progress or fully active sender/receiver use-case, or related to a particular feature of the data)
- [ ] We have coverage for a reasonable variety of messages.
- [ ] Non-standard mappings that only served to support outdated tests have been removed or adjusted (see [this comment](https://github.com/CDCgov/prime-reportstream/pull/11093#discussion_r1307582794) based on OBX.2)

bishoyayoub:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/10943?planning-poker) with Zenhub @arnejduranovic @brick-green @jack-h-wang @JessicaWNava @JFU-NAVA-PBC @jimmyfagan @luis-pabon-tf @thetaurean

bishoyayoub:
@arnejduranovic Can you help refine this ticket for grooming?

arnejduranovic:
Looking at this again, I think the AC is fine. It's also important to note that each sender/receiver translation schema will have tests run automatically once the following epic is implemented: https://github.com/CDCgov/prime-reportstream/issues/10481

[...]

PR #11566, opened 2023-09-27
Simplify OBX2 string mapping

thetaurean:
This PR cleans up the OBX2 mapping to remove a condition that was added to satisfy CO pilot.

Test Steps:
1. Run translation tests (`gradlew testIntegration`)

## Changes
- Removed dependency check on OBX5 within OBX2 mapping.

### Testing
- [x] Tested locally?
- [x] ~~Added~~ Modified tests

## Linked Issues
- #10943


jimmyfagan (changes requested):
The change looks good to me, however it breaks several integration tests. You'll need to determine whether to update those integration tests so they pass, or adjust the schemas/transforms to keep them working. My gut says we should update the expected output for the failing tests, the input for the failing tests probably came from the same sample messages as the CO ones that were removed, so the behavior here is likely not reflective of anything happening with SimpleReport.

thetaurean:
> The change looks good to me, however it breaks several integration tests. You'll need to determine whether to update those integration tests so they pass, or adjust the schemas/transforms to keep them working. My gut says we should update the expected output for the failing tests, the input for the failing tests probably came from the same sample messages as the CO ones that were removed, so the behavior here is likely not reflective of anything happening with SimpleReport.

That's exactly the green light I was looking for 😀 thanks for the sanity check.
I'll get those tests fixed up and re-request.

thetaurean:
Tests fixed!

arnejduranovic (approved):

prime-router/metadata/hl7_mapping/common/observation-result.yml
```diff
@@ -4,7 +4,7 @@ elements:
     hl7Spec: [ '%{hl7OBXField}-1' ]
 
   - name: result-value-type-st
-    condition: '(%resource.value.exists() and %resource.value.is(string)) or (%resource.dataAbsentReason.coding.where(extension(%`rsext-code-index-name`).value = "identifier").exists() and %resource.dataAbsentReason.coding.where(extension(%`rsext-code-index-name`).value = "identifier").code = "unknown")'
+    condition: '%resource.value is String'
```
jimmyfagan, line 7: After merging, due to the HAPI FHIR updates [here](https://github.com/CDCgov/prime-reportstream/pull/11406), this condition is no longer valid and smoke tests and integration tests are now failing. Kalish posted some tips [here](https://nava.slack.com/archives/C01HFHPNJEL/p1695841059693489) (not sure if the link is Nava-specific or if it'll work for you), but you should be able to just keep what's currently in `master` with the last part removed like before.

```suggestion
    condition: '%resource.value.exists() and %resource.value.is(string)'
```

thetaurean: Thankyou for the suggested fix!

jimmyfagan (approved):
Latest commits including merges look good!

Issue #11543, opened 2023-09-26
Create the schemaResource constant


JessicaWNava:
### User Story
As a RS developer, I want access to the resource initially passed into the file. 

### Description/Use Case
In `FhirToHl7Converter`, we would set a constant called `schemaResource` to 
equal the resource being passed into the file. That way, there would automatically be access in each file to the schema 
resource. It could then be used like so:
```
# resource: <Path to organization>
- name: performing-organization-name-pracrole
  condition: '%{schemaResource} is PractitionerRole'
  resource: '%resource.performer.resolve().organization.resolve()'
  schema: ./datatype/xon-organization
  constants:
    hl7OrgField: '%{hl7OBXField}-23'
```

See prime-router/docs/design/design/schema-resource-design.md for the full proposal.

### Acceptance Criteria
- [x] The initial resource can be accessed through a constant automatically added in the code rather than being defined in a file.

bishoyayoub:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/11543?planning-poker) with Zenhub @arnejduranovic @brick-green @jack-h-wang @JessicaWNava @JFU-NAVA-PBC @jimmyfagan @thetaurean

PR #11582, opened 2023-09-28
11543: set %context to the resource available before getting the focus resource

mkalish:
This PR updates how the fhir path engine to set the context constant resource.

Test Steps:
1. Run the smoke tests and integration tests

## Changes
- Updates the FHIRPathUtils invocation of the the fhir path engine to set the context parameter to the %resource that was set before evaluating the element

### Testing
- [x] Tested locally?
- [x] Added tests?

### Process

## Linked Issues
- #11543


prime-router/metadata/hl7_mapping/OML_O21/base/service-request-order-observation.yml
```diff
@@ -5,6 +5,6 @@ elements:
     schema: ../../common/observation-request
 
   - name: order-observation-date-time-start
-    condition: 'Bundle.entry.resource.ofType(Specimen).collection.collected.exists()'
-    value: [ 'Bundle.entry.resource.ofType(Specimen).collection.collected' ]
+    condition: '%rootResource.entry.resource.ofType(Specimen).collection.collected.exists()'
```
mkalish, line 8: TLDR; This is a side effect of no longer passing in `Bundle` as the context.  `%rootResource` is another reserved constant that should be used instead.

[This](https://github.com/hapifhir/org.hl7.fhir.core/blob/64f034ebbff80037d0fa24ef0f96031b02956d0a/org.hl7.fhir.r4/src/main/java/org/hl7/fhir/r4/utils/FHIRPathEngine.java#L803) array passed into the execution now contains the context and not bundle and the downstream implication is that [this](https://github.com/hapifhir/org.hl7.fhir.core/blob/64f034ebbff80037d0fa24ef0f96031b02956d0a/org.hl7.fhir.r4/src/main/java/org/hl7/fhir/r4/utils/FHIRPathEngine.java#L3084) while loop no longer finds `Bundle`

prime-router/metadata/hl7_mapping/common/observation-result.yml
```diff
@@ -212,7 +212,7 @@ elements:
     resourceIndex: contactIndex
 
   - name: observation-note
-    condition: '%observation.note.exists()'
+    condition: '%context.note.exists()'
```
mkalish, line 215: Changed this one spot so that there was a code example that could be referenced.

prime-router/docs/design/design/context-resource-design.md
```diff
@@ -38,27 +38,27 @@ We want to be able to access both the initial observation (`'%diagnostic.result[
 the resource that is the result of this condition 
 `%resource.result.resolve() | %service.supportingInfo.resolve().where(specimen.exists().not())`. 
 However, to do this, we are needing to set the observation constant to the value of the 
-schema resource. This is a problematic workaround because this piece of code isn't actually working how you might 
+contezt resource. This is a problematic workaround because this piece of code isn't actually working how you might
```
jimmyfagan, line 41:
```suggestion
context resource. This is a problematic workaround because this piece of code isn't actually working how you might 
```

prime-router/docs/getting-started/standard-operating-procedures/changing-transforms.md
```diff
@@ -93,6 +93,7 @@ Each element contains the following properties, listed in order of execution:
   constants defined at their schema level and will overwrite any that have the same name.
   - A reserved constant, `%resource`, is automatically provided, which maps to the focus resource
     (generally, the `resource` property).
+  - A reserved constant `%context` is provided and maps to the previous `%resource` value
```
jimmyfagan, line 96: I could see this being a little unclear, maybe something like:
```suggestion
  - A reserved constant `%context` is provided and maps to the previous `%resource` value when a new `resource` value is also being set in the same element.
```
Depending on whether `context` can be used anytime, or only when `resource` is set, that suggestion may still be misleading, haven't gotten far enough into the PR to know one way or the other.

prime-router/src/main/kotlin/fhirengine/translation/hl7/FhirToHl7Converter.kt
```diff
@@ -151,7 +151,7 @@ class FhirToHl7Converter(
         focusResources.forEachIndexed { index, singleFocusResource ->
             // The element context must now get the focus resource
             elementContext.focusResource = singleFocusResource
-            if (canEvaluate(element, bundle, singleFocusResource, elementContext)) {
+            if (canEvaluate(element, bundle, singleFocusResource, focusResource, elementContext)) {
```
jimmyfagan, line 154: What do you think of renaming `singleFocusResource` (and maybe `focusResources` too) for clearer distinction between that and `focusResource`? Maybe `elementResource`/`elementResources`?

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirPathUtils.kt
```diff
@@ -94,14 +94,15 @@ object FhirPathUtils : Logging {
     fun evaluateCondition(
         appContext: CustomContext?,
         focusResource: Base,
+        contextResource: Base,
```
jimmyfagan, line 97: Do you think it would make sense to make `contextResource` an optional parameter that defaults to `bundle`? Could even consider doing the same for `focusResource` and doing similar things for the params in `canEvaluate`. It seems like we have implicit defaults for these parameters, it's just being handled by the callers.

prime-router/metadata/hl7_mapping/OML_O21/base/service-request-order-observation.yml
jimmyfagan: A couple comments/questions on this:

1. We should also document `rootResource` as a reserved/special constant.
2. How do these work? I don't see any special handling for them on our end, so I guess they were already doing this in the library? If you could point me in the right direction for documentation or code that explains it, I'd appreciate it.
3. Should we add anything to prevent folks from explicitly assigning values to the reserved constants? Is that already handled by the library as well? Not sure what would take precedence in those cases.

jimmyfagan (commented):
I'm impressed you found such an efficient way of doing this! Left a handful of comments and questions, mostly around clarity/readability type things, nothing too critical though. Curious what you think, so let me know when this is ready for another look.

prime-router/docs/getting-started/standard-operating-procedures/changing-transforms.md
mkalish: Context is actually always set when evaluating a FHIR path.  When starting at the root, focusResource, rootResource and context will all be the same value

prime-router/metadata/hl7_mapping/OML_O21/base/service-request-order-observation.yml
mkalish: - Yep, I'll add another note on it (actually I'm going to break out a new section and enumerate all of them)
- The code in question is [here](https://github.com/hapifhir/org.hl7.fhir.core/blob/64f034ebbff80037d0fa24ef0f96031b02956d0a/org.hl7.fhir.r4/src/main/java/org/hl7/fhir/r4/utils/FHIRPathEngine.java#L1711)
- The way the code is written is that setting `%context` (or any other reserved constant) would fail silently as the the hosts constant resolver is invoked last.  It's a good idea to update our constant adding code to throw an exception if someone tried to do it

prime-router/src/main/kotlin/fhirengine/translation/hl7/FhirToHl7Converter.kt
mkalish: Yep, can do

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirPathUtils.kt
mkalish: I'm hesitant to add default parameters here because I do think callers should consider what value is appropriate for the three values.

prime-router/docs/getting-started/standard-operating-procedures/changing-transforms.md
jimmyfagan: Ah gotcha, "previous `%resource`" is a little confusing especially for cases when a new `resource` isn't being set, but I'm not sure how else to put it, so I think we're okay here.

prime-router/metadata/hl7_mapping/OML_O21/base/service-request-order-observation.yml
jimmyfagan: Sounds good, that third item can be a separate story if it wouldn't make sense to add here, sounds like we're no worse off than we were before. The rest of that makes sense, thanks!

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirPathUtils.kt
jimmyfagan: Ah yea that's fair. Readability-wise on the caller end, I'd prefer to use named parameters to make it clear which is which, especially in cases where we're using the same value for multiple params, but I acknowledge that might just be a personal preference, especially when IDEs help. So I think we're good either way.

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirPathUtils.kt
```diff
@@ -94,14 +94,15 @@ object FhirPathUtils : Logging {
     fun evaluateCondition(
```
jimmyfagan, line 94: Function comment probably could use an update to help define the differences between the 3 similar params.

prime-router/src/main/kotlin/fhirengine/translation/hl7/ConfigSchemaProcessor.kt
```diff
@@ -109,11 +109,12 @@ abstract class ConfigSchemaProcessor : Logging {
         element: ConfigSchemaElement,
```
jimmyfagan, line 109: We probably ought to at least mention `contextResource` in the function comment, though there's not much detail anyway, so this is more of a nit.

prime-router/metadata/hl7_mapping/OML_O21/base/service-request-order-observation.yml
mkalish: It wasn't too much work so I threw it in as part of this work

prime-router/docs/design/design/transformations.md
```diff
@@ -131,6 +131,39 @@ constants:
...
+    condition: '%context.note.exists()'
+    resource: '%resource.note.text.split(''\n'')'
+    schema: note
+    resourceIndex: noteIndex
+    constants:
+      hl7NotePath: '%{hl7ObservationPath}'
+      noteDetails: '%observation.note'
+  ```
+  where `%resource` will be the list of strings from splitting the note text and `%context` will be the observation
+- `%rootResource` will always map to the FHIR bundle that is being converted
```
jimmyfagan, line 165: This is so good 👏

jimmyfagan (approved):
Looks great, thanks for the quick turnaround!

prime-router/docs/design/design/transformations.md
```diff
@@ -131,6 +131,39 @@ constants:
 
 The constants can be used in any FHIR expression belonging to any element in the schema or any child schema. When a constant appears in a special string enclosed in %\`\`, the code will look for strings defined as constants and replace them with their values. Example: %\`rsext-software-vendor-org\`
 
+There are a set of reserved names used in the underlying HAPI FHIR library.  Attempting to add them in a schema will result
```
arnejduranovic, line 134: Amazing doc addition!

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/ConstantResolver.kt
```diff
@@ -31,13 +32,25 @@ data class CustomContext(
     val translationFunctions: TranslationFunctions? = Hl7TranslationFunctions()
 ) {
     companion object {
+
+        private val reservedConstantNames =
+            listOf("loinc", "ucum", "resource", "rootResource", "context", "us-zip", "`vs-", "`cs-", "`ext")
+
         /**
          * Add [constants] to a context.
          * @return a new context with the [constants] added or the existing context of no new constants are specified
```
arnejduranovic, line 41:
```suggestion
         * @return a new context with the [constants] added or the existing context if no new constants are specified
```

arnejduranovic (approved):

Issue #11113, opened 2023-08-28
Complete MSH Mapping Inventory


arnejduranovic:
### User Story
As ReportStream, I must implement the [HL7v2 -> FHIR Mapping inventory](https://docs.google.com/spreadsheets/d/1PaFYPSSq4oplTvw_4OgOn6h2Bs_CMvCAU9CqC4tPBgk/edit#gid=0) for the MSH segment in order to support onboarding future Senders to the Universal Pipeline.

**Note:** If a mapping does not exist in the inventory, only fields specified by NIST 2.5.1 should be considered to help limit the scope to a manageable level

### Description/Use Case
ORU-R01 calls for the following mappings:

- [ ] MSH[MessageHeader]
- [ ] MSH[Bundle]

#### Test Plan
Test Plan: https://github.com/CDCgov/prime-reportstream/blob/master/prime-router/docs/design/proposals/mapping-inventory-test-plan.md
PR That implements test plan, use as example: https://github.com/CDCgov/prime-reportstream/pull/11147

### Acceptance Criteria
- [ ] Segment mappings updated to match [HL7v2 -> FHIR Mapping inventory ORU_R01](https://docs.google.com/spreadsheets/d/1gHK6_PFyr7PXns7wLDs0LSLsbjm0x-4bWUu3crXMKMI/edit#gid=0)
- [ ] Tests created and passing, per the linked test plan

bishoyayoub:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/11113?planning-poker) with Zenhub @arnejduranovic @brick-green @jack-h-wang @JessicaWNava @JFU-NAVA-PBC @jimmyfagan @luis-pabon-tf @thetaurean @mkalish

Andrey-Glazkv:
Might be more than 5 points - needs to be refined

Andrey-Glazkv:
Move to Backlog for refinement and planning

arnejduranovic:
We are moving this from in progress to backlog because as we were working on this ticket we found that it is not worth completing this ticket before we complete epic #11655

PR #11647, opened 2023-10-05
11113: add HD->Organization->HD mappings

mkalish:
This PR completes the HL7->FHIR->HL7 for HD/Location, HD/Organization, XON/Organization

Test Steps:
1. Review the FHIR inventory
  - https://docs.google.com/spreadsheets/d/1-cdMWepRbnqJOVVZ-y70Qv9Ub7evClmbX9gYO0kHmpo/edit#gid=0
  - https://docs.google.com/spreadsheets/d/1thMC8Et_hA-VC-DyerLrpNDcNdzPOIql9Z7SP7u9h_w/edit#gid=0
  - https://docs.google.com/spreadsheets/d/126xjG_YCcKYzO7f10dtlBOBUMNZZXkteoE2zNu-_J40/edit#gid=0
3. Review the test files and make sure all the conditions are covered

## Changes
- Completes the HD -> Organization -> HD mapping
- Completes the HD -> Location -> HD mapping
- Completes the XON -> Organization -> XON mapping

### Testing
- [x] Tested locally?
- [x] Added tests?

### Process

## Linked Issues
- #11113

## To Be Done
- #11113 (This PR only addresses some of the datatype mappings used by MSH. We will revisit the MSH story once all DataType mappings have been completed. See: https://github.com/CDCgov/prime-reportstream/issues/11655)


prime-router/metadata/fhir_mapping/hl7/datatype/HDToExtensionAssigningAuthority.yml
mkalish, line 1: This is mostly the same as ExtensionHierarchiDesinator, but I opted to make a new one to avoid having to update everywhere it's used

prime-router/metadata/fhir_mapping/hl7/datatype/HDToExtensionAssigningAuthority.yml
```diff
@@ -0,0 +1,57 @@
...
+  expressionsMap:
+    url:
+      type: SYSTEM_URL
+      value: universal-id-type
+    valueCode:
+      type: STRING
+      valueOf: HD.3
+      expressionType: HL7Spec
+
+extension_4-universalIdUnknownType:
```
mkalish, line 44: The mapping says to not include HD.2 if it is not ISO or UUID, so I capture in this extension so it can get mapped back to HL7

prime-router/metadata/fhir_mapping/hl7/datatype/HDToLocation.yml
```diff
@@ -0,0 +1,59 @@
...
+      type: SYSTEM_URL
+      value: unknown-universal-id-type
+    valueString:
+      expressionType: HL7Spec
+      type: STRING
+      valueOf: HD.2
+  vars:
+    universalId: STRING, HD.3
+
+extension_2-universalIdType:
```
mkalish, line 42: The inventory does not specify capturing this at all, so extension added.

prime-router/metadata/fhir_mapping/hl7/datatype/Identifier_var2.yml
mkalish, line 1: Would love any naming suggestions here.  Essentially the problem is the linux for health library does not work if the schemas are used like identifier->location-identifier, so I created another Identifier that does not include location

prime-router/metadata/fhir_mapping/hl7/secondary/Source.yml
```diff
@@ -136,6 +136,7 @@ extension_3:
     urlValue: software-install-date
 
 extension_4:
+  condition: $sft1 NOT_NULL
```
mkalish, line 139: Without this condition we were frequently adding an SFT segment even when it did not exist in the original HL7

prime-router/metadata/hl7_mapping/common/message-header.yml
```diff
@@ -111,8 +100,8 @@ elements:
     hl7Spec: [ MSH-11-2 ]
 
   - name: version-id
-    value: ['"2.5.1"']
-    hl7Spec: [MSH-12]
+    value: [ '"2.5.1"' ]
```
mkalish, line 103: Formatting

prime-router/metadata/hl7_mapping/common/datatype/hd-location.yml
```diff
@@ -5,9 +5,9 @@ elements:
     hl7Spec: [ '%{hl7HDField}-1' ]
 
   - name: location-universal-id
-    value: [ '%resource.identifier.value.getId()' ]
+    value: [ '%resource.identifier.value', '%resource.extension(%`rsext-unknown-universal-id-type`).value' ]
     hl7Spec: [ '%{hl7HDField}-2' ]
 
   - name: location-universal-id-type
-    value: [ '%resource.identifier.value.getIdType()' ]
```
mkalish, line 12: This is incorrect

prime-router/metadata/fhir_mapping/hl7/datatype/HDToOrganization.yml
```diff
@@ -0,0 +1,38 @@
...
+  vars:
+    code: HD.3
+    valueIn: HD.2
+    hd3: String, HD.3
+  constants:
+    systemId: "urn:ietf:rfc:3986"
+    system: "http://terminology.hl7.org/CodeSystem/v2-0301"
+
+identifier_3:
+  condition: $hd3 NOT_EQUALS ISO && $hd3 NOT_EQUALS UUID
```
mkalish, line 29: Test case to see how missing HD.3 is handled

prime-router/metadata/fhir_mapping/hl7/datatype/XONToOrganization.yml
```diff
@@ -0,0 +1,80 @@
...
+  generateList: true
+  valueOf: extension/Extension
+  expressionType: resource
+  vars:
+    value: String, XON.9
+  constants:
+    KEY_NAME_SUFFIX: String
+    urlValue: name-representation-code
+
+extension_2-preferXon10:
```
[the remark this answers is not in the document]
jimmyfagan: The 3 cases to consider are:

1. XON.3 is set and XON.10 is empty
2. XON.10 is set and XON.3 is empty
3. XON.3 and XON.10 are set

This seems to work for case 1 and 3, but not case 2. In case 2 we end up putting XON.10 into `identifier.value`, and then in the output we populate XON.3 from that and leave XON.10 blank. It's tough to fix case 2 without breaking case 1 though... We might need to have extensions for both (I would recommend using `organization-id-number` and `organization-organization-identifier` for the extension names), and still using the logic from the mapping to populate `identifier.value` in addition to just always populating the extensions if the respective field is there. Then on the FHIR->HL7 mapping, we can populate XON.3/10 from their respective extensions, with a fallback to populate just XON.10 from `identifier.value` if the extensions aren't there.

prime-router/metadata/hl7_mapping/common/datatype/organization-xon.yml
```diff
@@ -0,0 +1,56 @@
...
+    constants:
+      cweFieldPath: '%{hl7XONField}-2'
+    schema: ce-coded-with-exceptions
+
+  - name: id-number
+    condition: '%resource.extension(%`rsext-organization-prefer-xon-10`).exists().not()'
+    value: [ '%resource.identifier[0].value' ]
+    hl7Spec: [ '%{hl7XONField}-3' ]
+
+  - name: id-number-xon-10-valuted
```
jimmyfagan, line 18:
```suggestion
  - name: id-number-xon-10-valued
```

prime-router/src/testIntegration/resources/datatests/mappinginventory/xon/xon-to-organization-xon10-populated.hl7
```diff
@@ -0,0 +1 @@
+MSH|^~\&#|||||20230501102531-0400||ORU^R01^ORU_R01|3003786103_4988249_33033||2.5.1|||||USA|UNICODE UTF-8||||Org^A^123^Check Digit^C1^Assigning Authority&2.1.4.1&ISO^MD^Hospital A&2.16.840.1.113883.9.11&ISO^^Hospital B
```
jimmyfagan, line 1: This sure is bare bones, which is good, but when I try to use the CLI to test the conversions outside of test cases, I get all kinds of validation issues. In various iterations of trying to run this MSH through the CLI I've gotten:
```
19:10:36.012 [main][1] WARN  gov.cdc.prime.router.fhirengine.translation.hl7.FhirToHl7Converter - Could not set HL7 value for spec MSH-22-3 for element id-number
ca.uhn.hl7v2.model.DataTypeException: ca.uhn.hl7v2.validation.ValidationException: Validation failed: Primitive value '123' requires to be empty because it is withdrawn from the current HL7 version and should not be used
```
and
```
19:36:02.084 [main][1] WARN  gov.cdc.prime.router.fhirengine.translation.hl7.FhirToHl7Converter - Could not set HL7 value for spec MSH-22-4 for element identifier-check-digit
ca.uhn.hl7v2.model.DataTypeException: ca.uhn.hl7v2.validation.ValidationException: Validation failed: Primitive value 'Check Digit' requires to be empty or a number with optional decimal digits
```
It seems like the `FHIRToHl7Converter`is going to prevent us from populating XON.3. I haven't spent enough time looking into it to know if that's just a CLI thing or if the same validation code is used in the running application. But yea we should probably figure out what validation is used in the CLI vs the application vs the translation tests, and maybe try to get those synced up. Even if the mappings work in the tests it might be wasted effort if the application would never be able to utilize those mappings due to this kind of validation.

To recreate what I'm seeing, you can copy/paste a valid hl7 message into `junk/input.hl7`, swap out MSH with the contents of this file, tweak things as desired, and run the following:
```
./prime fhirdata --input-file junk/input.hl7 --output-file junk/output.fhir --output-format FHIR
./prime fhirdata --input-file junk/output.fhir -s metadata/hl7_mapping/ORU_R01/ORU_R01-base.yml --output-file junk/output.hl7 --output-format HL7
```

jimmyfagan (commented):
Wanted to get some comments and discussion going before going OOO tomorrow, didn't quite get through the whole PR, but I'm going to call it there for now. Overall what I've seen is good, there's just a lot to discuss and figure out especially if this is going to be a kind of template for how to handle future stories.

prime-router/metadata/fhir_mapping/hl7/datatype/XONToOrganization.yml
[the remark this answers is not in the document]
mkalish: Ah good call, yes I'll update and add a test case as well

prime-router/metadata/fhir_mapping/hl7/codesystem/ExtensionUrlMapping.yml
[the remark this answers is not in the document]
mkalish: My inclination was that one of the last tickets for this epic would go through and clean up this file without an output being documentation of RS extensions needed to convert FHIR to HL7 reliably.

prime-router/metadata/fhir_mapping/hl7/datatype/HDToExtensionAssigningAuthority.yml
```diff
@@ -0,0 +1,72 @@
...
+  generateList: true
+  expressionType: nested
+  expressionsMap:
+    url:
+      type: SYSTEM_URL
+      value: namespace-id
+    valueString:
+      type: STRING
+      valueOf: HD.1
+      expressionType: HL7Spec
```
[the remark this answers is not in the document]
mkalish: Don't have a strong preference, though I personally find it easier to understand with `extension/Extension`.  Logistically it looks like there are a lot fewer uses of it as well.

prime-router/metadata/fhir_mapping/hl7/resource/MessageHeader.yml
```diff
@@ -15,11 +15,12 @@ source:
...
 
+#map MSH.4
 sender:
-  valueOf: resource/MessageSender
+  valueOf: datatype/HDToOrganization
   expressionType: reference
-  vars:
-    country: MSH.17
+  generateList: true
+  specs: MSH.4
```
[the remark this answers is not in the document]
mkalish: This is a great call out.  I think the three options are 

- open up HDToOrganization and additional fields that can be set via vars
- create a dedicated MSH -> MessageHeaderSender.yml that would map both MSH.4 and MSH.17
- fork the library to support the ability to create extended schemas

I lean slightly towards option 1 since it would still generic and the new fields would be "opt-in"

prime-router/metadata/fhir_mapping/hl7/datatype/XONToOrganization.yml
```diff
@@ -0,0 +1,80 @@
...
+      value: organization-name-type
+    valueCodeableConcept:
+      valueOf: datatype/CodeableConcept
+      generateList: true
+      expressionType: resource
+      specs: XON.2
+
+identifier_1:
+  condition: $xon10 NULL
+  valueOf: datatype/Identifier_var
```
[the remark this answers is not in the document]
mkalish: Big picture I agree that we definitely need to clean up how we map `Identifier` across the board, but I don't think I've seen enough of the inventory to truly know that XON.3-10 map to an Identifier uniquely or there are other cases where a different HL7 datatype will map similarly.  I can put some more thought into it.

WRT deleting files.  I'm overall trying to avoid deleting anything for now and we can hopefully get consensus on the directory structure so we'll know where the "cleaned" schemas are.

prime-router/src/testIntegration/resources/datatests/mappinginventory/xon/xon-to-organization-xon10-populated.hl7
mkalish: Hmm... let me play around with it, but I didn't think those messages actually caused any issues with the output?  Though agree that having them get produced is messy.

mkalish: Ah so, I actually don't think that the the errors you're highlighting is what prevents the CLI from finishing correctly.
```
Exception in thread "main" gov.cdc.prime.router.fhirengine.translation.hl7.RequiredElementException: Required element patient-information conditional was false or value was empty.
        at gov.cdc.prime.router.fhirengine.translation.hl7.FhirToHl7Converter.processElement$prime_router(FhirToHl7Converter.kt:148)
        at gov.cdc.prime.router.fhirengine.translation.hl7.FhirToHl7Converter.processSchema(FhirToHl7Converter.kt:123)
        at gov.cdc.prime.router.fhirengine.translation.hl7.FhirToHl7Converter.processSchema$default(FhirToHl7Converter.kt:104)
        at gov.cdc.prime.router.fhirengine.translation.hl7.FhirToHl7Converter.convert(FhirToHl7Converter.kt:94)
        at gov.cdc.prime.router.cli.ProcessFhirCommands.convertToHl7(ProcessFhirCommands.kt:145)
        at gov.cdc.prime.router.cli.ProcessFhirCommands.run(ProcessFhirCommands.kt:99)
        at com.github.ajalt.clikt.parsers.Parser.parse(Parser.kt:198)
        at com.github.ajalt.clikt.parsers.Parser.parse(Parser.kt:211)
        at com.github.ajalt.clikt.parsers.Parser.parse(Parser.kt:18)
        at com.github.ajalt.clikt.core.CliktCommand.parse(CliktCommand.kt:400)
        at com.github.ajalt.clikt.core.CliktCommand.parse$default(CliktCommand.kt:397)
        at com.github.ajalt.clikt.core.CliktCommand.main(CliktCommand.kt:415)
        at com.github.ajalt.clikt.core.CliktCommand.main(CliktCommand.kt:440)
        at gov.cdc.prime.router.cli.MainKt.main(main.kt:297)
```

Is also the exception I see.

The test codes actually uses `ORU_R01-test.yml` rather than `ORU_R01-base.yml` and makes some required elements not required.

jimmyfagan: Ah that's a good call. To clarify, I was just putting this MSH onto another valid message to get around those required elements, but the difference between `ORU_R01-test` and `ORU_R01-base` explains a lot. Try swapping `251` with `27` in `ORU_R01-test` and rerunning the CLI conversions, and that should show what I'm talking about, and it seems like as a result of it, MSH.22.3 or MSH.22.4 get set. In the case of MSH.22.4, I think we can resolve it by using a valid input for that field, but for MSH.22.3, it doesn't seem like v2.7 will ever let us set that.

prime-router/metadata/fhir_mapping/hl7/datatype/HDToExtensionAssigningAuthority.yml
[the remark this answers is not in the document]
jimmyfagan: I'd vote for `extension/Extension`, once you're familiar with how it works I agree that it's easier to understand and it saves a few lines per element. Also don't have a strong preference though.

prime-router/metadata/fhir_mapping/hl7/resource/MessageHeader.yml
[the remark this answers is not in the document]
jimmyfagan: I agree with option 1. I have a slight concern that some resources will end up with tons of these additional var fields that only apply to specific contexts, and that it won't look that different than what we have today, but maybe that concern is unwarranted. Also even if that was the case, having one template with a bunch of one-off vars is probably still preferable to having several one-off templates that are all largely similar as would be the case with option 2. Option 3 would be nice, but I doubt it'd be easy and it's probably not super justifiable to put that effort in.

prime-router/metadata/fhir_mapping/hl7/datatype/HDToExtensionAssigningAuthority.yml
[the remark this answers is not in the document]
mkalish: Hmmm... so I'm not actually sure we would always be able to use it though?  Do you know how it would work if the extension had a `valueCodeableConcept` where we need to reference a datatype?

prime-router/src/testIntegration/resources/datatests/mappinginventory/xon/xon-to-organization-xon10-populated.hl7
mkalish: Confirmed... just need to redo how we initially create the HL7 when converting from FHIR and these exceptions go away.

prime-router/metadata/fhir_mapping/hl7/datatype/HDToExtensionAssigningAuthority.yml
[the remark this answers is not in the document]
jimmyfagan: I don't know how/if we could use it in those cases, it seems to only be set up to work for primitive-valued extensions. Maybe that swings us back to not using `extension/Extension` to avoid having two different ways of doing things?

mkalish:
Summarizing notes from a slack huddle with @jimmyfagan:

- We will deprecate use of `extension/Extension` when creating extensions as it does not handle all uses cases
- I'll attempt to inline the XON Identifier mapping to remove the need for a separate file
- For XON.3 and XON.10, we'll create extensions that will store both values and use those exclusively when mapping back to HL7.  A FHIR sender would either need to implement them as well or we would need to write a FHIR transform to add them for a sender
- I'll update the testing code to test: HL7-> FHIR, FHIR->HL7 and HL7->FHIR->HL7 to eliminate any holes
- I'll disable the validation when converting FHIR-> HL7
- I'll add additional opt in vars to HDToOrganization
- We'll handle deletions of schemas that get orphaned at the end of the epic to avoid unintentionally breaking something

prime-router/metadata/fhir_mapping/hl7/resource/MessageHeader.yml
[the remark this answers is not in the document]
mkalish: Added this to HDToOrganization.yml

prime-router/metadata/hl7_mapping/common/datatype/extensionAssigningAuthority-hd.yml
```diff
@@ -0,0 +1,19 @@
+elements:
+
+  - name: assigning-authority-namespace-id
+    value: [ '%resource.extension(%`rsext-namespace-id`).value' ]
+    hl7Spec: [ '%{hl7HDField}-1' ]
```
brick-green, line 5: should there be a `condition` here?

```condition: '%resource.extension(%`rsext-namespace-id`).exists()'```

File: prime-router/metadata/hl7_mapping/common/datatype/organization-xon.yml [hunk 14]
```diff
@@ -0,0 +1,49 @@
...
+
+  - name: identifier-check-digit-scheme
+    value: [ '%resource.identifier.extension(%`rsext-check-digit-scheme`).value' ]
+    hl7Spec: [ '%{hl7XONField}-5' ]
+
+  - name: xon-assigning-authority
+    resource: '%resource.identifier.extension(%`rsext-assigning-authority`)'
+    constants:
+      hl7HDField: '%{hl7XONField}-6'
+    schema: extensionAssigningAuthority-hd
```
brick-green, line 29: Should there be a condition here to check that the new resource exists before including `extensionAssigningAuthority-hd`?

prime-router/metadata/hl7_mapping/common/datatype/extensionAssigningAuthority-hd.yml
mkalish: I think there's an open style question here, since the library won't set anything if the resource is missing

prime-router/src/testIntegration/kotlin/datatests/mappinginventory/hd/HDToLocationTests.kt
```diff
@@ -0,0 +1,80 @@
...
+        assert(
+            translateAndCompareFHIRToHL7(
+                "mappinginventory/hd/HD-to-Location-uuid.fhir",
+                "mappinginventory/hd/HD-to-Location-uuid.hl7",
+            ).passed
+        )
+    }
+
+    @Test
+    fun `test correctly handles and unknown universal id type`() {
```
brick-green, line 48:
```suggestion
    fun `test correctly handles unknown universal id type`() {
```

prime-router/metadata/hl7_mapping/common/datatype/extensionAssigningAuthority-hd.yml
brick-green: okay, I thought at first that since there are conditions for HD-2 there should be conditions on all. But I see now that HD-2 has conditions to make the two possibilities mutually exclusive

prime-router/metadata/fhir_mapping/hl7/datatype/HDToExtensionAssigningAuthority.yml
jimmyfagan: It'd be nice if we could combine this with extension_5 below, but I know the boolean logic handling in this library can be finnicky. If we can't combine them, I'd at least recommend renaming the extensions to emphasize the distinction, maybe `universalIdUnknownType`/`universalIdNullType`.

prime-router/metadata/fhir_mapping/hl7/datatype/XONToOrganization.yml
```diff
@@ -0,0 +1,124 @@
...
+    url:
+      type: SYSTEM_URL
+      value: organization-name-type
+    valueCodeableConcept:
+      valueOf: datatype/CodeableConcept
+      generateList: true
+      expressionType: resource
+      specs: XON.2
+
+identifier_1:
```
jimmyfagan, line 26: This is great 👏 One nit: the order of sub-expressions seems a little random, could we get those in order by HL7 field index to make this a little easier to take inventory of?

jimmyfagan (commented):
Another handful of comments and questions, but I love where we're headed here, great stuff! If you're eager to get this merged, I can do another review or get on a huddle after 4pm ET.

prime-router/metadata/fhir_mapping/hl7/datatype/HDToOrganization.yml
```diff
@@ -55,4 +55,10 @@ identifier_4:
     hd3: String, HD.3
   constants:
     system: "http://terminology.hl7.org/CodeSystem/v2-0301"
-    isUniversalId: "true"
\ No newline at end of file
+    isUniversalId: "true"
+
+address:
+  condition: $country NOT_NULL
```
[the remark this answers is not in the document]
mkalish: Yep, thats a good idea

prime-router/src/testIntegration/kotlin/datatests/mappinginventory/TranslateAndCompareFile.kt
```diff
@@ -24,4 +24,67 @@ fun translateAndCompareHL7ToFHIR(inputFile: String, expectedOutputFile: String):
     return TranslationTests().FileConversionTest(
         testConfig
     ).runTest()
+}
+
+fun verifyHL7ToFHIRToHL7Mapping(testFileName: String): CompareData.Result {
```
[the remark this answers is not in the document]
mkalish: Yea, I'm going to defer that for now to avoid more scope creep here.

prime-router/metadata/fhir_mapping/hl7/datatype/HDToOrganization.yml
```diff
@@ -0,0 +1,64 @@
+resourceType: Organization
+
+id:
+  type: STRING
+  valueOf: "GeneralUtils.generateResourceId()"
+  expressionType: JEXL
+
+identifier_1:
```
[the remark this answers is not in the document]
mkalish: Yea I'll go ahead and inline all the identifiers here

prime-router/metadata/fhir_mapping/hl7/datatype/XONToOrganization.yml
mkalish: Will update

prime-router/metadata/hl7_mapping/common/datatype/extensionAssigningAuthority-hd.yml
```diff
@@ -0,0 +1,19 @@
...
+
+  - name: assigning-authority-universal-id-known
+    condition: '%resource.extension(%`rsext-universal-id`).exists()'
+    value: [ '%resource.extension(%`rsext-universal-id`).value' ]
+    hl7Spec: [ '%{hl7HDField}-2' ]
+
+  - name: assigning-authority-universal-id-unknown
+    condition: '%resource.extension(%`rsext-universal-id`).exists().not()'
+    value: [ '%resource.extension(%`rsext-unknown-universal-id-type`).value' ]
+    hl7Spec: [ '%{hl7HDField}-2' ]
```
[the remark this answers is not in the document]
mkalish: I think that's a safe change, I'll try that out

prime-router/metadata/hl7_mapping/common/datatype/hd-location.yml
```diff
@@ -5,9 +5,9 @@ elements:
     hl7Spec: [ '%{hl7HDField}-1' ]
 
   - name: location-universal-id
-    value: [ '%resource.identifier.value.getId()' ]
+    value: [ '%resource.identifier.value', '%resource.extension(%`rsext-unknown-universal-id-type`).value' ]
```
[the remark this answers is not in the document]
mkalish: Will update

prime-router/metadata/fhir_mapping/hl7/datatype/HDToLocation.yml
```diff
@@ -0,0 +1,72 @@
...
+  expressionType: HL7Spec
+
+identifier_1:
+  expressionType: nested
+  generateList: true
+  vars:
+    universalId: String, HD.3
+  expressionsMap:
+    value_1:
+      condition: $universalId EQUALS ISO || $universalId EQUALS UUID
```
jimmyfagan, line 20: Is there significance to having the condition here vs at the top level? Want to be sure we don't add an empty identifier if the condition is false.

jimmyfagan (approved):
Just had one last question, also still have a minor formatting suggestion in `XONToOrganizationTests.kt`, but not anything that should block approval. I'll go ahead an mark approved, but I can take another look on Monday if you make any new big changes.

prime-router/metadata/fhir_mapping/hl7/datatype/HDToLocation.yml
mkalish: Mostly just wanted to be consistent, though long term the value would be if there was another field that needed to be added to the identifier, that had nothing to to with the condition for `value`

prime-router/metadata/hl7_mapping/ORU_R01/ORU_R01-base.yml
```diff
@@ -13,7 +13,10 @@ elements:
     schema: ../common/message-header
 
   - name: software-segment
-    condition: 'Bundle.entry.resource.ofType(MessageHeader).exists()'
+    condition: >
+      Bundle.entry.resource.ofType(MessageHeader).exists() and 
+      Bundle.entry.resource.ofType(MessageHeader).source.exists() and
+      Bundle.entry.resource.ofType(MessageHeader).source.extension(%`rsext-software-vendor-org`).exists()
```
arnejduranovic, line 19: This isn't the only place this applies, but I'm curious: If the sender is sending us FHIR and they do not include this custom RS extension, we will not map their message properly and the ONLY way to tell this is to manually compare the FHIR to the HL7? That means onboarding of FHIR senders will be even more tedious than onboarding HL7v2 and probably cannot be done in an automated fashion (or with the help of self-service tools)?

arnejduranovic (approved):

prime-router/metadata/hl7_mapping/ORU_R01/ORU_R01-base.yml
mkalish: I think onboarding any sender/receiver pair where the input format != the output format will be a little tedious.  I do think for this case of FHIR -> HL7 the solution is to produce the list of extensions that must be populated in order to reliably translate a FHIR message to HL7.

I'm not sure there is a way around this and explicitly always create HL7 from FHIR

mkalish: Yep. going to revert this change and move it into the test schema.  This will be covered by the mapping MSH/SFT->MessageHeader->MSH/SFT and I'm not convinced that this is the only extension we need to look for when deciding whether to create an SFT segment

PR #11704, opened 2023-10-09
HL7 SPM Mapping for Outbound OML

tjohnson7021:
This PR  adds mapping to HL7 SPM fields for the outbound OML message type

Test Steps:

2. Check that the output HL7 message has the SPM field

## Changes
- created `specimen.yml` schema to be consumed by the `OML_O21.yml` base

### Testing
- [x] Tested locally

## Linked Issues
[Create Outbound OML SPM Segment](https://app.zenhub.com/workspaces/cdc-ti-633de70d939f6c00155f3722/issues/gh/cdcgov/trusted-intermediary/494)


prime-router/metadata/hl7_mapping/common/specimen.yml
```diff
@@ -0,0 +1,33 @@
...
+  - name: specimen-collection-date/time
+    condition: '%resource.collection.collected.exists() and %resource.collection.collected is dateTime'
+    value: [ '%resource.collection.collected' ]
+    hl7Spec: [ '%{hl7SegmentGroup}-17' ]
+
+  - name: specimen-collection-time
+    condition: '%resource.collection.collected.exists() and %resource.collection.collected is dateTime'
+    value: [ '%resource.collection.collected' ]
+    hl7Spec: [ '%{hl7SegmentGroup}-17-1' ]
+
```
brick-green, line 23: Why do both `specimen-collection-date/time` and `specimen-collection-time` exist with the same `condition` and `value`?

brick-green (changes requested):
double checked the FHIR against the mapping inventory and it looks good. So then tested that the expected HL7 comes out and it does. But I don't see the last two elements in HL7 mapping being tested

prime-router/metadata/hl7_mapping/common/specimen.yml
mkalish, line 1: Can the existing `specimen` schema get used?
https://github.com/CDCgov/prime-reportstream/blob/d3de61ce3d246e70a2ec9e76e0f9709c13b59c74/prime-router/metadata/hl7_mapping/ORU_R01/base/specimen.yml

jcrichlake: I don't think so as it has a different path specification in the HL7 definition.  Ours is for OML where is the path listed is for ORU

mkalish: Right, but I think that on can get moved to common and used here

prime-router/metadata/hl7_mapping/common/specimen.yml
jorg3lopez: I deleted one entry, good catch!

prime-router/metadata/hl7_mapping/common/specimen.yml
```diff
@@ -0,0 +1,33 @@
...
+  - name: specimen-observation-date-time-start
+    condition: '%resource.collection.collected.exists() and %resource.collection.collected is Period'
+    value: [ '%resource.collection.collected.start' ]
+    hl7Spec: [ '%{hl7SegmentGroup}-17-1' ]
+
+  # Alternative to getting specimen collection time from DiagnosticReport instead
+  - name: specimen-collection-time-diagnostic
+    condition: '%resource.collection.collected.exists().not()'
+    value: [ '%diagnostic.effective' ]
+    hl7Spec: [ '%{hl7SegmentGroup}-17-1' ]
```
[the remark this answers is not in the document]
jorg3lopez: After carefully looking if our FHIR bundle has a `DiagnosticReport` resource, we noticed that it is missing in the bundle. We opt to delete this entry as we don't have that resource to pull the information from.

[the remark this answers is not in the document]
jorg3lopez: We are working on providing a test for the `Period` type.

prime-router/metadata/hl7_mapping/common/specimen.yml
jorg3lopez: We can move the specimen.yml to the `common` directory. The only thing is that some paths are catered to the ORU or the ORM  message due to difference in resource for some fields. I don't think it would affect the outcome if the resource is not found. We will work on figuring out a way to reuse the specimen.yml.

prime-router/metadata/hl7_mapping/common/specimen.yml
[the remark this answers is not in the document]
jorg3lopez: Added test

brick-green (approved):
passes tests. looks good to me

prime-router/metadata/hl7_mapping/OML_O21/OML_O21-base.yml
```diff
@@ -66,3 +66,10 @@ elements:
       hl7OBXField: '%{hl7SegmentGroup}/OBX'
     resourceIndex: resultIndex
 
+  - name: order-specimen
+    resource: Bundle.entry.resource.ofType(Specimen)
+    condition: '%resource.count() = 1'
```
mkalish, line 71: Nit: this could just be `%resource.exists()`.  Is there a situation where the FHIR would contain more than one `Specimen` and you only want the first?

mkalish (approved):
Ran the integration tests locally and they all passed.

Just wanted to note that the mapping of specimen won't support a case where the output OML message should contain multiple `ORDER` groups.  I don't think that needs to be tackled with this PR though

prime-router/metadata/hl7_mapping/OML_O21/OML_O21-base.yml
jorg3lopez: We are only expecting one specimen resource.

prime-router/metadata/hl7_mapping/OML_O21/OML_O21-base.yml
```diff
@@ -66,3 +66,10 @@ elements:
...
+    required: true
+    schema: ../common/specimen
+    constants:
+      hl7SpecimenFieldPath: /ORDER/OBSERVATION_REQUEST/SPECIMEN/SPM
```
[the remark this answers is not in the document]
tjohnson7021: At this time we are only expecting to receive 1 `SPECIMEN` in a FHIR bundle.

prime-router/metadata/hl7_mapping/OML_O21/OML_O21-base.yml
tjohnson7021: At this time we are only expecting to receive 1 `SPECIMEN` in a FHIR bundle.

PR #11745, opened 2023-10-11
11650: update CompareFhirData to handle all entries

jalbinson:
This PR updates the CompareFhirData#compareBundle function to treat all FHIR entries as root nodes in the hierarchy. This bug was exposed when changing the expected output of some HL7->FHIR mappings in XTNTests. Anything that was changed in the `Patient` resource had no effect on the test results due to the fact that `Patient` was not referenced anywhere else in the FHIR bundle.

I had to update a mapping and some expected output to get the existing tests to pass. I also added some helpful messaging when tests fail to show what values were expected vs actual.

Test Steps:
1. ./testIntegration

## Changes
- Removed `entriesToCompare` list and now just use all entries as roots
- Updated `ContactPoint` FHIR mapping to so an HL7 value of "ORN" -> FHIR "temp"
- Added primitive message to show what the actual and expected values are during a failure
- Fixed another test

### Testing
- [x] Tested locally?
- [ ] Added tests?

## Linked Issues
- Fixes #issue


prime-router/metadata/fhir_mapping/hl7/datatype/ContactPoint.yml
```diff
@@ -80,6 +80,13 @@ use_4:
   vars:
     useCode: String, XTN.2
 
+use_5:
+  condition: $useCode EQUALS ORN
+  type: STRING
+  value: "temp"
+  vars:
+    useCode: String, XTN.2
+
```
brick-green, line 89: Where does this mapping come from? Do you know why we have a test for "temp"?

prime-router/src/main/kotlin/fhirengine/utils/CompareFhirData.kt
```diff
@@ -217,7 +214,16 @@ class CompareFhirData(
         }
 
         if (!primitiveResult.passed) {
-            val msg = "FAILED: Property $primitiveIdPath $primitiveTypePath did not match"
+            val msg = if (expectedPrimitive is PrimitiveType<*>) {
+                val actualPrimitivesValues = actualPrimitive
+                    .filterIsInstance<PrimitiveType<*>>()
+                    .joinToString { it.asStringValue() }
+                "FAILED: Property $primitiveIdPath $primitiveTypePath did not match. " +
+                    "expected=${expectedPrimitive.asStringValue()}, actuals=$actualPrimitivesValues"
```
brick-green, line 222: Really like this update here!

```suggestion
                    "expected=${expectedPrimitive.asStringValue()}, actual=$actualPrimitivesValues"
```

prime-router/metadata/fhir_mapping/hl7/datatype/ContactPoint.yml
jalbinson: I got the mapping from how the test was set up and the FHIR -> HL7 mapping found below.

https://github.com/CDCgov/prime-reportstream/blob/master/prime-router/metadata/hl7_mapping/common/datatype/xtn-telecom.yml#L14

jimmyfagan: The particular mapping for ORN was introduced in [this commit](https://github.com/CDCgov/prime-reportstream/commit/fa3db4c1e1025d583df06d317fb2037ab1e44dff#diff-7cf6b3e50bf178ab7427e314589d85b4c3747ea2e8231c654385fa5afa3af784), but it's not clear why. It's not spelled out in [this sheet](https://docs.google.com/spreadsheets/d/1EE3a5UIGXLi8wMDr49fWTEVx8QS2GFqYgitE3vLOQTQ/edit#gid=0), but `other` seems more appropriate. Maybe `temp` was meant to be a temporary value?

brick-green: Still don't understand why temp exists in the mapping in the first place. But that's outside scope here so makes sense add the element.

brick-green:
I took the HL7 and made the FHIR:
```
./gradlew primeCLI --args="fhirdata --output-format FHIR --input-file src/testIntegration/resources/datatests/HL7_to_FHIR/sample_orm_20230809-001.hl7 --output-file junk/sample_orm_20230809-001.fhir"
```
I'm seeing some differences when comparing the output from the above to the FHIR located at 
`prime-router/src/testIntegration/resources/datatests/HL7_to_FHIR/sample_orm_20230809-001.fhir`

for example. in the fhir in the repo I see a smaller identifier than the identifier in the fhir from command output. Is that on purpopse?

fhir in repo:
image


fhir from command:
image

jalbinson:
@brick-green I didn't write those tests so I couldn't tell you. I believe the code I wrote to compare all entries in the bundle just exposed the fact that they were not equal.

brick-green:
> @brick-green I didn't write those tests so I couldn't tell you. I believe the code I wrote to compare all entries in the bundle just exposed the fact that they were not equal.

Okay, so this PR is not meant to fix the comparison? That the other ticket you're working on?

jalbinson:
This PR was made for fixing the issue where some FHIR entries were not even being compared.

In a few of the files I fixed (which someone else seems to have fixed in the meantime) there were differences in the `Patient` entry but with it being referenced nowhere else in the file, it was never actually compared.

brick-green (approved):
looks good to me

arnejduranovic (approved):

PR #11783, opened 2023-10-13
Dependency grab bag + Fix test suite

etanb:
fixes #11746 , fixes #11731 , fixes #11686 , fixes #11519 , fixes #11385


frontend-react/package.json
```diff
@@ -71,7 +71,7 @@
         "cypress:production:open": "echo 'not implemented yet'",
         "test": "jest",
         "test:debug": "jest --watch --runInBand",
-        "test:ci": "jest --ci --runInBand --coverage --verbose --silent",
```
jpandersen87, line 74: Why would we want to silence the output from the github action logs?

jpandersen87 (changes requested):

frontend-react/package.json
etanb: ah yeah, whoops. that was a vestige of experimentation. will remove (also `--verbose` + `--silent` == comedy)

etanb: Oh! I removed the `--silent` for that reason! Also `--silent` + `--verbose` doesn't make any sense together.

jpandersen87: Woops, went dyslexic on reading the diff line here. ignore

jpandersen87 (approved):

frontend-react/src/components/FileHandlers/FileHandler.test.tsx
```diff
@@ -120,10 +120,24 @@ describe("FileHandler", () => {
             );
             expect(screen.getByText("Select data model")).toBeVisible();
         });
+
+        test.only("allows the user to upload and file and shows the success screen", async () => {
```
penny-lischer, line 124: Why do you need the test.only here and above? You want to skip the others?

etanb: Yes, the others are broken. Instead of commenting out, I wanted to use the jest methods, `jest.skip` and `jest.only`

penny-lischer (approved):
LGTM

PR #11820, opened 2023-10-16
Fixing the Build Frontend workflow to fail if any of the tests fail

### LINKED ISSUE #11815: Investigate why the Workflow is not failing when the tests are
[Here is an example](https://github.com/CDCgov/prime-reportstream/actions/runs/6511576289/job/17687553835?pr=11783). It looks like the FE workflow is running but failing tests are not making the action fail. Etan can help if you have questions.

dkrylovsb:
This PR fixes the Build Frontend workflow to fail if any of the tests fail.

Test Steps:
1. Run the workflow with one of more FE test broken and make sure it fails. 

## Changes
- Explicitly setting `shell` to `bash --noprofile --norc -eo pipefail` when `retry` action is being used. GHA does this automatically; however, based on my analysis of the [source code](https://github.com/nick-fields/retry/blob/master/src/index.ts), nick-fields/retry _drops_ the extra `bash` options, so the exit code of the entire flow is the exit code of the last command. 

### Testing
- [x] Tested locally?

## Linked Issues
- Fixes #11815

### DIFF
--- a/.github/actions/build-frontend/action.yml
+++ b/.github/actions/build-frontend/action.yml
@@ -38,12 +38,13 @@ runs:
         max_attempts: 2
         retry_wait_seconds: 30
         command: |
+          set -e pipefail
           cd frontend-react
           echo "" >> .env.production
           echo "VITE_APPINSIGHTS_KEY=${{ inputs.appinsights-prod-key }}" >> .env.production
           yarn test:ci
           yarn build:production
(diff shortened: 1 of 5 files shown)


snesm (approved):

snesm:
Maybe create an issue/ticket or PR against nick-fields/retry?

PR #11855, opened 2023-10-18
dependabot workaround for docker-compose

### LINKED ISSUE #11861: Dependabot does not work with docker-compose
## Problem statement

Dependabot does not currently work with docker-compose. Migrate to dockerfiles as a workaround

JosiahSiegel:
This PR includes a workaround for dependabot's inability to work with docker-compose files. Versioned images are now placed into dockerfiles and dependabot config is updated to include these new files, as well as group a few PR updates where appropriate.

## Changes
- Move versioned docker-compose images to dockerfiles
- optimize dependabot config

## Linked Issues
- Fixes #11861

### DIFF
--- a/.github/workflows/release_to_github.yml
+++ b/.github/workflows/release_to_github.yml
@@ -108,7 +108,7 @@ jobs:
           prerelease: ${{ needs.wf_vars.outputs.prerelease }}
       - name: Build changelog
         id: build_changelog
-        uses: mikepenz/release-changelog-builder-action@713885393e391b4ce984457280aba0d24fbf0c87
+        uses: mikepenz/release-changelog-builder-action@6fd5cc6eaf7567dbd0f9666061215bb476f012fc
         with:
           configuration: '.github/changelog_config.json'
           # Allow changelog tests in forks


.github/dependabot.yml
```diff
@@ -281,26 +263,36 @@ updates:
     directory: "/operations/app/terraform/vars/demo"
     schedule:
       interval: "weekly"
+    ignore:
```
snesm, line 266: I think we should keep these enabled since according to their [lifecycle policy](https://support.hashicorp.com/hc/en-us/articles/360021185113-Support-Period-and-End-of-Life-EOL-Policy), version [2.99 released](https://github.com/hashicorp/terraform-provider-azurerm/blob/main/CHANGELOG-v2.md) in March of 2022 and may go unsupported by April 2024.

.github/dependabot.yml
```diff
@@ -5,108 +5,95 @@
 
 version: 2
 updates:
-  # Maintain dependencies for GitHub Actions Workflows
+
+  # Docker
+  - package-ecosystem: "docker"
+    directory: "/"
```
snesm, line 11: I do not believe dependabot will [search subdirectories/subfolders](https://github.com/dependabot/dependabot-core/issues/1015).  Each location with docker files needs to be called out individually.

snesm (changes requested):

snesm (approved):

arnejduranovic (approved):

PR #11906, opened 2023-10-20
Oo/engagement/jgilmore/10151 code to condition tables

### LINKED ISSUE #9248: Create proposal for determining condition given test code or other message data
### User Story

As an engineer on ReportStream, I need to design how to map codes (LOINC, et.al.) to condition names so I can implement a solution where a STLT can give RS a list of conditions they want to receive without worrying about sending specific LOINC codes.

### Description/Use Case

We should design a solution that is:
1. maintainable - we can easily update conditions with new codes (LOINC and otherwise)
2. extensible - we should be able to support more than just LOINC codes and be able to easily expand to others, like LDT or PLA

TODO: Prior stories have populated this research document, it should be referenced in the creation of this design

Additionally, we should review Patricia's research doc and investigate each item listed there (for example, RELMA DB). We should aim to reuse as many existing technologies and solutions as possible.

We need to make sure we understand how our solution fits in with everything else going on at the CDC, like for example, SHIELD.

### Acceptance Criteria

- Created design and reviewed with team and SHIELD project.

GilmoreA6:
This PR lays out a proposal for mapping and monitoring received observation codes to reportable conditions

## Linked Issues
- Fixes #9248


prime-router/docs/design/proposals/0023-condition-to-code-mapping/0023-condition-to-code-mapping.md
```diff
@@ -0,0 +1,383 @@
...
+| 2.16.840.1.113762.1.4.1146.799 | Influenza (Tests for influenza A virus Antigen)      | 88904-8 | Influenza virus A Ag [Presence] in Lower respiratory specimen by Immunofluorescence | LOINC       | 2.74    | Active | Infection caused by novel Influenza A virus variant (disorder) | 541000000000000  | SNOMEDCT              | 2023-03                       | RCTC         |
+|                                | Influenza - (ABC TESTING LABS)                       | 123456  | Influenza virus A                                                                   | LOCAL       |         | Active | Infection caused by novel Influenza A virus variant (disorder) | 541000000000000  | SNOMEDCT              | 2023-03                       | LOCAL        |   
+
+Example Action Log Error Message: "Missing mapping for: 123456 for table: Observation-Mapping"
```
arnejduranovic, line 261: Check if this applies to other places in the doc, but we need to update this example to include "code system" as well since "code" is not enough.

prime-router/docs/design/proposals/0023-condition-to-code-mapping/0023-condition-to-code-mapping.md
```diff
@@ -0,0 +1,383 @@
...
+
+Example query
+```yaml
+select * from action_log
+where type = 'mapping'
+and created_at >= current date -1
+```
+## Unmapped receiver queue
+
+Creation of a receiver with the correct filters will allow us to catch all messages with unmapped observations in a queue for Engagement to work. Combined with Action Log monitoring engagement will be able to update all mappings and resend affected messages as needed to STLTs. In order to create this receiver the following settings would be needed:
```
arnejduranovic, line 289: What if one observation was able to get mapped but another wasn't? That message would not end up going here? How do we handle this case?

prime-router/docs/design/proposals/0023-condition-to-code-mapping/0023-condition-to-code-mapping.md
```diff
@@ -0,0 +1,383 @@
...
+silent - do not generate diff or ask for confirmation
+activate - activate the table upon creation
+key - apikey to access VSAC FHIR terminology service
+OID - OID of valueset to update (default to all OID in table if blank)
+
+Example CLI command for update tool:
+
+./prime condition-mapping-update --activate --silent --environment staging --key <apikey>
+
+## Sender Compendium Comparison Utility
```
arnejduranovic, line 348: Should update this section to make it clear that we are checking code and codeSystem for equivalency and ignoring description, per our discussion

arnejduranovic (changes requested):

prime-router/docs/design/proposals/0023-condition-to-code-mapping/0023-condition-to-code-mapping.md
```diff
@@ -0,0 +1,383 @@
...
+## Out of Scope
+The below items are not covered in this proposal.
+
+1.) Mechanism to automate which conditions each state considers reportable.
+
+## Design
+
+Overview Diagram
+
+![img.png](code-to-condition diagram.png)
```
arnejduranovic, line 29: Probably should include the "pruning" feature in diagram? Also check for spelling mistakes.

prime-router/docs/design/proposals/0023-condition-to-code-mapping/0023-condition-to-code-mapping.md
```diff
@@ -0,0 +1,383 @@
...
+
+### Observation Mapping Table
+
+The Observation Mapping table will be made up of CSTE ValueSets and contain the following columns:
+
+| Column Name                   | Description                                   | Example                                                                        |
+|-------------------------------|-----------------------------------------------|--------------------------------------------------------------------------------|
+| Member OID                    | ValueSet Identifier                           | 2.16.840.1.113762.1.4.1146.239                                                 |
+| Name                          | ValueSet Name                                 | Chlamydia trachomatis Infection (Tests for Chlamydia trachomatis Nucleic Acid) |
+| Code                          | LOINC or SNOMED coded value                   | 16601-7                                                                        |
```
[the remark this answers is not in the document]
GilmoreA6: It could theoretically be any code system, but we expect that to be a rare edge case.

prime-router/docs/design/proposals/0023-condition-to-code-mapping/0023-condition-to-code-mapping.md
```diff
@@ -0,0 +1,383 @@
...
+
+The column names are taken directly from the [RCTC spreadsheet](https://docs.google.com/spreadsheets/d/1rO_p145xXO1AD76vx8vBqXgoQUnLqfc8/edit#gid=739612351) with the exception of "Value Source" and "Created At" which are additional columns added for administrative purposes that will be used when updating the table. Both LOINC and SNOMED codes are combined in this table and can be identified by column "Code System".
+The RCTC does a fairly good job of keeping up to date with LOINC and SNOMED codes and is regularly updated. It is anticipated that despite this there will be a requirement to map codes that are not present in the condition mapping table. These will have to be mapped manually after review by RS personnel in order to ensure that the proper condition code is mapped to the LOINC or SNOMED code. These codes can be submitted to CSTE valueset reviewers to be included in future releases. If a column is not applicable it can be left blank unless it is the "Code", "Condition Name", "Condition Code" or "Source" columns.
```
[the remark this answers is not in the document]
GilmoreA6: What's linked is the most recent RCTC that is released which is the source of truth for the base table.

prime-router/docs/design/proposals/0023-condition-to-code-mapping/0023-condition-to-code-mapping.md
```diff
@@ -0,0 +1,383 @@
...
+                "valueCodeableConcept": {
+                    "coding": [
+                        {
+                            "system": "http://snomed.info/sct",
+                            "code": "260373001",
+                            "display": "Detected"
+                        }
+                    ]
+                },
+                "meta": {
```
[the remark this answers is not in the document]
GilmoreA6: @arnejduranovic Is there a situation that we know of where an observation resource would have a meta.tag on it from a sender? If we define a custom RS extension then we would have to either create a generic RS extension for "mapping" or make an extension for each situation. The cool think about using meta.tag was the flexibility in that we could use it for more than just condition mapping (i.e. we could tag an observation as an AOE). Couldn't we also just strip out any observation meta.tags received from a sender?

prime-router/docs/design/proposals/0023-condition-to-code-mapping/0023-condition-to-code-mapping.md
```diff
@@ -0,0 +1,383 @@
...
+In order to account for both HL7 and FHIR input from senders, the lookup should occur after conversion of the incoming HL7 V2 message into a FHIR bundle. Since the sender transforms are used after conversion from HL7 to FHIR, adding the below elements to the default sender transform will add the condition tag to the observation resource.
+
+```yaml
+  - name: test-condition
+    resource: 'Bundle.entry.resource.ofType(Observation)'
+    condition: '%resource.code.coding.exists()'
+    bundleProperty: '%resource.meta.tag.code'
+    value: ['%resource.code.coding.code']
+    valueSet:
+      lookupTable:
```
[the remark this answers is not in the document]
GilmoreA6: It doesn't currently. One of the tickets I created #11934 addresses this. I agree, if we go with a custom extension than having it be a list of codes makes sense.

prime-router/docs/design/proposals/0023-condition-to-code-mapping/0023-condition-to-code-mapping.md
```diff
@@ -0,0 +1,383 @@
...
+  condition: '%resource.code.coding.exists() and %resource.meta.tag.code.not.exists()'
+  bundleProperty: '%resource.meta.tag.code'
+  value: ['%resource.valueCodeableConcept.coding.code']
+  valueSet:
+    lookupTable:
+      tableName: Condition-Mapping
+      keyColumn: code
+      valueColumn: condition_code
+```
+This element should be placed after the above elements as we only want to check the mapping for a result value if we have failed to find a match in the test ordered and test performed locations.
```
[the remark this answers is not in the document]
GilmoreA6: We would have to change the monitoring strategy if we did that. Otherwise we would be getting a lot of false positive for "unmapped" snomed codes when the condition was indicated by the LOINC code. To clarify for a particular set of observations the condition will be indicated by either the LOINC or SNOMED code, not both.

prime-router/docs/design/proposals/0023-condition-to-code-mapping/0023-condition-to-code-mapping.md
GilmoreA6: Need to think about this one since that is a complete change to how the lookuptables works. Honestly, probably not worth the effort and would make more sense to create a custom function that runs with the condition filter in that case. Brings up the question of what we do when two or more local codes have the same value and don't map/can't to a LOINC code (extreme edge case). Honestly in that case we should probably tell them to ask Regenstrief to create a LOINC code for them.

prime-router/docs/design/proposals/0023-condition-to-code-mapping/0023-condition-to-code-mapping.md
```diff
@@ -0,0 +1,987 @@
...
+OBX|1|CE|97099-6^COVID-19^LN||260415000^Not detected^SCT|||N^Normal^HL70078^^^^2.7|||F||||||^BD Veritor System for Rapid Detection of SARS-CoV-2 \T\ Flu A+B*^^^^^^^BD Veritor System for Rapid Detection of SARS-CoV-2 \T\ Flu A+B*||||||CSV uploads-11D1111111^^^^^CLIA&2.16.840.1.113883.4.7&ISO^XX^^^11D1111111|123 Main St^^^CA^94553^USA
+NTE|1|L|95419-8 Has symptoms related to condition of interest: No N HL70136
+SPM|1|1519ae86-98f4-40d8-9780-d5211186e5b5&CSV uploads&11D1111111&CLIA^1519ae86-98f4-40d8-9780-d5211186e5b5&CSV uploads&11D1111111&CLIA||697989009^Anterior Nasal Swab^SCT^^^^2.67|||||||||||||20230329051736-0700|20230329051742-0700
+BTS|1
+FTS|1
+```
+During the routing step the results of the condition filter are evaluated and for observations which pass the filter a reference to that observation is added to the FHIR bundle endpoint resource. During the Translate step any observations not referenced in the endpoint resource of the FHIR bundle will not be added to the outgoing report to the receiver. This will include both observations where a mapping could not be found as well as observations representing a reportable condition that is not of interest to the receiver.
```
arnejduranovic, line 911: When we receive FHIR, is it possible some conditions would already be in the endpoint resource? If so, we probably need a different way to communicate if an observation is wanted or not, maybe with another variable in the RS condition extension?

arnejduranovic (changes requested):

arnejduranovic (approved):
Looks great! Design may change slightly as we move forward with implementation, but I expect it'll follow the design laid out here pretty closely. Thank you for all the effort here!

PR #11929, opened 2023-10-23
Added DR datatype mappinng

JessicaWNava:
This PR maps the DR datatype

Test Steps:
1. Make sure the tests run

## Changes
Map Hl7 DR fields to FHIR and back again. 

### Testing
- [x] Tested locally?
- [x] Added tests?

#11817


prime-router/metadata/fhir_mapping/hl7/datatype/HumanName.yml
```diff
@@ -46,11 +46,19 @@ suffix_3:
...
+    valueOf: datatypes/DR/Period
+    condition: XPN.12 NULL && XPN.13 NULL
     expressionType: resource
     vars:
-       start: XPN.12 | XCN.19
-       end: XPN.13 | XCN.20
-     
\ No newline at end of file
+       start: XPN.10.1
+       end: XPN.10.2
```
brick-green, line 64: We had a slack thread going on whether to use a separate file for period or use nested elements. In my PR I went with the nested element strategy like this:
```
period:
  generateList: true
  expressionType: nested
  vars:
    start: XTN.13
    end: XTN.14
  expressionsMap:
    start:
      condition: $start NOT_NULL
      type: STRING
      valueOf: "GeneralUtils.dateTimeWithZoneId(start,ZONEID)"
      expressionType: JEXL
    end:
      condition: $end NOT_NULL
      type: STRING
      valueOf: "GeneralUtils.dateTimeWithZoneId(end,ZONEID)"
      expressionType: JEXL
```

I don't have a strong opinion on which strategy to use but I do think we should norm on it. What do you think?

JessicaWNava: Even if it is used in multiple places as it will eventually be? I thought the point was to have reusable code so it wouldn't need to be maintained in multiple places

brick-green: Yeah that was crux of the thread: https://agilesix.slack.com/archives/C026DV8R21W/p1697730231591709 (you'll probs have to change the url to nava.slack.com). You're definitely right that period is used all over the place so it could go in a datatypes/common folder. Plus there are probably other elements similar to this that could be added to common datatypes. But Kalish made an argument that the nested style keeps the mapping more aligned with the mapping inventory since the inventory doesn't reference a specific Period mapping. I leaned that way myself but it's just a preference thing.

prime-router/metadata/hl7_mapping/datatypes/Address/XAD.yml
```diff
@@ -53,6 +53,15 @@ elements:
     hl7Spec: [ '%{hl7SpecXADField}-11' ]
 
   # TODO: #11817
+  # TODO: #11664
```
arnejduranovic, line 56: Is this comment still relevant? Seems like both TODOs are addressed?

prime-router/metadata/fhir_mapping/hl7/datatypes/XAD/Address.yml
```diff
@@ -169,10 +169,21 @@ extension_7-xad11AddressRepresentationCode:
...
+  vars:
+    xad13: String, XAD.13
+    xad14: String, XAD.14
+    start: String, XAD.12.1
+    end: String, XAD.12.2
+  condition: $xad13 NULL || $xad14 NULL
+  expressionType: resource
+  valueOf: datatypes/DR/Period
+
+# TODO: #11664
```
arnejduranovic, line 184: What is there TODO here?

prime-router/metadata/hl7_mapping/datatypes/Address/XAD.yml
```diff
@@ -53,6 +53,15 @@ elements:
...
+  # XAD.12, XAD.13, XAD.14
+  - name: period
```
arnejduranovic, line 58: This doesn't necessarily result in byte-for-byte equivalence right? Period could have been populated by fields 13 and 14 and, if that's the case, then here you are setting XAD 12 to the values of 13 and 14 when the original HL7v2 did not have XAD 12 set in the first place. Sounds like something that can be solved by extensions for XAD 12, but I'm curious if we think this is worth it. My opinion is the added complexity is not worth it and this is fine as is. @mkalish thoughts?

prime-router/build.gradle.kts
```diff
@@ -829,10 +829,11 @@ dependencies {
     implementation("org.yaml:snakeyaml:2.2")
     implementation("io.github.linuxforhealth:hl7v2-fhir-converter") {
         version {
-            branch = "master"
+            branch = "kalish/use-sorted-map-for-processing-nestedexpression-expressions"
         }
     }
-    implementation("ca.uhn.hapi.fhir:hapi-fhir-structures-r4:6.8.4")
+//    implementation("com.github.focusconsulting:hl7v2-fhir-converter:v1.0.20-SNAPSHOT")
+    implementation("ca.uhn.hapi.fhir:hapi-fhir-structures-r4:6.8.2")
```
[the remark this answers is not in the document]
JessicaWNava: Idk. Locally it is 4 and it says those changes are committed, so I am rather confused.

prime-router/metadata/fhir_mapping/hl7/datatypes/XAD/Address.yml
JessicaWNava: This was leftover from merging in changes that were initially in Kalish's branch

arnejduranovic (approved):
Changes look good (match inventory) and verified tests are correct and testing the new mappings.

PR #11974, opened 2023-10-27
Fixed get_ssh_list.sh to work on Bash versions prior to v4.

dkrylovsb:
Fixed `get_ssh_list.sh` to work on Bash versions prior to v4.

Test Steps:
1. Run Terraform _or_ `get_ssh_list.sh` separately

## Changes
- This PR fixes `get_ssh_list.sh` to work on Bash versions prior to v4. `^^` case transformation does not work in Bash prior to v4, which is what comes with MacOS by default. I.e. this does not work: `upper_env=${environment^^}`

### Testing
- [x] Tested locally?


jeremy-page (commented):
Thank you Denis

Issue #11664, opened 2023-10-06
Update HL7v2 to FHIR to HL7v2 DataType Mappings: DTM / DT / TS


arnejduranovic:
### User Story

As a ReportStream engineer,
I want to update the ReportStream HL7v2 -> FHIR and FHIR -> HL7v2 mappings for the `DTM` and `DT` HL7v2 data types per the mapping inventory
so that I can ensure data does not get dropped during translation and enable the creation of mappings of ORU-R01 segments that reference these HL7v2 DataTypes.

### Description/Use Case

Given the ORU-R01 context, this HL7v2 DataType can map to the following resources (the sub list shows where exactly in what mapping file the DataType mapping appears in):

- [ ] DTM
    - [ ] dateTime
        - [ ] SPM[Specimen]-SPM-18
        - [ ] NTE[ServiceRequest]-NTE-6
        - [ ] OBX[Observation-Component]-OBX-14
        - [ ] OBX[Observation-Component]-OBX-19
        - [ ] OBR[DiagnosticReport]-OBR-7
        - [ ] OBR[DiagnosticReport]-OBR-8
        - [ ] PRT[Device]-PRT-17
        - [ ] PRT[Device]-PRT-18
        - [ ] ORC[ServiceRequest]-ORC-9
        - [ ] ORC[ServiceRequest]-ORC-15
        - [ ] PID[Patient]-PID-7
        - [ ] PID[Patient]-PID-29
        - [ ] PID[Provenance-Patient]-PID-33
        - [ ] PV2[Encounter]-PV2-8
        - [ ] PV2[Encounter]-PV2-9
        - [ ] OBR[Specimen]-OBR-7
        - [ ] OBR[Specimen]-OBR-8
        - [ ] PRT[PractitionerRole]-PRT-11
        - [ ] PRT[PractitionerRole]-PRT-12
        - [ ] PRT[RelatedPerson]-PRT-11
        - [ ] PRT[RelatedPerson]-PRT-12
        - [ ] OBX[Observation]-OBX-14
        - [ ] OBX[Observation]-OBX-19
        - [ ] PV1[Encounter]-PV1-44
        - [ ] PV1[Encounter]-PV1-45
        - [ ] SFT[MessageHeader]-SFT-6
        - [ ] HL7 Data Type - FHIR R4_ DR[DateTime] - Sheet1-DR.1
        - [ ] HL7 Data Type - FHIR R4_ DR[Period] - Sheet1-DR.1
[...]

PR #12065, opened 2023-11-02
DTM/DT/(TS) mapping updates

jimmyfagan:
This PR updates some DTM/DT mappings to match the v2-to-fhir inventory while preserving HL7 DTM/DTs in and out.

Test Steps:
1. Run the integration tests
2. Convert a sample hl7 message containing PID.7, NK1.8/9, and OBR.22 to fhir and back and observe the results

## Changes
- Updated mappings for PID.7, PID.29, and OBR.22 from v2-to-fhir to include extension with the original HL7 string (inlined)
- Updated mappings from HL7 back to PID.7, PID.29, and OBR.22 to use new sub-schema `DTM.yml`
- Added mapping for NK1.8 and NK1.9
- Added _many_ tests for different combinations of datatypes and dateTime precision
- Adjusted _many_ existing tests

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- Addresses #11664


prime-router/src/testIntegration/kotlin/datatests/mappinginventory/dtm/DTMTests.kt
```diff
@@ -0,0 +1,73 @@
...
+
+    @Test
+    fun `test dt to dateTime`() {
+        assert(verifyHL7ToFHIRToHL7Mapping("dt/dt-year-precision-to-dateTime").passed)
+        assert(verifyHL7ToFHIRToHL7Mapping("dt/dt-month-precision-to-dateTime").passed)
+        assert(verifyHL7ToFHIRToHL7Mapping("dt/dt-day-precision-to-dateTime").passed)
+    }
+
+//    @Test
+//    fun `test zero dt to dateTime`() {
```
jimmyfagan, line 70: @arnejduranovic Do you think this case supporting `0000` in NK1.8/9 is necessary? If we don't have to support values that register as non-valid dates, we don't have to use an extension for DT<->dateTime, but we can also just use the extension to make this work without too much trouble.

prime-router/metadata/hl7_mapping/datatypes/dateTime/DTM.yml
```diff
@@ -0,0 +1,7 @@
+elements:
+  - name: dateTime
+    value:
+      - '%resource.extension(%`rsext-hl7v2-date-time`).value'
+      - '%resource.extension("http://hl7.org/fhir/StructureDefinition/patient-birthTime").value'
+      - '%resource'
```
jimmyfagan, line 6: @mkalish The last two values here could be seen as logical fallbacks, but they are harmless in HL7->FHIR->HL7 cases, and they prevent us from needing to add transforms for SR and from needing to update basically every FHIR->HL7 test. I suppose the second value here is kind of optional, but I don't see any ambiguity around whether or not a FHIR dateTime's value should be used for the HL7 DTM in the absence of the rsext extension, so making this slight exception to the general approach seemed okay to me. What do you think?

jimmyfagan:
Added one last commit to remove an empty identifier from FHIR, and it required an update to tons of tests, feel free to ignore that commit (or review it independently) when doing your review.

prime-router/metadata/fhir_mapping/hl7/resource/Patient.yml
```diff
@@ -265,6 +265,39 @@ birthDate:
   valueOf: PID.7
   expressionType: HL7Spec
 
+_birthDate:
+  condition: $pid7 NOT_NULL
+  vars:
+    pid7: PID.7
+    dateTimeIn: PID.7, GeneralUtils.dateTimeWithZoneId(dateTimeIn,ZONEID)
+    isTime: PID.7, GeneralUtils.dateTimeWithZoneId(isTime,ZONEID).length() > 10
```
mkalish, line 273: This is good to know that this works, I think we had opened a bug, but might have been looking for a an operator rather than a Java solution.

I think this means we can remove this [now](https://github.com/CDCgov/prime-reportstream/blob/544cc8ccb5120c35653a21763e33a5835f502800/prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt#L335)?  Also, is there reason we need both extensions?  It looks like it would contain the same info.


Maybe just a comment as to the difference?

prime-router/metadata/fhir_mapping/hl7/secondary/PatientContact.yml
```diff
@@ -65,6 +65,13 @@ telecom_business:
   constants:
     use: work
 
+period:
```
mkalish, line 68: Just to confirm, this is here to test converting to Datetime?

prime-router/metadata/hl7_mapping/datatypes/dateTime/DTM.yml
mkalish: Avoiding having to make all the other changes sounds like a good reason to me, but maybe drop a comment around why the fallback exists and what the long term would look like?

prime-router/src/testIntegration/kotlin/datatests/mappinginventory/TranslateAndCompareFile.kt
```diff
@@ -26,65 +26,63 @@ fun translateAndCompareHL7ToFHIR(inputFile: String, expectedOutputFile: String):
...
-
-    val hl7ToFhirResult = TranslationTests().FileConversionTest(
-        hl7ToFhirConfig
-    ).runTest()
-
-    if (!hl7ToFhirResult.passed) {
-        return hl7ToFhirResult
+fun verifyHL7ToFHIRToHL7Mapping(
+    testFileName: String,
+    skipHl7ToFhir: Boolean = false,
```
mkalish, line 31: Doesn't look like these changes are used anywhere?  What's the use case you were envisioning?

prime-router/src/testIntegration/resources/datatests/HL7_to_FHIR_to_HL7/elims_1_72_3029202646_5532_NoPII-expected.hl7
```diff
@@ -4,26 +4,26 @@ PID|1||FPID00007844^^^STARLIMS.CDC.Prod&2.16.840.1.114222.4.3.3.2.1.1&ISO^PI~501
...
-OBR|2|23FMC-209M0173^SPHL-000008^2.16.840.1.114222.4.1.3666^ISO|72_3029202646_3029202646_716^STARLIMS.CDC.Prod^2.16.840.1.114222.4.3.3.2.1.1^ISO|PLT1141^JCV IgM XXX Ql IA^PLT^716^JC IgM^L^2.74^v unknown^^CDC-10282^Arbovirus Serology^L^^2.16.840.1.113883.6.1|||20230728151300+0000|||||||||FTC-3573^Novant Health Kernersville Medical Center^^^^^^^STARLIMS.CDC.Prod&2.16.840.1.114222.4.3.3.2.1.1&ISO^^^^XX||||||20230821155000-0400|||F
+OBR|2|23FMC-209M0173^SPHL-000008^2.16.840.1.114222.4.1.3666^ISO|72_3029202646_3029202646_716^STARLIMS.CDC.Prod^2.16.840.1.114222.4.3.3.2.1.1^ISO|PLT1141^JCV IgM XXX Ql IA^PLT^716^JC IgM^L^2.74^v unknown^^CDC-10282^Arbovirus Serology^L^^2.16.840.1.113883.6.1|||20230728151300+0000|||||||||FTC-3573^Novant Health Kernersville Medical Center^^^^^^^STARLIMS.CDC.Prod&2.16.840.1.114222.4.3.3.2.1.1&ISO^^^^XX||||||202308211550-0400|||F
```
mkalish, line 11: Nice

prime-router/metadata/fhir_mapping/hl7/resource/Patient.yml
mkalish: Nevermind this makes sense now for places with less precision

mkalish (approved):
Really good stuff!  Tests are very comprehensive and looks like we now have much closer 1:1 for ELIMS.

Two thoughts:
- Could consider splitting out the change to identifier to reduce the number of files; my main thought is just someone might was to look back at this PR to understand how dates work in the conversion
- Just confirming, that we're not going to handle TS.2 and maybe worth adding a note

prime-router/src/testIntegration/kotlin/datatests/mappinginventory/TranslateAndCompareFile.kt
jimmyfagan: Ah good call, I forgot to call this out in the description. This was useful for me while trying to make sure the HL7->HL7 was working without having to worry about the fhir files being present or up-to-date. It probably is mostly for flexibility of tests during development, but could theoretically be used to have more tests that only deal with a particular combination of the input/output formats to cut back on the volume of test files we have. Definitely not closely related to the goals of this PR, so I could see a case for removing it.

prime-router/metadata/fhir_mapping/hl7/secondary/PatientContact.yml
jimmyfagan: Specifically, this is to test converting to DateTime from a HL7 DT (Date), instead of a DTM (DateTime). DT is pretty uncommon in ORU_R01, so even though it's kind of a compound usage within `period`, this was the simplest case I could find.

prime-router/metadata/fhir_mapping/hl7/datatype/Identifier_SystemID.yml
```diff
@@ -110,6 +110,7 @@ extension_5:
   generateList: true
   valueOf: extension/Extension
   expressionType: resource
+  condition: $value NOT_NULL
```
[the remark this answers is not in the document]
jimmyfagan: Yea, definitely good point about splitting this out, just was touching a lot of these files already anyway. I'll think about it, maybe I can just link to the change view with the first few commits in the description to address that concern of somebody looking back at this PR without having to do too much reworking.

prime-router/src/testIntegration/kotlin/datatests/mappinginventory/dtm/DTMTests.kt
jimmyfagan: We talked about this over slack, and it sounds like there could be reason to support this. Will try to uncomment this test and update the NK1.8/9 mapping to support this case.

mkalish (approved):
New changes look good to me, going to get this merged since we'll want to reference this for other dates

PR #12169, opened 2023-11-08
Engagement/angela/11780/bugfix sent files

### LINKED ISSUE #11780: LA - Azure Alert - Production - Fatal Alert
## **Problem Statement**
On the Slack ReportStream onboarding-operations Channel received a "Found Fatal Alert" in Production FunctionApp.

## **What you need to know**
TenantId:	059ec6c1-498f-48d1-a59a-8df9bfdca183
TimeGenerated:	2023-10-13T14:52:20
Category:	Host.Function.Console
Location:	eastus
Message:	[1;31m14:52:20.432 [pool-2-thread-305][1810] FATAL-ALERT gov.cdc.prime.router.azure.SendFunction - Success: sftp upload of cdcprime_cdcprime_laoph_production_production_20231011213558.hl7 to SFTPTransportType(host=204.58.124.41, port=22, filePath=./, credentialName=null), FAILED Sftp upload of inputReportId f37664a4-fa8b-46e3-9fd0-f099a4ea61d3 to SFTPTransportType(host=204.58.124.41, port=22, filePath=./, credentialName=null) (orgService = la-doh.elr), Exception: Status code 409, '?<?xml version='1.0' encoding='utf-8'?><Error><Code>BlobAlreadyExists</Code><Message>The specified blob already exists.
HostVersion:	4.25.2.2
HostInstanceId:	d8b720dc-1518-4dbe-bc9c-488588e1e3ae
Level:	Information
AppName:	pdhprod-functionapp
RoleInstance	0d1d95cb5bca24eae97e7d7a3b08580f2c1be3b13b58a879b77193a70e7cfdf4
LevelId	2
ProcessId	25
SourceSystem	Azure
Type:	FunctionAppLogs
_ResourceId:	/subscriptions/7d1e3999-6577-4cd5-b296-f518e5c8e677/resourcegroups/prime-data-hub-prod/providers/microsoft.web/sites/pdhprod-functionapp

## **Acceptance Criteria**
Assign to Engineer
Investigate the cause/reason for the error message.
Determine if an action/resolution is needed.
Close ticket when resolved.

## **To Do**
Please reach out the Helpdesk if you need assistance in reaching out the State/STLT.

## **Additional Comments**
If applicable, add screenshots, links, etc. to further explain the ticket

adegolier:
This PR ...

The actual problem that was causing the failures in the linked ticket was that la-doh was having files generated with the same name for the elr and elr-secondary receiver. After sending, files were being uploaded to the blobstore in none/send. This PR changes the location files are uploaded to the blobstore to none/receiver.fullname which will put the files with the same name for different receivers into different folders. Added an EventAction.OTHER to be used as a default and changed the default action on BlobAccess.uploadBody() so that if it's called with no action for some reason, those files will be separated from files that have completed the pipeline. I also successfully built without the action to confirm that there was nowhere currently calling uploadBody() using the default action. The added test successfully failed in master and passes on this branch.

Test Steps:
1. run the test that was added
2. verify it passes
3. copy the test
4. check out master
5. paste the test into the appropriate file in master
6. run the test
7. verify it fails
8. (revert your code)
9. switch back to the branch
10. remove the default action from BlobAccess.uploadBody() and confirm that it still builds (I believe there may be one test somewhere that uses the default, so you may need to temporarily change that test also)
11. Unsure that manual testing can be performed locally.

## Changes
- Changed ActionHistory.trackSentReport() to call BlobAccess.uploadBody() with a subfolderName argument of receiver.fullName instead of "send".
- Changed the default action parameter in BlobAccess.uploadBody() from EventAction.NONE (which is what is passed by ActionHistory.trackSentReport()) to EventAction.OTHER.
- Added EventAction.OTHER and its related TaskAction.
- Added EventAction.OTHER to WorkflowEngine.finishedField() to be similar to EventAction.NONE. The differentiating point is that NONE is for when actions have been completed and no further action is needed and OTHER is meant to be used as a default in case something goes wrong.
- Added migration file to create the OTHER TaskAction. TaskAction is autogenerated by jooq.
- Updated ActionHistoryTests.test trackSentReport() to mock BlobAccess.uploadBody() using EventAction.NONE because that's what ActionHistory.TrackSentReport() is using.
- Added ActionHistoryTests.test trackSentReportSameFilenameDifferentReceiver() to test that the change to ActionHistory.trackSentReport() would successfully result in different blob urls for a file with the same name going to two receivers.
- Updated BlobAccessTests.upload body() to reflect the change to the default action in BlobAccess.uploadBody()

### Testing
- [X] Tested locally?
- [X] Added tests?

### Process

## Linked Issues
- Fixes #11780


prime-router/src/main/kotlin/azure/Event.kt
```diff
@@ -35,6 +35,7 @@ abstract class Event(val eventAction: EventAction, val at: OffsetDateTime?) {
         WIPE_ERROR, // Deprecated
         RESEND,
         REBATCH,
+        OTHER, // a default/unknown
```
mkalish, line 38: ❤️

prime-router/src/test/kotlin/azure/ActionHistoryTests.kt
```diff
@@ -465,4 +466,45 @@ class ActionHistoryTests {
...
+        every { inReportFile.itemCount } returns 15
+        val actionHistory1 = ActionHistory(TaskAction.receive)
+        actionHistory1.action
+        actionHistory1.trackSentReport(org.receivers[0], uuid, "filename1", "params1", "result1", header)
+        assertThat(actionHistory1.reportsOut[uuid]).isNotNull()
+        val actionHistory2 = ActionHistory(TaskAction.receive)
+        actionHistory2.action
+        actionHistory2.trackSentReport(org.receivers[1], uuid2, "filename1", "params1", "result1", header)
+        assertThat(actionHistory2.reportsOut[uuid2]).isNotNull()
+        assertNotEquals(blobUrls[0], blobUrls[1])
```
mkalish, line 508: Nit: could maybe check here that the `receiver.fullName` is in the urls

prime-router/src/main/kotlin/azure/Event.kt
```diff
@@ -55,6 +56,7 @@ abstract class Event(val eventAction: EventAction, val at: OffsetDateTime?) {
                 WIPE_ERROR -> TaskAction.wipe_error
                 RESEND -> TaskAction.resend
                 REBATCH -> TaskAction.rebatch
+                OTHER -> TaskAction.other
```
mkalish, line 59: Maybe just a comment here explaining that `OTHER` is not an expected value but rather a logical fallback default for `uploadBody`?

I could also see a tech debt ticket to investigate if `uploadBody` needs the default param at all

mkalish (approved):
:shipit: Great work!

I verified that the new test fails on `master` and also ran the smoketests and saw that the none directory was correctly populated

oslynn:
Tested. LGTM.  Thank you for the fix.

oslynn (approved):

prime-router/src/test/kotlin/azure/ActionHistoryTests.kt
adegolier: I think I didn't do that because I don't necessarily care that it has the receiver's name... just that they aren't going to the same place. But that is an easy change, so added and the test still passes.

prime-router/src/main/kotlin/azure/Event.kt
adegolier: I feel like in theory it shouldn't need the default, but.... there's defaults like this in a bunch of other places for some reason, which makes me feel like I might be missing something. Might be for graceful failure if something goes wonky with Azure/the queues.

PR #12190, opened 2023-11-11
Experience/12063-Create-custom-console-to-use-instead-of-global

### LINKED ISSUE #12063: Create custom console to use instead of global
## Impact
Ensures that whatever we send to App Insights is from within our application, and not from anything outside the application. Anything that currently uses the console is sending to App Insights.

This will ensure we are getting accurate data to App Insights. 

## Description / Background
Ott had gotten a pager duty alert on a FE alert that should not have been triggered.

Pager Duty uses App Insights to trigger alerts.

Our current implementation for intercepting calls to the global console to emit telemetry events has major flaws. 

- It requires modifying a native global variable which is bad practice
- It causes us to receive events for things that are outside our concern
  - user-side browser messages (ex: failed to load a valid css file which can be due to anything)
  - technically any extensions that the user has installed could spam our telemetry with emissions
 
We need to instead create a new custom global variable that mimics console and calls our function that handles calling the real console and emitting telemetry. With that, we can simply find+replace all current usages of global console within our app to the new custom console. This will ensure that we are only sending telemetry sourced from our app.

## Acceptance Criteria
- [ ] Custom object created that nothing else is using
- [ ] Using it within a context and not as a global variable

jpandersen87:
Fixes #12063 

This PR introduces and implements a new proposal for handling console and telemetry emissions.

- Replaces our custom Error Boundary component (RSErrorBoundary) with one derived from the `react-error-boundary` library
  - library recommended directly from react devs
- A new rsconsole object with methods that call to console and emit telemetry is available from session context
  - see proposal for more details
- All usages of global console either replaced with rsconsole where appropriate or removed entirely
- special console mock object available for import and usage in tests
  - initializes with spies on original methods (spied but NOT mocked)
  - has helper functions built-in to do certain actions on all spies (reset, restore, mock implementation)
- test utility function conditionallySupressConsole removed as it is no longer needed
  - as the new norm proposed would end all usage of console, we should rarely need to spy or mock console
  - rsconsole would be the correct mock to monitor per norm
  - use new console mock object to explicitly mock a particular method's implementation so as not to emit to terminal
- running tests should no longer emit garbage console data
  - any console emissions you see now should be legitimate issues needing investigation!
- AlertNotifications converted to be a Toast context available to use via appropriate hook
  - showError function removed
  - toast function updated
  - mock file that exposes a mock context to spy on toast function created
    - importing this mock context will automatically ensure the useToast hook is mocked with an implementation that returns the mock context
    - thus no additional work is needed when testing for proper toast calls (no need to wrap with provider) 
- all contexts moved into their own dedicated subfolders


frontend-react/src/TelemetryService.test.ts
penny-lischer, line 1: Are there any other tests that can be added here? Seems minimal.

frontend-react/src/components/Admin/AdminFormEdit.tsx
```diff
@@ -92,8 +92,9 @@ export const TextAreaComponent = (params: {
                             // checkJson made sure the following JSON.parse won't throw.
                             params.savefunc(JSON.parse(text));
                         } else {
-                            showError(
+                            showToast(
                                 `JSON data generated an error "${errorMsg}"`,
```
penny-lischer, line 96: Should we also display the `text` here? would give more insight to the error possibly.

frontend-react/src/contexts/Session/index.tsx
```diff
@@ -42,6 +44,7 @@ export interface RSSessionContext {
     setActiveMembership: (value: Partial<MembershipSettings> | null) => void;
     config: AppConfig;
     site: typeof site;
+    rsconsole: RSConsole;
```
penny-lischer, line 47: nit, should this be rsConsole?

penny-lischer (approved):
LGTM

frontend-react/src/contexts/Toast/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+            position: "bottom-center",
+            hideProgressBar: true,
+        },
+    );
+    toast.clearWaitingQueue(); // don't pile up messages
+};
+
+export const useToast = () => useContext(ToastContext);
+
+export function ToastProvider({ children }: React.PropsWithChildren) {
```
etanb, line 46: Style preference: we should use Direct Imports for all of our methods since it's much less verbose, and we can easily see at the top of the file what React imports we're using so: 

```
import React, { PropsWithChildren } from 'react';
```

etanb:
Just clarifying something: you're essentially replacing `AlertNotifications.tsx` with `frontend-react/src/contexts/Toast/index.tsx`?

jpandersen87:
> Just clarifying something: you're essentially replacing `AlertNotifications.tsx` with `frontend-react/src/contexts/Toast/index.tsx`?

Yes

frontend-react/src/contexts/Toast/index.tsx
etanb: ie, we use this syntax when using `useEffect`, `useState` etc.

frontend-react/src/components/Admin/AdminFormEdit.tsx
jpandersen87: The content of any current toast emissions should not have been touched. Any changes wanted on those should be done in another ticket.

etanb (approved):

Issue #12044, opened 2023-11-01
Test CDC-ELIMS messages: Part 7 - DateTimes


arnejduranovic:
### User Story
As ELIMS, I would like HL7v2 DateTime fields to get translated correctly when going from HL7v2 -> FHIR -> HL7v2

### Description/Use Case
Presently, there are two scenarios identified in this [spreadsheet](https://cdc.sharepoint.com/:x:/r/teams/ReportStream/Shared%20Documents/ELIMS-ELR/Gap%20Analysis/HL7%20IN%20vs%20HL7%20OUT%20_%20Gap_Analysis.xlsx?d=wfaed6c19ba414f03a1486c8490228a1c&csf=1&web=1&e=YXLfAV) that relate to DateTime differences:

1. Rows 6 - 14: RS is converting the timezone to +0000. The resulting time is correct, but we should preserve the original timezone.
2. Rows 15 - 21: When given a DateTime containing ONLY YY/MM/DD, we are adding time which is incorrect. The FHIR spec supports DateTime values that only contain YY/MM/DD.

**IMPORTANT NOTE:** The above two issues DO NOT appear when running the translation via prime CLI, like so: 

```
primeCLI --args='fhirdata --output-format HL7 -i /Users/arnej/Downloads/julie-date-issue/29_05059412_35054.HL7 -o /Users/arnej/Downloads/julie-date-issue/cmd-line-convert-out.hl7 -s /Users/arnej/Projects/prime-reportstream/prime-router/metadata/hl7_mapping/ORU_R01/ORU_R01-base.yml'
```

But they DO APPEAR when running the message through the pipeline via submission API. When I debugged, it seems like the issue is happening inside `FhirToHl7Converter.convert`, meaning the HL7v2->FHIR translation is NOT the problem, but rather something happens during the FHIR -> HL7v2 translation. At the start of this method, the bundle submitted via the cmd line and the bundle submitted via the API are identical but the value of `return message` is different.

### Acceptance Criteria
- [ ] [Sample test message](https://cdc.sharepoint.com/:f:/r/teams/ReportStream/Shared%20Documents/ELIMS-ELR/Gap%20Analysis/12044?csf=1&web=1&e=rOR9Y6) run through pipeline and output message confirmed to equal input message.
- [ ] Understand why CLI produces different output than running through pipeline. Issue fixed or new ticket created.

Andrey-Glazkv:
[...]

PR #12220, opened 2023-11-13
Timezones and times fixes for ELIMS

jimmyfagan:
This PR updates several mappings to make use of the new date/datetime conversions from HL7->FHIR->HL7. This covers OBR.7, OBR.8, ORC.9, OBX.14, OBX.19, OBX.5, MSH.7, ORC.15, OBR.14, SPM.17, SPM.18, and SFT.6. There was also a secondary change here to help ensure that running with the CLI and TranslationTests more closely mimics what our running application's behavior will be during translation.

Test Steps:
1. Run the integration tests and verify that they pass.

## Changes
- Updated several mappings
- Updated `ProcessFhirCommands.convertToHl7` and `TranslationTests.translateFromFhir` to ensure a `HL7TranslationConfig` is used in those paths.
- Updated `CustomTranslationFunctions.convertDateTimeToHL7` to ensure we don't accidentally use that function without a config present (updated a few tests to use `Hl7TranslationFunctions().convertDateTimeToHL7` instead since they didn't have a config)
- Updated many translation tests
  - Pared down some test files that were broader than they needed to be (NTE tests, multiple observation bundles)
  - Updated expected output for some HL7->HL7 tests to now expect preserved input/output for more date fields
  - Updated expected output for some FHIR->HL7 tests to expect a new timezone after the changes to the translation tests translation configs
  - Updated many many FHIR files for datatype tests to reflect MSH.7's fhir mapping

### Testing
- [x] Tested locally?

## Linked Issues
- Addresses #12044


prime-router/src/main/kotlin/fhirengine/engine/CustomTranslationFunctions.kt
```diff
@@ -28,38 +28,36 @@ class CustomTranslationFunctions(
         element: ConverterSchemaElement?,
         constantSubstitutor: ConstantSubstitutor?,
     ): String {
-        if (appContext?.config is HL7TranslationConfig) {
-            val receiver = appContext.config.receiver
-            val config = appContext.config.hl7Configuration
-            var dateTimeFormat = receiver?.dateTimeFormat
+        check(appContext != null)
```
arnejduranovic, line 31: Probably would be cleaner to change this method to make appContext not nullable?

prime-router/src/testIntegration/resources/datatests/HL7_to_FHIR_to_HL7/elims_2_40_05059364_34872_MIN-expected.hl7
```diff
@@ -3,6 +3,6 @@ SFT|CDC^^^^^CDC&2.16.840.1.114222.4&ISO^XX^^^CDC CLIA|ELIMS V11|STARLIMS|Binary
...
-SPM|1|^3015960902&STARLIMS.CDC.Stag&2.16.840.1.114222.4.3.3.2.1.2&ISO||119297000^Blood specimen^SCT^BLD^Blood^L^0912017^Adobe_Code^Blood||||||||||,||||20230823181001+0000
\ No newline at end of file
+OBR|1|CDC_3015960902^CDC^1.111.122311.222.44.2.3.3^ISO|40_3015960902_05059364_1087^STARLIMS.CDC.Stag^2.16.840.1.114222.4.3.3.2.1.2^ISO|PLT1228^Mold and Yeast XXX MS.MALDI-TOF^PLT^1087^MALDI-TOF-CLIA^L^2.74^v unknown^^^^^^2.16.840.1.113883.6.1|||0000|||||||||SPHL-000049^PR Public Health Lab^^^^^^^STARLIMS.CDC.Stag&2.16.840.1.114222.4.3.3.2.1.2&ISO^^^^XX|^NET^Internet^PRDHPHL@salud.gov.pr|||||202308231924-0400|||F
+OBX|1|CWE|PLT1228^Mold and Yeast XXX MS.MALDI-TOF^PLT^3562^MALDI-TOF-CLIA^L^2.74^v_unknown^MALDI-TOF-CLIA|ZZYGQ1OS-1|53326005^Candida albicans (organism)^SCT^^^^09012018^^Candida albicans||||||F|||0000|11D0668319^Centers for Disease Control and Prevention^CLIA^40^Fungus Reference Laboratory^L|NXQ0@cdc.gov^Anderson^Christopher|||20230823191530-0400||||Centers for Disease Control and Prevention^L^^^^CLIA&2.16.840.1.113883.4.7&ISO^XX^^^11D0668319|1600 Clifton Rd^^Atlanta^GA^30329^USA^B
```
arnejduranovic, line 7: So regarding the 0000 value, it seems like the only way we would have 0000 here is if that was the original value of the HL7v2 that was sent to us? What if the sender sent FHIR with a missing date, what would happen? According to HL7v2 spec, we should put `0000`.

arnejduranovic (commented):
My summary of my understanding of the changes, please correct me if I am wrong (or add more detail!) and in the future it would be good to include the owner's summary in the PR description so we can refer back to understand certain changesets:

We have updated the old/original HL7v2 mappings for datetimes to populate the RS extension preferred by the `DTMorDT.yml` FHIR->HL7v2 mapping logic, which was designed/implemented during our current mapping-inventory-alignment effort. This approach helps address the issues in the linked ticket and helps the current mappings align closer to the future mappings, making the migration of users easier in the future. The strategy, as determined by the mapping-inventory-alignment effort, leans towards lossless translation of datetimes as we move from HL7v2 to FHIR and back to HL7v2, hence the use of extensions.

prime-router/src/testIntegration/resources/datatests/HL7_to_FHIR_to_HL7/elims_2_40_05059364_34872_MIN-expected.hl7
arnejduranovic: I will try to find where this is documented and create a new ticket for this work.

arnejduranovic (approved):

PR #12240, opened 2023-11-14
remove env block, add better safety

etanb:
Hotfix to DAP Script issue.


jpandersen87 (approved):
Approving but expecting follow-up as this script shouldn't run in env dev.

penny-lischer (commented):
lgtm

PR #12258, opened 2023-11-15
11660: add EIP datatype mappings

jalbinson:
This PR adds the mappings for EIP to Identifier.

Test Steps:
1. Verify that mappings match the inventory
2. Verify the tests pass

## Changes
- Mapping for EIP -> Identifier

### Testing
- [x] Tested locally?
- [x] Added tests?

### Process

## Linked Issues
- #11660


prime-router/metadata/fhir_mapping/hl7/datatypes/EIP/Identifier-PlacerAssignedIdentifier.yml
```diff
@@ -0,0 +1,113 @@
+value:
+  type: STRING
+  valueOf: EIP.1.1
+  expressionType: HL7Spec
+
+type:
```
mkalish, line 6: It looks like this can be hardcoded?  The mapping specifies that this should always be the same value right?

mkalish (changes requested):
I think `type` in both mappings can be hardcoded

prime-router/metadata/fhir_mapping/hl7/datatypes/EIP/Identifier-PlacerAssignedIdentifier.yml
jalbinson: Yup it looks like I can remove the variables and just hardcode the strings

JessicaWNava (changes requested):
Looking at EIP-to-Identifier-PlacerAssignedIdentifier.hl7, there are 4 items for SPM2.1 when https://docs.google.com/spreadsheets/d/1Ez3350XkVw7cS5ufWsFVpVXV884WOtutsWFfEAZUUvo/edit?usp=sharing shows there should be 3. I also am concerned because I don't see SPM.2.2 where it is supposed to be. The value is CSV in HL7 but type.coding.code is set to PGN instead of CSV. System also doesn't  map to 11D1111111 which is the value in SPM 2.3. f34b0f57-1601-4480-ae8a-d4006e50f38d is also in value when it seems not to have a mapping so it should be in an extension.

jalbinson:
@JessicaWNava I talked with @mkalish about the mapping spreadsheet and we think it's incorrect. An `EIP` datatype contains 2 `EI` datatypes which each contain 4 strings (as shown in the NIST spec).

`type.coding.code` and `type.coding.system` are hardcoded to values according to the spreadsheet so they should not receive any mapped values from HL7.

Can you clarify what you mean by `SPM.2.3`? That should not exist and I double checked the mappings and they look like they're ending up in the right place.

JessicaWNava:
> @JessicaWNava I talked with @mkalish about the mapping spreadsheet and we think it's incorrect. An `EIP` datatype contains 2 `EI` datatypes which each contain 4 strings (as shown in the NIST spec).
> 
> `type.coding.code` and `type.coding.system` are hardcoded to values according to the spreadsheet so they should not receive any mapped values from HL7.
> 
> Can you clarify what you mean by `SPM.2.3`? That should not exist and I double checked the mappings and they look like they're ending up in the right place.
Sorry, for SPM.2.2, you have adb4a5cc-50ec-4f1e-95d7-0c1f77cacee1&CSV&11D1111111&CLIA which has 4 items, however, looking at https://docs.google.com/spreadsheets/d/1Ez3350XkVw7cS5ufWsFVpVXV884WOtutsWFfEAZUUvo/edit?usp=sharing there are only EIP.1 and EIP.2 for SPM.2 which I would think would mean that there would only be two items in your message.

jalbinson:
Yup thats why we think the spreadsheet is wrong. Looking at the NIST spec we see this which makes more sense given how an `EIP` datatype is defined.

Screenshot 2023-11-20 at 10 59 55 AM

prime-router/metadata/fhir_transforms/senders/SimpleReport/simple-report-sender-transform.yml
```diff
@@ -276,3 +277,48 @@ elements:
     resource: '%diagnosticReport.basedOn.resolve().requester.resolve().organization.resolve().telecom.where(system = "phone")'
     resourceIndex: telecomIndex
     schema: datatypes/simple-report-telecom
+
+  # populate Specimen identifier
+  - name: sr-specimen-placer-assigned-identifier-value
+    bundleProperty: '%specimen.identifier[0].value'
```
mkalish, line 283: Just to be safe, I would maybe moved `%specimen.identifier[0]` to be the resource and then have a condition to check for it?

You could also consider breaking this into a separate file, but up to you

prime-router/metadata/fhir_transforms/senders/SimpleReport/simple-report-sender-transform.yml
```diff
@@ -276,3 +277,48 @@ elements:
     resource: '%diagnosticReport.basedOn.resolve().requester.resolve().organization.resolve().telecom.where(system = "phone")'
     resourceIndex: telecomIndex
     schema: datatypes/simple-report-telecom
+
+  # populate Specimen identifier
```
mkalish, line 281: Maybe a little more comment around why these are required?  It looks like the schema was previously falling back to the the performer on the diagnostic report.

It would be good to capture that here so we can work with SR to update it

mkalish (approved):
Couple of small things, but LGTM

prime-router/metadata/fhir_transforms/senders/SimpleReport/simple-report-sender-transform.yml
jalbinson: I'm actually overwriting `%specimen.identifier[0]` on purpose.

In the example Simple Report FHIR files it is:
```
"identifier" : [
          {
            "value" : "5f0d29c2-4181-4a1e-8774-b88338602681"
          }
        ],
```

But that value never shows up in the expected HL7. Instead we see the `diagnosticReport.id`.

mkalish: ah ok... yea even more reason to try to distill what's going on here in a comment.

jalbinson: can do

JessicaWNava (approved):
LGTM! Thank you for the clarification.

PR #12259, opened 2023-11-15
Remove dummy action from build_hub

### LINKED ISSUE #12235: Remove dummy action from build_hub
As a devops engineer I want to make sure only properly tested code is deployed to production.  Remove the dummy workflow and use conditionals within the real workflow.

We have two GHA files -- build_hub.yml and build_hub_dummy.yml -- both of which define workflows with the same name: Build Hub. These workflows trigger upon commits to mutually exclusive set of paths; however, if a PR includes changes to both sets of path (as was the case), both workflows will trigger.

dkrylovsb:
This PR removes the "dummy" action from the Build Hub workflow. 

Test Steps:
1. Open up a PR with changes to one or more of the following paths:
            - 'prime-router/**'
            - 'operations/**'
            - '.github/actions/build-backend/action.yml'
            - '.github/actions/deploy-backend/action.yml'
            - '.github/workflows/build_hub.yml'
   Build Hub should run as part of the checks and actually build the router.
2. Open up a PR *without* any changes to the paths above. Build Hub should still run, but quicker: should skip building the router.

## Changes
- Removed `build_hub_dummy.yml` and changed build_hub.yml to _always_ kick off but conditionally skip the router build based on which files actually changed.

### Testing
- [x] Tested locally?
## Linked Issues
- Fixes #12235


.github/workflows/build_hub.yml
```diff
@@ -4,11 +4,9 @@ on:
   pull_request:
     branches:
       - master
-    paths:
-      - "prime-router/**"
```
snesm, line 8: deleting these will make the backend build on all frontend PR's

.github/workflows/build_hub.yml
```diff
@@ -22,7 +20,7 @@ jobs:
       has_router_change: ${{ steps.build_vars.outputs.has_router_change }}
     steps:
       - name: "Check out changes"
-        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # This is v4.1.1
```
snesm, line 23: No need for the comment, Dependabot is likely to replace the version "soon" anyways.

snesm (changes requested):

.github/workflows/build_hub.yml
dkrylovsb: @snesm Backend builds will be skipped because frontend changes won't flip `has_router_change` to true:

```    if: ${{ needs.pre_job.outputs.has_router_change == 'true' }}```

dkrylovsb: Correct @JosiahSiegel . It turned out the conditionals had been in place already before I started working on this. Pretty much all I had to do is delete the dummy workflow.

snesm:
@dkrylovsb there is also a frontend dummy: build_frontend_dummy.yaml
which presumably causes the same issue.

dkrylovsb:
@snesm That's right. It will be dealt with in a separate ticket, I assume.

snesm (approved):

PR #12277, opened 2023-11-16
The elims transform just negated the default transform and since none…

JessicaWNava:
… of the things in the default transform actually belong there, I renamed it to something more appropriate, deleted the elims sender transform, and updated the places that it needed to be updated in the code. Once this gets pushed out staging will need to be updated so that they no longer have a sender transform.

This PR renames the default transform to something more appropriate, removes it from use for ELIMS, deletes the ELIMS transform since that is only used to negate the default one.

Test Steps:
1. Verify that the ELIMS sender is not using a sender transform anymore.
2. Make sure the tests still pass

## Changes
- Renames the default transform to something more appropriate
- Removes it from use for ELIMS
- Deletes the ELIMS transform since that is only used to negate the default one.

### Testing
- [ X] Tested locally?
- [ ] Added tests?


JFU-NAVA-PBC (commented):
Ran find thru the prime-router/src, compared the renamed schema yml
grepped thru the src for references to the default schema yml
gone thru the git changed file list, doc and schema ymls look good

Running smoke, and tests now.........

JFU-NAVA-PBC (commented):
Ran into the same smoke error as the CI check:

***async end2end_up FAILED***: Convert report id null
*** Tests FAILED:  end2end_up ***

Probably broke something.

otherwise, looking good.

prime-router/docs/universal-pipeline/convert.md
```diff
@@ -7,26 +7,40 @@ The convert step is executed when the [Azure AQS message](README.md#aqs-usage) r
...
+> new formats, they would need to be transformed to FHIR by some new method.
 
 ## Sender Transforms
-After the conversion of an Item to the FHIR format, an additional transform can run on the resulting FHIR bundle. This transform is called a "Sender Transform" because it is used to make sender-specific adjustments to the FHIR bundle.
+After the conversion of an Item to the FHIR format, an additional transform can run on the resulting FHIR bundle. This 
+transform is called a "Sender Transform" because it is used to make sender-specific adjustments to the FHIR bundle.
 
-Sender Transforms are completely optional, but ReportStream has created a default sender transform that can be extended or used directly (`prime-router/metadata/fhir_transforms/senders/default-sender-transform.yml`). This is a good place to start to understand what kind of transforms a Sender may want performed on their messages.
+Sender Transforms are completely optional, but ReportStream has created a sender transform that mimics the behavior of 
+original pipeline (`prime-router/metadata/fhir_transforms/senders/default-sender-transform.yml`). This is a good place
```
JFU-NAVA-PBC, line 32: Should this be refactored to be consistent with the yml files?

JFU-NAVA-PBC (commented):
there seems a reference to default-sender-transform.yml in the convert doc.

JFU-NAVA-PBC:
smoke did not pass on local, have similar failures as in CI smoke:
*** TEST FAILED*** for ignore.CSV action send:  Expecting 5 item lineage records but got 0
Test passed: for ignore.HL7 action receive:  Expecting 5 item lineage records and got 5
Test passed: for ignore.HL7 action batch:  Expecting 5 item lineage records and got 5
*** TEST FAILED*** for ignore.HL7 action send:  Expecting 5 item lineage records but got 0
Test passed: for ignore.HL7_BATCH action receive:  Expecting 5 item lineage records and got 5
Test passed: for ignore.HL7_BATCH action batch:  Expecting 5 item lineage records and got 5
*** TEST FAILED*** for ignore.HL7_BATCH action send:  Expecting 5 item lineage records but got 0
Test passed: for ignore.HL7_NULL action receive:  Expecting 5 item lineage records and got 5
Test passed: for ignore.HL7_NULL action batch:  Expecting 5 item lineage records and got 5

prime-router/docs/universal-pipeline/convert.md
JFU-NAVA-PBC: > I'm not sure what you mean
should the referred schema file name be "original-pipeline-transforms.yml"?
"original pipeline (`prime-router/metadata/fhir_transforms/senders/default-sender-transform.yml`). This is a good place "

JFU-NAVA-PBC:
> > smoke did not pass on local, have similar failures as in CI smoke: *** TEST FAILED*** for ignore.CSV action send: Expecting 5 item lineage records but got 0 Test passed: for ignore.HL7 action receive: Expecting 5 item lineage records and got 5 Test passed: for ignore.HL7 action batch: Expecting 5 item lineage records and got 5 *** TEST FAILED*** for ignore.HL7 action send: Expecting 5 item lineage records but got 0 Test passed: for ignore.HL7_BATCH action receive: Expecting 5 item lineage records and got 5 Test passed: for ignore.HL7_BATCH action batch: Expecting 5 item lineage records and got 5 *** TEST FAILED*** for ignore.HL7_BATCH action send: Expecting 5 item lineage records but got 0 Test passed: for ignore.HL7_NULL action receive: Expecting 5 item lineage records and got 5 Test passed: for ignore.HL7_NULL action batch: Expecting 5 item lineage records and got 5
> 
> Interesting. Everything passes for me locally. I had to reload settings though. Did you try that?

when it failed for me locally, CI also failed on same errors, remember later on after some commits, the smoke tests did pass, so it should be OK now...

JFU-NAVA-PBC (approved):
LGTM, thanks for the good work, like all the green checks

victor-chaparro (approved):
LGTM

arnejduranovic:
Please link the ticket this is addressing in the description

prime-router/src/testIntegration/kotlin/datatests/TranslationTests.kt
```diff
@@ -460,7 +460,11 @@ class TranslationTests {
 
         private fun runSenderTransform(bundle: InputStream, schema: String): InputStream {
             val fhirBundle = FhirTranscoder.decode(bundle.bufferedReader().readText())
-            val transformedBundle = FhirTransformer(schema).transform(fhirBundle)
+            val transformedBundle = if (schema != "") {
```
arnejduranovic, line 463: usually something like .isNullOrEmpty is preferred here

prime-router/src/testIntegration/resources/settings/organizations.yml
```diff
@@ -172,7 +171,7 @@
       organizationName: "elims"
       format: "HL7"
       customerStatus: "testing"
-      schemaName: "metadata/fhir_transforms/senders/default-sender-transform"
+      schemaName: "metadata/fhir_transforms/senders/original-pipeline-transforms"
```
arnejduranovic, line 174: I'm not sure we want elims to be using the original pipeline transforms? I thought the point was to remove the dependency of ELIMS on the default sender transform? With the changes as is, all you're doing essentially is just removing the contents of elims-sender-transform from the transform that ultimately gets applied to their messages?

arnejduranovic (changes requested):
Left some comments. Maybe I am misunderstanding something, can you link the ticket this is fixing so I can double check my understanding?

prime-router/settings/organizations.yml
```diff
@@ -104,25 +104,24 @@
       organizationName: ignore
       topic: full-elr
       customerStatus: active
-      schemaName: metadata/fhir_transforms/senders/default-sender-transform
+      schemaName: metadata/fhir_transforms/senders/original-pipeline-transforms
```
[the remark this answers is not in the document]
jimmyfagan: These changes can't/shouldn't be made in staging until deployment, but after deployment, if anybody is referencing `default-sender-transform`, then things will break until we can update settings. It can happen more seamlessly if we keep a placeholder for `default-sender-transform` to prevent breakage, deploy, update the settings, and then go back and delete the `default-sender-transform` for the next deployment. That might be overkill, but you should probably check if we have any orgs referencing the default, and then we can decide how to proceed.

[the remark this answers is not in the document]
JessicaWNava: No, they have not been made in staging. They will need to be made there right before merging.

prime-router/src/testIntegration/resources/settings/organizations.yml
JessicaWNava: Odd it was removed at one point. Removed again.

prime-router/settings/organizations.yml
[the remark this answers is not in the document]
arnejduranovic: `you should probably check if we have any orgs referencing the default, and then we can decide how to proceed.`

I like this idea. Can we compile this list for both staging and production environments and put it in the ticket description? Also, can we write a plan in the ticket description of how we are going to proceed once this PR is merged. Once it is merged, we can update staging and run smokeTests, right?. Once we deploy, we can update production.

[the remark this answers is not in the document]
JessicaWNava: Compiled and added to the ticket. Jimmy actually put a plan in the comments in the ticket already, so I think we are good to go there as well.

Issue #12238, opened 2023-11-14
Improve the logging so when a sender/receiver transform fails the specific element is logged


mkalish:
### User Story
As a report stream developer, I would like to be able to diagnose what schema element prevent a receiver or sender transform from being successful.

### Description/Use Case
This is something that occasionally happen in PROD and it is extremely time consuming to track down what element is the issue as the log message contains on info.

We want to log at least the element and schema

### Acceptance Criteria
- [ ] Any exception generated when processing a schema element is caught and logged with details on the schema element that failed and then re-thrown

Andrey-Glazkv:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/12238?planning-poker) with Zenhub @arnejduranovic @jalbinson @JessicaWNava @jimmyfagan @thetaurean

Andrey-Glazkv:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/12238?planning-poker) with Zenhub @JFU-NAVA-PBC

PR #12325, opened 2023-11-20
12238: capture schema/element when an error is thrown during a transform

mkalish:
This PR updates the HL7->FHIR and FHIR->FHIR transform to capture any exceptions thrown while processing a schema element, wrapping it an ConfigSchemaProcessingException and re-throws.

Test Steps:
1. Running the smoke tests should be enough

## Changes
- Creates a new `ConfigSchemaProcessingException`
- Catches an exception generated while processing a FHIR->FHIR or FHIR->HL7 element and re-throws

### Testing
- [x] Tested locally?
- [x] Added tests?

### Process

## Linked Issues
- #12238


prime-router/src/main/kotlin/fhirengine/translation/hl7/schema/ConfigSchema.kt
```diff
@@ -161,6 +161,10 @@ abstract class ConfigSchemaElement(
 ) {
     private var validationErrors: MutableSet<String> = mutableSetOf()
 
+    override fun toString(): String {
```
mkalish, line 164: I just moved this up to the abstract class so both implementations will work the same way.

prime-router/src/test/kotlin/fhirengine/translation/hl7/FhirTransformerTests.kt
```diff
@@ -248,6 +250,34 @@ class FhirTransformerTests {
...
+                ".value.coding.version"
+        )
+
+        val schema = FhirTransformSchema(elements = mutableListOf(elemA))
+        schema.name = "ErrorSchema"
+
+        val ex = assertFailsWith<ConfigSchemaElementProcessingException> {
+            FhirTransformer(schema).transform(bundle)
+        }
+        assertThat(ex.message).isEqualTo(
```
mkalish, line 275: This is how the exception that occurred in PROD the other day would now be logged

prime-router/src/main/kotlin/fhirengine/translation/hl7/FhirToHl7Converter.kt
```diff
@@ -116,7 +117,11 @@ class FhirToHl7Converter(
         val schemaContext = CustomContext.addConstants(schema.constants, context)
 
         schema.elements.forEach { element ->
-            processElement(element, bundle, schemaResource, schemaContext, debug)
+            try {
+                processElement(element, bundle, schemaResource, schemaContext, debug)
+            } catch (ex: Exception) {
+                throw ConfigSchemaElementProcessingException(schema, element, ex)
```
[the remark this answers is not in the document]
jalbinson: Is there a more specific type we could use other than `Exception`? I'm guessing no but worth an ask.

prime-router/src/main/kotlin/fhirengine/translation/hl7/schema/ConfigSchema.kt
```diff
@@ -161,6 +161,10 @@ abstract class ConfigSchemaElement(
...
+        return "${if (schema != null) "$schema-" else ""}$name"
```
jalbinson, line 165: This could be a bit more readable by extracting out the condition from the string template.

prime-router/src/main/kotlin/fhirengine/translation/hl7/schema/ConfigSchema.kt
```diff
@@ -229,4 +233,19 @@ abstract class ConfigSchemaElement(
         if (overwritingElement.debug) this.debug = overwritingElement.debug
         if (overwritingElement.constants.isNotEmpty()) this.constants = overwritingElement.constants
     }
+}
+
+class ConfigSchemaElementProcessingException(
+    val schema: ConfigSchema<*>,
+    val element: ConfigSchemaElement,
+    private val originalException: Exception,
```
jalbinson, line 241: Why not use cause here and avoid having to override it in the body?
```kotlin
override val cause: Throwable
```

prime-router/src/main/kotlin/fhirengine/translation/hl7/schema/ConfigSchema.kt
```diff
@@ -229,4 +233,19 @@ abstract class ConfigSchemaElement(
...
+) :
+    Exception() {
```
jalbinson, line 243: Maybe `RuntimeException` or something more specific

prime-router/src/main/kotlin/fhirengine/translation/hl7/FhirToHl7Converter.kt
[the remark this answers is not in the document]
mkalish: I don't think so, especially since the goal is here to catch anything that might occur and log the element

jalbinson (approved):
Looks good to me with a couple tweaks

prime-router/src/test/kotlin/fhirengine/translation/hl7/FhirToHl7ConverterTests.kt
```diff
@@ -417,12 +419,16 @@ class FhirToHl7ConverterTests {
...
-        assertThat {
+        val missingDataEx = assertFailsWith<ConfigSchemaElementProcessingException> {
             FhirToHl7Converter(
                 "ORU_R01",
                 "src/test/resources/fhirengine/translation/hl7/schema/schema-read-test-01"
             ).convert(bundle)
-        }.isFailure()
+        }
+        assertThat(missingDataEx.message).isEqualTo(
+            "Error encountered while applying: ORU_R01/header-message-headers in ORU_R01 to FHIR bundle. \n" +
```
jimmyfagan, line 429: This is the test case I was thinking of in my other comment, the child schema in this case is `ORU_R01/header`, and the element name is `message-headers`, but it's hard to tell that from the message alone.

jimmyfagan (commented):
This is great, just had one comment for a potential improvement.

prime-router/src/main/kotlin/fhirengine/translation/hl7/schema/ConfigSchema.kt
mkalish: I think you're right... I'll just drop it and we can revisit it later.  The exception will now capture the enclosing schema.

prime-router/src/test/kotlin/fhirengine/translation/hl7/FhirToHl7ConverterTests.kt
mkalish: Agree, I think this will make debugging harder

prime-router/src/main/kotlin/fhirengine/translation/hl7/FhirToHl7Converter.kt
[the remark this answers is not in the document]
JFU-NAVA-PBC: the new more specific exception "ConfigSchemaElementProcessingException" seems fit the semantics well...

arnejduranovic (approved):

PR #12350, opened 2023-11-21
CQ and NM Mapping Inventory

brick-green:
This PR maps CQ and NM to Quantity

Test Steps:
1. Review mapping inventory:
    - Review PV2-11 for NM[Quantity-LengthOfStay] [LINK](https://docs.google.com/spreadsheets/d/1ILbJ8GmvNdEW4XmvrmEHtZxsRM1bk5hHl5WD2oDSa4Q/edit#gid=0)
    - Review SMP-12 for CQ[Quantity] [LINK](https://docs.google.com/spreadsheets/d/1RhRXhpavSHLckCYV2bJg_b0onlcjMTKLGbCb7ocsfCo/edit#gid=0)

## Linked Issues
- #11667


prime-router/metadata/fhir_mapping/hl7/datatypes/CQ/Quantity.yml
```diff
@@ -0,0 +1,32 @@
+value:
+  type: BIG_DECIMAL
+  expressionType: HL7Spec
+  valueOf: CQ.1
+
```
brick-green, line 5: 'BIG_DECIMAL' was added in august

prime-router/metadata/fhir_mapping/hl7/datatypes/CQ/Quantity.yml
```diff
@@ -0,0 +1,32 @@
...
+unit:
+  expressionType: nested
+  vars:
+    cwe_originalText: STRING, CQ.2.9
+  expressions:
+    - condition: $cwe_originalText NULL
+      type: STRING
+      expressionType: HL7Spec
+      valueOf: CQ.2.1
```
jack-h-wang, line 14: The mapping inventory says this should be CWE.2 but this seems to point to CWE.1. Was this intended?

jack-h-wang (commented):
Had a comment and a question on one thing:

prime-router/src/testIntegration/resources/datatests/mappinginventory/cq/quantity/cq_29_null.hl7
```diff
@@ -0,0 +1,4 @@
+MSH|^~\&#|||||20230501102531-0400||ORU^R01^ORU_R01|12345||2.5.1|||||USA|UNICODE UTF-8
```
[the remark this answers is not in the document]
brick-green: oh nice! Removed those from all the test files I'm adding

prime-router/metadata/fhir_mapping/hl7/datatypes/CQ/Quantity.yml
brick-green: Good catch!

prime-router/metadata/hl7_mapping/common/specimen.yml
```diff
@@ -64,6 +64,17 @@ elements:
...
+  - name: specimen-collection-amount-quantity
+    value: [ '%resource.collection.quantity.value' ]
+    hl7Spec: [ '%{hl7SpecimenFieldPath}-12-1' ]
+
+  - name: specimen-collection-amount-units
+    condition: '%context.collection.quantity.where(extension(%`rsext-cq-units`).value).exists()'
+    resource: '%resource.collection.quantity.extension(%`rsext-cq-units`).value'
+    constants:
+      cweFieldPath: '%{hl7SpecimenFieldPath}-12-2'
+    schema: ../datatypes/codeableConcept/CWE
```
jimmyfagan, line 76: Thoughts on extracting this into a `datatypes/Quantity/CQ.yml` and just passing `%resource.collection.quantity` through to that? Not sure how common this will be, or therefore how reusable that would be. Consider it optional, this works fine too.

prime-router/src/testIntegration/resources/datatests/mappinginventory/cq/quantity/cq_29_null.fhir
jimmyfagan, line 1: (nit) Would be good to run reformats on these fhir files

jimmyfagan (approved):
Left a few comments for some nits and optional adjustments, but this looks great, approved!

prime-router/metadata/fhir_mapping/hl7/datatypes/NM/Quantity-LengthOfStay.yml
```diff
@@ -0,0 +1,16 @@
+value:
+  type: STRING
+  valueOf: NM.1
+  expressionType: HL7Spec
+
+unit:
```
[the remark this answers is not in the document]
brick-green: There are other spots with this same problem. The POSITIVEINT doesn't really do anything here but I used that type as a mark for later. There's not a way to do math in the condition's currently. But since we forked the library we should be able to add that functionality. I'll make a ticket for it.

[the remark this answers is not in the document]
arnejduranovic: Can you please link the ticket in this thread when created?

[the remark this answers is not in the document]
brick-green: I was just writing the ticket for this and took a look at how we might implement and found out the functionality does in fact exist! Only string comparisons are documented in the library docs but there are also int comparisons (see`io.github.linuxforhealth.core.expression.condition.ConditionBiPredicates`). So I'm going to update the mappings here to use `GREATER_THAN`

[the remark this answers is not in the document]
brick-green: Ran into a snag here. If I set a condition with `GREATER_THAN 0` then the value/unit/system/code fields are not populated which is what we want. I added an extension to keep the original value in the event that a non-positive value is sent. The problem is `value` is a required field so if it's blank then nothing else in the datatype will be created including the extension. With that in mind, I think we should keep the mapping how it is in this PR and leave the condition to future validation work. Thoughts?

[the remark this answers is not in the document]
jimmyfagan: I agree, the mapping says we should "raise an error" in those cases, and that's not something we have any precedent for doing during the mapping, so it should be handled in validation. If we end up with negative values causing issues on receivers we can maybe handle that with receiver transforms down the line.

prime-router/metadata/hl7_mapping/common/specimen.yml
brick-green: added Quantity/CQ.yml since it will be used in two places. Good rec!

prime-router/src/testIntegration/resources/datatests/mappinginventory/cq/quantity/cq_29_null.fhir
brick-green: reformatted both fhir files

arnejduranovic (approved):

Issue #11051, opened 2023-08-24
Update How Route Function Queues Translate Function


arnejduranovic:
### User Story

As a pipeline step (route), I should follow the same pattern as other Universal Pipeline steps when calling the next step in the pipeline (translate) so that the pipeline can remain maintainable and reduce the probability of defects.

### Description/Use Case

Presently, there is an interesting unexpected behavior in the translate function, made apparent by #10928.

The route step will generate ONE report and add Endpoint Resources to the Provenance Resource, that the translate function will then iterate over to generate a new report for each receiver:

```kotlin
return receiverAndEndpoints
    .map {
        actionHistory.trackActionReceiverInfo(it.second.organizationName, it.second.name)
        it
    }
    .filter {
        it.second.topic.isUniversalPipeline
    }
    .map { (receiverEndpoint, receiver) ->
        val updatedBundle = pruneBundleForReceiver(bundle, receiverEndpoint)

        val bodyBytes = getByteArrayFromBundle(receiver, updatedBundle)
```

The interesting behavior that we want to solve for is when/if `val bodyBytes = getByteArrayFromBundle(receiver, updatedBundle)` throws an exception during translation (this can happen if the schema is "invalid" for the message that it is trying to process) everything the step did gets rolled back (see #10928), thus one "bad receiver" can prevent other receivers from receiving their data. Henceforth referred to as the "Rotten Apple Spoils the Bunch" problem.

If the Route step generated a separate report (and thus a separate call to the translation step) for each receiver it found for the report, this would not be an issue since the translate function would only ever run for one receiver. 

### Risks/Impacts/Considerations

As part of this update, we should update how we communicate where a report should be routed to. Destinations for a message should no longer be stamped to provenance resource in the route step, they should be communicated to the translate step via SQS queue params. This will work now because the translate step will receive a unique report for each destination.

### Dev Notes

One proposed solution is to update the Route step to generate a report and queue a Translate function for each receiver it identifies, INSTEAD OF using provenance and endpoint to indicate where a report should be routed to. This is how the other UP steps work, so implementing it this way would bring them in alignment and increase maintainability of the pipeline.

### Acceptance Criteria

- [ ] Rotten Apple problem remediated
- [ ] UP does not depend on provenance resource for routing/processing
- [ ] Route step documentation (and other documentation) updated to reflect new functionality

Andrey-Glazkv:
[...]

PR #12401, opened 2023-11-27
11051: route a message per receiver

mkalish:
This PR updates the route and translate steps such that the route step produces a message per receiver.

Test Steps:
1. Run the smoketests

Confirm that old messages still work:
1. Apply this patch while on `master`
```
index ae81aa9cd..7bca3acfd 100644
--- a/prime-router/src/main/kotlin/azure/QueueAccess.kt
+++ b/prime-router/src/main/kotlin/azure/QueueAccess.kt
@@ -46,7 +46,7 @@ object QueueAccess {
     fun sendMessage(
         queueName: String,
         message: String,
-        invisibleDuration: Duration = Duration.ZERO,
+        invisibleDuration: Duration = Duration.ofSeconds(30),
     ) {
         // Bug:  event.at is calculated before the call to workflowengine.recordHistory
         // In cases of very large datasets, that db write can take a very long time, pushing
```
2. Send an HL7 message through the UP that will get sent to a receiver
3. Immediately after sending the message, turn off reportstream (because of the azure functions bug, I had to go into Activity monitor and kill the java processes)
4. Wait 30 seconds and confirm that there is a message in the convert queue
5. Switch to this branch
6.  Turn report stream on and verify that the report step is delivered
7. Repeat for the route step and translate step (i.e. wait for the message to be in each queue)

## Changes
- The route step now dispatches a message per receiver
- The route step prunes the conditions before uploading the bundle

### Testing
- [ ] Tested locally?
- [ ] Added tests?

## Linked Issues
- #11051


prime-router/src/test/kotlin/fhirengine/utils/FHIRBundleHelpersTests.kt
```diff
@@ -336,6 +204,8 @@ class FHIRBundleHelpersTests {
         assertThat(bundle.getDiagnosticReportNoObservations()).isEmpty()
     }
 
+    // TODO: fix OOM exception, ticket:
+    @Ignore
```
mkalish, line 208: Still very confused to the why, but this test started causing an OOM exception after I made changes in the file that prevented the rest of the tests from passing.  I'm opting to ignore it since the following test covers the same functionality and will create a tech debt ticket to re-investigate/delete the test

prime-router/src/test/kotlin/fhirengine/engine/FhirTranslatorTests.kt
```diff
@@ -172,12 +230,21 @@ class FhirTranslatorTests {
         mockkObject(BlobAccess)
 
         // set up
-        val settings = FileSettings().loadOrganizations(colorado)
```
mkalish, line 175: Annoyingly a bunch of these tests weren't really doing anything.  None of the receivers in Colorado org match the receiver in the FHIR bundle.

prime-router/src/test/kotlin/fhirengine/utils/FHIRBundleHelpersTests.kt
```diff
@@ -336,6 +203,7 @@ class FHIRBundleHelpersTests {
         assertThat(bundle.getDiagnosticReportNoObservations()).isEmpty()
     }
 
+    // TODO: fix OOM exception, ticket:
```
jimmyfagan, line 206: What's the plan here? Wasn't sure if this test needs to be commented out or something in the meantime.

prime-router/docs/universal-pipeline/route.md
```diff
@@ -255,15 +268,15 @@ The processing mode of the data indicates the sender’s intended context for th
...
 ```
 
 ### Condition Filter
 
-Filter data based on the test identifiers. A receiver expecting flu results should only accept tests for flu. If the message contains multiple observations, some that pass the condition filter and others that do not, the condition filter will be used to identify the desired observations. Identifiers for the needed observations are added to the Endpoint which is then added to the Provenance resource.
-
-*The Translate step will review the Endpoints and remove any observations that are not identified.
+Filter data based on the test identifiers. A receiver expecting flu results should only accept tests for flu. If the
+message contains multiple observations, some that pass the condition filter and others that do not, the condition filter
+will pass if any the observations are of interest to the receiver.
```
jimmyfagan, line 279: Should we add a sentence to mention that observations not matching will be removed here? Before we mentioned that it was part of the translate step, so even though it's in routing now, we might still want to mention it.

jimmyfagan (approved):
Only very minor comments, but overall this looks good and is a great enhancement in my opinion. Good job cleaning up the tests and everything and marking the code that can be deleted later. I assume you'll add story number(s) to all the `TODO` comments before delivering, so I'm good to mark this approved. tests and integrationTests passed, running smokeTests now.

prime-router/src/test/kotlin/fhirengine/utils/FHIRBundleHelpersTests.kt
jimmyfagan: Looking again, I guess this didn't need to be skipped after increasing the heap size? But we still want a ticket to look for unnecessary memory usage?

jimmyfagan:
In your description you say:

> Confirm that old messages still work:
TBD

Any updates on that testing plan?

mkalish:
@jimmyfagan Completely blanked on the test plan for old messages,  Ill pull that together tonight for you to test tomorrow

prime-router/src/test/kotlin/fhirengine/utils/FHIRBundleHelpersTests.kt
mkalish: Yep... I'll still add a ticket

mkalish:
@jimmyfagan FYI added steps for confirming old messages continue to work

prime-router/docs/universal-pipeline/route.md
mkalish: Yes, I'll update this

jimmyfagan:
Followed steps for testing old messages in the queues and everything seems good (just waiting on a minor doc update still), this should be ready for @arnejduranovic at this point!

arnejduranovic (approved):

Issue #12034, opened 2023-11-01
Connect HCI to the UP as a Sender on Staging


brandonnava:
### User Story

As the HCI organization, I need the ability to send HL7v2.5.1 NIST ELR (ORU-R01) messages through RS to STLTs on ReportStream's staging environment universal pipeline. For the pilot effort, this will be NY.

### Description/Use Case

We need to configure RS:
1. Create a sender under the `hcintegrations` org in staging that can send HL7v2.5.1 NIST ELR messages
2. This sender should be configured to use the `full-elr` topic.
3. HCI will use RS's oauth, so contact Arnej for their public key and associate it with their org

### Acceptance Criteria

- [x] Added sample sender to org file. See https://github.com/CDCgov/prime-reportstream/pull/10426/files#diff-a3595b762b115fdaacb32ec9e8930def1a645002ecc8cc1ba9102d2ec5512fd8 to see how this was done for CDC ELIMS.
- [x] Updated existing receivers on UP to ignore messages coming from HCI
- [x] HCI Public key added to org 
- [x] Verify submissions page works for all HCI senders

arnejduranovic:
Before we can do this story, we need HCI to create a public/private key pair and send us the public key.

Andrey-Glazkv:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/12034?planning-poker) with Zenhub @arnejduranovic @brick-green @jack-h-wang @jalbinson @JessicaWNava @jimmyfagan @mkalish @thetaurean

Andrey-Glazkv:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/12034?planning-poker) with Zenhub @JFU-NAVA-PBC

jimmyfagan:
I misspoke in Standup today, the hci sender transform wasn't merged on Wednesday, so it wasn't in Thursday's deployment. It was merged today, but progress here will be slow until the next deployment which is next Tuesday.

PR #12422, opened 2023-11-28
Add sender transform for hci

jimmyfagan:
This PR adds a sender transform for HCI to allow for filters to be configured in staging to not include HCI messages. This conflicts with [this PR](https://github.com/CDCgov/prime-reportstream/pull/12277), so depending on which goes in first, this might require some minor changes.

Test Steps:
1. Run something like `./prime fhirdata -r metadata/fhir_transforms/senders/HCI/hci-sender-transform.yml --input-file junk/input.hl7 --output-file junk/output.fhir --output-format FHIR` with sample data and search the resulting fhir file for `hcintegrations`.

## Changes
- Added a hci sender transform to add a sender identifier to the fhir bundle in the UP (also added an element to the default sender transform to provide the baseline for this kind of transform and give consistency to how this can be utilized for future senders)
- Added senders for hci in the test _and_ local `organizations.yml` files
- Edited the existing translation test for hci to include the test sender

### Testing
- [x] Tested locally?
- [x] Added tests?

This is related to #12034, and is needed to be able to prevent existing UP senders from receiving HCI messages in staging.

## Settings Rollout Plan

Once this deploys to staging (and prod), I'll update the `hcintegrations.default-elr` sender to use the newly added `hci-sender-transform.yml` as a sender transform. No existing settings should be affected.


brick-green (approved):
looks good

Issue #12287, opened 2023-11-17
Handle HL7v2.5.1 messages without logging any errors


jimmyfagan:
### User Story
As a ReportStream maintainer, I want error messages in the application output to be reflective of real errors, and not produce error output for valid and expected inputs.

### Description/Use Case
When we read incoming HL7v2 messages, we first try to read them as v2.7. If that fails, we will then fall back to read them as v2.5.1. This works fine functionally, however currently, the failure to read messages as v2.7 results in error messages in the application output. For example, OBX.2 is allowed to be `CE` in v2.5.1 but not in v2.7, so a message with OBX.2 of `CE` results in:
```
11:15:26.681 [pool-2-thread-87][263] ERROR gov.cdc.prime.router.fhirengine.utils.HL7Reader - Failed to parse message
.............
Caused by: ca.uhn.hl7v2.HL7Exception: 'CE' in record 1 is invalid for version 2.7 at OBX-2(0)
```
We should only log message parsing errors in cases where the message can be read as neither v2.7 nor v2.5.1.

### Risks/Impacts/Considerations
No known risks, though the risk of not fixing this is that the error messages will cause false alarms among developers.

### Acceptance Criteria
- [ ] HL7v2.5.1 messages which are not valid HL7v2.7 (because of OBX.2 of `CE` for example) can be sent through the Universal Pipeline without any error messages appearing

arnejduranovic:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/12287?planning-poker) with Zenhub @brick-green @jack-h-wang @jalbinson @JFU-NAVA-PBC @mkalish @thetaurean @jimmyfagan @JessicaWNava

PR #12434, opened 2023-11-29
Updated the logging to only log an error if all message types fail to…

JessicaWNava:
… parse the message.

This PR updates the logging to only log an error if all message types fail to parse the message

Test Steps:
1. Ran a message through the pipeline before changes to verify that it produced the error. 
2. Ran a message through the pipeline after changes and verified that it now produced a warning, but that the message also still contained a warning. 
3. Ran the HL7 Reader Tests

## Changes
- Updates the logging to only log an error if all message types fail to parse the message

### Testing
- [ X] Tested locally?
- [ ] Added tests?

#issue #12287


JessicaWNava:
> Left one minor optional comment, but this looks good! Testing looked good, though I did notice the same warning showed up during the receive and the convert steps. Wondering if we need to log it in both places. I'll leave it to the code owner review to weigh in on that, and either way, it might be outside of the scope of this ticket, so I'll mark this approved.

Did you see a warning or an error? We are wanting a warning still.

arnejduranovic (approved):

PR #12441, opened 2023-11-29
11443: split batch function into universal and covid

### LINKED ISSUE #11443: Split the batch function such that the UP and legacy pipeline no longer share the same Azure function
### User Story
As a report stream developer, I would like a dedicated function for handling batching

### Description/Use Case
The batch function is "shared" between the legacy and UP pipeline, but this is mostly superficial as the code quickly bifurcates based on which pipeline the topic is configured with.  Splitting up these functions would further clarify the use of topic and allow the team to make changes to the UP without considering impacts to the legacy.

Existing batching is impacted while doing the split

### Dev Notes
- Most of the logic for the UP is already isolated in it's own [function](https://github.com/CDCgov/prime-reportstream/blob/463b9f0dc13139919fdfac2f52a058d2f00045c2/prime-router/src/main/kotlin/azure/BatchFunction.kt#L185)
- A new queue should get added to support the batch step, current understanding is that this will not require infra support but that should be confirmed

### Acceptance Criteria
- [ ] A new azure function and queue exist that is dedicated to batching data from the UP

jalbinson:
This PR splits the batch function into seperate Covid and Universal pipeline functions. It also seperates out the batch queue into seperate Covid and Universal pipeline queues.

It intentionally leaves existing functionality so we may churn through the existing queue once this is deployed. The deprecated batch queue and function may be safely removed after the next deploy.

Test Steps:
1. ./gradlew test
2. ./gradlew testIntegration

## Changes
- Moved batching code into a new package
- Split batch function into 2 new functions and a deprecated existing function
- Update queuing code to accept a new `queueName` parameter decoupled from the `Event` code.
- Updated tests to handle new classes

### Testing
- [ ] Tested locally?
- [ ] Added tests?

## Linked Issues
- Fixes #11443

## To Be Done
- #12462


prime-router/src/test/kotlin/azure/batch/UniversalBatchFunctionTests.kt
mkalish, line 1: Should we add some tests to cover the covid batch functino?

prime-router/src/main/kotlin/azure/batch/CovidBatchFunction.kt
mkalish, line 1: Nit: for the follow up maybe consider doing a git mv of `DeprecatedBatchFunction` to this one?  Mostly thinking of keeping the git history

mkalish (commented):
Few small nits, but looks great!

I'll pull this down and test tomorrow morning.

mkalish (changes requested):
Approval was preemptive.

Looks like at least one of the names for the new queues are invalid.

```
[2023-11-30T14:19:00.047Z] com.azure.storage.queue.models.QueueStorageException: Status code 400, "<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
[2023-11-30T14:19:00.047Z] <Error>
[2023-11-30T14:19:00.047Z]   <Code>InvalidResourceName</Code>
[2023-11-30T14:19:00.047Z]   <Message>The specifed resource name contains invalid characters.
[2023-11-30T14:19:00.047Z] RequestId:9e442850-e4f4-4a41-878a-9b08916cf72b
[2023-11-30T14:19:00.047Z] Time:2023-11-30T14:19:00.045Z</Message>
[2023-11-30T14:19:00.047Z] </Error>"
```

prime-router/src/test/kotlin/azure/batch/UniversalBatchFunctionTests.kt
jalbinson: can't hurt. I didn't add them in on first pass because there weren't any to begin with.

prime-router/src/main/kotlin/azure/batch/CovidBatchFunction.kt
jalbinson: I just tried to fix this but git is smartly considering the file with the least amount of changes the "moved" one and the other 2 files as the new ones.

PR #12500, opened 2023-12-04
Implement code to condition mapping and remerge CLI update tool

### LINKED ISSUE #11510: Build utility to auto-update Code to Condition table mappings
As a ReportStream engineer, I want to automatically update the CSTE valuesets on the code-to condition table using a CLI command.

**Background Info**
Since the bulk of the mapping table will be made up of values taken from CSTE codesets that are stored in the Value Set Authority Center (VSAC). We can make use of the VSAC's [FHIR Terminology Service](https://www.nlm.nih.gov/vsac/support/usingvsac/vsacfhirapi.html) to query and update the table. Using the "Member OID" and "Version" columns from the table we can create a utility to query each Member OID for which the "Value Source" column is "RCTC" to see if we have the latest version and if not to update to the latest version. This should be a CLI utility similar to update-livd-table command that can be run on-demand to update the table.

The FHIR Terminology Service for VSAC Resources is a RESTful API service for accessing the VSAC value sets and supported code systems.

Example parameters: </br>
environment - name of table to compare input to </br>
silent - do not generate diff or ask for confirmation
activate - activate the table upon creation
key - apikey to access VSAC FHIR terminology service
OID - OID of valueset to update (default to all OID in table if blank)

Example CLI command for update tool:

./prime condition-mapping-update --activate --silent --environment staging --key <apikey>

**Resources**
[example observation mapping table.xlsx](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNEJrQVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--2dbfd029a8f9b467a3832c1cdd130d4f4ac2db67/example%20observation%20mapping%20table.xlsx)

**additional info**
Column "Value Source" will indicate the source of the data in that row. "RCTC" indicates that the source of the data is the "Reportable Conditions Trigger Codes" table which is made up of CSTE valuesets in the VSAC. These valuesets can be found by searching the VSAC at https://vsac.nlm.nih.gov/valueset/expansions?pr=all and sorting the "steward' column by "CSTE Steward".

The following CLI command can be used to load the example table into your local database. Be sure to save the example table as a .CSV file before running the command.

./prime lookuptables create -i "input-file-location" -n "input-table-name" -a

**Acceptance Criteria**
-[ ] CLI command updates observation mapping table with values from relevant valuesets
-[ ] Valuesets not labeled RCTC are not affected by update process
-[ ] CLI can update one or more valuesets based on `--oids` parameter (or all `OID`s in table by default)
-[ ] Updated section `Update prod/staging tables using RESTful sources` in `docs/getting-started/standard-operating-procedures/observation-mapping-table.md`

### LINKED ISSUE #12275: Implement code to condition mapping
### User Story

As a sender of test results containing various conditions,
I want ReportStream to be able to route test results based on condition to receivers,
so I can meet by reporting requirements

### Description/Use Case

In order to be able to route messages based on condition, we first need to be able to stamp various observations in a message with the condition(s) they represent. The code to condition mapping [design](https://github.com/CDCgov/prime-reportstream/blob/5097e07acd10747c8cd2104c45bb986c29d623db/prime-router/docs/design/proposals/0023-condition-to-code-mapping/0023-condition-to-code-mapping.md#mapping-loincsnomed-codes-in-received-messagebundle-to-condition-code) discusses how to map test codes to condition codes in the internal FHIR resource. In short, this ticket should implement the code to condition stamping that is outlined in the design section titled `Mapping LOINC/SNOMED codes in received message/bundle to condition code`:
1. Update internal FHIR bundle observation resources with the appropriate condition code using a custom RS FHIR extension. If an observation is not a condition code, stamp it as "unknown" or "aoe"
    - Update action log when there is a failure to map an observation

This ticket shall implement the following sections of the design:
    - `Mapping LOINC/SNOMED codes in received message/bundle to condition code`
    - `Action Log Event`

All sections not in the list above are out of scope, including but not limited to:
    - `Filtering for condition`

Note, the design suggests where and how to implement code to condition mapping. This should be used as a guide and NOT as a hard requirement. We should implement code to condition mapping in a way that is most maintainable. This may mean refactoring current functionality.

### Acceptance Criteria

- [ ] ReportStream has the ability to stamp observations in the internal FHIR bundle with the condition it maps to
- [ ] ReportStream logs mapping failures
- [ ] Design documentation updated (if needed)

thetaurean:
This PR implements code to condition mapping and re-merges previously merged PR (#12327) that was mistakenly reverted as part of #12618.

Test Steps:
1. Run unit tests and integration tests
2. Submit a report and verify condition code extensions are added
   1. On `organizations.yml:114`, modify sender `development.dev-elims` to have a  `format` of `FHIR`
   2. Starting with a clean database and reloaded settings/tables, submit a report using the updated sender 
        ```zsh
        curl --location 'http://localhost:7071/api/reports' --header 'client: development.dev-elims' --header 'Content-Type: application/fhir+ndjson' --data-binary @src/test/resources/fhirengine/engine/routing/valid.fhir
        ```
   3. Open the azure storage explorer and browse to `reports/route/development.dev-elims`
   4. Download and view the most recent (and only) blob
   5. Verify there are new `condition-code` extensions on the fhir bundle. The full url is: 
        ```html
        https://reportstream.cdc.gov/fhir/StructureDefinition/condition-code
        ```

## Changes
- Implemented Observation extension function for adding mapped conditions
- Modified FHIRConverter to add mapped conditions to all observations
- Revive changes from #12327 

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- Fixes #12275 
- Fixes #11510


prime-router/src/test/resources/fhirengine/engine/bundle_condition_observation.fhir
mkalish, line 1: Nit: can we run a JSON formatter against this?

prime-router/src/main/kotlin/fhirengine/engine/FHIRConverter.kt
```diff
@@ -90,6 +93,12 @@ class FHIRConverter(
                 // conduct FHIR Transform
                 transformer?.transform(bundle)
 
+                bundle.entry.filter { it.resource is Observation }.forEach {
```
mkalish, line 96: Nit: I think think this can be
```suggestion
                bundle.entry.filterType<Observation>.forEach {
```

and then you can drop the cast

prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
```diff
@@ -27,6 +32,50 @@ import kotlin.reflect.jvm.isAccessible
  * A collection of helper functions that modify an existing FHIR bundle.
  */
 const val conditionExtensionurl = "https://reportstream.cdc.gov/fhir/StructureDefinition/reportable-condition"
+const val conditionCodeExtensionURL = "https://reportstream.cdc.gov/fhir/StructureDefinition/condition-code"
+
+private fun lookupCondition(code: Coding, metadata: Metadata): Coding? {
+    val mappingTable = metadata.findLookupTable("observation-mapping").also {
+        if (it == null) {
+            return null // TODO: better error if table is missing? log vs throw?
```
mkalish, line 40: I agree that we should just `throw` here

prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
```diff
@@ -27,6 +32,50 @@ import kotlin.reflect.jvm.isAccessible
...
+        }
+    }!!
+    val condition = mappingTable.dataRowsMap.find {
+        it[ObservationMappingConstants.TEST_CODE_KEY.lowercase()] == code.code // TODO: why lower
+    }
+    return if (condition.isNullOrEmpty()) {
+        null
+    } else {
+        Coding(
+            condition[ObservationMappingConstants.CONDITION_CODE_SYSTEM_KEY.lowercase()], // TODO: why lower
```
mkalish, line 50: Couple todos here that need to get cleaned...

Looks like the lookup table [code](https://github.com/CDCgov/prime-reportstream/blob/b3c5dcaba04e29395941c1b1e6c6596b091af988/prime-router/src/main/kotlin/metadata/LookupTable.kt#L149) is doing the lower casing

prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
```diff
@@ -27,6 +32,50 @@ import kotlin.reflect.jvm.isAccessible
...
+fun Observation.addMappedCondition(metadata: Metadata): List<ActionLogDetail> {
+    val codeSourcesMap = mapOf(
+        ObservationMappingConstants.BUNDLE_CODE_IDENTIFIER to this.code.coding,
+        ObservationMappingConstants.BUNDLE_CODEABLE_IDENTIFIER to this.valueCodeableConcept.coding
+    ).filterValues { it.isNotEmpty() }
+    if (codeSourcesMap.values.flatten().isEmpty()) return listOf(UnmappableConditionMessage())
+
+    return codeSourcesMap.mapNotNull { codeSourceEntry ->
+        codeSourceEntry.value.mapNotNull { code ->
+            lookupCondition(code, metadata).let { conditionCode ->
```
mkalish, line 66: I don't think you need the `let` here?

prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
```diff
@@ -27,6 +32,50 @@ import kotlin.reflect.jvm.isAccessible
...
+                if (conditionCode == null) {
+                    code.code
+                } else {
+                    code.addExtension(conditionCodeExtensionURL, conditionCode)
+                    null
+                }
+            }
+        }.let {
+            if (it.isEmpty()) null else UnmappableConditionMessage(it, codeSourceEntry.key)
```
mkalish, line 75: Is this logic potentially not granular enough?  If there was FHIR like

```json
"coding" : [
  {
    "code":  123 // Can be mapped
  },
  {
    "code"  456 // Cannot be mapped
  }
]
``` 
We would not return the error log?  Maybe I'm misreading this.

Either way it would be nice for the error message to also indicate which `code` was not ampped

prime-router/src/test/kotlin/fhirengine/azure/FhirFunctionTests.kt
```diff
@@ -736,5 +745,147 @@ class FhirFunctionTests {
...
+
+            val actionHistory = spyk(ActionHistory(TaskAction.receive))
+            val workflowEngine =
+                makeWorkflowEngine(
+                    metadata,
+                    settings,
+                    ReportStreamTestDatabaseContainer.testDatabaseAccess
+                )
+
+            val queueMessage = "{\"type\":\"raw\",\"reportId\":\"${report.id}\"," +
```
mkalish, line 863: Can we switch this out for the fhirConvert message type?

mkalish (changes requested):
Looking really good!  A couple of small comments, but main feedback is I think the unmapped condition logic might not be handling the case where the coding array has one code that can be mapped and another that can't.

Also, looks like the smoke tests might be broken.

prime-router/src/main/kotlin/fhirengine/engine/FHIRConverter.kt
thetaurean: Excellent suggestion, will do! I was really struggling to find that method call lol

prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
thetaurean: Yeah I suppose I was wondering why that is. It's causing inconsistencies that are wreaking havoc on my OCD. I kind of want to unilaterally use lower case for all column headers now. Or add a method/param to lookup table that doesn't lower case. 

Thoughts?

prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
thetaurean: If the lookup results in null we want to pass on the code.
Otherwise we want to add the extension and pass on a null that will get filtered.

Is there another scope function I should be using instead?

prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
thetaurean: Should behave as you expect/describ:

We would return the error log for the unmapped code in the 'chimera' FHIR, to include the code in the message and the origin field in the fieldMapping value

prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
mkalish: Got it...is there maybe a missing test then?  I dropped the let and I think they all passed

thetaurean:
Test suites are good now. Should be good to go 👍

thetaurean:
Test suites are happy now. Should be good to go 👍

thetaurean:
Updated PR description to account for remerging #12327

prime-router/src/main/kotlin/fhirengine/engine/FHIRConverter.kt
```diff
@@ -90,6 +93,15 @@ class FHIRConverter(
                 // conduct FHIR Transform
                 transformer?.transform(bundle)
 
+                // 'stamp' observations with their condition code
+                bundle.entry.map { it.resource }.filterIsInstance<Observation>().forEach {
+                    it.addMappedCondition(metadata).run {
+                        actionLogger.getItemLogger(bundleIndex + 1, it.id)
+                            .setReportId(message.reportId)
+                            .log(this, ActionLogLevel.warning)
```
mkalish, line 101: Nit: can this be switched to just using `warning` and then making `log` private again?

prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
```diff
@@ -25,6 +31,86 @@ import java.util.stream.Stream
...
+/**
+ * Retrieves loinc/snomed codes from [this] observation in known locations (code.coding and valueCodeableConcept.coding)
+ * @return a map of lists of codings keyed by their origin as a printable string
+ */
+fun Observation.getCodeSourcesMap(): Map<String, List<Coding>> {
+    val toReturn = mutableMapOf<String, List<Coding>>()
+    try {
+        toReturn[ObservationMappingConstants.BUNDLE_CODE_IDENTIFIER] = this.code.coding
+    } catch (error: FHIRException) {
+        if (error.message == null ||
```
mkalish, line 71: I don't think this check should be necessary? Looks like CodeableConcept is the only allowed value in the FHIR spec

prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
```diff
@@ -25,6 +31,86 @@ import java.util.stream.Stream
...
+            !error.message!!.startsWith("Type mismatch: the type CodeableConcept was expected")
+            ) {
+            throw error
+        }
+    }
+    try {
+        toReturn[ObservationMappingConstants.BUNDLE_CODEABLE_IDENTIFIER] = this.valueCodeableConcept.coding
```
mkalish, line 78: Nit: looking at this again I think this might be a better name (lines up with the actual FHIR property)
```suggestion
        toReturn[ObservationMappingConstants.BUNDLE_VALUE_IDENTIFIER] = this.valueCodeableConcept.coding
```

prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
```diff
@@ -25,6 +31,86 @@ import java.util.stream.Stream
...
+}
+
+/**
+ * For every snomed/loinc code in code or valueCodeableConcept, lookup a condition code and add it as an extension
+ * @param metadata metadata containing an observation-mapping lookup table
+ * @return a list of ActionLogDetail objects with information on any mapping failures
+ */
+fun Observation.addMappedCondition(metadata: Metadata): List<ActionLogDetail> {
+    val codeSourcesMap = this.getCodeSourcesMap().filterValues { it.isNotEmpty() }
+    if (codeSourcesMap.values.flatten().isEmpty()) return listOf(UnmappableConditionMessage()) // no codes found
```
mkalish, line 96: Nit: maybe a different class here since it's a little confusing since there isn't anything to map right here?

prime-router/src/test/kotlin/fhirengine/engine/FhirConverterTests.kt
```diff
@@ -340,4 +361,110 @@ class FhirConverterTests {
...
+            )
+        }
+
+        // assert
+        verify(exactly = 1) {
+            engine.getContentFromFHIR(any(), any())
+            actionHistory.trackExistingInputReport(any())
+            transformer.transform(any())
+            actionHistory.trackCreatedReport(any(), any(), blobInfo = any())
+            BlobAccess.Companion.uploadBlob(any(), FhirTranscoder.encode(bundle).toByteArray(), any())
```
mkalish, line 467: ➕

prime-router/src/test/kotlin/fhirengine/engine/FhirConverterTests.kt
```diff
@@ -340,4 +361,110 @@ class FhirConverterTests {
             actionHistory.trackCreatedReport(any(), any(), blobInfo = any())
         }
     }
+
+    @Test
+    fun `test condition code stamping`() {
```
mkalish, line 366: Nit: maybe one more test here or as a DB test that checks that the case when one condition can be mapped, but another can't be?

prime-router/src/main/kotlin/metadata/LookupTable.kt
```diff
@@ -152,6 +152,20 @@ open class LookupTable : Logging {
         }
     }
 
+    /**
+     * Returns the table as a sequence of map, where each key in the map is the column name. If you have multiple
+     * columns with the same name in a table, the last one wins. So don't do that.
```
mkalish, line 157: Nit: you could throw if `colName` is already in `rowMap` if you wanted

prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
```diff
@@ -25,6 +31,86 @@ import java.util.stream.Stream
...
+
+    return codeSourcesMap.mapNotNull { codeSourceEntry ->
```
mkalish, line 98: Nit: no need to change, but you could avoid the `null` return if you wanted by using `reduce` over an array rather than `map`

prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
```diff
@@ -25,6 +31,86 @@ import java.util.stream.Stream
...
 const val conditionExtensionurl = "https://reportstream.cdc.gov/fhir/StructureDefinition/reportable-condition"
+const val conditionCodeExtensionURL = "https://reportstream.cdc.gov/fhir/StructureDefinition/condition-code"
+
+/**
+ * Looks up a condition code for the passed in [code] (typically a test) using the given [metadata] object
+ * @param code test (or other type) code to look up a condition for
+ * @param metadata metadata containing an observation-mapping lookup table
+ * @return the condition code or null if no code was found
+ */
+private fun lookupCondition(code: Coding, metadata: Metadata): Coding? {
```
mkalish, line 42: No need to change it for now, but what do you think of a tech debt ticket to extract this out to more of a service?  Wondering if will have need to do this kind of work outside of processing a FHIR bundle.

mkalish (commented):
Great stuff!  Small comments, but I'm good to approve once I have a chance to test it locally.

On that topic, could you add instructions for how this could be verified manually? I'm thinking you could just send a message to a FHIR receiver

prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
thetaurean: When running smoke tests there's an instance where it's a String; I'll take another look and see if there's a different way to fix it perhaps?

prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
thetaurean: I'm not against it but I liked the elegance of using one class. In either case (no codes or codes without lookup values) the observation has an unmappable condition (reasoning on the name).

It also didn't make the query easier to distinguish between the two but maybe I was missing something there.

Your choice on this one either way.

prime-router/src/main/kotlin/metadata/LookupTable.kt
thetaurean: Ooh I didn't actually right this comment or the corresponding method I'm not sure why it came up that way in the diff.

That's a great point regardless though. I'll make the change and run tests.

prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
mkalish: No strong feelings from me!

prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
mkalish: Hmm.. that is strange.  I'm fine leaving it and opening a tech debt ticket to track it down

prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
thetaurean: After some more thought I think your suggestion is more scalable & testable so I went with that :).

prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
thetaurean: Ticket opened

prime-router/src/test/kotlin/fhirengine/engine/FhirConverterTests.kt
thetaurean: Added another test case to verify

prime-router/src/main/kotlin/metadata/LookupTable.kt
thetaurean: The behavior comes from an abstraction provided by a dependency and thus is tough to change.

prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
thetaurean: Tech debt ticket drafted

mkalish (approved):
:shipit: Looks great!  Successfully saw the FHIR bundle with the condition extension added

PR #12514, opened 2023-12-05
12427: refactor delete resource to avoid OOM

### LINKED ISSUE #12427: Investigate why Bundle.deleteResource uses so much memory and update the implementation
### User Story
As a report stream developer, I would like to avoid memory intensive bundle operations. 

### Description/Use Case
As part of updating the routing step, it was discovered that Bundle.deleteResources uses and holds onto a lot of memory such that it was causing OOM errors in the CI runner.  This was mitigated by upping the the heap size, but the method is self should be refactored.

### Dev Notes
There are likely a few low hanging changing items to look at:

- The recursive function [allocates memory](https://github.com/CDCgov/prime-reportstream/blob/3c5e4a3b224707549100a3c2a9f736b69fca5efa/prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt#L104) for the same data at each recursive step
- Reduce the amount of recursive calls
- This was only produced during the CI build

### Acceptance Criteria
- [ ] The settings to increase the heap size for the tests is removed

jalbinson:
This PR optimizes the `Bundle.deleteResource` function.

What happened during development:
- `Bundle.deleteResource` optimized and made much less resource intensive
- A different reason for the OOM error was discovered
  - The test to blame: `legacy- test full elr translation happy path, receiver with condition filter so extensions`
- This test was ignored during development of this PR
- This test was removed by [another PR](https://github.com/CDCgov/prime-reportstream/pull/12539/files#diff-12ddff7984dc8097c7fd0d865c501e8df72d00c62e9290c37ece36cb180fda5cL417)

Test Steps:
1. ./gradlew test

## Changes
- Extract out recursive call into internal function and pull the pieces out that do not need to be run every recursive call
- refactor reference deleting code to not use reflection
- clean out references only at the end when necessary
- clean up diagnostic report deletion code to not depend on FHIR path

### Testing
- [x] Tested locally?
- [ ] Added tests?

## Linked Issues
- Fixes #12427


jalbinson:
When profiling the test `Test Removing observation from diagnostic report with single observation`:

Before:
690 ms
298 MB

After:
18 ms
3.14 MB

prime-router/build.gradle.kts
```diff
@@ -168,8 +168,6 @@ tasks.test {
 
     // Set max parellel forks as recommended in https://docs.gradle.org/current/userguide/performance.html
     maxParallelForks = (Runtime.getRuntime().availableProcessors() / 2).takeIf { it > 0 } ?: 1
-    minHeapSize = "2g"
-    maxHeapSize = "3g"
```
JessicaWNava, line 172: Is this still necessary to remove after your changes?

JessicaWNava (approved):
One question, otherwise looks good.

arnejduranovic (approved):

Issue #12224, opened 2023-11-13
Design "original message Passthrough" Universal Pipeline Feature


arnejduranovic:
### User Story

As a Sender of HL7v2 to ReportStream who has my own data integrity requirements, I want ReportStream to not modify the message I send in any way so that the receiver can receive my message byte-for-byte the same way I sent it.

### Description/Use Case

In order to accommodate the use case above while retaining flexibility and the advantages of the Universal Pipeline, ReportStream should implement an "original message Passthrough" feature that is to be enabled via Sender settings. This setting can be called something like "send_original". If this setting is enabled, ReportStream will embed the original HL7v2 into the internal FHIR object during the translation in the `convert` step of the pipeline. Then, at the `translate` step, instead of translating the FHIR to HL7v2, the translate step will simply pull out the embedded HL7v2 from the FHIR object and send that along. Note, if this feature is enabled for a particular sender, then the message should:

- Not be batched with any other message.
- Have the same file name as the message that was received.

This ticket is to think through this use case and to elaborate on the design above and think through any edge cases or gaps. The design should be either documented in the repo or, if short, can be documented in the implementation ticket.

#### File name preservation request

ELIMS has also asked that the report we send to a receiver should have the same file name as the report we received from ELIMS. 

### Acceptance Criteria

- [ ] Design created for original message passthrough (if necessary)
- [ ] Implementation ticket created for original message passthrough which includes the design or a link to the design if a separate document

Andrey-Glazkv:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/12224?planning-poker) with Zenhub @arnejduranovic @jalbinson @JessicaWNava @jimmyfagan @thetaurean

Andrey-Glazkv:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/12224?planning-poker) with Zenhub @JFU-NAVA-PBC

Andrey-Glazkv:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/12224?planning-poker) with Zenhub @jsutantio @dylansmith-Nava

PR #12526, opened 2023-12-05
12224 Original Message Passthrough design proposal

jack-h-wang:
This PR adds a proposed design for the Original Passthrough feature.

## Changes
- Add proposed design for Original Passthrough.

## Linked Issues
- #12224

## To Be Done
- #12813 
- #12814 
- #12816

### DIFF
--- a/prime-router/docs/design/proposals/0024-original-message-passthrough.md
+++ b/prime-router/docs/design/proposals/0024-original-message-passthrough.md
@@ -1,0 +1,129 @@
+# Proposal for Original Message Passthrough
+
+## Background
+
+In order to meet data integrity requirements, a sender of HL7v2 has requested a feature in the Universal Pipeline to
+preserve the original unaltered message and convey it to receivers with no alterations, including preserving the
+original file name.
+
+## Assumptions
+
+This design is strictly for an HL7v2 sender and does not apply to a FHIR sender. Likewise, this does not apply to a FHIR
+receiver.
+
+## Possible Implementations
+
+We will need to store enough information in the pipeline to:
+
+* Identify an item as an original passthrough item
+* Reproduce the content of the original item
+
+The following proposed implementations can be grouped into two major approaches: storing the HL7v2 data within the FHIR
+bundle, or utilizing the blobURL of the original item upload.
+
+### Store HL7v2 in custom FHIR extension
+
+This set of implementations seeks to store the entire content of the incoming HL7v2 file within the FHIR bundle as a
+custom extension when the incoming file is received, after which the stored data would be used to reproduce the file.
+The custom extension is able to store the sender setting specifying original passthrough, the incoming HL7v2 filename,
+and the body of the file.
+
+#### Pros
+
+* The HL7v2 message in its entirety can be transmitted even to a FHIR receiver, should a receiver wish to see the source
+  file.
+    * It should be noted that if we elect to store the HL7v2 message for all incoming files that the message would not
+      reflect any filtering that would occur for a FHIR receiver.
+
+#### Cons
+
+* At a minimum, this will require adjusting the original HL7v2 message to serialize newline characters as `\n` to allow
+  the message to be stored in the FHIR bundle. This is a lossless encoding that will need to be handled when exporting
+  the file.
+
+#### Implementations
+
+* **Add feature to the FHIR converter library**:
+
+  Modify the CDCgov fork of the `hl7v2-fhir-converter` library to store the original filename and body of the input HL7
+  message inside the FHIR bundle as an extension within the bundle's Provenance resource.
+    * *Pro*: Storing the HL7v2 file body becomes automatic with minimal code change, as the FHIR converter library
+      already receives the HL7v2 body to process it.
+    * *Con*: The converter library currently does not receive the submission's original filename. We will need a means
+      of supplying this to the library.
+        * To consider: The FHIR converter is not intended to be specific to ReportStream and as such, any extensions to
+          it should be generic (e.g. pass in a settings JSON object to be included rather than creating settings
+          specific to ReportStream as named parameters).
+
+* **Add feature to the UP as a bundle enhancement**:
+
+  Add the original filename and the entire body of the received HL7v2 file to the FHIR bundle as an extension within the
+  bundle's Provenance resource.
+
+    * *Pro*: Unlike handling this feature in the FHIR converter library, the original HL7v2 filename can be more easily
+      made available when performing a bundle enhancement to add the extension.
+    * *Con*:
+
+### Retrieve the HL7v2 file from Azure blobstore
+
+The incoming HL7v2 file is stored to the Azure blobstore as part of the `receive` pipeline step. This can be utilized to
+extract the original HL7v2 file from the blobstore.
+
+#### Pros
+
+* This approach makes use of the existing copy of the original submission in the Azure blobstore. This guarantees a byte
+  for byte copy of the original submission as well as eliminating the need for an additional location to store the
+  original submission data.
+
+#### Cons
+
+* In this approach, the original file body is stored only in ReportStream's Azure blobstore, and will not be available
+  to downstream FHIR receivers should they wish to be able to view it in the future.
+
+#### Implementations
+
+* **Add feature to the UP as a bundle enhancement**:
+
+  We can introduce enhancing the FHIR bundle created in the `convert` pipeline step to add the sender setting specifying
+  original passthrough, the incoming HL7v2 filename, and the blobURL, as an extension within the bundle's Provenance
+  resource.
+
+    * *Con*: We need to be mindful of security concerns if URIs specific to ReportStream's Azure blobstore are to be
+      stored in the FHIR bundle. We may need to deidentify the URI to remove any sensitive information or strip this
+      information from the FHIR bundle when preparing to send it if a FHIR receiver is configured to receive it.
+
+* **Add feature to the UP translate step: retrieve blobURL from report_file DB table**:
+
+  In this approach, the process of creating the FHIR bundle is not altered in any way. Instead, the `translate` step is
+  altered to determine if original message passthrough is configured. If so, the `translate` step can then walk through
+  the `report_file` database table in a similar manner as done in the Submission History API to retrieve the blobURL and
+  the externalName of the original HL7v2 file and make a copy of it to send to the receiver.
+    * As information is not being added to the FHIR bundle, the determination that original passthrough is desired must
+      be made another way. This can take several forms, such as introducing a new `Report.Format` (e.g. `HL7_ORIGINAL`)
+      or having the `translate` function retrieve sender settings for the original submission.
+
+    * *Pro*: This approach potentially has the least amount of modifications to existing processes and makes use of
+      information that is already stored in the database. There is no need to store additional information about the
+      submission to facilitate the passthrough feature.
+    * *Pro*: There is no risk of leaking private information about ReportStream's storage as the FHIR bundle is not
+      modified.
+    * *Con*: This approach is highly reliant on the processes performed at the `receive` pipeline step remaining as they
+      are now. Any refactoring of this code must take original message passthrough into consideration.
+
+## Limitations
+
+* Filtering can still occur as a routing process, but the nature of sending an unmodified HL7v2 file will mean any
+  conditions that would have been eliminated by filtering would be sent regardless.
+* `MERGE` batching cannot be performed for original passthrough HL7v2 files.
+* Filenames in the Azure blobstore currently reflect pipeline specific information to more easily identify when and why
+  a file was created.
+    * There is currently a limitation that files must have unique names, and blobURLs cannot be reused.
+    * External filenames are in many cases based on the blobURL presently.
+
+## Questions
+
+* Will original passthrough utilize existing receiver filtering, or should we require a receiver specific to receiving
+  original passthrough files?
+    * If we utilize existing receivers, does the sender flag indicating original passthrough override receiver quality
+      filters?
+    * Alternatively, should a new `Report.Format` (e.g. `HL7_ORIGINAL`) be added to flag original passthrough items?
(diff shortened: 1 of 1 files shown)


arnejduranovic (changes requested):
Great work, it is apparent a lot of thought went into this. Kudos! 

I attempted to answer the open questions, or at least add my thoughts. I also sprinkled some additional thoughts throughout. Let's get on a call to discuss if needed! 

Requesting changes as a reminder to incorporate SOME answer to all the questions into the design and pick a design. Delete the questions afterwards.

prime-router/docs/design/proposals/0024-original-message-passthrough.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Background
+
+In order to meet data integrity requirements, a sender of HL7v2 has requested a feature in the Universal Pipeline to
+preserve the original unaltered message and convey it to receivers with no alterations, including preserving the
+original file name.
+
+## Assumptions
+
+This design is strictly for an HL7v2 sender and does not apply to a FHIR sender. Likewise, this does not apply to a FHIR
```
arnejduranovic, line 11: I would change this to "this design is for non-FHIR messages". It shouldn't matter if its HL7v2 or CSV or whatever, we should be able to stick it into the FHIR bundle (or pull it from azure).

jack-h-wang: This limitation is currently inherent to the Universal Pipeline not supporting CSV; to build this feature to support CSV would currently require extending the legacy pipeline as well.

prime-router/docs/design/proposals/0024-original-message-passthrough.md
```diff
@@ -0,0 +1,135 @@
...
+    * There is currently a limitation that files must have unique names, and blobURLs cannot be reused.
+    * External filenames are in many cases based on the blobURL presently.
+
+## Questions
+
+* Will original passthrough utilize existing receiver filtering, or should we require a receiver specific to receiving
+  original passthrough files?
+    * If we utilize existing receivers, does the sender flag indicating original passthrough override receiver quality
+      filters?
+    * How do we handle sending to a receiver configured for `MERGE` batching? Would the file be sent immediately or on
```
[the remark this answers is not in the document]
jack-h-wang: In the current design I landed on skipping the batch step altogether, but if you feel we need to reconsider this I'm open to discussing it.

prime-router/docs/design/proposals/0024-original-message-passthrough.md
```diff
@@ -0,0 +1,135 @@
...
+  conditions that would have been eliminated by filtering would be sent regardless.
+* `MERGE` batching cannot be performed for original passthrough HL7v2 files.
+* Filenames in the Azure blobstore currently reflect pipeline specific information to more easily identify when and why
+  a file was created.
+    * There is currently a limitation that files must have unique names, and blobURLs cannot be reused.
+    * External filenames are in many cases based on the blobURL presently.
+
+## Questions
+
+* Will original passthrough utilize existing receiver filtering, or should we require a receiver specific to receiving
```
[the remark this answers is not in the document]
jack-h-wang: I think I'd like to see more concrete examples of what we'd receive before we can plan out what needs to change in routing to support this.

prime-router/docs/design/proposals/0024-original-message-passthrough.md
```diff
@@ -0,0 +1,168 @@
...
+    * External filenames are in many cases based on the blobURL presently.
+
+## Feature Design
+
+After performing analysis of the possible implementations above, the **Add feature to the UP translate step: retrieve
+blobURL from report_file DB table** implementation was selected.
+
+The overall design of the original passthrough feature is as follows:
+
+* A new `Topic` will be used for original passthrough items. This guarantees that data that must be handled via original
```
arnejduranovic, line 145: I'm warming up to the idea of using Topics for the solution here. Instead of creating a whole new topic for this, what if we add a flag to all topics that indicate whether to essentially skip the translate step or not? It would just be another param akin to isUniversalPipeline:

```
/**
 * A submission with topic FULL_ELR will be processed using the full ELR pipeline (fhir engine), submissions
 * from a sender with topic COVID_19 will be processed using the covid-19 pipeline.
 */
enum class Topic(@JsonValue val jsonVal: String, val isUniversalPipeline: Boolean = true) {
    FULL_ELR("full-elr", true),
    ETOR_TI("etor-ti", true),
    ELR_ELIMS("elr-elims", true),
    COVID_19("covid-19", false),
    MONKEYPOX("monkeypox", false),
    TEST("test", false),
}
```

Then you can just use `Topic.sendOriginal` in translate function for your switch?

The reason I'm thinking more this way is because I'm thinking more about the existential question: "What is a topic?" and its been basically turning into a **contract** for a certain use case. For example, each Topic now indicates which pipeline it will use. So, when you select "FULL_ELR", you as the sender and receiver are both acknowledging and agreeing to the rules associated with that topic. So I think continuing to develop topics in this way (by adding another param called send_original) makes sense. The current topic "ELR_ELIMS" would be updated to have this set to true, the other topics will have it as false. If a sender and receiver both select this topic, they are agreeing to send/receive CDC ETOR ELIMS data from the STARLIMS system with the understanding the original message will be sent and transforms are disabled -- that makes sense to me?

arnejduranovic: If we go this route, that changes our "assumptions" a bit:

Original wording:

> This design is intended for a specific sender. Although the design can be expanded to encompass additional senders as needed, at this time it is less disruptive to assume this sender's requirements are unique. It should also be assumed that the long term goal for all senders on ReportStream is to eventually utilize ReportStream's filtering and routing
processes instead of bypassing these features.

Rough proposed wording:

> While this design is intended to support a specific use case, CDC ELIMS ETOR, it should be easily applicable to other use cases as well and fit into the existing architecture of the pipeline.

Thoughts:

I disagreed with a lot of the language in the original wording. If one use case is asking for this functionality, other use cases may ask for additional functionality as well, either the same or different as this. Thus, it becomes important to start thinking about how can we build this out while allowing us to further expand it in the future in a maintainable way. I'm not sure Topics in the way I propose we use them is the final solution here, but I think its a good first step.

prime-router/docs/design/proposals/0024-original-message-passthrough.md
```diff
@@ -0,0 +1,168 @@
...
+* `Report.formExternalFilename` uses the filename of the blobstore item, so there should not be any further
+  modifications necessary as long as the blobstore item is named correctly by the `translate` function.
+
+### Issues to be created
+
+* Add header to `receive` step to allow sender to supply original filename and store in database
+  field `report_file.external_name`
+    * Submissions that provide an `external_name` that already exists should be rejected.
+* Write function to look up original filename and blob URL for a given report ID
+* Add step to `translate` function to check message topic, copy the original blob with original filename (under receiver
```
arnejduranovic, line 164: I think this works, but double check: If a receiver comes back and says "Yo, I don't know the report ID but I got a file named File-ABC-123.hl7 from ya'll and it has these issues with it" do we have enough info to figure out where that came from? What if we sent many files with that name? It seems like we need to append the report ID to the file name, which ELIMS would be okay with, but we need to make sure file name length is still reasonable and won't cause issues.

arnejduranovic: Also, what if the the transport protocol is REST? Do we put the original file name in the header in that case?

prime-router/docs/design/proposals/0024-original-message-passthrough.md
```diff
@@ -0,0 +1,168 @@
...
+
+## Background
+
+In order to meet data integrity requirements, a sender has requested a feature in the Universal Pipeline to
+preserve the original unaltered message and convey it to receivers with no alterations, including preserving the
+original file name.
+
+## Assumptions
+
+This design is intended for a specific sender. Although the design can be expanded to encompass additional senders as
```
arnejduranovic, line 11: Right now, this is also an assumption, right?:

> The sender that has agreed to send"original functionality shall NOT send batched messages. Each submitted report shall contain only one "message", otherwise a receiver may receive duplicate messages. 

That said, I feel like we shouldn't trust the sender and verify this assumption in the pipeline, meaning not only is there special logic for this feature in the translate step but also receive/convert step, right?

arnejduranovic: Are there any other issues you can think of with any of the other pipeline steps?

arnejduranovic: Some implementation thoughts:

We could add ANOTHER param to the topic alongside `isUniversalPipeline` and `sendOriginal` called `allowBatch` and would default to true. This new param would be used in the receive or convert step and reject the message if it breaks the rule. I think its good that this decouples this logic from send_original, that way other topics can reuse the individual pieces as needed.

prime-router/docs/design/proposals/0024-original-message-passthrough.md
jack-h-wang: Unless I'm mistaken, REST transport already includes the filename under `HttpHeaders.ContentDisposition`

prime-router/docs/design/proposals/0024-original-message-passthrough.md
```diff
@@ -0,0 +1,172 @@
...
+In order to meet data integrity requirements, a sender has requested a feature in the Universal Pipeline to
+preserve the original unaltered message and convey it to receivers with no alterations, including preserving the
+original file name.
+
+## Assumptions
+
+* While this design is intended to support a specific use case, CDC ELIMS ETOR, it should be easily applicable to other
+  use cases as well and fit into the existing architecture of the pipeline.
+* Senders will send using a message format supported by the Universal Pipeline (currently HL7v2 and FHIR).
+* Senders using the sendOriginal functionality shall NOT send batched messages. Each submitted report shall contain only
```
arnejduranovic, line 14: no changes required here, just pointing out this assumption should be verified by the pipeline in the implementation and the report should be rejected if it contains more than one message.

prime-router/docs/design/proposals/0024-original-message-passthrough.md
```diff
@@ -0,0 +1,172 @@
...
+* Reproduce the filename and content of the original item
+
+A new mechanism needs to be provided to receive the original filename, as the current API does not provide a means to
+take it in. This can take the form of an additional header field.
+
+This information would be made available to the `translate` step of the UP, which would then be changed to retrieve the
+original item instead of performing receiver translation. Furthermore, the `translate` step would directly insert
+a `send` task for the item, bypassing the `batch` step.
+
+The following proposed implementations can be grouped into two major approaches: storing the HL7v2 data within the FHIR
```
arnejduranovic, line 31: Great table here!

prime-router/docs/design/proposals/0024-original-message-passthrough.md
```diff
@@ -0,0 +1,172 @@
...
+* As the `Topic` is unique, no other identifiers for sendOriginal items are needed. Configuring the sender and
+  receivers for the Topic with `sendOriginal` defined is all that is required.
+* The `translate` step will copy the original blob data from the blobstore instead of translating from the internal
+  FHIR bundle. Enhancements to `BlobAccess` that have been made in preparation for the `Manage Translation Schemas`
+  feature can be utilized to perform this operation. The `translate` step will then directly insert the resulting blob
+  to the `send` queue.
+* `Report.formExternalFilename` uses the filename of the blobstore item, so there should not be any further
+  modifications necessary as long as the blobstore item is named correctly by the `translate` function.
+
+### Issues to be created
```
arnejduranovic, line 162: I think the following tickets may be missing, based on the descriptions of the three tickets below:

- Update receive or convert step to verify submitted report only contains a single message when a report is received under a topic with that setting enabled.  We can discuss this during implementation, but imo this should be implemented using something like polymorphism so we don't have "if topic.setting == true" sprinkled everywhere in the future. Same idea goes for translate step.

prime-router/docs/design/proposals/0024-original-message-passthrough.md
```diff
@@ -0,0 +1,172 @@
...
+
+* Add header to `receive` step to allow sender to supply original filename and store in database
```
arnejduranovic, line 164: What will this header be called? The stakeholders are already asking me :P

prime-router/docs/design/proposals/0024-original-message-passthrough.md
arnejduranovic: Also, this should probably just be done with the translate step update ticket. And then the first two bullets you have should be done together. While the work can be split up the way you have it, its less overhead/management to combine them.

prime-router/docs/design/proposals/0024-original-message-passthrough.md
```diff
@@ -0,0 +1,172 @@
...
+  field `report_file.external_name`
+* Write function to look up original filename and blob URL for a given report ID
+* Add step to `translate` function to check message topic, copy the original blob with original filename (with report ID
+  appended), and insert into send queue
+    * The database currently requires blobURLs to be unique. Appending the report ID to the original filename should
```
arnejduranovic, line 169: I want to be clear that if a receiver gets a file from RS that has the ReportID included at the start or end of the original file name that is acceptable.

arnejduranovic (changes requested):

prime-router/docs/design/proposals/0024-original-message-passthrough.md
jack-h-wang: I don't feel strongly but I would suggest mirroring what is used in RESTTransport and use `Content-Disposition` with value `filename=<filename>`

prime-router/docs/design/proposals/0024-original-message-passthrough.md
jack-h-wang: I've added the additional requirement and consolidated the tickets; I will work on updating the tickets as written to reflect this.

arnejduranovic (approved):
Great work!

PR #12527, opened 2023-12-05
Add availability testing for the SFTP container

### LINKED ISSUE #10368: Create SFTP monitor for the SFTP containers in Stag/Prod so we know if the are unresponsive
Issue says it all - we want to know if the SFTP server is not working. Ideally we'd just SSH into it every 10 minutes or so

dkrylovsb:
This PR adds availability testing for the SFTP container. A new function in the admin app does the checking, which then feeds back into Application Insights as an additional availability check. 

Test Steps:
1. Do `terraform apply`. Ensure Application Insights display status for the SFTP container. 

## Changes
- A new admin app function: `confirm_sftp_access.py`.
- A new availability test in App Insights that uses this function to poll SFTP.

### Testing
- [x] Tested locally?
## Linked Issues
- Fixes #10368


jeremy-page (approved):
Thank you Denis

PR #12758, opened 2023-12-28
Adds enrichments

JessicaWNava:
This PR adds the ability to run enrichments on a bundle. 

Test Steps:
1. Make sure the unit tests and integration tests pass.
2. Run enrichments on a bundle via the command line. 

## Changes
- Allows enrichments on a bundle
- Adds the ability to add enrichments when running on the CLI
- Adds the ability to add enrichments to integration tests. 

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- #issue 12758


prime-router/src/main/kotlin/fhirengine/engine/FHIRTranslator.kt
```diff
@@ -108,22 +108,30 @@ class FHIRTranslator(
...
         receiver: Receiver,
         bundle: Bundle,
-    ) = when (receiver.format) {
-        Report.Format.FHIR -> {
-            if (receiver.schemaName.isNotEmpty()) {
-                val transformer = FhirTransformer(receiver.schemaName)
+    ): ByteArray {
+        if (receiver.enrichmentSchemaNames.isNotEmpty()) {
+            receiver.enrichmentSchemaNames.forEach { enrichmentSchemaName ->
+                val transformer = FhirTransformer(enrichmentSchemaName)
```
arnejduranovic, line 114: We should have a .info message here probably along the lines of "Applying schema <schema-name>". This will make more sense and be more helpful once we can add context to the log messages, like report_id.

JFU-NAVA-PBC (approved):
Great design and implementation works, in addition to global Provenance enrich, now per receiver fhir enrichment is a powerful feature to cater receiver specific fhir resource needs.

passed all tests and examed added unit tests and integration tests, looks good.

we now have a dedicated setting for integration tests:

./prime-router/src/integration/resources/settings/organizations.yml

if a sample receiver can be furnished there showing how a enrichment can be configured, that'd be great, optional since it is not ion AC.

JessicaWNava:
Added the test as requested @JFU-NAVA-PBC, thank you! 

@arnejduranovic those tests don't work like we talked about this morning. They do not work with the smoke tests, they work with the integration tests. I added a piece so that they work the same as schemaName which means that if enrichmentSchema isn't filled out explicitly, it will look to the receiver to see if it is specified there and use those.

JessicaWNava:
Ready for your review @arnejduranovic

prime-router/src/testIntegration/resources/datatests/translation-test-config.csv
```diff
@@ -1,29 +1,33 @@
...
+FHIR_to_HL7/simple_report_with_date_aoe_question.fhir,,FHIR_to_HL7/simple_report_with_date_aoe_question.hl7,metadata/hl7_mapping/receivers/STLTs/CA/CA-receiver-transform,HL7,PASS,"","","metadata/fhir_transforms/senders/SimpleReport/simple-report-sender-transform","","",""
+HL7_to_FHIR/sample_orm_20230809-001.hl7,,HL7_to_FHIR/sample_orm_20230809-001-with-enrichment.fhir,,FHIR,PASS,"","","","","","/metadata/enrichment/testing/testing2"
+HL7_to_FHIR_to_HL7/elims_40_4988249_33033.hl7,,HL7_to_FHIR_to_HL7/elims_40_4988249_33033_with_enrichment.hl7,metadata/hl7_mapping/ORU_R01/ORU_R01-test,HL7,PASS,"","test.dev-elims","","","","/metadata/enrichment/testing/testing2"
+HL7_to_FHIR_to_HL7/elims_40_4988249_33033.hl7,,HL7_to_FHIR_to_HL7/elims_40_4988249_33033_with_double_enrichment.hl7,metadata/hl7_mapping/ORU_R01/ORU_R01-test,HL7,PASS,"","test.dev-elims","","","","/metadata/enrichment/testing/testing,/metadata/enrichment/testing/testing2"
+HL7_to_FHIR_to_HL7/elims_40_4988249_33033.hl7,,HL7_to_FHIR_to_HL7/elims_40_4988249_33033_with_double_enrichment.hl7,metadata/hl7_mapping/ORU_R01/ORU_R01-test,HL7,PASS,"","test.dev-elims","","ca-test.enrichment-receiver","",""
```
arnejduranovic, line 33: I like this test and struggle to see the value of the other ones? I guess the one above this allows you to specify the enrichment schema name outside of the configured receiver, but even then, not sure we need that? Still not sure about the first two you added though. Seems to me if we updated this receiver to contains two enrichment schemas that would be all we need to test your new functionality?

arnejduranovic (changes requested):
Looks good! A few minor comments below.

prime-router/src/main/kotlin/Receiver.kt
```diff
@@ -52,6 +54,7 @@ open class Receiver(
     val description: String = "",
     val transport: TransportType? = null,
     val externalName: String? = null,
+    val enrichmentSchemaNames: List<String> = emptyList(),
```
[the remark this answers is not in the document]
JessicaWNava: Because we don't want the other settings that go along with TranslatorConfigurations, just the names to straight up apply the enrichments.

prime-router/metadata/enrichment/testing/testing.yml
```diff
@@ -0,0 +1,13 @@
+constants:
```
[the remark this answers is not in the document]
JessicaWNava: The problem is, that folder is not where the integration tests look to for getting the schemas. These can easily be removed once we have some real life cases available to us.

prime-router/src/testIntegration/resources/datatests/translation-test-config.csv
JessicaWNava: One checks it works on HL7 to FHIR (should be no difference, but figure it is still best to have a test in case someone mucks it up). One tests HL7 to HL7, but with only one schema name to make sure we don't get dependent on it being more than one name. One tests multiple schema names. One tests that we can use the configuration in the yml file as you asked me to do last week.

arnejduranovic: The following two can be combined by specifying multiple in the receiver config, no?
- `One tests that we can use the configuration in the yml file`
- `One tests multiple schema names.`

JessicaWNava: Sort of. It would mean that we aren't testing that the tests can handle both methods, but if you are fine with that, I can remove it.

prime-router/metadata/enrichment/testing/testing.yml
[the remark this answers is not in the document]
JessicaWNava: Nvm, as you said they could be moved, fixed!

prime-router/src/main/kotlin/Receiver.kt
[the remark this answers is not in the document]
arnejduranovic: This makes sense to me, given the use case. per the design, enrichments will always be FHIR transforms and should not require extra params. Let's go with this for now and we can have a new ticket if we find the need for it later.

arnejduranovic (approved):
Changes look good and all tests pass locally, including smokeTests.

Issue #11565, opened 2023-09-27
Evaluate the storage method for Azure blob store connection settings


jack-h-wang:
### User Story
As a ReportStream engineer, I want to follow best practices for storage of sensitive data so that I can be sure sensitive data remains secure.

### Description/Use Case
ReportStream stores connection settings to Azure blob stores for both accessing RS specific blob stores as well as connecting to non-RS blob stores when sending pipeline data. We should evaluate whether the storage of these settings is sufficiently secure or if we need to consider migrating these settings to a new location.

### Risks/Impacts/Considerations
Evaluate connection strings as they are retrieved by the BlobAccess class and the BlobStoreTransport method.

### Dev Notes
Consider use of the SecretHelper class, using the Email Engine functions as an example.

### Acceptance Criteria
- Evaluated the storage of Azure blob store connection settings and create a plan to migrate settings to a more secure location if needed.

bishoyayoub:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/11565?planning-poker) with Zenhub @arnejduranovic @jack-h-wang @JessicaWNava @JFU-NAVA-PBC @thetaurean @luis-pabon-tf

bishoyayoub:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/11565?planning-poker) with Zenhub @mkalish

Andrey-Glazkv:
Need to discuss this ticket - @arnejduranovic to schedule a call/discussion on this during sprint 79

JFU-NAVA-PBC:
Implementation Proposal:

1. Parameterize credentials, sensitive info as environment variables
2. Provision dedicated azure key vault for each environment:
     -  PROD
     -  STAGING
     -  TEST
     -  DEMO
     -  ... 
3. List of environment variables:
    - PROD:
       - RS_ENV_DB_USER =
       - RS-ENV_DB_PASSWORD =
       - RS_ENV_DB_URL =





2. Store the parameters in step on into Azure Key Vaults

JFU-NAVA-PBC:
Move review feedbacks and conversations here from PR (which is POC and will closed and deleted after research done).

=========== PR review feedbacks:
I think this is an interesting idea, though I'm not sure this is 100% the right path.

Some thoughts:

[...]

PR #12760, opened 2023-12-28
platform/jimfuqian/11565 POC: fetch server required creds, sensitive data from AZ key vault

JFU-NAVA-PBC:
This PR

Serves as a proof of concept for 

Test Steps:
1. Instruction to run locally is included in ticket #11565  

Note, no worries about CI failure, not suppose to run on CI due to sourcing ENV vars from a AZ key vault.

For local demo only.

## Changes

### Testing
- [ ] Tested locally?
- [ ] Added tests?

### Process

## Linked Issues
- Fixes #issue

## To Be Done
- #issue

    - new endpoint A
    - new endpoint B


mkalish (changes requested):
I think this is an interesting idea, though I'm not sure this is 100% the right path.  

Some thoughts:

- there is the intention that this be an open source project, so we need to avoid adding local configuration that would prevent anyone from cloning and working on this repo
- For local dev, I'm not seeing why we would need to fetch the credentials from azure rather than using the local vault implementation (this is how it currently works for fetching the blob transport params)
- I think we need to work more closely with devops since ultimately where these credentials live is there domain.  Taking a look in azure, it looks like value like the DB password and blob connection string exist as configuration on the function app
- I think the change in SftpTransport is actually undoing a working version of what we want to see where we dynamically load the credential out of the vault


My read on the scope of this work was to work with the devops team to see if it could make sense to make sensitive values no longer be an app config setting and fetch them out the existing azure vault, whether that be via a change to the application code or as a change in how the azure function works (i.e. using [managed identities](https://learn.microsoft.com/en-us/azure/app-service/tutorial-connect-msi-azure-database?tabs=sqldatabase-sc,userassigned-sc,dotnet,dotnet-mysql-mi,dotnet-postgres-mi,windowsclient) rather than passwords.)

JFU-NAVA-PBC:
> I think this is an interesting idea, though I'm not sure this is 100% the right path.
> 
> Some thoughts:
> 
> * there is the intention that this be an open source project, so we need to avoid adding local configuration that would prevent anyone from cloning and working on this repo

> * For local dev, I'm not seeing why we would need to fetch the credentials from azure rather than using the local vault implementation (this is how it currently works for fetching the blob transport params)
> * I think we need to work more closely with devops since ultimately where these credentials live is there domain.  Taking a look in azure, it looks like value like the DB password and blob connection string exist as configuration on the function app
> * I think the change in SftpTransport is actually undoing a working version of what we want to see where we dynamically load the credential out of the vault
> 
> My read on the scope of this work was to work with the devops team to see if it could make sense to make sensitive values no longer be an app config setting and fetch them out the existing azure vault, whether that be via a change to the application code or as a change in how the azure function works (i.e. using [managed identities](https://learn.microsoft.com/en-us/azure/app-service/tutorial-connect-msi-azure-database?tabs=sqldatabase-sc,userassigned-sc,dotnet,dotnet-mysql-mi,dotnet-postgres-mi,windowsclient) rather than passwords.)

arnejduranovic:
This is an interesting proof of concept and conversation that should be archived, but I don't see a need for immediate tickets to be generated or to proceed with this approach. We would first like to complete the AC of the ticket this is referencing, so I am closing this PR and we can proceed with fixing the ticket in a new PR (if a PR is even needed). I update the ticket to reference this PR for record keeping purposes.

PR #12817, opened 2024-01-03
Added documentation for retrieving unmapped bundles

### LINKED ISSUE #12283: Create receiver for messages containing unmapped conditions
### User Story

As a ReportStream engineer,
I would like to send messages that contain unmapped conditions to a special place,
so that I can easily review them

### Description/Use Case

Implement section `Unmapped receiver queue` of the code to condition [design](https://github.com/CDCgov/prime-reportstream/blob/5097e07acd10747c8cd2104c45bb986c29d623db/prime-router/docs/design/proposals/0023-condition-to-code-mapping/0023-condition-to-code-mapping.md#mapping-loincsnomed-codes-in-received-messagebundle-to-condition-code) document

### Dev Notes

OrgName suggestion: ReportStream
ReceiverName suggestion: UnmappedConditions

### Acceptance Criteria

- [ ] Receiver created/tested in staging and production.
- [ ] Set up azure container in RS for transport

thetaurean:
This PR adds documentation that covers retrieving source bundles with unmappable conditions.

Test Steps:
1. Submit `valid.fhir` to the receiver `development.dev-elims` (may have to change topic/format)
   ```zsh
    curl --location 'http://localhost:7071/api/reports' --header 'client: development.dev-elims' --header 'Content-Type:application/fhir+ndjson' --data-binary @/Users/gabriel/dev/focus/prime-reportstream/prime-router/src/test/resources/fhirengine/engine/routing/valid.fhir
   ```
2. Run the query and verify the output matches what is expected

## Changes
- Added SOP for retrieving unmappable conditions

### Testing
- [x] Tested locally?

### Process

## Linked Issues
- Fixes #12283

### DIFF
--- a/prime-router/docs/getting-started/standard-operating-procedures/observation-mapping-table.md
+++ b/prime-router/docs/getting-started/standard-operating-procedures/observation-mapping-table.md
@@ -105,3 +105,28 @@ spreadsheet using its code/OID. The resulting row is then mapped to the appropri
 Either copy the values from the website and map them into the appropriate columns  
 -OR-  
 Sign up for an account, download the CSV, and map the data from it
+
+## Checking for mapping failures
+Use the following query:
+```postgresql
+SELECT detail->>'message' as message, detail->>'fieldMapping' as field, action_log.report_id, report_file.body_url
+    FROM action_log
+    INNER JOIN report_file ON report_file.report_id = action_log.report_id
+    WHERE action_log.detail->>'errorCode' = 'INVALID_MSG_CONDITION_MAPPING'
+    ORDER BY action_log.created_at DESC
+    LIMIT 100;
+```
+
+Output will include the missing code, its origin, and the URL of the source data. Use the azure storage explorer
+or the azure portal to download the file being careful to observe PII precautions. 
+
+### Example output
+
+| message | field | report\_id | body\_url |
+| :--- | :--- | :--- | :--- |
+| Missing mapping for code\(s\): N | observation.valueCodeableConcept.coding.code | 3a947e0f-0832-403d-a9d8-92f9b88557a8 | http://localhost:10000/devstoreaccount1/reports/receive%2Fdevelopment.dev-elims%2FNone-3a947e0f-0832-403d-a9d8-92f9b88557a8-20240102233706.fhir |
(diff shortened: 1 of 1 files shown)


arnejduranovic (approved):
Tested locally and output of SQL query looked good.

brick-green (approved):
This works so approving. But what's the usefulness of the `field` column? Since this query is limited to condition mapping, will the value always be "observation.valueCodeableConcept.coding.code"? If so, it might be nice to replace with sender ID. Adding a created_at date column would be useful as well.

For future work, I'm wondering how engagement will know if a mapping has been added. This query would show past condition mapping errors but could be confusing if someone adds the mapping then another person runs this query and still sees the mapping error.

arnejduranovic:
> This works so approving. But what's the usefulness of the `field` column? Since this query is limited to condition mapping, will the value always be "observation.valueCodeableConcept.coding.code"? If so, it might be nice to replace with sender ID. Adding a created_at date column would be useful as well.
> 
> For future work, I'm wondering how engagement will know if a mapping has been added. This query would show past condition mapping errors but could be confusing if someone adds the mapping then another person runs this query and still sees the mapping error.

I like these points. I think the columns you suggested would be good additions. We were hoping to get feedback from engagement once they start using this feature but I'm not totally opposed to making these changes ahead of time. Regarding your second comment about how engagement will know if a mapping has been added, I think the answer right now is they would need to query the mapping table to see. I wonder if we need to adjust that table to have the date of when the code was added, so engagement can compare it to the timestamps of the log messages?

thetaurean:
> This works so approving. But what's the usefulness of the `field` column? Since this query is limited to condition mapping, will the value always be "observation.valueCodeableConcept.coding.code"? If so, it might be nice to replace with sender ID. Adding a created_at date column would be useful as well.

The field value could also be `observation.code.coding.code` so this distinguishes the origin. It comes from the pre-existing `fieldMapping` property on the log messages. I'm 100% with you on the `created_at` column, I'll update the query here and on metabase.

> 
> For future work, I'm wondering how engagement will know if a mapping has been added. This query would show past condition mapping errors but could be confusing if someone adds the mapping then another person runs this query and still sees the mapping error.

Good point and I share the concern. As @arnejduranovic mentioned, checking if a code is mapped requires manually checking the `observation-mapping` table. I'll look into adding the mentioned column to the output of the query and brainstorm if there's another way.

I think ideally we can also handle this with process: whoever investigates mapping errors leaves a message on slack. That will indicate all unmappeds before that timestamp have been addressed.

Issue #10490, opened 2023-07-18
Implement a function that accepts source and destination directories and recursively copies all the files


mkalish:
### User Story
As a report stream developer, I would like a function that recursively downloads all the files in directory 

### Description/Use Case
There will be downstream work that needs to download a whole set of files in a directory when syncing schemas

### Dev Notes
Signature will be similar to `BlobAccess.copyBlob` but should operate on a whole directory rather than single blob

### Acceptance Criteria
- [ ] A new function exists that can download the files recursively between azure containers

Andrey-Glazkv:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/10490?planning-poker) with Zenhub @arnejduranovic @brick-green @jack-h-wang @jalbinson @jimmyfagan @mkalish @thetaurean

JessicaWNava:
After writing a bunch of code, discovered there is already code that moves blobs form one blob dir to another, so just need to implement the download to the local dir.

PR #12854, opened 2024-01-04
Platform/jessica/10490 copy blobs to a local directory

JessicaWNava:
This PR allows you to download a directory of blobs to a local directory. Note, the ticket originally was also for downloading from one blob store dir and moving to another, but it turns out that work is already done.

Test Steps:
1. Make sure the unit tests pass.

## Changes
- Adds a function to download a directory of blobs to a local directory.

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- #10490


thetaurean (approved):
LGTM 👍

prime-router/src/main/kotlin/azure/BlobAccess.kt
```diff
@@ -336,6 +338,25 @@ class BlobAccess() : Logging {
             return toBlobUrl
         }
 
+        /**
+         * Download all blobs located at [sourceBlobDirectoryUrl] with container info [sourceBlobContainerInfo] to the
+         * [destinationDirectoryUrl].
+         */
+        fun downloadBlobsInDirectoryToLocal(
```
arnejduranovic, line 345: It appears this function would flatten the downloaded files into a single directory instead of preserving a potential virtual hierarchy. I think this is fine for the intended use case but something to note.

prime-router/src/main/kotlin/azure/BlobAccess.kt
```diff
@@ -336,6 +338,25 @@ class BlobAccess() : Logging {
...
+            sourceBlobDirectoryUrl: String,
+            sourceBlobContainerInfo: BlobContainerMetadata,
+            destinationDirectoryUrl: String,
```
arnejduranovic, line 348: This is not a url, maybe rename to `destinationDirectoryPath`?

prime-router/src/test/kotlin/azure/BlobAccessTests.kt
```diff
@@ -120,6 +122,43 @@ class BlobAccessTests {
...
+
+            BlobAccess.downloadBlobsInDirectoryToLocal("foo", sourceBlobContainerMetadata, ".")
+
+            contentToUpload.forEach { fileName ->
+                val file = File("./$fileName")
+                val content =
+                    file.inputStream()
+                        .readBytes().toString(Charsets.UTF_8)
+                assertEquals(fileName, content)
+                file.delete()
```
arnejduranovic, line 158: If assert fails does the file get deleted?

arnejduranovic (changes requested):

prime-router/src/main/kotlin/azure/BlobAccess.kt
JessicaWNava: No, it actually creates the directory structure because of how the function `writeByteArrayToFile` works

prime-router/src/test/kotlin/azure/BlobAccessTests.kt
```diff
@@ -120,6 +122,43 @@ class BlobAccessTests {
...
+                listOf("foo/item1.txt", "foo/item2.txt", "foo/item13.txt", "foo/baz/item3.txt")
+            contentToUpload.forEach { content ->
+                BlobAccess.uploadBlob(
+                    content,
+                    content.toByteArray(),
+                    sourceBlobContainerMetadata
+                )
+            }
+
+            BlobAccess.downloadBlobsInDirectoryToLocal("foo", sourceBlobContainerMetadata, ".")
```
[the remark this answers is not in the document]
JessicaWNava: Nothing happens, so there isn't really anything to test for. I did add a logged warning though for if there is nothing to copy.

arnejduranovic (approved):

PR #12957, opened 2024-01-16
change PD Alert to prime-reportstream-engineering

supriyaaddagada:
This PR is to change the on call schedule PD alert to go to prime-reportstream-engineering instead of going to pagerduty-alert-dump


jeremy-page (approved):
Thank you!

PR #12966, opened 2024-01-16
12722: Enrich logs with additional queryable data

### LINKED ISSUE #12722: Enrich log messages with searchable context
### User Story

As an engineer debugging the universal pipeline,
I want to be able to filter events (like trace or custom events) based on the report_id and message_id they are associated with,
so that I can quickly identify the problem.

### Description/Use Case

When a report is received by RS, RS assigns it a unique report_id. The next pipeline step that runs, convert, will generate one or more reports with new report_ids and the other steps will do the same until ultimately a report gets sent to a receiver. When debugging issues, it would be really helpful to filter log messages and events in azure logs by the report_id they are associated with. Special consideration should be paid to parent or child ids. To store all the info, we could decorate log messages with the following two searchable fields:
- `report_id` - the current reports id
- `submitted_report_id `- single value that has the id of the report that was originally submitted
- `parent_report_ids` - list of all reports this is a child of (report lineage in other words). Should be ordered.

Each report RS receives either contains a piece of data inside each item that is known as a "message_id", which is supposed to be a unique ID (like UUID) that identifies a message as unique. This will be unknown at the receive step, but will be known from the convert step on, and should decorate log messages and events where possible.

### Dev Notes

A popular way to enrich log messages with context is the [MDC pattern](https://www.baeldung.com/mdc-in-log4j-2-logback). Kotlin specific info can be found [here](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-slf4j/kotlinx.coroutines.slf4j/-m-d-c-context/). If we go this route, we need to make sure that these values will be searchable in Azure logs. If not, we may need to investigate how we can enrich trace messages with queryable key:value pairs using azure libraries.

### Acceptance Criteria

- [ ] Pattern for enriching log messages is designed and documented
- [ ] All log messages contain the report_id they are associated with, if applicable
- [ ] All log messages contain the message_id they are associated with, if applicable

jalbinson:
This PR adds structured logging in JSON to Report Stream. This allows Azure AppInsights to be able to pull apart different aspects of the logs into queryable data. The AppInsights Java agent will also inject additional fields (such as a trace id!) in the MDC.

This PR also adds helpful logging extension functions to able to take advantage of adding fields to the MDC. This allows us to query for logs based on custom fields defined in our source code. The fields will show up in the `customDimensions` property in the `traces` table

Test Steps:
1. Run Report Stream locally with the AppInsights java agent and configured AppInsights environment variable
2. Submit a Report
3. Query for that report in the `traces` table in AppInsights
4. See the receiver name and report id as queryable fields in `customDimensions`

## Changes
- Structured logging
- Extension functions for adding MDC fields

### Testing
- [x] Tested locally?
- [ ] Added tests?

## Linked Issues
- Fixes #12722


jalbinson:
Documentation still required

prime-router/src/main/kotlin/azure/SendFunction.kt
```diff
@@ -155,9 +157,13 @@ class SendFunction(private val workflowEngine: WorkflowEngine = WorkflowEngine()
         actionHistory: ActionHistory,
         isEmptyBatch: Boolean,
     ): ReportEvent {
+        val reportContext = mapOf(
```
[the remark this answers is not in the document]
jalbinson: Makes sense. I was trying to give fine-grained control of adding context params to specific log messages to be able to use `traces` to make graphs and visualizations. I've got custom events working well now so I think we can lean on that in a future PR for that use case.

I was concerned about the context leaking between requests but @mkalish assured me that by the nature of how an Azure functions app worked, that was basically impossible. I think if we want all logs to contain the same context we could just push directly to the MDC.

If we want certain blocks of logs to contain additional fields I can definitely add a similar function to `withLoggingContext`.

Is behavior like this what were looking for?
```kotlin

fun logging() {
    ThreadContext.put("everywhere", "value")
    logger.info("hello world!") // MDC = everywhere -> value
    withLoggingContext(mapOf("inside" to "other value")) {
        logger.info("inside") // MDC = everywhere -> value, inside -> other value
    }
    logger.info("back outside") // MDC = everywhere -> value
}
```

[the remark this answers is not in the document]
arnejduranovic: Yes your example is exactly what I had in mind. I would just add that if another function got called inside of that one that it would also have the enriched logs.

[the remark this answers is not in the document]
arnejduranovic:
```
fun logging() {
    ThreadContext.put("everywhere", "value")
    logger.info("hello world!") // MDC = everywhere -> value
    withLoggingContext(mapOf("inside" to "other value")) {
        logger.info("inside") // MDC = everywhere -> value, inside -> other value
    }
    logger.info("back outside") // MDC = everywhere -> value
    logging_two()
}

fun logging_two() {
    logger.info("hello two") // MDC = everywhere -> value
}
```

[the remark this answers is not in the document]
jalbinson: Sounds good to me!

prime-router/src/main/kotlin/common/LoggerExt.kt
```diff
@@ -0,0 +1,45 @@
...
+}
+
+fun KotlinLogger.fatal(msg: Any, t: Throwable?, context: Map<String, String>) = withContext(context) {
+    this.fatal(msg, t)
+}
+
+/**
+ * Logger function wrapper to push the context map to MDC and clear it after running the log function
+ */
+private fun withContext(context: Map<String, String>, logFn: () -> Unit) {
```
mkalish, line 41: Making me think it's worth adding https://github.com/oshai/kotlin-logging and we could sub this out for there function

prime-router/src/main/kotlin/common/LoggerExt.kt
```diff
@@ -0,0 +1,45 @@
...
+import org.apache.logging.log4j.ThreadContext
+import org.apache.logging.log4j.kotlin.KotlinLogger
+
+/**
+ * Collection of logger extension functions that accept a context map along with the message
+ *
+ * The context map fields will show up in the "customDimensions" collection in azure in the "traces"
+ * or "exceptions" table (depending on if an exception was passed)
+ */
+fun KotlinLogger.trace(msg: Any, context: Map<String, String>) = withContext(context) {
```
mkalish, line 12: Can we add follow up ticket, but I think it would be great to either add an overload that would allow passing in values like `Map<String, SomeSerializableType>`?  Mostly thinking it would be great to do something like `logger.info("Starting FHIR Convert", mapOf("message", FhirConvertMessage))`

prime-router/src/main/resources/jsonTemplate.json
```diff
@@ -0,0 +1,37 @@
+{
+  "mdc": {
```
mkalish, line 2: Is this getting mapped to `customDimensions` automatic with the java agent?

mkalish (commented):
This is great!  Are there instructions somewhere for setting this up locally to talk to azure?

Main thought is that if it's not a huge lift, worth bringing in that kotlin logging dependency.

prime-router/src/main/resources/jsonTemplate.json
jalbinson: Yup thats where it ends up.

prime-router/src/main/kotlin/common/LoggerExt.kt
jalbinson: Something like that is in the works for my events PR.

Unfortunately it appears that Azure (and `ThreadContext`) only accept a flat String to String map. The workaround there is to serialize a class to JSON and use it as the value. The upside there is with a few extra lines of query in the log explorer it can be extracted out into a new queryable column. The downside is the query is less efficient and requires parsing.

I'm building the PRs off each other so if this is something we want before merging it might make sense to combine the PRs.

prime-router/src/main/kotlin/common/LoggerExt.kt
jalbinson: I'm not against bringing in a library if it does all this work for me 😄

prime-router/src/main/kotlin/azure/SendFunction.kt
```diff
@@ -155,37 +156,44 @@ class SendFunction(private val workflowEngine: WorkflowEngine = WorkflowEngine()
...
-                actionHistory.setActionType(TaskAction.send_error)
-                actionHistory.trackActionResult(msg)
-                if (receiver.customerStatus == CustomerStatus.ACTIVE) {
-                    logger.fatal("${actionHistory.action.actionResult}")
+        withLoggingContext(
+            // TODO: update with data class after event code is merged
+            "report_id" to reportId.toString(),
+            "receiver" to receiver.fullName,
+        ) {
+            return if (nextRetryItems.isEmpty()) {
```
jalbinson, line 164: This is all indentation after adding `withLoggingContext`.

prime-router/src/main/resources/log4j2.xml
```diff
@@ -2,16 +2,7 @@
...
-                <!-- Might be able to use entity references to avoid repeating this pattern over and over. -->
-                <LevelPatternSelector defaultPattern="%d{HH:mm:ss.SSS} [%t][%T] %-5level %logger{36} - %msg - %X%n">
-                  <!-- Add some color to ERROR and WARN -->
-                  <PatternMatch key="ERROR" pattern="%highlight{%d{HH:mm:ss.SSS} [%t][%T] %-5level %logger{36} - %msg}{bright,red}%n" />
-                  <PatternMatch key="WARN" pattern="%highlight{%d{HH:mm:ss.SSS} [%t][%T] %-5level %logger{36} - %msg}{bright,red}%n" />
-                  <!-- Also add a unique string that can trigger PagerDuty  -->
-                  <PatternMatch key="FATAL" pattern="%highlight{%d{HH:mm:ss.SSS} [%t][%T] FATAL-ALERT %logger{36} - %msg}{bright,red}%n" />
-                </LevelPatternSelector>
-           </PatternLayout>
+            <JsonTemplateLayout eventTemplateUri="classpath:jsonTemplate.json" maxStringLength="14000" />
```
mkalish, line 5: Does this layout require a max string length?  It also looks like it might be bigger for azure?
https://github.com/MicrosoftDocs/azure-docs/blob/main/includes/application-insights-limits.md

mkalish (commented):
Changes look good, are there instructions somewhere for connecting the agent?

prime-router/src/main/resources/log4j2.xml
jalbinson: `14000` seemed to be the length suggested here: https://learn.microsoft.com/en-us/azure/spring-apps/structured-app-log

I'm unsure if the same limits apply outside of a Spring app.

jalbinson:
> Changes look good, are there instructions somewhere for connecting the agent?

You can find those here: https://github.com/CDCgov/prime-reportstream/blob/master/prime-router/docs/observability/local-dev-setup.md

arnejduranovic (approved):

PR #13018, opened 2024-01-18
Platform/jessica/12816 original message passthrough

JessicaWNava:
This PR implements the [design of original message passthrough](https://github.com/CDCgov/prime-reportstream/blob/master/prime-router/docs/design/proposals/0024-original-message-passthrough.md), which allows the original message sent by the sender to be what the receiver receives, instead of receiving a translated file.  There are numerous shortcomings (tech debt) in this ticket which will be picked up immediately in #13232 but the core functionality is tested and working. #13232 includes a pretty exhaustive list of what we would like to update.

Test Steps:
1. Send a message, either FHIR or HL7v2, under the` elr-elims` topic. Try this with and without setting `payloadName` in the request header.
3. Verify SFTP folder in build directory contains the expected message, that the message is IDENTICAL 100% to the original message you sent, and that the file name includes payloadName if you provided that in the submission

## Changes
- `Topic` enum updated with `isSendOriginal` property that modifies behavior of the translate function
- Update translate function to consider `isSendOriginal` topic property for business logic
- Updated "database integration test" 

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- #issue 12816


jimmyfagan (commented):
This looks good, it's cool that it doesn't actually take all that much code to get this to work. I left a handful of minor or optional comments.

Can you also link to the design in the description? I don't think I reviewed that, so just want a better idea of what was included in the design vs what was decided on here. In particular, using `topic` for this seems a little limiting in that one sender can't send original to some receivers while sending non-original to other receivers. Also, if it is a `topic`, then it seems like there's no reason for `send-original` senders or receivers to use transforms and maybe those fields should be disallowed for the topic. Anyway, just want to see how much of that was discussed in the design doc. Thanks!

jimmyfagan:
> This looks good, it's cool that it doesn't actually take all that much code to get this to work. I left a handful of minor or optional comments.
> 
> Can you also link to the design in the description? I don't think I reviewed that, so just want a better idea of what was included in the design vs what was decided on here. In particular, using `topic` for this seems a little limiting in that one sender can't send original to some receivers while sending non-original to other receivers. Also, if it is a `topic`, then it seems like there's no reason for `send-original` senders or receivers to use transforms and maybe those fields should be disallowed for the topic. Anyway, just want to see how much of that was discussed in the design doc. Thanks!

Update on this, I talked with Arnej about some of these things, and I'm good with what we have here. I think down the line, we should probably disallow transforms on this topic, but that's not in scope for this PR. Still would be good to link to the design doc in the PR description.

I think most of my other comments still apply, so let me know when this is ready for re-review!

JessicaWNava:
> This looks good, it's cool that it doesn't actually take all that much code to get this to work. I left a handful of minor or optional comments.
> 
> Can you also link to the design in the description? I don't think I reviewed that, so just want a better idea of what was included in the design vs what was decided on here. In particular, using `topic` for this seems a little limiting in that one sender can't send original to some receivers while sending non-original to other receivers. Also, if it is a `topic`, then it seems like there's no reason for `send-original` senders or receivers to use transforms and maybe those fields should be disallowed for the topic. Anyway, just want to see how much of that was discussed in the design doc. Thanks!

Here is the document Jimmy: https://github.com/CDCgov/prime-reportstream/blob/master/prime-router/docs/design/proposals/0024-original-message-passthrough.md

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
```diff
@@ -131,9 +133,29 @@ class FHIRRouter(
...
+                    ) {
+                        finalReceivers.add(receiver)
+                    }
+                }
+            } else {
+                finalReceivers = listOfReceivers.toMutableList()
+            }
+
+            if (finalReceivers.isEmpty()) {
+                ActionHistory(TaskAction.process_warning, true)
```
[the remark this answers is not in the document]
JessicaWNava: Arnej specifically said that we wanted an Action History created for this case.

prime-router/settings/organizations.yml
```diff
@@ -135,6 +135,16 @@
       customerStatus: active
       schemaName: metadata/fhir_transforms/senders/SimpleReport/simple-report-sender-transform
       format: FHIR
+    - name: ignore-fhir-passthrough      # Use this to test fhir passthrough
+      organizationName: ignore
+      topic: send-original
+      customerStatus: active
+      format: FHIR
+    - name: ignore-hl7-passthrough      # Use this to test fhir passthrough
```
JFU-NAVA-PBC, line 143: # Use this to test HL7 passthrough

?

JFU-NAVA-PBC (commented):
it might be good to add content about this new 'topic' in the docs, e.g. route.md

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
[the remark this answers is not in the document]
jimmyfagan: Ok, should that maybe just use the existing `actionHistory` that's passed into this function? Even if we do want to track this in the action history, I'm not sure this is the right way to do that, maybe I'm misunderstanding this, but it seems like we're just creating a new `ActionHistory` object and not doing anything with it.

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
```diff
@@ -135,22 +135,7 @@ class FHIRRouter(
...
-                    }
-                }
-            } else {
-                finalReceivers = listOfReceivers.toMutableList()
-            }
-
-            if (finalReceivers.isEmpty()) {
-                ActionHistory(TaskAction.process_warning, true)
-            }
+            finalReceivers = findReceiversForBundle(message, listOfReceivers, finalReceivers)
```
jimmyfagan, line 138: Sorry for the confusion, but I was actually suggesting combining this logic into the `findReceiversForBundle` function called at line 134. I think we ended up in the current branch having two `findReceiversForBundle` functions, so basically just combining those into one, so that here in `doWork` we only need to get one list of receivers, we don't need `listOfReceivers` and `finalReceivers`.

JFU-NAVA-PBC (approved):
Given this as an important feature asked from external user, it might worth adding comprehensive smoke tests.

Could use a separate ticket if this is not in current AC.

JFU-NAVA-PBC (commented):
Passed all tests, and did the following manual tests:

1. Post FHIR report payload with client = ignore.ignore-fhir-passthrough
2. Post HL7V2 report payload with client = ignore.ignore-hl7-passthrough

here are the received messages at the SFTP folder:

`-rw-r--r--   1 jimfuqian-mac-pro  staff  283865 Jan 23 14:54 -61deb386-f820-43a8-ba3f-0fa095bc6ea2-20240123145401.fhir
-rw-r--r--   1 jimfuqian-mac-pro  staff   23797 Jan 23 15:03 ca-receiver-transform-04073855-a777-4b80-bf28-2846650ae816-20240123150302.hl7`

the contents are the same as originals, the hl7 message name is odd, is it intended?

JFU-NAVA-PBC:
FYI, checked the intermediate data in containers, and noticed that:

1. For FHIR passthrough, for each message POSTed, there are 2 records produced in containers - batch, route, translate, there is 1 record in: none, ready, receive.
2. For HL7 passthrough, for each message POSTed, there is only 1 entries produced in containers - batch, none, ready, receive, route, translate.

Not able to get my head around on this, but just some behaviors to check that are expected.

Screen shots:

2 records in route for receiver=ignore-fhir-passthrough:

image

1 record in route for receiver=ignore-hl7-passthrough:

image

Similar observation for container "translate", etc.

JFU-NAVA-PBC:
Comment on steps:

Since some steps are not applicable to pass through, e.g. batch, none, ..., is it easy to skip them to save on IO? or just put a small record with sufficient envelop info to trigger next step? hope this won't make the UP steps logic messy...

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
JessicaWNava: Oh, the reason these are separate is because Arnej asked for it to be separate because he wanted there to be a separate ActionHistory for it and for it not to be seen as part of the filters. Unless I misunderstood his and my conversation @arnejduranovic ?

JessicaWNava:
> Passed all tests, and did the following manual tests:
> 
> 1. Post FHIR report payload with client = ignore.ignore-fhir-passthrough
> 2. Post HL7V2 report payload with client = ignore.ignore-hl7-passthrough
> 
> here are the received messages at the SFTP folder:
> 
> `-rw-r--r-- 1 jimfuqian-mac-pro staff 283865 Jan 23 14:54 -61deb386-f820-43a8-ba3f-0fa095bc6ea2-20240123145401.fhir -rw-r--r-- 1 jimfuqian-mac-pro staff 23797 Jan 23 15:03 ca-receiver-transform-04073855-a777-4b80-bf28-2846650ae816-20240123150302.hl7`
> 
> the contents are the same as originals, the hl7 message name is odd, is it intended?

What makes the message name odd @JFU-NAVA-PBC ?

jimmyfagan:
I think Jim is saying that the output hl7 message name includes the receiver transform name, which is odd since that transform isn't used in the passthrough case. I agree that it's odd, but I think that can be handled down the line, it is basically expected based on how the names are currently constructed, even if it should eventually be changed.

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
```diff
@@ -244,6 +253,32 @@ class FHIRRouter(
         }
     }
 
+    private fun findReceiversForBundle(
```
arnejduranovic, line 256: I wasn't picturing we would need to do anything with the route step. It seems like this check to make sure the original format matches the receiver's accepted format could be done more cleanly in the translate step?

I think its cleaner if route step just check route filters. If original passthrough is enabled and it doesn't match receiver's specifications then that sounds like a translate step problem since when we are in the route step we don't care about format, just the data.

JessicaWNava:
> I think Jim is saying that the output hl7 message name includes the receiver transform name, which is odd since that transform isn't used in the passthrough case. I agree that it's odd, but I think that can be handled down the line, it is basically expected based on how the names are currently constructed, even if it should eventually be changed.

Got it, thank you Jimmy!

arnejduranovic (changes requested):

JFU-NAVA-PBC:
> > I think Jim is saying that the output hl7 message name includes the receiver transform name, which is odd since that transform isn't used in the passthrough case. I agree that it's odd, but I think that can be handled down the line, it is basically expected based on how the names are currently constructed, even if it should eventually be changed.
> 
> Got it, thank you Jimmy!

yea, that's what I meant ;-), minor stuff

prime-router/src/main/kotlin/Report.kt
```diff
@@ -1651,6 +1651,8 @@ class Report : Logging {
             metadata: Metadata,
             actionHistory: ActionHistory,
             topic: Topic,
+//            originalReportId: UUID,
```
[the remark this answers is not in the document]
JessicaWNava: Because none of it ended up being needed to pass the information through to where it was needed to be used.

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
JessicaWNava: Here is where that decision is being made as to who to send it to, so I think it makes the most sense to do it here. I am not sure how I would do this in the translate step.

prime-router/src/main/kotlin/SubmissionReceiver.kt
```diff
@@ -229,7 +229,15 @@ class TopicReceiver : SubmissionReceiver {
...
 
-        val processEvent = ProcessEvent(Event.EventAction.PROCESS, report.id, options, defaults, routeTo)
+        val processEvent = ProcessEvent(
+            Event.EventAction.PROCESS,
+            report.id,
+            options,
+            defaults,
+            routeTo,
+            null,
+            null
```
arnejduranovic, line 239: Why is this change needed? It appears these two are already defaulted to null?

prime-router/src/main/kotlin/SubmissionReceiver.kt
```diff
@@ -267,13 +275,18 @@ class UniversalPipelineReceiver : SubmissionReceiver {
...
         val report: Report
+        var originalReportFormat: Format = Report.Format.INTERNAL
 
         when (sender.format) {
             Sender.Format.HL7 -> {
                 val messages = HL7Reader(actionLogs).getMessages(content)
                 val isBatch = HL7Reader(actionLogs).isBatch(content, messages.size)
                 // create a Report for this incoming HL7 message to use for tracking in the database
 
+                if (!isBatch) {
```
arnejduranovic, line 286: If it is not batched, why do we explicitly need to set it to Report.Format.HL7? These types of changes should have a comment to explain the why when the why is not clear.

prime-router/src/main/kotlin/SubmissionReceiver.kt
```diff
@@ -341,7 +355,15 @@ class UniversalPipelineReceiver : SubmissionReceiver {
...
         // add task to task table
-        val processEvent = ProcessEvent(eventAction, report.id, options, defaults, routeTo)
+        val processEvent = ProcessEvent(
+            eventAction,
+            report.id,
+            options,
+            defaults,
+            routeTo,
+            null,
+            null
```
arnejduranovic, line 365: Why is this change needed? It appears these two are already defaulted to null?

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
arnejduranovic: Yes the decision who to send it to is done here, and format is not a consideration that the routing step takes into account. It only worries about the configuration of the routing filters.

That said, can't you basically delete all changes in this function and in translate just do something like:

```
if (receiver.format == message.originalReportFormat) {
    getOriginalMessage(message.originalReportId as ReportId, WorkflowEngine())
} else {
    // log err via action log
    // throw exception
}
```

That seems a lot cleaner to me and now the translate function is doing what its supposed to do: make sure the format of the message is in a format the receiver can accept, otherwise throw an error.

prime-router/src/main/kotlin/fhirengine/engine/QueueMessage.kt
```diff
@@ -81,7 +82,16 @@ interface ReportIdentifyingInformation {
     val topic: Topic
 }
 
-abstract class ReportPipelineMessage : ReportIdentifyingInformation, WithDownloadableReport, QueueMessage()
+interface OriginalMessageInformation {
```
arnejduranovic, line 85: I see value in passing the original report ID to every step for purposes of lineage tracking via logs, less so for the original format. How difficult would it be to put both in a class called "OriginalMessageInfo", that way we can expand this class later if we want without having to explicitly alter all these classes every time? We also wouldn't have to explicitly pass originalReportFormat to two functions that don't care about it just to get it to translate, but if we wrap it in a class like OriginalMessageInfo maybe it makes more sense?

arnejduranovic (changes requested):
Please see comments below and also please check tests again.

prime-router/src/main/kotlin/fhirengine/engine/FHIRTranslator.kt
```diff
@@ -68,7 +72,14 @@ class FHIRTranslator(
                 val receiver = settings.findReceiver(message.receiverFullName)
                     ?: throw RuntimeException("Receiver with name ${message.receiverFullName} was not found")
                 actionHistory.trackActionReceiverInfo(receiver.organizationName, receiver.name)
-                val bodyBytes = getByteArrayFromBundle(receiver, bundle)
+
+                val originalMessage = getOriginalMessage(message.reportId, WorkflowEngine())
```
arnejduranovic, line 76: Can we move this inside the if below and change the if to:

`if (message.topic.isSendOriginal) {`

This way, we can save some computation if isSendOriginal is not true

prime-router/src/main/kotlin/fhirengine/engine/FHIRTranslator.kt
```diff
@@ -68,7 +72,14 @@ class FHIRTranslator(
...
+
+                val bodyBytes = if (originalMessage.schemaTopic.isSendOriginal) {
```
arnejduranovic, line 78: Also since we're touching this function, can we update the else below to:

```
throw RuntimeException(
                    "Message was not a FhirTranslateQueueMessage and cannot be processed by FHIRTranslator: $message"
            )
```

prime-router/src/main/kotlin/fhirengine/engine/FHIRTranslator.kt
```diff
@@ -98,6 +109,26 @@ class FHIRTranslator(
         }
     }
 
+    /**
+     * Takes a [reportId] and returns the content of the original message as a ByteArray
+     */
+    internal fun getOriginalMessage(reportId: ReportId, workflowEngine: WorkflowEngine): ReportFile {
```
arnejduranovic, line 115: Both of these helpers should be moved somewhere more general so other services can make use of them as well.

arnejduranovic (changes requested):

prime-router/src/test/kotlin/fhirengine/engine/FhirTranslatorTests.kt
```diff
@@ -142,6 +183,130 @@ class FhirTranslatorTests {
...
+        every { actionHistory.trackCreatedReport(any(), any(), blobInfo = any()) }.returns(Unit)
+        every { actionHistory.trackExistingInputReport(any()) }.returns(Unit)
+        every { actionHistory.trackActionReceiverInfo(any(), any()) }
+            .returns(Unit)
+
+        accessSpy.transact { txn ->
+            engine.run(message, actionLogger, actionHistory, txn)
+        }
+
+        verify(exactly = 1) {
```
arnejduranovic, line 262: I'm not sure what the value of this test is? Seems like we may as well just leave it to the smokeTests at this point.

prime-router/src/test/kotlin/fhirengine/engine/FhirRouterTests.kt
```diff
@@ -789,6 +810,100 @@ class FhirRouterTests {
         }
     }
 
+    @Test
+    fun `success - everything passes, send original`() {
```
arnejduranovic, line 814: We did not change anything with the route function, why does this need a new test? What value does this add?

prime-router/src/main/kotlin/fhirengine/azure/FHIRFunctions.kt
```diff
@@ -24,6 +24,7 @@ import gov.cdc.prime.router.fhirengine.engine.elrTranslationQueueName
 import org.apache.commons.lang3.StringUtils
 import org.apache.logging.log4j.kotlin.Logging
 
+const val sendQueueName = "send"
```
arnejduranovic, line 27: Remove this and move it to `FHIREngine.kt` with the others. Then, import it above with the rest.

prime-router/src/main/kotlin/fhirengine/engine/FHIRTranslator.kt
```diff
@@ -68,36 +71,76 @@ class FHIRTranslator(
...
+                    "Message was not a FhirTranslateQueueMessage and cannot be processed by FHIRTranslator: $message"
                 )
             }
         }
     }
 
+    /**
+     * Takes a [reportId] and returns the content of the original message as a ByteArray
+     */
+    internal fun getOriginalMessage(
```
arnejduranovic, line 124: rename to getOriginalReport since you changed what this returns now. Also, @mkalish , do you think these two helpers would fit better in WorkflowEngine? Or any other ideas?

prime-router/src/main/kotlin/fhirengine/engine/FHIRTranslator.kt
```diff
@@ -68,36 +71,76 @@ class FHIRTranslator(
...
                     ?: throw RuntimeException("Receiver with name ${message.receiverFullName} was not found")
                 actionHistory.trackActionReceiverInfo(receiver.organizationName, receiver.name)
-                val bodyBytes = getByteArrayFromBundle(receiver, bundle)
+
+                var nextAction = Event.EventAction.BATCH
+                var externalName = ""
+                var queueMessage: ReportEventQueueMessage? = null
+                val bodyBytes = if (message.topic.isSendOriginal) {
+                    nextAction = Event.EventAction.SEND
+                    val originalMessage = getOriginalMessage(message.reportId, WorkflowEngine())
```
arnejduranovic, line 80: rename to `originalReport`

prime-router/src/main/kotlin/fhirengine/engine/FHIRTranslator.kt
```diff
@@ -68,36 +71,76 @@ class FHIRTranslator(
...
                 throw RuntimeException(
-                    "Message was not a FhirConvert and cannot be processed: $message"
+                    "Message was not a FhirTranslateQueueMessage and cannot be processed by FHIRTranslator: $message"
                 )
             }
         }
     }
 
+    /**
+     * Takes a [reportId] and returns the content of the original message as a ByteArray
```
arnejduranovic, line 122: Takes a [reportId] and returns the content of the first ancestor as submitted by the sender as a ByteArray

prime-router/src/main/kotlin/fhirengine/engine/FHIRTranslator.kt
```diff
@@ -68,36 +71,76 @@ class FHIRTranslator(
...
+        workflowEngine: WorkflowEngine,
+    ): ReportFile {
+        val rootReportId = findRootReportId(reportId, workflowEngine)
+        return workflowEngine.db.fetchReportFile(rootReportId)
+    }
+
+    /**
+     * Takes a [reportId] and returns the ReportId of the original message that was sent
+     */
+    fun findRootReportId(reportId: ReportId, workflowEngine: WorkflowEngine): ReportId {
```
arnejduranovic, line 135: IntelliJ wants this to be private? Ideally this function, and the one above, would be unit tested.

prime-router/src/main/kotlin/Report.kt
```diff
@@ -1714,11 +1715,16 @@ class Report : Logging {
                     )
                 }
 
+            var blobName = report.name
```
arnejduranovic, line 1718: `val blobName = if (!externalName.isNullOrEmpty()) externalName + report.name else report.name` better?

arnejduranovic (changes requested):

prime-router/src/main/kotlin/fhirengine/engine/FHIRTranslator.kt
arnejduranovic: I will create a follow-up ticket that includes both of these items and link it here. We can move forward without making this change.

prime-router/src/main/kotlin/fhirengine/engine/FHIRTranslator.kt
JessicaWNava: They are unit tested in FhirTranslatorTests

prime-router/settings/organizations.yml
mkalish, line 1: Just confirming that these are just formatting changes and nothing that needs to be pushed to stage/prod?

prime-router/src/main/kotlin/fhirengine/engine/FHIREngine.kt
```diff
@@ -20,6 +20,7 @@ import java.time.OffsetDateTime
 const val elrConvertQueueName = "elr-fhir-convert"
 const val elrRoutingQueueName = "elr-fhir-route"
 const val elrTranslationQueueName = "elr-fhir-translate"
+const val elrSendQueueName = "send"
```
mkalish, line 23: Nit: this is also the queue that covid pipeline uses.  If we're adding a tech debt ticket to split them though can leave this for now

prime-router/src/main/kotlin/fhirengine/engine/FHIRTranslator.kt
```diff
@@ -68,36 +70,75 @@ class FHIRTranslator(
...
-                val bodyBytes = getByteArrayFromBundle(receiver, bundle)
+
+                var nextAction = Event.EventAction.BATCH
+                var externalName: String? = null
+                var queueMessage: ReportEventQueueMessage? = null
+                val bodyBytes = if (message.topic.isSendOriginal) {
+                    nextAction = Event.EventAction.SEND
+                    val originalReport = getOriginalReport(message.reportId)
+                    externalName = originalReport.externalName
+                    queueMessage = ReportEventQueueMessage(nextAction, false, message.reportId, "")
```
mkalish, line 81: Nit: I think this can be simplified by not instantiating this here?  Seems a little unsafe since `message.reportId` is the wrong value right?

prime-router/src/main/kotlin/fhirengine/azure/FHIRFunctions.kt
```diff
@@ -127,7 +128,13 @@ class FHIRFunctions(
         fhirEngine: FHIRTranslator,
         actionHistory: ActionHistory = ActionHistory(TaskAction.translate),
     ) {
-        runFhirEngine(message, dequeueCount, fhirEngine, actionHistory)
+        val messagesToDispatch = runFhirEngine(message, dequeueCount, fhirEngine, actionHistory)
```
mkalish, line 131: Can we add a comment that this only dispatches an event if the message is going through the send original logic?

prime-router/src/test/kotlin/fhirengine/engine/FhirRouterTests.kt
```diff
@@ -102,14 +102,14 @@ data object SampleFilters {
         "%patient.birthDate.exists()",
         "%specimen.type.exists()",
         "(%patient.address.line.exists() or " +
-                "%patient.address.postalCode.exists() or " +
-                "%patient.telecom.exists())",
+            "%patient.address.postalCode.exists() or " +
```
mkalish, line 105: Just confirming that this is just formatting?

prime-router/src/test/kotlin/fhirengine/engine/FhirTranslatorTests.kt
```diff
@@ -68,7 +70,7 @@ class FhirTranslatorTests {
                 Topic.FULL_ELR,
                 CustomerStatus.ACTIVE,
                 ORU_R01_SCHEMA,
-                format = Report.Format.HL7_BATCH,
+                format = Report.Format.HL7,
```
mkalish, line 73: Why did this need to change?

prime-router/src/test/kotlin/fhirengine/azure/FhirFunctionTests.kt
```diff
@@ -732,30 +802,162 @@ class FhirFunctionTests {
...
+
+                val sendReportFile =
+                    DSL.using(txn).select(REPORT_FILE.asterisk())
+                        .from(REPORT_FILE)
+                        .where(REPORT_FILE.REPORT_ID.eq(sendTask!!.reportId))
+                        .fetchOneInto(REPORT_FILE)
+                assertThat(sendReportFile).isNotNull()
+            }
+            // verify we downloaded the original report in the Translate step
+            verify(exactly = 1) {
```
mkalish, line 938: Could we add an additional verification here that we uploaded to the blob store with the correct file contents?

prime-router/src/test/kotlin/fhirengine/azure/FhirFunctionTests.kt
```diff
@@ -732,30 +802,162 @@ class FhirFunctionTests {
...
+            // verify sendMessage did not get called because next action should be Batch and the batch decider function,
+            // which runs on a timer, will pick up the task from the DB instead of being called directly by the
+            // Translate function
             verify(exactly = 0) {
                 queueMock.sendMessage(any(), any())
             }
         }
 
+        @Test
+        fun `test successfully processes a translate message when isSendOriginal is true`() {
```
mkalish, line 840: Nit: maybe also one checking that HL7 works as well?

mkalish (commented):
LGTM!  Can we add the tech debt ticket before merging?

I'm pulling this down and testing it now

mkalish (changes requested):
Something seems a little off with how the filename is created.  Sending a report without `payloadname` resulted in this `oru_r01-base-05063187-9c8f-40db-9ca3-5af652b94eb3-20240202130514.hl7` which is maybe ok since the behavior is undefined here?

However sending one where I did set the `payloadname` looked incorrect to me `elims_special_name-oru_r01-base-04434322-b9d9-4288-97dd-592b5d277c0a-20240202130652.hl7`

mkalish (approved):
:shipit: Decision was made in a huddle to update how the filename is created in the immediate follow up ticket

prime-router/settings/organizations.yml
arnejduranovic: Correct.

prime-router/src/main/kotlin/fhirengine/engine/FHIREngine.kt
arnejduranovic: Yep, see #13232

prime-router/src/main/kotlin/fhirengine/engine/FHIRTranslator.kt
arnejduranovic: Yes wrong value that gets updated below (will fix in #13232)

prime-router/src/test/kotlin/fhirengine/engine/FhirRouterTests.kt
arnejduranovic: Correct.

prime-router/src/test/kotlin/fhirengine/engine/FhirTranslatorTests.kt
arnejduranovic: Will revert these changes, I missed these during cleanup. Good catch!

prime-router/src/test/kotlin/fhirengine/azure/FhirFunctionTests.kt
arnejduranovic: Will revisit this in #13232 and also add smokeTests that verify contents.

prime-router/src/test/kotlin/fhirengine/azure/FhirFunctionTests.kt
arnejduranovic: Will revisit in #13232

arnejduranovic (approved):
Update PR description, see for more details.

Issue #12268, opened 2023-11-16
HL7-FHIR-HL7 ORU-R01 Mapping: Provenance


mkalish:
### User Story

As a Sender of ORU-R01 HL7v2 messages to RS,
I expect my HL7v2 messages get delivered to the intended receiver in either the FHIR or HL7v2 format without losing data

### Description/Use Case

To ensure RS is mapping ORU-R01 messages losslessly, RS needs to map HL7v2 segment to FHIR resource and FHIR resources to HL7v2 segments as described by the [mapping inventory](https://docs.google.com/spreadsheets/d/1gHK6_PFyr7PXns7wLDs0LSLsbjm0x-4bWUu3crXMKMI/edit#gid=0). 

This ticket shall focus on creating HL7v2->FHIR and FHIR->HL7v2 mappings for the following resource(s):
- Provenance
- Provenance-Source
- Provenance-Transformation
- Provenance-Patient

Please reference the aforementioned ORU-R01 mapping inventory for guidance on what segments and resources need to be mapped.

### Dev Notes

- Reuse the datatypes that were created as part of epic #11655
- Mapping decisions or interpretations made due to ambiguity in the mapping inventory should be documented to enable future follow-up with the owning organization

### Acceptance Criteria

- [ ] Mappings created for segments/resource
- [ ] Decision points documented

PR #13102, opened 2024-01-25
Update provenance mapping(s)

jimmyfagan:
This PR implements the mappings for [Provenance-Source](https://docs.google.com/spreadsheets/d/1F5aYk6tFCYTQd_qEaEc5G85ZcCm98R5B-sq2JGqUagk/edit#gid=0), [Provenance-Transformation](https://docs.google.com/spreadsheets/d/1byfzqOfOvIVdRkHv2Tto5a-a0YMYWWP0eryaZBvibIo/edit#gid=0), and [Provenance-Patient](https://docs.google.com/spreadsheets/d/1n_mRDFUD-LI6_fVxNurlTCU53Rn6AfDnE75wPH6hC4I/edit#gid=0). It also removed the existing provenance mapping reference from ORU_R01. Since these Provenance mappings are based from MSH and PID, both of which are fully mapped in HL7v2->FHIR->HL7v2, the Provenance mappings only are needed for the HL7v2-to-FHIR translation, and not the other way around.

No new tests were needed since the existing tests covered all the necessary combinations of fields for these mappings. Every existing test needed new FHIR output to be generated as a result of these changes.

Test Steps:
1. Run the integration tests.

## Changes
- Slight update to the PRT device mapping to be less impacted by non-PRT-devices existing or not
- Cleaned up the MSH.7 hl7 mapping slightly
- Added a schema to add a RS organization, currently in the MSH fhir mapping folder
- Added "recorded" to the ignored FHIR fields for FHIR comparison, since one of these mappings includes using the current time, and that's hard to verify in tests.

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- Addresses #12268


prime-router/metadata/fhir_mapping/hl7/message/ORU_R01.yml
```diff
@@ -8,11 +8,13 @@ resources:
     additionalSegments:
       - SFT
 
-  - resourceName: Provenance
+  - resourceName: Provenance-Source
```
mkalish, line 11: So I think this is suppose to be the name of the FHIR resource... did you hit an issue if they all have the same name?

prime-router/metadata/hl7_mapping/resources/MessageHeader/MSH.yml
```diff
@@ -30,23 +30,20 @@ elements:
     constants:
       hl7HDField: 'MSH-6'
 
-  - name: datetime-of-message
```
mkalish, line 33: Nit: I'm actually wondering if we should keep this one and delete the other two?  It looks like they all come from MSH.7 and this extension will always get set if that is valued.

prime-router/metadata/fhir_mapping/hl7/segments/MSH/Provenance-Source.yml
```diff
@@ -0,0 +1,95 @@
+resourceType: Provenance
+id:
+  type: STRING
+  valueOf: "GeneralUtils.generateResourceId()"
+  expressionType: JEXL
+
+target:
```
mkalish, line 7: It looks like this should also point to the created MessageHeader?

prime-router/metadata/fhir_mapping/hl7/segments/MSH/Provenance-Source.yml
```diff
@@ -0,0 +1,95 @@
...
+  expressionType: nested
+  generateList: true
+  expressionsMap:
+    role:
+      type: STRING
+      valueOf: "source"
+
+# second entity value may be added after a mapping update based on MSH.3
+
+agent_1:
```
mkalish, line 23: Nit: I think we've been trying to collapse these into one schema element using `expressions`.  I think long term it forces keeping items like these close together so that we can verify the conditions remain mutually exclusive

prime-router/metadata/fhir_mapping/hl7/segments/MSH/Provenance-Transformation.yml
```diff
@@ -0,0 +1,57 @@
...
+              type: STRING
+              valueOf: "assembler"
+            system:
+              type: SYSTEM_URL
+              valueOf: "http://terminology.hl7.org/CodeSystem/provenance-participant-type"
+    who:
+      valueOf: segments/MSH/datatypes/HD/ReportStreamOrganization
+      expressionType: reference
+
+entity:
```
mkalish, line 52: I'm wondering if we should leave this off since we're not populating the document reference

prime-router/metadata/fhir_mapping/hl7/segments/MSH/datatypes/HD/ReportStreamOrganization.yml
mkalish, line 1: Nit: wondering if this should be in an `HD` directory?  If it really is an HD we might want to use some of the extensions because it currently wouldn't have enough data to go back into an HD

prime-router/metadata/fhir_mapping/hl7/segments/PID/Provenance-Patient.yml
```diff
@@ -0,0 +1,60 @@
...
+  valueOf: "GeneralUtils.generateResourceId()"
+  expressionType: JEXL
+
+target:
+  valueOf: datatype/Reference
+  generateList: true
+  expressionType: resource
+  specs: $Patient
+
+recorded:
```
mkalish, line 13: Can we add a tech debt ticket that would allow us to set a transform start variable that would be available globally?  I think technically our `recorded` values might differ by a very tiny amount

mkalish (commented):
Great work here!  Just a few small comments, but I think we're missing test cases to verify some of the inventory rules?

mkalish:
@jimmyfagan Also looks like there might be a failing smoke test

prime-router/metadata/hl7_mapping/resources/MessageHeader/MSH.yml
jimmyfagan: Yep that works for HL7 senders, but it would change the requirements of FHIR senders, in particular some of these changes were needed for the OML FHIR sender tests. We can make those changes via transforms and go that route though, do you think that's the way we should go?

prime-router/metadata/fhir_mapping/hl7/segments/MSH/Provenance-Source.yml
jimmyfagan: Good catch, also didn't verify this in the integration tests, but it seems like the bundle reference isn't working either. Will have to investigate and regenerate FHIR again.

prime-router/metadata/fhir_mapping/hl7/segments/MSH/datatypes/HD/ReportStreamOrganization.yml
jimmyfagan: It's not really an HD, was kind of shoehorning it in there since the values were based on an HD compatible transform. I think maybe I'll just move it out of the directory, it's probably a standalone kind of concept.

prime-router/metadata/fhir_mapping/hl7/segments/MSH/Provenance-Source.yml
jimmyfagan: Ah right, I've been falling back onto whatever I was using as a starting point, but good call I'll rework this.

prime-router/metadata/fhir_mapping/hl7/message/ORU_R01.yml
mkalish: So I think we should update it to match the actual resource name just based on the documentation

prime-router/metadata/hl7_mapping/resources/MessageHeader/MSH.yml
mkalish: That's my thought?  Though might not be a huge deal, but the case I was guarding against was MSH.7 not being set in the original HL7 and we then populate it on the way out

prime-router/metadata/fhir_mapping/hl7/segments/MSH/Provenance-Transformation.yml
jimmyfagan: Hmm, yea maybe, the same would probably apply to Provenance-Source, and then I'd need to add an extension to be able to reliably distinguish between the different provenance types. That might be a good idea anyway, the existing fhir path to make that distinction is pretty clunky. I'll work through that.

prime-router/metadata/fhir_mapping/hl7/segments/PID/Provenance-Patient.yml
jimmyfagan: It's only down to second precision, so I haven't been able to hit that, but it does seem possible. I can see about setting the transformation time as a variable at the base transform level and using that in each spot, it might only be a couple lines and save us the ticket.

jimmyfagan:
Noticed the failing smoke test late yesterday, will try to investigate and resolve that along with the other feedback, will re-request a review when it's ready again.

prime-router/metadata/fhir_mapping/hl7/segments/PID/Provenance-Patient.yml
jimmyfagan: Didn't work, so opened a ticket [here](https://app.zenhub.com/workspaces/platform-6182b02547c1130010f459db/issues/gh/cdcgov/prime-reportstream/13134)

prime-router/settings/organizations.yml
```diff
@@ -960,7 +960,6 @@
       customerStatus: active
       jurisdictionalFilter: [ "(%performerState.exists() and %performerState = 'CA') or (%patientState.exists() and %patientState = 'CA')" ]
       qualityFilter:
-        - "(%hl7MessageType.exists() and %hl7MessageType = 'R01')"
```
mkalish, line 963: Does this change need to be made in staging/prod?

mkalish (approved):
Nice work! LGTM!

prime-router/settings/organizations.yml
jimmyfagan: It could go either way, it's just repeated in another line that is part of the same filter below (gets cut off by the github view). Just was cleaning it up while I was in the area, but it's functionally the same either way.

arnejduranovic (approved):

PR #13067, opened 2024-01-23
platform/jimfuqian/10508-Update-all-schemas-references-to-URIs

JFU-NAVA-PBC:
This PR:

Change all the schema reference from file path to URI
 
Test Steps:
1. Pass all tests

## Changes
- Going through all schema files (.yml), do the requested change: i.e. change any schema reference in path syntax to URI - e.g. - schema: classpath:/metadata/hl7_mapping/resources/Location/HD.yml
- Going through all kotlin sources, do the requested change to any schema references
- Copy metadata/hl7_mapping and metadata/fhir_transforms to src/main/resources/metadata to satisfy resource lookup in the URI based resources (schemas) resolution.

### Testing
- [X] Tested locally?
- [X] Ran `./prime test` or `./gradlew testSmoke` against local Docker ReportStream container?
- [ ] Added tests?

### Process
- [ ] Are there licensing issues with any new dependencies introduced?
- [ ] Includes a summary of what a code reviewer should test/verify?
- [ ] Updated the release notes?
- [ ] Database changes are submitted as a separate PR?
- [ ] DevOps team has been notified if PR requires ops support?

## Linked Issues
- Fixes #issue

## To Be Done
*Create GitHub issues to track the work remaining, if any*
- #issue

## Specific Security-related subjects a reviewer should pay specific attention to
- Does this PR introduce new endpoints?
    - new endpoint A
    - new endpoint B
- Does this PR include changes in authentication and/or authorization of existing endpoints?
- Does this change introduce new dependencies that need vetting?
- Does this change require changes to our infrastructure?
- Does logging contain sensitive data?
- Does this PR include or remove any sensitive information itself?

If you answered '_yes_' to any of the questions above, conduct a detailed Review that addresses at least:

- What are the potential security threats and mitigations? Please list the _STRIDE_ threats and how they are mitigated
    - **S**poofing (faking authenticity)
        - Threat _T_, which could be achieved by _A_, is mitigated by _M_
    - **T**ampering (influence or sabotage the integrity of information, data, or system)
    - **R**epudiation (the ability to dispute the origin or originator of an action)
    - **I**nformation disclosure (data made available to entities who should not have it)
    - **D**enial of service (make a resource unavailable)
    - **E**levation of Privilege (reduce restrictions that apply or gain privileges one should not have)
- Have you ensured logging does not contain sensitive data?
- Have you received any additional approvals needed for this change?

PR #13103, opened 2024-01-25
platform/jimfuqian/10508 Change Schema Refs to URI

JFU-NAVA-PBC:
This PR:

Change all the schema reference from file path to URI

This is a re-work of [PR 13067](https://github.com/CDCgov/prime-reportstream/pull/13067) to keep the source structures remain the same.

Test Steps:
1. Pass all tests

## Changes
- Going through all schema files (.yml), do the requested change: i.e. change any schema reference in path syntax to URI - e.g. - schema: classpath:/metadata/hl7_mapping/resources/Location/HD.yml
- Going through all kotlin sources, do the requested change to any schema references
- Changed gradle DSL to properly provision the schemas during build time

### Testing
- [X] Tested locally?
- [ ] Added tests?

### Process

## Linked Issues
- Fixes #issue

## To Be Done
- #issue

    - new endpoint A
    - new endpoint B


prime-router/build.gradle.kts
```diff
@@ -549,6 +565,16 @@ tasks.register("package") {
 tasks.register("quickPackage") {
     group = rootProject.description ?: ""
     description = "Package the code and necessary files to run the Azure functions skipping unit tests and migration"
+//    sourceSets {
```
mkalish, line 568: Remove?

prime-router/src/main/kotlin/fhirengine/translation/hl7/schema/ConfigSchemaReader.kt
```diff
@@ -44,6 +44,29 @@ object ConfigSchemaReader : Logging {
...
+        schemaClass: Class<out ConfigSchema<out ConfigSchemaElement>>,
+    ): ConfigSchema<*> {
+        // Load a schema including any parent schemas.  Note that child schemas are loaded first and the parents last.
+//        val schemaList = when (schemaUri.scheme) {
+//            null -> fromRelative(schemaName, folder, schemaClass)
+//            else -> fromUri(schemaUri, schemaClass)
+//        }
+
+        val schemaList = when (schemaUri.scheme) {
+            "classpath" -> fromUri(schemaUri, schemaClass)
```
mkalish, line 58: I don't think this is right... this means we can only support cases where it's a classpath.  The goal of this is epic is to support many schemes?

prime-router/src/main/kotlin/fhirengine/translation/hl7/schema/ConfigSchemaReader.kt
```diff
@@ -44,6 +44,29 @@ object ConfigSchemaReader : Logging {
         return mergedSchema
     }
 
+    fun fromFile(
+        schemaUri: URI,
+        schemaClass: Class<out ConfigSchema<out ConfigSchemaElement>>,
+    ): ConfigSchema<*> {
+        // Load a schema including any parent schemas.  Note that child schemas are loaded first and the parents last.
```
mkalish, line 51: I'm not following why this change is required?

prime-router/src/main/kotlin/fhirengine/translation/hl7/schema/ConfigSchemaReader.kt
```diff
@@ -134,7 +157,8 @@ object ConfigSchemaReader : Logging {
                 readOneYamlSchema(file.inputStream(), schemaClass)
             }
             "classpath" -> {
-                val input = javaClass.classLoader.getResourceAsStream(schemaUri.path.substring(1))
+                val schemaYml = if (schemaUri.path.endsWith(".yml")) schemaUri.path else schemaUri.path + ".yml"
```
mkalish, line 160: Shouldn't we need this change in all the cases here?  Also, can we just put `.yml` in the schema and avoid this extra branch?

prime-router/src/main/kotlin/fhirengine/translation/hl7/schema/converter/ConverterSchemaReader.kt
```diff
@@ -2,14 +2,28 @@ package gov.cdc.prime.router.fhirengine.translation.hl7.schema.converter
...
+import java.net.URI
 
 /**
  * Read a converter schema [schemaName] from a file given the root [folder].
  * @return the validated schema
  * @throws Exception if the schema is invalid or is of the wrong type
  */
 fun converterSchemaFromFile(schemaName: String, folder: String? = null): ConverterSchema {
-    val schema = ConfigSchemaReader.fromFile(schemaName, folder, schemaClass = ConverterSchema::class.java)
+    val schemaUri = getURI(folder, schemaName)
```
mkalish, line 14: I'm not following this change either.  What case is this handling?

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/SchemaUtils.kt
```diff
@@ -0,0 +1,15 @@
+package gov.cdc.prime.router.fhirengine.translation.hl7.utils
+
+/**
+ * helper
+ */
+fun getURI(folder: String?, schemaName: String): String {
```
mkalish, line 6: I don't think this should be required... the class path should contain all the info about the found and we shouldn't be constructing it dynamically.  I think you're editing the old code path that supports relative pathing.

prime-router/src/test/kotlin/fhirengine/translation/hl7/schema/ConfigSchemaReaderTests.kt
```diff
@@ -89,14 +89,15 @@ class ConfigSchemaReaderTests {
     @Test
     fun `test read schema tree from file`() {
         // This is a good schema
-        val schema = ConfigSchemaReader.readSchemaTreeRelative(
-            "ORU_R01",
-            "src/test/resources/fhirengine/translation/hl7/schema/schema-read-test-01"
+        val schema = ConfigSchemaReader.readSchemaTreeUri(
```
mkalish, line 92: This should still work... the existing code supports both relative and reading by absolute URI.

mkalish (changes requested):
I think closing in here, but still some work to be done.

- I really think we should just use `git mv` to get the mappings files into the classpath, relying on that new gradle task is brittle and unnecessary
- I don't think you need any of the code changes you made.  This ticket is 100% move and update schemas

prime-router/build.gradle.kts
```diff
@@ -537,10 +543,20 @@ tasks.azureFunctionsPackage {
     finalizedBy("copyAzureScripts")
 }
 
+// copy metadata to main resources
```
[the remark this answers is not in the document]
JFU-NAVA-PBC: This is to leverage gradle resources provision practice - copying metadata from source tree to build directory - since we switched to URI (resources) as schema references.

this is to avoid change the current source tree structures:

<local-repo-base>
             |
             |----------> metadata
             |                           |
             |                           |--------> fhir_transforms
             |                           |
             |                           |--------> hl7_mapping
             |                           |-------------> other sub dirs
             | 
             |--------- src
             |                  |
             |                  |----------> main
             |                  |                       |
             |                  |                       |--------> resources
             |                  |                       .......

 the git mv approach is OK, since the metadata hl7_mapping, fhir_transforms will be copied to build area by convention, but the un-desirable is that it break the top level metadata folders original purpose - keep the schemas etc. in one place, now hl7_mapping, fhir_transforms are moved out, RS "metadata" are scattered around in multiple locations in source tree, I think it reduces the managibility.

later on if RS adding more "metadata", where should they go? go to top level metadata dir? or go into src/main/resources/metadata?

prime-router/src/main/kotlin/fhirengine/translation/hl7/schema/ConfigSchemaReader.kt
JFU-NAVA-PBC: how many scheme we are going to support by this PR?

prime-router/build.gradle.kts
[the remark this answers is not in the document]
mkalish: Can we not just put them in a directory called metadata?  I'm not sure anything else has to change.

prime-router/src/main/kotlin/fhirengine/translation/hl7/schema/ConfigSchemaReader.kt
mkalish: The code already supports all the schemes

prime-router/src/main/kotlin/fhirengine/translation/hl7/schema/ConfigSchemaReader.kt
JFU-NAVA-PBC: yea, here I put some heuristic tweaking, the code could be tolerant or restrict, can mandate all schema reference to yml file be explicitly ending with .yml

but even with that required in authoring schema yaml files, do we want to be tolerant (guess a suffix) or throw error e.g. can not read schema: classpath: /.../xyz ??

prime-router/src/main/kotlin/fhirengine/translation/hl7/schema/converter/ConverterSchemaReader.kt
JFU-NAVA-PBC: getURI(folder, schemaName) is an adaptor func I added, it will adapt some existing local file system semantics based schema loading code so that they will work with schemas references in URI format.

I think those schema loading funcs with folder, schemaName parameters - should be deprecated since it is switched to resource URI semantics, but could be done later, for now I just want to make things working.

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/SchemaUtils.kt
JFU-NAVA-PBC: yup, adapting old code with URI semantics....

prime-router/src/test/kotlin/fhirengine/translation/hl7/schema/ConfigSchemaReaderTests.kt
JFU-NAVA-PBC: why still keep the relative path tests since we are switching to schema reference with resource URI?

think some legacy tests either be refactored or replaced with URI based

prime-router/build.gradle.kts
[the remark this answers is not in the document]
JFU-NAVA-PBC: "Can we not just put them in a directory called metadata? I'm not sure anything else has to change."

Can do that.

prime-router/src/main/kotlin/fhirengine/translation/hl7/schema/ConfigSchemaReader.kt
mkalish: My current thinking is that we should insist on the extension, but happy to discuss further.  Just seems clearer, since `classpath: /.../xyz` isn't actually a file and I was thinking we should make folks be explicit

prime-router/src/main/kotlin/fhirengine/translation/hl7/schema/converter/ConverterSchemaReader.kt
mkalish: Agreed, though there is a clean up ticket at the end of this epic that capture that work.  The scope here was to try to just get the existing schemas going through the new code path

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/SchemaUtils.kt
mkalish: See comment above, but the old code _should_ be able to exist along side this new URI pattern, but let me know if that isn't quite working out

prime-router/src/test/kotlin/fhirengine/translation/hl7/schema/ConfigSchemaReaderTests.kt
mkalish: Mostly just to keep the scope of the work smaller and making rollback easier if we need to an update

arnejduranovic:
Closing since another PR solving the same ticket has been approved by codeowners and @JFU-NAVA-PBC : #13231

PR #13192, opened 2024-01-31
12774 - Daily data API updates

### LINKED ISSUE #12774: API work for Daily Data search feature
## Problem statement
The Daily Data search feature will require some backend API work to ensure it works properly. This is the ticket for the backend API work.

## What you need to know
- We need to make changes to the API on the Daily Data page to handle the new reportId or fileName parameters.
- Check with Platform (perhaps Michael) if we need to allow only reportID or fileName, OR if we can allow both
- Check with Platform if we should be creating a new API with a new version

## Acceptance criteria
- [x] Update Swagger
- [x] Update API Unit Test cases 
- [x] API will have two new optional parameters (reportId and fileName)
- [ ] Test in Production to ensure time is not too long of a response

penny-lischer:
This PR ...

Updates the deliveries endpoint to allow for reportId and fileName as optional parameters.

Test Steps:
1. Within a HTTP client you can now call the following:
/api/waters/org/ca-dph.elr/deliveries?fileName=
/api/waters/org/ca-dph.elr/deliveries?reportId=

## Changes
- Added reportId and fileName optional parameters
Screenshot 2024-02-06 at 9 38 00 AM

Screenshot 2024-02-06 at 9 38 24 AM

Screenshot 2024-02-06 at 9 38 33 AM

Screenshot 2024-02-06 at 9 38 57 AM

Screenshot 2024-02-06 at 9 39 19 AM

Screenshot 2024-02-09 at 9 37 36 AM

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- Fixes #12774


prime-router/src/main/kotlin/history/azure/DeliveryFacade.kt
```diff
@@ -53,6 +57,7 @@ class DeliveryFacade(
         require(since == null || until == null || until > since) {
             "End date must be after start date."
         }
+        val reportId = if (reportIdStr != null) UUID.fromString(reportIdStr) else null
```
mkalish, line 60: You might want to put this in a try/catch since `UUID.fromString` will throw if it's not a valid UUID

prime-router/src/main/kotlin/history/azure/HistoryDatabaseAccess.kt
```diff
@@ -4,6 +4,7 @@ import gov.cdc.prime.router.azure.DatabaseAccess
 import gov.cdc.prime.router.azure.db.Tables.ACTION
 import gov.cdc.prime.router.azure.db.Tables.REPORT_FILE
 import gov.cdc.prime.router.common.BaseEngine
+import org.apache.commons.lang3.compare.ComparableUtils.ge
```
mkalish, line 7: I don't think this is used?

prime-router/src/main/kotlin/history/azure/HistoryDatabaseAccess.kt
```diff
@@ -136,12 +139,22 @@ abstract class HistoryDatabaseAccess(
...
         orgService: String?,
+        reportId: UUID?,
+        fileName: String?,
         since: OffsetDateTime?,
         until: OffsetDateTime?,
         showFailed: Boolean,
     ): Condition {
         var filter = this.organizationFilter(organization, orgService)
 
+        if (reportId != null) {
```
mkalish, line 150: Optional, but this does still feel weird since these are mutually exclusive?  I could see up dating the HTTP function to check the parameters and return a `422` if both are provided.

Otherwise, I just think we need really good documentation on what the outcome would be if you provide both

mkalish (commented):
Looking good!

prime-router/src/main/kotlin/history/azure/HistoryDatabaseAccess.kt
penny-lischer: Agree, should this be checked in the ReportFileFunction.kt -> getListByOrg() method or here?

mkalish: My two cents would be close to the Function itself to make it easier to generate the correct HTTP response

jpandersen87 (approved):

victor-chaparro (approved):
Looks good! Tested it locally and works as expected.

prime-router/src/main/kotlin/history/azure/DeliveryFacade.kt
```diff
@@ -54,6 +62,13 @@ class DeliveryFacade(
             "End date must be after start date."
         }
 
+        var reportId: UUID?
+        try {
+            reportId = if (reportIdStr != null) UUID.fromString(reportIdStr) else null
+        } catch (e: IllegalArgumentException) {
+            throw HttpException("Invalid format for report ID: $reportIdStr", HttpStatus.UNPROCESSABLE_ENTITY)
```
mkalish, line 69: I think doing something like `return HttpUtilities.bad(request, Invalid format for report ID: $reportIdStr)` is more common (I think it's fine to return bad response rather 422, but you could also add a new helper)

prime-router/src/test/kotlin/history/azure/DeliveryFacadeTests.kt
mkalish, line 1: Maybe one test that documents the behavior that a date filter will still be in effect even if searching by id?

mkalish (commented):
Looks great!  Couple of small things, but nothing blocking from me

prime-router/src/main/kotlin/history/azure/DeliveryFacade.kt
penny-lischer: That won't work here since the return type of this method is a List and not a response.

penny-lischer: I updated the logic. Let me know if this is ok.

prime-router/src/main/kotlin/history/azure/DeliveryFacade.kt
```diff
@@ -54,6 +59,13 @@ class DeliveryFacade(
             "End date must be after start date."
         }
 
+        var reportId: UUID? = null
+        try {
+            reportId = if (reportIdStr != null) UUID.fromString(reportIdStr) else null
```
mkalish, line 64: This doesn't seem quite to me?  At least testing locally, if I pass in an invalid UUID I now just get all the results?

penny-lischer: fixed

mkalish: Nice!  Just tested it out and I got a 400 message with the invalid ID in it

mkalish (approved):
:shipit:

PR #13249, opened 2024-02-05
13160: create report route event

### LINKED ISSUE #13160: Metrics Dashboard MvP Backend: Part 1
### User Story

As ReportStream,
I need to provide various product-level and application performance metrics,
so that stakeholders can better understand the system and its operation.

### Description/Use Case

This design should identify how the backend should handle collecting, storing, and surfacing the data required to answer the following questions outlined in the FY24 OKR Metrics Dashboard Requirements:

- Number of STLTs receiving two or more conditions from ReportStream

Also keep in mind the Observability requirements as a whole and make sure our MvP approach does not lock us out of supporting the additional questions outline therein:  [observability requirements document](https://cdc.sharepoint.com/:x:/r/teams/ReportStream/_layouts/15/Doc.aspx?sourcedoc=%7B50A50DD5-3353-43D3-89E6-9E3D6E7DD0D8%7D&file=Observability%20Product%20Requirements.xlsx&nav=MTVfezYzNTdENzdBLTQ0RjMtMTE0Mi05RjlCLTFEMTFEQ0JDMEQwN30&action=default&mobileredirect=true).

- These metrics should be able to be retrieved from somewhere in RS Azure App Insights
- We should leverage Azure logs/metrics to power these queries.

### Acceptance Criteria

- [ ] Azure App Insights Graph updated to answer the questions outlined above

jalbinson:
This PR adds a new Azure custom event for routed reports. It includes necessary fields to answer questions on what conditions are going where.

Test Steps:
1. Send data to staging that you know will have conditions routed to a receiver
2. Query the `customEvents` table in Azure to find the emitted event

## Changes
- Added new event
- Added a way to test events
- Updated dependencies on `FHIREngine` to accept the `AzureEventService` so we can emit events at various stages

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- Fixes #13160


prime-router/src/main/kotlin/azure/observability/event/ReportRouteEvent.kt
```diff
@@ -0,0 +1,17 @@
+package gov.cdc.prime.router.azure.observability.event
+
+import gov.cdc.prime.router.ReportId
+import gov.cdc.prime.router.Topic
+
+/**
+ * Event definition for when a report gets routed to a receiver
+ */
+data class ReportRouteEvent(
```
jalbinson, line 9: The actual important part of this PR!

Do we need any additional fields here for better future-proofing?

arnejduranovic (changes requested):

prime-router/src/main/kotlin/fhirengine/engine/FHIREngine.kt
```diff
@@ -95,13 +107,15 @@ abstract class FHIREngine(
...
+                    blobAccess ?: BlobAccess(),
+                    azureEventService ?: AzureEventServiceImpl()
                 )
                 TaskAction.route -> FHIRRouter(
                     metadata ?: Metadata.getInstance(),
                     settingsProvider!!,
                     databaseAccess ?: databaseAccessSingleton,
-                    blobAccess ?: BlobAccess()
+                    blobAccess ?: BlobAccess(),
+                    azureEventService ?: AzureEventServiceImpl()
```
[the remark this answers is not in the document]
jalbinson: Sounds good.

jalbinson:
@arnejduranovic @brick-green @mkalish 

Adding in a bunch of additional fields for future use is fine with me. The only potential issue I see us running into is the size limits of what AppInsights can injest.

https://github.com/MicrosoftDocs/azure-docs/blob/main/includes/application-insights-limits.md
Max size of telemetry item: 64 KB
key length: 150
value length: 8192

If we put as many fields at the top level as possible it shouldn't be a problem but just dropping in an object that serializes to json could potentially blow up if the object is really large. We would have to run some tests with common reports to see what kind of sizes we would be working with.

prime-router/src/main/kotlin/azure/observability/event/ReportRouteEvent.kt
jalbinson: @arnejduranovic @brick-green @mkalish 

Adding in a bunch of additional fields for future use is fine with me. The only potential issue I see us running into is the size limits of what AppInsights can injest.

https://github.com/MicrosoftDocs/azure-docs/blob/main/includes/application-insights-limits.md
Max size of telemetry item: 64 KB
key length: 150
value length: 8192

If we put as many fields at the top level as possible it shouldn't be a problem but just dropping in an object that serializes to json could potentially blow up if the object is really large. We would have to run some tests with common reports to see what kind of sizes we would be working with.

arnejduranovic: If we serialize an object to JSON, would be still be able to query based on the keys in it? I guess I assumed we could but I'm not sure

arnejduranovic: Also, this ticket doesn't need to add all the fields, but it should set up the architecture so we have the ability to add more very easily in the future without having to refactor.

jalbinson: > If we serialize an object to JSON, would be still be able to query based on the keys in it? I guess I assumed we could but I'm not sure. Seems like something we should verify as well.

We can query it but there needs to be an additional step in the query to parse that json. It could slow down the overall query so my recommendation is to keep fields we expect to query a lot at the top level.

arnejduranovic: I would like to see what format this message is (hl7 vs FHIR) and then also the value of field MSH.21. When we add this, would we just add more params here? That doesn't seem maintainable to me. Sounds like we may want to use a builder pattern here so we don't have to change the constructor all the time? https://www.baeldung.com/kotlin/builder-pattern

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
```diff
@@ -261,6 +271,9 @@ class FHIRRouter(
             // ensure tracking is set
             actionHistory.trackCreatedReport(nextEvent, report)
 
+            // send event to Azure AppInsights
+            emitAzureEvent(report, message, null, bundle)
```
arnejduranovic, line 275: Interesting, so this means if receiver is null the report didn't go anywhere. I think that's probably a fine way to communicate that.

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
```diff
@@ -694,4 +707,24 @@ class FHIRRouter(
...
+        report: Report,
+        message: ReportPipelineMessage,
+        receiver: Receiver?,
+        bundle: Bundle,
+    ) {
+        val sender = reportService.getSenderName(message.reportId)
+        val conditions = bundle.getAllMappedConditions()
+
+        azureEventService.trackEvent(
+            ReportRouteEvent(
```
arnejduranovic, line 721: If we used builder pattern, we could just call this `UniversalPipelineEvent` instead, have a field for "type" (route, convert, etc), and then just set the params that make sense for that particular case. We wouldn't have to update all the usages everywhere. What do you think about that? Any other advantages or disadvantages? Perhaps this is something we just wait to refactor if we find we need to.

jalbinson: I think events should be different enough that making a single class with a builder might end up too large and confusing.

I also think the compiler errors when we update an event definition are beneficial! We don't want to add a new field and possibly miss a spot and then end up emitting bad events for months unknowingly.

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
jalbinson: That was my thought. We could also omit sending this event entirely if we are only interested in truly "routed" reports. It would be easy enough to add a line in the query though to filter these ones out.

arnejduranovic (approved):
SmokeTests: Passed
IntegrationTests: Passed

I checked out the branch locally, connected to app insights test, and verified the correct events and correct number of events made it up and that all data was correct. I also updated my passthrough cleanup code to use the new getRootReport function and all functionality on my branch continued to work, including my own getRootReport unit test I wrote. 

Great work!

PR #13253, opened 2024-02-05
Refactored mapped condition filter; added fhir expression + condition lookup filters

### LINKED ISSUE #12705: Combine mappedConditionFilter with legacy conditionFilter
### User Story
As a dev, I want to avoid creating variants of existing filters
to keep the codebase consolidated condition filters consolidated 

### Description/Use Case
During #12279 we added a new type of condition filter for filtering observations based on stamped condition codes. To avoid

### LINKED ISSUE #12706: Implement condition lookup filter for condition-based routing
### User Story
As a person configuring ReportStream, I want to filter Observations by a condition name so that configuring condition filters remains scalable and terse.

### Description/Use Case
The current implementation of mapped condition filtering relies on configuring lists of condition codes per receiver. We would like to add a new type of filter that looks up a list of codes (in a new lookup table) given a condition name.

- May want to reconfigure to leverage the new filter

### Dev Notes
- Consider inheriting from CodeStringConditionFilter and loading the CodeString in the constructor

### Acceptance Criteria
- New condition filter implemented supporting a condition name

thetaurean:
This PR introduces a new design pattern for filters in ReportStream and adds legacy `conditionFilter` support to the recently added mappedConditionFilter (which has been refactored to use the new design pattern). This adds two new types of filter: `FHIRExpressionConditionFilter` and `ConditionLookupConditionFilter`.

Test Steps:
1. Run full test suite
2. Manually test a legacy condition filter
3. Manually test a mapped condition filter
4. Manually test a combination of 2 & 3

## Changes
- Refactor abstract `ConditionFilter` and related implementations
- Update `FHIRRouter` to utilize new filter design pattern properly
- Added `FHIRExpressionConditionFilter` and `ConditionLookupConditionFilter`

### Testing
- [x] Tested locally?
- [ ] Added tests? One still needs fixed and a couple more added

## Linked Issues
- Fixes #12705 
- Fixes #12706 

## To Be Done
Before Merge / Review
- Address any outstanding TODOs
- Check for any commented tests to be fixed

Tech Debt
- Refactor rest of filters (create ticket)

Deployment Strategy
- Code: Add temporary no-op on mappedConditionFilter
  - **DEPLOY** Prime can now handle the new `mappedConditionFilter` but is not using them
- Config: Duplicate production `conditionFilter`s as `mappedConditionFilter`s
- Code: Remove legacy `conditionFilter` code and temporary no-op
  - **DEPLOY** Prime is now using only the new condition filters but with the `mapped` moniker
- Config: Configure production conditionFilter as duplicates of mappedConditionFilter
- Code: Adjust implementation to use `conditionFilter` instead of `mappedConditionFilter`
  - **DEPLOY** Prime is now using the new condition filters without the `mapped` moniker
- Config: Remove `mappedConditionFilter` from production config
- Code: Remove `mappedConditionFilter` from codebase


prime-router/src/main/kotlin/ReportStreamFilter.kt
```diff
@@ -11,7 +22,37 @@ import kotlin.reflect.full.memberProperties
...
-fun ReportStreamConditionFilter.codes(): List<String> = this.flatMap { it.codes() }
+data class ConditionFilterResult(
+    val pass: Boolean = false,
+    val log: List<ActionLogDetail> = emptyList(),
+    val fails: Map<String, List<Observation>> = emptyMap(),
+)
+
+fun ConditionFilterResult.and(other: ConditionFilterResult) = this.andOr(other, true)
+fun ConditionFilterResult.or(other: ConditionFilterResult) = this.andOr(other)
+fun ConditionFilterResult.andOr(other: ConditionFilterResult, useOr: Boolean = true) = ConditionFilterResult(
```
mkalish, line 33: Should this be private?  The default param does not reflect what this does

prime-router/src/main/kotlin/ReportStreamFilter.kt
```diff
@@ -11,7 +22,37 @@ import kotlin.reflect.full.memberProperties
...
 typealias ReportStreamConditionFilter = List<ConditionFilter>
 
-fun ReportStreamConditionFilter.codes(): List<String> = this.flatMap { it.codes() }
+data class ConditionFilterResult(
+    val pass: Boolean = false,
+    val log: List<ActionLogDetail> = emptyList(),
+    val fails: Map<String, List<Observation>> = emptyMap(),
+)
+
+fun ConditionFilterResult.and(other: ConditionFilterResult) = this.andOr(other, true)
```
mkalish, line 31: Any reason these need to be extensions vs just being add to the class?

prime-router/src/main/kotlin/ReportStreamFilter.kt
```diff
@@ -20,21 +61,183 @@ fun ReportStreamConditionFilter.codes(): List<String> = this.flatMap { it.codes(
...
 abstract class ConditionFilter(val value: String) {
-    // TODO: change to .filter() or .passes() method and combine with legacy filter
-    abstract fun codes(): List<String>
+    abstract fun evaluate(
+        bundle: Bundle,
+        filter: Boolean = true,
+    ): ConditionFilterResult
+}
+
+abstract class ObservationConditionFilter(value: String) : ConditionFilter(value) {
```
mkalish, line 70: The naming here is getting a little confusing to me, what other kind of `*ConditionFilter` are you envisioning?

prime-router/src/main/kotlin/ReportStreamFilter.kt
```diff
@@ -20,21 +61,183 @@ fun ReportStreamConditionFilter.codes(): List<String> = this.flatMap { it.codes(
...
+            val obsResult = evaluateObservation(bundle, observation)
+            if (obsResult.pass) {
+                passedObservations.add(observation)
+            } else if (filter) {
+                bundle.deleteResource(observation)
+            }
+            result.or(obsResult)
+        }
+        if (result.pass && passedObservations.all { it.getMappedConditions().all { code -> code == "AOE" } }) {
+            // TODO: all AOE add appropriate message
```
mkalish, line 85: TODO here

mkalish (changes requested):
Super neat stuff, I really like the direction where this is heading.  Have some comments, let me know if you want to chat about any of them

- Missing a lot of documentation; I'm liking the changes but I think KDoc would be really helpful
- I think a lot of places the default params can get dropped
- I'm not sure bundling in the removing of observations into the "filter" behavior makes sense (and realistically is a problem that exists before this change).  The code uses the filter to mean two different things
    - Filter the list of the receivers for a topic; for the condition case that means a receiver is filtered if none of the condition filters evaluate to true
    - For an individual receiver, the bundle is then filtered such that the a observation is only included if any of the condition filters evaluate to true
I think these distinctions need to be better surfaced; for example the idea of a "not" condition filter might be difficult to configure correctly to work for both of the cases
- I'd like to see a lot more unit tests around all the new condition code

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
```diff
@@ -373,12 +369,11 @@ class FHIRRouter(
             // TODO: merge with condition filter (see https://github.com/CDCgov/prime-reportstream/issues/12705)
```
JFU-NAVA-PBC, line 369: can be removed?

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
```diff
@@ -170,7 +167,7 @@ class FHIRRouter(
                 // TODO: merge with mapped condition filter (see https://github.com/CDCgov/prime-reportstream/issues/12705)
                 // If the receiver does not have a condition filter set send the entire bundle to the translate step
                 var receiverBundle = if (receiver.conditionFilter.isEmpty()) {
-                    bundle
+                    bundle.copy()
```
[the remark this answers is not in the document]
JFU-NAVA-PBC: +++

JFU-NAVA-PBC (commented):
it might be good to add tests for new code coverage

JFU-NAVA-PBC (commented):
1. passed all tests locally
2. I saw migration of existing settings on deployed ENVs w.r.t this filter changes being planned
3. Intro of abstract and open classes on condition filter look neat and could help potential future impl of condition filtering of different flavors

Nothing to add w.r.t coding style as Micheal already covered that.

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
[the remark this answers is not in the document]
thetaurean: The bundle must get copied before any modification to prevent filters bleeding across receivers. But looking at it again, this should move to the code block that evaluates the `mappedConditionFilter` (and only happen if that filter will run).

I'll make the update.

prime-router/src/main/kotlin/ReportStreamFilter.kt
```diff
@@ -11,7 +22,37 @@ import kotlin.reflect.full.memberProperties
...
+            resultMap[it.key] = resultMap.getOrDefault(it.key, emptyList()) + it.value
+        }
+    }
+)
+
+fun List<ConditionFilter>.evaluate(
+    bundle: Bundle,
+    filter: Boolean = true,
+    useOr: Boolean = true,
+    perResult: (
```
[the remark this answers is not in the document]
thetaurean: It was being used but not anymore following a refactor. I left it in but will remove.

prime-router/src/main/kotlin/ReportStreamFilter.kt
thetaurean: I don't envision any others. As this pattern extends to the rest of the filters I was envisioning `ConditionFilter` becomes `Filter` or `ReportStreamFilter` and `ObservationConditionFilter` becomes `ConditionFilter` or `ObservationFilter`.

The intent was to provide a reusable `evaluate` loop where child classes override `evaluateObservation`.

I think `ObservationFilter` makes the most sense given the above? And then rename the base abstract class after legacy is gone.

prime-router/src/main/kotlin/ReportStreamFilter.kt
thetaurean: I wasn't planning on adding any logging here (there isn't in current master) but wanted to double-check.
@arnejduranovic do you think we should log if bundle doesn't get sent to a receiver due to all AOE observations?

We don't currently log anything of the sort when deciding what receivers to route to. Only when pruning.

prime-router/src/main/kotlin/fhirengine/engine/FHIRRouter.kt
thetaurean: Should probably leave it in until the deployment process finishes and the old code is removed.

thetaurean:
> 1. passed all tests locally
> 
>     2. I saw migration of existing settings on deployed ENVs w.r.t this filter changes being planned
> 
>     3. Intro of abstract and open classes on condition filter look neat and could help potential future impl of condition filtering of different flavors
> 
> 
> Nothing to add w.r.t coding style as Micheal already covered that.

I'll ping you (and @mkalish) once tests changes and code comments are up.

thetaurean:
> Super neat stuff, I really like the direction where this is heading. Have some comments, let me know if you want to chat about any of them
> 
>     * Missing a lot of documentation; I'm liking the changes but I think KDoc would be really helpful

Heard and addressing.

>     * I think a lot of places the default params can get dropped

👍 
 
>     * I'm not sure bundling in the removing of observations into the "filter" behavior makes sense (and realistically is a problem that exists before this change).  The code uses the filter to mean two different things

In a previous version of this I had two abstract methods: `evaluate` (for qualifying receivers) and `filter` (for filtering the bundle). I'm not against splitting them, but implementation ends up being the same algorithm just returning different data (and noop for the filter).
1. Run all these filters and let me know if any of them passed
2. Run all these filters and let me know what observations failed which filters

>       * Filter the list of the receivers for a topic; for the condition case that means a receiver is filtered if none of the condition filters evaluate to true
>       * For an individual receiver, the bundle is then filtered such that the a observation is only included if any of the condition filters evaluate to true
>         I think these distinctions need to be better surfaced; for example the idea of a "not" condition filter might be difficult to configure correctly to work for both of the cases

Could a wrapper class or a an added class member (val not: Boolean = false) achieve this effectively?

> 
>     * I'd like to see a lot more unit tests around all the new condition code

Wrapping up, thanks for reviewing what's up :).

.environment/pre-commit.runner.sh
```diff
@@ -14,8 +14,8 @@ pushd "${REPO_ROOT?}" 1>/dev/null 2>&1
 CHECKS_TO_RUN=(
     #${REPO_ROOT}/.environment/gitleaks/run-gitleaks.sh
     ${REPO_ROOT}/.environment/ktlint/run-ktlintCheck.sh
-    ${REPO_ROOT}/.environment/terraform-fmt/run-terraform-fmt.sh
-    ${REPO_ROOT}/.environment/frontend/run-frontend.sh
+    #${REPO_ROOT}/.environment/terraform-fmt/run-terraform-fmt.sh
+    #${REPO_ROOT}/.environment/frontend/run-frontend.sh
```
snesm, line 18: Please restore these test.  Both should short-circuit and exit quickly if you have no relevant changes.

snesm (changes requested):

.environment/pre-commit.runner.sh
thetaurean: Thanks for the callout! Didnt intend to leave these committed they were temporary following a merge.

arnejduranovic (commented):
Great work! My main question is this: Should we wait to push this and make even more changes in order to simplify and refactor the way routing works even more. My thought is that we would want to do this eventually and I don't like disrupting engagement multiple times. We should come up with something stable we think will be flexible and maintainable enough going forward for all filters.

prime-router/src/main/kotlin/ReportStreamFilter.kt
```diff
@@ -2,39 +2,206 @@ package gov.cdc.prime.router
...
+) : BundleResourceFilter<Observation>(observationFilter)
+
+class BundleConditionFilter(
+    val conditionFilter: BundlePrunable<Observation>,
+) : BundleResourceFilter<Observation>(conditionFilter) {
+    override fun pass(bundle: Bundle): Boolean =
+        conditionFilter.fetchResources(bundle).filter { observation ->
+            conditionFilter.evaluateResource(bundle, observation)
+        }.let { observations ->
+            // never pass a bundle with only AOE conditions
```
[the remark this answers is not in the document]
thetaurean: This has been reworked so that the AOE check is no longer necessary.

The business logic is that we want to include AOEs if we found a condition we're interested in. So instead of configuring them in the filter as a code that passes (which results in the edge case of a bundle with all AOEs), the filter logic adds AOEs only if it found a condition that passes.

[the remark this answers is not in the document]
arnejduranovic: Nice! And if a receiver for some reason does not want any or certain AOEs, we can have a separate "pruner" for that.

prime-router/settings/organizations.yml
```diff
@@ -1087,9 +1087,11 @@
       qualityFilter:
         - "true"
       processingModeFilter: [ "true" ]
-      mappedConditionFilter: # this filter will only pass observations stamped with condition codes AOE or 840539006
-        - type: codeString
-          value: 840539006,AOE
+      observationFilter: # this filter will only pass observations stamped with condition codes AOE or 840539006
```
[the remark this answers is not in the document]
thetaurean: > Have we looked into integrating something like https://pkl-lang.org/index.html for development-time settings checking?

This looks extra neat.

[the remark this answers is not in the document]
thetaurean: I feel like there should be a way to denest but the only path I found was a custom deserializer. I think there's possibly a way to do it with annotations.

With that said I think it's a lot of work to hide something that shouldn't be hidden.

Perhaps a name update `resourceFilter` - > `resourceEvaluator`
```
        - action:
          type: BundleResourceFilter
          resourceEvaluator:
            type: conditionCodePruner
            codes: 840539006,1234
```
A BundlePrunable evaluates resources and prunes them from the bundle.
A BundleResourceFilter wraps a BundlePrunable and uses it to evaluate pass/fail logic.

De-nesting means losing jackson's subtype object mapping (or hacking it back into the custom deserializer) and hiding what's happening under the hood.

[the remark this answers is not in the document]
thetaurean: Some other thoughts:

- I have a version of this code that has `Filterable<T>` (resulting in `Filterable<Bundle>` and `Filterable<Observation>`) and does away with the prune concept. But currently evaluating a fhir expression is **tightly** coupled to a bundle, even if it's been evaluated against an observation. Decoupling that requires us to refactor the `CustomContext` class to not require a Bundle.

- We don't need to handle wrapping the pruner at the config level. We just need to know whether to evaluate the the pruner as a pruner or as part of a resource filter. Currently pruners are always evaluated as filters AND run as pruners, so we could also make that assumption. But I like the idea that you could filter, then prune, then filter again based on the config.

PR #13308, opened 2024-02-08
Add hospital receiver to Flexion org settings for results flow

basiliskus:
This PR adds the `simulated-hospital` receiver settings for the TI results use case as described in [TI's issue #641](https://github.com/CDCgov/trusted-intermediary/issues/641), along with test ORU files for the translation test. It also fixes the looping issue with TI by filtering using the `ETOR` meta tag and adds the routing filter for the receivers

### Steps to test Results:
1. Run `./prime multiple-settings set --input ./settings/staging/0166-flexion-staging-results-handling.yml`
2. Choose an ORU message to send to `flexion.simulated-lab`. You may use `prime-router/src/testIntegration/resources/datatests/HL7_to_FHIR/etor_ORU_20240220.hl7` as a sample ORU
4. Change the Receiving Facility ID (`MSH-6.2`) in the ORM to `simulated-hsopital-id`
5. Send the ORU to the waters endpoint:
    `curl --header 'Content-Type: application/hl7-v2' --header 'Client: flexion.simulated-lab' --header 'Authorization: Bearer token' --data-binary '@/path/to/etor_ORU_20240220.hl7' 'http://localhost:7071/api/waters'`
4. Verify the request went through without any errors and uploaded to the SFTP for the hospital. The delivered `ORU` message should map everything in the original `ORU` with the exception of `OBX.4` and `OBX.29`

### Steps to test Orders:
1. Make sure `./settings/staging/0166-flexion-staging-results-handling.yml` has been applied locally
2. Choose an ORM message to send to `flexion.simulated-hospital`. You may use `prime-router/src/testIntegration/resources/datatests/HL7_to_FHIR/sample_orm_20230809-001.hl7` as a sample ORM
4. Change the Receiving Facility ID (`MSH-6.2`) in the ORM to `simulated-lab-id`
5. Send the ORM to the waters endpoint:
    `curl --header 'Content-Type: application/hl7-v2' --header 'Client: flexion.simulated-hospital' --header 'Authorization: Bearer token' --data-binary '@/path/to/sample_orm_20230809-001.hl7' 'http://localhost:7071/api/waters'`
6. Verify the request went through without any errors and uploaded to the SFTP for the lab. The message should be delivered only once

### Steps to test final destination routing
1. Change `MSH-6.2` in the message to anything other than what's instructed above and send the message
2. The message should not be delivered to its final destination

### Looping issue
If you send a FHIR message as `etor-service-sender` without the `ETOR` meta tag, the message should be sent back to TI instead of routed to deliver to destination. You could use this [ORU FHIR message](https://github.com/CDCgov/trusted-intermediary/blob/main/examples/MN/004_MN_ORU_R01_NBS_1_hl7_translation.fhir) that doesn't have the ETOR tag 

### Testing
- [x] Tested locally?
- [x] Ran `./gradlew testSmoke`
- [x] Ran `./gradlew testIntegration`


prime-router/settings/staging/0166-flexion-staging-results-handling.yml
```diff
@@ -146,3 +149,30 @@
...
+    # RS => Hospital: translates ORU_R01 from FHIR to HL7 and uploads the message to the SFTP folder
+    - name: "simulated-hospital"
+      externalName: "Simulated Hospital ETOR Receiver - Receiving ORU_R01 HL7 results"
+      organizationName: "flexion"
+      topic: "etor-ti"
+      customerStatus: "active"
+      jurisdictionalFilter:
+        - "Bundle.entry.resource.ofType(MessageHeader).event.code = 'R01'" # ORU_R01
+        - "Bundle.entry.resource.ofType(MessageHeader).meta.tag.where(system = 'http://localcodes.org/ETOR').code = 'ETOR'"
+        # - "Bundle.entry.resource.ofType(MessageHeader).destination.receiver.resolve().identifier.where(value = 'simulated-hospital').exists()"
```
saquino0827, line 161: Do we know which receiver we're filtering for? Does this need to stay commented until we know?

basiliskus: Hmmm I was planning to leave it commented just as an example that we can follow for actual partner settings, but it may be a good idea to use it. `simulated-hospital` and `simulated-hospital` will only be used by us for testing purposes, so we can decide on an id that we'll use in the `MSH-6.2` for our testing messages

basiliskus: I uncommented the filters. When sending test messages to staging we'll now need to use `simulated-hospital-id` in `MSH-6.2` when sending results and `simulated-lab-id` in `MSH-6.2` when sending orders

oslynn:
@basiliskus please add testing steps and FHIR OML message that we can send to flexion.simulated-lab.  Therefore, we can test 
 the looping issue with TI by filtering using the ETOR meta tag and adding the routing filter for the receivers.

oslynn (approved):

prime-router/src/main/resources/metadata/hl7_mapping/receivers/Flexion/TIResults.yml
```diff
@@ -0,0 +1 @@
+extends: classpath:/metadata/hl7_mapping/ORU_R01/ORU_R01-base.yml
```
victor-chaparro, line 1: Why is this file needed, if all it's doing is extending  ORU_R01-base? Why not use use ORU_R01 in the receiver settings?

basiliskus: For now it's not used but we may need it in the future in case our ORU mapping diverges. It's following the convention we have for Orders `prime-router/src/main/resources/metadata/hl7_mapping/receivers/Flexion/TILabOrder.yml`. Having said that, I don't mind removing this for now and adding it back once we need it

victor-chaparro (approved):
Approving, but

PR #13316, opened 2024-02-09
14415: clean up exception logging for hl7 parsing

### LINKED ISSUE #13315: Reduce noisy expected exceptions for hl7 parsing to reduce PagerDuty alerts
## Describe the bug
Reduce logging of hl7 parsing exceptions when they are expected.

## Impact on ReportStream
Our alerting rules look at logged exceptions so we are continuing to falsely alert that there are issues.

## Expected behavior
Exceptions that trigger alerts are unexpected.

jalbinson:
This PR stops logging the expected exception during HL7 parsing to reduce PagerDuty alerts.

I also took the liberty of cleaning up the logging logic to use the log4j `Level` class rather than depending on some logging class found in an SSH library.

Test Steps:
1. See that parsing errors are no longer ending up in the exceptions table in Azure

## Changes
- Remove exception from logger call
- Use `Level` class from log4j

### Testing
- [ ] Tested locally?
- [ ] Added tests?

## Linked Issues
- Fixes #13315


mkalish (approved):
:shipit:

PR #13390, opened 2024-02-15
Experience/13388-App-and-related-componentservice-simplification-and-mock-cleanup

### LINKED ISSUE #13388: App and related component/service simplification and mock cleanup
Parent #12426

## Description
Some of our components and services can be simplified to require less custom code and scaffolding, as well as cleaning up mocking in these areas.

## Impact
- Time savings for FE engineers (auto-mocking rather than wrapping things, with custom function, with the necessary library code). This eliminates from having to run the custom function in a bunch of different areas/tests.

## Acceptance Criteria
- [ ] App is in control of rendering at root, not react-router
- [ ] AppInsights customization is based off officially provided classes and patterns so as not to stray if/when we can drop customizations
- [ ] Objects used for creating mock objects moved to fixture-named files/folders
- [ ] Start usage of global automocks for common libraries/local modules

jpandersen87:
Fixes #13388

- App once again at root of rendering (takes config and routes and creates everything needed and renders)
- Bulk of App logic moved to SessionContext
- App/SessionContext no longer need split subcomponents
- Configs for AppInsights/Okta moved to app config
- Context hooks moved to dedicated files from Providers
- Start of moving objects intended to aid in mocking other objects to fixture files/folders
- ApplicationInsights, Okta React, and Session Context are always automocked
- Start of removing context components from test render wrapper
- telemetry service moved to utils
- tests updated accordingly with new locations and mocks


penny-lischer (approved):
LGTM

etanb (approved):

etanb:
If we're checking in all these 4 spaced files, we should check in some lint config that enforces this as well.

Issue #12750, opened 2023-12-26
Investigate Unit Testing FHIR String Value


thetaurean:
### User Story
As a developer, I want isolate unit tests with invalid data sane from code they are not testing so that they are testing what they intend to.

### Description/Use Case
During implementation of #12275 we added a try/catch on extension function `Observation.getCodeSourcesMap()` to resolve a weird testing issue where `Observation.code.coding` and `Observation.valueCodeableConcept.coding` were strings, leading to type errors.

Investigate the underlying cause and determine if something needs to be fixed.

### Dev Notes
Remove the try/catch from the extension function mentioned (see `FHIRBundleHelpers.kt`) and track down the error that occurs when running smoke tests.

### Acceptance Criteria
- The try/catch is no longer necessary and removed -OR- the reason for the `coding` resolving to a String is documented.

PR #13398, opened 2024-02-16
12750: only map conditions for Observation.value if its CodeableConcept

mkalish:
This PR updates the condition mapping code to only map `Observation.code` if it's not null and `Observation.value` if it's a CodeableConcept.

Test Steps:
1. Apply the following diff
```
diff --git a/prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt b/prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
index 3f4fe3d91..74cb81882 100644
--- a/prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
+++ b/prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
@@ -16,7 +16,6 @@ import gov.cdc.prime.router.fhirengine.utils.FHIRBundleHelpers.Companion.getChil
 import io.github.linuxforhealth.hl7.data.Hl7RelatedGeneralUtils
 import org.hl7.fhir.r4.model.Base
 import org.hl7.fhir.r4.model.Bundle
-import org.hl7.fhir.r4.model.CodeableConcept
 import org.hl7.fhir.r4.model.Coding
 import org.hl7.fhir.r4.model.DateTimeType
 import org.hl7.fhir.r4.model.DiagnosticReport
@@ -74,9 +73,10 @@ fun Observation.getCodeSourcesMap(): Map<String, List<Coding>> {
         toReturn[ObservationMappingConstants.BUNDLE_CODE_IDENTIFIER] = this.code.coding
     }
 
-    if (this.value is CodeableConcept) {
-        toReturn[ObservationMappingConstants.BUNDLE_VALUE_IDENTIFIER] = this.valueCodeableConcept.coding
-    }
+    toReturn[ObservationMappingConstants.BUNDLE_VALUE_IDENTIFIER] = this.valueCodeableConcept.coding
+//    if (this.value is CodeableConcept) {
+//        toReturn[ObservationMappingConstants.BUNDLE_VALUE_IDENTIFIER] = this.valueCodeableConcept.coding
+//    }
 
     return toReturn
 }
```
and then run the `addMappedCondition supports processing values other than codeable concept` test.  It should fail.
2. Revert the patch and the test should now pass.

## Changes
- Only try to map the condition code for Observation.code if code is not null
- Only try to map the condition code for Observation.value is value is a CodeableConcept

### Testing
- [ ] Tested locally?
- [ ] Added tests?

### Process

## Linked Issues
- #12750

### DIFF
--- a/prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
+++ b/prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
@@ -14,9 +14,9 @@ import gov.cdc.prime.router.fhirengine.translation.hl7.utils.CustomContext
 import gov.cdc.prime.router.fhirengine.translation.hl7.utils.FhirPathUtils
 import gov.cdc.prime.router.fhirengine.utils.FHIRBundleHelpers.Companion.getChildProperties
 import io.github.linuxforhealth.hl7.data.Hl7RelatedGeneralUtils
-import org.hl7.fhir.exceptions.FHIRException
 import org.hl7.fhir.r4.model.Base
 import org.hl7.fhir.r4.model.Bundle
+import org.hl7.fhir.r4.model.CodeableConcept
 import org.hl7.fhir.r4.model.Coding
 import org.hl7.fhir.r4.model.DateTimeType
 import org.hl7.fhir.r4.model.DiagnosticReport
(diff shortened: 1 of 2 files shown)


prime-router/src/test/kotlin/fhirengine/utils/FHIRBundleHelpersTests.kt
```diff
@@ -782,4 +783,90 @@ class FHIRBundleHelpersTests {
...
+        code.addCoding(Coding("system", "80382-5", "display"))
+        entry.setCode(code)
+
+        entry.setValue(StringType("A string value"))
+
+        entry.addMappedCondition(metadata)
+
+        val extension = code.coding.first().extension.first()
+        assertThat(extension.url).isEqualTo(conditionCodeExtensionURL)
+        assertThat((extension.value as? Coding)?.code).isEqualTo("6142004")
```
jalbinson, line 828: You can chain these assertions to have better output if it were to fail.

```kotlin
assertThat(extension.value)
  .isInstanceOf<Coding>()
  .transform { it.code }
  .isEqualTo("6142004")
```

If `extension.value` changes its type you would get a better error message than what would currently happen which is saying something like `null does not equal "6142004"`.

jalbinson (approved):
Looks good to me besides a unit test suggestion.

prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
```diff
@@ -67,24 +67,17 @@ private fun lookupCondition(code: Coding, metadata: Metadata): Coding? {
  */
 fun Observation.getCodeSourcesMap(): Map<String, List<Coding>> {
     val toReturn = mutableMapOf<String, List<Coding>>()
```
jalbinson, line 69: not crazy about the variable name `toReturn`. Maybe `codeSourcesMap` since thats what were getting?

prime-router/src/test/kotlin/fhirengine/utils/FHIRBundleHelpersTests.kt
mkalish: Oh that's neat, will change that

prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
mkalish: Agreed, will update the name

Issue #13020, opened 2024-01-18
Update the HL7Converter to support loading different schema mappings


mkalish:
### User Story
As a sender, I would like to ReportStream to have the flexibility to support converting multiple HL7 conformance profiles

### Description/Use Case
Currently, when an HL7 message is translated, the translator object is just a [singleton](https://github.com/CDCgov/prime-reportstream/blob/b2aaee0cff36f6e7bb1a448fa67cbee122c5ad8c/prime-router/src/main/kotlin/fhirengine/engine/FHIRConverter.kt#L203) with it's configuration hardcoded.  In order to support multiple HL7 specs, the codebase will need to create instances of the translator with different properties, specifically where the mapping schemas for that profile are loaded.

Each new instance should load the default datatypes in addition into a profile specific ones as this will substantially reduce the amount of duplicated files.

### Dev Notes
- [This](https://github.com/CDCgov/prime-reportstream/blob/b2aaee0cff36f6e7bb1a448fa67cbee122c5ad8c/prime-router/src/main/kotlin/fhirengine/translation/HL7toFhirTranslator.kt#L26) is the init logic.  As part of this work it would be great to also switch to loading them out of the classpath
- The HL7toFhirTranslator constructor should also load the "default" set of datatypes, the `additional.resources.location` property can likely be used

### Acceptance Criteria
- [ ] The schemas are loaded out of the classpath (or if not feasible a follow up ticket created to handle that)
- [ ] A `HL7toFhirTranslator` object can be created with a specified location for it's schema files as well as the default datatypes
- [ ] The current HL7 mapping schemas are moved into a new directory `catchall`
- [ ] The the `FhirConverter` code creates a `HL7toFhirTranslator` with the `catchall` schemas loaded

Andrey-Glazkv:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/13020?planning-poker) with Zenhub @brick-green

PR #13457, opened 2024-02-22
13020 load schema mapping support

jack-h-wang:
This PR introduces the ability to dynamically configure the location of HL7 to FHIR conversion schemas based on the message type and profile.

This should be reviewed with https://github.com/CDCgov/hl7v2-fhir-converter/pull/8

Test Steps:
1. Run Integration tests and confirm they pass.

## Changes
- Added data structures to support dynamic schema mapping by MSH9/MSH21
- Refactored HL7toFhirTranslator to not rely on a singleton instance
- HL7toFhirTranslator instances can now be constructed with a configuration file path

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- #13020

### DIFF
--- a/prime-router/build.gradle.kts
+++ b/prime-router/build.gradle.kts
@@ -852,7 +852,7 @@ dependencies {
     implementation("org.yaml:snakeyaml:2.2")
     implementation("io.github.linuxforhealth:hl7v2-fhir-converter") {
         version {
-            branch = "master"
+            branch = "platform/jwang/13020-runtime-constructors"
         }
     }
     implementation("ca.uhn.hapi.fhir:hapi-fhir-structures-r4:6.10.4")
--- a/prime-router/build.gradle.kts
+++ b/prime-router/build.gradle.kts
@@ -852,7 +852,7 @@ dependencies {
     implementation("org.yaml:snakeyaml:2.2")
     implementation("io.github.linuxforhealth:hl7v2-fhir-converter") {
         version {
-            branch = "platform/jwang/13020-runtime-constructors"
+            branch = "master"
         }
     }
     implementation("ca.uhn.hapi.fhir:hapi-fhir-structures-r4:6.10.4")
--- a/prime-router/metadata/test_fhir_mapping/config.properties
+++ b/prime-router/metadata/test_fhir_mapping/config.properties
@@ -1,0 +1,5 @@
+base.path.resource=./metadata/test_fhir_mapping
+supported.hl7.messages=ORU_R01,ORM_O01
+default.zoneid=+00:00
+additional.conceptmap.file=
+additional.resources.location=./metadata/fhir_mapping
--- a/prime-router/metadata/test_fhir_mapping/config.properties
+++ b/prime-router/metadata/test_fhir_mapping/config.properties
@@ -2,4 +2,4 @@ base.path.resource=./metadata/test_fhir_mapping
 supported.hl7.messages=ORU_R01,ORM_O01
 default.zoneid=+00:00
 additional.conceptmap.file=
-additional.resources.location=./metadata/fhir_mapping
+additional.resources.location=
--- a/prime-router/metadata/test_fhir_mapping/hl7/message/ORU_R01.yml
+++ b/prime-router/metadata/test_fhir_mapping/hl7/message/ORU_R01.yml
@@ -1,0 +1,12 @@
+resources:
+
+  - resourceName: MessageHeader
+    segment: MSH
+    resourcePath: segments/MSH/MessageHeader
+    repeats: false
+    isReferenced: true
+    additionalSegments:
+      - SFT
+
+
+# This test should only process MSH to produce minimal results
--- a/prime-router/src/main/kotlin/cli/ProcessFhirCommands.kt
+++ b/prime-router/src/main/kotlin/cli/ProcessFhirCommands.kt
@@ -234,7 +234,7 @@ class ProcessFhirCommands : CliktCommand(
             val msh = message.get("MSH") as Segment
             Terser.set(msh, 2, 0, 1, 1, "^~\\&#")
         }
-        return Pair(HL7toFhirTranslator.getInstance().translate(message), message)
+        return Pair(HL7toFhirTranslator().translate(message), message)
     }
 
     /**
--- a/prime-router/src/main/kotlin/cli/ProcessFhirCommands.kt
+++ b/prime-router/src/main/kotlin/cli/ProcessFhirCommands.kt
@@ -236,7 +236,7 @@ class ProcessFhirCommands : CliktCommand(
         }
         val hl7profile = HL7Reader.getMessageProfile(message.toString())
         // search hl7 profile map and create translator with config path if found
-        return when (val configPath = HL7Reader.messageProfileMap[hl7profile]) {
+        return when (val configPath = HL7Reader.profileDirectoryMap[hl7profile]) {
             null -> Pair(HL7toFhirTranslator().translate(message), message)
             else -> Pair(HL7toFhirTranslator(configPath).translate(message), message)
         }
--- a/prime-router/src/main/kotlin/cli/ProcessFhirCommands.kt
+++ b/prime-router/src/main/kotlin/cli/ProcessFhirCommands.kt
@@ -234,7 +234,12 @@ class ProcessFhirCommands : CliktCommand(
             val msh = message.get("MSH") as Segment
             Terser.set(msh, 2, 0, 1, 1, "^~\\&#")
         }
-        return Pair(HL7toFhirTranslator().translate(message), message)
+        val hl7profile = HL7Reader.getMessageProfile(message.toString())
+        // search hl7 profile map and create translator with config path if found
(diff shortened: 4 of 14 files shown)


prime-router/build.gradle.kts
```diff
@@ -852,7 +852,7 @@ dependencies {
     implementation("org.yaml:snakeyaml:2.2")
     implementation("io.github.linuxforhealth:hl7v2-fhir-converter") {
         version {
-            branch = "master"
+            branch = "platform/jwang/13020-runtime-constructors"
```
jack-h-wang, line 855: Flip this back to master when changes to the HL7 to FHIR converter have been merged

prime-router/metadata/test_fhir_mapping/config.properties
```diff
@@ -0,0 +1,5 @@
+base.path.resource=./metadata/test_fhir_mapping
+supported.hl7.messages=ORU_R01,ORM_O01
+default.zoneid=+00:00
+additional.conceptmap.file=
+additional.resources.location=./metadata/fhir_mapping
```
jack-h-wang, line 5: Test configuration references the original location for additional resources, so none need to be declared here

prime-router/metadata/test_fhir_mapping/hl7/message/ORU_R01.yml
```diff
@@ -0,0 +1,12 @@
...
+  - resourceName: MessageHeader
+    segment: MSH
+    resourcePath: segments/MSH/MessageHeader
+    repeats: false
+    isReferenced: true
+    additionalSegments:
+      - SFT
+
+
+# This test should only process MSH to produce minimal results
```
jack-h-wang, line 12: This only differs from the normal mappings in that MSH is the only segment to be processed, which results in a much smaller output file.

prime-router/src/main/kotlin/cli/ProcessFhirCommands.kt
```diff
@@ -234,7 +234,12 @@ class ProcessFhirCommands : CliktCommand(
             val msh = message.get("MSH") as Segment
             Terser.set(msh, 2, 0, 1, 1, "^~\\&#")
         }
-        return Pair(HL7toFhirTranslator.getInstance().translate(message), message)
+        val hl7profile = HL7Reader.getMessageProfile(message.toString())
```
jack-h-wang, line 237: Support for additional schema mapping added to CLI commands here

prime-router/src/main/kotlin/fhirengine/engine/FHIRConverter.kt
```diff
@@ -204,7 +204,9 @@ class FHIRConverter(
         // create the hl7 reader
         val hl7Reader = HL7Reader(actionLogger)
         // get the hl7 from the blob store
-        val hl7messages = hl7Reader.getMessages(queueMessage.downloadContent())
+        val hl7rawmessages = queueMessage.downloadContent()
+        val hl7profile = HL7Reader.getMessageProfile(hl7rawmessages)
```
jack-h-wang, line 208: Support for additional schema mapping added to queue processing here

prime-router/src/main/kotlin/fhirengine/translation/HL7toFhirTranslator.kt
```diff
@@ -16,36 +17,16 @@ import org.hl7.fhir.r4.model.Bundle
 /**
  * Translate an HL7 message to FHIR.
  */
-class HL7toFhirTranslator internal constructor(
+class HL7toFhirTranslator(
+    private val configFolderPath: String = "./metadata/fhir_mapping",
```
jack-h-wang, line 21: Instead of a singleton instance that is used globally, the caller should now create a HL7toFhirTranslator instance to call the translate function. There should be no change in behavior if constructed without parameters.

prime-router/src/main/kotlin/fhirengine/translation/HL7toFhirTranslator.kt
```diff
@@ -56,7 +37,18 @@ class HL7toFhirTranslator internal constructor(
         hl7Message: Message,
     ): HL7MessageModel {
         val messageTemplateType = getMessageTemplateType(hl7Message)
-        return defaultMessageTemplates[messageTemplateType]
+
+        // if the requested message templates have been previously loaded, return the stored templates.
+        // otherwise, load the templates
+        val messageTemplate = if (messageTemplates.containsKey(configFolderPath)) {
```
jack-h-wang, line 43: Templates are stored for performance reasons; otherwise integration test runtime is greatly extended while new instances of ResourceReader load the templates over and over again

prime-router/src/main/kotlin/fhirengine/translation/HL7toFhirTranslator.kt
```diff
@@ -77,6 +69,9 @@ class HL7toFhirTranslator internal constructor(
         // https://github.com/LinuxForHealth/hl7v2-fhir-converter/blob/d5e43fffa96654e7c5bc896e020ff2fa8aac4ff2/src/main/java/io/github/linuxforhealth/hl7/HL7ToFHIRConverter.java#L135-L159
         // If timezone specification is needed it can be provided via a custom HL7MessageEngine with a custom FHIRContext that has the time zone ID set
 
+        // reinitialize the ResourceReader to ensure the correct singleton instance for the job is loaded
+        ResourceReader(ConverterConfiguration(configFolderPath))
```
jack-h-wang, line 73: In a number of places processing switches from an local instance of ResourceReader to the singleton; reinitializing here to make sure the two are equivalent.

prime-router/src/main/kotlin/fhirengine/utils/HL7Reader.kt
```diff
@@ -196,6 +196,12 @@ class HL7Reader(private val actionLogger: ActionLogger) : Logging {
     }
 
     companion object {
+        // map of HL7 message profiles: maps name of profile to configuration directory path
+        val messageProfileMap: Map<MessageProfile, String> = emptyMap()
```
jack-h-wang, line 200: The mapping of message profile to configuration path is defined here. This is intentionally left empty for now.

prime-router/src/test/kotlin/fhirengine/engine/FhirConverterTests.kt
```diff
@@ -270,6 +271,36 @@ class FhirConverterTests {
...
+        )
+
+        every { message.downloadContent() }
+            .returns(validHl7)
+
+        val result = engine.getContentFromHL7(message, actionLogger)
+
+        // the test fhir mappings produce far fewer entries than the production ones
+        assertThat(result).isNotEmpty()
+        assertThat(result[0].entry.size).isEqualTo(5)
```
jack-h-wang, line 301: Open to suggestions to make this test less brittle. This currently relies on the test mappings producing a specific number of entries.

prime-router/src/main/kotlin/fhirengine/translation/HL7toFhirTranslator.kt
mkalish: This seems a bit off?  Is there a unit test that would fail if I uncommented this line out?

prime-router/src/main/kotlin/fhirengine/utils/HL7Reader.kt
mkalish, line 200: Nit: maybe update the var name to reflect it's pointing at a directory?

prime-router/src/test/kotlin/fhirengine/engine/FhirConverterTests.kt
```diff
@@ -270,6 +272,38 @@ class FhirConverterTests {
...
+
+        mockkClass(HL7Reader::class)
+        mockkObject(HL7Reader.Companion)
+        every { HL7Reader.Companion.messageProfileMap.get(testProfile) } returns "./metadata/test_fhir_mapping"
+
+        val result2 = engine.getContentFromHL7(message, actionLogger)
+
+        // the test fhir mappings produce far fewer entries than the production ones
+        assertThat(result2).isNotEmpty()
+        assertThat(result2[0].entry.size).isLessThan(result[0].entry.size)
```
mkalish, line 304: I think we can go further with these assertions and actually compare the contents

prime-router/src/test/kotlin/fhirengine/utils/HL7ReaderTests.kt
```diff
@@ -271,6 +271,20 @@ OBX|1|test|94558-4^SARS-CoV-2 (COVID-19) Ag [Presence] in Respiratory specimen b
         assertThat(type).isEqualTo("ORU")
     }
 
+    @Test
+    fun `test getMessageProfile`() {
```
mkalish, line 275: Can we add tests to cover cases when profile is null?

mkalish (changes requested):
Really good stuff!  Nothing major from me, just want to see a few more tests.

The only meaningful comment I had was on that initialization of `ResourceReader`

mkalish (approved):
Test updates look great!

PR #13378, opened 2024-02-14
13136: metrics part 2

jalbinson:
This PR adds a new event `ReportAcceptedEvent` which is emitted when a report is ready to be broken down by receiver.

Test Steps:
1. Run the application locally with the application insights jar and properly set up environment variable
2. Run the smoke tests
3. See events in the Application Insights dashboard

## Changes
- New event
- Added parent report ID to other event to facilitate better querying

### Testing
- [x] Tested locally?
- [x] Ran `./prime test` or `./gradlew testSmoke` against local Docker ReportStream container?
- [x] Added tests?

## Linked Issues
- Fixes #13136

PR #13470, opened 2024-02-22
13136: add human readable condition display to azure events

jalbinson:
This PR adds human readable conditions to Azure events rather than just the condition code.

Test Steps:
1. Run the project with Azure AppInsights jar setup
2. Run the smoke tests
3. Check Azure logs to see the new condition display present

## Changes
- Added `ConditionSummary` data class
- Added `AzureEventUtils` to handle packaging up Conditions into a form Azure can handle
- Refactored some bundle helper functions

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- Fixes #13378


jalbinson:
Prod how it is now with only codes
Screenshot 2024-02-22 at 3 13 45 PM

Test with condition display strings
Screenshot 2024-02-22 at 3 14 37 PM

jalbinson:
It took me a bit to figure out how to handle dynamic types in KQL. For a quick rundown of the query:

- Only grab `ReportAcceptedEvents`
- Take the dynamic type `customDimensions` and grab `conditions` off of it 
- Convert that to a string before converting it back into a dynamic/json type (important!) 
- Use the `mv-expand` command to break apart the array into 1 row per entry in the array ([see docs for more info](https://learn.microsoft.com/en-us/azure/data-explorer/kusto/query/mv-expand-operator))
- Add a new column with only the `display` field on the `conditions` object
- Run a summarize count operation by your final display field

This will give you a fancy chart of total number of conditions flowing into the system!

mkalish (approved):
:shipit: Saw the events with the display
Screenshot 2024-02-22 at 4 02 52 PM

prime-router/src/main/kotlin/azure/observability/event/AzureEventUtils.kt
```diff
@@ -0,0 +1,20 @@
...
+object AzureEventUtils {
+
+    /**
+     * Retrieves all mapped conditions from a [Bundle] and converts them to a [ConditionSummary]
+     */
+    fun getConditions(bundle: Bundle): Set<ConditionSummary> {
+        return bundle
+            .getAllMappedConditions()
+            .map(ConditionSummary::fromCoding)
+            .toSet()
```
jalbinson, line 18: Don't use a `Set` here

PR #13581, opened 2024-03-04
Experience/12978/filter status

etanb:
This PR ...

Adds filter status functionality which includes:

- Always showing the _number_ of results even if there are 0
- Showing the receiver
- If selected, show the date in `mm:dd:yy` format
- If selected, show the selected time, as it appears in the dropdown, in `hh:mm a' format

Demo of features:

https://github.com/CDCgov/prime-reportstream/assets/516877/e7394ef0-a74f-41b4-b4bb-873279d1329a


audreykwr:
- [ ] Update from hyphen to an en dash to separate date and time ranges (I do not think this was noted in the design, so that is on me. But I noticed when testing)

Screenshot 2024-03-06 at 11 17 36 AM

jpandersen87 (approved):

frontend-react/src/hooks/filters/UseDateRange.ts
```diff
@@ -61,13 +61,13 @@ const FALLBACK_FROM_STRING = "01/01/2000";
 const FALLBACK_FROM = new Date(FALLBACK_FROM_STRING).toISOString();
 const FALLBACK_TO_STRING = "01/01/3000";
 const FALLBACK_TO = new Date(FALLBACK_TO_STRING).toISOString();
-const DEFAULT_TIME = "0:0";
+const DEFAULT_FROM_TIME = "0:0";
+const DEFAULT_TO_TIME = "23:59";
```
penny-lischer, line 65: Do we need to include the seconds and milliseconds to be accurate?

etanb: Nope, this is just a default string. `date-fns` doesn't care if it's `0:0` or `00:00` for midnight.

penny-lischer (approved):
lgtm

PR #13593, opened 2024-03-06
Experience/13480/deliveries receiver status update

### LINKED ISSUE #13480: Add Receiver Status (active or inactive) to the API return call to ensure only active receivers will be shown on Daily Data
## Problem statement

## What you need to know

## Acceptance criteria

penny-lischer:
This PR ...

Add Receiver Status parameter to the API return call to allow for only active receivers on Daily Data

Test Steps:

1. In HTTP client append the `receivingOrgSvcStatus` param to the deliveries call i.e
`/api/waters/org/ca-dph/deliveries?sortDir=DESC&sortColumn=batchReadyAt&since=2023-11-30T00:00:00.000Z&until=2023-12-31T00:00:00.000Z&receivingOrgSvcStatus=active`

## Changes
Added optional receivingOrgSvcStatus parameter to the deliveries API so that the daily data page can display only active and testing records.

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- Fixes #13480

### DIFF
--- a/frontend-react/yarn.lock
+++ b/frontend-react/yarn.lock
@@ -7985,10 +7985,10 @@ __metadata:
   languageName: node
   linkType: hard
 
-"dompurify@npm:^3.0.8":
-  version: 3.0.8
-  resolution: "dompurify@npm:3.0.8"
-  checksum: cac660ccae15a9603f06a85344da868a4c3732d8b57f7998de0f421eb4b9e67d916be52e9bb2a57b2f95b49e994cc50bcd06bb87f2cb2849cf058bdf15266237
+"dompurify@npm:^3.0.9":
+  version: 3.0.9
+  resolution: "dompurify@npm:3.0.9"
+  checksum: 09794f2e40a0003d36e3cd70be1036f83de39d9d4a0f199c45e0b9962ce8ae5f9fe424d7edc4c8ff9ec615be9744a140d66788f0925913ac482d4628b283cae5
   languageName: node
   linkType: hard
 
@@ -14566,7 +14566,7 @@ __metadata:
     cross-env: ^7.0.3
     date-fns: ^3.3.1
     date-fns-tz: ^2.0.0
-    dompurify: ^3.0.8
+    dompurify: ^3.0.9
     dotenv-flow: ^4.1.0
     downloadjs: ^1.4.7
     eslint: ^8.57.0
--- a/frontend-react/yarn.lock
+++ b/frontend-react/yarn.lock
@@ -7985,10 +7985,10 @@ __metadata:
   languageName: node
   linkType: hard
 
-"dompurify@npm:^3.0.9":
-  version: 3.0.9
-  resolution: "dompurify@npm:3.0.9"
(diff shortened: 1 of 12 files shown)


victor-chaparro (approved):
LGTM

prime-router/docs/api/waters-delivery.yml
```diff
@@ -30,6 +30,22 @@ paths:
...
+          in: query
+          description: the fileName within the deliveries of the organization and client
+          schema:
+            type: string
+        - name: receivingOrgSvcStatus
+          in: query
+          description: "'active' or 'inactive' status of the service used by the organization. If no status is provided then all services will be returned."
+          example: "active"
+          schema:
+            type: string
```
mkalish, line 48: Nit: could you `enum: [active, inactive]` here to tighten this up

prime-router/src/main/kotlin/history/azure/HistoryDatabaseAccess.kt
```diff
@@ -72,23 +75,38 @@ abstract class HistoryDatabaseAccess(
...
+                organization, orgService, receivingOrgSvcStatus, reportId, fileName, since, until, showFailed
+            )
 
         return db.transactReturning { txn ->
             val query = DSL.using(txn)
                 // Note the report file and action tables have columns with the same name, so we must specify what we need.
                 .select(
                     ACTION.ACTION_ID, ACTION.CREATED_AT, ACTION.SENDING_ORG, ACTION.SENDING_ORG_CLIENT,
                     REPORT_FILE.RECEIVING_ORG, REPORT_FILE.RECEIVING_ORG_SVC,
+                    DSL.jsonbGetAttributeAsText(SETTING.VALUES, "customerStatus").`as`("receiving_org_svc_status"),
```
mkalish, line 92: Nit: does this value need to get returned?

prime-router/src/main/kotlin/history/azure/HistoryDatabaseAccess.kt
```diff
@@ -72,23 +75,38 @@ abstract class HistoryDatabaseAccess(
         klass: Class<T>,
         reportId: UUID? = null,
         fileName: String? = null,
+        receivingOrgSvcStatus: String? = null,
```
mkalish, line 78: Not a blocker, but what do you think about switching this to the enum to help catch any typo errors?

mkalish (approved):
No blockers from me, only thought would be if it makes sense to tighten the `receivingOrgSvcStatus` param everywhere to be the enum value rather than just a string

frontend-react/yarn.lock
mkalish, line 1: Should this be in this PR?

penny-lischer: It gets updated when doing a yarn install. I can remove it from this PR.

prime-router/src/main/kotlin/history/azure/HistoryDatabaseAccess.kt
penny-lischer: Yes. If we pass `active` then both `active` and `testing` status will get returned and we may want a way to view this in the UI.

victor-chaparro (approved):
LGTM

mkalish (approved):
:shipit:

PR #13595, opened 2024-03-06
Implemented support for stamping multiple condition

### LINKED ISSUE #13572: Implement multiple condition support
### User Story
As an engineer or end user, 
I want to stamp all conditions associated with a LOINC code on an observation
so that they all can be used for research and reporting.

### Description/Use Case
We need to stamp all the conditions associated with an observation's LOINC code during the FHIR convert. We curently only stamp the first matching code.

- This will impact reporting on the conditions
- Be sure to update any code that retrieves mapped conditions to retrieve all

### Dev Notes
Look at `addMappedConditions` in `FhirBundleHelpers.kt` as well as the `getMappedConditions` helpers.

### Acceptance Criteria
- FHIR convert step stamps all mapped conditions if more than one is present
- Helper methods that retrieve conditions are updated as necessary

thetaurean:
This PR adds support for stamping multiple conditions. It also ensures first-order side effects of this change are handled from a routing and pruning perspective. It does not address any observability/reporting concerns.

Test Steps:
1. Running added unit tests is enough

## Changes
- Stamp all corresponding condition codes rather than the first found

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- Fixes #13572


prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
```diff
@@ -36,28 +36,30 @@ const val conditionExtensionurl = "https://reportstream.cdc.gov/fhir/StructureDe
...
-    return if (condition.isNullOrEmpty()) { // could not find the code
-        null
-    } else { // code found; create Coding instance to return
-        Coding(
-            condition[ObservationMappingConstants.CONDITION_CODE_SYSTEM_KEY],
-            condition[ObservationMappingConstants.CONDITION_CODE_KEY],
-            condition[ObservationMappingConstants.CONDITION_NAME_KEY]
-        )
+    }.toList()
+    return if (conditions.isEmpty()) { // no codes found
```
mkalish, line 53: Nit: I think you can skip this conditional and just call `map` after `filter`?

prime-router/src/test/kotlin/fhirengine/utils/FHIRBundleHelpersTests.kt
```diff
@@ -777,14 +781,65 @@ class FHIRBundleHelpersTests {
...
+                        "Condition Code System",
+                        "Condition Name"
+                    )
+                )
+            )
+        )
+
+        val entry = Observation()
+        val code = CodeableConcept()
+        code.addCoding(Coding("system", "80382-5", "display"))
```
mkalish, line 824: This might be me not grokking the test/fix here, but I thought we were handling situations where the same test code maps to multiple conditions? i.e. `"80382-5"` maps to Influenza (disorder) and Influenza A

mkalish (commented):
Let me know if we should discuss, but code change makes sense, I wonder though if there is a missing test case?

prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
jalbinson: Adding to this for your functional programming knowledge:

Your standard list higher-order functions (`map`, `filter`, `fold`) will be no-ops on empty lists. In this instance if your `filter` returns an empty list your map function will do nothing. This is very handy for chaining together a bunch of functions without having to deal with conditionals.

thetaurean: Yeah this was just a derp. Good catch thanks all 👌

prime-router/src/test/kotlin/fhirengine/utils/FHIRBundleHelpersTests.kt
thetaurean: Derp; nice catch!

mkalish (approved):
Thanks for the updates!

Issue #13021, opened 2024-01-18
Create the NIST ELR 2.5.1 message and data types


mkalish:
### User Story
As a report stream sender, I would like to be able to have an exact mapping for a HL7 message that matches the NIST ELR 2.5.1 conformance profile.

### Description/Use Case
This is covered more in depth in the design spec, but in order to guarantee being able to losslessly and deterministically map from HL7->FHIR->HL7 for a give conformance profile, there needs to exist java classes that match that profile.  The work here is to create all the java code needed for NIST ELR 2.5.1

### Dev Notes
- This `platform/kalish/12752-poc-hapi-generated-nist-elr-message` branch is a good starting point.  The classes there were generated from the NIST ELR 2.5.1 conformance profile and the HAPI HL7 source gen tool

### Acceptance Criteria
- [ ] All the custom datatypes used in the the NIST ELR 2.5.1 are created
- [ ] A class that extends `AbstractMessage` exactly matches the NIST ELR 2.5.1 conformance profile
- [ ] The class can be used to parse a valid NIST ELR 2.5.1 message

Andrey-Glazkv:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/13021?planning-poker) with Zenhub @arnejduranovic @jack-h-wang @jalbinson @JessicaWNava @JFU-NAVA-PBC @jimmyfagan @mkalish @thetaurean

Andrey-Glazkv:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/13021?planning-poker) with Zenhub @brick-green

PR #13618, opened 2024-03-08
13021 nist elr java class

jack-h-wang:
This PR adds a new ORU_R01 AbstractMessage implementation for the NIST ELR 2.5.1 spec.

Test Steps:
1. CI should be enough
2. Verify that the classes match what is in the NIST ELR profile https://hl7v2-elr-testing.nist.gov/mu-elr/

## Changes
- Add NIST ELR 2.5.1 ORU_R01 message java class
- Add tests

### Testing
- [ ] Tested locally?
- [x] Added tests?

## Linked Issues
- #13021


prime-router/src/main/java/fhirengine/translation/hl7/structures/nistelr251/datatype/CE.java
```diff
@@ -0,0 +1,208 @@
+/*
+ * This class is an auto-generated source file for a HAPI
+ * HL7 v2.x standard structure class.
+ *
+ * For more information, visit: http://hl7api.sourceforge.net/
```
arnejduranovic, line 5: We don't need full docs now, but we should update the PR description to include steps on how you generated these (what inputs and tools did you use, etc) so we can review that process as well.

prime-router/src/main/java/fhirengine/translation/hl7/structures/nistelr251/datatype/CE.java
```diff
@@ -0,0 +1,208 @@
+/*
```
arnejduranovic, line 1: nit: these are in the translation folder but have nothing to do with translation proper?

prime-router/src/test/kotlin/fhirengine/translation/NistElrTests.kt
```diff
@@ -0,0 +1,39 @@
...
+import ca.uhn.hl7v2.util.Hl7InputStreamMessageIterator
+import ca.uhn.hl7v2.validation.impl.ValidationContextFactory
+import kotlin.test.Test
+import fhirengine.translation.hl7.structures.nistelr251.message.ORU_R01 as NIST_ELR_ORU_R01
+
+class NistElrTests {
+    @Test
+    fun `test nist elr java class`() {
+        val rawMessage = """
+MSH|^~\&|CDC PRIME - Atlanta, Georgia (Dekalb)^2.16.840.1.114222.4.1.237821^ISO|Avante at Ormond Beach^10D0876999^CLIA|PRIME_DOH|Prime ReportStream|20210210170737||ORU^R01^ORU_R01|371784|P|2.5.1|||NE|NE|USA||||PHLabReportNoAck^ELR_Receiver^2.16.840.1.113883.9.11^ISO
```
arnejduranovic, line 18: I'm assuming this message contains all the possible fields the classes support? How was it created?

arnejduranovic:
Also, where is this AC criterion met? 

`A class that extends AbstractMessage exactly matches the NIST ELR 2.5.1 conformance profile`

prime-router/src/main/java/fhirengine/translation/hl7/structures/nistelr251/datatype/CE.java
mkalish: So I'm not sure if I would recommend doing that moving forward?  Maybe we can do a quick 16th on this.

I'm also realizing that we should confirm that these actually differ from the base CE

prime-router/src/test/kotlin/fhirengine/translation/NistElrTests.kt
mkalish: I _think_ Jack grabbed it out of one of the elims messages, but I can take a second look and make sure it's comprehensive

mkalish:
@arnejduranovic That AC is met by the ORU_R01.java file

prime-router/src/main/java/fhirengine/translation/hl7/structures/nistelr251/datatype/CE.java
mkalish: Not sure I follow, this is the same package with the `HL7ToFhirTranslator`?

mkalish:
@arnejduranovic This is ready for another review

brick-green:
The only missing datatype I see is `CWE_ELR_OBX-5`. But the only difference with `CWE_ELR` are in the predicate column in the nist profile which don't impact the java class. Is that why you didn't include it?

arnejduranovic (approved):

mkalish:
@brick-green I think it didn't get picked up by the code generator and then when I went through the segment's it didn't appear (because OBX-5 varies).  Looking at it now, I think we can safely not create a separate file since the shape is the same?

.github/actions/sonarcloud/action.yml
```diff
@@ -21,7 +20,7 @@ runs:
         SONAR_TOKEN: ${{ inputs.sonar-token }}
       with:
         args: >
-          -Dsonar.coverage.exclusions=prime-router/src/test/**,prime-router/src/testIntegration/**,prime-router/src/main/kotlin/cli/tests/**,frontend-react/**/__mocks__/**,frontend-react/**/mocks/**,frontend-react/**/*.test.*
+          -Dsonar.coverage.exclusions=prime-router/src/test/**,prime-router/src/testIntegration/**,prime-router/src/main/kotlin/cli/tests/**,frontend-react/**/__mocks__/**,frontend-react/**/mocks/**,frontend-react/**/*.test.*,prime-router/src/main/java/**
```
snesm, line 23: Why exclude the 51 java files in `prime-router/src/main/java/**` ?

mkalish: These are pretty just going to be POJO files/configuration and there is limited value in them getting scanned.  I don't feel super strongly about it though, so I'm happy to add a follow up ticket that will set up sonar to scan java files (it wasn't working off the bat which is why I excluded them)

brick-green:
> @brick-green I think it didn't get picked up by the code generator and then when I went through the segment's it didn't appear (because OBX-5 varies). Looking at it now, I think we can safely not create a separate file since the shape is the same?

Yeah I agree

.github/actions/sonarcloud/action.yml
snesm: This only excludes them from coverage statistics.  Maybe the coverage supporting files are not generated for this new folder.  Please ticket for investigation.

snesm (approved):

PR #13625, opened 2024-03-08
FHIR Extension Registry

brick-green:
This PR adds a file to document all the custom FHIR extensions created and used in ReportStream.


prime-router/docs/design/design/RS-Extention-Registry.yml
```diff
@@ -0,0 +1,5930 @@
...
+                      }
+                    ]
+                  }
+                ]
+              }
+            ]
+          }
+        }
+      Extension:
+        - id: sender-id
```
mkalish, line 134: I think this is missing in the example above?  But also thinking that maybe we cannot include it, I think we only use this internally

prime-router/docs/design/design/RS-Extention-Registry.yml
```diff
@@ -0,0 +1,5930 @@
...
+          description: Identifies the entity sending the bundle to ReportStream
+        - id: encoding-characters
+          url: "https://reportstream.cdc.gov/fhir/StructureDefinition/encoding-characters"
+          valueType: String
+          hl7Use: MSH.2
+        - id: character-set
+          url: "https://reportstream.cdc.gov/fhir/StructureDefinition/character-set"
+          valueType: String
+          hl7Use: MSH.18
+        - id: MSH.7
```
mkalish, line 147: Wondering if this might be a little confusing since this is technically nested in the msh-message-header extension?  Maybe we can follow the same pattern as MSH.24?

prime-router/docs/design/design/RS-Extention-Registry.yml
```diff
@@ -0,0 +1,5930 @@
...
+          valueType: String
+          hl7Use: MSH.7
+        - id: MSH.13
+          url: "MSH.13"
+          valueType: String
+          hl7Use: MSH.13
+        - id: MSH.14
+          url: "MSH.14"
+          valueType: String
+          hl7Use: MSH.14
```
mkalish, line 158: Nit: maybe these should be `hl7v2Field` to line up with the extension?

prime-router/docs/design/design/RS-Extention-Registry.yml
```diff
@@ -0,0 +1,5930 @@
...
+          valueType: String
+          hl7Use: SFT.6
+    HD:
+      Example:
+        {
+          "fullUrl": "MessageHeader/827ccb0e-ea8a-306c-8c34-a16891f84e7b",
+          "resource": {
+            "resourceType": "MessageHeader",
+            "id": "827ccb0e-ea8a-306c-8c34-a16891f84e7b",
+            "extension": [
```
mkalish, line 274: Nit: maybe delete the bit here that ins't from an `HD` or comment where the HD is?

prime-router/docs/design/design/RS-Extention-Registry.yml
```diff
@@ -0,0 +1,5930 @@
...
+          url: "https://reportstream.cdc.gov/fhir/StructureDefinition/universal-id-type"
+          valueType: String
+          hl7Use: HD.3
+        - id: hl7v2Field
+          url: "https://reportstream.cdc.gov/fhir/StructureDefinition/hl7v2Field"
+          valueType: String
+          hl7Use: NA
+          constant: "MSH.3", "MSH.24"
+
+- Resource: Any
```
mkalish, line 325: Should this be `Attachment`?

prime-router/docs/design/design/RS-Extention-Registry.yml
```diff
@@ -0,0 +1,5930 @@
...
+              }
+            ]
+          }
+        }
+      Extension:
+        - id: hl7v2Field
+          url: "https://reportstream.cdc.gov/fhir/StructureDefinition/hl7v2Field"
+          valueType: String
+          hl7Use: NA
+          description: Identifies the HL7 origin/destination of the data.
```
mkalish, line 707:
```suggestion
          description: Identifies the HL7 field of the data.
```

prime-router/docs/design/design/RS-Extention-Registry.yml
```diff
@@ -0,0 +1,5930 @@
...
+      Extension:
+        - id: coding-system-oid
+          url: "https://reportstream.cdc.gov/fhir/StructureDefinition/coding-system-oid"
+          valueType: OID
+          hl7Use: CWE.14
+        - id: cwe-coding
+          url: "https://reportstream.cdc.gov/fhir/StructureDefinition/cwe-coding"
+          valueType: String
+          hl7Use: NA
+          constant: "coding"
```
mkalish, line 1119: Can't this also be `alt-coding` etc.?

mkalish (commented):
Really great stuff here!  I think this will really be useful for anyone (including ourselves) understanding how these should work.    I did a once over, but will do a line by line review when you're ready to move this out of draft.  Few thoughts:

- Just making sure theres consistency on nesting the extensions (noticed a discrepancy on MSH)
- Maybe subbing hl7v2Field in to line up with the extensions
- I was a little confused  by the `Resource` and the `FHIR-Type` attributes and what they represent

prime-router/docs/design/design/RS-Extention-Registry.yml
brick-green: yep, and I was missing `secondary-alt-coding` as an option. Added both

prime-router/docs/design/design/RS-Extention-Registry.yml
brick-green: I didn't include this as Resource since we don't use it as a resource in our mapping. We only ever use `valueAttachment` in extensions

prime-router/docs/design/design/RS-Extention-Registry.yml
brick-green: I left MessageHeader resource in formation in the drive home that this is only on the MessageHeader resource. The individual HD fields are marked in the Extension section. Think that's sufficient?

prime-router/docs/design/design/RS-Extention-Registry.yml
brick-green: Yup, that was a mistake. Fixed

prime-router/docs/design/design/RS-Extention-Registry.yml
brick-green: good call. Removed

prime-router/docs/design/design/RS-Extention-Registry.yml
```diff
@@ -0,0 +1,3773 @@
...
+                  "system": "http://terminology.hl7.org/CodeSystem/v3-ActCode",
+                  "code": "AMB",
+                  "display": "ambulatory"
+                }
+              }
+            ]
+          }
+        }
+      Extension:
+        - id: hl7v2Field
```
mkalish, line 120: This doesn't have a description, which could make sense, but I'm wondering if we want to provide maybe a section at the top that explains these generic extensions?

prime-router/docs/design/design/RS-Extention-Registry.yml
```diff
@@ -0,0 +1,3773 @@
...
+                  hl7Use: HD.1
+                - id: universal-id
+                  url: "https://reportstream.cdc.gov/fhir/StructureDefinition/universal-id"
+                  valueType: String
+                  hl7Use: HD.2
+                - id: universal-id-type
+                  url: "https://reportstream.cdc.gov/fhir/StructureDefinition/universal-id-type"
+                  valueType: String
+                  hl7Use: HD.3
+    SFT:
```
mkalish, line 617: We should update the SFT->Device ticket to mention updating this

prime-router/docs/design/design/RS-Extention-Registry.yml
```diff
@@ -0,0 +1,3773 @@
...
+        {
+          "fullUrl": "Organization/1707860303759373000.da28a601-4c4b-4b7e-84f5-b71cd42e27ab",
+          "resource": {
+            "resourceType": "Organization",
+            "id": "1707860303759373000.da28a601-4c4b-4b7e-84f5-b71cd42e27ab",
+            "extension": [
+              {
+                "url": "https://reportstream.cdc.gov/fhir/StructureDefinition/cwe-organization",
+                "valueCodeableConcept": {
+                  # CodeableConcept Data
```
mkalish, line 811: Nit: is there another example we could use or delete the comment?

prime-router/docs/design/design/RS-Extention-Registry.yml
```diff
@@ -0,0 +1,3773 @@
...
+              }
+            },
+            "extension": [
+              {
+                "url": "https://reportstream.cdc.gov/fhir/StructureDefinition/pid-patient",
+                "extension": [
+                  {
+                    "url": "PID.8",
+                    "valueCodeableConcept": {
+                      # CWE to CodeableConcept
```
mkalish, line 1100: Nit: same thought here

prime-router/docs/design/design/RS-Extention-Registry.yml
```diff
@@ -0,0 +1,3773 @@
...
+              }
+            ],
+            "value": "Entity ID",
+            "assigner": {
+              "reference": "Organization/1707769195125082000.40e4d3f4-bbd6-43af-aefe-5894b31e9806"
+            }
+          }
+        ]
+      Extension:
+        - id: ei-assigner-organization-type
```
mkalish, line 3037: Not sure I have a great solution, but this was a tad confusing.  This extension is one where the value is either "default" or "organization"?  Maybe they should be collapsed and it should be `constants`?

prime-router/docs/design/design/RS-Extention-Registry.yml
```diff
@@ -0,0 +1,3773 @@
...
+              url: "XTN.9"
+              valueType: String
+              hl7Use: XTN.9
+            - id: XTN.12
+              url: "XTN.12"
+              valueType: String
+              hl7Use: XTN.12
+
+- Resource: Any
+  FHIR-Type: Destination
```
mkalish, line 3365: Nit: should this be `BackboneElement` or `MessageHeader.destination`?

mkalish (approved):
Really really good stuff.

My only thought is if there is worth adding a section at the beginning (or a readme) that explains how this should be interpreted.  I'm thinking it would include explanations of why there is a need for the `hl7v2field` extension.

PR #13679, opened 2024-03-14
13505: Replace e2e up test with more performant version

### LINKED ISSUE #13505: Create Performant End-to-End UP Tests
### User Story

As a developer of the Universal Pipeline,
I want to have tests that ensure reports are being delivered successfully and have the expected contents,
so that I can confirm the system is working properly.

### Description/Use Case

Presently, `testEnd2EndUP` takes a lot of time to execute due to inefficient polling and is only testing lineage, something that can be more efficiently tested on a lower level, like UP step-specific integration tests. This ticket aims to create a new end2end test that executes faster and tests that a report was correctly received instead of querying to check lineage is correct. 

This new set of end2end tests should:
- Verify Report is correctly sent to expected receiver for full-elr topic
- Verify Report is correctly sent to expected receiver for elr-elims topic
- Verify contents of received report are as expected
- have a low execution time

This new set of end2end should NOT:
- Thoroughly test various parts of the pipeline, like routing and translation. That can be left up to the integration tests of each of the various pipeline steps

### Acceptance Criteria

- [ ] New set of performant end2end tests created for UP that verifies the basic functionality of the system

JFisk42:
This PR removes the old Universal Pipeline end2end test and replaces it with a more performant test. The goal of this new test is not to verify each step in the report lifecycle or lineage, but to verify that it is sent to the correct receivers and that the contents match what we expect them to be.

**Note: Still new here so please feel free to correct any wording issues in documentation/logging/variable naming that may stem from incomplete understanding.**

Test Steps:
1. Run `./gradlew testEnd2EndUP`
2. Expect to see two reports created. One for the FULL_ELR workflow and the other for ELIMS.
3. Verify the receivers and sent file names in the output via `curl --location 'http://localhost:7071/api/waters/report/[report_id]/history/'`
4. Running in staging requires setting the AzureWebJobsStorage environment variable. Because that variable contains semicolons it will need to be set in CLI like `export AzureWebJobsStorage="<key from azure>"`

## Changes
- Replaced the old UP e2e test with a test that posts a report. It then verifies the sent reports were sent to the correct receivers. It also pulls the resulting files from the blob store and verifies that the contents are what we expect.
- This allowed a reduction in local test run time from about 8 minutes to under 1.5 minutes.

### Testing
- [x] Tested locally?

### Process

## Linked Issues
- Fixes #13505

## TODO
- https://app.zenhub.com/workspaces/platform-6182b02547c1130010f459db/issues/gh/cdcgov/prime-reportstream/14336


prime-router/src/main/kotlin/cli/tests/BasicTests.kt
```diff
@@ -66,172 +66,94 @@ class Ping : CoolTest() {
...
+         passed = passed and examinePostResponse(json, false)
+         if (!passed) {
+             bad("***end2end_up FAILED***: Error in post response")
+         }
+
+         val reportId = getReportIdFromResponse(json)
+             ?: return bad("***end2end_up FAILED***: Did not find report id")
+
+         // Need just a little wait for the db to catch up
+         waitABit(10, environment)
```
JFisk42, line 134: Based on my understanding of how [waitABit](https://github.com/CDCgov/prime-reportstream/blob/791e8b0c79c97238239ea894bda6541c590be3b8/prime-router/src/main/kotlin/cli/tests/TestReportStream.kt#L449) works, this should be fine. **However**, there may be edge cases in how the UP system works that this won't be a long enough wait for, or it may be way too long of a wait. Opinions welcome.

prime-router/src/main/kotlin/cli/tests/BasicTests.kt
```diff
@@ -66,172 +66,94 @@ class Ping : CoolTest() {
 }
 
 /**
- * End to end test for the universal pipeline. This test should read in a submission, verify it reached all steps
- * (receive, convert, route, translate, batch, send) and that it generated the correct lineage records.
+ * Updated end2end test for the universal pipeline. This test posts a report and verifies that the correct
+ * receiver received the file and that the correct file is in the sftp directory.
+ * This test is repeated: Once for the normal UP workflow and once for the ELIMS workflow.
  */
 class End2EndUniversalPipeline : CoolTest() {
```
JFisk42, line 73: I considered renaming the test name and commands, but I couldn't think of any conflict with keeping them the same.

prime-router/src/main/kotlin/cli/tests/BasicTests.kt
mkalish: So as alternative you could use the pattern [here](https://github.com/CDCgov/prime-reportstream/blob/1e725139b15d2438bd2e743bf6f377fa5a58baf0/prime-router/src/main/kotlin/cli/tests/SyncTranslationSchemasTest.kt#L116) which is essentially just a coroutine with a timeout.

It might be a little more reliable since it can complete quicker if the report is ready.

I also don't think the comments in `waitABit` actually work since the output is printed after the tests are done so you don't actually get feedback

prime-router/src/main/kotlin/cli/tests/BasicTests.kt
```diff
@@ -66,172 +66,94 @@ class Ping : CoolTest() {
...
+         }
+
+         val reportId = getReportIdFromResponse(json)
+             ?: return bad("***end2end_up FAILED***: Did not find report id")
+
+         // Need just a little wait for the db to catch up
+         waitABit(10, environment)
+
+         db = WorkflowEngine().db
+         db.transact { txn ->
```
mkalish, line 137: I'm wondering if there is a way we can figure out another way to do this without having to connect to the db.  This would simplify running these tests against staging.

prime-router/src/main/kotlin/cli/tests/BasicTests.kt
```diff
@@ -66,172 +66,94 @@ class Ping : CoolTest() {
...
-                        )
-                    val sendResults = pollForStepResult(translateReportId, TaskAction.send)
-                    // verify each result is valid
-                    for (result in sendResults.values)
-                        passed = passed && examineStepResponse(result, "send", sender.topic)
-                    if (!passed) {
-                        bad("***async end2end_up FAILED***: Send result invalid")
-                    }
+                // Check that created file exists in sftp directory
+                val createdFile = File("./build/sftp/$fileName")
```
mkalish, line 149: So I'm actually not sure what this should be, but I don't know how this would work when running against the staging environment.  Probably a good 17th discussion...

prime-router/src/main/kotlin/cli/tests/BasicTests.kt
```diff
@@ -66,172 +66,94 @@ class Ping : CoolTest() {
...
+                if (createdFile.exists()) {
+                    good("File sftp succeeded with the creation of $fileName")
```
mkalish, line 151: Nit: What do you think about checking contents to make sure that they match what should have been generated?  My thoughts are that for the a topic where we're sending the original file we should double check that we're sending that exactly and then for a topic where we're applying a transform were correctly doing that.

mkalish (commented):
Awesome stuff!  Super exciting to see these getting improved.  Let me know if you want to chat about any of my comments.

arnejduranovic (changes requested):
Looking a lot nicer than the other tests! Great job reducing the runtime.

prime-router/src/main/kotlin/cli/tests/BasicTests.kt
```diff
@@ -115,125 +138,162 @@ class End2EndUniversalPipeline : CoolTest() {
...
+
+        expectedReceivers.forEach { expectedReceiver ->
+            // Retrieve external filenames from the history endpoint
+            val actualFilename = findReportExternalNames(response, expectedReceiver)
+            if (actualFilename.isNotEmpty()) {
+                good("The report was received by ${expectedReceiver.name}")
+            } else {
+                bad("***$name FAILED***: Did not find the sent report name for receiver ${expectedReceiver.name}")
+                passed = false
+                return@forEach
```
JFisk42, line 181: Still wrapping my brain around Kotlin so let me know if there's a better way to do this. My thought here is that this part of the test happens in a few seconds, so the test should still go through the motion of checking each sent report even if one fails.

prime-router/src/main/kotlin/cli/tests/BasicTests.kt
JFisk42: Discussed in standup: We are going to keep the old test as removed.

prime-router/src/main/kotlin/cli/tests/BasicTests.kt
```diff
@@ -66,172 +66,94 @@ class Ping : CoolTest() {
...
+     override val status = TestStatus.SMOKE
+
+     override suspend fun run(environment: Environment, options: CoolTestOptions): Boolean {
+         initListOfGoodReceiversAndCountiesForUniversalPipeline()
+
+         ugly("Running $name asynchronously")
+         var passed = true
+
+         ugly("Starting $name Test: send ${fullELRSender.fullName} data to $allGoodCounties")
+         passed = passed and end2EndUP(
```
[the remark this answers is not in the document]
JFisk42: I took a little different approach. Let me know what you think: https://github.com/CDCgov/prime-reportstream/pull/13679#discussion_r1558272977. This gets it under 90 seconds every time.

arnejduranovic (changes requested):

prime-router/src/main/kotlin/cli/tests/TestReportStream.kt
```diff
@@ -870,10 +871,10 @@ abstract class CoolTest {
                 ?: error("Unable to find sender $fullELRSenderName for organization ${org1.name}")
         }
 
-        const val etorTISenderName = "ignore-etor-ti"
-        val etorTISender by lazy {
-            settings.findSender("$org1Name.$etorTISenderName") as? UniversalPipelineSender
-                ?: error("Unable to find sender $etorTISenderName for organization ${org1.name}")
+        const val fhirSenderName = "ignore-fhir-e2e"
```
JFisk42, line 874: There were a few variables declared here that weren't being used, so I cleaned them up.

prime-router/src/main/kotlin/cli/tests/TestReportStream.kt
```diff
@@ -958,11 +950,39 @@ abstract class CoolTest {
             allGoodCounties = allGoodReceivers.joinToString(",") { it.name }
         }
 
-        fun initListOfGoodReceiversAndCountiesForUniversalPipeline() {
-            allGoodReceivers = mutableListOf(universalPipelineReceiver1)
-            allGoodCounties = allGoodReceivers.joinToString(",") { it.name }
+        fun initTestDataForUniversalPipeline() {
```
JFisk42, line 953: Again, Kotlin noob here, let me know if I'm doing anything weird with this.

prime-router/src/main/kotlin/cli/tests/BasicTests.kt
```diff
@@ -115,126 +137,194 @@ class End2EndUniversalPipeline : CoolTest() {
...
+     * process will be skipped. Otherwise, the file will be converted to FHIR if necessary. Sender enrichments will be
+     * run. Then translated based on the expected receiver format.
+     * Caveats: This does not perform all transformations from the UP. Notably, batch headers/footers are absent.
+     * @param originalFile the file to transform
+     * @param expectedReceiver determines resulting data format
+     * @param sender used for the sender enrichment schema
+     * @param blobMetadata necessary for the FhirTransformer
+     * @return the transformed ByteArray in the proper data format of HL7 or FHIR
+     */
+    private fun performFileTransforms(
```
brick-green, line 293: Is this function necessary? Could you not just create the control file manually then supply that in to compare to the output? 

If there was a bug introduced to another part of the code, like maybe in `FhirToHl7Converter()`, this function wouldn't catch that, right?  To me it looks like it would make a new control file with the code that has a bug then compare that to the file made by the pipeline that also uses the code with a bug. So the files would be the same and the test would pass.

brick-green (commented):
one question and one suggestion

prime-router/src/main/kotlin/cli/tests/BasicTests.kt
```diff
@@ -115,126 +137,194 @@ class End2EndUniversalPipeline : CoolTest() {
...
-                if (!receiverNames.contains(receiver.fullName)) {
-                    bad("***async end2end_up FAILED***: expected ${receiver.fullName} to have received a report")
-                    passed = false
-                }
+        echo("\nPolling for history endpoint for report ID: $reportId. (Max poll time $maxPollSecs seconds)")
+        while (timeElapsedSecs <= maxPollSecs) {
+            delay(pollSleepSecs.toLong() * 1000)
+            timeElapsedSecs += pollSleepSecs
+
+            val getUrl = "${environment.url}/api/waters/report/$reportId/history"
```
JFisk42, line 167: So this call is returning a 401 in staging.

prime-router/src/main/kotlin/cli/tests/BasicTests.kt
JFisk42: That's a good question. Because it's using the same classes as the actual posted report, any bugs should be mirrored. Off the top of my head I think the main issue would be if there are any changes to conversion/transformation in the pipeline, or the sender/receiver, then the expected file would have to be updated.

prime-router/src/test/resources/fhirengine/smoketest/valid_hl7.hl7
```diff
@@ -1,6 +1,6 @@
...
+PID|1||100001^^^MEDITECH&2.16.840.1.114222.4.3.2.2.1.321.111&ISO^MR^COCAA&&ISO~285-64-4175^^^MEDITECH&2.16.840.1.114222.4.3.2.2.1.321.111&ISO^SS^COCAA&&ISO~X605236^^^MEDITECH&2.16.840.1.114222.4.3.2.2.1.321.111&ISO^PI^COCAA&&ISO~E40014904640^^^MEDITECH&2.16.840.1.114222.4.3.2.2.1.321.111&ISO^AN^COCAA&&ISO||BIERLY^LONNIE^^^^^L||19361023|M||2106-3^White^HL70005^W^White^L^2.5.1^5.67|4673 ELM ST^CHERRY CRK NURSING CTR^CANTERBURY^IG^80014^USA^H||^PRN^PH^^1^303^6930111|^WPN^PH^^1^303^9999999^^Business||S|NOE^Nonreligious^HL70006|||||N^Not Hispanic or Latino^HL70189^2^Not Hispanic or Latin^L^2.5.1^5.61||||||||N|||20210809085200-0600|COCAA^^ISO|337915000^Homo sapiens (organism)^SCT^Human^Human^L^1^1
 NTE|1|P|EH|RE^Remark^HL70364^C^Comment^L^2.5.1^V1
 NK1|1|SURYAN^GENARO^^^^^L|OTH^Other^HL70063^OT^OTHER RELATIONSHIP^L^2.5.1^5.67|4861 20TH AVE^^THUNDER MOUNTAIN^NV^99999^USA^H|^PRN^PH^^1^720^3013954||||||||||||||||||||||||||^WPN^PH^^1^720^3013954|4861 20TH AVE^^AURORA^CO^99999^USA^H
```
JFisk42, line 5: Going to need to change the state from CO here. Noticed when running the test in staging that a receiver for the Colorado Dept of Health is also receiving the data. Because the patient state is IG and org state is CO its getting received by multiple orgs' receivers. Dunno if how bad this is but it's something I think should be proactively changed.

prime-router/src/main/kotlin/cli/tests/BasicTests.kt
```diff
@@ -115,125 +138,162 @@ class End2EndUniversalPipeline : CoolTest() {
...
+
+    private suspend fun end2EndUP(
+        file: File,
+        environment: Environment,
+        sender: Sender,
+        expectedReceivers: List<Receiver>,
+        reportId: ReportId,
+    ): Boolean {
+        var passed = true
+        val connectionInfo = "DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;" +
```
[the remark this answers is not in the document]
JFisk42: The answer to this was absolutely it will break. This has now been addressed.

prime-router/src/main/kotlin/cli/tests/BasicTests.kt
```diff
@@ -115,125 +138,162 @@ class End2EndUniversalPipeline : CoolTest() {
...
-                    // verify each result is valid
-                    for (result in sendResults.values)
-                        passed = passed && examineStepResponse(result, "send", sender.topic)
-                    if (!passed) {
-                        bad("***async end2end_up FAILED***: Send result invalid")
+        try {
+            val jsonResponse = jacksonObjectMapper.readTree(response)
+
+            if (!jsonResponse.isEmpty && !jsonResponse["destinations"].isEmpty) {
+                jsonResponse["destinations"].forEach { destination ->
```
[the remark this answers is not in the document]
JFisk42: There doesn't appear to be anything in Kotlin to easily read in json to collections without first building up some classes/structures to contain the expected data. I went with a different json library that is cleaner than this original implementation.  https://github.com/CDCgov/prime-reportstream/pull/13679/files?w=0#diff-4025d6168d38b38532a410c9be4bcfa26e83dfb2239af85cf8569c57a77374a8R254

prime-router/src/main/kotlin/cli/tests/TestReportStream.kt
JFisk42: I like this a lot more! Array/Lists and the creation thereof in Kotlin is something that still feels weird to me.

prime-router/src/main/kotlin/cli/tests/BasicTests.kt
JFisk42: Discussed this in stand up today. The general choices are to generate the expected file in the test and provide extra test coverage for translation schemas or use a saved expected which provides extra test coverage for the various steps of the UP. It was decided that saved expected files were the route to go, but that the merit of generating them could be useful in future tests so this method has been preserved. https://github.com/CDCgov/prime-reportstream/pull/13679/files?w=0#diff-4025d6168d38b38532a410c9be4bcfa26e83dfb2239af85cf8569c57a77374a8R276

prime-router/src/main/kotlin/cli/tests/BasicTests.kt
JFisk42: I fixed this by adding an access token that is generated when logging into staging from the Prime CLI https://github.com/CDCgov/prime-reportstream/pull/13679/files?w=0#diff-4025d6168d38b38532a410c9be4bcfa26e83dfb2239af85cf8569c57a77374a8R163. Also swapped out a different HTTP util call.

prime-router/src/test/resources/fhirengine/smoketest/valid_hl7.hl7
JFisk42: Fixed this but something within the change in the valid_hl7.hl7 file made other tests failed. Took the route of making a copy of this file for just e2e tests. https://github.com/CDCgov/prime-reportstream/pull/13679/commits/29518aa9ed2e39fa79b5fd35a9803796705a0f97

brick-green (approved):
looking good to me!

prime-router/src/main/kotlin/cli/tests/BasicTests.kt
```diff
@@ -115,126 +131,214 @@ class End2EndUniversalPipeline : CoolTest() {
...
-            if (routeReportIds.size < expectedReceivers.size) {
-                return bad(
-                    "***async end2end_up FAILED***: Expected at least ${expectedReceivers.size} route" +
-                        "report id(s), but got ${routeReportIds.size}."
+    /**
+     * Runs validations after a report has been submitted to the UP and marked as Delivered.
+     * For all expected receivers this will validate:
+     * - An external filename matching that receiver is present in the history endpoint response
+     * - Download the file from BlobStorage
+     * - Compare the expected data with the actual data returned from BlobStorage
```
arnejduranovic, line 194: We should also test that the name of the downloaded file is what we expect (I think there is a bug and you will find it's not what we expect?)

prime-router/src/main/kotlin/cli/tests/BasicTests.kt
```diff
@@ -66,45 +72,55 @@ class Ping : CoolTest() {
...
- * (receive, convert, route, translate, batch, send) and that it generated the correct lineage records.
+ * An End-to-End test for the Universal Pipeline. This test posts a report and verifies that the file was sent to
+ * the expected receiver(s), then verifies the contents of the file match the expected contents.
+ * This test runs through four scenarios:
+ * - A report that is submitted as HL7 and is sent as HL7 (Topic: FULL-ELR).
+ * - A report that is submitted as HL7 and is sent as FHIR (Topic: FULL-ELR).
+ * - A report that is submitted as HL7 and is sent as HL7 but has the send original flag (Topic: ELIMS).
+ * - A report that is submitted as FHIR and is sent as FHIR (Topic: FULL-ELR).
  */
 class End2EndUniversalPipeline : CoolTest() {
```
arnejduranovic, line 83: Right now the test is set up in such a way where it will fail early if any one test failed, but it would be better if we could run all of them. This would lead to faster turn around if more than one test was failing?

arnejduranovic (changes requested):

prime-router/settings/organizations.yml
```diff
@@ -243,6 +255,75 @@
         port: 22
         filePath: ./upload
         credentialName: DEFAULT-SFTP
+    - name: FULL_ELR_E2E      # For the E2E UP test
+      externalName: Ignore FULL_ELR_E2E
+      organizationName: ignore
+      topic: full-elr
+      customerStatus: active
+      qualityFilter:
+        - "%messageId.exists()"
```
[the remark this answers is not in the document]
JFisk42: I guess it just depends on what role if any you want for the quality filters in these tests. At the moment these exist simply because of the receivers I was copying from. I can modify it into something like you're suggesting and this would be an extra layer tested in these now.

prime-router/build.gradle.kts
```diff
@@ -954,7 +954,7 @@ dependencies {
     testImplementation("io.ktor:ktor-client-mock:$ktorVersion")
     testImplementation("org.jetbrains.kotlinx:kotlinx-coroutines-core:1.8.0")
     testRuntimeOnly("org.junit.jupiter:junit-jupiter-engine:5.10.2")
-    testImplementation("org.junit.jupiter:junit-jupiter:5.10.2")
+    implementation("org.junit.jupiter:junit-jupiter:5.10.2")
```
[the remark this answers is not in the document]
JFisk42: Good catch. This was required at one point: https://nava.slack.com/archives/C026DV8R21W/p1710800247416919 but after looking at it now it's no longer required. I have reverted this.

prime-router/src/main/kotlin/cli/tests/BasicTests.kt
JFisk42: Good idea. Currently the only spot that would cause the whole test to not run is if one of the reports never got a `overallStatus` of `Delivered`. I've changed that to log as a failure and continue the test. https://github.com/CDCgov/prime-reportstream/pull/13679/commits/caafe52d3289e87769aaae7bb31f24f14f6d6c55

prime-router/src/main/kotlin/cli/tests/BasicTests.kt
```diff
@@ -66,45 +72,55 @@ class Ping : CoolTest() {
...
+    override val description = "e2e tests that verify reports are correctly uploaded, routed, and received"
     override val status = TestStatus.SMOKE
 
     override suspend fun run(environment: Environment, options: CoolTestOptions): Boolean {
-        initListOfGoodReceiversAndCountiesForUniversalPipeline()
-        ugly("Starting $name Test: send ${fullELRSender.fullName} data to $allGoodCounties")
-
-        // run through the universal pipeline test
-        ugly("Running end2end_up asynchronously")
+        initTestDataForUniversalPipeline()
```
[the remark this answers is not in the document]
JFisk42: Sounds good! https://github.com/CDCgov/prime-reportstream/pull/13679/commits/39831faf5441d7aa958d4e4571012d4fd8c23cee

prime-router/settings/organizations.yml
```diff
@@ -111,6 +111,12 @@
       topic: mars-otc-elr
       customerStatus: active
       format: HL7
+    - name: ignore-full-elr-e2e  # For the E2E UP test
```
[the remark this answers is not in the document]
arnejduranovic: Yep that makes sense, we just want to make sure staging config matches this one.

prime-router/src/main/kotlin/cli/tests/BasicTests.kt
```diff
@@ -66,45 +72,55 @@ class Ping : CoolTest() {
...
 /**
- * End to end test for the universal pipeline. This test should read in a submission, verify it reached all steps
- * (receive, convert, route, translate, batch, send) and that it generated the correct lineage records.
+ * An End-to-End test for the Universal Pipeline. This test posts a report and verifies that the file was sent to
+ * the expected receiver(s), then verifies the contents of the file match the expected contents.
+ * This test runs through four scenarios:
+ * - A report that is submitted as HL7 and is sent as HL7 (Topic: FULL-ELR).
+ * - A report that is submitted as HL7 and is sent as FHIR (Topic: FULL-ELR).
+ * - A report that is submitted as HL7 and is sent as HL7 but has the send original flag (Topic: ELIMS).
+ * - A report that is submitted as FHIR and is sent as FHIR (Topic: FULL-ELR).
```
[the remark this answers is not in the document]
JFisk42: Ticket created here https://app.zenhub.com/workspaces/platform-6182b02547c1130010f459db/issues/gh/cdcgov/prime-reportstream/14336

JFisk42:
Did final verification of the test in staging and updating deployment guide. Also noting here: The new step will be setting the AzureWebJobsStorage env var before trying to run the test in staging.

PR #13678, opened 2024-03-13
Revert "Experience/13580/e2e submissions"

penny-lischer:
Reverts CDCgov/prime-reportstream#13653

PR #13684, opened 2024-03-14
Revert "Revert "Experience/13580/e2e submissions""

penny-lischer:
Should not have been reverted Reverts CDCgov/prime-reportstream#13678


GilmoreA6 (approved):
incorrectly reverted as part of troubleshooting submissions api smoke test issue.

PR #13695, opened 2024-03-14
Experience/13480/deliveries receiver status

### LINKED ISSUE #13480: Add Receiver Status (active or inactive) to the API return call to ensure only active receivers will be shown on Daily Data

penny-lischer:
This PR ...

Add Receiver Status parameter to the API return call to allow for only active receivers on Daily Data

Test Steps:

1. In HTTP client append the `receivingOrgSvcStatus` param to the deliveries call i.e
`/api/waters/org/ca-dph/deliveries?sortDir=DESC&sortColumn=batchReadyAt&since=2023-11-30T00:00:00.000Z&until=2023-12-31T00:00:00.000Z&receivingOrgSvcStatus=ACTIVE&receivingOrgSvcStatus=INACTIVE`

## Changes
Added optional receivingOrgSvcStatus parameter to the deliveries API so that the daily data page can display only active and testing records.

Decoupled the fetchActions method since deliveries required more logic than submissions.
Screenshot 2024-03-21 at 12 13 13 PM

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- Fixes #13480


prime-router/src/main/kotlin/history/ReportHistory.kt
```diff
@@ -37,6 +38,7 @@ abstract class ReportHistory(
     var reportId: String? = null,
     var topic: Topic? = null,
     var reportItemCount: Int? = null,
+    var receivingOrgSvcStatus: String? = null,
```
mkalish, line 41: Now that we've split up the DB actions, maybe makes sense to only do this for the a Delivery?

I see this
```
{
        "submissionId": 427,
        "timestamp": "2024-03-13T12:16:54.826Z",
        "id": "cb7451ff-929a-4633-ae51-658fdc51ca9d",
        "topic": "elr-elims",
        "reportItemCount": 1,
        "httpStatus": 201,
        "receivingOrgSvcStatus": null,
        "sender": "ignore.ignore-elr-elims"
    },
```
locally when using the submissions API

prime-router/src/main/kotlin/history/azure/ReportFileFunction.kt
```diff
@@ -313,6 +317,12 @@ abstract class ReportFileFunction(
             fun extractShowFailed(query: Map<String, String>): Boolean {
                 return query["showfailed"]?.toBoolean() ?: false
             }
+
+            fun extractReceivingOrgSvcStatus(query: Map<String, String>): CustomerStatus? {
```
mkalish, line 321: Sorry, I missed this in the first PR, but is the intention to let through a request with an invalid CustomerStatus?  Currently, this returns a `400` and
```
{
    "error": "No enum constant gov.cdc.prime.router.CustomerStatus.INACTIV"
}
```

prime-router/src/main/kotlin/history/azure/HistoryDatabaseAccess.kt
```diff
@@ -148,6 +229,20 @@ abstract class HistoryDatabaseAccess(
     ): Condition {
         var filter = this.organizationFilter(organization, orgService)
 
+        if (receivingOrgSvcStatus != null) {
+            filter = if (receivingOrgSvcStatus.toString() == CustomerStatus.ACTIVE.toString()) {
```
mkalish, line 233: I think you can drop the `toString` here

mkalish: Actually, can this not all be simplified?

```
filter.and(
                    DSL.jsonbGetAttributeAsText(SETTING.VALUES, "customerStatus")
                    .notEqual(customerStatus.name)
                )
```
?

mkalish (changes requested):
Confirmed that the updates to split the functions maintains both APIs working.  Had a few thoughts, I apologize if I missed these when initially reviewing them

arnejduranovic (commented):
Did we add a test to catch the issue? Instead of the issue being caught by an Auth test, it should be caught by a test that runs on the CI build.

penny-lischer:
> Did we add a test to catch the issue? Instead of the issue being caught by an Auth test, it should be caught by a test that runs on the CI build.

There was no error. The submissions were just returning an empty array.

prime-router/src/main/kotlin/history/azure/ReportFileFunction.kt
penny-lischer: This gets caught in the try/catch of the `getListByOrg` function. Is there a way to catch the error before that?

prime-router/src/main/kotlin/history/azure/HistoryDatabaseAccess.kt
penny-lischer: still need the lowercase() since the customerStatus is stored in the dbs as lowercase.

mkalish: If you use `name` instead of `toString` I think it will work?

prime-router/src/main/kotlin/history/azure/HistoryDatabaseAccess.kt
```diff
@@ -148,6 +229,20 @@ abstract class HistoryDatabaseAccess(
     ): Condition {
         var filter = this.organizationFilter(organization, orgService)
 
+        if (receivingOrgSvcStatus != null) {
+            filter = if (receivingOrgSvcStatus == CustomerStatus.ACTIVE) {
```
mkalish, line 233: Sorry if my previous comment was confusing, but I meant you could collapse all of this into one statement
```suggestion
           filter = filter.and(
                DSL.jsonbGetAttributeAsText(SETTING.VALUES, "customerStatus")
                    .eq(receivingOrgSvcStatus.name.lowercase())
            )
```

mkalish (commented):
Left one more comment, sorry if my previous one was confusing

prime-router/src/main/kotlin/history/azure/HistoryDatabaseAccess.kt
mkalish: Ah, ok didn't know that requirement...so I'm fine leaving this as is, but would argue that the API is now confusing.  I would not expect `TESTING` to come back if I pass `ACTIVE`.  I'm wondering if what you would want to do is support multiple values for this param? i.e.

`?receivingOrgSvcStatus=ACTIVE&receivingOrgSvcStatus=TESTING`

mkalish: If we want to keep it this way, I think we need to capture this behavior via a test.  I think this would be really easy to silently break

prime-router/docs/api/waters-delivery.yml
```diff
@@ -30,6 +30,26 @@ paths:
...
+          schema:
+            type: string
+        - name: fileName
+          in: query
+          description: the fileName within the deliveries of the organization and client
+          schema:
+            type: string
+        - name: receivingOrgSvcStatus
+          in: query
+          description: "Customer status of the service used by the organization."
```
mkalish, line 45: Nit: maybe a small update here indicating multiple values can be passed here

mkalish (approved):
:shipit:

victor-chaparro (commented):
I went through the Unit tests, but I didn't see a test for the sending both 'active' and 'testing'. Did I miss it?

prime-router/docs/api/waters-delivery.yml
penny-lischer: Will add

victor-chaparro (approved):
LGTM!

PR #13773, opened 2024-03-20
SFT to Device Mapping

brick-green:
This PR maps SFT to Device as described in the [ORU_R01](https://docs.google.com/spreadsheets/d/1gHK6_PFyr7PXns7wLDs0LSLsbjm0x-4bWUu3crXMKMI/edit#gid=0) mapping


prime-router/metadata/fhir_mapping/hl7/segments/MSH/Provenance-Source.yml
```diff
@@ -1,3 +1,8 @@
+# Entity.role is required if SFT exists. The mapping inventory has the following direction:
+#   - If the software does not represent the original source system then Entity.role=derivation
+#   - If the software does represent the original source system then Entity.role=source
+# Either source or derivation will need to be set for each sender
```
mkalish, line 4: Comment is a little confusing since we're hardcoding to "source"... is this saying that we would use a sender transform to change it?

prime-router/src/main/resources/metadata/hl7_mapping/ORU_R01/ORU_R01-base.yml
```diff
@@ -13,10 +13,17 @@ elements:
     schema: classpath:/metadata/hl7_mapping/resources/MessageHeader/MSH.yml
 
   - name: software-segment
-    condition: 'Bundle.entry.resource.ofType(MessageHeader).exists()'
+    resource: 'Bundle.entry.resource.ofType(Provenance).where(entity.exists()).entity.what.resolve()'
+    schema: classpath:/metadata/hl7_mapping/resources/Device/SFT.yml
+
+  - name: software-segment-legacy
```
mkalish, line 19: Can this not get dropped in favor of a sender transform?

mkalish (approved):
Nice work!

Two main comments, but I don't want to be a blocker so approving it:
- can we drop that "legacy"?
- Do we need a dedicated test for this mapping?

prime-router/metadata/fhir_mapping/hl7/segments/MSH/Provenance-Source.yml
brick-green: Oh yep, forgot about this. Removed this comment and added info to HL7v2-FHIR-Inventory.md

PR #13791, opened 2024-03-22
Map Clearing Bug

jorg3lopez:
This PR addresses the issue where the concurrency test fails when packaging the application. The reason is that a for loop is used to cycle through the keys of a map to delete them. Though the syntax is Kotlin, the under-the-hood collection and behavior are provided by java. 

Test Steps:
1. ./gradlew clean package

## Changes
- Use `.clear()` instead of iterator + `.remove(key)`

### Testing
- [x] Tested locally?


snesm (approved):

jack-h-wang (approved):
LGTM, all tests pass locally

arnejduranovic (approved):

PR #13792, opened 2024-03-22
13480 - Added conditional to receivingOrgSvcStatus

### LINKED ISSUE #13480: Add Receiver Status (active or inactive) to the API return call to ensure only active receivers will be shown on Daily Data

penny-lischer:
This PR ...

Adds conditional logic for the deliveries receivingOrgSvcStatus param.

Test Steps:
1. Both of these API calls should return data
`/api/waters/org/ca-dph/deliveries?sortDir=DESC&sortColumn=batchReadyAt&since=2023-11-30T00:00:00.000Z&until=2023-12-31T00:00:00.000Z&receivingOrgSvcStatus=ACTIVE&receivingOrgSvcStatus=TESTING`

`/api/waters/org/ca-dph/deliveries?sortDir=DESC&sortColumn=batchReadyAt&since=2023-11-30T00:00:00.000Z&until=2023-12-31T00:00:00.000Z`

### Testing
- [x] Tested locally?
- [ ] Added tests?

## Linked Issues
- Fixes #13480


victor-chaparro (approved):

prime-router/src/main/kotlin/history/azure/HistoryDatabaseAccess.kt
```diff
@@ -229,11 +229,13 @@ abstract class HistoryDatabaseAccess(
     ): Condition {
         var filter = this.organizationFilter(organization, orgService)
 
-        var statusList = receivingOrgSvcStatus?.map { it.name.lowercase() }
-        filter = filter.and(
+        if (receivingOrgSvcStatus != null) {
```
arnejduranovic, line 232: Are there tests that verify this new logic?

arnejduranovic (commented):
Changes look good but need to make sure we have tests

prime-router/src/main/kotlin/history/azure/HistoryDatabaseAccess.kt
penny-lischer: There are existing tests where if you do not pass this, the response status is validated.

arnejduranovic (approved):
Will add database tests for HistoryDatabaseAccess in another ticket

PR #13866, opened 2024-03-29
Platform/jimfuqian/11928 refactor code to remove old dependencies resurrect

### LINKED ISSUE #11928: Re-surrect-PR: Refactor code to remove old dependencies
### User Story
As a ReportStream engineer, I want to keep our dependencies current so that use of unmaintained code is minimized.
Note, this round of work is to re-surrect the PR which was reverted from last de[ployment.

### Description/Use Case
A number of dependencies (see Dev Notes below for a list) appear to be no longer maintained or public. We should remove these dependencies and refactor any code that uses them to use alternatives.

### Dev Notes
Mo supplied these notes:

- `com.github.kittinunf.fuel:fuel-jackson:2.3.1` - we use (used?) fuel to invoke web services and use Jackson to serialize data to and from JSON/Yaml. This library helped bridge those two. We could replace fuel with Ktor and yoink this one
- `com.github.KennethWussmann:mock-fuel:1.3.0` - ditto above. We needed this to mock our Fuel calls. I don't think this is maintained anymore 
- `com.jcraft:jsch:0.1.55` - older SSH/SFTP library that's also no longer maintained. This one can probably get pulled entirely (and should be) it's almost certainly not being used actively. Prune it.

Additionally, `mock-fuel` appears to be the only remaining use of `jitpack`, which should also be removed (see #11938).

### Acceptance Criteria
- [x] Dependency `com.github.kittinunf.fuel:fuel-jackson:2.3.1` is removed or an explanation for its continued use is provided.
- [x] Dependency `com.github.KennethWussmann:mock-fuel:1.3.0` is removed or an explanation for its continued use is provided.
- [x] Dependency `com.jcraft:jsch:0.1.55` is removed or an explanation for its continued use is provided.
- [x] Dependency `jitpack.io` is removed or an explanation for its continued use is provided.

thetaurean:
This PR:

Brings back a past branch that was unable to past server2server auth tests. The problem was an unregistered callback for handling a refresh token event. Original PR text below for clarity / information.

Removes unmaintained libraries around Fuel http: https://github.com/kittinunf/fuel and fuel mocks used with tests.
This removal improves report stream code base security posture.

the replacement is ktor http: https://github.com/ktorio/ktor

Test Steps:
1. Run full test suite to include server2server auth tests. 
  ```
   ./prime test --env staging --run=server2serverauth --key $AZURE_key
   ```

## Changes
- Changed gradle.build.kts to remove Fuel and all related dependencies including fuel mocks and added any additional dependencies required by ktor http
- Changed main code to use ktor http in place where Fuel http used
- Created HttpClientUtils.kt where GET, POST, PUT, HEAD, DELETE, SubmitForm are wrapped and exposed as handy utilities functions to the report stream main code, ktor http client "Content Negotiation" is configured with the ktor http client to promote clean code with regard to response body to serializable classes.
- Refactor tests accordingly
- Note: there were transports classes already using ktor http instead of Fuel before this PR, those are not touched, but tickets are logged for future consolidation
- Note, there are other http client used in the code base, e.g. azure http, those are not touched since there is no concern about the log term maintain etc.

### Testing
- [X] Tested locally?
- [X] Added tests? some old tests are replaced with tests using ktor mock engine

### Process

## Linked Issues
- Fixes #11928 

## To Be Done
- Investigate other uses of HttpClient and make sure refresh token is adequately handled #issue


prime-router/build.gradle.kts
```diff
@@ -449,16 +449,9 @@ tasks.register("testEnd2End") {
     finalizedBy("primeCLI")
 }
 
-tasks.register("testEnd2EndUP") {
```
jalbinson, line 452: Is this task no longer necessary?

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
```diff
@@ -304,26 +304,32 @@ class EmailScheduleEngine {
...
 
-                for (user in response.get().array()) emails.add(
+                val users = JSONArray(respStrJson)
+                for (user in users) emails.add(
                     (user as JSONObject).getJSONObject("profile").getString("email")
                 )
             }
         } catch (ex: Throwable) {
-            logger.warning("Error in fetching emails")
+            logger.warning("Error in fetching emails, exception: ${ex.message}")
```
jalbinson, line 332: move this to 

```kotlin
logger.warning("Error in fetching emails, exception", ex)
```

So it triggers errors and we catch that something is amiss.

prime-router/build.gradle.kts
```diff
@@ -957,5 +933,8 @@ dependencies {
     testImplementation("org.testcontainers:junit-jupiter:1.19.6")
     testImplementation("org.testcontainers:postgresql:1.19.6")
 
+    testImplementation("io.ktor:ktor-server-test-host:$ktorVersion")
+    testImplementation("org.jetbrains.kotlin:kotlin-test:$kotlinVersion")
+    testImplementation("com.github.stefanbirkner:system-lambda:1.2.1")
```
jalbinson, line 938: Why did we add these libraries?

Generally anything scoped to just tests is not a big deal but it adds more dependencies to keep up to date which we're specifically trying to reduce in this ticket.

prime-router/src/test/kotlin/cli/LookupTableCommandsTest.kt
```diff
@@ -86,116 +158,84 @@ class LookupTableCommandsTest {
...
-        assertFailsWith<IOException>(
-            block = {
-                LookupTableEndpointUtilities.checkCommonErrorsFromResponse(mockResultFailure, mockResponse)
-            }
+        assertTrue(
+            exception.message.toString().contains(
+                "New Lookup Table Table is identical to existing table version 1"
+            )
         )
+    }
```
jalbinson, line 178: Its better to assert the exception type than the message within

prime-router/src/main/kotlin/common/HttpClientUtils.kt
```diff
@@ -0,0 +1,582 @@
...
+            url: String,
+            tokens: BearerTokens? = null,
+            headers: Map<String, String>? = null,
+            acceptedContent: ContentType? = ContentType.Application.Json,
+            timeout: Long? = REQUEST_TIMEOUT_MILLIS,
+            queryParameters: Map<String, String>? = null,
+            jsonPayload: String? = null,
+            httpClient: HttpClient? = null,
+        ): HttpResponse {
+            return runBlocking {
```
jalbinson, line 499: We have a single threaded model at the moment but in general it would be better to make these functions async.

```kotlin
     private suspend fun invoke() {
     // runBlocking no longer needed and will be handled at the caller (or not if we're doing things the right way)
```

prime-router/src/main/kotlin/common/HttpClientUtils.kt
```diff
@@ -0,0 +1,582 @@
...
+                        bearer {
+                            loadTokens {
+                                bearerTokens
+                            }
+                            refreshTokens {
+                                bearerTokens
+                            }
+                        }
+                    }
+                }
```
jalbinson, line 557: Since this does a ton under the hood it would probably be easier to just nix it and add the header by hand.

```kotlin
bearerTokens?.accessToken?.let {
    headers {
        append("Authorization", "Bearer $it")
    }
}
```

prime-router/build.gradle.kts
thetaurean: Not sure why it was removed. I recall discussion around merging the two -- testEnd2End and testEnd2EndUP -- but I haven't found anything concrete there so I've added it back in for now.

prime-router/build.gradle.kts
thetaurean: Nice catch! These were not necessary. I've reached out to Jim to find out why they were added but for now I've removed them.

prime-router/src/test/kotlin/cli/LookupTableCommandsTest.kt
thetaurean: There's an assertion above on the exception type. This assertion is further verifying the message content which seems appropriate in the context. We should make that exception unique but feels out of scope for now.

prime-router/src/main/kotlin/common/HttpClientUtils.kt
thetaurean: Changing this litters the code with `runBlocking {}` calls. I suggest we cross this bridge when we get there, it seems be simpler to transition to multithread as currently written.

prime-router/src/main/kotlin/azure/EmailEngineFunction.kt
thetaurean: I wasn't able to chain that exception to `logger.warn`, maybe I'm missing something?

prime-router/src/main/kotlin/common/HttpClientUtils.kt
jalbinson: I agree. It would be nice to at some point to correctly handle coroutines but this is not the ticket to do it in. I was just bringing this design decision up.

prime-router/src/main/kotlin/common/HttpClientUtils.kt
jalbinson: I think thats a premature optimization though. We currently don't use refresh tokens and it caused us an issue. I think these few lines are what we actually want to do at the moment and if we do want to use refresh tokens in the future it will be easy to find where the auth header is currently being appended.

thetaurean: Yeah you're right. Copacetic :).

jalbinson (approved):
Looks good. I'd recommend getting another pair of eyes on this before you merge though since the changes are so far reaching.

prime-router/src/main/kotlin/cli/LookupTableCommands.kt
```diff
@@ -56,7 +49,11 @@ import java.time.Instant
  * If a [useThisToken] is not specified, then attempt to get a token from Otka.
  * Otherwise, [useThisToken] is sent as the bearer token to the ReportStream server.
  */
-class LookupTableEndpointUtilities(val environment: Environment, val useThisToken: String? = null) {
+class LookupTableEndpointUtilities(
+    val environment: Environment,
+    val useThisToken: String? = null,
+    val httpClient: HttpClient? = null,
```
arnejduranovic, line 55: I don't understand why we need this now?

prime-router/src/main/kotlin/transport/GAENTransport.kt
```diff
@@ -41,7 +41,7 @@ import java.io.ByteArrayInputStream
  * See [Google API for Exposure Notification](https://https://developers.google.com/android/exposure-notifications/exposure-notifications-api) for details on GAEN.
  * See [Issue API](https://https://github.com/google/exposure-notifications-verification-server/blob/main/docs/api.md#apiissue) for the details on the issue API ReportStream calls.
  */
-class GAENTransport : ITransport, Logging {
+class GAENTransport(val httpClient: HttpClient? = null) : ITransport, Logging {
```
arnejduranovic, line 44: I don't understand this change. httpClient is only used once below and it is ignore and overwritten?

prime-router/src/main/kotlin/credentials/HashicorpVaultCredentialService.kt
```diff
@@ -1,44 +1,76 @@
...
+    }
 
-        // Initialize the API token used to secure requests
-        manager.baseHeaders = mapOf("X-Vault-Token" to VAULT_TOKEN)
+    /**
+     * object specific impl - also adapted to unit tests
+     */
+    fun fetchCredentialHelper(
+        connectionId: String,
+        httpClient: HttpClient? = null,
```
arnejduranovic, line 27: I don't understand this change. httpClient is only used once below and it is ignore and overwritten?

arnejduranovic (approved):

PR #13886, opened 2024-04-01
BUG: No way to control date range for TableFilters component

### LINKED ISSUE #13885: BUG: No way to control date range for TableFilters component
Currently, TableFilters does not use the date range data from the FilterManager prop to initialize internal date range variables (initializing to undefined instead). This causes the default values for the date range components to use the current local datetime. Tests that render TableFilters and test the presented range dates are failing once the current month changes (and would technically fail on year change). TableFilters needs to accept a provided default range. However, we do not want to stray from current designer-approved user experiences with date ranges (meaning for now we need to check if the FilterManager ranges are the year 2000-3000 fallbacks, and initialize to undefined to imitate current live behavior). This will allow tests against custom range scenarios to pass while fallback scenarios are unchanged.

# Acceptance Criteria
- [] TableFilters uses the ranges from FilterManager unless they're the fallbacks
- [] Tests pass using predefined custom date ranges
- [] UI/UX on pages unchanged

jpandersen87:
Fixes #13885

- TableFilters uses FilterManager date range if not fallback
- mock -> fixtures
- fallback variable names made clearer
- TestTable folded into table test file
- new retries added to usepagination tests
- test env timezone set to UTC via globalSetup file so as not to require being set before running vitest binary (ex: IDEs preconfigured to run vitest bin directly instead of using package.json test script)


penny-lischer (approved):
LGTM

PR #13905, opened 2024-04-02
Handle case where Observation.device has values that don't map to HL7

mkalish:
This PR updates the mapping for OBX.18 to support what SR is currently sending and avoiding generating an empty repeating component

Test Steps:
CI covers this

## Changes
- Addes a condition for mapping OBX.18 to only map if value exists
- Updates the HL7 equality check to use encodePreservingChars

### Testing
- [x] Tested locally?
- [x] Added tests?

### Process


prime-router/src/main/kotlin/fhirengine/translation/hl7/FhirToHl7Converter.kt
```diff
@@ -91,7 +92,7 @@ class FhirToHl7Converter(
     }
 
     override fun checkForEquality(converted: Message, expectedOutput: Message): Boolean {
-        return converted.encode() == expectedOutput.encode()
```
mkalish, line 94: Not necessarily related to the fix, but noticed that this wasn't using the extension method when debugging this

prime-router/src/main/resources/metadata/hl7_mapping/resources/Observation/OBX.yml
```diff
@@ -59,6 +59,7 @@ elements:
 
   - name: obx-equipment-instance-identifier
     resource: '%resource.device.resolve().identifier.union(%resource.extension(%`rsext-obx-observation`).extension.where(url = "OBX.18").tail().value.resolve().identifier)'
+    condition: '%resource.value.exists() or %resource.extension(%`rsext-assigning-authority`).exists()'
```
mkalish, line 62: This condition handles us getting sent FHIR like so 
```
{
    "fullUrl" : "Device/1710881998384474000.9752cf60-d1b6-4cbe-98b2-9984718ee4ed",
    "resource" : {
      "resourceType" : "Device",
      "id" : "1710881998384474000.9752cf60-d1b6-4cbe-98b2-9984718ee4ed",
      "identifier" : [ {
        "extension" : [ {
          "url" : "https://reportstream.cdc.gov/fhir/StructureDefinition/assigning-authority",
          "extension" : [ {
            "url" : "https://reportstream.cdc.gov/fhir/StructureDefinition/namespace-id",
            "valueString" : "Equipment"
          }, {
            "url" : "https://reportstream.cdc.gov/fhir/StructureDefinition/universal-id",
            "valueString" : "1.111.122311.222.44.2.3.3"
          }, {
            "url" : "https://reportstream.cdc.gov/fhir/StructureDefinition/universal-id-type",
            "valueCode" : "ISO"
          } ]
        } ],
        "value" : "CDC_3015822725"
      } ]
    }
  }
```
where the second identifier is missing a value

arnejduranovic (approved):

prime-router/src/main/resources/metadata/hl7_mapping/resources/Observation/OBX.yml
brick-green: I'm only seeing a one identifier in this snippet. Where's the second identifier? Also the resource on the element references `extension.where(url = "OBX.18")` but I don't see that extension in the test FHIR data. Is that where the second identifier should come from?

mkalish: so `%resource.device.resolve().identifier` will grab all the identifiers on the device and then map them, so in this case there are no values in the OBX.18 extension.

That case actually only comes up when mapping from HL7->FHIR->HL7 since `device` has cardinality of 1

brick-green (approved):
lgtm

PR #13939, opened 2024-04-05
Bug: create blob endpoint from suffix and account name

mkalish:
This PR accounts for the the fact that azure connection strings have a different shape local vs. the higher environments. 

Test Steps:
1. CI should cover this

## Changes
- Updates the `getBlobEndpoint` method to look for account name and endpointsuffix in the higher environments

### Testing
- [x] Tested locally?
- [x] Added tests?

### Process

### DIFF
--- a/prime-router/src/main/kotlin/azure/BlobAccess.kt
+++ b/prime-router/src/main/kotlin/azure/BlobAccess.kt
@@ -82,6 +82,8 @@ class BlobAccess() : Logging {
              * Regex used to extract the URL for blob storage
              */
             private val blobEndpointRegex = Regex(";BlobEndpoint=(?<blobEndpoint>[^;]+);")
+            private val blobAccountNameRegex = Regex(";AccountName=(?<accountName>[^;]+);")
+            private val blobEndpointSuffixRegex = Regex(";EndpointSuffix=(?<endpointSuffix>[^;]+);?")
 
             /**
              * Builds a [BlobContainerMetadata] object. [envVar] will be resolved to the blobstore connection string.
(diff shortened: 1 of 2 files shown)


prime-router/src/main/kotlin/azure/BlobAccess.kt
```diff
@@ -105,10 +107,18 @@ class BlobAccess() : Logging {
...
-            return "$blobStorageUrl/$containerName"
+            val blobEndpointMatch = blobEndpointRegex.find(connectionString)
+            val blobAccountNameMatch = blobAccountNameRegex.find(connectionString)
+            val blobEndpointSuffixMatch = blobEndpointSuffixRegex.find(connectionString)
+            val blobStorageUrl = blobEndpointMatch?.groups?.get("blobEndpoint")?.value
+            val blobAccountName = blobAccountNameMatch?.groups?.get("accountName")?.value
+            val blobEndpointSuffix = blobEndpointSuffixMatch?.groups?.get("endpointSuffix")?.value
+            if (blobStorageUrl != null) {
+                return "$blobStorageUrl/$containerName"
+            } else if (blobAccountName != null && blobEndpointSuffix != null) {
```
mkalish, line 118: I was not sure if including the endpoints in the source code was no good from a security front.  We could also consider following up and adding another ENV var that has this value, since it can be derived from the connection string it might not be worth it

prime-router/src/test/kotlin/azure/BlobAccessTests.kt
```diff
@@ -448,6 +448,17 @@ class BlobAccessTests {
         assertThat(metadata.getBlobEndpoint()).isEqualTo("http://localhost:54321/devstoreaccount1/test")
     }
 
+    @Test
+    fun `test get blob endpoint when accountname and suffix are used`() {
+        val endpoint =
+            """
+                DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=keydevstoreaccount1;EndpointSuffix=core.windows.net
```
mkalish, line 455: I confirmed this shape by looking at what is the staging app configuration

arnejduranovic: Only comment is perhaps name the vars `deployedEnvEndpoint` and `localEnvEndpoint` or something more descriptive.

arnejduranovic (approved):

prime-router/src/main/kotlin/azure/BlobAccess.kt
```diff
@@ -105,10 +107,18 @@ class BlobAccess() : Logging {
          * @return the URL for the storage and container that this object represents
          */
         fun getBlobEndpoint(): String {
-            val match = blobEndpointRegex.find(connectionString)
-            val blobStorageUrl = match?.groups?.get("blobEndpoint")?.value
-                ?: throw RuntimeException("Connection string is misconfigured and does not contain a blob endpoint URL")
-            return "$blobStorageUrl/$containerName"
+            val blobEndpointMatch = blobEndpointRegex.find(connectionString)
```
jalbinson, line 110: Maybe I'm just a regex hater but I feel like it would simpler to just parse the string into a map.

Something simple like this
```kotlin
val map = string
  .split(";")
  .map { it.split("=") }
  .associate { it.first() to it.last() }

val blobEndpoint = map["BlobEndpoint"]
// etc
```

jalbinson (approved):

PR #13972, opened 2024-04-08
Add Ingestion Time to the Deliveries Endpoint

halprin:
This PR adds root report IDs and ingestion times to the Delivery API.

Test Steps:
1. Have a receiver with batching enabled.
2. Send a report into RS twice in a row.  The sender/receiver config should be such that these reports will be delivered to the aforementioned receiver.
3. Call the `/waters/report/{id}/delivery` API, using the send pipeline step reportId.
4. See the new `originalIngestion` field that includes all the possible root `reportId`s and their ingestion times (`createdAt`).

## Changes
- Added a new field to the `DeliveryHistory` class to contain the ingestion report IDs and times.
- Created a new `getRootReports` (notice the plurality) method, and added it to `ReportGraph` and `ReportService`.  These methods return a list of `ReportFile`s and work even when given a `reportId` from the batch or send step.
- Call the new `getRootReports` and extract the `reportId` and `createdAt` and put this information in the new `DeliveryHistory` field.
- Some kick ass unit tests.

### Testing
- [x] Tested locally?
- [x] Added tests?

### Process
- [x] Database changes are submitted as a separate PR?

## Linked Issues

- Helps address https://github.com/CDCgov/trusted-intermediary/issues/989


prime-router/src/main/kotlin/history/azure/DeliveryFacade.kt
```diff
@@ -95,11 +97,17 @@ class DeliveryFacade(
...
     ): DeliveryHistory? {
-        return dbDeliveryAccess.fetchAction(
+        val deliveryHistory = dbDeliveryAccess.fetchAction(
             deliveryId,
             orgName = null,
             DeliveryHistory::class.java
         )
+        val reportId = deliveryHistory?.reportId
+        if (reportId != null) {
+            val ingestionTime = reportService.getRootReport(UUID.fromString(reportId)).createdAt
```
arnejduranovic, line 107: @halprin This is incorrect conceptually -- apologies if I forgot to mention this in our last meeting (I had it in my notes but not sure if I articulated it). 

You can not call "getRootReport" in the delivery function because the reportId that ultimately gets passed in may have MORE THAN ONE ROOT REPORT due to ReportStream's batching feature. The areas in the code where "getRootReport" is called today are guaranteed to only have one root report because those areas are called before batching happens (like the Translation function).

So, one possible solution is to implement the following:

- `getRootReports` - returns one or more root reports
- `DeliveryHistory` - existing class has a new property called `originatingReports` or something that contains the following properties: `ingestionTime`, `reportId`

Please think through the above and how it would work with your product requirements. Let's discuss on call if needed!

arnejduranovic (changes requested):

prime-router/src/main/kotlin/history/azure/DeliveryFacade.kt
halprin: Thanks for the feedback, @arnejduranovic!  We addressed it.  Please re-review.  :)

prime-router/src/main/kotlin/history/azure/DeliveryFacade.kt
```diff
@@ -95,11 +97,22 @@ class DeliveryFacade(
...
             orgName = null,
             DeliveryHistory::class.java
         )
+        val reportId = deliveryHistory?.reportId
+        if (reportId != null) {
+            val roots = reportService.getRootReports(UUID.fromString(reportId))
+            deliveryHistory.originalIngestion = roots.map {
+                mapOf(
+                    "reportId" to it.reportId,
+                    "ingestionTime" to it.createdAt,
```
arnejduranovic, line 111: While perhaps not immediately required, I think the following would also be useful for the user:

`"sendingOrg" to it.sendingOrg`

halprin: Done, @arnejduranovic!

arnejduranovic (approved):

arnejduranovic:
@penny-lischer @jpandersen87 Just FYI this PR adds a new property to the deliveryHistory object that is returned.

Issue #13796, opened 2024-03-22
Integrate NIST Profile Validation to UP


arnejduranovic:
### User Story
As a report stream sender,
I would like ReportStream to validate my message against the profile I specify in MSH 21,
so I can ensure receivers receive valid RadxMars messages

### Description/Use Case
Implement the RadxMars [software requirements document](https://cdc.sharepoint.com/:w:/r/teams/ReportStream/Shared%20Documents/Engineering/Design/SRD-RadxMars-OTC.docx?d=wfb06850638704ddb896ef6f02a1428ef&csf=1&web=1&e=qzVVGh)  sections:

- III. Routing support - essentially just create the following topic: `mars-otc-elr`.
- V. Validation support - logic on how, where, and when we apply validation

### Acceptance Criteria
- [ ] All requirements in sections III, and V of SRD implemented
- [ ] NIST validation library integrated to ReportStream
- [ ] Sender validation occurs on HL7v2 messages in Convert Function after de-batching

Andrey-Glazkv:
Messages we got were not valid

PR #14016, opened 2024-04-10
13796: write convert tests that document current behavior

mkalish:
This PR adds additional test cases that document the differing behavior processing bulk FHIR vs HL7 messages

Test Steps:
n/a

## Changes
- Adds new tests that document what happens when batches of FHIR or HL7 reports with a mix of valid and invalid items go through the convert step

### Testing
- [x] Tested locally?
- [x] Added tests?

### Process

## Linked Issues
- #13796


prime-router/src/test/kotlin/fhirengine/azure/FhirFunctionIntegrationTests.kt
```diff
@@ -383,6 +462,303 @@ class FhirFunctionIntegrationTests() {
         }
         verify(exactly = 1) {
             QueueAccess.sendMessage(elrRoutingQueueName, any())
+            BlobAccess.uploadBody(Report.Format.FHIR, any(), any(), any(), any())
+        }
+    }
+
+    @Test
+    fun `test successfully processes a convert message for bulk HL7 message`() {
+        val validBatch = cleanHL7Record + "\n" + invalidHL7Record
```
arnejduranovic, line 471: So this batch's second item is invalid but it still goes through because the particular record its not considered invalid by the parser?

arnejduranovic (approved):

arnejduranovic:
@GilmoreA6 @victor-chaparro @chris-kuryak tests that show that:

- If a batched FHIR Report has "invalid" items, the valid items will get through
- If a batched HL7 Report has "invalid" items, no items will get through

The parser is weird and so "valid" and "invalid" depends on the parser's implementation. This is why you can see in the first test that HL7 batch makes it through with an invalid record because the parser does not consider that particular record invalid.

arnejduranovic (approved):

prime-router/src/test/kotlin/fhirengine/azure/FhirFunctionIntegrationTests.kt
mkalish: Yep "invalid" in this case means having a segment that does not exist in ORU_R01

PR #14020, opened 2024-04-11
Fix duplicate auth header

thetaurean:
This PR fixes a bug where "Bearer" is errantly prepended to an access token.

Test Steps:
1. Run test suite to include server2server auth tests against staging

## Changes
- Adds the ktor auth handler dependency and uses it to perform bearer auth
  - Registers the refresh callback to prevent clearing tokens
 
### Testing
- [x] Tested locally?

## Linked Issues
- Fixes #issue


arnejduranovic (commented):
Test steps say to run auth to auth tests against staging but those were passing prior to this change, correct? If we are making this change in response to a defect, it sounds like we are missing some tests?

arnejduranovic (approved):

JFisk42 (approved):

PR #14152, opened 2024-04-22
Adding function endpoints for calling the intermediary

somesylvie:
This PR adds metadata endpoints to the History and Delivery functions, which call the intermediary to retrieve this data
https://github.com/CDCgov/trusted-intermediary/issues/690

Test Steps:
1. Create test data - Submit reports that will go to the trusted intermediary
2. Using the IDs created in the previous steps, query the `waters/report/{id}/delivery/etorMetadata` and `waters/report/{id}/history/etorMetadata` with the relevant IDs

## Changes
- Add new TI Metadata endpoints to delivery and submission functions
- Add a shared function to ReportFileFunction to handle retrieving metadata from TI
- Add tests
- Add terraform env variables for the TI base URL
- Modified ReportGraph to retrieve descendent reports correctly
- Added helper query to ReportGraph to support other changes

### Testing
- [x] Tested locally?
- [x] Added tests?

### Process
- [x] Are there licensing issues with any new dependencies introduced? - added `kotlinx-coroutines-core` to enable unit testing coroutines. This is an Apache 2.0 license and an official Kotlin product
- [x] Database changes are submitted as a separate PR? - NA
- [x] DevOps team has been notified if PR requires ops support? - NA

    - `waters/report/{id}/delivery/etorMetadata`
    - `waters/report/{id}/history/etorMetadata`
    - These endpoints use the existing auth structure in their respective files
    - No
    - No
    - No, but added a new environment variable in terraform
    - No
    - No

    -  The new endpoints leverage the existing auth infrastructure


prime-router/build.gradle.kts
```diff
@@ -112,6 +112,7 @@ val env = mutableMapOf<String, Any>(
     "VAULT_API_ADDR" to "http://localhost:8200",
     "SFTP_HOST_OVERRIDE" to "localhost",
     "SFTP_PORT_OVERRIDE" to "2222",
+    "ETOR_TI_baseurl" to "http://host.docker.internal:8080",
```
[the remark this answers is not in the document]
arnejduranovic: Also, I think TI-integration should be optional. That means our startup scripts should not automatically spin up an instance of TI (which it doesn't) but that also means not including `ETOR_TI_baseurl` in gradle or docker-compose. TI functions should assert the ETOR_TI integration is properly configured before doing its logic.

prime-router/src/main/kotlin/history/azure/ReportFileFunction.kt
```diff
@@ -190,6 +204,75 @@ abstract class ReportFileFunction(
...
+        return HttpUtilities.okResponse(request, responseBody)
+    }
+
+    /**
+     * Use the specified report ID to find the lookup ID for the ETOR TI to use to retrieve metadata
+     *
+     * @param id DB Action that we are reviewing
+     * @return the string lookup ID if found, otherwise and empty string
+     */
+    abstract fun getLookupId(id: String): String
```
arnejduranovic, line 274: It may be cleaner instead of this abstract method to have `retrieveMetadata` take in a function parameter for this lookup id logic.

arnejduranovic: Also, rename `id` to `reportId` if that's what you are expecting here. Ditto for the other places in this changeset.

prime-router/src/main/kotlin/history/azure/SubmissionFunction.kt
```diff
@@ -126,4 +128,31 @@ class SubmissionFunction(
...
+        ) request: HttpRequestMessage<String?>,
+        @BindingName("id") id: String,
+        context: ExecutionContext,
+    ): HttpResponseMessage {
+        return this.retrieveMetadata(request, id, context, null)
+    }
+
+    override fun getLookupId(id: String): String {
+        val submissionActionId = this.actionFromId(id)
+        val submissionHistory = submissionsFacade.findDetailedSubmissionHistory(submissionActionId)
```
arnejduranovic, line 148: It seems like it would be cleaner to use logic similar to findDetailedDeliveryHistory here, where we simply, in this case, walk DOWN the tree (instead of up like in findDetailedDeliveryHistory). ReportGraph.kt should ultimately be used by both of your endpoints because it is more straight forward.

So, I am suggesting a new method in ReportGraph called something like "getDescendentReports" and "getDescendentReport" that uses `reportDescendantGraphCommonTableExpression`

Also, in findDetailedDeliveryHistory, why not just pass in the reportID?

It seems the logic in the change set related to getLookupId could be greatly simplified if the above is implemented.

somesylvie: Just to check, after we create a new method in ReportGraph, are you suggesting having SubmissionFunction call ReportGraph directly? Other methods in SubmissionFunction call SubmissionFacade which calls ReportService which calls ReportGraph

prime-router/src/main/kotlin/history/azure/ReportFileFunction.kt
jcrichlake: We've refactored the id parameter name, that's a good catch.  We've observed the established pattern in report stream is to use an abstract function for example `actionIsValid` or `singleDetailedHistory` in SubmissionFunction.  Are you sure you want us to stray from the established pattern?

prime-router/build.gradle.kts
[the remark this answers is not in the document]
jcrichlake: We've added the endpoints to the .yml API files.  Would it be possible to leave the environment variable in build.gradle and the docker-compose files? Currently it shouldn't impact any of the report stream developers and it would prevent the intermediary developers from having to manually add this environment variable every time we need to to do development or testing on one of these two endpoints.  Additionally we would avoid running a check for the environment variables existence in each of the deployed API calls for a value that absolutely should be there.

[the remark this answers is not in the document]
jcrichlake: The two bullet points for documentation have been addressed

mkalish:
Example for handling transactions

```kt
// this goes  right as the beginning of your azure function
workflowEngine.db.transactReturning { txn ->
 /// all downstream callers take in the txn
val report = reportGraph.getDescendantReports(childId: UUID, txn: DataAccessTransaction)
}
```

I'm trying to convert places over to use this pattern, so no worries about using code that does it wrong just trying to avoid having more examples get added.

prime-router/docs/api/delivery.yml
```diff
@@ -41,6 +41,29 @@ paths:
           description: The search request could not be parsed
         '404':
           description: The receiver could not be found
+  /waters/report/{reportId}/delivery/etorMetadata:
+    get:
+      summary: Retrieves ETOR metadata for a report
+      description: Allows fetching of ETOR intermediary metadata by report ID
```
arnejduranovic, line 47: The summary and description for this and `/waters/report/{reportId}/history/etorMetadata` are the same, can we be more detailed in each case?

prime-router/docs/design/design/rs-ti-integration.md
```diff
@@ -0,0 +1,13 @@
+# Report Stream and Intermediary Integration
+
+## ETOR
+
+For the purpose of supporting our partners on the ETOR initiative, there is a need to make certain data available to our
```
arnejduranovic, line 5: Can we be more detailed? What is the "certain data" and why was a new endpoint needed? Why do the existing delivery and history endpoints not cut it?

prime-router/src/main/kotlin/history/azure/DeliveryFunction.kt
```diff
@@ -188,6 +190,49 @@ class DeliveryFunction(
...
+        context: ExecutionContext,
+    ): HttpResponseMessage {
+        return this.retrieveETORIntermediaryMetadata(request, reportId, context, null)
+    }
+
+    /**
+     * Function for finding the associated report ID that the intermediary knows about given the report ID that the receiver is
+     * given from report stream
+     */
+    override fun getLookupId(reportId: UUID): UUID? {
```
arnejduranovic, line 218: Can you please explain the implementation of this method? In your use-case, there should only ever be one root report because we will not be batching messages to receivers, correct? IF we are batching messages to receivers, then there will be multiple root reports, but I fail to see how the oldest one would necessarily be the correct one?

This comment is based on my understanding that `getRootReports` returns reports that do not have a parent (If receivers have batching turned off this will never be the case because each report containing a single message can only ultimately come from a single submitted report). Based on the logic I see here, I think your understanding is that it returns ALL ancestors of a report?

That said, this logic could simply be `return reportGraph.getRootReport(reportId)` because getRootReport will throw tooManyRows exception if there is more than one result. Then, I recommend updating `retrieveETORIntermediaryMetadata` to take in the single line logic as a function param instead of the current `override fun getLookupId` pattern.

arnejduranovic: BONUS: Notice the description of getRootReport:
```
     * Recursively goes up the report_linage table from any report until it reaches
     * a report with an action type of "receive" (the root report)
     *
     * This will return null if no report with action type "receive" is present or if
     * the root is passed in
```

prime-router/src/main/kotlin/history/azure/ReportFileFunction.kt
```diff
@@ -190,6 +205,85 @@ abstract class ReportFileFunction(
...
+                    restTransportInfo,
+                    reportId.toString(),
+                    jksCredential,
+                    credential,
+                    logger
+                )
+            }
+        }
+
+        val lookupId = this.getLookupId(reportId)
```
arnejduranovic, line 248: Intellij recommends:

        val lookupId = this.getLookupId(reportId) ?: return HttpUtilities.notFoundResponse(request, "lookup Id not found")

prime-router/src/main/kotlin/history/azure/SubmissionFunction.kt
```diff
@@ -126,4 +131,42 @@ class SubmissionFunction(
     ): HttpResponseMessage {
         return this.getDetailedView(request, id)
     }
+
+    /**
+     *Endpoint for intermediary receivers to verify status of messages. It passes
```
arnejduranovic, line 136: This comment is the same as the comment for `DeliveryFunction.getEtorMetadata`. Can we be more specific in each description so its easier to tell what the difference between the two is? For example, are both actually for "intermediary `receivers` to verify status of messages" or is one supposed to say `senders`?

prime-router/src/main/kotlin/history/azure/SubmissionFunction.kt
```diff
@@ -126,4 +131,42 @@ class SubmissionFunction(
...
+    }
+
+    /**
+     * Function for finding the associated report ID that the intermediary knows about given the report ID that the sender is
+     * given from report stream
+     */
+    override fun getLookupId(reportId: UUID): UUID? {
+        val reportGraph = ReportGraph(workflowEngine.db)
+        var descendants: List<ReportFile> = emptyList()
+        workflowEngine.db.transactReturning { txn ->
```
arnejduranovic, line 163: Why is this `getLookupId` using `transactReturning` but the one in `DeliveryFunction` isn't? `getRootReports` uses `transactReturning`, why are we doing it differently with `getDescendantReports`?

prime-router/src/main/kotlin/history/db/ReportGraph.kt
```diff
@@ -144,6 +145,24 @@ class ReportGraph(
...
+    /**
+     * Recursively goes down the report_linage table from any report until it reaches
+     * all descendant reports with the specified action type(s)
+     *
+     * This will return an empty list if no report with the specified action type is present or if
+     * the ID of the final descendant is passed in
+     *
+     * If the passed in report ID has multiple descendant reports, they will all be returned
+     */
+    fun getDescendantReports(
```
arnejduranovic, line 157: Nice

arnejduranovic (changes requested):

prime-router/docs/api/delivery.yml
halprin: Fixed.

prime-router/docs/design/design/rs-ti-integration.md
halprin: Fixed.

prime-router/src/main/kotlin/history/azure/ReportFileFunction.kt
halprin: Fixed.

prime-router/src/main/kotlin/history/azure/SubmissionFunction.kt
somesylvie: Updated to say `senders` in SubmissionFunction

prime-router/src/main/kotlin/history/azure/DeliveryFunction.kt
halprin: Yes, you are correct.  We technically don't support batching as of this moment and would require additional work on our end to support in the future.  Given that, we have migrated to using `getRootReport` (singular).

halprin: I also have an [outstanding question to you in Slack](https://flexion.slack.com/archives/C048S3GPSQG/p1714510584647929?thread_ts=1714503553.513999&cid=C048S3GPSQG) about changing the pattern for `getLookupId`.

prime-router/src/main/kotlin/history/azure/SubmissionFunction.kt
halprin: We originally did it the same as `getRootReports` and `getRootReport`, but @mkalish [recommended](https://github.com/CDCgov/prime-reportstream/pull/14152#issuecomment-2083631465) that we do it this new way for the new `getDescendantReports` method.  We decided to not touch the pre-existing `getRootReports` or `getRootReport` methods.

mkalish: @arnejduranovic We should probably open a ticket to fix that one spot too (I think it got added recently), but the transaction getting created there makes the function unsafe to be called since you can use it while an existing transaction is in place.

prime-router/build.gradle.kts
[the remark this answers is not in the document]
mkalish: In the spirit of this being an open source project, I tend to agree with arnej that we shouldn't add new code/configuration that won't work for everyone.  Let me know if you want some help brainstorming how to make it easier to spin up locally, but a few things you could do:

- new gradle task that sets up the environment
- modify the local.env and just leave it there

I do think we should be checking for the service being available.  As written a new developer ramping up here would hit the endpoint and after a timeout get a strange error message

prime-router/src/main/kotlin/history/azure/DeliveryFunction.kt
```diff
@@ -188,6 +189,41 @@ class DeliveryFunction(
...
+        // the delivery endpoint is called by the final receiver with a sent report ID, where TI
+        // knows about the related submission report ID
+
+        val reportGraph = ReportGraph(workflowEngine.db)
+        val root = reportGraph.getRootReport(reportId)
```
mkalish, line 222: I think we can sub this out for `ReportService.getRootReport`

prime-router/src/main/kotlin/history/azure/DeliveryFunction.kt
```diff
@@ -188,6 +189,41 @@ class DeliveryFunction(
...
+
+        return root?.reportId
```
mkalish, line 224: This seems like an exceptional case that should be at least logged if not generate an error message right?  Isn't this where we delivered a report but then there is record of it getting submitted?

prime-router/src/main/kotlin/history/azure/ReportFileFunction.kt
```diff
@@ -190,6 +205,83 @@ abstract class ReportFileFunction(
...
+        val client = if (engine == null) HttpClient() else HttpClient(engine)
+        val restTransportInfo = receiver?.transport as RESTTransportType
+        val (credential, jksCredential) = RESTTransport().getCredential(restTransportInfo, receiver)
+        val logger: Logger = context.logger
+        var authPair: Pair<Map<String, String>?, String?> =
+            Pair(null, null)
+
+        var responseBody = ""
+
+        runBlocking {
```
mkalish, line 236: Nit: I think the more kotlinnier way might be something like and avoid having the mutable state

```suggestion
         val authPair = runBlocking {
            async {
                authPair = RESTTransport().getOAuthToken(
                    restTransportInfo,
                    reportId.toString(),
                    jksCredential,
                    credential,
                    logger
                )
            }.await()
        }
```

prime-router/src/main/kotlin/history/azure/ReportFileFunction.kt
```diff
@@ -190,6 +205,83 @@ abstract class ReportFileFunction(
...
+                    logger
+                )
+            }
+        }
+
+        val lookupId = this.getLookupId(reportId)
+            ?: return HttpUtilities.notFoundResponse(request, "lookup Id not found")
+
+        var status: HttpStatusCode = HttpStatusCode.NotFound
+        runBlocking {
```
mkalish, line 252: Same nit here on using `async` rather than.

prime-router/src/main/kotlin/history/db/ReportGraph.kt
```diff
@@ -360,26 +379,44 @@ class ReportGraph(
...
+     * Fetches all descendant report records in a recursive manner.
+     *
+     * @param txn the data access transaction
+     * @param cte the common table expression for report lineage
+     * @return the descendant report records
+     */
+    private fun descendantReportRecords(
+        txn: DataAccessTransaction,
+        cte: CommonTableExpression<Record1<UUID>>,
+        searchedForTaskActions: Set<TaskAction>,
```
mkalish, line 411: Nice!

prime-router/src/test/kotlin/history/azure/DeliveryFunctionTests.kt
```diff
@@ -636,6 +668,164 @@ class DeliveryFunctionTests : Logging {
...
+        action.sendingOrg = organizationName
+        action.actionName = TaskAction.send
+
+        mockkConstructor(ReportGraph::class)
+
+        val firstReport = ReportFile()
+        firstReport.reportId = UUID.randomUUID()
+        firstReport.createdAt = OffsetDateTime.parse("2023-04-18T23:36:00Z")
+
+//        val secondReport = ReportFile()
```
mkalish, line 708: Some commented code here

prime-router/src/main/kotlin/history/azure/ReportFileFunction.kt
```diff
@@ -190,6 +205,83 @@ abstract class ReportFileFunction(
...
+            launch {
+                response = client.get("${System.getenv("ETOR_TI_baseurl")}/v1/etor/metadata/" + lookupId) {
+                    authPair.first?.forEach { entry ->
```
mkalish, line 255:
```suggestion
                    authPair.first.forEach { entry ->
```

mkalish (commented):
Nice work!  Really coming along here, summary of some thoughts I had
- I agree that we should figure out a more creative way to support this optional env var and not bake into the app since it strictly optional
- Not being able to find the root report for an id might make sense to log an error since we should investigate that
- An integration test that only mocks out the TI endpoint would be great to lock the behavior, but I think that can be a fast follow/happy to pair on that

prime-router/src/test/kotlin/history/azure/DeliveryFunctionTests.kt
jcrichlake: Fixed

prime-router/src/main/kotlin/history/azure/SubmissionFunction.kt
arnejduranovic: Ah, makes sense. Good catch! Will make a note to create a new ticket.

prime-router/src/main/kotlin/history/azure/ReportFileFunction.kt
jcrichlake: Fixed

halprin: Fixed.

prime-router/src/main/kotlin/history/azure/ReportFileFunction.kt
halprin: Fixed.

prime-router/src/main/kotlin/history/azure/ReportFileFunction.kt
halprin: Fixed.

prime-router/src/main/kotlin/history/azure/DeliveryFunction.kt
halprin: Good point.  Fixed.

prime-router/src/main/kotlin/history/azure/DeliveryFunction.kt
halprin: Fixed.

Issue #13507, opened 2024-02-27
Create Integration Tests for FHIRConverter


arnejduranovic:
### User Story

As a developer of the Universal Pipeline,
I want to have integration tests that ensure `FHIRConverter` updates the system state correctly and produces correct outputs,
so that I can confirm the pipeline step is working properly.

### Description/Use Case

The integration tests should use both the database and azure blob test containers and verify both are updated correctly given certain inputs to the function.

### Acceptance Criteria

- [ ] Created one or more integration tests for the pipeline step that gives confidence the step is working as expected

Andrey-Glazkv:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/13507?planning-poker) with Zenhub @adegolier @arnejduranovic @brick-green @jack-h-wang @jalbinson @JFisk42 @mkalish @thetaurean

PR #14183, opened 2024-04-23
13507: FHIR Converter integration tests

mkalish:
This PR applies a fix for which message model is chosen when parsing an HL7 message by deferring to the topic as well as adding comprehensive integration tests for the FHIR converter

Test Steps:
1. Send one of the elims messages through the pipeline and verify that it gets delivered
2. Send one of the MarsOtc messages use to test validation through the pipeline and verify it gets delivered

Verify the coverage (generated from only running the new tests)
Screenshot 2024-04-26 at 2 49 48 PM

## Changes
- Adds the ability to use an HL7MessageParseConfiguration set on the topic (required to support the RadXMars spec), otherwise uses the current fallback behavior of 
  - Any ORU_R01 message is parsed as 27 and otherwise the default HAPI behavior is use
- Adds a comprehensive set of integration tests for the FHIR Converter

### Testing
- [x] Tested locally?
- [x] Added tests?

### Process

## Linked Issues
- #13507


david-navapbc:
in addition to the integration test failures - there is an additional test failure that only presents when `gradlew package` or `gradlew build` is invoked. When `gradlew test` is invoked the test does not fail. 

```gov.cdc.prime.router.fhirengine.translation.SyncSchemaE2ETests test end-to-end sync schemas workflow() FAILED (10s)

  gov.cdc.prime.router.fhirengine.translation.TranslationSchemaManager$Companion$TranslationSyncException: Validation state was invalid, the valid blob is misconfigured. 
  It is either duplicated or not present when the state is not being validated
      at app//gov.cdc.prime.router.fhirengine.translation.SyncSchemaE2ETests.test end-to-end sync schemas workflow(SyncSchemaE2ETests.kt:87)```

File: prime-router/settings/organizations.yml [hunk 1]
```diff
@@ -106,6 +106,11 @@
       customerStatus: active
       schemaName: classpath:/metadata/fhir_transforms/senders/original-pipeline-transforms.yml
       format: HL7
+    - name: ignore-mars-otc
```
mkalish, line 109: A little unrelated, but I realized we should have a mars otc sender for test purposes

prime-router/src/main/kotlin/Sender.kt
```diff
@@ -43,6 +43,7 @@ import java.time.OffsetDateTime
     JsonSubTypes.Type(value = UniversalPipelineSender::class, name = "full-elr"),
     JsonSubTypes.Type(value = UniversalPipelineSender::class, name = "etor-ti"),
     JsonSubTypes.Type(value = UniversalPipelineSender::class, name = "elr-elims"),
+    JsonSubTypes.Type(value = UniversalPipelineSender::class, name = "mars-otc-elr"),
```
mkalish, line 46: I had initially missed that this needs to get added in addition to topic

prime-router/src/main/resources/metadata/fhir_transforms/senders/SimpleReport/simple-report-sender-transform.yml
```diff
@@ -588,6 +588,7 @@ elements:
     bundleProperty: '%resource.extension("https://reportstream.cdc.gov/fhir/StructureDefinition/hl7v2-date-time").value[x]'
     value: [ '%diagnosticReport.issued' ]
 
+  # TODO: ticket.  This transform causes a circular JSON structure when applied to a FHIR bundle generated from converting HL7
```
mkalish, line 591: I originally encountered this when trying to write tests for an HL7 message with a sender transform, but I worked around it by using the original-pipeline-transform (there were othe rreasons that it made impossible to use), but I didn't want to lose track of this and will open a bug before merging.

prime-router/src/test/kotlin/fhirengine/azure/FHIRConverterIntegrationTests.kt
```diff
@@ -0,0 +1,456 @@
...
+                    10001
+                )
+            }/devstoreaccount1;"""
+        return BlobAccess.BlobContainerMetadata(
+            "container1",
+            blobConnectionString
+        )
+    }
+
+    // TODO convert between sender format and report form
```
mkalish, line 155: I'll add a ticket for this as well, so both formats don't need to be passed

adegolier:
Haven't been able to do the test steps yet because I'm having problems building/running the tests locally.

prime-router/src/test/kotlin/common/TestcontainersUtils.kt
```diff
@@ -15,12 +15,6 @@ object TestcontainersUtils {
         customImageName: String,
         customEnv: Map<String, String> = emptyMap(),
     ): GenericContainer<*> {
-        val containerKey = Pair(customImageName, customEnv)
```
mkalish, line 18: The caching here was not safe across multiple gradle test runners and can result in random tests beginning to fail even if they were previously passing.

prime-router/src/main/kotlin/SettingsProvider.kt
```diff
@@ -53,6 +54,7 @@ enum class Topic(
     val isUniversalPipeline: Boolean = true,
     val isSendOriginal: Boolean = false,
     val validator: IItemValidator = NoopItemValidator(),
+    val hl7ParseConfiguration: HL7Reader.Companion.HL7MessageParseAndConvertConfiguration? = null,
```
arnejduranovic, line 57: Why wasn't MARS_OTC_ELR updated to configure `HL7MessageParseAndConvertConfiguration`?

prime-router/src/main/kotlin/fhirengine/utils/HL7Reader.kt
```diff
@@ -270,11 +270,12 @@ class HL7Reader(private val actionLogger: ActionLogger) : Logging {
...
-        fun parseHL7Message(rawHL7: String): Pair<Message, HL7MessageParseAndConvertConfiguration?> {
+        fun parseHL7Message(
+            rawHL7: String,
+            parseConfiguration: HL7MessageParseAndConvertConfiguration?,
+        ): Message {
             val hl7MessageType = getMessageType(rawHL7)
-            val parseConfiguration = messageToConfigMap[hl7MessageType]
-            val message = getHL7ParsingContext(hl7MessageType, parseConfiguration).pipeParser.parse(rawHL7)
-            return Pair(message, parseConfiguration)
+            return getHL7ParsingContext(hl7MessageType, parseConfiguration).pipeParser.parse(rawHL7)
```
arnejduranovic, line 278: I think we need to look into `getHL7ParsingContext` and evaluate the business logic there and align with product. I'll think on it and maybe create a ticket, but I think this is fine for now if its well tested.

prime-router/src/test/kotlin/fhirengine/azure/FHIRConverterIntegrationTests.kt
arnejduranovic: reminder to create ticket and then update this comment with the number

prime-router/src/main/kotlin/SettingsProvider.kt
mkalish: I was holding off on adding that until the mappings and java file is done.

prime-router/src/main/kotlin/fhirengine/utils/HL7Reader.kt
mkalish: Yep... I think we have it isolated well now so that we make tactical changes here

arnejduranovic (approved):
Changes look good and tests pass locally

prime-router/src/main/kotlin/SettingsProvider.kt
mkalish: FYI, my thought is that the fast follow is done in [this](https://github.com/CDCgov/prime-reportstream/pull/14255) PR, let me know if you have a different preference

Issue #13794, opened 2024-03-22
Create the RadxMars 2.5.1 ORU-R01 message and data types


arnejduranovic:
### User Story
As a report stream sender,
I would like to be able to have an exact mapping for a HL7 message that matches the RadxMars OTC 2.5.1 conformance profile
so that ReportStream does not drop any data when serializing my messages

### Description/Use Case
This is covered more in depth in the design spec, but in order to guarantee being able to losslessly and deterministically map from HL7->FHIR->HL7 for a give conformance profile, there needs to exist java classes that match that profile.  The work here is to create all the java code needed for RadxMars OTC 2.5.1

### Dev Notes
- See how the PR for #13021 accomplished this for NIST 2.5.1
- See output of #13024 for details on how to do this work

### Acceptance Criteria
- [ ] All the custom datatypes used in the the RadxMars OTC 2.5.1 are created
- [ ] A class that extends `AbstractMessage` exactly matches the RadxMars 2.5.1 conformance profile
- [ ] The class can be used to parse a valid RadxMars OTC 2.5.1 message

Andrey-Glazkv:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/13794?planning-poker) with Zenhub @adegolier @arnejduranovic @brick-green @jack-h-wang @jalbinson @JFisk42 @mkalish @thetaurean

PR #14185, opened 2024-04-24
13794 add radxmars hl7 java class

jack-h-wang:
This PR adds a Java class to support RADx MARS HL7.

Ideally, this class should also support all other ORU-R01 HL7 when catchall mappings are updated.

Test Steps:
1. Verify the Java class is compatible with the RADx MARS spec. Deviations from the spec in terms of allowing deprecated fields or the use of forwards compatible types are expected.

## Changes
- Add a Java class based on HL7 v2.7 that makes changes needed to support RADx MARS HL7.

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- #13794

## To be done
- #14249


prime-router/src/main/java/fhirengine/translation/hl7/structures/radxmars251/segment/OBX.java
```diff
@@ -0,0 +1,1464 @@
...
+      this.add(CWE.class, false, 0, 0, new Object[]{getMessage()}, "Observation Method");
+      this.add(EI.class, false, 0, 0, new Object[]{getMessage()}, "Equipment Instance Identifier");
+      this.add(DTM.class, false, 1, 0, new Object[]{getMessage()}, "Date/Time of the Analysis");
+      this.add(CWE.class, false, 0, 0, new Object[]{getMessage()}, "Observation Site");
+      this.add(EI.class, false, 1, 0, new Object[]{getMessage()}, "Observation Instance Identifier");
+      this.add(CNE.class, false, 1, 0, new Object[]{getMessage()}, "Mood Code");
+      this.add(XON.class, false, 1, 0, new Object[]{getMessage()}, "Performing Organization Name");
+      this.add(XAD.class, false, 1, 0, new Object[]{getMessage()}, "Performing Organization Address");
+      this.add(XCN.class, false, 1, 0, new Object[]{getMessage()}, "Performing Organization Medical Director");
+      this.add(ID.class, false, 1, 10, new Object[]{getMessage(), Integer.valueOf(909)}, "Patient Results Release Category");
```
brick-green, line 120: Are there supposed to be two `Patient Results Release Category` types?

jack-h-wang: Good catch, that's a mistake

prime-router/src/main/java/fhirengine/translation/hl7/structures/radxmars251/segment/OBX.java
```diff
@@ -0,0 +1,1463 @@
...
+      this.add(CWE.class, false, 1, 0, new Object[]{getMessage()}, "Root Cause");
+      this.add(CWE.class, false, 0, 5, new Object[]{getMessage()}, "Local Process Control");
```
jack-h-wang, line 122: Choosing to defer to the HL7 v2.8 spec in classifying OBX-28 as CWE. RADxMARS classes this as ID, but as an unused optional field and it can be processed as CWE with no issue.

brick-green (approved):
looking good

prime-router/src/test/kotlin/fhirengine/translation/RadxMarsTests.kt
```diff
@@ -0,0 +1,43 @@
...
+import ca.uhn.hl7v2.validation.impl.ValidationContextFactory
+import gov.cdc.prime.router.fhirengine.engine.encodePreserveEncodingChars
+import kotlin.test.Test
+import fhirengine.translation.hl7.structures.radxmars251.message.ORU_R01 as RADXMARS_ORU_R01
+
+class RadxMarsTests {
+    @Test
+    fun `test radxmars java class`() {
+        val rawMessage = """
+MSH|^~\&|MMTC.PROD^2.16.840.1.113883.3.8589.4.2.106.1^ISO|CAREEVOLUTION^00Z0000024^CLIA|AIMS.INTEGRATION.PRD^2.16.840.1.114222.4.3.15.1^ISO|AIMS.PLATFORM^2.16.840.1.114222.4.1.217446^ISO|20240404110305+0000||ORU^R01^ORU_R01|20240404110305_433848dc33ca4ce89ff4773fed48e8f6|P|2.5.1|||NE|NE|||||PHLabReport-NoAck^ELR251R1_Rcvr_Prof^2.16.840.1.113883.9.11^ISO
```
mkalish, line 19: Can we update one of the OBX segments to include the fields that we had to add?  Noticed this will setting up some testing

prime-router/src/main/java/fhirengine/translation/hl7/structures/radxmars251/group/ORU_R01_VISIT.java
mkalish, line 1: Why do we need this class?  `PV1` and `PV2` are the same right?

mkalish (changes requested):
Ok... after playing around a bit I think some things need to get updated

- Our new java classes should all return version `2.7` currently the message is now returning a mix of `2.5.1` and `2.7`
- I believe the parsing issue I'm hitting is because of some differences between what you have required and what is required in 2.7.  I think you need to go back and have it match with the 2.7 copy
- Do we need all the new groups?  I think some of them would be the same as in 2.7

prime-router/src/main/java/fhirengine/translation/hl7/structures/radxmars251/group/ORU_R01_VISIT.java
jack-h-wang: I rebased the message and groups from the 2.7 class and those which do not bring in OBX are no longer included.

mkalish (approved):
Updates worked for me!

Issue #14114, opened 2024-04-18
Update the condition mapping code to indicate the correct index in the original report for an item with an unmappable code


mkalish:
### User Story
As a report stream sender, I would like to know which item in my report contained an unmappable condition

### Description/Use Case
Currently, the index set in the action log detail is the index out of the produced FHIR bundles which does not map to the actual index in the report.  The mapping logic needs to be shifted so that a sender can correctly identify the item in the report they sent that had an unmappable condition.

### Dev Notes
- This will likely require adjusting where this code runs in the conversion step

### Acceptance Criteria
- [ ] The index reported in the action log detail for an unmappable condition always reflects the index in the original report

Andrey-Glazkv:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/14114?planning-poker) with Zenhub @adegolier @arnejduranovic @brick-green @jack-h-wang @jalbinson @JFisk42 @mkalish @thetaurean

Andrey-Glazkv:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/14114?planning-poker) with Zenhub @david-navapbc

PR #14284, opened 2024-05-02
Update condition mapping to indicate correct item index in log

brick-green:
This PR moves the condition mapping code to the process() function in FHIRConverter. This was needed as the process() function can remove items in a report when an item is not parsable. So moving the condition mapping functionality into process() ensures the original index would still be accessible.

Test Steps:
1. run ./prime test
2. review `test fully unmapped condition code stamping logs errors` in FHIRConverterTests()

## Changes
- move condition mapping functionality into the process() function
- remove some unnecessary asserts from test
- add some asserts to test index logging

## Linked Issues
- #14114


thetaurean (approved):
Looks good! Thanks for cleaning up those assertions 🙂👌

prime-router/src/test/kotlin/fhirengine/engine/FhirConverterTests.kt
```diff
@@ -466,7 +467,7 @@ class FhirConverterTests {
 
     @Test
     fun `test fully unmapped condition code stamping logs errors`() {
-        val fhirData = File(VALID_DATA_URL).readText()
+        val fhirData = File(BATCH_VALID_DATA_URL).readText()
```
mkalish, line 470: Can we also add this in to the new integration tests?

PR #14404, opened 2024-05-15
add duration logging for process function and condition mapping

brick-green:
This PR adds duration logging to the convert function with the intent to measure the process function and within that the condition mapping. In my local testing the `process()` function accounted for 90%+ of the fhir convert function. Within the `process()` function the condition mapping code accounted for 80%+. 

## Changes
- Add duration logging to FHIR Converter process()
- Add duration logging to lookupConditionCodings()
- Add duration logging to Observation.addMappedConditions()


snesm (approved):

jeremy-page (approved):

prime-router/src/main/kotlin/fhirengine/engine/FHIRConverter.kt
```diff
@@ -125,7 +125,11 @@ class FHIRConverter(
             // TODO: https://github.com/CDCgov/prime-reportstream/issues/14287
             FhirPathUtils
 
-            val fhirBundles = process(format, queueMessage, actionLogger)
+            val fhirBundles =
+                LogMeasuredTime.measureAndLogDurationWithReturnedValue(
+                    "Process queue message",
+                    mapOf("reportId" to queueMessage.reportId.toString(), "topic" to queueMessage.topic.name)
```
mkalish, line 131: nit: Both of these will get automatically added by the MDC utils call above

mkalish (commented):
Changes look good, but maybe can't hurt to also throw some logging around the other pieces of in the convert step just so we have a complete picture?

PR #14498, opened 2024-05-21
Upgrade Frontend for Okta OIE

### LINKED ISSUE #14497: Update Frontend Okta Widget for OIE
With both okta environments migrated to the Identity engine, we can update our widget settings to take full advantage of it (disabling the "classic" engine mode it is currently running under). We also need to update the frontend to support receiving self-service parameters, which will allow our widget to perform the self-service steps instead of the okta-hosted page (having our emails reference the website to do this will happen later).

# Acceptance Criteria
- [ ] Frontend okta widget's Classic engine mode disabled
- [ ] Login page supports `otp` and `token` url parameters and passes on to widget
- [ ] Check all existing user flows to ensure working properly with new changes:
  - [ ]  Login
  - [ ]  Lock-out
  - [ ]  Reset Password
  - [ ] MFA Factors
  - [ ] New account first login setup

jpandersen87:
Fixes #14497

This PR updates the login page to pass special URL parameters to the Okta widget and to use the default OIE mode, which will allow it to handle self-service workflows.


frontend-react/src/pages/Login.tsx
```diff
@@ -12,6 +12,15 @@ export function Login() {
     const { oktaAuth, authState, config } = useSessionContext();
     const location: Location<{ originalUrl?: string } | undefined> =
         useLocation();
+    const [searchParams] = useSearchParams();
+    const finalConfig = useMemo(
```
penny-lischer, line 16: Will these values get cleared on logout and on session expiration?

jpandersen87: They are only URL params collected on the login page. There is no persistence and natural login process will redirect the user after successful auth which won't have any params.

etanb (approved):

penny-lischer (approved):
lgtm

PR #14559, opened 2024-05-29
14168 FHIR-to-FHIR transform validation

### LINKED ISSUE #14168: Create Validation pipeline for FHIR->FHIR transforms
### User Story
We want to check FHIR transforms for potential errors during development

Only handle FHIR->FHIR transforms

### Dev Notes
- Follow pattern set up for organizations
- Add json schema for FHIR transforms
  - Include references to schema at the top of all fhir transform files
- Add Konform validations for FHIR transforms

### Acceptance Criteria
- FHIR transforms now have Intellij autocomplete
- FHIR transforms will now be checked to have valid structure and values

jalbinson:
This PR adds validation for FHIR-to-FHIR transforms.

Test Steps:
1. `./gradlew primeCLI --args='validate-yaml --type FhirToFhirTransform --dir src/main/resources/metadata/fhir_transforms'`

## Changes
- Added JSON schema for FHIR-to-FHIR transforms
- Added FHIR path validation to specific fields within FHIR-to-FHIR transforms
- Added a step to CI to ensure all newly added transforms are valid before merging  

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- Fixes #14168


prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirPathUtils.kt
```diff
@@ -60,6 +60,16 @@ object FhirPathUtils : Logging {
         }
     }
 
+    /**
+     * Is the provided path a valid FHIR path given the evaluation context?
+     */
+    fun validatePath(path: String, evaluationContext: FHIRPathEngine.IEvaluationContext): Boolean {
+        return runCatching {
+            pathEngine.hostServices = evaluationContext
```
jalbinson, line 68: I'm not a huge fan of mutating `pathEngine.hostServices` in the `FhirPathUtils` object before every call. It could leave it in a bad state. It also looks like its used outside the object in `ProcessFhirCommands` which could cause some undefined behavior.

prime-router/src/main/kotlin/fhirengine/engine/LookupTableValueSet.kt
```diff
@@ -27,8 +27,8 @@ data class LookupTableValueSetConfig(
  */
 class LookupTableValueSet
     (@JsonProperty("lookupTable") private val configData: LookupTableValueSetConfig) : ValueSetCollection {
-    private val metadata = Metadata.getInstance()
     private val mapVal: SortedMap<String, String> by lazy {
+        val metadata = Metadata.getInstance()
```
jalbinson, line 31: I dropped this into the `lazy` block to avoid starting up the whole database connection pool during the validation step where all I care about is the parsed values from YAML.

snesm (approved):

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirPathUtils.kt
thetaurean: What about keeping the existing hostServices in a temp variable then resetting it afterwards?

Though this would only happen when running a `validate-yaml` command from the cli right?

Depending on above, maybe this warrants adding a TODO or a tech-debt ticket for a temporary context change.

thetaurean (approved):
👍  Looks good to me. I left a reply on your `pathEngine.hostServices` concern which might prompt a change.

I was able to test out the validate command locally.

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirPathUtils.kt
jalbinson: It's safe currently because the calls to mutate it happen during cli commands which runs a fresh version of the app every time.

The code called during the pipeline always resets the state to what it needs through appContext. It's not a danger yet but it is definitely a code-smell that could lead to a weird bug in the future. I'll leave a comment to think about how we could fix it.

arnejduranovic: Reminder to leave comment

arnejduranovic (changes requested):

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirPathUtils.kt
thetaurean: Yeah I figured this wouldn't persist. I still think it would be a good idea to temporarily `hostServices` and then set it back after the function call.

arnejduranovic (approved):

prime-router/src/main/kotlin/fhirengine/translation/hl7/utils/FhirPathUtils.kt
jalbinson: Done.

Issue #13598, opened 2024-03-06
Update the code so that long schema names do not cause errors when saving a report_file


mkalish:
### User Story
As a developer, I would like the report_file.schema_name to support the length of schemas that we have

### Description/Use Case
This column was originally setup to store the schema names from the covid pipeline which had a different convention that was much shorter (i.e. `waters/waters-covid-19`).  The schema names in the universal pipeline are now paths to schemas which can be substantially longer, but the column is only configured to allow for 63 characters.  In some places we are over the limit already `metadata/fhir_transforms/senders/SimpleReport/simple-report-sender-transform`, (through a current bug/quirk this is not recorded in the db) and in other places we are extremely close `metadata/hl7_mapping/receivers/STLTs/VI/VI-receiver-transform`.  Adding in `classpath:/` and `.yml` tipped it so that the schemas we do record are greater than 63

### Dev Notes
- Altering the length of a varchar **should** not require a lock/rewrite on the table so this should be safe.  The migration would be something like `ALTER TABLE report_file ALTER COLUMN schema_name TYPE VARCHAR(5000);`
  - We should make sure to ask devops to check this beforehand since a table lock would require downtime

### Acceptance Criteria
- [ ] The temporary code is removed
- [ ] Longer schema names do not cause an issue when saving a report_file

arnejduranovic:
PR https://github.com/CDCgov/prime-reportstream/pull/14624 exists to implement this ticket but has received pushback from DevOps. TBD how to proceed.

PR #14624, opened 2024-06-04
13598 address database varchar size limits

jack-h-wang:
This PR extends the length of the `report_file`.`schema_name` column so longer strings can be stored and removes temporary code that truncates the data to be stored in this column. It additionally extends the length of many other varchar columns throughout the database.

Test Steps:
1. Run `clean` followed by `package` to apply db migrations and run tests.
2. Confirm the database has been updated by running the following query and verifying the value is null:
```
SELECT character_maximum_length 
FROM information_schema.columns 
WHERE table_schema = 'public' 
AND table_name  = 'report_file' 
AND column_name  = 'schema_name';
```

## Changes
- Add migration to alter size of `report_file`.`schema_name` and many other varchar database columns
- Remove temporary code to truncate schema name prior to inserting to database
- Fix actionHistory to correctly handle varchar columns without a size limit
- Add and update tests

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- #13598

## To be done
- #14721


prime-router/src/main/resources/db/migration/V71__extend_schema_name.sql
```diff
@@ -0,0 +1,15 @@
...
+ *
+ * Copy a version of this comment into the next migration
+ *
+ */
+
+/*
+ * Extend the length of report_file.schema_name to accommodate expected data length.
+ */
+
+ALTER TABLE report_file ALTER COLUMN schema_name TYPE VARCHAR(1000);
```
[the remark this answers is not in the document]
mkalish: Might be easiest to just not have a limit at all so we don't have to up the value again (though 1000 is probably long enough).  Postgres recommends not to set a length limit unless there is an application reason to enforce a length.

https://wiki.postgresql.org/wiki/Don%27t_Do_This#Don.27t_use_varchar.28n.29_by_default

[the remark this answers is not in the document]
arnejduranovic: As part of this ticket, @jack-h-wang can you please go through the DB and apply Michael's proposed change to all varchar fields and we can include this link in the PR as justification? I'm also open to another ticket, but make sure to include all changes in a single PR so we have a single migration.

prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -188,9 +185,18 @@ class ActionHistory(
             }
         }
         // capture the azure client IP but override with the first forwarded for if present
-        action.senderIp = request.headers["x-azure-clientip"]?.take(ACTION.SENDER_IP.dataType.length())
+        action.senderIp = if (ACTION.SENDER_IP.dataType.length() > 0) {
```
jack-h-wang, line 188: This and the handling of several other strings needed to change in order to not break when the database column has no length limit. `.length()` returns 0 in this case, which caused the string to be truncated to 0 characters.

prime-router/src/main/resources/db/migration/V71__extend_character_limits.sql
```diff
@@ -0,0 +1,178 @@
...
+DROP INDEX IF EXISTS idx_vw_livid_test_performed_loinc_code;
+DROP INDEX IF EXISTS idx_vw_livid_test_orderd_loinc_code;
+DROP INDEX IF EXISTS idx_vw_livid_otc;
+
+/* Remove the view */
+DROP MATERIALIZED VIEW IF EXISTS vw_livd_table;
+
+
+ALTER TABLE action
+ALTER COLUMN action_params TYPE VARCHAR,
```
snesm, line 28: Please provide the original COLUMN types for each of these with reasoning for using an effectively unlimited field.

snesm (changes requested):

prime-router/src/main/resources/db/migration/V71__extend_character_limits.sql
arnejduranovic: I think adding original column types is a good idea for documentation purposes. @jack-h-wang the reasoning that applies to most of these is "Previous limit was arbitrary and there is not a solid application requirement for limiting this field to a certain length". 

The only one I think that does not have non-arbitrary limits they cannot exceed lest they be invalid and could have a length constraint is `sender_ip` (39 characters to support IPv6), but I think we can just leave is without the limit like you have here.

arnejduranovic (approved):

prime-router/src/main/resources/db/migration/V71__extend_character_limits.sql
jack-h-wang: The migration now includes the original types for all fields that are changed. Justification has been added to the beginning comments of the migration.

prime-router/src/test/kotlin/azure/ActionHistoryTests.kt
```diff
@@ -562,11 +582,4 @@ class ActionHistoryTests {
         assertContains(blobUrls[0], org.receivers[0].fullName)
         assertContains(blobUrls[1], org.receivers[1].fullName)
     }
-
-    @Test
-    fun `test trimSchemaNameToMaxLength malformed URI`() {
-        val longMalformedURI = ":very_very:_long_name//with a badly formed URI that causes a parse exception"
```
adegolier, line 568: Do we still have/need handling of malformed URIs? If so, we should probably be keeping that part of the test at least.

adegolier:
Would definitely like to see this branch updated with the latest from master, just to make sure everything is okay after the changes Kalish made with the project structure.

prime-router/src/main/resources/db/migration/V71__extend_character_limits.sql
```diff
@@ -0,0 +1,183 @@
...
+DROP MATERIALIZED VIEW IF EXISTS vw_livd_table;
+
+
+ALTER TABLE action
+ALTER COLUMN action_params TYPE VARCHAR, --previously VARCHAR(2048)
+ALTER COLUMN sender_ip TYPE VARCHAR, -- previously VARCHAR(39)
+ALTER COLUMN sending_org TYPE VARCHAR, -- previously VARCHAR(63)
+ALTER COLUMN sending_org_client TYPE VARCHAR, -- previously VARCHAR(63)
+ALTER COLUMN receiving_org TYPE VARCHAR, -- previously VARCHAR(63)
+ALTER COLUMN receiving_org_svc TYPE VARCHAR; -- previously VARCHAR(63)
```
arnejduranovic, line 37: Matt: Consider where clause performance for some of these

jack-h-wang:
Superceded by #18116.

PR #14645, opened 2024-06-06
Filenames should prefix with none when translation schema is not present

### LINKED ISSUE #14455: Delivered filename for a FHIR receiver is wrong
### User Story
As a FHIR receiver, I would like the filename to look like `{reportId}-{datetime}.fhir`.  

### Description/Use Case
Currently, the format is `-{reportId}-{datetime}.fhir`

### Dev Notes

To recreate:
1. Send a report that will go to a FHIR SFTP receiver
2. Look at file in build/sftp

Expect the filename to be: `{reportId}-{datetime}.fhir`
In reality: `-{reportId}-{datetime}.fhir`

### Acceptance Criteria
- [ ] Filename for a delivered FHIR report is updated

JFisk42:
This PR fixes the issue where FHIR filenames were being formed incorrectly. The standard filename format is `{translationSchema}-{reportId}-{dateTime}.{fileType}`. Because most (all?) FHIR receivers do not utilize a translation schema, this was being left as a blank string and formatting the filename as `-{reportId}-{dateTime}.{fileType}`. 

Regression tests/other scenarios that were considered:
- Topics with `isSendOriginal` set to true. 
  - These end up with a filename format of `None-{reportId}-{dateTime}.{fileType}`. This is due to where the SchemaName is set in Report.kt and a comment was added to highlight where this is happening.
  - Alternatively, reports with this topic setting can provide a `payloadname` when being submitted. This will result in an external name added to the file and this scenario **does not** use the logic in FileNameTemplate.kt. This scenario is [tested elsewhere](https://github.com/CDCgov/prime-reportstream/blob/2569a589d330c300477bbbff96694335dae2e84d/prime-router/src/test/kotlin/ReportTests.kt#L1071).
- Alternative filename template formats.
  - We have a [yml for defining alternative filename formats](https://github.com/CDCgov/prime-reportstream/blob/2569a589d330c300477bbbff96694335dae2e84d/prime-router/metadata/file_name_templates/file-name-templates.yml).
  - This fix only touches filename formats which utilize the `SchemaBaseName()` element. Filename templates with this element are `standard` and `standard_secondary`.
  - **Test have been added to cover these scenarios**.

Test Steps:
1. (_optional_) To reproduce the issue 
   1. Checkout master and run `./gradlew testEnd2EndUP`
   2. Look over the history endpoint results for the tests. You will notice a few things:
      1. Reports routed to receivers with no translation schema (can be verified by checking organizations.yml) will have the incorrect formatting of `-{reportId}-{dateTime}.{fileType}`. This will be the case for all receivers setup to receiver FHIR for this test.
      2. Reports for ELIMS (the isSendOriginal set to true scenario) will be formatted as `None-{reportId}-{dateTime}.{fileType}`.
      3. Reports routed to receivers with translation schemas listed will be properly formatted as `{translationSchema}-{reportId}-{dateTime}.{fileType}`. This will be the case for all receivers that receive HL7 in this test.
2. To verify the change
   1. Checkout the branch and run `./gradlew testEnd2EndUP`
   2. Look over the history endpoint results for the tests.
      1. Reports routed to FHIR receivers should now be prefixed with `none` and formatted as `none-{reportId}-{dateTime}.fhir`.
         1. The reason for lowercase is that the standard filename template is set to lowercase.
      2. Reports for ELIMS should be unchanged and will be formatted as `None-{reportId}-{dateTime}.hl7`.
      3. Reports routed to HL7 receivers should be unchanged and will be formatted as `{translationSchema}-{reportId}-{dateTime}.hl7`.

## Changes
- When a filename is generated for a receiver with no SchemaBaseName, this will be populated as `None` instead of an empty string.

### Testing
- [x] Tested locally?
- [x] Added tests?

### Process

## Linked Issues
- Fixes #14455


prime-router/src/testIntegration/kotlin/FileNameTemplateIntegrationTests.kt
```diff
@@ -349,6 +359,80 @@ class FileNameTemplateIntegrationTests {
...
+        }
+        assertThat(metadata.fileNameTemplates).containsKey(key)
+        val standardNameFormat = metadata.fileNameTemplates[key]
+
+        // act
+        val fileName = standardNameFormat?.getFileName(config, reportId = reportId)
+            ?: assertk.fail("error getting standard file name")
+
+        // assert
+        assertThat(fileName).startsWith("oru_r01-base.yml-$reportId-$formattedDate")
```
brick-green, line 377: realize this is out of scope for the ticket. But it seems odd to include the ext for the schema file in the filename. Maybe we need a ticket to investigate that?

brick-green (commented):
Do you also need to check for a blank `schemaName` in `Report.formExternalFilename()` in this block 

```
if (translationConfig == null) {
                val formatter = DateTimeFormatter.ofPattern("yyyyMMddHHmmss")
                "${Schema.formBaseName(schemaName)}-$reportId-${formatter.format(createdAt)}.$nameSuffix"
            }
```

JFisk42:
> Do you also need to check for a blank `schemaName` in `Report.formExternalFilename()` in this block
> 
> ```
> if (translationConfig == null) {
>                 val formatter = DateTimeFormatter.ofPattern("yyyyMMddHHmmss")
>                 "${Schema.formBaseName(schemaName)}-$reportId-${formatter.format(createdAt)}.$nameSuffix"
>             }
> ```

I looked over it again and this logic branch gets used in a few different instances:
- When reports are first received
- Just after the conversion to FHIR happens
- When routing ELIMS reports (sendOriginal = true)

It looks like in all of these cases that the schema name is getting pulled in as "None" set in one of two places in Report.kt ([example](https://github.com/CDCgov/prime-reportstream/blob/master/prime-router/src/main/kotlin/Report.kt#L408), I know this is the line where the schema name prefix is set for ELIMS). So I think we're good there.

prime-router/src/testIntegration/kotlin/FileNameTemplateIntegrationTests.kt
JFisk42: Agreed. I will log this tomorrow.

brick-green (approved):
lgtm

prime-router/src/testIntegration/kotlin/FileNameTemplateIntegrationTests.kt
JFisk42: Logged: https://app.zenhub.com/workspaces/platform-6182b02547c1130010f459db/issues/gh/cdcgov/prime-reportstream/14703

arnejduranovic (approved):

prime-router/src/main/kotlin/FileNameTemplate.kt
```diff
@@ -76,7 +76,14 @@ class SchemaBaseName : FileNameElement {
     override val name = "schemaBaseName"
 
     override fun getElementValue(args: List<String>, translatorConfig: TranslatorConfiguration?): String {
-        return Schema.formBaseName(translatorConfig?.schemaName ?: "")
+        val myString = if (translatorConfig?.schemaName.isNullOrEmpty()) {
```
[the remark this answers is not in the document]
JFisk42: Whoops, somehow turned a blind eye to my temporary variable name, fixed this.

prime-router/src/main/kotlin/Report.kt
```diff
@@ -405,6 +405,7 @@ class Report : Logging {
     ) {
         this.id = UUID.randomUUID()
         // UP submissions do not need a schema, but it is required by the database to maintain legacy functionality
+        // topics with isSendOriginal set to true will use this name when generating the filename
```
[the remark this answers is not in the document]
JFisk42: Yeah when you put it like that I agree it doesn't really have a place here. It's also easy enough to find if you don't know where to look. Removed.

Issue #14330, opened 2024-05-07
Update Internal Blobstore Naming Strategy


arnejduranovic:
### User Story

As a blob in ReportStream's Storage Container,
I want to ensure my name does not collide with any other blob name,
so that the pipeline step that generates me does not fail and throw and exception.

### Description/Use Case

Each step in the universal pipeline generates one or more reports. These reports are then stored in the proper folder in the Azure Storage Container. The blobs are named with complex naming logic that may cause collisions and lead to the upload to fail and the pipeline step to then fail as well. 

There are two important features in ReportStream concerning filenames:

1. ReportStream allows receivers to specify the format of their filename and what info their filename contains (see `prime-router/metadata/file_name_templates/file-name-templates.yml`)
2. ReportStream allows senders to specify the name of the submitted report in the header of their submission request (`payloadName`). If `payloadName` is specified AND the topic has isSendOriginal enabled, then the receiver's filename convention should be ignored and the name of the blob should equal: `reportID + payloadName`.

The above two use cases shall still be supported, but whatever the filename a receiver will receive, it shall not be the name of the blob in ReportStream's internal blob store. ReportStream's internal blob store shall only include the report's UUID in the name and nothing else. This is to avoid potential naming collisions.

### Risks/Impacts/Considerations

- Consider how we link sent filenames to the internal blob name - verify/confirm

### Acceptance Criteria

- [ ] Filenaming logic updated to reflect requirements above

Andrey-Glazkv:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/14330?planning-poker) with Zenhub @adegolier @arnejduranovic @brick-green @jack-h-wang @jalbinson @mkalish @thetaurean @david-navapbc @JFisk42

PR #14649, opened 2024-06-07
Internal blobstore names to reportid.ext and report_file.external to sent or received filename

brick-green:
This PR sets the uploaded blob filename to reportID.ext for all steps while saving the submission payload name and sent filenames using the report file.externalName field. This also adds two azure events: ReportReceivedEvent and ReportSentEvent

### For filename changes
1. Run all tests
2. review tests that I changed to ensure the behavior is captured. This includes ensuring that the submission payload name is used as the sent filename when the topic has send original.

### For azure events
1. configure system locally according to [local-dev-setup.md](https://github.com/CDCgov/prime-reportstream/blob/master/prime-router/docs/observability/local-dev-setup.md)
2. get your SU password
3. log in to azure and open pdhtest-appinsights
4. Open Logs --> customEvents
5. Check that ReportReceivedEvent and ReportSentEvent are present 
    - **Note:** Sometimes this takes a few minutes to show up after the report has been processed
6. Review properties in the two new events ensuring they are correct and identifying if any new properties would be useful

## Changes
- Added ReportReceivedEvent and ReportSentEvent azure events
- Changed uploadBody() to only use reportID.ext as the filename
- Changed formExternalFilename() to use submission payload name when isSendOriginal is true

### Testing
- [X] Tested locally?

## Linked Issues
- Linked #14330


prime-router/src/main/kotlin/Report.kt
```diff
@@ -196,6 +197,17 @@ class Report : Logging {
                     else -> throw IllegalArgumentException("Unexpected extension $ext.")
                 }
             }
+
+            fun valueOfFromBodyFormat(bodyFormat: String): Format {
```
mkalish, line 201: Could this be replaced with `Format.valueOf()`?

prime-router/src/main/kotlin/Report.kt
```diff
@@ -1608,13 +1582,24 @@ class Report : Logging {
             header: WorkflowEngine.Header,
             metadata: Metadata? = null,
         ): String {
-            return if (header.reportFile.bodyUrl != null) {
-                BlobAccess.BlobInfo.getBlobFilename(header.reportFile.bodyUrl)
+            return if (header.receiver?.topic?.isSendOriginal == true) {
+//                the externalName of the root report should equal the submission payload name parameter
+                ReportService().getRootReport(header.reportFile.reportId).externalName ?: formExternalFilename(
```
mkalish, line 1587: Nit: maybe pass this in?  Seems a little weird to instantiate it here.

prime-router/src/main/kotlin/Report.kt
```diff
@@ -1608,13 +1582,24 @@ class Report : Logging {
             header: WorkflowEngine.Header,
             metadata: Metadata? = null,
         ): String {
-            return if (header.reportFile.bodyUrl != null) {
-                BlobAccess.BlobInfo.getBlobFilename(header.reportFile.bodyUrl)
+            return if (header.receiver?.topic?.isSendOriginal == true) {
```
mkalish, line 1585: Maybe as a follow up, but we might want to collapse/rename this with other `formExternalFilename` since they have different logic and potentially one makes a DB call

prime-router/src/main/kotlin/Report.kt
```diff
@@ -1631,18 +1616,23 @@ class Report : Logging {
...
+            metadata: Metadata? = Metadata.getInstance(),
+            nameFormat: String = "standard",
+            translationConfig: TranslatorConfiguration? = null,
         ): String {
+            val nameSuffix = format.ext
             return if (bodyUrl != null) {
                 BlobAccess.BlobInfo.getBlobFilename(bodyUrl)
+            } else if (translationConfig == null) {
+                val formatter = DateTimeFormatter.ofPattern("yyyyMMddHHmmss")
+                // This file-naming format is used for all INTERNAL files, and whenever there is no custom format.
```
mkalish, line 1628: Is this comment still accurate?

prime-router/src/main/kotlin/Report.kt
```diff
@@ -270,11 +282,7 @@ class Report : Logging {
     val name: String
```
mkalish, line 282: It looks like this actually might not be used anymore?  I only saw usages in tests.

prime-router/src/main/kotlin/azure/observability/event/AzureEventUtils.kt
```diff
@@ -34,4 +38,25 @@ object AzureEventUtils {
...
+    /**
+     * Returns a ReportSentEvent
+     */
+    fun createReportSentEvent(
+        receiver: Receiver,
+        reportFiles: List<ReportFile?>,
+        reportId: ReportId,
+        externalFilename: String,
+    ): ReportSentEvent {
+        val reportFile = if (reportFiles.size == 1) reportFiles.first() else null
```
mkalish, line 51: I'm not sure this right? For a batch of items wouldn't it be possible for there be more than one root report?

prime-router/src/main/kotlin/azure/observability/event/AzureEventUtils.kt
```diff
@@ -34,4 +38,25 @@ object AzureEventUtils {
...
+        return ReportSentEvent(
+            reportFile?.reportId,
+            reportId,
+            receiver.topic,
+            Sender.createFullName(reportFile?.sendingOrg, reportFile?.sendingOrgClient),
```
mkalish, line 56: Following on from above, I think this should be a list?

Also, I would be a fan of throwing an error (or at least logging one) if the reportFile is missing the sender info since that would mean the data is in an a broken state.

prime-router/src/main/kotlin/azure/observability/event/ReportSentEvent.kt
```diff
@@ -0,0 +1,18 @@
...
+
+import gov.cdc.prime.router.ReportId
+import gov.cdc.prime.router.Topic
+
+/**
+ * Event definition for when a report has been successfully sent to a receiver.
+ * [trasnportType] is nullable as some receivers will purposefully not have transports configured
+ */
+data class ReportSentEvent(
+    val sentReportId: ReportId?,
```
mkalish, line 11: When would this be nullable?

prime-router/src/main/kotlin/azure/observability/event/ReportReceivedEvent.kt
```diff
@@ -0,0 +1,12 @@
...
+
+import gov.cdc.prime.router.ReportId
+import gov.cdc.prime.router.Topic
+
+data class ReportReceivedEvent(
+    val reportId: ReportId,
+    val topic: Topic,
+    val format: String,
+    val senderName: String,
+    val payload: String?,
```
mkalish, line 11: Hmm, wondering if this might be crazy big and not good to put in the events?  I'm thinking like a 1000+ ndjson FHIR bundle

prime-router/src/main/kotlin/cli/ProcessDataCommands.kt
```diff
@@ -324,20 +323,9 @@ class ProcessData(
                 val outputFile = if (outputFileName != null) {
                     File(outputFileName!!)
                 } else {
-                    // generate a translation config if we don't have
-                    val translationConfig = report.destination?.translation ?: CustomConfiguration(
```
mkalish, line 328: Nice, so much simpler

prime-router/src/main/kotlin/fhirengine/engine/FHIRConverter.kt
```diff
@@ -171,7 +171,7 @@ class FHIRConverter(
                         val blobInfo = BlobAccess.uploadBody(
                             Report.Format.FHIR,
                             bodyBytes,
-                            report.name,
+                            report.id.toString(),
```
mkalish, line 174: I left a comment earlier that `report.name` isn't used anymore so maybe it could be just repurposed to be `report.id.toString()`?

prime-router/src/main/kotlin/history/DeliveryHistory.kt
```diff
@@ -83,13 +81,7 @@ class DeliveryHistory(
...
     val fileName: String
         get() {
-            return Report.formExternalFilename(
-                this.bodyUrl,
-                ReportId.fromString(this.reportId),
-                this.schemaName,
-                Report.Format.safeValueOf(this.bodyFormat),
-                this.createdAt
-            )
+            return this.bodyUrl ?: ""
```
mkalish, line 84: This doesn't seem quite right?  Doesn't the frontend show the something like "report.hl7"?

prime-router/src/testIntegration/resources/quicktests/schema-translations.csv
mkalish, line 1: Not super familiar with these tests, but just confirming that the changes here are just reflecting the internal updates?

mkalish (changes requested):
Nice work!  Lots of meaty changes here.

Main change I saw was some adjustments to the new events and then maybe some doc updates that explain how names work now?

I'll plan to run testing locally after some updates.

prime-router/src/main/kotlin/Report.kt
brick-green: My thought was that it's only needed if `isSendOriginal == true` so it was better here. What do you think?

prime-router/src/main/kotlin/azure/observability/event/ReportReceivedEvent.kt
brick-green: changed to payloadName

prime-router/src/main/kotlin/Report.kt
brick-green: removed property

prime-router/src/main/kotlin/Report.kt
brick-green: updated comment

prime-router/src/main/kotlin/Report.kt
brick-green: moved the getRootReport() call

prime-router/src/main/kotlin/Report.kt
brick-green: replaced with valueOfIgnoreCase()

prime-router/src/main/kotlin/history/DeliveryHistory.kt
brick-green: updated to get filename from bodyUrl

prime-router/src/main/kotlin/azure/observability/event/ReportSentEvent.kt
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Event definition for when a report has been successfully sent to a receiver.
+ * [trasnportType] is nullable as some receivers will purposefully not have transports configured
+ */
+data class ReportSentEvent(
+    val rootReportId: List<ReportId>,
+    val sentReportId: ReportId,
+    val topic: Topic,
+    val senderName: List<String?>,
```
mkalish, line 17: Nit: maybe just filter out the nulls?

prime-router/src/main/kotlin/azure/observability/event/ReportSentEvent.kt
```diff
@@ -0,0 +1,38 @@
...
+        receiver: Receiver,
+        reportFiles: List<ReportFile>,
+        reportId: ReportId,
+        externalFilename: String,
+    ) :
+        this(
+            reportFiles.map { it.reportId },
+            reportId,
+            receiver.topic,
+            reportFiles.map { Sender.createFullName(it.sendingOrg, it.sendingOrgClient) },
```
mkalish, line 33: (I think thats the keywrod)

```suggestion
            reportFiles.mapNotNull { Sender.createFullName(it.sendingOrg, it.sendingOrgClient) },
```

prime-router/src/main/kotlin/history/SubmissionHistory.kt
```diff
@@ -79,16 +77,7 @@ open class SubmissionHistory(
...
-                return Report.formExternalFilename(
-                    this.bodyUrl,
-                    ReportId.fromString(this.reportId),
-                    this.schemaName,
-                    Report.Format.safeValueOf(this.bodyFormat),
-                    this.createdAt
-                )
-            }
-            return ""
+            return this.bodyUrl ?: ""
```
mkalish, line 80: Does this need to line up with DeliveryHistory?

mkalish (approved):
Great work!

Smoke tests all pass, events look good and the names look good in the azure storage.

arnejduranovic (approved):

prime-router/src/main/kotlin/Report.kt
```diff
@@ -162,11 +159,7 @@ data class ReportStreamFilterResult(
  * unique id and name as well as list of sources for the creation of the report.
  */
 class Report : Logging {
-    enum class Format(
-        val ext: String,
-        val mimeType: String,
-        val isSingleItemFormat: Boolean = false,
-    ) {
+    enum class Format(val ext: String, val mimeType: String, val isSingleItemFormat: Boolean = false) {
```
adegolier, line 162: I know this is nitpicky and I wasn't going to say anything about it, but it looks like you just changed this parameter list from being vertical to being horizontal and in another spot I'll call out below it looks like the only change is from horizontal to vertical.

prime-router/src/main/kotlin/Report.kt
```diff
@@ -494,21 +472,20 @@ class Report : Logging {
 
     @Suppress("Destructure")
     private fun createTable(schema: Schema, values: List<List<String>>): Table {
-        fun valuesToColumns(schema: Schema, values: List<List<String>>): List<Column<*>> {
-            return schema.elements.mapIndexed { index, element ->
+        fun valuesToColumns(
+            schema: Schema,
+            values: List<List<String>>,
```
adegolier, line 477: Horizontal to vertical, though I guess there is also the equals in here.

prime-router/src/main/kotlin/Report.kt
brick-green: ah yeah that's my auto formatter at work. I'll got back and try to fix it

adegolier: Noooo not the auto formatter changing some to horizontal and some to vertical 😭

PR #14729, opened 2024-06-13
Fixed a bug while running the validate schemas command

victor-chaparro:
This PR fixes a bug while running the command `./prime validateSchemas`. The FhirToHl7Converter was missing a context parameter that is used to translate messages.

Test Steps:
1. Run `./prime validateSchemas` against staging and verify that schemas are valid

## Changes
- Fixed a bug while running the command `./prime validateSchemas`.


mkalish (approved):
👍🏻

PR #14869, opened 2024-06-24
REST Transport - Issuer and Audience Customization

jorg3lopez:
This PR introduces the capability to use a map in the settings file to generate JWTs for REST transport. The implementation focuses on retrieving values under the `iss` (issuer) and `aud` (audience) keys. This enhancement allows for the creation of JWTs with custom issuer and audience fields, applicable specifically to the `"two-legged"` `authType`.

This feature is necessary to establish a connection with EPIC - UCSD using `OAuth2`. Although TI is the issuer of the token, we need to utilize the client ID provided by EPIC to create the JWT. With these settings, we can modify the issuer from TI-sender to EPIC's client ID and set the audience to EPIC's token endpoint.

Test Steps:
1. Run `gradlew :prime-router:clean package`
2. Run `gradlew :prime-router:testIntegration`
3. Create receiver with REST transport with `authType: "two-legged`
4. Add `jwtParams:` under `transport:`
5. Stub `"iss":""` and `"aud":""` with mock entries, under `jwtParams` ** see [Reference section](#reference) **
7. Fire up the containers
8. Add the new receiver 
9. Send file to api
10. Make sure jwt gets created using the `iss` and `aud` from the `jwtParams` map

## Changes
- `TransportType.kt`
  - Added `jwtParams` map as part of constructor
- `RESTTransport.kt`
  - Added functionality to use the `jwtParams` map to generate jwt
- `RESTTransportIntegrationTests.kt`
  - Added integration test to account when using `jwtParams` to create jwt
- `gitleaks-config.toml`
  - Added regex to allow line of code to be excluded from gitleaks

### Testing
- [x] Tested locally?
- [x] Added tests?

<a name="reference"></a>
## Reference
transport settings using `jwtParams`
image


prime-router/src/main/kotlin/transport/RESTTransport.kt
```diff
@@ -263,16 +264,22 @@ class RESTTransport(private val httpClient: HttpClient? = null) : ITransport {
...
+                Organization(
+                    receiver.name, receiver.fullName,
+                    Organization.Jurisdiction.FEDERAL, null, null, null, null, null
+                ),
+                "", privateKey, ""
+            )
+        } else {
+            val issuer = jwtPamams["issuer"] ?: receiver.name
+            val audience = jwtPamams["audience"] ?: ""
+            AuthUtils.generateToken(issuer, audience, privateKey, "", null)
```
jorg3lopez, line 281: We need to add the jti. I think we can do something similar like `generateOrganizationToken` does it, thoughts?

jorg3lopez: Decided to pass the UUID instead of generating it here so we keep the functionality intact

prime-router/src/main/kotlin/TransportType.kt
```diff
@@ -126,6 +126,8 @@ data class RESTTransportType
     val tlsKeystore: String? = null,
     /** [parameters] The map of parameters to be sent in the message */
     val parameters: Map<String, String> = emptyMap(),
+    /** The map of parameters for constructing the jwt */
+    val jwtParams: Map<String, String> = emptyMap(),
```
oslynn, line 130: Why we can't use "parameters:" instead of introduce a new "jwtParams:"?

jorg3lopez: Good question. I think "jwtParams:" is used specifically to customize the jwt. The "parameters:" map is used to construct the header. My thought process was to keep the responsibilities separated.

jorg3lopez:
Thoughts on adding documentation under the [rest.md](https://github.com/CDCgov/prime-reportstream/blob/master/prime-router/docs/onboarding-users/transport/rest.md) file?

snesm (approved):

oslynn (approved):

prime-router/src/main/kotlin/transport/RESTTransport.kt
```diff
@@ -255,16 +257,22 @@ class RESTTransport(private val httpClient: HttpClient? = null) : ITransport {
...
+            AuthUtils.generateOrganizationToken(
+                Organization(
+                    receiver.name, receiver.fullName,
+                    Organization.Jurisdiction.FEDERAL, null, null, null, null, null
+                ),
+                "", privateKey, ""
+            )
+        } else {
+            val issuer = jwtParams["iss"] ?: receiver.name
+            val audience = jwtParams["aud"] ?: ""
```
victor-chaparro, line 273: I'm trying to figure out if the issuer and the audience is sensitive information and should be securely stored along the receiver credentials of having it on the receiver settings is fine?
Would it be better to have a new auth-type called "two-legged-jwt" and use that instead of checking if jwt-params is not empty?
Is it safe to default to the receiver.name  if issuer is empty?

jorg3lopez: These are excellent questions @victor-chaparro . Thank you for your patience. We are currently determining if the issuer field constitutes sensitive information. If it is sensitive, it should not be included in the repository. For now, the issuer is used only for testing EPIC's Sandbox environment and will differ in the production environment. If we retain this configuration, we can update the settings file via the UI or CLI and include the `issuer` there.

Regarding the `audience`, EPIC has it disclosed on their public [documentation ](https://fhir.epic.com/Documentation?docId=oauth2&section=BackendOAuth2Guide), so it is not considered sensitive information. 

We default the issuer to `receiver.name` to maintain pre-existing functionality and behavior. `AuthUtils.generateOrganizationToken` uses  `receiver.name`  for the `issuer` and leaves the `audience` blank. If the jwtParams map is empty, we revert to the previous settings as part of the null handling process.

I am open to the suggestion of creating a new `authType`. However, adding a map for JWT customization keeps this feature within the `two-legged` `authType` without creating a duplicate `authType` with an additional feature. This approach also allows for future extension to other `authType` values that generate JWTs. For now, it will be part of the `two-legged` `authType`, but we can extend it to other types if needed. Thoughts?

victor-chaparro: Thanks for the details responses @jorg3lopez. I'll go ahead and approve this.

victor-chaparro (approved):

Issue #14721, opened 2024-06-12
Ensure JOOQ is serializing fields `sender_ip` and `external_name` in a safe manner and that said values are valid. 


jack-h-wang:
### User Story
As ReportStream, I want inputs from external sources sanitized so that I can be certain data is valid and does not pose a security concern.

### Description/Use Case
As a best practice, we should ensure that all database access is sanitized so that unintended consequences cannot occur. We should look at both incoming data when performing writes as well as the methods by which the database is accessed when processing API calls.

### Dev Notes
Some examples that may need to be addressed:
- `action`.`sender_ip` is in some cases taken from HttpRequestMessage headers. We could attempt to validate the headers contain valid IP addresses.
- `report_file`.`external_name` is in some cases taken in by the submission endpoint as `payloadName`. We could attempt to ensure this name does not contain invalid characters.
- Ensure database access is strictly via jOOQ using methods that do not introduce SQL injection vulnerabilities. See [jOOQ documentation](https://www.jooq.org/doc/latest/manual/reference/security/security-sql-injection/)

### Acceptance Criteria
- [ ] The serialization of `sender_ip` to the database from submissions is being serialized by JOOQ safely. If this is not the case refactor so we're using JOOQ safely. 
- [ ] prior to serialization ensure `sender_ip` is a valid ip address ([commons](https://commons.apache.org/proper/commons-validator/apidocs/org/apache/commons/validator/routines/InetAddressValidator.html) has a ready bake way to do this). If validation fails ~reject operation on error~ don't set the ip. 
- [ ] The serialization of `external_name` to the database from submissions is being serialized by JOOQ safely. If this is not the case refactor so we're using JOOQ safely. 
- ~prior to serializtion of `external_name` to the database ensure the value is a valid unix/windows filename. You should be able to use one of the implementations of Path to check for valid filename w/o touching the actual file system. If validation fails reject operation on error~
- [ ] Unit test ensuring malicious input to those fields is sanitized before it lands in the db

Andrey-Glazkv:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/14721?planning-poker) with Zenhub @adegolier @arnejduranovic @brick-green @david-navapbc @jack-h-wang @jalbinson @JFisk42 @mkalish @thetaurean

david-navapbc:
[...]

PR #14889, opened 2024-06-25
14721: verify action params and SQL param escaping

mkalish:
This PR adds some data sanitization and validation around the sender ip and external name fields that get in the action log and report file tables respectively.

Test Steps:
There is decent test coverage, but the external name can be tested

1. Send a report with the `payloadname` header with a value ~3000 characters long
2. Make sure the report is not accepted

## Changes
- Validates that the sender IP is a real IP before inserting it
- Errors the report if the payloadname is larger the external name db column
- Switches to use jackson for some of the action params
- Adds some more tests

### Testing
- [ ] Tested locally?
- [ ] Added tests?

### Process

## Linked Issues
- #14721


prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -150,43 +153,61 @@ class ActionHistory(
...
+        val url: String,
+        @JsonProperty("Headers")
+        val headers: Map<String, String>,
+        @JsonProperty("QueryParameters")
+        val queryParameters: Map<String, String>,
+    ) {
+        companion object {
+            // These header/query params can potentially contain auth information, and we do not want to log them
+            // gitleaks:allow
+            private val parameterNamesToExclude = listOf(
```
jalbinson, line 171: could save a few cycles by using `setOf` since were iterating through this (albeit short) list for every header.

prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -195,11 +216,17 @@ class ActionHistory(
...
-        request.headers["x-forwarded-for"]?.let {
-            action.senderIp = it.split(",").firstOrNull()?.trim()?.take(ACTION.SENDER_IP.dataType.length())
+        val senderIp =
+            (
+                (
+                    request.headers["x-forwarded-for"]?.split(",")
+                        ?.firstOrNull()
+                    )?.take(ACTION.SENDER_IP.dataType.length()) ?: request.headers["x-azure-clientip"]
+                )
+        if (senderIp != null && InetAddressValidator.getInstance().isValid(senderIp)) {
```
jalbinson, line 226: always love when apache commons handles the tricky stuff

prime-router/src/main/kotlin/azure/RequestFunction.kt
```diff
@@ -50,14 +50,24 @@ abstract class RequestFunction(
             ?: request.queryParameters.getOrDefault(CLIENT_PARAMETER, "")
     }
 
+    class InvalidExternalPayloadException(val externalName: String) : RuntimeException()
```
jalbinson, line 53: You should pass the `externalName` along to the `RuntimeException` constructor as the message so we see it when it gets thrown.

prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -150,43 +153,61 @@ class ActionHistory(
...
+            )
 
-        // Extract query parameters
-        val filteredParams = request.queryParameters.entries
-            .filter { (key, _) ->
-                notAllowedParameterParts.none { part ->
-                    key.contains(part, ignoreCase = true)
+            fun removeExcludedParameters(it: Map.Entry<String, String>) =
+                !parameterNamesToExclude.any { excluded ->
+                    it.key.contains(
```
jalbinson, line 180: We could mistakenly be filtering out any header that contains the substring listed out in `parameterNamesToExclude`.

Example: `WWW-Authenticate` is gone because it contains `auth`.

prime-router/src/main/kotlin/azure/RequestFunction.kt
```diff
@@ -50,14 +50,24 @@ abstract class RequestFunction(
...
      */
-    protected fun extractPayloadName(request: HttpRequestMessage<String?>): String? {
+    internal fun extractPayloadName(request: HttpRequestMessage<String?>): String? {
         // payloadName can be in the header or in the url parameters.  Return null if not found.
-        return request.headers[PAYLOAD_NAME_PARAMETER]
+        val payloadName = request.headers[PAYLOAD_NAME_PARAMETER]
             ?: request.queryParameters[PAYLOAD_NAME_PARAMETER]
+
+        payloadName?.let {
+            if (it.length > ReportFile.REPORT_FILE.EXTERNAL_NAME.dataType.length()) {
```
jalbinson, line 65: Does `ReportFile.REPORT_FILE.EXTERNAL_NAME.dataType.length() == 2048` because the column is a `varchar(2048)`?

prime-router/src/main/kotlin/azure/RequestFunction.kt
mkalish: woops meant to do that

prime-router/src/main/kotlin/azure/ActionHistory.kt
mkalish: I agree...but I wanted to add the underlying logic the same

prime-router/src/main/kotlin/azure/ActionHistory.kt
mkalish: will update

prime-router/src/main/kotlin/azure/RequestFunction.kt
mkalish: Yes that is correct

jalbinson (approved):

arnejduranovic (approved):

PR #14900, opened 2024-06-26
Add validation during org creation via CLI tool

### LINKED ISSUE #14764: Add validation during organization creation CLI tool
### User Story
We want organizations attempting to be created to be validated.

### Description/Use Case
During a CLI call to created any organization/sender/receiver, call our validation code on the file passed in. Add in a new flag to skip this process if necessary.

- We need a flag to bypass the validation step just in case there is a situation where an update is needed NOW (like during a meeting with a STILT) and we don't have time to update the json schem or konform validation.

### Dev Notes
- Call our validation code in `PutSettingCommand`
- Skip validation if our new flag is present (something like --no-validation)
- print out errors if there are any so the user can fix them

### Acceptance Criteria
- validation code called on all creation and updates to organizations
- new flag for skipping validation works
- errors are printed
- unit tests

adegolier:
Added validation to the settings commands. It works with `./prime multiple-settings set` and `./prime organization set` right now since we don't have a validation profile for just a receiver or just a sender, and will attempt to validate the input file before attempting to set it.

Test Steps:
1. run SettingCommandsTests.kt

## Changes
- Added validation for settings commands before a settings file is set
- Added tests for settings commands, specifically checking the validation
- Pulled shared logic shared between the settings commands and the validation commands out into ValidateUtilities.kt
- Fixed help descriptions on validation commands
- Enabled the UNWRAP_SINGLE_VALUE_ARRAYS deserialization feature in the yaml mapper to allow yaml files with only a single organization to be read

### Testing
- [X] Tested locally?
- [X] Added tests?

## Linked Issues
- Fixes #14764


prime-router/src/main/kotlin/cli/SettingCommands.kt
```diff
@@ -619,10 +631,37 @@ abstract class PutSettingCommand(
...
+        if (skipValidation) {
+            echo(yellow("${inputFile.name} will not be validated."))
+        } else if (useJson) {
+            echo(
+                red(
+                    "JSON files cannot be validated at this time! " +
+                        "You can use the --no-validation flag to avoid this prompt."
+                )
+            )
+            confirm("Proceed anyway?")
```
jalbinson, line 649: the confirmation check is a nice touch.

prime-router/src/main/kotlin/cli/ValidateUtilities.kt
```diff
@@ -0,0 +1,61 @@
...
+import gov.cdc.prime.router.config.validation.ConfigurationType
+import gov.cdc.prime.router.config.validation.ConfigurationValidationFailure
+import gov.cdc.prime.router.config.validation.ConfigurationValidationResult
+import gov.cdc.prime.router.config.validation.ConfigurationValidationService
+import gov.cdc.prime.router.config.validation.ConfigurationValidationServiceImpl
+import gov.cdc.prime.router.config.validation.ConfigurationValidationSuccess
+import java.io.File
+
+class ValidateUtilities() {
+    companion object ValidateUtilities {
```
jalbinson, line 15: rather than adding a companion object to a class, you can just make the object top level.

```kotlin
object ValidateUtilities {
  // ...
}
```

prime-router/src/main/kotlin/cli/ValidateUtilities.kt
```diff
@@ -0,0 +1,61 @@
...
+        val service: ConfigurationValidationService = ConfigurationValidationServiceImpl()
```
jalbinson, line 16: this should be passed in as a dependency which would mean this logic should not be in a companion object.

```kotlin
class ValidateUtilities(
  val service: ConfigurationValidationService = ConfigurationValidationServiceImpl()
) {
  // ...
}
```

prime-router/src/main/kotlin/cli/ValidateUtilities.kt
```diff
@@ -0,0 +1,61 @@
...
+
+        fun validateFiles(
```
jalbinson, line 18: Is there any way to make calls to `ValidateYAMLCommand` since this code all seems pretty similar?

Maybe theres a way to run a command from a command otherwise you could directly call some of the functions.

jalbinson: ah hadn't scrolled down enough. Looks like that was the impetus for this object.

prime-router/src/main/kotlin/cli/ValidateUtilities.kt
adegolier: I made it consistent with the other utilities from the CLI. Why would it be passed in as a dependency if we want the validation to function the same way from different commands in the CLI?

arnejduranovic: The other example I saw in the CLI project was `CommandUtilities`, which does not take in a service. What Jamie is suggesting is more flexible and more easily unit testable (just pass in a mock when creating a ValidateUtilities object).

My vote is to implement Jamie's suggestion.

arnejduranovic (changes requested):

arnejduranovic (approved):

PR #14909, opened 2024-06-27
Experience/14762/get deliveries history api

### LINKED ISSUE #14762: Update Daily Data Table API to enable pagination
## Problem statement
Since discovering the bug on Daily Data which is causing a pagination issue #14742, we have rolled back to a previous version of the table and pagination. Now we need to fix the pagination and update the table component so it works as planned.

## What you need to know

- #14240 is the original table component update that caused the issue
- #14742 addresses the issue that was found and rolled back the updated table component
- Per discussion within FE engineering, the pagination should be handled by the API/Backend. That will need to be updated first, and then the rest can be 

## Acceptance criteria

- [x] Pagination is handled by the API/Backend and is no longer complex code handled by the front-end
- [x] Create a new deliveries endpoint in case we need to point to the existing one
- [x] Unit tests
- [x] Update swagger docs

penny-lischer:
This PR ...

Creates a new deliveries endpoint to allow the API to handle the pagination and sorting of fetched data.

Test Steps:
1. In your favorite API tester call the following:
http://localhost:7071/api/v1/waters/org/ignore/deliveries?receivingOrgSvcStatus=ACTIVE,TESTING
with the following body:
```
{
    "sort": {
        "direction": "DESC",
        "property": "createdAt"
    },
    "pagination": {
        "page": 1,
        "limit": 10
    },
    "filters": [
        {
            "filterName": "SINCE",
            "value": "2000-01-01T00:00:00.000Z"
        },
        {
            "filterName": "UNTIL",
            "value": "2024-05-01T08:00:00.000Z"
        }
    ]
}
```

## Changes
- New azure function called `getDeliveriesHistory`
- Will return pagination and total count in the response
Screenshot 2024-07-01 at 12 47 44 PM

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- Fixes #14762


prime-router/src/test/kotlin/history/azure/DeliveryFunctionTests.kt
```diff
@@ -1555,6 +1568,422 @@ class DeliveryFunctionTests : Logging {
...
+            val jwt = mapOf("organization" to listOf(oktaSystemAdminGroup), "sub" to "test@cdc.gov")
+            val claims = AuthenticatedClaims(jwt, AuthenticationType.Okta)
+
+            mockkObject(AuthenticatedClaims)
+            every { AuthenticatedClaims.Companion.authenticate(any()) } returns claims
+            mockkConstructor(DeliveryHistoryDatabaseAccess::class)
+            every {
+                anyConstructed<DeliveryHistoryDatabaseAccess>().getDeliveries(
+                    any(),
+                    organization1.name, null, null, null, null
```
mkalish, line 1708: It doesn't look like this matches what the mock is getting called with (looks like receiver parameter is getting passed).

After that and fixing the assertion on 1764L the test passed for me

jpandersen87 (approved):

prime-router/src/main/kotlin/history/db/DeliveryHistoryDatabaseAccess.kt
```diff
@@ -0,0 +1,316 @@
...
+        orgService: String?,
+        receivingOrgSvcStatus: List<CustomerStatus>?,
+        reportIdStr: String?,
+        fileName: String?,
+    ): ApiSearchResult<DeliveryHistory> {
+        require(organization.isNotBlank()) {
+            "Invalid organization."
+        }
+
+        var reportId: UUID?
```
mkalish, line 219: nit: this doesn't need to a `var` can can just be assigned to the try

prime-router/src/main/kotlin/history/db/DeliveryHistoryDatabaseAccess.kt
```diff
@@ -0,0 +1,316 @@
...
+        organization: String,
+        orgService: String?,
+        receivingOrgSvcStatus: List<CustomerStatus>?,
+        reportId: UUID?,
+        fileName: String?,
+    ): Condition {
+        var filter = databaseDeliveryAccess.organizationFilter(organization, orgService)
+
+        if (receivingOrgSvcStatus != null) {
+            var statusList = receivingOrgSvcStatus.map { it.name.lowercase() }
```
mkalish, line 299: nit: `val`

shared/build.gradle.kts
mkalish, line 1: Can we revert all of these changes?  To run formatting against just your work it should be `./gradlew :prime-router:ktlintFormat`

mkalish (approved):
Couple of small things, but great work!

Please revert the changes in the other projects before merging

mkalish:
@penny-lischer FYI, there are some compile errors

mkalish (approved):
Good stuff!

victor-chaparro (approved):
LGTM!

Issue #14350, opened 2024-05-09
Clean up the submission/delivery code


mkalish:
### User Story
As a RS developer, I would like the delivery submission endpoints to work effectively and to be well tested

### Description/Use Case
The code for these endpoints is brittle and does not support easily iterating on the pipeline.  Since the report lineage is agnostic of the steps the code should be able to be refactored to not need to specifically pull data out of the specifically named steps.

### Risks/Impacts/Considerations
- The endpoints no longer working as expected; however the understanding is that they are rarely used so the risk is low

### Dev Notes
- The recursive query can likely be replaced in favor the new report graph code

The problem code is below.  The issue is that the route steps are hardcoded rather than being flexible to support arbitrary steps
```kt
descendants.filter { it.actionName == TaskAction.translate }.forEach { descendant ->
                enrichWithTranslateAction(descendant)
            }
            descendants.filter { it.actionName == TaskAction.route }.forEach { descendant ->
                enrichWithRouteAction(descendant)
            }
            descendants.filter { it.actionName == TaskAction.convert }.forEach { descendant ->
                enrichWithConvertAction(descendant)
            }
```

### Acceptance Criteria
- [ ] The delivery/submission endpoints are flexible enough to support any adjustments to the UP steps without having to be amended
- [ ] Database tests are written to cover the functionality

Andrey-Glazkv:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/14350?planning-poker) with Zenhub @adegolier @arnejduranovic @brick-green @jack-h-wang @jalbinson @mkalish @thetaurean

Andrey-Glazkv:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/14350?planning-poker) with Zenhub @JFisk42

Andrey-Glazkv:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/14350?planning-poker) with Zenhub @david-navapbc

david-navapbc:
talked with @mkalish and got some clarity as to the scope of the ticket 

[...]

PR #14951, opened 2024-07-01
14350: update submission to be agnostic of steps

mkalish:
This PR updates the submission history endpoint to be agnostic of the underlying action names

Test Steps:
On master
1. Send a submission
2. Get the report Id and execute the submission endpoint
```
curl --location 'http://localhost:7071/api/waters/report/{REPORT_ID}/history' \
--header 'authentication-type: okta' \
--header 'Authorization: Bearer dummy'
```

Do the same on this branch and confirm the outputs are the same (they might differ in order FYI)

## Changes
- Updates how the submission history is generated
- Adds the intellij project settings
- Adds DB tests for the submission history endpoints

### Testing
- [x] Tested locally?
- [ ] Added tests?

### Process

## Linked Issues
#14350


.idea/codeStyles/Project.xml
mkalish, line 1: Unrelated, but noticed that these should live a directory above to take affect for everyone

prime-router/src/main/kotlin/history/db/ReportGraph.kt
```diff
@@ -408,15 +410,22 @@ class ReportGraph(
...
+            .withRecursive(cte)
+            .select(REPORT_FILE.asterisk())
+            .distinctOn(REPORT_FILE.REPORT_ID)
+            .from(cte)
+            .join(REPORT_FILE)
+            .on(REPORT_FILE.REPORT_ID.eq(cte.field(0, UUID::class.java)))
+            .join(ACTION)
+            .on(ACTION.ACTION_ID.eq(REPORT_FILE.ACTION_ID))
+
+        if (searchedForTaskActions != null) {
```
mkalish, line 425: In the case for the submission history we want to be look at the entire report graph

prime-router/src/test/kotlin/azure/HttpTestUtils.kt
```diff
@@ -15,13 +15,14 @@ import kotlin.collections.Map
...
     var httpStatus: HttpStatusType = HttpStatus.OK
     var content: Any? = null
+    var headers: MutableMap<String, String> = mutableMapOf()
 
     override fun getStatus(): HttpStatusType {
         return this.httpStatus
     }
 
     override fun getHeader(var1: String): String {
-        return "world"
```
mkalish, line 24: Not sure what the thought was here, but this needed to change in order to test that the location was returned correctly

prime-router/src/test/kotlin/history/SubmissionHistoryTests.kt
mkalish, line 1: I pretty aggressively deleted code here, but there is definitely more clean up to do here.  I've tried to keep the scope small and only delete stuff that no longer applies

prime-router/src/test/kotlin/history/azure/SubmissionFunctionIntegrationTests.kt
```diff
@@ -0,0 +1,541 @@
...
+import org.junit.jupiter.api.extension.ExtendWith
+import org.testcontainers.junit.jupiter.Testcontainers
+import java.time.OffsetDateTime
+import java.util.UUID
+
+@Testcontainers
+@ExtendWith(ReportStreamTestDatabaseSetupExtension::class)
+class SubmissionFunctionIntegrationTests {
+
+    class ReportNodeBuilder {
```
mkalish, line 45: Leaving this here for now and it can get moved once another test wants similar setup

prime-router/src/main/kotlin/azure/DatabaseAccess.kt
```diff
@@ -1347,8 +1347,8 @@ class DatabaseAccess(val create: DSLContext) : Logging {
     /**
      * Inserts the provided [actionLog] using [txn] as the data context.
      */
-    private fun insertActionLog(actionLog: ActionLog, txn: Configuration) {
-        val detailRecord = DSL.using(txn).newRecord(Tables.ACTION_LOG, actionLog)
+    fun insertActionLog(actionLog: ActionLog, txn: Configuration) {
```
mkalish, line 1350: This is the only change, the rest are ktlint cleaning up

prime-router/src/main/kotlin/history/SubmissionHistory.kt
```diff
@@ -240,47 +239,76 @@ class DetailedSubmissionHistory(
...
+                } else {
+                    existingDestination.sentReports.addAll(sentReports)
+                    existingDestination.downloadedReports.addAll(downloadedReports)
+                    if (report.nextActionAt != null) {
+                        existingDestination.sendingAt = report.nextActionAt
+                    }
+                }
             }
 
+            // TODO: does this ever run?
```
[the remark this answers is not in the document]
arnejduranovic: Also, where else would things like topic and sender be populated if this code doesn't run?

arnejduranovic (commented):
Changes and tests look good! However, do we need to update the end2end tests to validate the submission history response to ensure everything works as expected with the new route step refactor? I think it'll have to be a new ticket unless we want to slightly refactor it after the refactor?

.idea/codeStyles/Project.xml
```diff
@@ -0,0 +1,25 @@
+<component name="ProjectCodeStyleConfiguration">
```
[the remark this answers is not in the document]
mkalish: This is actually a new file.  It had previously been in the the prime-router directory, but since we open the project here now I'm re-adding it

prime-router/src/main/kotlin/history/SubmissionHistory.kt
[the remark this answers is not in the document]
mkalish: Sorry, I had meant to delete this. I had confirmed that this does run

mkalish:
@arnejduranovic The end2end_up is now written in such a way that is also agnostic of the steps and is simply using the history endpoint to get the name of the file to check in the SFTP folder, so no changes required.

david-navapbc:
when I attempt to build this branch I get a database flyway error. 

```
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':prime-router:flywayMigrate'.
> Error occurred while executing flywayMigrate
  Unable to obtain connection from database (jdbc:postgresql://localhost:5432/prime_data_hub) for user 'prime': Connection to localhost:5432 refused. Check that the hostname and port are correct and that the postmaster is accepting TCP/IP connections.
  ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
  SQL State  : 08001
  Error Code : 0
  Message    : Connection to localhost:5432 refused. Check that the hostname and port are correct and that the postmaster is accepting TCP/IP connections.
  
  Connection to localhost:5432 refused. Check that the hostname and port are correct and that the postmaster is accepting TCP/IP connections.
  Connection refused
  ```

**Comment by david-navapbc** [comment]
why is the app returning a url for history that uses the reportId vs the subissionId? semantically does it not make more sense to use the id that tags the entire job vs the id that tags the report artifact from the first stage of processing?

**Comment by david-navapbc** [comment]
[in.txt](https://github.com/user-attachments/files/16185818/in.txt)

[out.txt](https://github.com/user-attachments/files/16185819/out.txt)

"in.txt" is an HL7 message I submitted to the app. 

"out.txt" is the response from a query to the history endpoint 

why is `reportItemCount` "1"?  it looks like there are several observations in the hl7 message going in.

File: prime-router/src/main/kotlin/history/azure/SubmissionsFacade.kt [hunk 7]
```diff
@@ -116,30 +122,56 @@ class SubmissionsFacade(
...
-            action.sendingOrg,
-            DetailedSubmissionHistory::class.java
-        )
-        submission?.actionsPerformed?.add(action.actionName)
-
-        // Submissions with a report ID (means had no errors) can have a lineage
-        submission?.reportId?.let {
-            val relatedSubmissions = dbSubmissionAccess.fetchRelatedActions(
-                UUID.fromString(it),
+        if (id == null) {
```
david-navapbc, line 129: this is a nit but we have a lot of "id" object floating around. It might be worthwhile to specify exactly which "id" we're dealing with

prime-router/src/main/kotlin/history/db/ReportGraph.kt
david-navapbc: wasn't the code that was there returning the entire report graph already? Assuming the caller used the endpoint given to them in response to a successful submission did not the legacy code return the descendants in order?

mkalish: It mostly did, but also didn't select out the ReportFile object.  Big picture trying to have the recursive queries be abstracted in one spot.

mkalish:
@david-navapbc For your question on report id vs action id, I think we more commonly talk in terms of report id internally and then when debugging with senders/receivers rather than action id.

david-navapbc (approved):

arnejduranovic (approved):

PR #14976, opened 2024-07-01
Ignore linting/testing steps in build-frontend when in trialfrontend env

### LINKED ISSUE #14975: Ignore linting/testing steps in build-frontend when in trialfrontend env
As it is expected that our PR workflow will handle ensuring branches pass linting/testing before allowing a merge, these steps are unnecessary when deploying to the trialfrontend and their times greatly shrunk if conditionally disabled. This will also allow deployments even if there are problems with staging (causing e2e to fail).

# Acceptance Criteria
- [ ] linting and testing steps in build-frontend action updated with `if` condition checking if env-name does not start with `trialfrontend`

jpandersen87:
Fixes #14975

This PR add `if` conditionals on linting and test steps in order to skip them if the env-name starts with `trialfrontend`. It also:
- removes an obsolete step setting a LOG_FILE env var
- adds check if there are any warnings before outputting frontend-warnings file in link checking test


penny-lischer (approved):
LGTM

snesm (approved):

Issue #14701, opened 2024-06-11
BUG: Authenticated navbar unavailable if not associated with any admin/receiver/sender groups


jpandersen87:
The auth navbar isn't rendered unless a user is part of the three current group types (admin, receiver, sender). This prevents the logout button from being available. The base bar should always be available after login (with logout button), with aspects of it hidden depending on permissions (aka group).

# How to recreate
1. Log in as a user that is not part of any group in Okta
2. The Auth nav bar does not appear

# Acceptance Criteria
- [ ] Auth navbar with user display and logout button available to all logged in users
- [ ] group-based navigation links/sections available to the user if they are associated
  - [ ] If user is not part of any group, they should only see email and logout in the Auth Nav Bar
- [ ] Create user that is not part of any Okta group for E2E tests 
- [ ] Update E2E

penny-lischer:
@chris-kuryak do we need design for this?

chris-kuryak:
Joseph has identified that the fix is a one-line code change.
Bulk of work will be getting with DevOps to check on the change.

PR #15019, opened 2024-07-02
BUG: Login state param initialization of state for widget isn't passed to auth client

### LINKED ISSUE #15018: BUG: Login state param initialization of state for widget isn't passed to auth client
It seems that the okta auth client the widget creates is not properly initializing with the state we're passing. This causes magiclink usage to end with a fallback message of manually using the OTP code. We need to pass out pre-created `oktaAuth` client to the widget, and manually set the state inside it for the widget to properly complete login.

# Acceptance Criteria
- [ ] `oktaAuth` client from session context passed as `authClient` property for widget during initialization
- [ ] `state` url param set to `oktaAuth.options.state` manually

jpandersen87:
Fixes #15018 #14701

This PR (hopefully) fixes magiclinks so that state is properly passed to the okta auth client. It also fixes the nav bar so that it will appear for all authenticated users (regardless of groups). The bar only shows the email and logout buttons for users that do not have the necessary groups for anything else.

### DIFF
--- a/frontend-react/src/AppRouter.tsx
+++ b/frontend-react/src/AppRouter.tsx
@@ -103,6 +103,7 @@ const ReportStreamApiDocumentationPayloads = lazy(
 
 /* Public Pages */
 const TermsOfService = lazy(() => import("./pages/TermsOfService"));
+
 const LoginCallback = lazy(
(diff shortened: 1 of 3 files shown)


penny-lischer (approved):
LGTM

PR #15054, opened 2024-07-03
CDC CA cert updates

supriyaaddagada:
This PR is to to include certs for California into the dockerfile for connecting to https://saphireapistage.cdph.ca.gov/.


jeremy-page (changes requested):
This is kind of a pain but would you please rename the certs "CDC-CA-xxx" to "CA-CA-xxx"? These certificates come from the state of California, not the CDC.

jeremy-page (approved):

victor-chaparro (approved):

PR #15069, opened 2024-07-04
ETOR: Update `la-phl` org settings

basiliskus:
A few things have changed since the `la-phl` org settings were first created. This PR updates all properties with the correct value. Also cleaned up the file to remove any property that's not changing the default value.

Also had to add a missing schema for `encryptionKeyUrl` in `prime-router/metadata/json_schema/organizations/organizations.json` for it to pass the yaml validation 

Test Steps:
1. Load the settings locally: `/prime multiple-settings set -s -i ./settings/STLTs/LA/la-phl.yml`
2. Submit an OML message with the expected structure. We used this one:
[29d4cf15-7504-46b6-9139-78b0037ee5e3.txt](https://github.com/user-attachments/files/16133117/29d4cf15-7504-46b6-9139-78b0037ee5e3.txt)
3. Check that the message was delivered

### Testing
- [x] Tested locally?


prime-router/settings/staging/0162-la-phl-staging-new-rest-setting.yml
```diff
@@ -1,104 +1,49 @@
-# To submit this to localhost, run:
```
oslynn, line 1: We move all of STLTs setting to new location in settings.  In your case, it should be:  ./settings/STLTs/FEXION/la-phl.yml

oslynn (approved):

victor-chaparro (approved):

PR #15125, opened 2024-07-09
Add frontend_ci to frontend changed files list in build-vars

### LINKED ISSUE #15124: BUG: frontend-ci workflow's build-vars action requires pull request read permissions
According to `dorny/paths-filter`'s documentation, it requires pull request read permissions to properly read changed files from the base branch within a pull request context. This may be the cause of erroneous skips that have occurred.

# Acceptance Criteria
- [ ] pull request read granted to `build-vars` step action for `frontend-ci` workflow

jpandersen87:
This PR adds frontend_ci to the list of watched files in build_vars for frontend.


jpandersen87:
> Looks like the job was still skipped? https://github.com/CDCgov/prime-reportstream/actions/runs/9861243019/job/27229266134?pr=15125

That was correct behavior, as the build-vars frontend check didn't include the `frontend_ci` workflow. I just added a commit to add that file to the list.

.github/workflows/frontend_ci.yml
```diff
@@ -23,6 +23,8 @@ jobs:
   pre-job:
     name: Pre Job
     runs-on: ubuntu-latest
+    permissions:
```
snesm, line 26: I believe we already have read/write (the default for [permissive](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) mode), so you are reducing permissions here, rather than adding.

snesm (changes requested):
I''m not quite sure what problem we are trying to solve here.

snesm (approved):

Issue #13925, opened 2024-04-04
Refactor default httpclient as a singleton


thetaurean:
### User Story
As a dev, I want to use a singleton for shared & reused objects so that the codebase remains readable and maintainable.

### Description/Use Case
In #11928 we introduced an HTTPClient that is used throughout the project for making RESTful and other HTTP calls. That client can be specified by way of a parameter to various functions / class constructors in the project. But the parameter is only used for testing. Runtime code is passing in null for the client, which causes `HTTPClientUtils` to create a new default client for every request.

Instead, this client should be instantiated as a singleton and perhaps an inherited member or function of the BaseEngine.

### Risks/Impacts/Considerations
- Be sure to look at the duplicated code in `RESTTransport.kt`

### Dev Notes
- Retrieval or instantiation of the client should be mockable to facilitate simple testing

### Acceptance Criteria
- [ ] No more HTTPClient parameters everywhere
- [ ] Singleton reusable HTTPClient for majority of use casees
- [ ] Can still create one-time use HTTPClients for edge cases
  - Most (all?) cases can just configure the request and override singleton HTTPClient settings.
- [ ] All clients are closed when done

arnejduranovic:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/13925?planning-poker) with Zenhub @brick-green @david-navapbc @jack-h-wang @jalbinson @JFisk42 @mkalish @thetaurean

david-navapbc:
I've refactored `HttpClientUtils.getDefaultHttpClient` to return a singleton instead of a new instance. I've also refactored `invoke()` to include an "also" call to close the client after use. 

The ticket speaks to `RESTTrasport` having duplicate code and indeed it does. The ACs don't speak specifically to migrating that code so I posted a thing on slack asking after it. If there's no objections I'm going to migrate that code to `HttoClientUtils` and ensure all calls to it are properly re-routed. 

Lastly I still need to look into how the clients are being used. The current call to `close()` is only happening if the caller is using the invoke() method inside HttpClientUtilts or in `RESTTransport`

david-navapbc:
I **think** I've run to ground all non-test references to HttpClientUtils. As far as I can tell literally none of them are being created with an httpClient object and thus are are all using the singleton. 

Here is the rundown: 

* LookupTableCommands
[...]

PR #15133, opened 2024-07-10
Platform/13925/singleton http client

david-navapbc:
This PR ...

Addresses issue [13925](https://github.com/CDCgov/prime-reportstream/issues/13925). All callers that use `HttpClientUtils` will get:

* an httpclient that ~~is closed when they are done with it~~ frees resources automatically after request has been made 
* ~~a singleton non-auth http client that can be reused~~
* ~~a semi-singleton auth http client that can be reused so long as each subsequent calls  use the same auth credentials. in case of call that employs different auth credentials the client is remade with the new credentials baked in~~ 

Test Steps:
1. run the tests. there are several new ones that cover the new behavior 

## Changes
- all uses of the client attached to `HttpClientUtils` free connection resources upon completion of request because we're now using the Apache5 engine in place of Apache4 ~~are "closed" when the network call is complete~~ 
- ~~the default non-auth client is now a singleton~~ 
- ~~the default with-auth client is a semi singleton that gets remade only when a caller asks for a client with auth credentials different than the previous client~~
- the httpclient is created once and the method to get the default client is now a simple getter  
- new tests ensuring the new behavior is working as expected

### Testing
- [ x] Tested locally?
- [ x] Ran `./prime test` or `./gradlew testSmoke` against local Docker ReportStream container?
- [ x] Added tests?


prime-router/src/main/kotlin/common/HttpClientUtils.kt
```diff
@@ -530,44 +558,96 @@ class HttpClientUtils {
...
 
-                install(HttpTimeout)
-                engine {
-                    followRedirects = true
-                    socketTimeout = TIMEOUT
-                    connectTimeout = TIMEOUT
-                    connectionRequestTimeout = TIMEOUT
-                    customizeClient {
+        /**
+         * Called by getDefaultHttpClient as a helper to handle clients with auth tokens. Caller handles thread safety.
```
adegolier, line 597: Minor point of confusion... here it says "Caller handles thread safety" but later seems to say that safety is enforced by the way this class is written?

prime-router/src/main/kotlin/common/HttpClientUtils.kt
```diff
@@ -25,13 +25,41 @@ import kotlinx.serialization.json.Json
...
          * timeout for http calls
          */
         private const val TIMEOUT = 50_000
         const val REQUEST_TIMEOUT_MILLIS: Long = 130000 // need to be public to be used by inline
         const val SETTINGS_REQUEST_TIMEOUT_MILLIS = 30000
 
+        /**
+         * resets client and token hash to default
+         */
+        private fun reset() {
```
adegolier, line 57: Is this only here for testing?

prime-router/src/test/kotlin/common/HttpClientUtilsTests.kt
```diff
@@ -0,0 +1,110 @@
...
+
+        // httpClientWithAuth should still be null
+        val httpClientWithAuth: Field = HttpClientUtils::class.java.getDeclaredField("httpClientWithAuth")
+        httpClientWithAuth.trySetAccessible()
+        assertEquals(null, httpClientWithAuth.get(HttpClient::class))
+
+        // accessTokenHash should still be -1
+        val accessTokenHash: Field = HttpClientUtils::class.java.getDeclaredField("accessTokenHash")
+        accessTokenHash.trySetAccessible()
+        assertEquals(0, accessTokenHash.get(Int::class))
```
adegolier, line 60: Comment says the accessTokenHash should be -1, but the assert says it should be 0

adegolier (approved):
Lgtm aside from minor notes, and all the tests pass. Added tests seem appropriate to cover expected behavior.

prime-router/src/main/kotlin/common/HttpClientUtils.kt
david-navapbc: yup. this is a static/companion class and this was the simplest and most grokkable way I could think of to ensure the tests had a "fresh" copy of those vars for each test run.

prime-router/src/test/kotlin/common/HttpClientUtilsTests.kt
david-navapbc: oh! good catch. the hash for a null obj reference is always 0 and I didn't know that until I wrote this which is why it's not defaulting to 0. I'll take care of the comment.

prime-router/src/main/kotlin/common/HttpClientUtils.kt
david-navapbc: I can rework the language. The intention of that comment is two fold: 

1. to alert future-David and other devs that the helper method doesn't enforce thread safety on client obj creation, the caller of the helper does by way of the `synchronized` block. 
2. to remind future-David and other devs that java/kotlin work in a way that prevents obj arguments passed to a method from changing unexpectedly by subsequent calls that change the source of that obj argument and thus THIS race-condition concern is not a concern. 

in other words I'm attempting to document the two likely sources of race-condition shennanigans: 

(1) speaks to what happens when two threads call the "get" method at the same time and why there's not going to be a race condition resulting in a caller receiving an object with arguments baked into it that it doesn't expect to be there. 

(2) speaks to what happens when a caller receives a client obj from the "get" method and, before it can use it to make a network call, a second caller receives a client obj with different arguments from the "get" method. 

I'll update the comment block to see if I can't make it more lucid.

prime-router/src/main/kotlin/common/HttpClientUtils.kt
thetaurean: I would just make this a public method and rename it to `destroy` or `deinit`. It's a functionality that could be useful in the code and cleans up access in the tests. And maybe add a quick unit tests that makes sure destroy functions as expected and doesn't break the class somehow.

thetaurean (approved):
Good stuff 👍 This looked like a fun one. Left some minor nits to discuss.

prime-router/src/main/kotlin/common/HttpClientUtils.kt
arnejduranovic: I don't see a reason why this can't be public? This seems like reasonable functionality a potential client (calling code) would use?

prime-router/src/main/kotlin/common/HttpClientUtils.kt
```diff
@@ -521,7 +549,7 @@ class HttpClientUtils {
                     jsonPayload?.let {
                         setBody(jsonPayload)
                     }
-                }
+                }.also { httpClient?.close() }
```
arnejduranovic, line 552: I think this is the biggest value add of this ticket.

prime-router/src/main/kotlin/common/HttpClientUtils.kt
```diff
@@ -530,44 +558,96 @@ class HttpClientUtils {
...
+                                )
+                            }
+                        install(HttpTimeout)
+                        engine {
+                                followRedirects = true
+                                socketTimeout = TIMEOUT
+                                connectTimeout = TIMEOUT
+                                connectionRequestTimeout = TIMEOUT
+                                customizeClient {
+                                }
```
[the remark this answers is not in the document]
david-navapbc: I did not look closely at it. Do you want it gone? I'm not wed to it being there. I carried it over as you said

prime-router/src/main/kotlin/common/HttpClientUtils.kt
david-navapbc: we can easily make it public in future if some other prod code needs it. Until we actually need that functionality though I'd rather keep it private and prevent future shenanigans. The point of this ticket was to reduce the number of times the client gets re-created vs re-used and given that we don't have a specific need for a non-test caller to reset the state of the companion fields I worry that some future dev is going to - with good intentions - activate reset() when it's not necessary and thus reduce the value of the work in this PR .

prime-router/src/main/kotlin/common/HttpClientUtils.kt
```diff
@@ -530,44 +558,96 @@ class HttpClientUtils {
...
+         *
+         * In other words, in the case where two callers attempt, one immediately after the other, to create and use an
+         * httpClientWithAuth object with differing auth token values, the first caller has a COPY of the stack
+         * reference to the ORIGINAL object, NOT the actual reference which would point to the NEW obj once the NEW obj
+         * is created. Even in the case of an immediate subsequent caller to this method that provides an auth token
+         * value different than the first caller, the ORIGINAL caller has a copy of the reference to the still-in-scope
+         * ORIGINAL obj in the heap, and thus, is using the ORIGINAL obj with the auth token it provided, and the
+         * subsequent caller is using the NEW stack reference copy that points to the NEW client obj that contains the
+         * NEW auth token it expects to be there.
+         */
```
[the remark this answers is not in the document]
david-navapbc: actually - what if I moved it a few lines down so it's out of the docstring? Would that work for you? I'm not fighting for it to be in the docstring. I'm fighting for it to be there co-located with the code. If we move it a few lines down it doesn't show up in any generated javadoc and any dev who looks at the source will still see it.

prime-router/src/main/kotlin/common/HttpClientUtils.kt
thetaurean: MEMOERY LEAKS HATE THIS ONE SIMPLE TRICK

prime-router/src/main/kotlin/common/HttpClientUtils.kt
thetaurean: Renaming it `destroy` or `deinit` might address this ambiguity and reduce/eliminate the likelihood of misuse.

Retrieving the method by name and modifying its visibility in tests screams antipattern/smell to me (I don't feel strongly enough to get my crusading gear though).

There are situations where a `destroy` or `deinit` might be needed: in a previous PR this client used a bearer auth handler that would discard its token if a request resulted in a 403. Since the experience team might further maintain/refine this, we might consider packaging the functionality now (with an informative docstring even).

prime-router/src/main/kotlin/common/HttpClientUtils.kt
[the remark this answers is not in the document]
thetaurean: For the docstring I would take a step back and focus on your key points
- This function provides thread-safe reusable HTTP clients implemented as a private singleton
- Clients requiring authentication are instantiated/reused on a per-credential basis using a token hash
- A description of behavior on repeated calls (with less focus on OO concepts)

I think think these points should move outside the docstring (or remove altogether)
- Description of kotlin byref/byval semantics
- Assurance that previously returned clients aren't destroyed due to above
- Explaining encapsulation / OO facets

>          * **NOTE**  Java and Kotlin both use pass-by-value with reference copy to pass arguments to a method. There is
>          * therefore NO risk of one caller having an httpClientWithAuth obj change out from under them by a subsequent
>          * caller who provides a different auth token.
>          *
>          * The client objects in this class are private and there is no direct reference to them outside the
>          * "getter" methods which are written in a manner that ensures they are thread safe. All the places where we
>          * actually use the httpClientWithAuth obj in this class are scoped to within a method call using a provided
>          * reference copy passed to the method as an argument. All external callers have no access to private members
>          * and thus are forced to use the objs in the same safe manner.
>          *
>          * In other words, in the case where two callers attempt, one immediately after the other, to create and use an
>          * httpClientWithAuth object with differing auth token values, the first caller has a COPY of the stack
>          * reference to the ORIGINAL object, NOT the actual reference which would point to the NEW obj once the NEW obj
>          * is created. Even in the case of an immediate subsequent caller to this method that provides an auth token
>          * value different than the first caller, the ORIGINAL caller has a copy of the reference to the still-in-scope
>          * ORIGINAL obj in the heap, and thus, is using the ORIGINAL obj with the auth token it provided, and the
>          * subsequent caller is using the NEW stack reference copy that points to the NEW client obj that contains the
>          * NEW auth token it expects to be there.

prime-router/src/main/kotlin/common/HttpClientUtils.kt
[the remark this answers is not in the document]
thetaurean: It would be nice to clean up while we're in here.

I did a cursory search through [docs](https://ktor.io/docs/client-engines.html#java) and I don't think an empty `customizeClient {}` block does anything. So we should be able to remove it without any impact.

Or if something does break (ideally not in prod 😅) we can add a comment there.

prime-router/src/main/kotlin/common/HttpClientUtils.kt
david-navapbc: ok I'll make the change

prime-router/src/main/kotlin/common/HttpClientUtils.kt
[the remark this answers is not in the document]
david-navapbc: I'm not ok removing it but it sounds like you're ok with my suggestion of removing it from the docstring which I'll take care of

prime-router/src/main/kotlin/common/HttpClientUtils.kt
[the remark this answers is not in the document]
david-navapbc: I'll take care of it

prime-router/src/main/kotlin/common/HttpClientUtils.kt
david-navapbc: @arnejduranovic the reason is it's called out by Oracle as "best practice" 

> Use the most restrictive access level that makes sense for a particular member. Use private unless you have a > good reason not to.

https://docs.oracle.com/javase/tutorial/java/javaOO/accesscontrol.html

that being said y'all seem to want it public so I'll make it public

david-navapbc: @thetaurean [YAGNI](https://en.wikipedia.org/wiki/You_aren't_gonna_need_it) is why I was looking to keep it private. nowhere in the app is that method being called outside of test code and we can change it to public when we need to. 

that being said - y'all seem to want it public so I'll make it public.

arnejduranovic: Good and valid points, David. I'm happy to approve the PR with either method, I think this is minor and don't think its a show-stopper. 

To me, you ARE using this method somewhere (in test code). The test code is the client calling code of this object and to me it makes sense for this method to be public because of that. I don't think we should add methods to classes just for testing purposes, but in this case I think a "reset" method makes sense for the public facing interface of this class given what it does.

arnejduranovic: Actually, I want to edit and say I think we should not push code that is using reflection, even if its test code. So we should address this before merging.

david-navapbc:
@arnejduranovic made the changes

prime-router/src/test/kotlin/common/HttpClientUtilsTests.kt
```diff
@@ -0,0 +1,37 @@
...
+
+    @BeforeEach
+    fun setup() {
+        HttpClientUtils.reset()
+    }
+
+    @Test
+    fun `test default values for client is null`() {
+        // httpClient should be initialized as null
+        val httpClientField: Field = HttpClientUtils::class.java.getDeclaredField("httpClient")
```
arnejduranovic, line 19: We should avoid reflection where possible. We treat it as a work around for when we need to do something with a library we don't have source code access to, and should limit it to that use case. What we can do here instead is leverage Spyk (see for usages in code) to assert certain branches of getDefaultHttpClient get called instead of verifying the private `httpClient` directly. Although, we may not need this if we apply the suggested change to default httpClient to an actual object instead of null.

prime-router/src/main/kotlin/cli/tests/AuthTests.kt
```diff
@@ -1172,25 +1170,14 @@ class Server2ServerAuthTests : CoolTest() {
...
-        val response = runBlocking {
-            client.get(orgEndpoint) {
-                timeout {
-                    requestTimeoutMillis = 45000
-                    // default timeout is 15s; raising higher due to slow Function startup issues
-                }
-                accept(ContentType.Application.Json)
-            }
-        }
+        val response = HttpClientUtils.get(
```
arnejduranovic, line 1175: I think this is a good change.

prime-router/src/main/kotlin/common/HttpClientUtils.kt
```diff
@@ -312,7 +327,7 @@ class HttpClientUtils {
             httpClient: HttpClient? = null,
```
arnejduranovic, line 327: The suggested patch removes the ability to pass in a custom client, so this deviates from that. I think not applying that change is probably fine, seems more flexible to give clients the option of providing their own client for now. I can't think of any downsides at the moment.

prime-router/src/main/kotlin/common/HttpClientUtils.kt
```diff
@@ -496,32 +515,36 @@ class HttpClientUtils {
...
             return runBlocking {
-                (httpClient ?: createDefaultHttpClient(accessToken)).request(url) {
+                (httpClient ?: getDefaultHttpClient()).request(url) {
                     this.method = method
                     timeout {
                         requestTimeoutMillis = timeout
                     }
                     url {
                         queryParameters?.forEach {
-                            parameter(it.key, it.value.toString())
```
arnejduranovic, line 506: nice :)

prime-router/src/main/kotlin/common/HttpClientUtils.kt
```diff
@@ -530,43 +553,32 @@ class HttpClientUtils {
          * note: most configuration parameters are overridable
          * e.g. expectSuccess default to false because most of the time
          * the caller wants to handle the whole range of response status
-         * @param bearerTokens null default, the access token needed to call the endpoint
+         * @param bearerTokens the access token needed to call the endpoint
```
arnejduranovic, line 556: remove this line?

prime-router/src/main/kotlin/common/HttpClientUtils.kt
```diff
@@ -530,43 +553,32 @@ class HttpClientUtils {
...
-                // install(Logging) {
-                //     logger = Logger.SIMPLE
-                //     level = LogLevel.INFO
-                // }
-                // not using Bearer Auth handler due to refresh token behavior
-                accessToken?.let {
-                    defaultRequest {
-                        header("Authorization", "Bearer $it")
+        fun getDefaultHttpClient(): HttpClient {
+            synchronized(this) {
```
arnejduranovic, line 560: This relates to my comment about having httpClient initialized at the creation of the object instead of here and make it not resettable: we should evaluate if we really need this if we apply the suggest httpClient change.

prime-router/src/test/kotlin/common/HttpClientUtilsTests.kt
```diff
@@ -0,0 +1,37 @@
...
+        httpClientField.trySetAccessible()
+        assertEquals(null, httpClientField.get(HttpClient::class))
+    }
+
+    @Test
+    fun `test httpClient is a singleton`() {
+        // create the httpClient singleton
+        val httpClient1 = HttpClientUtils.getDefaultHttpClient()
+        val httpClientObjRef1 = System.identityHashCode(httpClient1)
```
arnejduranovic, line 28: I don't think this extra step to get the hash is necessary? Have you tried comparing the objects directly? iirc `==` should check for equality by matching references.

david-navapbc: good call - will make the change

prime-router/src/main/kotlin/common/HttpClientUtils.kt
david-navapbc: ack

prime-router/src/main/kotlin/common/HttpClientUtils.kt
david-navapbc: ack

prime-router/src/main/kotlin/common/HttpClientUtils.kt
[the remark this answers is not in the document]
david-navapbc: ack

prime-router/src/main/kotlin/common/HttpClientUtils.kt
david-navapbc: ack

prime-router/src/main/kotlin/common/HttpClientUtils.kt
```diff
@@ -312,35 +330,38 @@ class HttpClientUtils {
             httpClient: HttpClient? = null,
         ): HttpResponse {
             return runBlocking {
-                (httpClient ?: createDefaultHttpClient(accessToken)).submitForm(
-                    url,
-                    formParameters = Parameters.build {
-                        formParams?.forEach { param ->
-                            append(param.key, param.value)
+                // use{...} will automatically close the client
```
arnejduranovic, line 333: Let's remove these comments because use is not anything special to this library, it works on anything that is `Closable`. see: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io/use.html. We should prefer not to document Kotlin features, it's easy enough to look up imo.

prime-router/src/main/kotlin/common/HttpClientUtils.kt
```diff
@@ -496,78 +517,47 @@ class HttpClientUtils {
...
             return runBlocking {
-                (httpClient ?: createDefaultHttpClient(accessToken)).request(url) {
-                    this.method = method
-                    timeout {
-                        requestTimeoutMillis = timeout
-                    }
-                    url {
-                        queryParameters?.forEach {
-                            parameter(it.key, it.value.toString())
+                // use{...} will automatically close the client
```
arnejduranovic, line 520: Remove comment.

prime-router/src/test/kotlin/common/HttpClientUtilsTests.kt
arnejduranovic: bump

prime-router/src/main/kotlin/common/HttpClientUtils.kt
```diff
@@ -496,78 +517,47 @@ class HttpClientUtils {
...
+                            setBody(jsonPayload)
+                        }
                     }
                 }
             }
         }
 
         /**
-         * Create a http client with sensible default settings
+         * Get a http client with sensible default settings
```
arnejduranovic, line 554: This comment block no longer works imo since we turned this into a simple getter now and create the client elsewhere.

arnejduranovic (changes requested):

prime-router/src/test/kotlin/common/HttpClientUtilsTests.kt
david-navapbc: made the change and I agree with the change. in answer to your question - the intention behind that is "System.identityHashCode(...)" can't be overriden in the way ".equals()" and ".hashCode()" can be. 

That being said - I agree the likelihood of us running into trouble because someone overrode ".equals()" in a weird manner is very very low.

prime-router/src/main/kotlin/common/HttpClientUtils.kt
david-navapbc: ack

prime-router/src/main/kotlin/common/HttpClientUtils.kt
david-navapbc: ack

prime-router/src/main/kotlin/common/HttpClientUtils.kt
arnejduranovic: This looks good to me now, I either commented on the wrong thing or commented on an old revision maybe.

prime-router/build.gradle.kts
```diff
@@ -881,7 +881,13 @@ dependencies {
     implementation("io.github.java-diff-utils:java-diff-utils:4.11")
     implementation("io.ktor:ktor-client-core:$ktorVersion")
     implementation("io.ktor:ktor-client-cio:$ktorVersion")
+    // the apache4 client will NOT free resources automatically between requests
```
arnejduranovic, line 884: Before merging this PR, these comments should be removed and a ticket created instead to update other usages (wait until PR gets fully approved though in case something comes up)

thetaurean (approved):
Looks good 👍 the invoke magic is interesting

prime-router/build.gradle.kts
mkalish: @david-navapbc Bumping this comment

prime-router/src/main/kotlin/common/HttpClientUtils.kt
```diff
@@ -32,10 +30,30 @@ class HttpClientUtils {
         const val REQUEST_TIMEOUT_MILLIS: Long = 130000 // need to be public to be used by inline
         const val SETTINGS_REQUEST_TIMEOUT_MILLIS = 30000
 
+        private var httpClient: HttpClient =
```
mkalish, line 33: Nit 
```suggestion
        private val httpClient: HttpClient =
```

mkalish (approved):
Changes look good to me, I just think we need to have the ticket created and the code tagged with it before merging.

prime-router/build.gradle.kts
david-navapbc: ack

prime-router/src/main/kotlin/common/HttpClientUtils.kt
david-navapbc: ack

arnejduranovic (approved):

Issue #15157, opened 2024-07-11
NM - Result File Name Changes


Jcavallo7:
## Problem statement

" It looks like we have started to receive files that do not adhere to the conventional file naming. Here are some examples…

covid-19-7fb3d14a-bfc6-485a-b245-e22af307d336-20240630232501.hl7

covid-19-e27662b4-92e9-4280-832c-ae9d110f478f-20240628152506.hl7

covid-19-0c974780-3383-4337-be7a-29571a027549-20240627212515.hl7

It appears that some came in on 6/28, 6/30, 7/1-7/8."

## What you need to know

Communication found in ReportStream Inbox:

" Please also confirm the following settings: 
Quality Filters Default quality filters 
Secondary receiver: gets all reports in CSV format available to download through ReportStream 
MSH-3: CDC PRIME - Atlanta, Georgia (Dekalb)^2.16.840.1.114222.4.1.237821^ISO 
MSH-5: NMDOH^2.16.840.1.113883.3.5364^ISO 
MSH-6: NMDOH^2.16.840.1.113883.3.5364^ISO 
PID-10: If Unknown set to blank 
Suppress HL7 fields: OBX-17-1 
Truncate HD Namespace Ids: True 
**File name format: cdcprime_elr_p_{created date}.hl7** 
Format: HL7 
Transport: SFTP 
SFTP Server: sftp.syncronys.org 
Path: ./incoming/ELR/ 
Batch messages – 12 times per day "

## Acceptance criteria

Verify file name format for NM/Syncronys

## To do

Correct and resend any errors.

chris-kuryak:
@victor-chaparro is looking into it with the Platform team

chris-kuryak:
Platform has a ticket to fix this #15164

chris-kuryak:
@victor-chaparro to check if this is working now

chris-kuryak:
@TaneaY @Jcavallo7 
The issue has been corrected since July 16.

Questions:
1. Are they receiving the correct filenames now?
2. If so, would they like us to resend the messages that had filenames that did not match their settings, or is there no more action needed here?

Jcavallo7:
Follow up email sent to the user.

TaneaY:
Follow up was received ticket resolved.

Hello Juan,

I triple checked and have not seen anymore come in in the last week. I think we a re good now. If any do come in I will let you know.

Thank you for working on this!

Issue #15051, opened 2024-07-03
Reports on the Daily Data page cannot be downloaded


arnejduranovic:
## Describe the bug
<img width="1785" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBL3V2QlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--b6ea75f6218e1b286c6f40f243f367e2783a9324/image%20%283%29.png" alt="image (3).png" />

On the Daily Data page on the ReportStream authenticated website, a user can click "Filename" to download the report, but since PR #14649 this functionality is broken (see screenshot above)

## Impact on ReportStream
Users that do not have transport protocol set up and rely on the website to download their reports cannot

## Steps to reproduce
1. Start up the backend AND frontend locally
2. Submit a report
3. Go the the Daily Data page for the appropriate receiver
4. Try to click the filename and see it cannot be downloaded

## Expected behavior
The Report should be downloaded when "Filename" link is clicked.

## Additional context
The current thought is that [this](https://github.com/CDCgov/prime-reportstream/pull/14649/files#diff-875ca850bfb82db016e5551261d625268dc4[…]rmExternalFilename(,),-val%20content%20%3D) specific change broke this functionality. We should work to find a solution that meets the AC of the original ticket that introduced this change (#)

arnejduranovic:
This bug does not seem to appear for ELIMS receiver

PR #15164, opened 2024-07-11
15051,15157: download report and file name fix

mkalish:
This PR fixes two issues from the updates to how the filename is stored.

Test Steps:
Download issue
1. With any report id generated from the translate, batch or send step call
```
curl --location 'http://localhost:7071/api/history/report/{REPORT_ID}' \
--header 'organization: ignore' \
--header 'Authorization: Bearer dummy'
```
on master, you'll get a 404 and on this branch you'll get the contents

File format
1. Setup a receiver to use one of the custom formats in `file-name-templates`
2. Send a report that will go to that receiver
3. Verify that the file name in the SFTP bucket matches the configured format

## Changes
- Updates getReportId to form the external name if one does not exist on the report
- Updates Report.formExternalFileName to pass in the nameFormat and the translation config

### Testing
- [x] Tested locally?
- [x] Added tests?

### Process

## Linked Issues
- #15051 
- #15157


prime-router/src/main/kotlin/Report.kt
mkalish, line 1: I'm going to try track down why there are so many changes.  I made a small edit in the file but ktlint made a a ton of changes

prime-router/src/main/kotlin/azure/HistoryFunctions.kt
```diff
@@ -386,7 +386,13 @@ open class BaseHistoryFunction : Logging {
                         }
                     )
                     .content(String(contents))
-                    .fileName(requestedReport.externalName)
+                    .fileName(
+                        if (requestedReport.externalName.isNullOrBlank()) {
```
mkalish, line 390: This will "fix" the bug, unclear if it really matters that much what the file name is?

prime-router/src/main/kotlin/Report.kt
```diff
@@ -1511,25 +1511,29 @@ class Report : Logging {
...
+            formExternalFilename(
+                header.reportFile.reportId,
+                header.reportFile.schemaName,
+                MimeFormat.valueOfIgnoreCase(header.reportFile.bodyFormat),
+                header.reportFile.createdAt,
+                metadata = metadata ?: Metadata.getInstance(),
+                translationConfig = header.receiver?.translation,
+                nameFormat = header.receiver?.translation?.nameFormat
+            )
+        }
```
brick-green, line 1536: Do we need the `else if` and `else` blocks at all now? How should the outbound filename depend on the `isSendOriginal` setting?

mkalish: I think the logic could maybe get cleaned up here... but I'm hesitant to make additional adjustments for a bug fix.

I'm overall not sure when we want to return the `externalName`

brick-green (approved):
lgtm

arnejduranovic (approved):

Issue #14602, opened 2024-06-03
Implement Secure Upload Service API


arnejduranovic:
### User Story

As a stakeholder of ReportStream,
I want a SECURE, RELIABLE, SCALABLE, and DECOUPLED way to handle incoming upload requests,
so that we can minimize potential of unrelated-to-uploads ReportStream issues preventing senders from submitting reports.

### Description/Use Case

Presently, ReportStream has two upload APIs:
- "reports API" - uses Azure Functions Key Auth
- "Waters API" - uses OAuth2.0 with homemade custom Authorization server embedded in the application

The reports API is coupled to an insecure auth method and both endpoints are tightly coupled to the main Azure Functions App where all the other services are running alongside. This ticket is to implement a new upload API, specific to UP, that is decoupled from the main app/database and uses the new secure Auth service.

### Dev Notes

The implementation details of this service have (mostly) been thought through. Please see the Receive Step Software Requirements section in the [UP Software Requirements Document](https://cdc.sharepoint.com/:w:/r/teams/ReportStream/Shared%20Documents/Engineering/Design/SRD/Universal%20Pipeline%20SRD.docx?d=w5eff6d937ce34bfa8651e588bdd4ea44&csf=1&web=1&e=IYm2Ol).

### Acceptance Criteria

- [ ] Upload Service (Spring App) implemented per the SRD. Any changes to SRD reviewed and Approved by @arnejduranovic 
- [ ] SRD updated (if needed) and Approved by @arnejduranovic 
- [ ] Auth skipped for now

Andrey-Glazkv:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/14602?planning-poker) with Zenhub @adegolier @arnejduranovic @brick-green @david-navapbc @jack-h-wang @jalbinson @JFisk42 @mkalish @thetaurean

PR #15187, opened 2024-07-15
Spring Submissions API

brick-green:
This PR adds a submission REST api to the submissions module. It will accept a body and three header values. With that information these processes will be included:

1. validate the `Content-Type` and `client_id` headers
2. upload body of request to blob storage
3. upload message to `elr-fhir-convert` queue
4. insert record to submission table in azure storage resource (same resource as the blob storage)
5. create ReportReceivedEvent custom event and send to app insights
6. return Response to submitter

### Unit tests

1. review SubmissionControllerTest.kt
2. run tests

### Process test
1. go get your SU pw and log in to azure portal
2. open pdhteststorageaccount then go to Security + networking --> Access keys --> key 1 connection string
3. `export AZURE_STORAGE_CONNECTION_STRING="<connection string from step 2>"`
4. open pdhtest-appinsights then copy `Connection String` from the top of overview page
5. `export APPLICATIONINSIGHTS_CONNECTION_STRING="<connection string from step 4>"`
6. `./gradlew bootRun`
7. submit to `http://localhost:8880/api/v1/reports` with these headers:
  - Content-Type: application/hl7-v2
  - client_id: test-sender
8. confirm upload to blob to blob storage in reports container
9. confirm record insertion in azure blob table (to view this I had to download the azure storage explorer desktop app
10. confirm azure event added to customEvents in azure app insights

*A new queue(elr-fhir-convert) will be created in subsequent tickets. So that cannot be tested in yet.

## Linked Issues
- #14602


snesm:
Does this conform to prime-router/docs/api/reports.yml, or is there a new spec?

arnejduranovic:
Should we add a new end2end test that submits using new endpoint?

brick-green:
> Does this conform to prime-router/docs/api/reports.yml, or is there a new spec?

This is a new spec. I'll add a doc. Good catch

brick-green:
> Should we add a new end2end test that submits using new endpoint?

Figured that would go along with the convert step update ticket, right?

prime-router/src/main/kotlin/azure/batch/BatchDeciderFunction.kt
mkalish, line 1: These are just formatting changes?

submissions/build.gradle.kts
```diff
@@ -15,16 +15,25 @@ extra["springCloudAzureVersion"] = "5.13.0"
 dependencies {
     implementation("org.springframework.boot:spring-boot-starter-web")
     implementation("com.azure.spring:spring-cloud-azure-starter-storage")
+    implementation("com.microsoft.azure:applicationinsights-runtime-attach:3.5.3")
```
mkalish, line 18: I think is already in the project conventions?

submissions/build.gradle.kts
```diff
@@ -15,16 +15,25 @@ extra["springCloudAzureVersion"] = "5.13.0"
...
     testImplementation("org.jetbrains.kotlin:kotlin-test-junit5")
+    testImplementation("org.mockito.kotlin:mockito-kotlin:4.0.0")
     testRuntimeOnly("org.junit.platform:junit-platform-launcher")
     implementation("org.jetbrains.kotlinx:kotlinx-coroutines-reactor:1.8.1")
     implementation(project(":shared"))
 }
 
+configurations.all {
+    exclude(group = "org.apache.logging.log4j", module = "log4j-to-slf4j")
+    exclude(group = "ch.qos.logback")
```
mkalish, line 34: Maybe just a comment indicating why this needs to get excluded?

submissions/src/main/kotlin/gov/cdc/prime/reportstream/submissions/config/AzureConfig.kt
```diff
@@ -0,0 +1,58 @@
...
+            .buildClient()
+
+    @Bean
+    fun tableClient(): TableClient = TableClientBuilder()
+            .connectionString(connectionString)
+            .tableName(tableName)
+            .buildClient()
+
+    @Bean
+    fun containerName(): String = containerName
```
mkalish, line 46: Doesn't look like these are used?

submissions/src/main/kotlin/gov/cdc/prime/reportstream/submissions/controllers/SubmissionController.kt
```diff
@@ -0,0 +1,185 @@
...
+            blobClient.upload(dataByteArray.inputStream(), dataByteArray.size.toLong())
+            logger.info("Uploaded report to blob storage: blobUrl=${blobClient.blobUrl}")
+
+            // Create the message for the queue
+            val message = mapOf(
+                "reportId" to reportId.toString(),
+                "blobUrl" to blobClient.blobUrl,
+                "headers" to filterHeaders(headers)
+            )
+            val objectMapper = jacksonObjectMapper().registerModule(JavaTimeModule())
```
mkalish, line 91: Maybe as a follow up, but we should just move the jackson utilities to the shared project so we don't need to make one again here

submissions/src/main/kotlin/gov/cdc/prime/reportstream/submissions/controllers/SubmissionController.kt
```diff
@@ -0,0 +1,185 @@
...
+            logger.debug("Converted report data to ByteArray")
+
+            // Upload to blob storage
+            val blobContainerClient = blobServiceClient.getBlobContainerClient(containerName)
+            val blobClient = blobContainerClient.getBlobClient(formBlobName(reportId, headers))
+            blobClient.upload(dataByteArray.inputStream(), dataByteArray.size.toLong())
+            logger.info("Uploaded report to blob storage: blobUrl=${blobClient.blobUrl}")
+
+            // Create the message for the queue
+            val message = mapOf(
```
mkalish, line 86: Can we pull out a data class for this?  We'll want to be able to use that same class in prime-router to deserialize

submissions/src/main/kotlin/gov/cdc/prime/reportstream/submissions/controllers/SubmissionController.kt
```diff
@@ -0,0 +1,185 @@
...
+            logger.debug("Created message for queue")
+
+            // Upload to Queue
+            val queueClient = queueServiceClient.getQueueClient(queueName)
+            queueClient.createIfNotExists()
+            queueClient.sendMessage(messageString)
+            logger.info("Sent message to queue: queueName=$queueName")
+
+            // Insert into Table
+            // TableEntity() sets PartitionKey and RowKey. Both are required by azure and combine to create the PK
```
mkalish, line 102: What do you think about using the reportReceivedTime as the partition key?

submissions/src/main/kotlin/gov/cdc/prime/reportstream/submissions/controllers/SubmissionController.kt
```diff
@@ -0,0 +1,185 @@
...
+                headers = filterHeaders(headers),
+                sender = clientId!!,
+                senderIP = headers["x-azure-clientip"].toString(),
+                fileSize = headers["content-length"].toString(),
+                blobUrl = blobClient.blobUrl
+            )
+            logger.debug("Created ReportReceivedEvent")
+
+            // Log to Application Insights
+            telemetryClient.trackEvent(
```
mkalish, line 128: Maybe as a follow up, but we should move the azure event service over to shared, since all the event stuff should work the same way

submissions/src/main/kotlin/gov/cdc/prime/reportstream/submissions/controllers/SubmissionController.kt
```diff
@@ -0,0 +1,185 @@
...
+            val response =
+                CreationResponse(
+                    reportId,
+                    status,
+                    OffsetDateTime.now(),
+                )
+            logger.info("Report submission successful: reportId=$reportId")
+
+            return ResponseEntity(response, HttpStatus.CREATED)
+        } catch (e: Exception) {
```
mkalish, line 144: I think this could be simplified by using a spring `ExceptionHandler`

https://docs.spring.io/spring-framework/reference/web/webmvc/mvc-controller/ann-exceptionhandler.html

submissions/src/main/kotlin/gov/cdc/prime/reportstream/submissions/controllers/SubmissionController.kt
```diff
@@ -0,0 +1,185 @@
...
+
+    private fun formBlobName(
+        reportId: UUID,
+        headers: Map<String, String>,
+    ): String {
+        val senderName = headers["client_id"]?.lowercase()
+        return when (headers["Content-Type"]?.lowercase()) {
+            "application/hl7-v2" -> "receive/$senderName/$reportId.hl7"
+            "application/fhir+ndjson" -> "receive/$senderName/$reportId.fhir"
+            else -> "receive/$senderName/$reportId"
```
mkalish, line 163: I think we should throw?  Though based on the `consumes` annotation I guess this is  not possible?

submissions/src/main/kotlin/gov/cdc/prime/reportstream/submissions/controllers/SubmissionController.kt
```diff
@@ -0,0 +1,185 @@
...
+                // throw IllegalArgumentException("Unsupported content-type: $contentType")
+        }
+    }
+
+    private fun validateHeaders(clientId: String?): ResponseEntity<String>? {
+
+        if (clientId.isNullOrEmpty()) {
+            logger.warn("Missing required header: client_id")
+            return ResponseEntity.badRequest().body("Missing required header: client_id.")
```
mkalish, line 172: Nit: could also handle this via an exception handler rather than returning null.  Might make the flow a little clearer

submissions/src/test/kotlin/SubmissionControllerTest.kt
```diff
@@ -0,0 +1,290 @@
...
+    private lateinit var objectMapper: ObjectMapper
+
+    private lateinit var blobContainerClient: BlobContainerClient
+    private lateinit var blobClient: BlobClient
+    private lateinit var queueClient: QueueClient
+    private lateinit var sendMessageResult: SendMessageResult
+
+    @BeforeEach
+    fun setUp() {
+        MockitoAnnotations.openMocks(this)
```
mkalish, line 70: Any reason to use mockito over mockk?  Might just be confusing going between the projects

submissions/src/test/kotlin/SubmissionControllerTest.kt
```diff
@@ -0,0 +1,290 @@
...
+                .content(requestBody)
+                .contentType(MediaType.valueOf("application/hl7-v2"))
+                .header("client_id", "testClient")
+                .header("payloadname", "testPayload")
+                .header("x-azure-clientip", "127.0.0.1")
+        )
+            .andExpect(MockMvcResultMatchers.status().isCreated)
+
+        verify(blobClient).upload(any(ByteArrayInputStream::class.java), anyLong())
+        verify(queueClient).sendMessage(anyString())
```
mkalish, line 137: Nit: could we harden these tests?  This doesn't verify that the queue message is actually correct

submissions/src/test/kotlin/SubmissionControllerTest.kt
```diff
@@ -0,0 +1,290 @@
...
+        verify(tableClient).createEntity(any())
```
mkalish, line 138: Same here, we're not verifying that the entity we added is correct

submissions/src/test/kotlin/SubmissionControllerTest.kt
```diff
@@ -0,0 +1,290 @@
...
+            blobContainerClient,
+            blobClient,
+            queueClient
+        )
+    }
+
+
+    @Test
+    fun `submitReport should return CREATED status with content-type hl7-v2`() {
+        val data = mapOf("key" to "value")
```
mkalish, line 121: nit: Maybe bring a small HL7 message in here instead?

submissions/src/test/kotlin/SubmissionControllerTest.kt
```diff
@@ -0,0 +1,290 @@
...
+
+        mockMvc.perform(
+            MockMvcRequestBuilders.post("/api/v1/reports")
+                .content(requestBody)
+                .contentType(MediaType.valueOf("application/fhir+ndjson"))
+                .header("client_id", "testClient")
+                .header("payloadname", "testPayload")
+                .header("x-azure-clientip", "127.0.0.1")
+        )
+            .andExpect(MockMvcResultMatchers.status().isCreated)
```
mkalish, line 158: We should also verify the contents of the response

submissions/src/main/kotlin/gov/cdc/prime/reportstream/submissions/config/AzureConfig.kt
```diff
@@ -0,0 +1,58 @@
...
+    private lateinit var containerName: String
+
+    @Value("\${azure.storage.queue-name}")
+    private lateinit var queueName: String
+
+    @Value("\${azure.storage.table-name}")
+    private lateinit var tableName: String
+
+    @Bean
+    fun blobServiceClient(): BlobServiceClient = BlobServiceClientBuilder()
```
mkalish, line 30: Any reason to make this the `bean` rather than the blobcontainer client?

submissions/src/main/kotlin/gov/cdc/prime/reportstream/submissions/config/AzureConfig.kt
```diff
@@ -0,0 +1,58 @@
...
+            .connectionString(connectionString)
+            .buildClient()
+
+    @Bean
+    fun queueServiceClient(): QueueServiceClient = QueueServiceClientBuilder()
```
mkalish, line 35: Same here, this will only use the one queue so we could make that instead?

mkalish (changes requested):
Great stuff!  Loving seeing this new stuff come together.  Few overall thoughts:
- I think the testing needs to be harden a bit; I also do think at least one test using test containers that checks the blob upload and table upload would be a great way to harden this
- It looks like there are is a lot of reformatting unrelated, could we split that out to keep the number of files smaller?
- I think it might be a little cleaner to have a blobcontainer, queueclient etc. injected into the controller rather than injecting the name.  I would look into using `@Component` and `@Autowired`
  - Related to that, I could see a case for abstracting a layer on top of the azure stuff, but maybe that could be a follow up

prime-router/src/main/kotlin/azure/batch/BatchDeciderFunction.kt
brick-green: undid all the formatting changes. Intellij was auto-formatting so I turned that off too.

submissions/build.gradle.kts
brick-green: This isn't in project conventions, that's  `applicationinsights-core`. But I reviewed and didn't actually need this one so I removed.

submissions/build.gradle.kts
brick-green: added a comment here

submissions/src/main/kotlin/gov/cdc/prime/reportstream/submissions/config/AzureConfig.kt
brick-green: removed all of these in favor of creating the clients in the config class

submissions/src/main/kotlin/gov/cdc/prime/reportstream/submissions/controllers/SubmissionController.kt
brick-green: created the data class

submissions/src/main/kotlin/gov/cdc/prime/reportstream/submissions/controllers/SubmissionController.kt
brick-green: yeah that's a bit of an improvement, added it

submissions/src/main/kotlin/gov/cdc/prime/reportstream/submissions/controllers/SubmissionController.kt
brick-green: added an exception handling class. Good call. It's wayyyyyy better

submissions/src/main/kotlin/gov/cdc/prime/reportstream/submissions/controllers/SubmissionController.kt
brick-green: yeah I like throwing here too. It shouldn't ever happen due to `consumes` attribute but seems clearer this way

submissions/src/main/kotlin/gov/cdc/prime/reportstream/submissions/controllers/SubmissionController.kt
brick-green: another good call. Added to exception handling class

submissions/src/test/kotlin/SubmissionControllerTest.kt
brick-green: I can't remember the specifics now. But I started using mockk and hit some issues then all the guides and online advice for sping were using mockito so it felt more inline with Spring. What do you think?

submissions/src/test/kotlin/SubmissionControllerTest.kt
brick-green: added a listener and some verifys on on the inputs

submissions/src/main/kotlin/gov/cdc/prime/reportstream/submissions/config/AzureConfig.kt
brick-green: moved all of them to the client instead of the service. Definitely cleaner

submissions/src/main/kotlin/gov/cdc/prime/reportstream/submissions/controllers/SubmissionController.kt
```diff
@@ -0,0 +1,228 @@
...
+        logger.debug("Created message for queue")
+
+        // Upload to Queue
+        queueClient.createIfNotExists()
+        queueClient.sendMessage(messageString)
+        logger.info("Sent message to queue: queueName=${queueClient.queueName}")
+
+        // Insert into Table
+        // TableEntity() sets PartitionKey and RowKey. Both are required by azure and combine to create the PK
+        val tableEntity = TableEntity(reportReceivedTime.toString(), reportId.toString())
```
mkalish, line 94: I would maybe switch this to using `Instant` rather than `Offset`.  I think our containers will always run in the east coast data center but if that changed I think the partition would be less effective with different offsets

shared/src/main/kotlin/gov/cdc/prime/reportstream/shared/SubmissionQueueMessage.kt
```diff
@@ -0,0 +1,5 @@
+package gov.cdc.prime.reportstream.shared
+
+import java.util.UUID
+
+data class SubmissionQueueMessage(val reportId: UUID, val blobUrl: String, val headers: Map<String, String>)
```
mkalish, line 5: You're probably in a better place to evaluate this when integrating it into the convert step, but should we special case `clientId` since we absolutely need that to start the next step?

mkalish (commented):
Updates look great!

I did hit two snags when testing locally (I was using the local azurite instance)
- `x-azure-clientip` is required, maybe not a big deal but it might be nice to not have requirement locally
- I got an authentication failure for adding the entity to the table. 

Let me know if a screenshare would help triage these.

mkalish (commented):
Updates look great on the tests.

Still hitting one issue locally
```
java.lang.IllegalStateException: agent was not initialized
        at com.microsoft.applicationinsights.agent.internal.telemetry.TelemetryClient.getActive(TelemetryClient.java:125)
        at com.microsoft.applicationinsights.agent.internal.classicsdk.BytecodeUtilImpl.trackEvent(BytecodeUtilImpl.java:103)
        at com.microsoft.applicationinsights.agent.bootstrap.BytecodeUtil.trackEvent(BytecodeUtil.java:73)
        at com.microsoft.applicationinsights.TelemetryClient.agent$trackEventTelemetry(TelemetryClient.java)
        at com.microsoft.applicationinsights.TelemetryClient.track(TelemetryClient.java)
        at com.microsoft.applicationinsights.TelemetryClient.trackEvent(TelemetryClient.java:78)
        at com.microsoft.applicationinsights.TelemetryClient.trackEvent(TelemetryClient.java:58)
        at gov.cdc.prime.reportstream.submissions.controllers.SubmissionController.submitReport(SubmissionController.kt:121)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
```
here's part of the stack.  I think the issue is when you don't have the connection string set.

You might need abstract out the telemetry service into your own service where you can check for the connection string and maybe just log.

Two other things (maybe just follow up tickets):
- were you able to have any luck with an integration test?
- noticed that we should update the logging config to do JSON like prime-router

submissions/src/main/kotlin/gov/cdc/prime/reportstream/submissions/TelemetryService.kt
```diff
@@ -0,0 +1,30 @@
+package gov.cdc.prime.reportstream.submissions
+
+import com.microsoft.applicationinsights.TelemetryClient
+import org.slf4j.LoggerFactory
+import org.springframework.stereotype.Service
+
+@Service
```
mkalish, line 7: 🙌🏻

mkalish (approved):
Fantastic stuff here!

Was able to successfully submit a report and confirmed I saw in azure storage.

PR #15202, opened 2024-07-16
Add dummy filters to avoid routing all ETOR ORUs to these receivers

basiliskus:
We currently don't have the receiver identifier information for the Ochsner and Oracle recivers. In the meantime, we need to add a dummy filter so all ETOR ORUs are not routed to them


victor-chaparro (approved):
You can edit these settings through the UI. You don't really have to create PRs for these small changes.

Issue #14574, opened 2024-05-31
Update Integration Tests for FHIRTranslator


arnejduranovic:
### User Story

As a developer of the Universal Pipeline,
I want to have integration tests that ensure `FHIRTranslator` updates the system state correctly and produces correct outputs,
so that I can confirm the pipeline step is working properly.

### Description/Use Case

There are some FHIRTranslator integration tests located in `FhirFunctionIntegrationTests.kt`. These should be split out into its own class, following the pattern established by `FHIRRouterIntegrationTests` and `FHIRConverterIntegrationTests`. There may be a lot of test harnessing that can be reused between the three, and this should be broken out into a shared location instead of being duplicated. 

The integration tests should use both the database and azure blob test containers and verify both are updated correctly given certain inputs to the function. The following should be verified:
- report lineage is correctly updated
- item lineage is correctly updated
- task table is correctly updated
- action and action_log table is correctly updated
- report_file table is correctly updated
- Azure Queue Message is correctly generated
- Azure Events are properly generated
- Passthrough functionality works correctly
- Translation functionality works correctly

### Acceptance Criteria

- [ ] Created one or more integration tests for the pipeline step that gives confidence the step is working as expected

Andrey-Glazkv:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/14574?planning-poker) with Zenhub @adegolier @arnejduranovic @brick-green @david-navapbc @jack-h-wang @jalbinson @JFisk42 @mkalish @thetaurean

PR #15304, opened 2024-07-23
14574 fhir translator tests

jack-h-wang:
This PR adds new tests for the FHIR translate step and removes deprecated tests.

Test Steps:
1. Verify `FHIRTranslatorIntegrationTests` complete successfully and cover expected usage scenarios.

## Changes
- Add `FHIRTranslatorIntegrationTests` and update `UniversalPipelineTestUtils`.
- Remove old `FhirFunctionTests` and `FhirFunctionIntegrationTests`.

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- #14574


prime-router/src/test/kotlin/common/UniversalPipelineTestUtils.kt
```diff
@@ -184,10 +184,17 @@ object UniversalPipelineTestUtils {
     val fhirSenderWithNoTransform = UniversalPipelineSender(
         "fhir-elr-no-transform",
         "phd",
-        MimeFormat.HL7,
+        MimeFormat.FHIR,
```
jack-h-wang, line 187: `fhirSenderWithNoTransform` was essentially a near duplicate of `hl7SenderWithNoTransform` with no ill effect across several test suites and I'm a little confused why. This value can be flipped back to HL7 without affecting test results.

thetaurean: I think at this point in the pipeline we've passed all the points where that value would be checked or impact control flow. That makes sense to me since at this point it doesn't really matter what the original format was.

prime-router/src/test/kotlin/fhirengine/azure/FHIRTranslatorIntegrationTests.kt
```diff
@@ -0,0 +1,364 @@
...
+
+            // verify message format is HL7
+            assertThat(batchTask.bodyFormat).isEqualTo("HL7")
+
+            // verify message does not match the original FHIR input
+            val translatedValue = BlobAccess.downloadBlobAsByteArray(
+                sendReportFile!!.bodyUrl,
+                UniversalPipelineTestUtils.getBlobContainerMetadata(azuriteContainer)
+            )
+            assertThat(translatedValue).isNotEqualTo(reportContents.toByteArray())
```
thetaurean, line 195: Is there any way we could ensure enrichments were applied correctly? Perhaps apply them to the parsed original bundle and then assert with that?

thetaurean (changes requested):
Do we also need to differentiate between HL7/FHIR when isSendOriginal is true? I see that there's no branching code there but it might be good to have a test just to make sure format never impacts sending the original.

I see that `FHIRTranslator.kt` doesn't really log or emit events, but maybe we should assert/check any added actionHistory?

What do you think about expanding the assertion on the resulting batch queue message? We could check that it is the correct type and verify its values.
```
assertThat(batchTask)
   .isInstanceOf<BatchQueueMessage>()
   .matchesPredicate
```
We could also verify that the method to queue it got called. We do this a bunch in the filtering integration tests, might be a good place to borrow from.

jack-h-wang:
> Do we also need to differentiate between HL7/FHIR when isSendOriginal is true? I see that there's no branching code there but it might be good to have a test just to make sure format never impacts sending the original.

Additional tests have been added to test both.

> I see that `FHIRTranslator.kt` doesn't really log or emit events, but maybe we should assert/check any added actionHistory?

I've added the check. Also realized in the process I misunderstood how much setup was actually required for the test; the initial version generated too many reports.

> What do you think about expanding the assertion on the resulting batch queue message? We could check that it is the correct type and verify its values.
> 
> ```
> assertThat(batchTask)
>    .isInstanceOf<BatchQueueMessage>()
>    .matchesPredicate
> ```
> 
> We could also verify that the method to queue it got called. We do this a bunch in the filtering integration tests, might be a good place to borrow from.

I think the way the tests currently check specific values in the batch/send messages is sufficient. I don't think we need to check which methods were called for these tests; checking the queue itself should be all that's needed.

thetaurean (approved):
👍

prime-router/src/test/kotlin/fhirengine/azure/FHIRTranslatorIntegrationTests.kt
```diff
@@ -0,0 +1,611 @@
...
+            val translatedValue = BlobAccess.downloadBlobAsByteArray(
+                sendReportFile!!.bodyUrl,
+                UniversalPipelineTestUtils.getBlobContainerMetadata(azuriteContainer)
+            )
+            assertThat(translatedValue).isEqualTo(reportContents.toByteArray())
+        }
+    }
+
+    @Test
+    fun `successfully translate for FHIR receiver when isSendOriginal is true`() {
```
mkalish, line 538: Maybe one more test verifying the sendOriginal behavior when the original is HL7, but the receiver is configured to get FHIR?

prime-router/src/test/kotlin/fhirengine/azure/FHIRTranslatorIntegrationTests.kt
mkalish: Looks like this is unresolved

jack-h-wang: Discussed with Gabriel and added some clarifying comments on what this test is meant to accomplish.

prime-router/src/test/kotlin/fhirengine/azure/FHIRTranslatorIntegrationTests.kt
```diff
@@ -0,0 +1,611 @@
...
+
+        // verify task and report_file tables were updated correctly in the Translate function (new task and new
+        // record file created)
+        ReportStreamTestDatabaseContainer.testDatabaseAccess.transact { txn ->
+            val batchTask = DSL.using(txn).select(Task.TASK.asterisk()).from(Task.TASK)
+                .where(Task.TASK.NEXT_ACTION.eq(TaskAction.batch))
+                .fetchOneInto(Task.TASK)
+
+            // verify batch queue task exists
+            assertThat(batchTask).isNotNull()
```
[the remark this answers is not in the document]
jack-h-wang: The batch task object retrieved here is used as the base for the subsequent tests, which check body format and the message uploaded to the blobstore. I also added a check that the receiver is as expected.

PR #15412, opened 2024-07-31
Refactor condition mapping lookups into service-like class

### LINKED ISSUE #12906: Use service design pattern for code-to-condition lookups
### User Story
As a dev I want blocking data retrieval to be wrapped in a service design pattern
so that data retrieval is reusable, failure methods are well defined, and business logic stays decoupled and isolated.

### Description/Use Case
The lookup of condition codes from the `observation-mapping` table is currently a rudimentary function call and lookup in a data structure. This should be wrapped into a service and use the appropriate design pattern.

### Acceptance Criteria
- `observation-mapping` table data is accessed using a service design pattern

thetaurean:
This PR refactors code to condition lookups into their own class utilizing a service-like design pattern.

Test Steps:
1. Run test suites

## Changes
- Code to condition mapping has been refactored into its own service-like class

### Testing
- [x] Tested locally?

## Linked Issues
- Fixes #12906


david-navapbc (approved):
some things on style but nothing worth holding up the PR over. good work!

prime-router/src/main/kotlin/cli/LookupTableCommands.kt
```diff
@@ -553,50 +554,6 @@ class LookupTableGetCommand(httpClient: HttpClient? = null) : GenericLookupTable
     }
 }
 
-class ObservationMappingConstants {
```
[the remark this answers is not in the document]
thetaurean: I moved them to `ConditionMapper.kt`. I originally put them in `LookupTableCommands.kt` on a whim or lack of a better place.

prime-router/src/main/kotlin/azure/ConditionMapper.kt
```diff
@@ -0,0 +1,138 @@
...
+import org.hl7.fhir.r4.model.Bundle
+import org.hl7.fhir.r4.model.Coding
+import org.hl7.fhir.r4.model.Observation
+
+interface ConditionMapper {
+    /**
+     * Attempt to find diagnostic conditions for a given test [code]
+     * @return a list of diagnostic conditions identified (could be empty)
+     */
+    fun lookupCondition(code: String): List<Coding>
```
mkalish, line 17: Nit: I would consider dropping this?  Might encourage callers to use the bulk impl

prime-router/src/main/kotlin/azure/ConditionMapper.kt
```diff
@@ -0,0 +1,138 @@
...
+
+    /**
+     * Lookup condition codes on every observation in a [bundle] and add them as custom extensions
+     * @param bundle the bundle that will be stamped
+     * @return a list of ObservationMappingFailure objects with information on any mapping failures and their source
+     */
+    fun stampBundle(bundle: Bundle): List<LookupTableConditionMapper.ObservationMappingFailure>
+
+    /**
+     * Lookup condition codes for an [observation] and add them as custom extensions
```
mkalish, line 33:
```suggestion
     * Lookup condition codes for an [Observation] and add them as custom extensions
```

prime-router/src/main/kotlin/azure/ConditionMapper.kt
```diff
@@ -0,0 +1,138 @@
...
+     * @return a map associating test [codes] to their diagnostic conditions as Coding's
+     */
+    fun lookupConditions(codes: List<String>): MutableMap<String, List<Coding>>
+
+    /**
+     * Lookup condition codes on every observation in a [bundle] and add them as custom extensions
+     * @param bundle the bundle that will be stamped
+     * @return a list of ObservationMappingFailure objects with information on any mapping failures and their source
+     */
+    fun stampBundle(bundle: Bundle): List<LookupTableConditionMapper.ObservationMappingFailure>
```
mkalish, line 30: This doesn't seem to fit here?  The interface is returning a class defined under the concrete impl?

prime-router/src/main/kotlin/azure/ConditionMapper.kt
```diff
@@ -0,0 +1,138 @@
...
+
+    /**
+     * Lookup condition codes for an [observation] and add them as custom extensions
+     * @param observation the observation that will be stamped
+     * @return a list of ActionLogDetail objects with information on any mapping failures
+     */
+    fun stampObservation(observation: Observation): List<ActionLogDetail>
+
+    companion object {
+        const val TEST_CODE_KEY = "Code"
```
mkalish, line 40: This also seems a little unexpected here since these are tied directly to the lookup table right?

prime-router/src/main/kotlin/azure/ConditionMapper.kt
```diff
@@ -0,0 +1,138 @@
...
+     * @return a list of ObservationMappingFailure objects with information on any mapping failures and their source
+     */
+    fun stampBundle(bundle: Bundle): List<LookupTableConditionMapper.ObservationMappingFailure>
+
+    /**
+     * Lookup condition codes for an [observation] and add them as custom extensions
+     * @param observation the observation that will be stamped
+     * @return a list of ActionLogDetail objects with information on any mapping failures
+     */
+    fun stampObservation(observation: Observation): List<ActionLogDetail>
```
mkalish, line 37: Might make more sense for this to accept the observation and codes to add as an extension?  Feels like making the interface return an ActionLogDetail is limiting.

mkalish (changes requested):
I like where this is heading!  My thoughts are mainly around more opportunities to further isolate the new service.

prime-router/src/main/kotlin/azure/ConditionMapper.kt
thetaurean: I think this should match the parameter name (which is lower case)

prime-router/src/main/kotlin/azure/ConditionMapper.kt
thetaurean: Doh.

prime-router/src/main/kotlin/azure/ConditionMapper.kt
thetaurean: DId a 2nd pass at reorganizing these

prime-router/src/main/kotlin/azure/ConditionMapper.kt
thetaurean: On second thought this probably doesn't belong in this interface whatsoever. More to come.

thetaurean:
Thanks for the great feedback! I took a 2nd pass at the class hierarchy, the root purpose of the interface, and the return types. Things should be a bit more reusable and forward-minded now, I think.

Looking forward to see your thoughts.

mkalish (approved):
LGTM!

Issue #15397, opened 2024-07-30
Correct valueset lookup table loading


jack-h-wang:
### User Story
As a RS engineer, I want to ensure lookup tables are loaded properly so that transforms function as expected.

### Description/Use Case
There seems to be a bug that prevents a single valueset lookup table to be loaded more than once. See #14901 for details.

### Dev Notes
PR #14933 was merged as an interim solution. This work is to remove the interim solution and load the lookup tables as expected.

### Acceptance Criteria
- [ ] Revert #14933 and make changes to the valueset lookup function in a way that tests continue to pass.

Issue #14901, opened 2024-06-26
Update lookupTable function to allow to lookup more than one column of a Table


oslynn:
## Problem statement

lookTable only allows to lookup one column. If we try to lookup another column, it throws "Value column not found in specified lookup table" at LookupTableValueSet.kt: 38.  During the first lookup, it filters the table to specified key:valueColumn at line 43.  

## Impact
- Example: Data from SimpleReport has the LOINC code but does not include OBR-4.7 Coding System Version ID. This value is found in the LIVD table. When we go to look up the OBR-4.7, we cannot because we are only limited to looking up one column of the LIVD table at a time.

## How to reproduce
To duplicate the problem at the sr-service-request-code-extension-version element to simple-report-sender-transform.yml after element: sr-service-request-code-extension-display as below:

  - name: sr-service-request-code-extension-display
    condition: '%serviceRequest.code.coding[0].display.empty()'
    bundleProperty: '%serviceRequest.code.coding[0].display'
    value: [ '%serviceRequest.code.coding.code' ]
    valueSet:
      lookupTable:
        tableName: LOINC
        keyColumn: LOINC_NUM
        valueColumn: LONG_COMMON_NAME

  - name: sr-service-request-code-extension-version
    condition: '%serviceRequest.code.coding[0].version.empty()'
    bundleProperty: '%serviceRequest.code.coding[0].version'
    value: [ '%serviceRequest.code.coding.code' ]
    valueSet:
      lookupTable:
        tableName: LOINC
        keyColumn: LOINC_NUM
        valueColumn: VersionLastChanged

./gradlew package

Use Postman to send
[SR-TEST.txt](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMG1RQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--d7e85a4117a1b9f2235a513e7e8289c646855357/SR-TEST.txt)

You should see the exception below:
<img width="1346" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMGFRQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--7f84064c51e945282c72a0b62792c3108e577711/image.png" alt="image.png" />

## Needed behavior
[...]

PR #15425, opened 2024-08-01
15397 valueset loading fix

jack-h-wang:
This PR reverts the previously merged fix to an issue with multiple valuesets using the same source table and implements a new fix.

Test Steps:
1. Covered by tests, but another reproduction of the issue is detailed in #14901

## Changes
- Reverted #14933
- A copy of the lookup table is now made instead of processing the lookup table in metadata directly
- Updated tests

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- #15397

### DIFF
--- a/prime-router/src/main/kotlin/fhirengine/engine/LookupTableValueSet.kt
+++ b/prime-router/src/main/kotlin/fhirengine/engine/LookupTableValueSet.kt
@@ -40,7 +40,7 @@ class LookupTableValueSet
             throw SchemaException("Value column not found in specified lookup table")
         }
 
-        val filterTable = lookupTable.table.copy().retainColumns(configData.keyColumn, configData.valueColumn)
+        val filterTable = lookupTable.table.selectColumns(configData.keyColumn, configData.valueColumn)
         val result = mutableMapOf<String, String>()
         filterTable.forEach { row ->
             result[row.getString(configData.keyColumn)] = row.getString(configData.valueColumn)
--- a/prime-router/src/main/kotlin/fhirengine/engine/LookupTableValueSet.kt
+++ b/prime-router/src/main/kotlin/fhirengine/engine/LookupTableValueSet.kt
@@ -40,8 +40,9 @@ class LookupTableValueSet
             throw SchemaException("Value column not found in specified lookup table")
         }
 
+        val filterTable = lookupTable.table.retainColumns(configData.keyColumn, configData.valueColumn)
         val result = mutableMapOf<String, String>()
-        lookupTable.table.forEach { row ->
+        filterTable.forEach { row ->
             result[row.getString(configData.keyColumn)] = row.getString(configData.valueColumn)
         }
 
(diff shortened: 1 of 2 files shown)


prime-router/src/main/kotlin/fhirengine/engine/LookupTableValueSet.kt
```diff
@@ -40,8 +40,9 @@ class LookupTableValueSet
             throw SchemaException("Value column not found in specified lookup table")
         }
 
+        val filterTable = lookupTable.table.copy().retainColumns(configData.keyColumn, configData.valueColumn)
```
jack-h-wang, line 43: Originally, `retainColumns` was called directly on `lookupTable.table`, which was modifying the copy of the lookup table in metadata. Explicitly making a copy of the original table to work from instead resolves the issue.

prime-router/src/test/kotlin/fhirengine/engine/LookupTableValueSetTests.kt
```diff
@@ -77,88 +78,61 @@ class LookupTableValueSetTests {
         mockkObject(Metadata)
         every { Metadata.getInstance() } returns UnitTestUtils.simpleMetadata
 
-        var bundle = Bundle()
+        val bundle = Bundle()
```
jack-h-wang, line 81: Updated this test to more closely resemble the conditions under which the issue was discovered and to remove mutability.

prime-router/src/main/kotlin/fhirengine/engine/LookupTableValueSet.kt
```diff
@@ -40,8 +40,9 @@ class LookupTableValueSet
             throw SchemaException("Value column not found in specified lookup table")
         }
 
+        val filterTable = lookupTable.table.selectColumns(configData.keyColumn, configData.valueColumn)
```
thetaurean, line 43:
```suggestion
        val filterTable = lookupTable.table.selectColumns(configData.keyColumn, configData.valueColumn)
```

This might avoid an explicit copy (already addressed)

prime-router/src/test/kotlin/fhirengine/engine/LookupTableValueSetTests.kt
thetaurean: Nice catch

thetaurean (approved):
LGTM  👍

Requested 1 small change which you already addressed :).

prime-router/src/main/kotlin/fhirengine/engine/LookupTableValueSet.kt
jack-h-wang, line 43: Originally, `retainColumns` was called directly on `lookupTable.table`, which was modifying the copy of the lookup table in metadata. Using `selectColumns` is more appropriate for filtering the table.

arnejduranovic (approved):

PR #14696, opened 2024-06-10
MN UP Receiver Migration Pre-Work

oslynn:
This PR MN UP Receiver migration pre-work

Test Steps:
1. *Include steps to test these changes*

## Changes
- *Include a comprehensive list of changes in this PR*
- *(For web UI changes) Include screenshots/video of changes*

### Testing
- [ ] Tested locally?
- [ ] Ran `./prime test` or `./gradlew testSmoke` against local Docker ReportStream container?
- [ ] (For Changes to /frontend-react/...) Ran `npm run lint:write`? 
- [ ] Added tests?

### Process
- [ ] Are there licensing issues with any new dependencies introduced?
- [ ] Includes a summary of what a code reviewer should test/verify?
- [ ] Updated the release notes?
- [ ] Database changes are submitted as a separate PR?
- [ ] DevOps team has been notified if PR requires ops support?

## Linked Issues
- Fixes #issue

## To Be Done
*Create GitHub issues to track the work remaining, if any*
- #issue

## Specific Security-related subjects a reviewer should pay specific attention to
- Does this PR introduce new endpoints?
    - new endpoint A
    - new endpoint B
- Does this PR include changes in authentication and/or authorization of existing endpoints?
- Does this change introduce new dependencies that need vetting?
- Does this change require changes to our infrastructure?
- Does logging contain sensitive data?
- Does this PR include or remove any sensitive information itself?

If you answered '_yes_' to any of the questions above, conduct a detailed Review that addresses at least:

- What are the potential security threats and mitigations? Please list the _STRIDE_ threats and how they are mitigated
    - **S**poofing (faking authenticity)
        - Threat _T_, which could be achieved by _A_, is mitigated by _M_
    - **T**ampering (influence or sabotage the integrity of information, data, or system)
    - **R**epudiation (the ability to dispute the origin or originator of an action)
    - **I**nformation disclosure (data made available to entities who should not have it)
    - **D**enial of service (make a resource unavailable)
    - **E**levation of Privilege (reduce restrictions that apply or gain privileges one should not have)
- Have you ensured logging does not contain sensitive data?
- Have you received any additional approvals needed for this change?

PR #15528, opened 2024-08-06
Added ability to send reports using PUT

victor-chaparro:
This PR adds the ability to send reports using PUT through the REST transport and also sets the accessToken if provided in the credential.
Minnesota's REST API only accepts reports be sent to them using PUT. The `transport.headers.method` setting must be set to `PUT` so that the REST transport can configure the HTTP client to send via PUT, otherwise it will use POST by default.

Test Steps:
1. Run `./prime multiple-settings set --input ./settings/STLTs/MN/mn-doh.yml`
2. Send test report routed to MN
3. Verify the report gets sent to MN
4. Query the `action` table by running `select * from action order by created_at desc`. Wait for an `action_name` with `send_error` entry to show up. The report fails because we're sending to a mock API that only validates the message and our messages fail that validation. The errors will be fixed, but for now this is only to test we can send reports to their REST API. We can verify that by looking at the `action_result` and making sure we get validation errors back.

## Changes
- Added ability to specify REST method inside transport.headers

### Testing
- [x] Tested locally?
- [ ] Added tests?

## Linked Issues
- Fixes #14696


JessicaWNava (approved):
Couple of minor changes. Mostly I need to know how to verify that the message sent to them. I didn't see any errors when I sent it, but is there another way to verify?

JessicaWNava (changes requested):
Oops meant to request changes

JessicaWNava (approved):
Changes good, just waiting on instructions to figure out how to verify that the message actually sent.

JessicaWNava:
When I tested this I did not see a send_error action in the table. Seems to be stuck on waiting to deliver despite the numberPerDay being set to 1440

prime-router/src/main/kotlin/transport/RESTTransport.kt
```diff
@@ -592,84 +595,106 @@ class RESTTransport(private val httpClient: HttpClient? = null) : ITransport {
...
-                        if (key == "Content-Length" && value == "<calculated when request is sent>") {
-                            Pair(key, message.size.toString())
-                        } else {
-                            Pair(key, value)
-                        }
-                    }.toMap()
-                )
+            val response: HttpResponse = if (headers["method"] == "PUT") {
+                client.put(restUrl) {
+                    build(logger, headers, message, fileName, boundary, reportCreateDate)
```
GilmoreA6, line 601: can we add a test to cover this logic split here?

GilmoreA6 (approved):

PR #15533, opened 2024-08-07
Fix item graph recursive query and re-enable event sending

mkalish:
This PR adjusts the item lineage recursive query to get postgres to use the index after the table has been vacuumed and analyzed

Test Steps:
CI covers it

## Changes
- Re-enables sending events 
- Updates the item lineage recursive query to that postgres will optimize it correctly
### Testing
- [x] Tested locally?
- [ ] Added tests?

### Process


jalbinson (approved):

arnejduranovic (approved):
Context:

This query was really slow and after looking at the execution plan we learned it was doing a sequence scan and not utilizing the index on item_lineage table. Once we removed what we determined to be the unnecessary additional REPORT_FILE query, the execution plan changed to using an index with the actual query completing in metabase with sample data in 6ms whereas before it wasn't.

PR #15547, opened 2024-08-08
fix circular dependency

etanb:
This PR fixes an issue with our e2e tests where we had a circular dependency of `helpers/utils.ts` -> `pages/daily-data.ts` -> `pages/BasePage.ts` and around again. Image below generated with `madge`:

graph


penny-lischer (approved):
lgtm

PR #15572, opened 2024-08-09
Feature/1472 metabase maintenance

supriyaaddagada:
This PR is to get inactive users from Metabase and send an alert to slack about inactive users.


.github/scripts/MetabaseUsers/MBUserAPI.ps1
```diff
@@ -0,0 +1,58 @@
+# param (
+#     [string]$key
+# )
+$key = 'mb_h7RFQv6J2zsrmlDkyE08aNWB/YPKirsK1OPO4DqCBNQ='
```
dkrylovsb, line 4: Key should come from GitHub secrets.

dkrylovsb (approved):

PR #15613, opened 2024-08-14
Platform/13199/fhir function dynamic message routing REDUX 

david-navapbc:
This PR ...

Supersedes the [original PR](https://github.com/CDCgov/prime-reportstream/pull/15273) as it became stale. 

Refactors how queue message routing happens such that the behavior (1) no longer hard coded (2) baked into the queue message itself. ~~The trigger for changing default routing behavior is expected to be the something in the content of the message itself and thus it makes sense for the routing logic to be colocated with the content of the message itself.~~ 

## Changes
~~QueueMessage has been updated with a "when" clause (switch/case for Java-heads) that sets default routing behavior for a given queue message type. Should a short circuit of this behavior be required the germane "when" block for that message type may be updated to route the message to a different queue topic. the queueAccess object is provided to the a new function, QueueMessage.send() to keep separation of concern in tact.  The QueueMessage object handles routing itself with routing plumbing provided by FHIRFunctions.~~

`QueueMessage`, `FhirFunctions`, and associated tests have been updated. 

`QueueMessage` was changed to attach the correct queue name to the QueueMessage obj typed for that kind of QueueMessage. Functionally speaking, the queue message itself knows where it's supposed to go. 

In `FhirFunctions` the calls to `doRoute(), doConvert(), etc` have been consolidated into a single common method. The test updates were made to accommodate this change. 

### Testing
- [x ] Tested locally?


arnejduranovic (approved):
Good work!

PR #15646, opened 2024-08-19
Engagement/jgilmore/15635 update sr transform patient

GilmoreA6:
This PR updates the SR transform to use the accession number as patient-id if one is not received.

Test Steps:
1. ./gradlew testIntegration
2. ./prime test

## Changes
- adds new element to use accession number as patient-id if one is not received and associated integration tests


victor-chaparro (approved):
Nice!

oslynn (approved):

PR #15658, opened 2024-08-20
Platform/david navapbc/13925/singleton http client/redux

david-navapbc:
This PR ...
keeps every change from the previous PR except for the Apache5 client. see ticket for details as to why. this version of code confirmed to pass all tests against stage env. 

### Testing
- [x] Tested locally?


prime-router/src/main/kotlin/cli/tests/AuthTests.kt
```diff
@@ -1172,107 +1170,84 @@ class Server2ServerAuthTests : CoolTest() {
...
-                    requestTimeoutMillis = 45000
-                    // default timeout is 15s; raising higher due to slow Function startup issues
-                }
-                accept(ContentType.Application.Json)
-            }
-        }
+        val response = HttpClientUtils.get(
+            url = orgEndpoint,
+            accessToken = userToken,
+            timeout = 45000, // default timeout is 15s; raising higher due to slow Function startup issues
```
jalbinson, line 1178: Looks like this was there previously but 45 seconds is a very long timeout especially when it comes to tests

prime-router/src/main/kotlin/common/HttpClientUtils.kt
```diff
@@ -32,10 +30,30 @@ class HttpClientUtils {
...
+                            ignoreUnknownKeys = true
+                        }
+                    )
+                }
+                install(HttpTimeout)
+                engine {
+                    followRedirects = true
+                    socketTimeout = TIMEOUT
+                    connectTimeout = TIMEOUT
+                    connectionRequestTimeout = TIMEOUT
```
jalbinson, line 49: Any use in making these values different or is `TIMEOUT` just the max length for any kind of timeout?

prime-router/src/main/kotlin/common/HttpClientUtils.kt
```diff
@@ -32,10 +30,30 @@ class HttpClientUtils {
...
+                }
+            }
+
         /**
          * GET (query resource) operation to the given endpoint resource [url]
          * @param url: required, the url to the resource endpoint
-         * @param tokens: null default, the access token needed to call the endpoint
+         * @param accessToken: null default, the access token needed to call the endpoint
```
jalbinson, line 56: much more clear variable name 👍

prime-router/src/main/kotlin/common/HttpClientUtils.kt
```diff
@@ -312,7 +330,7 @@ class HttpClientUtils {
             httpClient: HttpClient? = null,
         ): HttpResponse {
             return runBlocking {
-                (httpClient ?: createDefaultHttpClient(accessToken)).submitForm(
+                (httpClient ?: getDefaultHttpClient()).submitForm(
```
jalbinson, line 333: Is this pattern still required given we have a singleton http client now? Or are there still some specialized clients sprinkled around the codebase?

prime-router/src/main/kotlin/common/HttpClientUtils.kt
```diff
@@ -496,24 +518,28 @@ class HttpClientUtils {
...
                     headers?.let {
                         headers {
                             headers.forEach {
                                 append(it.key, it.value)
                             }
                         }
                     }
+                    accessToken?.let {
+                        headers {
+                            append("Authorization", "Bearer $accessToken")
```
jalbinson, line 540: Could there be another type of auth needed in the future outside of `Bearer`?

prime-router/src/main/kotlin/cli/tests/AuthTests.kt
david-navapbc: I don't have context for the comment 

```
// default timeout is 15s; raising higher due to slow Function startup issues
```

but I imagine it's there for a reason. My gutcheck feeling is to leave the value alone but if you feel strongly I'm happy to change it to something else.

prime-router/src/main/kotlin/common/HttpClientUtils.kt
david-navapbc: I left it set to what it was set prior. If there's something you'd like me to set one or all of these too I'm open to suggestion

prime-router/src/main/kotlin/common/HttpClientUtils.kt
david-navapbc: I don't think there are callers supplying their own clients - but - I'm loathe to make too many changes until we dig in and refactor all of this (including the sister class HttpUtils) into one, unified, set of http utilities.

prime-router/src/main/kotlin/common/HttpClientUtils.kt
david-navapbc: not in the current design no. 

@arnejduranovic mentioned something about - going forward - each microservice is going to have it's own client for managing this stuff. I imagine in that case if client [x] is doing auth differently than [y] we could make those changes locally to each microservice.

prime-router/src/main/kotlin/common/HttpClientUtils.kt
jalbinson: ok in that case I'd suggest just keeping that in mind when refactoring.

jalbinson (approved):

arnejduranovic (approved):

PR #15675, opened 2024-08-23
Engagement/jessica/15406 cli download report

JessicaWNava:
This PR adds a CLI command to download a report from a specified environment while optionally (unless it is prod) removing PII.

Test Steps:
1. Get a report id to query for, make sure it is returned with and without the PII removed as you specify in the command
2. Get a report id that is for an HL7 report and make sure it returns a message saying it is the wrong type. 
3. Try to download a prod report with remove-pii set to false and make sure it does not allow this.
Example query: downloadMessage -o /Users/jessicawillis/git/newnew/prime-reportstream/prime-router/src/testIntegration/resources/datatests/FHIR_to_HL7/downloaded.fhir --env local -r 777fc1d7-dd51-4f01-ad56-6e51b4721caa --remove-pii false

## Changes
- CLI command added to download a report

### Testing
- [ ] Tested locally?
- [ ] Added tests?

### Process

## Linked Issues
- Fixes #issue

## To Be Done
- #issue

    - new endpoint A
    - new endpoint B


prime-router/src/main/kotlin/azure/ReportFunction.kt
```diff
@@ -85,6 +91,73 @@ class ReportFunction(
...
+     * GET report to download
+     *
+     * @see ../../../docs/api/reports.yml
+     */
+    @FunctionName("getMessagesFromTestBank")
+    fun downloadReport(
+        @HttpTrigger(
+            name = "downloadReport",
+            methods = [HttpMethod.GET],
+            authLevel = AuthorizationLevel.ANONYMOUS,
```
victor-chaparro, line 104: I think it would be better to use an azure function key to call this API, since we will have to manually call this API every time we need a report and we won't be using the front end to call this API. It will make it easier to use and we wouldn't have to get a bearer token every time we wanted to use it.
There's an example inside `ReportFunction.kt` in line 70. You'd just have to set to authLevel to `authLevel = AuthorizationLevel.FUNCTION` and once it's in staging and prod we can generate a random key for it.
If we do it that way we can remove the authentication logic and just rely on azure's authorization.

victor-chaparro (approved):

PR #15793, opened 2024-09-05
testing re enabling file download

### LINKED ISSUE #15612: Add File Download testing to Daily Data smoke test user flow
## Problem statement
Checking for File Download capability was removed from the previous ticket #15211. So we need to add that capability back in.

## What you need to know

 We need to select a Covid pipeline receiver in order to ensure that the Report details lists facility data.
 As an admin, select the ignore org and navigate to the Daily Data page.

- UP receiver: ignore.FULL_ELR
- CP receiver: ignore.CSV
- ELIMS receiver: ignore.ELR_ELIMS
 
## Acceptance criteria

For a Universal pipeline receiver:
- [ ] First record downloads when filename clicked
- [ ] Downloaded filename matches filename shown in Filename column

etanb:
The underlying issue is that we can't test downloading of a REAL file when the data is mocked. However, when we're pointing to staging, IE in the Github Action Smoke Tests, the logic in the test is sound, and we can test against a real file download.

Fixes #15612


penny-lischer (approved):
lgtm

PR #15814, opened 2024-09-08
Platform/david navapbc/14568/update fhir resource delete strategy

david-navapbc:
This PR ...
Updates the logic in `FHIRReceiverFilter.evaluateObservationConditionFilters()` to be functionally equivalent but more succinct and more computationally efficient. As per the scope of the ticket I did not address upstream refactoring (ie - how condition filtering works) and nor did I change any of the methods in FHIRBundleHelpers. 

No new methods added so no new tests. Existing tests cover changes in this PR: 
```

FHIRBundleHelperTests
--
"test filterObservations"
"test filterMappedObservations"

FhirReceiverFilterTests
---
"test filterMappedObservations"
"fail - mapped condition filter fails"
"fail - bundle of only AOEs do not pass mappedConditionFilter"
"success - jurisfilter, qualfilter, routing filter, proc mode passes, and condition filter passes"
"test the bundle queued for the translate function is filtered to conditions the receiver wants"
"test the bundle queued for the translate function is filtered to mapped conditions the receiver wants"
"test a receiver can receive a report when no condition filters are configured"

FhirReceiverFilterIntegrationTests
---
"should send valid FHIR report filtered by condition filter"
"should not send report fully pruned by condition filter"
"should send valid FHIR report with no condition related filtering"
"should not send report where the only unpruned observations are AOE"
"should send valid FHIR report filtered by mapped condition filter"
"should not send report fully pruned by mapped condition filter"
```

### Testing
- [x ] Tested locally?

### DIFF
--- a/prime-router/src/main/kotlin/fhirengine/engine/FHIRReceiverFilter.kt
+++ b/prime-router/src/main/kotlin/fhirengine/engine/FHIRReceiverFilter.kt
@@ -29,7 +29,6 @@ import gov.cdc.prime.router.azure.observability.event.AzureEventService
 import gov.cdc.prime.router.azure.observability.event.AzureEventServiceImpl
 import gov.cdc.prime.router.azure.observability.event.ReportStreamEventName
 import gov.cdc.prime.router.azure.observability.event.ReportStreamEventProperties
-import gov.cdc.prime.router.codes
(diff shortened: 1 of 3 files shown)


arnejduranovic:
No longer needed

Issue #15645, opened 2024-08-19
[BUG] Submission History API doesn't show Filtered reports


victor-chaparro:
## Describe the bug
When filters are triggered for receivers they do not show up when calling Submission History API `{{url}}/waters/report/{reportId}/history`. The submission status says `Waiting to Deliver` and should say `Failed to Deliver` and under destinations filteredReports should be populated.

## Impact on ReportStream

Senders, if using the history endpoint, may not get accurate information as to the real status of their submitted report and may lead to confusion.

Engagement team uses API to troubleshoot submission and ensure filters are working as expected.

## Steps to reproduce
Send a report locally and make sure the submission triggers a qualityFilter. You can do this by updating a receiver's quality filter to false:`qualityFilter: ["false"]`
Call the Submission History API  `{{url}}/waters/report/2558/history`

It seems like the new steps added to the Universal Pipeline are breaking this API. The API was probably looking at the Action Log Route step records to return this information.

## Expected behavior
- "Failed to deliver" should be the status that is displayed when a report fails quality filters.
- Destinations object should return filters triggered.

arnejduranovic:
@victor-chaparro when a report fails quality filters (or any routing filter) should we use status "not delivering / partially delivered" instead of "failed to deliver". I don't think failed to deliver is a status?

Issue #15524, opened 2024-08-06
[BUG]: History endpoint intermediate overallStatus can be "Not Delivering" and then change later


arnejduranovic:
## Describe the bug
`end2end_up` tests were failing during deployment when run against the staging environment because the history endpoint that the tests rely on were not showing that the tests were delivered where we expected them to be. In fact, in one case, the history endpoint showed an overall status of "Not Delivering" when a query of the customEvents showed the reports was correctly delivered. Manually hitting the history endpoint later showed expected results: overallStatus was "delivered" and the receivers were correct. 

TLDR; It seems like "not delivering" can sometimes be an "intermediary" result whereas it should be a final result. Meaning if a report is marked as "not delivering", then its status should not change at some later point in time.

## Impact on ReportStream
Senders, if using the history endpoint, may not get accurate information as to the real status of their submitted report and may lead to confusion.

## Steps to reproduce
Hard to reproduce since it relies on the processing speed of the report. It may be easier to comb through overallStatus calculation logic and identify the defect that way.

## Expected behavior
- "Not Delivering" should not be an intermediary status, only a final one.
- Create test that reproduces this issue

adegolier:
I've run into this and it's easier to reproduce running against Staging.... Made a ticket (#15654), but closed as duplicate

PR #15846, opened 2024-09-11
Track filter results for reports failing receiver filters

thetaurean:
This PR adds code to properly track filter results so they can be found by the submission history API. This is a partial fix for #15645 , with a follow-up PR to come that fixes the logic for `overallStatus` calculation (and logic for related bug #15524).

Test Steps:
1. Run tests
2. Run the pipeline and submit a report that will fail a ReceiverFilter (I duplicated ignore.FULL_ELR and set arbitrary filters of `true` and `false` accordingly)

## Changes
- Add filter results to generated empty report on filter failure

### Testing
- [x] Tested locally?
- [ ] Added tests?

## To Be Done
- #issue Ticket for adding end2end integration tests for submission history (and fixing bugs mentioned above)


david-navapbc (approved):
LGTM - nice work!

prime-router/src/main/kotlin/Report.kt
```diff
@@ -134,8 +134,8 @@ data class ReportStreamFilterResult(
     val filteredTrackingElement: String,
     val filterType: ReportStreamFilterType?,
     val filteredObservationDetails: String? = null,
+    override val scope: ActionLogScope = ActionLogScope.translation,
```
arnejduranovic, line 137: Can you explain why this is set to translation? I'm not quite following the reason for this, and it seems wrong? We're not translating anything here....

Edit: Okay I see it was like this before, all you did was move it up to the constructor so in FHIRReceiverFIlter you can pass in `ActionLogScope.report`, which makes more sense in that context.

prime-router/src/main/kotlin/fhirengine/engine/FHIRReceiverFilter.kt
```diff
@@ -434,6 +438,19 @@ class FHIRReceiverFilter(
...
+                    // add filter results
+                    emptyReport.filteringResults.add(
+                        ReportStreamFilterResult(
+                            receiver.fullName,
+                            db.fetchReportFile(queueMessage.reportId).itemCount,
+                            filterResult.failingFilter.filters.joinToString("\n"),
+                            emptyList(),
+                            AzureEventUtils.getIdentifier(bundle).value ?: "",
+                            filterResult.failingFilter.filterType,
+                            scope = ActionLogScope.report
```
arnejduranovic, line 450: What's the difference if we put `item` here instead of `report`?

arnejduranovic (approved):

prime-router/src/main/kotlin/fhirengine/engine/FHIRReceiverFilter.kt
thetaurean: I just thought report makes more sense because the whole report is getting filtered out. No issues or side effects if changed to item.

There's a separate ActionLog ReceiverItemFiltered (per item filter3d) that is scoped to item that we should maybe get rid of in favor of a summertime ReportStreamFilterResult.

PR #15893, opened 2024-09-16
Fix submission history bugs and add integration test cases

### LINKED ISSUE #15524: [BUG]: History endpoint intermediate overallStatus can be "Not Delivering" and then change later

### LINKED ISSUE #15645: [BUG] Submission History API doesn't show Filtered reports

thetaurean:
This PR fixes two bugs related to submission history and adds relevant tests to prevent reoccurrence.

Test Steps:
1. Run test suites

## Changes
- Submission history checks `nextAction` on latest report in lineage for reports without destinations
- Submission history checks latest action instead of earliest

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- Fixes #15645 
- Fixes #15524


prime-router/src/test/kotlin/history/SubmissionHistoryIntegrationTests.kt
brick-green, line 1: You've got a happy path test. Should there be an unhappy test?

prime-router/src/main/kotlin/history/SubmissionHistory.kt
```diff
@@ -384,7 +398,7 @@ class DetailedSubmissionHistory(
              * the receivers.
              */
             return if (
-                reports.size > 1
+                reports.size > 1 && !hasNextAction
```
brick-green, line 401: Do you know why this is checking if `reports.size is > 1` instead of `reports.size is > 0`?

brick-green (commented):
a couple questions

prime-router/src/main/kotlin/history/SubmissionHistory.kt
```diff
@@ -182,6 +182,12 @@ class DetailedSubmissionHistory(
     @JsonIgnore
     var nextActionScheduled = false
 
+    /**
+     * Flag to check if there's a next action for a report related to this submission
+     */
+    @JsonIgnore
+    var hasNextAction = false
```
mkalish, line 189: I totally forget how the `nextActionScheduled` is used, but maybe they can get collapsed?

prime-router/src/test/kotlin/history/SubmissionHistoryIntegrationTests.kt
mkalish, line 1: Curious if there is a way we can avoid this test since its pretty heavy.  Is there a reason we can't do this by just seeding the db directly like the other tests do?

prime-router/src/test/kotlin/history/azure/SubmissionFunctionIntegrationTests.kt
```diff
@@ -237,6 +240,9 @@ class SubmissionFunctionIntegrationTests {
                     log(ActionLog(InvalidParamMessage("log"), type = ActionLogLevel.warning))
                     reportGraphNode {
                         action(TaskAction.route)
+                        reportGraphNode {
```
mkalish, line 243: Should there be another assertion change?

prime-router/src/main/kotlin/Receiver.kt
```diff
@@ -115,7 +116,8 @@ open class Receiver(
         timeZone = timeZone,
         dateTimeFormat = dateTimeFormat,
         reverseTheQualityFilter = reverseTheQualityFilter,
-        enrichmentSchemaNames = enrichmentSchemaNames
+        enrichmentSchemaNames = enrichmentSchemaNames,
+        transport = transport
```
mkalish, line 120: Is this used?

mkalish (commented):
Main thought is if we can add more tests to the existing tests file rather than introducing a new pattern.

prime-router/src/main/kotlin/Receiver.kt
thetaurean: Yes used for setting the transport on a receiver to `NullTransport` so I can run the send step.

If I change the tests to not run the whole pipeline it will no longer be needed.

prime-router/src/main/kotlin/fhirengine/engine/FHIRConverter.kt
```diff
@@ -165,7 +165,7 @@ class FHIRConverter(
                                 ),
                                 metadata = this.metadata,
                                 topic = queueMessage.topic,
-                                nextAction = TaskAction.none
+                                nextAction = TaskAction.destination_filter
```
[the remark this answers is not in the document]
thetaurean: Good catch this was left in by mistake; removing.

prime-router/src/main/kotlin/history/SubmissionHistory.kt
thetaurean: I believe the assumption is that if we only have one report we must have only received it so far.

prime-router/src/test/kotlin/history/SubmissionHistoryIntegrationTests.kt
thetaurean: The intent was to create tests that couple the SubmissionHistory to the pipeline steps.

Thinking it over I could have added simpler tests to ensure the pipeline steps output the expected state (with respect to ActionHistory, etc.), then add corresponding tests to the Submission side with that expected state.

Working on this now, though I will leave this new pattern in a branch somewhere in case it becomes useful later on. Arnej has some vision for testing various `SubmissionHistory` scenarios that might require it.

prime-router/src/test/kotlin/history/azure/SubmissionFunctionIntegrationTests.kt
thetaurean: I took a look and did not think of any in the context of status/history.

I was thinking maybe an assertion to ensure the lineage is "terminated" so to to speak, but that would just assert against data we inserted.

prime-router/src/test/kotlin/history/SubmissionHistoryIntegrationTests.kt
thetaurean: The only problem I see with this approach is that if someone changes the pipeline such that the outputs (ActionLogs, etc.) change, the SubmissionHistory tests won't break because their input is agnostic to any changes in the pipeline.  The associated pipeline tests will break (and get fixed) but the SubmissionHistory tests stay happy.

Maybe that's okay though, thoughts? We could always add a comment to the relevant pipeline tests to be wary of the loose coupling to history tests.

mkalish: So I think thats ok?  My two cents is that we're just testing that we can correctly read and parse what's been inserted into the DB and that a broader smoketest is what verifies that the whole thing working together.

thetaurean: The broader smoketest suite does not catch these bugs. So maybe enhancement is needed there?

Thought for one of these bugs I'm not sure current smoketest design pattern can catch the issue. Since it only presents at a specific point in the pipeline (after receiver step, before destination filter completion) and smoke tests can't reliably time such a history API call.

Issue #15805, opened 2024-09-07
Surface version info in more places


jack-h-wang:
### User Story
As an RS engineer, I want to see which version of the app performed logged actions, so that more diagnostic information is available.

### Description/Use Case
A `version.json` file is now available to identify the version of the functionapp. Adding this to the information logged to Azure would allow us to track which version performed any given action.

### Dev Notes
Adding this information could take a few forms, such as:
- Automatically inserting version within Azure custom events.
- Writing a log entry with the information.
Use cases for this information should be considered when deciding where the version is to be supplied.

### Acceptance Criteria
- [ ] Version information is visible when viewing events within the Azure portal

arnejduranovic:
- Where and how often is this referenced?
- What stats can we pull from the current implementation to inform risk?

jack-h-wang:
With #15805 merged I confirmed this is functioning as expected: the commitId that now appears in customEvents matches the content of `version.json` in staging.

PR #15902, opened 2024-09-17
15805 version in azure events

jack-h-wang:
This PR makes the current commit hash available to the main project and adds it to Azure events.

Test Steps:
1. Run Gradle task `package` or `quickPackage` and ensure the project builds successfully.

## Changes
- Add Gradle task to generate a Version object usable in the main project.
- Add version to `ReportEventData` and supply the commit hash from the Version object.

### Testing
- [x] Tested locally?
- [ ] Added tests?

## Linked Issues
- #15805

## To be done
- #15989

### DIFF
--- a/prime-router/build.gradle.kts
+++ b/prime-router/build.gradle.kts
@@ -48,7 +48,7 @@ plugins {
     id("io.swagger.core.v3.swagger-gradle-plugin") version "2.2.23"
 }
 
-val commitId = {
+val commitId by lazy {
     val stdout = ByteArrayOutputStream()
     exec {
         commandLine("git", "rev-parse", "--short", "HEAD")
--- a/prime-router/build.gradle.kts
+++ b/prime-router/build.gradle.kts
@@ -48,8 +48,17 @@ plugins {
     id("io.swagger.core.v3.swagger-gradle-plugin") version "2.2.23"
 }
 
+val commitId = {
+    val stdout = ByteArrayOutputStream()
+    exec {
+        commandLine("git", "rev-parse", "--short", "HEAD")
+        standardOutput = stdout
+    }
+    stdout.toString(StandardCharsets.UTF_8).trim()
+}
+
 group = "gov.cdc.prime.reportstream"
-version = "0.2-SNAPSHOT"
+version = "0.2-SNAPSHOT-$commitId"
 description = "prime-router"
 val azureAppName = "prime-data-hub-router"
 val azureFunctionsDir = "azure-functions"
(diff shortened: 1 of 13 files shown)


prime-router/build.gradle.kts
```diff
@@ -48,6 +48,16 @@ plugins {
...
+// retrieve the current commit hash
+val commitId by lazy {
+    val stdout = ByteArrayOutputStream()
+    exec {
+        commandLine("git", "rev-parse", "--short", "HEAD")
+        standardOutput = stdout
+    }
+    stdout.toString(StandardCharsets.UTF_8).trim()
+}
+
```
thetaurean, line 60: Does this end up running in CI? Are we sure that all the contexts this runs in has git?

As an alternative to running git, any thoughts on reading `.git/HEAD` and `.git/refs/<current-HEAD`?

prime-router/build.gradle.kts
```diff
@@ -502,16 +516,29 @@ tasks.azureFunctionsPackage {
...
+        package gov.cdc.prime.router.version
+        
+        /**
+         * Supplies information for the current build.
+         * This file is generated via Gradle task prior to compile time and should always contain the current commit hash.
+         */
+        object Version {
+          const val commitId = "$commitId"
+        }
+    """.trimIndent()
```
thetaurean, line 538: Seems a tad sketchy to be writing out Kotlin here.

Any thoughts on making the code static and read from the version.json file that is output? Or perhaps read directly from `.git/HEAD` and use that info to get the hash from `.git/refs` ?

prime-router/src/test/kotlin/azure/ActionHistoryTests.kt
```diff
@@ -352,7 +352,8 @@ class ActionHistoryTests {
             Topic.TEST,
             "http://blobUrl",
             TaskAction.send,
-            OffsetDateTime.now()
+            OffsetDateTime.now(),
+            ""
```
thetaurean, line 356: Any reason we shouldn't put at least a dummy version here? Ditto in other 4 locations.

thetaurean (commented):
Just a couple minor questions / suggestions.

prime-router/build.gradle.kts
jack-h-wang: This approach was suggested by @mkalish and is not dissimilar to the way jOOQ generates objects representing the database. I'm also not entirely certain the application can read files from those locations - the pivot to this approach was done amid difficulty attempting to implement exactly that.

prime-router/src/test/kotlin/azure/ActionHistoryTests.kt
jack-h-wang: It didn't seem relevant to these specific tests; I only added references where it was needed.

prime-router/build.gradle.kts
thetaurean: Nah if we know it works in all environments (which is obvious if already in master) then I'm good 👍.

thetaurean (approved):
👍

arnejduranovic (commented):
Once deployed, let's verify that the commit in the events matches the commit in the json file by logging in to SSH (process described in deployment doc)

prime-router/build.gradle.kts
```diff
@@ -502,16 +516,29 @@ tasks.azureFunctionsPackage {
 
 tasks.register("generateVersionFile") {
```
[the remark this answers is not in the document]
jack-h-wang: Added the TODO and created #15989.

arnejduranovic (approved):

PR #15928, opened 2024-09-19
move initializeQueueMessages to companion object

brick-green:
This PR moves initializeQueueMessages to companion object in BaseEngine

Test Steps:
1. run integration and unit tests


mkalish (approved):
👍🏻

PR #15990, opened 2024-09-24
update submission received event and delete v1 azure events

### LINKED ISSUE #13161: Update ReportFunction ReportReceivedEvent
### User Story

As an Engineer on ReportStream,
I would all ReportReceivedEvent Azure Events to have the same info,
so that debugging can be consistent across all submission endpoints.

### Description/Use Case

With the creation of a new microservice in  the`submissions` spring app, there exist two implementations of ReportReceivedEvent that we should consolidate. 

```
NEW PROPOSED EVENT OBJECT

data class ReportReceivedEvent(
    val timeStamp: Instant,
    val reportId: UUID,
    val parentReportId: UUID,
    val rootReportId: UUID,
    val headers: Map<String, String>,
    val queryParameters: Map<String, String>,
    val method: HttpMethod,
    val url: String,
    val sender: String,
    val senderIP: String,
    val fileSize: String,
    val blobUrl: String,
)
```

```
prime-router/src/main/kotlin/azure/observability/event/ReportReceivedEvent.kt

data class ReportReceivedEvent(
    val reportId: ReportId,
    val sender: Sender,
    val submissionDetails: ActionHistory.ReceivedReportSenderParameters,
    val senderIP: String,
    val fileSize: String,
) : AzureCustomEvent
```

```
submissions/src/main/kotlin/gov/cdc/prime/reportstream/submissions/ReportReceivedEvent.kt

data class ReportReceivedEvent(
    val timeStamp: Instant,
    val reportId: UUID,
    val parentReportId: UUID,
    val rootReportId: UUID,
    val headers: Map<String, String>,
    val sender: String,
    val senderIP: String,
    val fileSize: String,
    val blobUrl: String,
)
```

### Dev Notes

The microservice does not make use of the `AzureCustomEvent`, and it does not necessarily need to be updated to do so, as there is no immediate need for this since we expect ReportFunction to be deprecated relatively soon.
### Acceptance Criteria

- [ ] `prime-router/src/main/kotlin/azure/observability/event/ReportReceivedEvent.kt` contains the same info as the `submissions/ReportReceivedEvent`

brick-green:
This PR deletes the v1 azure event data classes and improves the event emitted from the submission controller. It now includes queryParamters, httpMethod, and full request url. 

Test Steps:
1. `./gradlew bootRun` to make sure it compiles
2. run submissionController tests
3. run submissionControllerIntegration tests

## Changes
- update SubmissionReceivedEvent
- add retrieval and filtering for query parameters
- retrieve url and http method
- delete unused azure event data classes

### Testing
- [X] Tested locally?

## Linked Issues
- Linked #13161


prime-router/docs/observability/azure-events.md
```diff
@@ -202,55 +93,43 @@ customEvents
...
 ```
 customEvents
-| where name == "ReportAcceptedEvent"
-| extend observations = parse_json(tostring(customDimensions.observations))
-| mv-expand observations
-| extend conditions = parse_json(tostring(observations.conditions))
-| mv-expand conditions
-| extend conditionDisplay = tostring(conditions.display)
+| where name == "ITEM_ACCEPTED"
+| extend conditionDisplay = tostring(parse_json(tostring(parse_json(tostring(parse_json(tostring(parse_json(tostring(parse_json(tostring(customDimensions.params)).bundleDigest)).observationSummaries))[0].testSummary))[0].conditions))[0].display)
```
jalbinson, line 97: there has to be a better way 😆 

If it works, it works though.

submissions/src/main/kotlin/gov/cdc/prime/reportstream/submissions/controllers/SubmissionController.kt
```diff
@@ -237,6 +247,11 @@ class SubmissionController(
         return headers.filter { it.key.lowercase() in headersToInclude }
     }
 
+    private fun filterQueryParameters(queryParameters: Map<String, String>): Map<String, String> {
+        val headersToInclude = emptyList<String>() // update this list as new QueryParameters are needed
```
jalbinson, line 251: This should be moved to configuration.

jalbinson: Same with headers as well.

prime-router/docs/observability/azure-events.md
brick-green: I just used the UI in azure to extend on that field and that's what it gave me. I agree it looks kinda messy though

submissions/src/main/kotlin/gov/cdc/prime/reportstream/submissions/controllers/SubmissionController.kt
```diff
@@ -98,11 +102,17 @@ class SubmissionController(
...
             parentReportId = reportId,
             rootReportId = reportId,
-            headers = filterHeaders(headers),
-            sender = clientId,
-            senderIP = senderIp,
-            fileSize = contentLength,
-            blobUrl = blobClient.blobUrl
+            requestParameters = SubmissionDetails(
+                filterHeaders(headers),
+                filterQueryParameters(request.parameterMap.mapValues { it.value.joinToString(",") })
```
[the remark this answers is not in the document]
brick-green: I don't understand. Those two filter functions return maps of the headers and parameters so that they are separated for querying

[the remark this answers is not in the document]
jalbinson: if we have the query string `?param=a&param=b&param=c` then your code will serialize it to: 
```kotlin
mapOf(
  "param" to "a,b,c"
)
```
So then say we want to look for all events where `param` contains `b` we would have to break out "a,b,c" in KQL.

My suggestion is make the fields in `SubmissionDetails` to `Map<String, List<String>>` and let jackson do the serialization work so its easier to parse it back to json in KQL.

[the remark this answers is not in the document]
brick-green: I set it up to account for a single query parameter to have multiple values. I did not set it up for headers though as I don't see a use case for that yet. I also think it's a little out of scope for this ticket as I would need to update the queue message then also update the FHIRReceiver to account for the new type. If we ever actually need to accept query parameters we'd also need to do the same.

submissions/src/main/resources/application.properties
```diff
@@ -4,4 +4,15 @@ azure.storage.connection-string=${AZURE_STORAGE_CONNECTION_STRING:DefaultEndpoin
 azure.storage.container-name=${AZURE_STORAGE_CONTAINER_NAME:reports}
 azure.storage.queue-name=${AZURE_STORAGE_QUEUE_NAME:elr-fhir-receive}
 azure.storage.table-name=${AZURE_STORAGE_TABLE_NAME:submission}
-spring.security.oauth2.resourceserver.jwt.issuer-uri=https://reportstream.oktapreview.com/oauth2/ausekaai7gUuUtHda1d7
\ No newline at end of file
+spring.security.oauth2.resourceserver.jwt.issuer-uri=https://reportstream.oktapreview.com/oauth2/ausekaai7gUuUtHda1d7
+
+# Allowed headers
+allowed.headers.client_id=client_id
```
jalbinson, line 10: Few things:

This being a map doesn't seem to serve a purpose since both key/value are the same(ish) and you ignore the key in code anyway. In code the type will just be `List<String>`.

I think adding an additional prefix would help clarify the configuration as well. Something like `event.allowed.headers`?

Last bit is I think yml really simplifies this file over a properties file. Here's what the result would look like:
```yml
event.allowed:
  headers:
    - client_id
    - content-type
    - payloadname
    - x-azure-clientip
    - content-length
  queryParameters:
    - ???
```

With a properties file there is much more duplication and you need to add indices.
```properties
event.allowed.headers[0]=client_id
event.allowed.headers[1]=content-type
...
```

submissions/src/main/kotlin/gov/cdc/prime/reportstream/submissions/controllers/SubmissionController.kt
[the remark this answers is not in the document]
jalbinson: works for me.

submissions/src/main/resources/application.properties
brick-green: got it fixed like this. I did leave it as just `allowed` since this list is also used to filter out parameters for the queue message

jalbinson (approved):
Looks good! Thanks for bearing with me over the config stuff. Sets us up nicely for down the road.

submissions/src/main/kotlin/gov/cdc/prime/reportstream/submissions/config/AllowedParametersConfig.kt
```diff
@@ -0,0 +1,41 @@
...
+ * @property queryParameters A list of allowed query parameter names that can be used in incoming requests.
+ */
+@Component
+@ConfigurationProperties(prefix = "allowed")
+class AllowedParametersConfig {
+    /**
+     * A list of allowed HTTP headers that can be accepted by the API.
+     * Each entry in the list represents a header name expected in the incoming request.
+     */
+    var headers: List<String> = emptyList()
```
arnejduranovic, line 34: This shouldn't be empty, right? Only queryParams for now? Off the top of my head, payloadName should be supported, along with a few others?

brick-green: This is the field that holds the injected list of allowable header. It's defaulted to an empty list so that if someone in the future wants to allow all headers then they can. The list of allowable headers, including payloadName, is in the `application.yml` file.

arnejduranovic (approved):

PR #16002, opened 2024-09-25
Address intermediate "Not Delivering" before routing a received report

### LINKED ISSUE #15524: [BUG]: History endpoint intermediate overallStatus can be "Not Delivering" and then change later

### LINKED ISSUE #15645: [BUG] Submission History API doesn't show Filtered reports

thetaurean:
This PR adjusts submission history calculation logic to take into account next actions that are not scheduled for any particular time (that only use `nextAction` and don't set a `nextActionAt`).

Test Steps:
1. Run tests

## Changes
- SubmissionHistory
  - Adds logic to submission history to check nextAction as well as nextActionAt
  - Fixes bug in submission history logic to find last action
- Test and test utility changes
  - Adds a submission history integration test for a received & converted report
  - Adds logic to `ReportGraphBuilder`/`ReportNodeBuilder`to set `createdAt` to avoid same value for all nodes
  - Adds logic to `ReportGraphBuilder`/`ReportNodeBuilder` to explicity set `nextAction`
  - Sets `nextAction` to `none` for tests cases where reports will not be delivered

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- Fixes #15524 
- Fixes #15645


adegolier (commented):
Ran new test against master and it failed, passes against this branch, so 👍 there

prime-router/src/main/kotlin/history/SubmissionHistory.kt
```diff
@@ -182,6 +182,12 @@ class DetailedSubmissionHistory(
     @JsonIgnore
     var nextActionScheduled = false
 
+    /**
+     * Flag to check if there's a next action for the newest report in this submission
+     */
+    @JsonIgnore
+    var hasNextAction = false
```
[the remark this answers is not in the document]
thetaurean: `nextActionScheduled` tracks scheduled next actions -- ones that use the `nextActionAt` field. Currently that's only the send step (maybe batch also but I don't think so).

`hasNextAction` tracks `nextAction`, looking at the latest report in the lineage.

prime-router/src/test/kotlin/common/ReportNodeBuilder.kt
```diff
@@ -130,7 +142,14 @@ class ReportGraphBuilder {
...
             .setTransportResult(node.theTransportResult)
-            .setNextAction(node.reportGraphNodes.firstOrNull()?.theAction)
+            .setNextAction(
+                if (node.theNextAction != null) {
+                    node.theNextAction
+                } else {
+                    node.reportGraphNodes.firstOrNull()?.theAction
+                }
+            )
+            .setCreatedAt(graph.node.createdAt.plusMinutes(1))
```
thetaurean, line 152: This was added because the nodes were generating with the same timestamp. This caused a sorting algorithm in the submission history function to fail which prevented it from finding the correct last action/report in the lineage.

prime-router/src/test/kotlin/fhirengine/azure/FHIRReceiverFilterIntegrationTests.kt
thetaurean, line 1: This adds a couple more assertions for #15846

adegolier (approved):
lgtm

arnejduranovic (approved):

Issue #15223, opened 2024-07-17
Support Generating an Empty Report in Convert Step


arnejduranovic:
# User Story
As an RS developer, I would like to be able to generate an empty report so that lineages work correctly.

# Description/Use Case
Currently, the code only support generating an empty report as part of sending something to receiver (i.e. they're on batch schedule and always want a report even if there are no items). However, the convert step also needs to be able to generate an empty report in the following situations:
- the submitted report was completely empty
- the hl7 message was unparseable

The empty report must be generated because otherwise the report lineage doesn't work and the action logs from the convert step are not visible.

# Dev Notes
ActionHistory.trackGeneratedEmptyReport already exists and overload likely can get created to support this. Another option is to remove exceptions that are based on assumptions that a report cannot have empty item lineages.

Search for ticket number in code, there is a comment.

# Acceptance Criteria
- [ ] Convert step updated to generate an empty report and report lineage (no item lineage) when certain conditions are met

Andrey-Glazkv:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/15223?planning-poker) with Zenhub @adegolier @arnejduranovic @brick-green @david-navapbc @jack-h-wang @jalbinson @JFisk42 @mkalish @thetaurean

PR #16137, opened 2024-10-07
15223: generate empty in convert

mkalish:
This PR updates the code so that when an empty report (or any situation where the number of items cannot be determined) makes it to the convert step.

Test Steps:
There are a integration tests that covers this, but as the pipeline is currently structured this is hard to test manually as the report function will synchronously fail. 

## Changes
- Adds a new `trackEmptyReport` that can be used anywhere in the pipeline
- Updates the convert step to use it if it cannot determine the number of items

### Testing
- [x] Tested locally?
- [x] Added tests?

### Process

## Linked Issues
- #15223

### DIFF
--- a/prime-router/src/main/kotlin/azure/ActionHistory.kt
+++ b/prime-router/src/main/kotlin/azure/ActionHistory.kt
@@ -452,14 +452,11 @@ class ActionHistory(
         generatingEmptyReport = true
         val reportFile = ReportFile()
         reportFile.reportId = report.id
-
-        reportFile.nextAction = TaskAction.send
-        reportFile.schemaName = trimSchemaNameToMaxLength(report.schema.name)
         reportFile.schemaTopic = report.schema.topic
-        reportFile.itemCount = report.itemCount
+        reportFile.itemCount = 0
         reportFile.bodyFormat = report.bodyFormat.toString()
-        reportsOut[reportFile.reportId] = reportFile
         reportFile.nextAction = TaskAction.none
+        reportsOut[reportFile.reportId] = reportFile
     }
 
     /**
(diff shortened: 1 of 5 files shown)


prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -448,6 +448,17 @@ class ActionHistory(
         }
     }
 
+    fun trackEmptyReport(report: Report) {
```
mkalish, line 451: This is generic so we can track an empty report anywhere we need to do

prime-router/src/test/kotlin/common/UniversalPipelineTestUtils.kt
```diff
@@ -257,12 +257,13 @@ object UniversalPipelineTestUtils {
...
     /**
-     * fetch child reports associated with a [parent] report and ensure we find an [expected] number of children
+     * fetch child reports associated with a [parent] report and ensure we find an [expectedItems] number of children
      */
     fun fetchChildReports(
         parent: Report,
         txn: DataAccessTransaction,
-        expected: Int? = null,
+        expectedItems: Int? = null,
+        expectedReports: Int = 1,
```
mkalish, line 266: This needed to get adjusted since its now possible to generate a report with no lineages (i.e. an empty report).

thetaurean (approved):
LGTM 👍 One small comment:

I wonder if there's any tech debt in `ActionHistory.kt` worth ticketing. I remember getting a little lost / confused when tracking a created empty report in other steps.

Separate but related, should we scope some tickets for using this new function in other places we generate an empty report (e.g. `FHIRDestinationFilter.kt` and `FHIRReceiverFilter.kt`)? Will spend a little more time thinking this over tomorrow.

mkalish:
@thetaurean 100% think there are tech debt tickets to be made for action history but I have always been unsure where to start since I think there could be an argument to simply rewrite the whole thing.

For the other pipeline steps, maybe?  This spot in in convert is a little unique in the sense that its only part of the pipeline where the number of items is unknown in case of an error

arnejduranovic (approved):

Issue #14609, opened 2024-06-04
When trying to view delivery details for reports that are not batched but sent as single messages#


JessicaWNava:
### User Story
As a receiver who has useBatchHeaders set to false. I want to be able to see my reports in Daily Data without errors. 

### Description/Use Case
When there are multiple reports sent to a receiver at the exact same time (which occurs when there is a receiver not using batch headers, but is instead receiving each message as it's own report). The delivery details link returns a 500 error with the exception:

`"org.jooq.exception.TooManyRowsException Message: Cursor returned more than one result"`

This can by observed logging in as ar-phd in prod, going to [Daily Data](https://reportstream.cdc.gov/daily-data), and then selecting the "reportID" link to see more details for report_ID "18955639-1c6d-4182-a078-f91f6ecbdc25"

### Dev Notes
This occurs because the query in the fetchAction method in the DatabaseDeliveryAccess class throws an exception because there is more than one report with the same action id.

It also looks like this bug is manifesting in other ways, see: https://nava.slack.com/archives/C048S3GPSQG/p1725376883033999. The TLDR of this one is that may be the root cause of these issues: 

> I think this is a bug when the receiver has batching disabled and has multiple reports that need to be sent out at the same time. They are sent individually, but still get the same action id for the batch step

### Acceptance Criteria
- [ ] The delivery details page loads for reports that originated from a report with multiple messages to a receiver who has useBatchheaders set to false.
- [ ] The fetchAction method can handle the situation where there are multiple reports returned OR multiple reports being returned are eliminated as a possibility. 
- [ ] Investigated to see if it makes sense to make the action ID in the batch step unique so these errors do not occur

Andrey-Glazkv:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/14609?planning-poker) with Zenhub @adegolier @arnejduranovic @brick-green @david-navapbc @jack-h-wang @jalbinson @JFisk42 @mkalish @thetaurean

PR #16188, opened 2024-10-09
14609: findDetailedDeliveryHistory handles actions that produce multiple reports

mkalish:
This PR updates fetching detailed delivery to count for actions that produce more than on report

Test Steps:
Follow these steps on main and they'll error, but on this branch they'll pass
1. Send a report through with at least two times from `ignore.ignore-fhir-e2e`
2. Wait for the items to get delivered
3. Find a report ID from the batch step going to the `ignore.FULL_ELR_FHIR` receiver
4. Run
```sh
curl --location 'http://localhost:7071/api/waters/report/afbeca26-8b1a-4a57-9749-8000c731e473/delivery' \
--header 'Authorization: Bearer dummy' \
```
For the success case, you should get back a single delivered item

## Changes
- Updates the findDetailedDeliveryHistory to individually retrieve the report and action if the requested id is a report id

### Testing
- [x] Tested locally?
- [x] Added tests?

### Process

## Linked Issues
- [Fixes #issue](https://github.com/CDCgov/prime-reportstream/issues/14609)

## To Be Done
- https://github.com/CDCgov/prime-reportstream/issues/16191

### DIFF
--- a/prime-router/src/main/kotlin/history/azure/DeliveryFacade.kt
+++ b/prime-router/src/main/kotlin/history/azure/DeliveryFacade.kt
@@ -91,6 +91,7 @@ class DeliveryFacade(
     /**
      * Get expanded details for a single report
      *
+     * @param id  either a report id (UUID) or action id (Long)
      * @param deliveryId id for the delivery being used
      * @return Report details
      */
--- a/prime-router/src/main/kotlin/history/azure/DeliveryFacade.kt
+++ b/prime-router/src/main/kotlin/history/azure/DeliveryFacade.kt
@@ -19,7 +19,7 @@ import java.util.*
  */
 class DeliveryFacade(
     private val dbDeliveryAccess: DatabaseDeliveryAccess = DatabaseDeliveryAccess(),
-    dbAccess: DatabaseAccess = BaseEngine.databaseAccessSingleton,
+    val dbAccess: DatabaseAccess = BaseEngine.databaseAccessSingleton,
     val reportService: ReportService = ReportService(),
 ) : ReportFileFacade(
     dbAccess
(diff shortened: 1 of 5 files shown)


prime-router/src/main/kotlin/history/azure/DeliveryFacade.kt
```diff
@@ -19,7 +19,7 @@ import java.util.*
  */
 class DeliveryFacade(
     private val dbDeliveryAccess: DatabaseDeliveryAccess = DatabaseDeliveryAccess(),
-    dbAccess: DatabaseAccess = BaseEngine.databaseAccessSingleton,
+    val dbAccess: DatabaseAccess = BaseEngine.databaseAccessSingleton,
```
jalbinson, line 22: Make this a `private val` so `dbAccess` cannot be accessed outside the class

prime-router/src/main/kotlin/history/azure/DeliveryFacade.kt
```diff
@@ -95,13 +95,27 @@ class DeliveryFacade(
...
     ): DeliveryHistory? {
-        val deliveryHistory = dbDeliveryAccess.fetchAction(
-            deliveryId,
-            orgName = null,
-            DeliveryHistory::class.java
-        )
+        val reportFileId = try {
+            UUID.fromString(id)
+        } catch (ex: IllegalArgumentException) {
+            null
```
jalbinson, line 104: log some kind of warning (but don't include the exception for now since it'll trigger pagerduty)

prime-router/src/main/kotlin/history/azure/DeliveryFacade.kt
```diff
@@ -95,13 +95,27 @@ class DeliveryFacade(
      * @return Report details
      */
     fun findDetailedDeliveryHistory(
+        id: String,
```
jalbinson, line 98: can this parameter be changed to `reportFileId` and then internally we use `parsedReportFileId`? At a glance from the function signature it's hard to tell what `id` is referring to. Update the javadoc comments as well.

prime-router/src/main/kotlin/history/azure/DeliveryFacade.kt
mkalish: I actually don't think we should log since its not unexpected that this could be an action id rather than a report UUID

prime-router/src/main/kotlin/history/azure/DeliveryFacade.kt
mkalish: Yep this is because this can be iether a UUID or a Long.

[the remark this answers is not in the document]
mkalish: I'll update the KDOC, but I believe the intention was that this method works with either a report id or an action id (UUID vs Long)

jalbinson: could split out into 2 functions and use the types and have a private shared implementation
```kotlin
fun findDetailedDeliveryHistory(
  reportId: UUID,
  deliveryId: Long,
)

fun findDetailedDeliveryHistory(
  actionId: Long,
  deliveryId: Long,
)
```

mkalish: So honestly it just pushes the same issue further up?  The root cause is the shared inheritance between the submission and delivery fetching, but I view that as out of scope for this work

jalbinson: Looks like the design of this from the ground up has report ID and action ID mashed together so I agree it doesn't make sense to tackle it here.

Prying those types back apart in multiple places via UUID parsing exceptions sets off my code smell alarm bells though. An option for tech debt tickets down the line I guess 🤷

mkalish: I'll give myself a TODO to open a tech debt ticket for this.

jalbinson (approved):

arnejduranovic (approved):

PR #16257, opened 2024-10-18
unifying local and in memory azure event service classes into one

### LINKED ISSUE #15337: Remove the two test implementations of azure event service
### User Story
As a dev, I would like that there would only be one concrete test implement of AzureEventService

### Description/Use Case
There are two concrete implementations of `AzureEventService` when there only needs to be one.

### Dev Notes
- `AzureEventServiceImpl` is the implementation used in production 
- The two test implementations that need to be collapsed:
  - `LocalAzureEventServiceImpl`
  - `InMemoryAzureEventService`

### Acceptance Criteria
- [ ] There is only one concrete implementation of `AzureEventService` that is used in the tests

kant777:
This PR unifies local and in memory azure event service classes into one

Test Steps:
1. gradle build

image

## Changes

### Testing
- [x] Tested locally?
- [ ] Added tests?

## Linked Issues
- Fixes #15337


david-navapbc (approved):
unit and integration tests pass for me - changes look ok. 

couple of quick things about the PR itself for future reference 

* in the "steps to replicate" section you instruct to use the "build" target. The team has had prior conversations about this and according to @mkalish we shouldn't be using the "build" target but instead the "package" target. NBD and this is partly why we were having dialogue today about build file and documenting what the right thing to do is with respect to build tasks. 
* there's a doc somewhere that lays out how branches should be named. @MichaelEsuruoso @arnejduranovic or @mkalish  will know where it is. My understanding is branches should be named as such `platform/kalish/16084-sftp-ingestion-for-reports` 
* under "changes" you left the default text there. Please in future either remove that section or populate it with ticket-specific changes. 

🍻  and nice job!

arnejduranovic (approved):

Issue #16140, opened 2024-10-07
Update UP Exception Handling Strategy


arnejduranovic:
### User Story

As the owner of the Universal Pipeline,
I want exceptions to be caught by default and handled gracefully,
so that I do not have to rely on the built-in azure retry logic which comes with its own downsides.

### Description/Use Case

See [Receiver Enrichment SRD](https://cdc.sharepoint.com/:w:/r/teams/ReportStream/_layouts/15/doc2.aspx?sourcedoc=%7BCD15C571-D563-4135-B293-6FFAA77826AE%7D&file=SRD-ReceiverEnrichment.docx&action=default&mobileredirect=true) for the requirements of this feature. Each step's requirements in this document outlines the strategy like so:

1. Not modify the database at all for any item (i.e. not process any items) 
2. Generate a `PIPELINE_EXCEPTION` Azure Event 
3. Manually populate `<specific-step>-poison` Azure Queue with the passed in `QueueMessage`  
4. Exit the pipeline step gracefully 

See the "Handling Failures" section of the document for more details.

### Acceptance Criteria

- [ ] Updated UP exception handling per [Receiver Enrichment SRD](https://cdc.sharepoint.com/:w:/r/teams/ReportStream/_layouts/15/doc2.aspx?sourcedoc=%7BCD15C571-D563-4135-B293-6FFAA77826AE%7D&file=SRD-ReceiverEnrichment.docx&action=default&mobileredirect=true) for all UP steps.
- [ ] Update the blob trigger handling for the SFTP ingestion

MichaelEsuruoso:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/16140?planning-poker) with Zenhub @adegolier @arnejduranovic @david-navapbc @jack-h-wang @JFisk42 @kant777 @mkalish @thetaurean @jalbinson

MichaelEsuruoso:
@mkalish closing out ticket. Does not require a demo

PR #16258, opened 2024-10-18
16140: manually populate poison queue when a step encounters an error

mkalish:
This PR updates the FHIR function error handling to manually add most error cases to the poison queue

Test Steps:
This is mostly covered by automation, but to test locally.

1. In FHIRConverter, update the `doWork` to throw a Runtime exception
2. Verify that the poison queue is populated and the message is not retried

## Changes
- Updates how the FHIR functions run each FHIR engine so that only data access exceptions are retried and all other are manually moved to the poison queue

### Testing
- [x] Tested locally?
- [x] Added tests?

### Process

## Linked Issues
- #16140


prime-router/src/test/kotlin/fhirengine/azure/FHIRConverterIntegrationTests.kt
```diff
@@ -158,7 +158,7 @@ class FHIRConverterIntegrationTests {
     @BeforeEach
     fun beforeEach() {
         mockkObject(QueueAccess)
-        every { QueueAccess.sendMessage(any(), any()) } returns Unit
+        every { QueueAccess.sendMessage(any(), any()) } returns ""
```
adegolier, line 161: Should this be actually returning something besides empty String now?

adegolier (approved):
One small question, but otherwise lgtm. Tests all passed for me and made the change to FHIRConverter as indicated in the test steps and saw it go to the poison queue instead of being retried as expected.

prime-router/src/test/kotlin/fhirengine/azure/FHIRConverterIntegrationTests.kt
mkalish: I think in this case it doesn't actually matter since we only need the return value for reporting purposes if we encounter an exception

prime-router/src/main/kotlin/azure/QueueAccess.kt
```diff
@@ -66,13 +66,14 @@ object QueueAccess {
...
         val timeToLive = invisibleDuration.plusDays(timeToLiveDays)
-        createQueueClient(queueName).sendMessageWithResponse(
+        val response = createQueueClient(queueName).sendMessageWithResponse(
             message,
             duration,
             timeToLive,
             null,
             null
         )
+        return response.value.messageId
```
arnejduranovic, line 76: Nice! Given an Azure Error Event, this allows us to easily find the associated queue message in the poison queue.

arnejduranovic (approved):

PR #16292, opened 2024-10-21
15864 Spring Cloud Gate configuration with Swagger

### LINKED ISSUE #15864: Investigate Spring Cloud Gateway support for only allowing the proxying of certain URLs
### User Story
Our current auth POC uses a simple global wildcard to forward all requests to their correct destination whether it exists or not. This is causing issues with serving swagger ui and also may be forwarding requests to invalid endpoints that could be handled at the gateway level.

### Description/Use Case
See if requests could be set up via configuration as seen in [Spring Cloud Gateway documentation](https://spring.io/guides/gs/gateway)

### Dev Notes
Ensure Swagger can be used if added
Investigate if Azure FrontDoor could be used for this.

### Acceptance Criteria
if possible...
-POC updated to route specific configuration (remove wildcard request mapping)
-Requests that are forwarded are still being checked for authentication and 401 if missing/invalid
-Swagger loads
-unit tests continue to pass

if not possible...
-Document reasoning in project and write ticket for alternative way to limit scope of possible forwarded requests

jalbinson:
This PR updates the Auth service to use Spring Cloud Gateway configuration to forward requests rather than a wildcard request mapping. It also enabled swagger when running the application locally.

Test Steps:
To ensure Gateway is working as expected
1. Run `gradle quickRun` to start up ReportStream
2. Run `gradle submissions:bootRun`
3. Run `gradle auth:bootRun`
     (Contact me if having an issue with a missing secret and I will share the staging okta secret with you)
4. Run Arnej token script to retrieve access token
    (If this is a huge pain I can send you a valid access token)
6. Make a request to POST http://localhost:9000/api/v1/reports
7. If you get a 415 Unsupported Media Type response it made it to submissions

To see Swagger
1. Run `gradle auth:bootRun`
2. Access http://localhost:9000/swagger/ui.html
3. See Swagger!

## Changes
- Removed unneeded code
- Implemented configuration based gateway pattern
- Included Swagger

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- Fixes #15864


auth/src/test/kotlin/gov/cdc/prime/reportstream/auth/controller/HealthControllerTest.kt
```diff
@@ -0,0 +1,32 @@
...
+
+@ExtendWith(SpringExtension::class)
+@SpringBootTest
+@AutoConfigureWebTestClient
+class HealthControllerTest @Autowired constructor(
+    private val webTestClient: WebTestClient
+) {
+
+    @Test
+    fun `successful healthcheck`() {
```
jalbinson, line 22: Added a simple test here just to make sure project is compiling and healthcheck endpoint is functional.

auth/src/test/kotlin/gov/cdc/prime/reportstream/auth/controller/AuthControllerTest.kt
jalbinson, line 1: I removed the AuthControllerTest as the AuthController no longer exists. I believe this style of test would be better suited for an end-to-end test in the future.

Since the auth and cloud gateway functionality are all now configuration classes I don't think unit tests are much help. The internal functionality of those plugins should already be tested by their own authors.

auth/src/main/kotlin/gov/cdc/prime/reportstream/auth/config/SubmissionsConfig.kt
```diff
@@ -0,0 +1,13 @@
...
+import org.springframework.boot.context.properties.ConfigurationProperties
+import org.springframework.context.annotation.Configuration
+
+/**
+ * Configuration for Submissions microservice
+ */
+@Configuration
+@ConfigurationProperties(prefix = "submissions")
+data class SubmissionsConfig(
+    var baseUrl: String = "http://localhost:8080"
```
mkalish, line 12: Just confirming, the default isn't strictly required and this is ready from the config?

auth/src/main/resources/application.yml
```diff
@@ -1,26 +1,28 @@
 spring:
   application:
     name: "auth"
-  profiles:
```
mkalish, line 4: Isn't this how we would want to handle this in the other environments?  Fine if we're punting for now...

auth/src/main/resources/application.yml
```diff
@@ -1,26 +1,28 @@
...
-    baseUrl: http://localhost:8880
+app:
+  environment: local
+
+# submissions microservice configuration
+submissions:
+  baseUrl: http://localhost:8080
+
+# Ensure these are disabled in production
+springdoc:
```
mkalish, line 21: I'm not sure what we're getting out of the swagger docs, what are the other endpoints that would ultimately be here?

mkalish (commented):
Changes look good to me, other than the CSRF message we chatted on.

I'll take a final look when you push that update

auth/src/main/kotlin/gov/cdc/prime/reportstream/auth/config/SubmissionsConfig.kt
jalbinson: There some weirdness without the default value. I can look more deeply into it. I think using those annotations is written to work well with standard java getters/setters and not as cleanly with a data class.

Screenshot 2024-10-22 at 2 29 31 PM

auth/src/main/resources/application.yml
jalbinson: profiles seemed to be a nice way to have different beans injected via environment but now the project is so simple that it doesn't require it.

auth/src/main/resources/application.yml
jalbinson: It would be more helpful to write an openapi spec for the URLs were expecting and have swagger load that.

auth/src/main/kotlin/gov/cdc/prime/reportstream/auth/config/SubmissionsConfig.kt
jalbinson: Looks like theres a workaround I'll try

https://stackoverflow.com/questions/76626376/configurationproperties-without-default-values-in-spring-boot-3-using-kotlin

jalbinson: that workaround worked great. It also let me use vals instead of vars 👍

mkalish (approved):
Looks good to me, was able to auth successfully and get a 401 without the token

auth/src/main/resources/application.yml
arnejduranovic: Ya the thought was we would document the APIs this proxy supports in the proxy itself since the other endpoints will not be available to the outside world

jalbinson: I just found this article which would be nice way to handle this!

https://medium.com/@erkndmrl/swagger-3-with-spring-cloud-gateway-a-comprehensive-guide-f355459d914a

tl;dr Setting up swagger on downstream services and configuring the auth project to load their specs

arnejduranovic (approved):

PR #16299, opened 2024-10-21
Update ORM mappings

JFisk42:
This PR updates our ORM mappings to use several of the newer catchall mappings (Provenance, Encounter, etc.). It also adds an ORM java class so that we can leverage HL7 v2.7 mapping datatypes for better compatibility. 

Test Steps:
1. Run the full gamut of tests: `./gradlew test`, `./gradlew testSmoke`, `./gradlew testIntegration`. 

## Changes
- Utilization of newer catchall mappings (`Provenance-Source`, `Provenance-Transformation`, `Provenance-Patient`, `Encounter`, as well as adding in segments to existing mappings.
- Creation of an ORM java class. Note: ORM does not exist as an official message type for HL7 v2.7. This Java class is labeled as 2.7 so that it can leverage forward and backward compatibility. ORM using 2.5.1 hapi structures was the root cause of the data loss that has been documented by ETOR. 
- Creation of an ORM integration tests fully populated with all currently available segments, and all currently mapped fields on those segments. 

### Testing
- [x] Tested locally?
- [ ] Added tests?

## To Be Done
TODO: I will be logging a ticket to add new mapping fields for various reasons. None of these are essential for the core ORM mappings.


prime-router/metadata/HL7/catchall/hl7/message/ORM_O01.yml
```diff
@@ -38,15 +49,16 @@ resources:
       - .ORC
       - .ORDER_DETAIL.OBR
       - .ORDER_DETAIL.OBSERVATION.NTE
-      - MSH
 
-  - resourceName: Specimen
```
JFisk42, line 43: Removed specimen here because it wasn't in the mapping inventory and only generates an essentially empty specimen resource because SPM shouldn't be in ORM.

prime-router/metadata/HL7/catchall/hl7/segments/OBX/Observation.yml
```diff
@@ -10,14 +10,6 @@ resourceType: Observation
...
 # OBX.20 is repeatable but maps to bodySite on Observation which is not, so it's all captured in OBXExtension
 # Not implementing OBX.22 not in mapping deprecated in NIST
-# Not implementing OBX.26 not in mapping not in NIST
-# Not implementing OBX.27 not in mapping not in NIST
-# Not implementing OBX.28 not in mapping not in NIST
-# Not implementing OBX.29 not in mapping not in NIST
-# Not implementing OBX.31 not in mapping not in NIST
-# Not implementing OBX.32 not in mapping not in NIST
-# Not implementing OBX.33 as it's not in the NIST spec the logic in the inventory cannot be implemented in the library
-
```
JFisk42, line 13: At some point we started supporting all of these fields, mostly through extensions.

prime-router/build.gradle.kts
```diff
@@ -900,7 +900,9 @@ dependencies {
     implementation("ca.uhn.hapi.fhir:org.hl7.fhir.r4:6.3.24")
     implementation("ca.uhn.hapi:hapi-base:2.5.1")
     implementation("ca.uhn.hapi:hapi-structures-v251:2.5.1")
+    implementation("ca.uhn.hapi:hapi-structures-v26:2.5.1")
```
JFisk42, line 903: I made a comment in `ORM_O01_PATIENT.java` about this. Basically, the 2.6 structure library is only necessary because `ORM_O01_INSURANCE.class` doesn't exist in 2.7. Once we've implemented the INS resources this can be removed.

david-navapbc (changes requested):
my understanding is that this PR updates our handling of translating hl7->fhir messages such that we're not dropping "unit" data from "ORM_001" messages due to catch-all mappings vs the mappings in this PR that specifically handles the "[CE](https://hl7.eu/HL7v2x/v24/std24/ch02.htm#Heading84)" field.  

On my workstation:

* `./gradle clean package` produces "Version" error we've experienced due to build file issues (see [slack thread](https://nava.slack.com/archives/C026DV8R21W/p1727983911032999?thread_ts=1727807919.641029&cid=C026DV8R21W)) 
* `./gradle clean` then `./gradle package` compiles successfully 
* `./gradle test` produces this error
    * [test_report.zip](https://github.com/user-attachments/files/17484007/test_report.zip)
* `./gradle testSmoke` passes. 

I would like to understand why `./gradle test` produces a failure before approving the PR. 

Otherwise all looks good to me 🍻

prime-router/src/main/java/fhirengine/translation/hl7/structures/fhirinventory/group/ORM_O01_PATIENT.java
```diff
@@ -0,0 +1,328 @@
...
+    init(factory);
+  }
+
+  private void init(ModelClassFactory factory) {
+    try {
+      this.add(PID.class, true, false, false);
+      this.add(PD1.class, false, false, false);
+      this.add(NTE.class, false, true, false);
+      this.add(ORM_O01_PATIENT_VISIT.class, false, false, false);
+      // If IN1 related classes are built out and ORM_O01_INSURANCE replaced, all hapiv26 dependencies can be removed
```
mkalish, line 77: Makes sense, I think there is a ticket Arnej wrote (will write?), could we tag it here?

prime-router/src/main/java/fhirengine/translation/hl7/structures/fhirinventory/group/ORM_O01_ORDER_DETAIL.java
```diff
@@ -0,0 +1,360 @@
...
+
+  /**
+   * Creates a new ORM_O01_ORDER_DETAIL group
+   */
+  public ORM_O01_ORDER_DETAIL(Group parent, ModelClassFactory factory) {
+    super(parent, factory);
+    init(factory);
+  }
+
+  private void init(ModelClassFactory factory) {
```
mkalish, line 75: Could we add a comment here indicating that caristix says there is an `ORDER_DETAIL_SEGMENT` which doesn't actually get modeled?

prime-router/src/main/java/fhirengine/translation/hl7/structures/fhirinventory/group/ORM_O01_PATIENT.java
mkalish: Also, just to confirm, the mappings for IN1 still need to be built right?

mkalish (approved):
Changes make sense to me!  Couple of comments.

Might make sense to merge `master` back into your branches to... I changed the location of the generated `Version` file when I fixed the gradle build which was causing me an error if I didn't clean

david-navapbc (approved):
Something odd is going on with that test such that it's failing but intermittently in a manner that I can't reliably reproduce. I *think* there's something stateful going on that has nothing to do with this branch. Approving this PR and will keep an eye out for this failure in particular in future. My hunch is that giving the build file(s) some ❤️ 's might help with whatever is triggering the issue. 

nice work and we're gtg on my end to merge this 🍻

prime-router/src/main/java/fhirengine/translation/hl7/structures/fhirinventory/group/ORM_O01_PATIENT.java
JFisk42: Talked to Arnej earlier this week and it sounds like building out more java classes is likely to be completed before the IN1 work. I linked [that ticket](https://github.com/CDCgov/prime-reportstream/issues/16307) in the comment and am leaving comments there and on the ticket for building out INx hapi classes.

JFisk42: [Added Comment](https://github.com/CDCgov/prime-reportstream/pull/16299/commits/45354ae50d215895943668c32bdcc50d1106f4bd)

prime-router/src/main/java/fhirengine/translation/hl7/structures/fhirinventory/group/ORM_O01_ORDER_DETAIL.java
JFisk42: Good idea, [added comment](https://github.com/CDCgov/prime-reportstream/pull/16299/commits/f129f2c62d0e9ca370d91468e08c3e13d6b7bc4c).

PR #16308, opened 2024-10-22
Platform/angela/16191/always execute send

### LINKED ISSUE #16191: A report for a receiver should always go through a SEND transport
### User Story
As a receiver of Reports,
I would like reports destined to me to always go through a send transport and default to the NullTransport if I do not have one configured,
so that my data can show up in the Daily Data page correctly.

### Description/Use Case
The transports are responsible for handling a lot of record keeping when they process the report (such as lineage tracking uploading the contents of the sent report) that does not happen if a receiver is not configured with any transport.  This leads to downstream complications such as the daily data fetching a list of reports from the batch step rather than the send step leading to slightly inaccurate data.

Does this impact the following?: https://github.com/CDCgov/prime-reportstream/issues/15644

### Dev Notes
- Which transport is used is configured [here](https://github.com/CDCgov/prime-reportstream/blob/51a8a01f93260e3dcce71ba7962a85163d35c532/prime-router/src/main/kotlin/azure/SendFunction.kt#L157)
- [NullTransport](https://github.com/CDCgov/prime-reportstream/blob/51a8a01f93260e3dcce71ba7962a85163d35c532/prime-router/src/main/kotlin/transport/NullTransport.kt#L15)

### Acceptance Criteria
- [ ] A report for a receiver uses a default Null transport when no transport is configured

adegolier:
This PR ...

Ensures that files destined for receivers with no transport configured still get an action logged for the send step. It also logs a send warning for receivers that have a transport type that is not yet supported by the SendFunction.

Test Steps:
1. Send a file on the covid-19 topic with ordering facility county set to HL7_NULL (Or set up another receiver with a null transport and send a file that will go to that receiver)
2. Verify that the file gets a record in the Action table with action_name "send" and action_result stating that it is going to the Null Transport. (I find it easiest to call the history API after the send step has executed, search the report_file table for the report_id of the file sent to the null transport receiver, and then check the associated action_id on the action table)

## Changes
- Adds a new field to Receiver for the TransportType which will return NullTransportType if there is no TransportType configured
- Modifies the SendFunction to use the new field to determine the transport that should be used
- Modifies the send_warning on the SendFunction to be more informational in the case that a TransportType is added that is not supported by the SendFunction's Transports.

### Testing
- [X] Tested locally?
- [X] Added tests?

## Linked Issues
- Fixes #16191


prime-router/src/main/kotlin/azure/SendFunction.kt
```diff
@@ -104,13 +103,15 @@ class SendFunction(
                 val nextRetryItems = mutableListOf<String>()
                 val externalFileName = Report.formExternalFilename(header, workflowEngine.reportService)
                 val sentReportId = UUID.randomUUID() // each sent report gets its own UUID
-                if (receiver.transport == null) {
+                val retryItems = retryToken?.items
+                val transport = getTransport(receiver.transportType)
+                if (transport == null) {
```
JFisk42, line 108: I think this logic isn't reachable anymore? Sorry my brain is kinda going in loops here. But wouldn't `receiver.transportType` set that value to `NullTransportType` which wouldn't evaluate to `null`?

prime-router/src/main/kotlin/transport/NullTransport.kt
```diff
@@ -26,7 +26,7 @@ class NullTransport : ITransport {
     ): RetryItems? {
         if (header.content == null) error("No content for report ${header.reportFile.reportId}")
         val receiver = header.receiver ?: error("No receiver defined for report ${header.reportFile.reportId}")
-        val msg = "Sending to Null Transport"
+        val msg = "Sending to Null Transport. File can be downloaded by Receiver until it expires."
```
JFisk42, line 29: Is this a message that is ever viewed by our consumers, or just our team? If viewed by consumers "Sending to Null Transport" could be confusing (especially if it's not a developer).

prime-router/src/test/kotlin/azure/SendFunctionTests.kt
```diff
@@ -134,6 +175,68 @@ class SendFunctionTests {
...
+        every { workflowEngine.db } returns mockk<DatabaseAccess>()
+        mockkObject(Report.Companion)
+        every { Report.formExternalFilename(any(), any(), any(), any(), any(), any(), any()) } returns ""
+
+        // Invoke
+        val event = ReportEvent(Event.EventAction.SEND, reportId, false)
+        SendFunction(workflowEngine).run(event.toQueueMessage(), context)
+
+        // Verify
+        verify { nullTransport.send(any(), any(), any(), any(), any(), any(), any(), any(), any()) }
```
JFisk42, line 203: I'm still learning my junit/kotlin unit testing. I suspect this verify isn't verifying anything? If we want to verify that the send method was called on nullTransport I think it needs something more, but totally not sure.

JFisk42 (commented):
Didn't have time to test this or look at it long enough to attempt to answer my own questions. But because I'm about to be out the rest of the day I wanted to post these incase they aren't just a case of "Josh not getting it".
I'll finish up reviewing/testing in the morning.

prime-router/src/main/kotlin/azure/SendFunction.kt
adegolier: So, that's for the case of a transport type is added (which is something that is on the receiver) but the SendFunction isn't updated with a transport to support it (I think it's weird that they're not tied together well, but since they are very separate, that seemed like a possibility and I wrote a test for that case)

prime-router/src/main/kotlin/transport/NullTransport.kt
adegolier: I'm pretty sure that's just in the DB for our logs. I can confirm that it doesn't show up on the History API, and I'm not sure what else might show that, since it's logged on the Action table.

prime-router/src/test/kotlin/azure/SendFunctionTests.kt
adegolier: My understanding is that's exactly what verify is for.... I could be wrong, tho

prime-router/src/main/kotlin/transport/NullTransport.kt
JFisk42: Right on. I'm very unfamiliar with what messages get viewed by just our engagement team or which might be seen by STLTs or other consumers so I just wanted to check.

prime-router/src/test/kotlin/azure/SendFunctionTests.kt
JFisk42: Yup this was noobie confusion and I didn't take the time to look it up myself because I was rushed yesterday. Our verify statements, and mocking in general, uses the `MockK` library for Kotlin. This blog is a pretty good overview of mocking using it: https://notwoods.github.io/mockk-guidebook/docs/mocking/verify/

> Using verify to verify that a function was called looks a lot like using every for stubbing. A simple example is to call a method and immediately check that it was called.

JFisk42 (approved):
Verified the test steps: 
- Sent the contents of `test-0001-input-covid-19.hl7` through `ignore.ignore-hl7`. 
  - There's a lot of messages going to a lot of places so I also posted a report using only [the message with an ORC-22.9 of `HL7_NULL` by itself](https://github.com/CDCgov/prime-reportstream/blob/a1627fa0a0518a7f1617122c31f36c6f9d26c61f/prime-router/src/main/resources/clitests/compare-test-files/test-0001-input-covid-19.hl7#L54-L70) to see it working in isolation. 
- History endpoint verified HL7_NULL as the receiver:
```    
"destinations": [
        {
            "organization": "FOR TESTING ONLY",
            "organization_id": "ignore",
            "service": "HL7_NULL",
            "itemCount": 1,
            "sending_at": "2024-10-23T23:50:00.000Z",
            "filteredReportRows": [],
            "filteredReportItems": [],
            "sentReports": [],
            "downloadedReports": []
        }
    ],
```
- Looking in the DB verified that on the action table, the corresponding action_name was `send` and action_result was `Sending to Null Transport. File can be downloaded by Receiver until it expires.`

prime-router/src/main/kotlin/azure/SendFunction.kt
```diff
@@ -162,7 +164,7 @@ class SendFunction(
             is SoapTransportType -> workflowEngine.soapTransport
             is GAENTransportType -> workflowEngine.gaenTransport
             is RESTTransportType -> workflowEngine.restTransport
-            is NullTransportType -> NullTransport()
+            is NullTransportType -> workflowEngine.nullTransport
```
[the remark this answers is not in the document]
adegolier: What should I be doing with the email transport type, then?

prime-router/src/main/kotlin/azure/SendFunction.kt
```diff
@@ -104,13 +103,15 @@ class SendFunction(
...
                 val externalFileName = Report.formExternalFilename(header, workflowEngine.reportService)
                 val sentReportId = UUID.randomUUID() // each sent report gets its own UUID
-                if (receiver.transport == null) {
+                val retryItems = retryToken?.items
+                val transport = getTransport(receiver.transportType)
+                if (transport == null) {
                     actionHistory.setActionType(TaskAction.send_warning)
-                    actionHistory.trackActionResult("Not sending $inputReportId to $serviceName: No transports defined")
+                    actionHistory
+                        .trackActionResult("Not sending $inputReportId to $serviceName: Unsupported transport type")
```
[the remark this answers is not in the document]
adegolier: Left it in because I'd wanted to make sure we had some kind of alert in the case of someone adding a transport type, but not adding a corresponding transport for it. I didn't know I could make it just not compile with a sealed class, so changed to that per your suggestion.

prime-router/src/main/kotlin/azure/SendFunction.kt
```diff
@@ -154,16 +150,18 @@ class SendFunction(
...
             is SFTPTransportType -> workflowEngine.sftpTransport
             is BlobStoreTransportType -> workflowEngine.blobStoreTransport
             is AS2TransportType -> workflowEngine.as2Transport
             is SoapTransportType -> workflowEngine.soapTransport
             is GAENTransportType -> workflowEngine.gaenTransport
             is RESTTransportType -> workflowEngine.restTransport
-            is NullTransportType -> NullTransport()
-            else -> null
+            is NullTransportType -> workflowEngine.nullTransport
+            is EmailTransportType -> workflowEngine.emailTransport
```
mkalish, line 164: Nice catch!

mkalish (approved):
Changes look great to me, I didn't pull it locally, so relying on Josh's testing.

Awesome stuff on making it a `sealed` class, this makes it way more error proof moving forward

Issue #16163, opened 2024-10-08
Update the deployment_rollback ALLOWED_USERS list to reflect current resources


emvaldes:
The GitHub Workflow "deployment_rollback" is still hardcoded in the ALLOWED_USERS variable to:
./.github/workflows/deployment_rollback.yml:18:   ALLOWED_USERS: "lucasdze,jeremy-page,JosiahSiegel,snesm,supriyaaddagada"

**Note**: We should/must change this approach and base it on GitHub Groups.
Unless, this is not possible so that next time we do not need to create a change.
Are these hardcoded because of the way this workflow requires these to be setup?

I am recommending to change it to something like:
${{ env.ALLOWED_USERS }}

emvaldes:
This is the PR for this GitHub Issue:
[Modifying the Allowed Users settings to use env.ALLOWED_USERS](https://github.com/CDCgov/prime-reportstream/pull/16335)

emvaldes:
A decision was made to disregard this recommendation and just update the existing list of allowed users to reflect current/active resources.

Previous: "lucasdze,jeremy-page,JosiahSiegel,snesm,supriyaaddagada"
Current: "devopsmatt,emvaldes,scott-aquia,bethbeza

PR #16335, opened 2024-10-23
Modifying the Allowed Users settings to use env.ALLOWED_USERS

emvaldes:
This PR is aimed to fix the existing hardcoded configuration of allowed users for deployment's rollback to be driven by an environment variable which will in time allow for this variable to be modified without affecting or triggering any build process and enhance compliance with active-users.

Note: This is not an exploitable issue or any security concerns but rather an administrative functionality that must be addressed to maintain accuracy on which resources are actively involved in the process considering the dynamics of resources management in this project.

Configuration Steps:
- [x] Created an repository variable ALLOWED_USERS with the existing users to avoid any process disruption.

## Changes
- [x] Requesting feedback from all teams on who should be added to this list (it must be updated A.S.A.P.)
**Current users**: lucasdze,jeremy-page,JosiahSiegel,snesm,supriyaaddagada

### Testing
- No testing is involved in this PR other than notifying all involved parties.

### Process
- [x] Creating the repository variable ALLOWED_USERS
- [x] Updating the repository variable with the correct and authorized users
- [x] Updated the performed changes to the workflow with relevant notes
- [x] Development and DevSecOps team has been notified

## Linked Issues
- This is not a fix to any existing issue.

## To Be Done
* [x] A GitHub issues is already created: [Update the deployment_rollback ALLOWED_USERS list to reflect current resources](https://github.com/CDCgov/prime-reportstream/issues/16163)

## Security-related subjects:
- This PR does not introduce any new dependencies that need vetting and it does not log any sensitive data but it does provide certain level of masking to those users allowed to rollback deployments.


.github/workflows/deployment_rollback.yml
```diff
@@ -15,19 +15,19 @@ jobs:
   DeployToCandidateSlot:
     runs-on: ubuntu-latest
     env:
-      ALLOWED_USERS: "lucasdze,jeremy-page,JosiahSiegel,snesm,supriyaaddagada"
+      ALLOWED_USERS: ${{ env.ALLOWED_USERS }}
```
devopsmatt, line 18: Listing the members in the workflow itself is more transparent than in GH settings, which are not visible to all committers to this repo.

devopsmatt (changes requested):

.github/workflows/deployment_rollback.yml
emvaldes: But doing that causes a problem as any changes will require a build which is not a reasonable operation.
Anyone can print out the value of an repository variable.

emvaldes: These users list is used to identify if the github.actor is part of that list but at no point the information of who this user is gets printed or displayed.

I am proposing to simply use ${{ env.ALLOWED_USERS }} and have that driven as an abstraction that we can maintain without having to initiate any sort of change in the build process.

This is only referenced in this specific workflow:
``` console
$ find . -not -iwholename '*/.git/*' -type f | xargs -I {} egrep -iHn "allowed_users" {} ;
./.github/workflows/deployment_rollback.yml:18:   ALLOWED_USERS: "lucasdze,jeremy-page,JosiahSiegel,snesm,supriyaaddagada"
./.github/workflows/deployment_rollback.yml:22:    if: ${{ contains(env.ALLOWED_USERS, github.actor) }}
./.github/workflows/deployment_rollback.yml:26:    if: ${{ !contains(env.ALLOWED_USERS, github.actor) }}
(edited)
```

emvaldes: Let me know what's your decision in this regard.

devopsmatt: Moving the list of allowed users into GH Settings variables is not transparent enough for people who want to find someone who can run the action and do not have access to repository settings.  
I'd prefer to code it in the same manner it was done before, just update the list of users to the current team members.

devopsmatt (changes requested):

.github/workflows/deployment_rollback.yml
emvaldes: @devopsmatt has pointed out to leave this references in plain text to serve as a reference to others in the development team on who is responsible for the deployments rollbacks.
I will update this list with current/active users and provide references on how to reach out and contact them.

emvaldes: Updated list with current resources:
**ALLOWED_USERS**: "devopsmatt,emvaldes,scott-aquia,bethbeza"

devopsmatt (approved):
LGTM

Issue #16148, opened 2024-10-08
Update Lineage Tracking in FHIRReceiver Function


arnejduranovic:
### User Story

As the owner of the ReportStream Architecture,
I want the FHIRReceiver pipeline step to implement lineage tracking correctly,
so that FHIRReceiver behaves the same as other UP steps in this regard.

### Description/Use Case
> **CONTEXT:** Ticket #16011 was created following the [initial FHIRReceiver implementation](https://github.com/CDCgov/prime-reportstream/pull/15621) and the [QueueMessage followup](https://github.com/CDCgov/prime-reportstream/pull/15928) to rethink how the "submission specific" events should be designed. Upon further investigation, it has been decided Ticket #16011 should be closed and instead the FHIRReceiver should be updated so that a new event is not required. The initial implementation deviated from the normal report and lineage creation/handling and should be brought back in line. Once this is done, the submission specific events can go away and reuse the currently existing Report events.

### Acceptance Criteria
  - [x] FHIRReceiver does not duplicate logic in the FHIR Converter
  - [x] A missing sender is a handled error that results in the message being put on the poison queue

MichaelEsuruoso:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/16148?planning-poker) with Zenhub @adegolier @arnejduranovic @david-navapbc @jack-h-wang @jalbinson @JFisk42 @kant777 @mkalish @thetaurean

PR #16370, opened 2024-10-28
16148: update FHIR reciever logic to be simpler

mkalish:
This PR simplifies the FHIR receiver and removes the few bugs in sender parsing and event tracking.  The simplification can be done because the FHIR Converter handles the same cases more robustly

Test Steps:
1. Turn on the submissions service and reportstream
2. Send a report to submissions
3. Verify it gets picked up by report stream

## Changes
- Drops event handling for errors in favor of putting messages on to the queue directly
- Removes the functionality duplicated with FHIR converter

### Testing
- [x] Tested locally?
- [x] Added tests?

### Process

## Linked Issues
- #16148 

## To Be Done
TBD: but there might be another follow up ticket to address this more efficiently and remove the FHIR Receiver step, but it would require some updates to the how the item lineage works


prime-router/src/main/kotlin/azure/observability/event/ReportStreamEventService.kt
```diff
@@ -296,28 +276,6 @@ class ReportStreamEventService(
         ).send()
     }
 
-    override fun sendSubmissionProcessingError(
```
mkalish, line 299: This didn't align with how the other reports were going to work and likely would have led to confusion as this was inverting parameters compared to other builders

prime-router/src/main/kotlin/fhirengine/engine/FHIRReceiver.kt
```diff
@@ -97,94 +91,31 @@ class FHIRReceiver(
...
-    ): Sender? {
-        val clientId = queueMessage.headers[clientIdHeader]
-        val sender = clientId?.takeIf { it.isNotBlank() }?.let { settings.findSender(it) }
-
-        actionHistory.trackActionParams(queueMessage.headers.toString())
-
-        // Handle case where sender is not found
-        return if (sender == null) {
-            // Send an error event
-            reportEventService.sendSubmissionProcessingError(
```
mkalish, line 148: Rather than emit an event, we're now adding an item to the poison queue which will cause an alert and enables the engagement team to fix the configuration and requeue the message.

prime-router/src/main/kotlin/fhirengine/engine/FHIRReceiver.kt
```diff
@@ -97,94 +91,31 @@ class FHIRReceiver(
...
-                Submission(
-                    queueMessage.reportId.toString(), "Rejected",
-                    queueMessage.blobURL,
-                    "Sender not found matching client_id: ${queueMessage.headers[clientIdHeader]}"
-                )
-            submissionTableService.insertSubmission(submission)
-            null
-        } else {
-            // Handle case where sender is inactive
-            if (sender.customerStatus == CustomerStatus.INACTIVE) {
```
mkalish, line 176: Dropped this since the UP via the ReportFunction silently just drops a message.  I'm of the opinion that if a sender is not suppose to sending reports we should revoke there access rather than have a flag that gets checked so late in the process.

arnejduranovic:
Closing in favor # 16339 which removes the receive step entirely. This was done because it was decided the receive step need not exist long-term.

PR #16397, opened 2024-10-30
Adding test member oid as the extension to all observations in the el…

### LINKED ISSUE #14511: Map CSTE Member OID to member-oid extension during convert-step
### User Story
As a STLT receiving ELR,
I want to only receive test results that are antigen tests for a given condition,
so that my programs have the correct type of tests we are interested in analyzing.

### Description/Use Case
Currently the covid pipeline uses the Covid LIVD table to determine if reports are for antigen tests. This is needed because some receivers only want to receive antigen tests. We need the ability to do this in the UP, but can't rely on the LIVD table since that only supports Covid results and this needs to work for any condition. 

**For the UP, we need to implement a way to route items based on whether they are antigen or another type. This information is reflected by the Member OID the test performed belongs to**

### Dev Notes
Implementation checklist:

1.Add new extension to map test code member oid (`Member OID` column in RS condition mapping table): `https://reportstream.cdc.gov/fhir/StructureDefinition/test-performed-member-oid`

2. Modify convert step to stamp ELR messages with test-performed-member-oid
3. Ensure this data is NOT in Azure Events

With the above implemented, the following routing filter expression should be functional:

```
%resource.code.coding.extension('https://reportstream.cdc.gov/fhir/StructureDefinition/condition-code').value.where(code  in ('840539006')).exists() and %resource.interpretation.coding.code = 'A' and %resource.code.coding.extension('https://reportstream.cdc.gov/fhir/StructureDefinition/test-performed-member-oid').value.where(code  in ('2.16.840.1.113762.1.4.1146.1142')).exists()
```

**Docs** on our condition to code mapping dataset can be found here: `prime-router/docs/design/proposals/0023-condition-to-code-mapping/0023-condition-to-code-mapping.md`

**CSV version of the code to condition mapping table** in our repo that is used to seed a dev DB: `prime-router/metadata/tables/local/observation-mapping.csv`

**Official dataset our condition mapping table is derived from**: https://docs.google.com/spreadsheets/d/1rO_p145xXO1AD76vx8vBqXgoQUnLqfc8/edit?gid=739612351#gid=739612351

### Acceptance Criteria

- [x] `test-performed-member-oid` stamped on Observation resources in ELR messages
- [x] Results can be filtered out based on test type (member oid) as determined by CSTE value sets
- [x] The process to determine results are antigen is documented (@victor-chaparro agreed engagement will do this AC in follow up)
- [x] **NEW:** The current path of condition code must not change so as to not break existing filters when deployed
- [x] Azure Events shall not change as to not break our current dashboard

kant777:
…r messages

This PR  adds the member OID as the extension to all observations in the elr messages.

Test Steps:

image

## Changes
-  The ConditionStamper will now handle both condition lookups and Member OID stamping based on the logic and constants from ObservationMappingConstants

### Testing
- [x] Tested locally? yes
- [x] Ran `./prime test` or `./gradlew testSmoke` against local Docker ReportStream container? yes
- [x] Added tests? yes

## Linked Issues
- Fixes #14511

### DIFF
--- a/prime-router/src/test/kotlin/fhirengine/azure/FHIRReceiverFilterIntegrationTests.kt
+++ b/prime-router/src/test/kotlin/fhirengine/azure/FHIRReceiverFilterIntegrationTests.kt
@@ -25,8 +25,6 @@ import gov.cdc.prime.router.Topic
 import gov.cdc.prime.router.azure.ActionHistory
 import gov.cdc.prime.router.azure.BlobAccess
 import gov.cdc.prime.router.azure.ConditionStamper
-import gov.cdc.prime.router.azure.ConditionStamper.Companion.CONDITION_CODE_EXTENSION_URL
-import gov.cdc.prime.router.azure.ConditionStamper.Companion.MEMBER_OID_EXTENSION_URL
 import gov.cdc.prime.router.azure.DatabaseLookupTableAccess
 import gov.cdc.prime.router.azure.Event
 import gov.cdc.prime.router.azure.LookupTableConditionMapper
--- a/prime-router/src/test/kotlin/fhirengine/engine/FhirConverterTests.kt
+++ b/prime-router/src/test/kotlin/fhirengine/engine/FhirConverterTests.kt
@@ -507,7 +507,7 @@ class FhirConverterTests {
             assertEquals("OID12345", (memberOidExtension!!.value as StringType).value)
         }
     }
-    
+
     @Test
     fun `test fully unmapped condition code stamping logs errors`() {
         val fhirData = File(BATCH_VALID_DATA_URL).readText()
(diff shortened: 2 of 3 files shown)


david-navapbc (changes requested):
I don't see where AC  "The process to determine results are antigen is documented" is met. 

I also don't see any new tests. 

Please address the above and I'll take another look. 

ty!

prime-router/src/main/kotlin/azure/ConditionMapper.kt
```diff
@@ -34,16 +42,32 @@ class LookupTableConditionMapper(metadata: Metadata) : IConditionMapper {
...
+                    acc[testCode] = memberOid
+                }
+                acc
+            }
+    }
 }
 
 class ConditionStamper(private val conditionMapper: IConditionMapper) {
     companion object {
         const val conditionCodeExtensionURL = "https://reportstream.cdc.gov/fhir/StructureDefinition/condition-code"
```
arnejduranovic, line 62: We can fix the IntelliJ warning on this var name while we're here?

arnejduranovic (commented):
Changes look good, but I agree with the need for tests. At a minimum, an integration test similar to: `should send valid FHIR report filtered by mapped condition filter`. This would also show what a FHIRExpression filter would look like.

PR #16446, opened 2024-11-04
Experience/16445-Update-gitleaks-config-allowlist

### LINKED ISSUE #16445: Update gitleaks-config allowlist
It appears our gitleaks config has some files, including frontend, under a `files` allowlist that may no longer be used (as I can only find reference to the `paths` list that we also have). We need to update this config so that those files are properly allowed (including our yarn lock file which always throws a false positive). The regexes for the paths in that list also need to be properly escaped for special symbols (`/` and `.`).

- [ ] Relevant paths from `files` list moved to `paths`
- [ ] `paths` list properly escaped

jpandersen87:
Fixes #16445 

This PR updates the `gitleaks-config.toml` file.
- update allowlist `paths` list with currently relevant entries in the `files` list (which does not seem to be used)
- escapes the paths properly and uses beginning-of-line and end-of-line symbols where appropriate
- comments out echo of non-existent log file path in error output


devopsmatt (approved):
LGTM

scott-aquia (commented):
I have reviewed these changes from a security perspective

PR #16550, opened 2024-11-15
Platform/bill/16144

wcutshall:
This PR contains coding changes for user story 16144.

Test Steps:
1. POST a message through Postman (or equivalent) and observe the log files.

## Changes
- Modified FHIRTranslator and associated tests.
  - Added new ITEM_TRANSLATE azure event in translate function.
  -  Check for matching file digest SHAs.

### Testing
- [x] Tested locally?
- [x] Added tests?

### Process
- [ ] Are there licensing issues with any new dependencies introduced?  **N/A**
- [ ] Updated the release notes? **N/A**
- [ ] Database changes are submitted as a separate PR? **N/A**
- [ ] DevOps team has been notified if PR requires ops support?  **N/A**

## Linked Issues
- N/A

## To Be Done
- N/A

- Does this PR introduce new endpoints? **No**
- Does this PR include changes in authentication and/or authorization of existing endpoints? **No**
- Does this change introduce new dependencies that need vetting? **No**
- Does this change require changes to our infrastructure?  **No**
- Does logging contain sensitive data?  **No**
- Does this PR include or remove any sensitive information itself?  **No**


prime-router/src/main/kotlin/fhirengine/engine/FHIRTranslator.kt
```diff
@@ -114,6 +118,10 @@ class FHIRTranslator(
         logger.trace("Preparing to send original message")
         val originalReport = reportService.getRootReport(message.reportId)
         val bodyBytes = BlobAccess.downloadBlobAsByteArray(originalReport.bodyUrl)
+        val localDigest = BlobUtils.digestToString(sha256Digest(bodyBytes))
+        check(message.digest == localDigest) {
```
arnejduranovic, line 122: This seems wrong to me. The `digest` value in the message I believe would be the digest of the blob associated with the passed in report (so the report that is downloaded in `sendTranslated`). Here, you are downloading the ROOT report, which would have a different digest.

prime-router/src/main/kotlin/fhirengine/engine/FHIRTranslator.kt
```diff
@@ -163,6 +171,18 @@ class FHIRTranslator(
             topic = message.topic
         )
 
+        val builder = ReportStreamItemEventBuilder(
```
arnejduranovic, line 174: We should prefer the style that is shown in FHIRDestinationFilter where we pass in the builder directly to `sendItemEvent` (so we remain consistent):

```
reportEventService.sendItemEvent(
            eventName = ReportStreamEventName.ITEM_ROUTED,
            childReport = report,
            pipelineStepName = TaskAction.destination_filter
        ) {
            parentReportId(queueMessage.reportId)
            params(
                mapOf(
                ReportStreamEventProperties.RECEIVER_NAME to receiver.fullName,
                ReportStreamEventProperties.BUNDLE_DIGEST
                    to bundleDigestExtractor.generateDigest(bundle)
            )
            )
            trackingId(bundle)
        }
```

prime-router/src/test/kotlin/fhirengine/engine/FhirTranslatorTests.kt
```diff
@@ -106,10 +112,11 @@ class FhirTranslatorTests {
         val actionLogger = mockk<ActionLogger>()
         val engine = makeFhirEngine()
 
+        val reportId = UUID.randomUUID()
         val message =
             spyk(
                 FhirTranslateQueueMessage(
-                    UUID.randomUUID(),
+                    reportId,
```
arnejduranovic, line 119: I think what happened here is we coded a bug and then got the tests to pass with the bug! This is one of the problems with our "unit tests". They require so much mocking that at some point it becomes hard to understand what we are even testing and if its correct. Thus, a lot of our Universal Pipeline steps have integration tests that do not mock and its easier to test things like this.

That said, my recommendation is to update the tests in `FHIRTranslatorIntegrationTests.kt` in addition to these.

arnejduranovic (changes requested):

prime-router/src/test/kotlin/fhirengine/engine/FhirTranslatorTests.kt
wcutshall: Are you saying that changing it to reportId from random UUID is the correct way to go?

prime-router/src/main/kotlin/fhirengine/engine/FHIRTranslator.kt
wcutshall: So, if I'm understanding correctly, we should be comparing message.digest with blobInfo.digest (after the file is uploaded)?

arnejduranovic: `message.digest` is already used in `sendTranslated` and that's the only place it should be used. Here, you are downloading the root report, so you need to to compare the downloaded root report to the root report's digest (which is stored in the DB report_file table)

wcutshall: Latest changes should address this.

prime-router/src/main/kotlin/fhirengine/engine/FHIRTranslator.kt
```diff
@@ -114,6 +121,11 @@ class FHIRTranslator(
         logger.trace("Preparing to send original message")
         val originalReport = reportService.getRootReport(message.reportId)
         val bodyBytes = BlobAccess.downloadBlobAsByteArray(originalReport.bodyUrl)
```
arnejduranovic, line 123: nit: Have we considered updating this to use `BlobAccess.downloadBlob(...)` which will do the digest check for us?

prime-router/src/main/kotlin/fhirengine/engine/FHIRTranslator.kt
wcutshall: Should be in there now.

prime-router/src/main/kotlin/fhirengine/engine/FHIRTranslator.kt
wcutshall: Made the change.

prime-router/src/test/kotlin/fhirengine/azure/FHIRTranslatorIntegrationTests.kt
```diff
@@ -191,6 +196,20 @@ class FHIRTranslatorIntegrationTests : Logging {
...
+        assertThat(
+            azureEventService
+                .reportStreamEvents[ReportStreamEventName.ITEM_TRANSFORMED]!!.first()
+        ).isInstanceOf<ReportStreamItemEvent>()
+        val event = azureEventService
+            .reportStreamEvents[ReportStreamEventName.ITEM_TRANSFORMED]!!.first() as ReportStreamItemEvent
+        assertThat(event.params[ReportStreamEventProperties.ORIGINAL_FORMAT]).isEqualTo("FHIR")
+        assertThat(event.params[ReportStreamEventProperties.TARGET_FORMAT]).isEqualTo("HL7")
+        assertThat(event.params[ReportStreamEventProperties.RECEIVER_NAME]).isEqualTo("phd.x")
+        val enrichments = event.params[ReportStreamEventProperties.ENRICHMENTS] as List<*>
```
arnejduranovic, line 210: nit: should be able to verify the actual value here, not just that it exists?

wcutshall: De-nit'ed with latest code push.

arnejduranovic (approved):
great work!

PR #16595, opened 2024-11-19
remove usage of downloadjs

### LINKED ISSUE #16309: [Bug]: Downloading a HL7 file from Daily Data garbles the 'µ' character in the content of the file
### Contact Details

cdc-ti@flexion.us

### What happened?

When downloading a HL7 file that contains 'µ', the character gets garbled to �. We have confirmed that downloading the same file directly from the Azure blob storage encodes the file correctly, so it only happens when downloading from Daily Data.

We look into the issue and found the RS frontend is using the [download](https://github.com/rndme/download) library in Daily Data, which has this open issue since 2017: [1.4.6+ broke downloading files with non-ASCII charachers](https://github.com/rndme/download/issues/56). We've also confirmed the HTTP response that gets the content of the file before downloading, has the 'µ' shown correctly, so everything seems to point to this bug in the download library.

We're sending the 'µ' character in a `ORU` message as part of the `µmol/L` unit in `OBX-6`

### Version

1.0.2 (Default)

### What browsers are you seeing the problem on?

Chrome

### Relevant log output

```shell
No logs to show
```

## Acceptance Criteria
- [ ] Try testing first by calling the API directly to see if the same response/error happens. Note that response in the ticket
- [ ] Download functionality works as expected without error
- [ ] Point this ticket with proper amount of work estimate before closing

etanb:
This PR removes our usage of `downloadjs` which is a long-ago deprecated library that can be replicated by native JS functions.

Fixes #16309


jpandersen87 (approved):
🎉

PR #16689, opened 2024-12-03
reset pagination upon filtering

### LINKED ISSUE #16385: [BUG] Daily data filter don't work after navigating to a different page
## Problem statement

When loading the daily data page and navigating to a different page to look at additional results filtering by dates does not work.

## What you need to know

Steps to reproduce:

- Load daily data page with ignore data.
- Select FULL_ELR receiver.
- Enter 10/19/2024 on From
- Enter 10/22/2024 on To
- Click apply and you get back some results.
- Click reset
- Select FULL_ELR receiver.
- Go to page 4.
- Enter 10/19/2024 on From
- Enter 10/22/2024 on To
- Click apply and you get no results.

## Acceptance criteria

- [x] Filtering on Daily data works
- [x] Point this ticket with time/effort it took before closing (1-2 = Low, 3-5 = medium, 8-13 = high)

etanb:
This PR makes sure we reset pagination to page 1 upon applying filters. IE, if a user applied a filter but was on page 4 at the time, the API would use the correctly filtered date/time BUT would also apply that page=4 pagination.

fixes #16385


victor-chaparro (approved):
Reviewed with Etan and it looks good!

Issue #16143, opened 2024-10-07
Implement Convert Step Updates


arnejduranovic:
### User Story

As the owner of the RS Architecture,
I want the Convert step to implement the latest SRD,
so that it can function as intended.

### Description/Use Case

The  [ReceiverEnrichment SRD](https://cdc.sharepoint.com/:w:/r/teams/ReportStream/_layouts/15/doc2.aspx?sourcedoc=%7BCD15C571-D563-4135-B293-6FFAA77826AE%7D&file=SRD-ReceiverEnrichment.docx&action=default&mobileredirect=true) outlines new requirements for the convert step in the section titled "Convert Step Software Requirements". These new requirements should be implemented and the existing ones should be verified.

### Acceptance Criteria

- [ ] Implemented Convert Step requirements outlined in [ReceiverEnrichment SRD](https://cdc.sharepoint.com/:w:/r/teams/ReportStream/_layouts/15/doc2.aspx?sourcedoc=%7BCD15C571-D563-4135-B293-6FFAA77826AE%7D&file=SRD-ReceiverEnrichment.docx&action=default&mobileredirect=true)
- [ ] Update convert step to only stamp extensions for the appropriate message type (ex: condition codes only should be stamped on ELR messages, not test orders)

MichaelEsuruoso:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/16143?planning-poker) with Zenhub @adegolier @arnejduranovic @david-navapbc @jack-h-wang @jalbinson @JFisk42 @kant777 @mkalish @thetaurean

MichaelEsuruoso:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/16143?planning-poker) with Zenhub @wcutshall

MichaelEsuruoso:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/16143?planning-poker) with Zenhub @adegolier

MichaelEsuruoso:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/16143?planning-poker) with Zenhub @JFisk42

PR #16710, opened 2024-12-04
[16143] Implement Convert Step Updates

wcutshall:
This PR contains coding changes for user story [16144](https://github.com/CDCgov/prime-reportstream/issues/16143).

Test Steps:
1. Debug unit tests or integration tests locally with breakpoints set in affected code to verify functionality.
2. POST a message through Postman (or equivalent) locally and observe the log files.
3. Connect dev env to testing app insights and verify events show up properly.

## Changes
- Modified FHIRConverter and associated tests to:
  - Only add extensions if message is an ELR.
  - Send an ITEM_TRANSFORMED Azure event.

### Testing
- [x] Tested locally?
- [x] Added tests?

### Process
No changes.

### DIFF
--- a/prime-router/src/main/kotlin/fhirengine/engine/FHIRConverter.kt
+++ b/prime-router/src/main/kotlin/fhirengine/engine/FHIRConverter.kt
@@ -114,6 +114,7 @@ class FHIRConverter(
         companion object {
 
             private val clientIdHeader = "client_id"
+            private val payloadNameHeader = "payloadname"
 
             /**
              * Converts a [FhirConvertQueueMessage] into the input to the convert processing
@@ -160,6 +161,7 @@ class FHIRConverter(
                 val blobSubFolderName = message.blobSubFolderName
 
                 val clientId = message.headers[clientIdHeader]
+                val payloadName = message.headers[payloadNameHeader]
                 val sender = clientId?.takeIf { it.isNotBlank() }?.let { settings.findSender(it) }
                 if (sender == null) {
                     throw SubmissionSenderNotFound(clientId ?: "", reportId, blobUrl)
@@ -181,7 +183,8 @@ class FHIRConverter(
                 // is properly recorded in the report file table with the correct sender
                 actionHistory.trackExternalInputReport(
                     report,
-                    BlobAccess.BlobInfo(format, blobUrl, blobDigest.toByteArray())
+                    BlobAccess.BlobInfo(format, blobUrl, blobDigest.toByteArray()),
+                    payloadName
                 )
                 actionHistory.trackActionSenderInfo(sender.fullName)
 
--- a/prime-router/src/test/kotlin/common/UniversalPipelineTestUtils.kt
+++ b/prime-router/src/test/kotlin/common/UniversalPipelineTestUtils.kt
@@ -441,7 +441,7 @@ object UniversalPipelineTestUtils {
         fileName: String = "mr_fhir_face.fhir",
     ): Report {
         val blobUrl = BlobAccess.uploadBlob(
-            "${TaskAction.receive.literal}/$fileName",
+            "${previousAction.literal}/$fileName",
             reportContents.toByteArray(),
             getBlobContainerMetadata(azuriteContainer)
         )
--- a/prime-router/src/main/kotlin/history/db/ReportGraph.kt
+++ b/prime-router/src/main/kotlin/history/db/ReportGraph.kt
@@ -123,11 +123,10 @@ class ReportGraph(
     }
 
     /**
-     * Recursively goes up the report_linage table from any report until it reaches
-     * a report with an action type of "receive" (the root report)
+     * Recursively goes up the report_lineage table from any report until it reaches
+     * a report that does not appear in report_lineage as a child report (the root report)
      *
-     * This will return null if no report with action type "receive" is present or if
-     * the root is passed in
+     * This will return null if the root is passed in
      */
     fun getRootReport(childReportId: UUID): ReportFile? {
         return db.transactReturning { txn ->
@@ -172,19 +171,19 @@ class ReportGraph(
             .from(cte)
             .join(REPORT_FILE)
             .on(REPORT_FILE.REPORT_ID.eq(ItemGraphTable.ITEM_GRAPH.PARENT_REPORT_ID))
-            .join(ACTION)
-            .on(ACTION.ACTION_ID.eq(REPORT_FILE.ACTION_ID))
-            .where(ACTION.ACTION_NAME.eq(TaskAction.receive))
+            .leftJoin(REPORT_LINEAGE)
+            .on(REPORT_FILE.REPORT_ID.eq(REPORT_LINEAGE.CHILD_REPORT_ID))
+            .where(REPORT_LINEAGE.PARENT_REPORT_ID.isNull())
+            .orderBy(REPORT_FILE.ACTION_ID.asc())
             .fetchOneInto(Item::class.java)
         return rootItem
     }
 
     /**
-     * Recursively goes up the report_linage table from any report until it reaches
-     * all reports with an action type of "receive" (the root report)
+     * Recursively goes up the report_lineage table from any report until it reaches
+     * all reports that do not appear in report_lineage as a child report (the root report)
      *
-     * This will return null if no report with action type "receive" is present or if
-     * the root is passed in
+     * This will return null if the root is passed in
      *
      * If the passed in report ID has multiple root reports, they will all be returned
      */
@@ -476,9 +475,10 @@ class ReportGraph(
         .from(cte)
         .join(REPORT_FILE)
         .on(REPORT_FILE.REPORT_ID.eq(cte.field(0, UUID::class.java)))
-        .join(ACTION)
-        .on(ACTION.ACTION_ID.eq(REPORT_FILE.ACTION_ID))
-        .where(ACTION.ACTION_NAME.eq(TaskAction.receive))
+        .leftJoin(REPORT_LINEAGE)
+        .on(REPORT_FILE.REPORT_ID.eq(REPORT_LINEAGE.CHILD_REPORT_ID))
+        .where(REPORT_LINEAGE.PARENT_REPORT_ID.isNull())
+        .orderBy(REPORT_FILE.ACTION_ID.asc())
 
     /**
      * Accepts a list of ids and walks down the report lineage graph
--- a/prime-router/src/test/kotlin/fhirengine/azure/FHIRTranslatorIntegrationTests.kt
+++ b/prime-router/src/test/kotlin/fhirengine/azure/FHIRTranslatorIntegrationTests.kt
@@ -640,7 +640,7 @@ class FHIRTranslatorIntegrationTests : Logging {
     }
 
     @Test
-    fun `successfully translate HL7 for FHIR receiver when isSendOriginal is true`() {
+    fun `successfully translate HL7 for FHIR receiver when isSendOriginal is true from receive step`() {
         // set up
         val receiverSetupData = listOf(
             UniversalPipelineTestUtils.ReceiverSetupData(
@@ -720,7 +720,7 @@ class FHIRTranslatorIntegrationTests : Logging {
     }
 
     @Test
-    fun `successfully translate for FHIR receiver when isSendOriginal is true`() {
+    fun `successfully translate for FHIR receiver when isSendOriginal is true from receive step`() {
         // set up
         val receiverSetupData = listOf(
             UniversalPipelineTestUtils.ReceiverSetupData(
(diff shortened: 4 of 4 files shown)


prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
```diff
@@ -116,6 +117,26 @@ fun Bundle.addProvenanceReference() {
     }
 }
 
+/**
+ * Return true if Bundle contains an ELR in the MessageHeader.
+ *
+ * @return true if has a MesssageHeader that contains an R01 or ORU_R01, otherwise false.
+ */
+fun Bundle.isElr(): Boolean {
```
arnejduranovic, line 125: RS currently receives test orders and results. Instead of `isElr`, perhaps a method like `getMessageType` that returns an enum value like `LAB_TEST_RESULT` or `LAB_TEST_ORDER` or `UNKOWN` would be more reusable? Then, I would say let's stamp the message if we get back either `LAB_TEST_RESULT` or `UNKOWN` just to be safe?

Implementing detection for `LAB_TEST_ORDER` is optional at this point as there is no need to distinguish this as of yet.

prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
```diff
@@ -116,6 +117,26 @@ fun Bundle.addProvenanceReference() {
...
+    var isElr = false
+    if (this.type == Bundle.BundleType.MESSAGE && this.entry.isNotEmpty()) {
+        // By rule, the first entry must be a MessageHeader
```
arnejduranovic, line 128: Can you link to documentation that says this? This makes me uneasy. See fun `Bundle.getDiagnosticReportNoObservations(): List<Base> {` below to see how we can utilize FHIRPath here for a better solution imo.

wcutshall: Look at the documentation around "entry" for the [Bundle](https://www.hl7.org/fhir/bundle.html) resource

_This repeating element order: For bundles of type 'document' and 'message', the first resource is special (must be Composition or MessageHeader respectively). For all bundles, the meaning of the order of entries depends on the bundle type._

prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
wcutshall: Added some code for this with only LAB_RESULT or UNKNOWN as options.  Future features will need to add additional values to the enum.

prime-router/src/test/kotlin/fhirengine/engine/FhirConverterTests.kt
```diff
@@ -544,6 +646,12 @@ class FhirConverterTests {
         }
     }
 
+//    @Test
```
arnejduranovic, line 649: Please delete

arnejduranovic (changes requested):

prime-router/src/main/kotlin/fhirengine/engine/FHIRConverter.kt
```diff
@@ -508,24 +514,32 @@ class FHIRConverter(
                 }
                 // 'stamp' observations with their condition code
                 if (item.bundle != null) {
+                    val isElr = if (item.bundle!!.getRSMessageType() == RSMessageType.LAB_RESULT) true else false
```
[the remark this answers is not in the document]
wcutshall: Yup.

prime-router/src/main/kotlin/fhirengine/engine/FHIRConverter.kt
```diff
@@ -508,24 +514,32 @@ class FHIRConverter(
...
-                            val logger = actionLogger.getItemLogger(item.index + 1, observation.id)
-                            if (result.failures.isEmpty()) {
-                                logger.warn(UnmappableConditionMessage())
-                            } else {
-                                logger.warn(
-                                    result.failures.map {
-                                    UnmappableConditionMessage(
-                                        it.failures.map { it.code },
-                                        it.source
+                        // Only do this if it is an ELR item.
```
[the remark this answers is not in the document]
wcutshall: Cleaned it up.

prime-router/src/test/kotlin/fhirengine/engine/FhirConverterTests.kt
wcutshall: Removed.

prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
```diff
@@ -116,6 +118,38 @@ fun Bundle.addProvenanceReference() {
     }
 }
 
+/**
+ * Return true if Bundle contains an ELR in the MessageHeader.
+ *
+ * @return true if has a MesssageHeader that contains an R01 or ORU_R01, otherwise false.
+ */
+fun Bundle.isElr(): Boolean {
```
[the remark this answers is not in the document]
wcutshall: My thought on this is that, if we intend to add several potential groupings (i.e., enums in RSMessageType), then getRSMessageType would become quite large.  Keeping just meaningful function names related to the message grouping type makes the switch statement clearer in my opinion.  Additionally, there is an old programming rule of thumb that goes something like "no subroutine should be longer than one screen in length".  This, obviously, is rather vague but speaks to keeping code segments small to increase "understandability".  You'll often see this argument backed by references like [The Magic Number Sever, Plus or Minus Two](https://en.wikipedia.org/wiki/The_Magical_Number_Seven,_Plus_or_Minus_Two).

prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
```diff
@@ -116,6 +118,38 @@ fun Bundle.addProvenanceReference() {
...
+}
+
+/**
+ * Return RSMessageType based on grouping logic.
+ *
+ * @return RSMessageType of this Bundle.
+ */
+fun Bundle.getRSMessageType(): RSMessageType {
+    when {
+        isElr() -> return RSMessageType.LAB_RESULT
```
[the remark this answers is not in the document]
wcutshall: See my comment about your suggestion above.

prime-router/src/test/kotlin/common/UniversalPipelineTestUtils.kt
```diff
@@ -43,14 +43,27 @@ const val validFHIRRecord1 =
...
...vation","id":"d683b42a-bf50-45e8-9fce-6c0531994f09","status":"final","code":{"coding":[{"system":"http://loinc.org","code":"80382-5"}],"text":"Flu A"},"subject":{"reference":"Patient/9473889b-b2b9-45ac-a8d8-191f27132912"},"performer":[{"reference":"Organization/1a0139b9-fc23-450b-9b6c-cd081e5cea9d"}],"valueCodeableConcept":{"coding":[{"system":"http://snomed.info/sct","code":"260373001","display":"Detected"}]},"interpretation":[{"coding":[{"system":"http://terminology.hl7.org/CodeSystem/v2-0078","code":"A","display":"Abnormal"}]}],"method":{"extension":[{"url":"https://reportstream.cdc.gov/fhir/StructureDefinition/testkit-name-id","valueCoding":{"code":"BD Veritor System for Rapid Detection of SARS-CoV-2 & Flu A+B_Becton, Dickinson and Company (BD)"}},{"url":"https://reportstream.cdc.gov/fhir/StructureDefinition/equipment-uid","valueCoding":{"code":"BD Veritor System for Rapid Detection of SARS-CoV-2 & Flu A+B_Becton, Dickinson and Company (BD)"}}],"coding":[{"display":"BD Veritor System for Rapid Detection of SARS-CoV-2 & Flu A+B*"}]},"specimen":{"reference":"Specimen/52a582e4-d389-42d0-b738-bee51cf5244d"},"device":{"reference":"Device/78dc4d98-2958-43a3-a445-76ceef8c0698"}}}]}"""
```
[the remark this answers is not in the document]
wcutshall: This was something that I meant to ask before I committed the code.  We never reached a definitive conclusion about whether or not we could expect the event.Coding in the MessageHeader.  If I just add it to the existing variables then it will break other tests.  The 1a versions were added to support my tests without breaking the other existing tests.

prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
```diff
@@ -116,6 +118,38 @@ fun Bundle.addProvenanceReference() {
...
+ * @return true if has a MesssageHeader that contains an R01 or ORU_R01, otherwise false.
+ */
+fun Bundle.isElr(): Boolean {
+    var isElr = false
+    if (this.type == Bundle.BundleType.MESSAGE && this.entry.isNotEmpty()) {
+        // By rule, the first entry must be a MessageHeader
+        val resource = this.entry[0].resource
+        if (resource is MessageHeader) {
+            val event = resource.event
+            if (event is Coding && ((event.code == "R01") || (event.code == "ORU_R01"))) {
```
jalbinson, line 133: are there other codes that could be ELR in the future? Throwing those codes into a set to check against would make this a bit more readable.

prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
jalbinson: I played around with FHIRPath a bit and I think it makes it a bit more readable and flat. I've also learned working here that specs are often not followed as closely as we anticipate.

```kotlin
val code = FhirPathUtils.evaluate(
        null,
        this,
        this,
        "Bundle.entry.resource.ofType(MessageHeader).event.code"
    )
        .filterIsInstance<CodeType>()
        .firstOrNull()
        ?.code
```

prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
```diff
@@ -116,6 +118,38 @@ fun Bundle.addProvenanceReference() {
...
+    return isElr
+}
+
+/**
+ * Return RSMessageType based on grouping logic.
+ *
+ * @return RSMessageType of this Bundle.
+ */
+fun Bundle.getRSMessageType(): RSMessageType {
+    when {
```
jalbinson, line 147: kotlin tip: `when` expressions return values so you can omit your returns in each branch and return the whole expression.

```kotlin
return when {
    isElr() -> RSMessageType.LAB_RESULT
    else -> RSMessageType.UNKNOWN
}
```

prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
wcutshall: That's unknown at this point.  I suspect not right now, though.

prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
wcutshall: The Gradle build fails when your suggested change is made.

```
> Task :prime-router:compileKotlin FAILED
e: file:///Users/bill/projects/report-stream/prime-reportstream/prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt:151:1 A 'return' expression required in a function with a block body ('{...}')
```

wcutshall: Oops my mistake, apologies.  Made the change as requested.

prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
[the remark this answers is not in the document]
arnejduranovic: Sure, I just think in practice in this case this won't be longer than a page etc and the downside is you'll be copy pasting this function most likely and just changing the if check in it (violating the DRY principal). That said, there is no repeated code right now, so I'll leave this as a "nit" and we can refactor if needed when we expand if you'd like.

prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
arnejduranovic: Unlikely

prime-router/src/test/kotlin/common/UniversalPipelineTestUtils.kt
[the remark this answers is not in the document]
arnejduranovic: We should try to not keep duplicating these test messages for specific cases and instead prefer to fix existing tests if it makes to do so.

> We never reached a definitive conclusion about whether or not we could expect the event.Coding in the MessageHeader.

Can you remind me what the issue was here? AFAIK:

- SR is only FHIR sender right now, and they are populating it
- Our HL7v2->FHIR mappings populate it? (did you confirm this?)

Given the above are true, I think we should expect that to be populated in MessageHeader?

prime-router/src/main/kotlin/fhirengine/engine/FHIRConverter.kt
```diff
@@ -508,24 +514,32 @@ class FHIRConverter(
...
+                                                it.source
+                                            )
+                                        }
                                     )
                                 }
-                                )
                             }
                         }
                     }
+                    logger.info(
```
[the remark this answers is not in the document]
wcutshall: Removed.

prime-router/src/test/kotlin/common/UniversalPipelineTestUtils.kt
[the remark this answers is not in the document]
wcutshall: You are correct in that the example from SimpleReport does contain ORU_R01:

```
{
      "fullUrl": "MessageHeader/a9922147-0ae7-4b2c-bb50-2d373a7efb18",
      "resource": {
        "resourceType": "MessageHeader",
        "id": "a9922147-0ae7-4b2c-bb50-2d373a7efb18",
        "meta": {
          "tag": [
            {
              "system": "http://terminology.hl7.org/CodeSystem/v2-0103",
              "code": "P",
              "display": "Production"
            }
          ]
        },
        "eventCoding": {
          "system": "http://terminology.hl7.org/CodeSystem/v2-0003",
          "code": "ORU_R01",
          "display": "ORU/ACK - Unsolicited transmission of an observation message"
        },
        "destination": [
          {
            "name": "PRIME ReportStream",
            "endpoint": "https://prime.cdc.gov/api/reports?option=SkipInvalidItems"
          }
        ],
        "sender": {
          "reference": "Organization/719ec8ad-cf59-405a-9832-c4065945c130"
        },
        "source": {
          "extension": [
            {
              "url": "https://reportstream.cdc.gov/fhir/StructureDefinition/software-binary-id",
              "valueString": "3fe4aeb"
            },
            {
              "url": "https://reportstream.cdc.gov/fhir/StructureDefinition/software-install-date",
              "valueInstant": "2024-11-14T16:38:42Z"
            },
            {
              "url": "https://reportstream.cdc.gov/fhir/StructureDefinition/software-vendor-org",
              "valueReference": {
                "reference": "Organization/07640c5d-87cd-488b-9343-a226c5166539"
              }
            }
          ],
          "software": "PRIME SimpleReport",
          "version": "3fe4aeb",
          "endpoint": "https://simplereport.gov/"
        },
        "focus": [
          {
            "reference": "Provenance/1810d743-5440-470c-8afc-528b9b35784c"
          },
          {
            "reference": "DiagnosticReport/2b99e03c-b797-497e-8f93-0d71354b22d4"
          }
        ]
      }
    },
```

And our HL7->FHIR mappings do seem to bring forward the HL7 trigger type in the MessageHeader eventCoding.  The issue is, since it is not strictly a required field, can/should we rely on it being populated for all senders?  Do we have the ability to compel them to do so?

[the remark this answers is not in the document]
arnejduranovic: We do have the ability to compel them to do so (via FHIR profile enforcement, which is not implemented yet) and also during the onboarding process. Do a lot of tests fail if you update the existing bundles in here to include the eventCoding? and why?

prime-router/src/main/kotlin/fhirengine/utils/FHIRBundleHelpers.kt
```diff
@@ -116,6 +118,40 @@ fun Bundle.addProvenanceReference() {
...
+    val code = FhirPathUtils.evaluate(
+        null,
+        this,
+        this,
+        "Bundle.entry.resource.ofType(MessageHeader).event.code"
+    )
+        .filterIsInstance<CodeType>()
+        .firstOrNull()
+        ?.code
+    if (code != null && ((code == "R01") || (code == "ORU_R01"))) {
```
jalbinson, line 137: nit: you can just return this expression rather using the mutable `isElr` variable.

jalbinson: and you can also skip the null check since `null == "RO1` will be false as well

wcutshall: Noice!  Made the change.

arnejduranovic (approved):

PR #16723, opened 2024-12-05
16409 do not use receive step to determine root report

### LINKED ISSUE #16409: Identify and update DB queries to be independent of the Receive step
### User Story

As a developer of ReportStream,
I want to ensure all DB queries are independent of the receive step,
so that the pipeline steps can be more flexible and support greater use cases.

### Description/Use Case

With the addition of the new submission microservice, the first step that will be recorded in the DB is the `Convert` step, instead of the `Receive` step. Because of this, all queries that rely on the `Receive` step as a kind of base-case need to be rewritten to be agnostic of the Receive step. 

Some areas to investigate:
- getRootReport and supporting functions (detailed below in the Dev Notes)
- get original report functionality in the translate step

### Dev Notes

We will need to look to identify all queries/feature that apply. A precursory search lends the following two cases:

```
    fun getRootItem(childReportId: UUID, childIndex: Int, txn: DataAccessTransaction): Item? {
        val cte = itemAncestorGraphCommonTableExpression(
            childReportId,
            childIndex
        )
        val rootItem = DSL.using(txn)
            .withRecursive(cte)
            .select(
                ItemGraphTable.ITEM_GRAPH.asterisk()
            )
            .from(cte)
            .join(REPORT_FILE)
            .on(REPORT_FILE.REPORT_ID.eq(ItemGraphTable.ITEM_GRAPH.PARENT_REPORT_ID))
            .join(ACTION)
            .on(ACTION.ACTION_ID.eq(REPORT_FILE.ACTION_ID))
            .where(ACTION.ACTION_NAME.eq(TaskAction.receive))
            .fetchOneInto(Item::class.java)
        return rootItem
    }
```
```
    private fun rootReportRecords(
        txn: DataAccessTransaction,
        cte: CommonTableExpression<Record1<UUID>>,
    ) = DSL.using(txn)
        .withRecursive(cte)
        .select(REPORT_FILE.asterisk())
        .from(cte)
        .join(REPORT_FILE)
        .on(REPORT_FILE.REPORT_ID.eq(cte.field(0, UUID::class.java)))
        .join(ACTION)
        .on(ACTION.ACTION_ID.eq(REPORT_FILE.ACTION_ID))
        .where(ACTION.ACTION_NAME.eq(TaskAction.receive))
```

The reason `rootReportRecords` can return more than one record is because a report ID that was created in the send step can contain items from multiple submitted reports (thanks to the batch step). My initial thought here is that instead of relying on the WHERE that looks at the TaskAction, we need to consult with the `report_lineage` table and only pick reports where REPORT_FILE.REPORT_ID does not have an entry in that table as a child. This may require a new CTE?

### Acceptance Criteria

- [ ] Queries that rely on the receive step are updated to work for any step
- [ ] Submissions API verified to be working correctly (Submissions page on website)

jack-h-wang:
This PR adjusts the search for root reports to not rely on explicitly searching for the `receive` step report. Additionally, the root report from the submissions endpoint should now include the sender-supplied `payloadname` if present.

Test Steps:
1. The tests in `ReportGraphTest` and `FhirTranslatorIntegrationTests` are sufficient

## Changes
- Do not use `receive` step as indicator of a root report.
  - Instead, a root report is determined by a report ID that does not appear in the `report_lineage` table as a child report.
- Retrieve `payloadname` from submission endpoint message headers when creating root report.

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- Fixes #16409


prime-router/src/main/kotlin/fhirengine/engine/FHIRConverter.kt
```diff
@@ -178,7 +180,8 @@ class FHIRConverter(
                 // is properly recorded in the report file table with the correct sender
                 actionHistory.trackExternalInputReport(
                     report,
-                    BlobAccess.BlobInfo(format, blobUrl, blobDigest.toByteArray())
+                    BlobAccess.BlobInfo(format, blobUrl, blobDigest.toByteArray()),
+                    payloadName
```
jack-h-wang, line 184: This change to carry over `payloadname` from the submissions API is needed to continue supporting the `sendOriginal` function. This adds the information to the root report being created here.

prime-router/src/main/kotlin/history/db/ReportGraph.kt
```diff
@@ -172,19 +171,19 @@ class ReportGraph(
             .from(cte)
             .join(REPORT_FILE)
             .on(REPORT_FILE.REPORT_ID.eq(ItemGraphTable.ITEM_GRAPH.PARENT_REPORT_ID))
-            .join(ACTION)
-            .on(ACTION.ACTION_ID.eq(REPORT_FILE.ACTION_ID))
-            .where(ACTION.ACTION_NAME.eq(TaskAction.receive))
+            .leftJoin(REPORT_LINEAGE)
+            .on(REPORT_FILE.REPORT_ID.eq(REPORT_LINEAGE.CHILD_REPORT_ID))
+            .where(REPORT_LINEAGE.PARENT_REPORT_ID.isNull())
+            .orderBy(REPORT_FILE.ACTION_ID.asc())
```
jack-h-wang, line 177: The previous query was to join the `action` table in order to check if the `action_name` is `receive`. The new query instead filters the list of reports to only include those that do not appear in `report_lineage` as a child. This list is then ordered by `action_id` to ensure the results are output in a consistent order. The join to the `action` table is removed since it is no longer necessary.

prime-router/src/test/kotlin/fhirengine/azure/FHIRTranslatorIntegrationTests.kt
```diff
@@ -798,4 +798,166 @@ class FHIRTranslatorIntegrationTests : Logging {
             assertThat(translatedValue).isEqualTo(reportContents.toByteArray())
         }
     }
+
+    @Test
+    fun `successfully translate HL7 for FHIR receiver when isSendOriginal is true from convert step`() {
```
jack-h-wang, line 803: These tests ensure `sendOriginal` items still function when `convert` is the root report, as the case would be when the submissions API is used.

prime-router/src/main/kotlin/history/db/ReportGraph.kt
JFisk42: Thanks, this explanation was very helpful!

JFisk42 (approved):
Code looks good to the best of my understanding. 
I ran the tests locally and everything is as expected.

Side note: I wanted to use this PR as an excuse to try out the new submission microservice. I was not able to get any reports sent to it because of setup issues. Most likely this is because of the extra steps currently required to get the authorization service running locally. This should no block this PR.

prime-router/src/main/kotlin/fhirengine/engine/FHIRConverter.kt
```diff
@@ -111,6 +111,7 @@ class FHIRConverter(
         companion object {
 
             private val clientIdHeader = "client_id"
+            private val payloadNameHeader = "payloadname"
```
jalbinson, line 114: This doesn't need to be done now, but since in the future these header names will be used across multiple services they should be defined in the shared project. It makes it less prone to errors if we have to change them for some reason in the future.

Maybe you could throw a `TODO` on there with your ticket for submissions service updates.

prime-router/src/main/kotlin/history/db/ReportGraph.kt
jalbinson: Just to clarify, we are now recursively going up the report lineage table until we find a parent report id which is null?

jack-h-wang: The recursive query (`cte`) was not altered, but that is the current and previous behavior. The change here was the in the way the results from the recursive query are filtered.

jalbinson: Got it 👍

jalbinson (approved):

Issue #16755, opened 2024-12-06
Use the ReportStream HL7 ORU_R01 Java class in FHIR->HL7 conversions


JFisk42:
### User Story
As a HL7 receiver
I want to have the maximum amount of data that is transmitted by senders, 
so that the data I receive is accurate and complete.

### Description
Because of how the HAPI library handles data for withdrawn fields, it is possible that subfields of a complex datatype for withdrawn fields to be either incorrectly mapped or dropped entirely. (Ex: OBR-15-1-2; IN1-38-1-2. [OBR-15 is SPS](https://hl7-definition.caristix.com/v2/HL7v2.5.1/DataTypes/SPS) and OBR-15-1 is a CWE; [IN1-38 is CP](https://hl7-definition.caristix.com/v2/HL7v2.5.1/DataTypes/CP) and IN1-38-1 is MO)

Scenario A (incorrectly modified data):
IN1-38 input from sender: `|22.22&USD|`
IN1-38 output to receiver: `|22.22^USD|`
_Note that IN1-38-1-2 has been incorrectly assigned to the position for IN1-38-2._

Scenario B (dropped data):
IN1-38 input from sender: `|22.22&USD^P2|`
IN1-38 output to receiver: `|22.22^P2|`
_Note that IN1-38-1-2 has been entirely dropped._

Currently, our FHIR->HL7 mappings use the HAPI declared Java classes and not our RS "catchall" Java classes. This proposed change would update the ORU_R01 mappings for FHIR->HL7 to avoid any potential clipping of data. Ideally, all FHIR->HL7 transform files should be updated to utilize the RS Java classes.

### Risks/Impacts/Considerations
- At the moment, the only fields that have been concretely identified as affected by this issue are OBR-15 and IN1-38/40/41 (ie: Fields withdrawn in 2.7 and have a complex datatype where the subfield also has a complex datatype). The IN1 fields are not mapped in production for any message types. OBR-15 is only now mapped in production with this weeks' deployment. 
- It should be noted that this ticket only applies to ORU_R01 mappings as none of our STLTs use other message types at the moment. The reason STLTs specifically are being called out in this ticket is that each STLT has its own FHIR->HL7 mapping file where the HAPI Java classes are specified. This issue is currently being fixed for OML/ORM message types with [a ticket currently in progress](https://app.zenhub.com/workspaces/platform-6182b02547c1130010f459db/issues/gh/cdcgov/prime-reportstream/15490).
[...]

Issue #15500, opened 2024-08-05
Create mapping of HL7 IN1 -> FHIR Coverage -> HL7


mkalish:
### User Story
As a user, I would like to have ADT data correctly mapped to FHIR and then back to HL7

### Description/Use Case
ADT maps IN1 to Coverage and we'll need to write that implementation

### Dev Notes
[Mapping inventory](https://docs.google.com/spreadsheets/d/1-8U35_8X9wQTQAchyVAujF4TFrbcjyDeloU2gpHT2_o/edit#gid=0)

### Acceptance Criteria
- [x] IN1 segments in an ADT_* message are mapped to a Coverage according to the inventory and then map into HL7 without any data loss
- [x] If ADT message support is not complete or underway, implement IN1 for OML/ORM.

MichaelEsuruoso:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/15500?planning-poker) with Zenhub @adegolier @arnejduranovic @david-navapbc @jack-h-wang @JFisk42 @kant777 @mkalish @thetaurean

JFisk42:
It was mentioned on the PR for CPExtension that the existing extension mappings for IN1/Coverage.yml will need to be moved to IN1Extension. This did not get completed before code owner review so should be done as part of this ticket. https://github.com/CDCgov/prime-reportstream/pull/16767#discussion_r1881179199

PR #16767, opened 2024-12-10
Add CP to Extension mappings

### LINKED ISSUE #15490: Create mapping for HL7 CP -> FHIR extension -> HL7 CP
### User Story
As a user, I would like to have ADT data correctly mapped to FHIR and then back to HL7

### Description/Use Case
ADT contains the CP datatype but it does not having an entry in the mapping inventory so an extension will need to be created for it 

- IN1.40 is a ST in 2.7.1

### Dev Notes
- See how [MOC](https://github.com/CDCgov/prime-reportstream/blob/93e350a7c23ec0e0263722eae7c2d90f13bf4f19/prime-router/metadata/HL7/catchall/hl7/datatypes/MOC/Extension.yml) was handled as an example
- IN1.40 is an CP

### Acceptance Criteria
- [ ] CP datatypes in an ADT_* message are mapped to an extension and then map into HL7 without any data loss

JFisk42:
This PR adds HL7 -> FHIR -> HL7 mappings for the HL7 Datatype CP. These are currently only being mapped for OML_O21 messages. 

Test Steps:
1. Integration tests pass

## Changes
- Added `composite-price` in the extension registry as a CP to Extension item.
- Renamed the MOCExtension file from `datatypes/MOC/Extension.yml` to `datatypes/MOC/MOCExtension.yml` to better fit our naming convention.
- Updated OML_O21 FHIR -> HL7 mappings to use the RS specific java classes

### Testing
- [x] Tested locally?
- [x] Added tests?

### Process

## Linked Issues
- Fixes #15490


prime-router/docs/design/design/RS-Extention-Registry.yml
```diff
@@ -3418,6 +3418,99 @@
               valueType: String
               hl7Use: XTN.12
 
+- Resource: Any
```
JFisk42, line 3421: As noted in standup, this pattern is somewhat cobbled together from other extension mapping patterns because of its uniqueness. 

CP is somewhat unique in that:
- It does not exist in the mapping inventory
- It is a datatype that exists on multiple HL7 segments

With that in mind, the idea here is to create something that is more formalized than simply adding each field to Resource.extension by itself. This allows CP to be noted as such when it shows up, and makes it easier to send CP data if a FHIR sender ever needed to do so.

Still, this is what made sense to me, so if there is a format that better fits existing/future patterns I'm open to it.

prime-router/src/main/resources/metadata/hl7_mapping/OML_O21/OML_O21-base.yml
```diff
@@ -1,6 +1,6 @@
 # $schema: ./../../../../../../metadata/json_schema/fhir/fhir-to-hl7-mapping.json
 
-hl7Class: ca.uhn.hl7v2.model.v27.message.OML_O21
+hl7Class: fhirengine.translation.hl7.structures.fhirinventory.message.OML_O21
```
JFisk42, line 3: This was changed to get around some very uncommon edge cases. There is a ticket to update ORU-based mappings with the same change: #16755

This scenario appears to only occur for fields which have been Withdrawn between HL7 2.5.1 and 2.7. This causes the field type to be set to NULLDT by Hapi. This hasn't been an issue with other fields as Hapi will still handle them appropriate when going from a complex fhir type to NULLDT. However, in the case where the withdrawn field has complex subfields (Ex: IN1.38), there will be data modification or loss.

jack-h-wang (approved):
All looks good to me!

prime-router/metadata/HL7/catchall/hl7/segments/IN1/Coverage.yml
```diff
@@ -13,4 +13,36 @@ extension:
   expressions:
     - expressionType: resource
       valueOf: datatypes/AUI/AUIExtension
-      specs: IN1.14
\ No newline at end of file
+      specs: IN1.14
+    - expressionType: resource
```
jack-h-wang, line 17: I think these values will move to `IN1Extension` but that doesn't have to be done now.

arnejduranovic (approved):

prime-router/metadata/HL7/catchall/hl7/segments/IN1/Coverage.yml
JFisk42: I didn't get to this before the codeowner review was left. In the interest of not making reviewers revise an already approved PR I left a comment to have this be done as part of the main IN1 work. https://github.com/CDCgov/prime-reportstream/issues/15500

PR #16792, opened 2024-12-13
Experience/16070/message bank page

### LINKED ISSUE #16070: Create Test Message Bank Page
## Problem statement
The backend work is complete. Now we need to create the front-end pages.

## What you need to know
- [Design in Figma](https://www.figma.com/design/iIGPoS3V2MCVsOOUZKlgiV/Receiver-Message-Testing?node-id=538-29418&node-type=canvas&t=pIFOXdcguJDjtl6W-0)
- Go to with Jessica Willis or Victor Chaparro if you need help interfacing with the APIs that Jessica Willis created
- @victor-chaparro to add API details to the ticket here

## Acceptance criteria
- [x] Message test bank screen is created per design
- [x] When a user clicks "View Message" they see the full content of an individual message in a new tab
- [x] Design, Content, and Functionality reviewed by Design lead (Jillian)
- [x] Reviewed with Tech Lead (Victor) and Product Lead (Chris) before merging

etanb:
This PR creates the initial functionality for the Message Testing feature, and completes the following objectives:

- Create an initial entry point for a user to test messages at the URL: `http://localhost:3000/admin/orgreceiversettings/org/${ORG-NAME}/receiver/${RECEIVER-NAME}/action/edit/message-testing`
- Display all messages for an Org
- Ability to add a custom test message (will only persist within that single browser tab)
- Ability to open message in a new tab

**VIDEO:**

https://github.com/user-attachments/assets/fd73d621-6b50-47fa-8cc1-c2b6529046da

Fixes #16070


frontend-react/src/AppRouter.tsx
```diff
@@ -6,7 +6,7 @@ import { RequireGate } from "./shared/RequireGate/RequireGate";
 import { SenderType } from "./utils/DataDashboardUtils";
 import { lazyRouteMarkdown } from "./utils/LazyRouteMarkdown";
 import { PERMISSIONS } from "./utils/UsefulTypes";
```
jpandersen87, line 8: nit: add spacing after imports

frontend-react/src/components/Admin/MessageTesting.tsx
```diff
@@ -0,0 +1,176 @@
...
+                    </Button>
+                    <Button type="button" onClick={handleAddCustomMessage} disabled={text.length === 0}>
+                        Add
+                    </Button>
+                </div>
+            </div>
+        );
+    };
+
+    const RadioField = ({ title, body, index }: { title: string; body: string; index: number }) => {
```
jpandersen87, line 95: Is there a blocker to using the radio field from the react uswds library? I think you should be able to pass a component as the label, thus reducing the area of concern to the label only.

frontend-react/src/components/Admin/MessageTesting.tsx
```diff
@@ -0,0 +1,176 @@
...
+    const handleSelect = (event: ChangeEvent<HTMLInputElement>) => {
+        setSelectedOption(event.target.value);
+    };
+
+    const handleAddCustomMessage = () => {
+        setSelectedOption(null);
+        setOpenCustomMessage(true);
+    };
+
+    const CustomMessage = () => {
```
jpandersen87, line 46: This should be broken out into it's own component file taking what is needed as props (may require creating a subdirectory for MessageTesting and moving related files).

frontend-react/src/components/Admin/MessageTesting.tsx
jpandersen87, line 95: This should be broken out into it's own component file taking what is needed as props (may require creating a subdirectory for MessageTesting and moving related files).

frontend-react/src/hooks/api/messages/UseMessageTesting.ts
```diff
@@ -0,0 +1,34 @@
+import { useSuspenseQuery } from "@tanstack/react-query";
+import { useCallback } from "react";
+import { reportsEndpoints, RSMessage } from "../../../config/endpoints/settings";
+import useSessionContext from "../../../contexts/Session/useSessionContext";
+import { Organizations } from "../../UseAdminSafeOrganizationName/UseAdminSafeOrganizationName";
+
+const { testing } = reportsEndpoints;
+
+const useReportTesting = () => {
```
jpandersen87, line 9: Name does not convey what it's for specifically. Unless this mimics what the test message entity is named in backend code (oh boy if it does...) it should be something similar to `useTestMessages`.

frontend-react/src/hooks/api/messages/UseMessageTesting.ts
jpandersen87, line 1: Filename needs to reflect final name of hook.

jpandersen87 (changes requested):

frontend-react/src/components/Admin/MessageTesting.tsx
```diff
@@ -0,0 +1,190 @@
...
+                formattedContent = JSON.stringify(JSON.parse(body), null, 2);
+            } catch {
+                formattedContent = body;
+            }
+
+            const blob = new Blob([formattedContent], { type: "text/plain" });
+
+            const url = URL.createObjectURL(blob);
+
+            window.open(url, "_blank");
```
jpandersen87, line 110: Theoretical concern: is this potentially a race condition if we don't wait for the page to finish loading via the window handle returned before releasing the ObjectURL?

etanb: I looked into that, and that's only a problem with much larger files, usually images, which we'd use an extremely short `setTimeout` to accommodate.

frontend-react/src/components/Admin/MessageTesting.tsx
etanb: Yup, will make the replacement!

frontend-react/src/hooks/api/messages/UseMessageTesting.ts
etanb: Changed it all to the Message nomenclature. We chatted about this via Slack, but there's a discrepancy between the API naming `/api/reports/testing` and the design "Message testing", ie, Reports vs. Messages. We're going to go with how the Design looks for the F/E and keep the majority of the naming conventions within the React app Messages but will add a JSDoc comment for this context.

jpandersen87 (approved):

victor-chaparro (approved):
LGTM

jillian-hammer (approved):
looks good! thanks @etanb

PR #16836, opened 2024-12-18
15500 in1 mappings

jack-h-wang:
This PR implements IN1 mappings for OML_O21 and ORM_O21 messages.

Test Steps:
1. Review mappings to verify they match the mapping inventory (unless noted otherwise).
2. Confirm all integration tests pass.

## Changes
- implement IN1 and IN1Extension mappings 
- add hl7v2date extension
- add DT to Period mapping 
- add XPN to RelatedPerson mapping 
- update AUIExtension mapping 
- update ORM_021 mapping
- update tests

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- #15500


prime-router/metadata/HL7/catchall/hl7/codesystem/ExtensionUrlMapping.yml
```diff
@@ -297,9 +303,11 @@
   url: "https://reportstream.cdc.gov/fhir/StructureDefinition/hl7v2Component"
 
 # Date time as hl7v2 string
-# This extension exists because HL7 dates have very little restrictions (i.e. YYYY-MM, YYYY-MM-DD HH, YYYY-MM-DD HH:MM)
-# are all valid, but the FHIR specs has much stricter rules.  In order to reliably generate the same HL7 message, this extension
-# is used to capture what was in the original HL7 message and then exclusively used when mapping FHIR->HL7
+# These extensions exist because HL7 dates have very little restrictions (i.e. YYYY-MM, YYYY-MM-DD HH, YYYY-MM-DD HH:MM)
+# are all valid, but the FHIR specs has much stricter rules.  In order to reliably generate the same HL7 message, these
+# extensions are used to capture what was in the original HL7 message and then exclusively used when mapping FHIR->HL7
+- id: "hl7v2-date"
```
jack-h-wang, line 309: This date only extension was added for two reasons: HL7 DT (date) and DTM (datetime) are distinct datatypes and should have separate handling, and storing it as a different extension makes it so transforms that should only be applied to datetimes (such as timezone conversion) would not apply. So far, only the IN1 mappings have been updated to use this extension.

prime-router/metadata/HL7/catchall/hl7/datatypes/AUI/AUIExtension.yml
```diff
@@ -2,7 +2,7 @@
 
 url:
   type: STRING
-  valueOf: IN1.14
+  valueOf: $auiExtensionName
```
jack-h-wang, line 5: Unsure why this was hardcoded to `IN1.14` previously, but this now better resembles other extension datatypes in that it receives the extension name from the caller.

prime-router/metadata/HL7/catchall/hl7/datatypes/DT/Period.yml
```diff
@@ -0,0 +1,47 @@
+# $schema: ./../../../../../json_schema/fhir/hl7-to-fhir-mapping-resource-template.json
```
jack-h-wang, line 1: This is a so-far unique case where a `Period` is created from two `DT` datatypes. `Period.start` and `Period.end` are of the `datetime` type so dates must be converted to datetime; I have reservations about doing this, but the FHIR inventory specifies this and at least in the extensions the original date is preserved.

prime-router/metadata/HL7/catchall/hl7/segments/IN1/IN1Extension.yml
```diff
@@ -0,0 +1,701 @@
...
+    - expressionType: nested
+      vars:
+        in133: IN1.33
+      condition: $in133 NOT_NULL
+      generateList: true
+      expressionsMap:
+        url:
+          type: STRING
+          value: IN1.33
+        valueString:
```
JFisk42, line 429: Should we consider mapping this and IN1.34 to `valueDecimal` instead? 
Edit: On second thought all these extension fields fall into "whose line is it anyway" territory. So there may be more accurate ways to represent it but unless anyone is attempting send these fields it's going to make no difference.

JFisk42 (commented):
Everything is looking good so far! I got to everything but a couple of the files today, I'll finish those up Monday. 

One extra thing: We should either log a small ticket to add IN1 to ORM mappings or chuck it in on this review.

prime-router/metadata/HL7/catchall/hl7/datatypes/DT/Period.yml
jack-h-wang: That particular method is in `Hl7RelatedGeneralUtils.java` in the HL7 > FHIR converter library. Looking closer at the FHIR output, it does seem like that method is happy to take only a date and format it accordingly (`20240701` became `2024-07-01`).

prime-router/metadata/HL7/catchall/hl7/segments/IN1/Coverage.yml
```diff
@@ -2,47 +2,94 @@
...
+  expressionType: reference
+  condition: $in111 NOT_NULL
+  valueOf: datatypes/XON/Organization
+  specs: IN1.11
+  vars:
+    in111: STRING_ALL, IN1.11
+
+relationship:
+  expressionType: resource
+  valueOf: datatypes/CWE/CodeableConcept
```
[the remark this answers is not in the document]
jack-h-wang: I don't think I've seen such functionality on the HL7 > FHIR library. I think we'd have to do it as a sender transform, but I also think this is a bigger ticket where we have to also decide how to handle cases where this mapping would cause data loss.

prime-router/metadata/HL7/catchall/hl7/segments/IN1/IN1Extension.yml
jack-h-wang: That second thought was my thought process on how to handle `NM`, `ID`, etc. since our experience has been that basically anything can be put in these fields regardless of how the specification says they should be used, and storing a simple string representation is the simplest way to carry that all the way through the pipeline.

prime-router/metadata/HL7/catchall/hl7/segments/IN1/IN1Extension.yml
```diff
@@ -0,0 +1,701 @@
...
+    - expressionType: nested
+      vars:
+        in124: IN1.24
+      condition: $in124 NOT_NULL
+      generateList: true
+      expressionsMap:
+        url:
+          type: STRING
+          value: IN1.24
+        valueString:
```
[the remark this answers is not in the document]
jack-h-wang: We could probably squeeze the `DT` boilerplate mapping here, but my thought process was we have no real reason to perform transforms on `DT` values (unlike `DTM`) so I didn't see a use case where we'd need to convert it when it doesn't map to an existing FHIR attribute.

prime-router/src/main/resources/metadata/hl7_mapping/OML_O21/OML_O21-base.yml
```diff
@@ -40,6 +40,7 @@ elements:
     #          https://github.com/CDCgov/prime-reportstream/issues/15500
```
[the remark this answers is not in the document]
jack-h-wang: I'm having trouble figuring out what adding `RelatedPerson` to the scope actually does for us; as it is, there's no explicit link between the RP and the patient. Writing the scope this way effectively boils down to the logic of include this coverage if `IN1.16` was populated, and don't include it if it wasn't. That doesn't seem right to me either. (For the record, your pseudocode works but the or operator is `or` rather than `||`.)

Looking closer at the `Coverage` resource, I wonder if it makes the most sense to unconditionally map `Coverage.beneficiary` to the patient and use that as the link... that logically makes sense to me, but that's not a mapping that anything is telling us to do.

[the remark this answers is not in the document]
JFisk42: I'm thinking we should keep as it and removing/updating todo because I don't think what is listed in the mapping inventory makes sense. 

The references column for `IN1[Coverage]` states [Coverage.subscriber=Patient[PID]](https://docs.google.com/spreadsheets/d/1cSYLIXVfIjl7ib2DEGxD96LD9-YVJNQtKul5IwzJtts/edit?gid=0#gid=0&range=L31). My understanding of the References column is that it is supposed to represent a required binding. IE How this should read is that we map Coverage back to IN1 if and only if Coverage.subscriber is a reference to the only patient in the bundle. (Please let me know if this is wrong).

So but this only works if IN1-17 is present and 'patient'. So putting anything more restrictive here would be a problem unless we did something radical like just adding a patient reference to Coverage.extension or Coverage.subscriber.extension or Coverage.beneficiary. There's lots of potential for confusion there and all of this is just to prevent if a fhir bundle is sent with irrelevant coverage resources which sounds very unlikely.

prime-router/metadata/HL7/catchall/hl7/segments/IN1/Coverage.yml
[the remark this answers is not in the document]
JFisk42: Apparently I am just not aware of how we handle ValueSet mappings. From what I remember the answer is that we usually try not to, but there are definitely a few places where we are doing it but I cannot figure out the mechanism for it. 

I'll bring this up at the next office hours and don't think it should hold up this pr. I'll take it as a todo to bring up and log a follow up if need be. 

Just to document it somewhere, here is one example from our code base:
[Encounter.status has a required binding. ](http://hl7.org/fhir/R4/encounter.html)
Mapping inventory does have value set mappings for it. [[1](https://docs.google.com/spreadsheets/d/1Sqt8f_LpBVMIdJxcAlB4emDOBqXV8rhy60V2_yaDp7c/edit?gid=0#gid=0), [2](https://docs.google.com/spreadsheets/d/1-g6mrN1Q_Y_foJjlOXY32UAtMWwPi245mh3kRc8FBQ4/edit?gid=0#gid=0)]
And we do get these mapped values from somewhere as it's [in some of the fhir](https://github.com/CDCgov/prime-reportstream/blob/345f93ccd4561677350297d27eae4b9ac006f3ef/prime-router/src/testIntegration/resources/datatests/mappinginventory/catchall/pv1pv2/PV1-to-Encounter.fhir#L401) we've generated for tests.

prime-router/src/main/resources/metadata/hl7_mapping/OML_O21/OML_O21-base.yml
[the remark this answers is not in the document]
jack-h-wang: Looking a little more, `PV1[Coverage]` has `Coverage.beneficiary.reference=Patient[1].id` listed as reference. (This is a really sparse mapping file, only specifying that `PV1-20` maps to `Coverage.type`.) If the implication is that both are supposed to update the same `Coverage`, then creating the `Coverage.beneficiary` link makes sense...

prime-router/metadata/HL7/catchall/hl7/segments/IN1/IN1Extension.yml
JFisk42: That's a really good point! Getting more specific with the value format might be accurate but it forces us to be less flexible on fields that we can't really predict how they will be used.

prime-router/src/main/resources/metadata/hl7_mapping/OML_O21/OML_O21-base.yml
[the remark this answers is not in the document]
JFisk42: Yeah actually I think you're right. Sorry, my brain has been flipping on the references column and I'm overthinking it. Looking at some examples in our code this column is generally treated as something that should be mapped, so adding a reference to the patient on Coverage.beneficiary and then having the mapping here require that sounds correct to me.

JFisk42 (approved):
Looks great!

prime-router/metadata/HL7/catchall/hl7/codesystem/ExtensionUrlMapping.yml
```diff
@@ -52,6 +52,8 @@
   url: "http://hl7.org/fhir/StructureDefinition/patient-animal"
 - id: "relationship"
   url: "https://hl7.org/fhir/StructureDefinition/relationship"
+- id: "extension-subscriber-id"
+  url: "http://hl7.org/fhir/R5/StructureDefinition/extension-subscriberId"
```
jalbinson, line 56: what is the significance of the `R5` path segment? I see one other instance of it and it uses `https` scheme.

jack-h-wang: It's a literal string provided in the [inventory mappings](https://docs.google.com/spreadsheets/d/1-8U35_8X9wQTQAchyVAujF4TFrbcjyDeloU2gpHT2_o/edit?gid=0#gid=0) (row 12, column Q)

prime-router/src/testIntegration/resources/datatests/mappinginventory/catchall/in1/IN1-to-Coverage-IN1-10-SN-IN1-17-SEL.fhir
```diff
@@ -0,0 +1,1260 @@
...
+                ]
+              },
+              "value": "555221234"
+            }
+          },
+          {
+            "url": "https://reportstream.cdc.gov/fhir/StructureDefinition/in1-coverage",
+            "extension": [
+              {
+                "url": "IN1.2",
```
jalbinson, line 224: If I'm reading this correctly, the vast majority of the `IN1` segment is stored in a giant extension on the `Coverage` resource?

jack-h-wang: Correct; we need a location to store any values that either didn't have a FHIR mapping or the FHIR mapping was insufficient for being able to rebuild the HL7. This is the same pattern as other mappings (OBX, PRT, etc).

Issue #15731, opened 2024-08-30
Implement HL7v2 Message Response in Submission API


arnejduranovic:
### User Story

As a sender of Lab Orders using the EPIC system,
I require ReportStream's submission API to return an HL7v2 formatted response instead of the normal JSON,
so that the EPIC system can properly acknowledge the submission was successful. 

### Description/Use Case

Certain senders may require RS response objects to be in a certain format, which may potentially limit the information RS can provide. This ticket is specifically asking for a  HL7v2 formatted message to support epic systems sending lab orders. It is likely this HL7v2 message will be the following: https://hl7-definition.caristix.com/v2/HL7v2.5.1/Segments/MSA

### Dev Notes

- The submission API in RS returns a JSON object with a lot of properties, and we may not need to port all of them to the new message type. This is pending requirements.

- Epic has indicated they are willing to populate “x-application” with a value of “hl7-v2+er7” to indicate to RS it should return an HL7v2 message in the body of the response. Another option is we can create a new sender setting that indicates what format senders want their responses it.

### Acceptance Criteria

- [ ] Submission endpoint supports returning HL7v2 formatted body in HTTP response

arnejduranovic:
Waiting on ETOR-TI team to provide exact requirements for the HL7v2 message for this use case.

JohnNKing:
Draft ETOR ACK requirements: https://cdc.sharepoint.com/:w:/r/teams/OLSS_ETOR_ReportStream/Shared%20Documents/ReportStream%20ETOR%20ACK%20Requirements.docx?d=w93bd416399924de6a301626414d6cea2&csf=1&web=1&e=OEaZKR

Feedback during the LA SME call today indicated we were on the right track. However, we may receive more partner feedback in the coming week.

JohnNKing:
New version of the requirements doc: https://cdc.sharepoint.com/:w:/r/teams/OLSS_ETOR_ReportStream/_layouts/15/Doc.aspx?sourcedoc=%7BE5A9FAE6-8CD2-4C99-A757-F027562C1737%7D&file=ReportStream%20ETOR%20ACK%20Requirements.docx&nav=eyJjIjoxNjg2ODc2OTM0fQ&action=default&mobileredirect=true

CDC is reaching out to the hospital to confirm they have no additional feedback.

JohnNKing:
Feedback from one of our partners (Ochsner): ACKs in staging are working as expected, for orders sent from Epic! 🎉

[...]

PR #16859, opened 2024-12-20
Platform/angela/14116/read hl7

### LINKED ISSUE #14116: As a RS developer, I want a consistent way to read HL7 messages
### User Story
As a RS developer, I want a consistent way to read HL7 messages

### Description/Use Case
As part of refactoring the FhirConverter function an updated method of being able to parse HL7 messages was introduced, but only used in the new code.  This function should be used throughout the code base where we read HL7 messages

### Dev Notes

`HL7Reader` now contains two main methods for reading an HL7v2 message:
- `getMessages` - the old one to be replaced that supports multiple different class definitions per message type
- `parseHL7Message` the new one that supports only one "mega" class definition per message type

All usages of `getMessages` should be replaced with `parseHL7Message`. Once that is done, all methods that were used only by `getMessages` should be removed.

### Acceptance Criteria
- [x] `parseHL7Message` is used everywhere HL7 is parsed in the universal pipeline. Search for this in codebase: `// TODO: https://github.com/CDCgov/prime-reportstream/issues/14116`
- [x] `HL7Reader` has any unused code cleaned up

adegolier:
This PR ...

Removes HL7Reader.getMessages() in favor of HL7Reader.parseHL7Message()

Test Steps:
1. Run tests

## Changes
- Removed getMessages() from HL7Reader and changed references to it to use HL7Reader.parseHL7Message()
- Removed getMessageModelClasses() from HL7Reader because it was only used by getMessages
- Moved additional functions from HL7Reader into the companion object since there is no need to have multiple instances of HL7Reader
- Removed HL7Reader.messageToConfigMap because it was no longer being used
- Marked profileDirectoryMap and getMessageProfile() in HL7Reader as deprecated as they are only used in the CLI tool (which was outside the scope of the ticket) and added corresponding suppression of deprecated warnings for their usage to ProcessFhirCommands.convertHl7ToFhir()

### Testing
- [ ] Tested locally?
- [ ] Added tests?

## Linked Issues
- Fixes #14116


prime-router/src/main/kotlin/fhirengine/utils/HL7Reader.kt
```diff
@@ -21,198 +20,23 @@ import gov.cdc.prime.router.ActionLogger
 import gov.cdc.prime.router.InvalidReportMessage
 import org.apache.commons.lang3.exception.ExceptionUtils
 import org.apache.logging.log4j.Level
-import org.apache.logging.log4j.kotlin.Logging
+import org.apache.logging.log4j.kotlin.logger
 import java.util.Date
-import ca.uhn.hl7v2.model.v251.message.ORU_R01 as v251_ORU_R01
 import ca.uhn.hl7v2.model.v251.segment.MSH as v251_MSH
```
jack-h-wang, line 25: Is there a way we can import only the `v27` version of MSH? I think this is another remnant of the old message model class behavior.

jack-h-wang (changes requested):
Couple quick comments that stood out to me on a quick look, but overall looks good to me!

prime-router/src/main/kotlin/SubmissionReceiver.kt
```diff
@@ -268,14 +269,14 @@ class UniversalPipelineReceiver : SubmissionReceiver {
 
         when (sender.format) {
             MimeFormat.HL7 -> {
-                val messages = HL7Reader(actionLogs).getMessages(content)
-                val isBatch = HL7Reader(actionLogs).isBatch(content, messages.size)
-                // create a Report for this incoming HL7 message to use for tracking in the database
+                val messageCount = Hl7InputStreamMessageStringIterator(content.byteInputStream()).asSequence().count()
```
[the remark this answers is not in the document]
adegolier: I'm a little unclear on what changing a one-liner to a separate method call improves, but I'm not super familiar with Kotlin.

prime-router/src/main/kotlin/fhirengine/utils/HL7Reader.kt
adegolier: Not if we want to keep the check that it is not v2.5

jack-h-wang: Yep, I don't think we need that anymore. There shouldn't be a functional difference between the `v251` and `v27` versions of this, and we're actively looking to delete the `NIST` class entirely.

prime-router/src/main/kotlin/SubmissionReceiver.kt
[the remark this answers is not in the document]
jack-h-wang: I think as general practice it's good that any bit of repeated code is consolidated down to a single call. It makes it much more maintainable in the long run; if for some reason this has to be changed in the future we'll have only one place to change the code or the imports it uses if we drilled it down to a single helper function on our end rather than introducing an explicit external dependency in multiple places in our code.

prime-router/src/main/kotlin/fhirengine/utils/HL7Reader.kt
```diff
@@ -350,6 +151,7 @@ class HL7Reader(private val actionLogger: ActionLogger) : Logging {
         }
 
         // map of HL7 message profiles: maps profile to configuration directory path
+        @Deprecated("This field is only in use for the CLI", level = DeprecationLevel.WARNING)
```
arnejduranovic, line 154: Is this something that needs to be looked into? If so, can we create a story to address this and add it to the PR description? It looks like there's other stuff in here that could probably get deleted that I'm not clear why it exists anymore, like: oidProfileMap (this is leftover from the old strategy iirc)

prime-router/src/test/kotlin/fhirengine/azure/FHIRConverterIntegrationTests.kt
```diff
@@ -1142,4 +1142,33 @@ class FHIRConverterIntegrationTests {
             assertThat(report.bodyFormat).isEqualTo("HL7")
         }
     }
+
+    @Test
+    fun `test should gracefully handle a case with an unsupported contents`() {
```
arnejduranovic, line 1147: This test already exists as part of: `should successfully convert HL7 messages`?

This is that tests setup code:
```
val receivedReportContents =
            listOf(cleanHL7Record, invalidHL7Record, unparseableHL7Record, badEncodingHL7Record)
                .joinToString("\n")
```
notice: invalidHL7Record, unparseableHL7Record, badEncodingHL7Record

prime-router/src/main/kotlin/SubmissionReceiver.kt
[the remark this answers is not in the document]
adegolier: Yes, I'm aware of that. I was just under the impression that it might be different in Kotlin for some reason since that was exactly what I said about wanting to modify getMessages() instead of deleting it altogether.

prime-router/src/main/kotlin/fhirengine/utils/HL7Reader.kt
adegolier: For some reason I thought this was in a test. This is not in scope for the ticket.

prime-router/src/main/kotlin/fhirengine/utils/HL7Reader.kt
adegolier: If it wasn't in the CLI (and dealing with that out of scope for this ticket) I would have went ahead and removed it, but I did want to keep other people from adding more stuff that uses it. I can add a ticket, yeah.

prime-router/src/main/kotlin/fhirengine/utils/HL7MessageHelpers.kt
```diff
@@ -103,4 +103,8 @@ object HL7MessageHelpers : Logging {
 
         return builder.toString()
     }
+
+    fun messageCount(rawHl7: String): Int {
```
jack-h-wang, line 107: This function should be in a companion object; that way, a new instance of `HL7MessageHelpers` doesn't need to be instantiated every time this function is called.

prime-router/src/main/kotlin/SubmissionReceiver.kt
```diff
@@ -290,12 +291,14 @@ class UniversalPipelineReceiver : SubmissionReceiver {
...
-                messages.forEachIndexed {
-                    idx, element ->
-                    MessageType.validateMessageType(element, actionLogs, idx + 1)
+                if (messageCount == 0 && !actionLogs.hasErrors()) {
+                    actionLogs.error(InvalidReportMessage("Unable to find HL7 messages in provided data."))
                 }
+                // check for valid message type
+                Hl7InputStreamMessageStringIterator(content.byteInputStream()).asSequence()
+                    .forEachIndexed { index, rawItem ->
+                        MessageType.validateMessageType(HL7Reader.parseHL7Message(rawItem, null), actionLogs, index + 1)
```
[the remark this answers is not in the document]
adegolier: I think I got sidetracked trying to understand if this would affect the ack or not

prime-router/src/main/kotlin/SubmissionReceiver.kt
```diff
@@ -290,12 +291,14 @@ class UniversalPipelineReceiver : SubmissionReceiver {
...
-
-                // check for valid message type
-                messages.forEachIndexed {
-                    idx, element ->
-                    MessageType.validateMessageType(element, actionLogs, idx + 1)
+                if (messageCount == 0 && !actionLogs.hasErrors()) {
+                    actionLogs.error(InvalidReportMessage("Unable to find HL7 messages in provided data."))
                 }
+                // check for valid message type
+                Hl7InputStreamMessageStringIterator(content.byteInputStream()).asSequence()
```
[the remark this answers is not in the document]
adegolier: You can't do that because the sequence can be consumed only once.

prime-router/src/main/kotlin/fhirengine/utils/HL7MessageHelpers.kt
adegolier: HL7MessageHelpers is an object. As far as I understand it, classes can have companion objects, but objects cannot.

jack-h-wang (approved):

prime-router/src/main/kotlin/SubmissionReceiver.kt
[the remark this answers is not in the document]
JFisk42: Marking this resolved because the ask is for a functional change in the SubmissionReceiver. A ticket was logged for this: https://github.com/CDCgov/prime-reportstream/issues/16902

prime-router/src/main/kotlin/azure/service/SubmissionResponseBuilder.kt
```diff
@@ -106,12 +106,11 @@ class SubmissionResponseBuilder(
...
-            val hl7Reader = HL7Reader(ActionLogger())
-            val messages = hl7Reader.getMessages(requestBody)
-            val isBatch = hl7Reader.isBatch(requestBody, messages.size)
+            val messageCount = Hl7InputStreamMessageStringIterator(requestBody.byteInputStream()).asSequence().count()
+            val isBatch = HL7Reader.isBatch(requestBody, messageCount)
 
-            if (!isBatch && messages.size == 1) {
-                val message = messages.first()
+            if (!isBatch && messageCount == 1) {
+                val message = HL7Reader.parseHL7Message(requestBody, null)
```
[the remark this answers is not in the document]
JFisk42: Marking this resolved as a ticket was logged: https://github.com/CDCgov/prime-reportstream/issues/16904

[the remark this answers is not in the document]
adegolier: Fixed it after all.

prime-router/src/main/kotlin/fhirengine/utils/HL7Reader.kt
adegolier: Fixed this now

prime-router/src/main/kotlin/cli/ProcessFhirCommands.kt
```diff
@@ -516,16 +516,15 @@ class ProcessFhirCommands : CliktCommand(
      * look like.
      * @return a FHIR bundle and the parsed HL7 input that represents the data in the one HL7 message
      */
+    @Suppress("DEPRECATION")
```
arnejduranovic, line 519: What about the code changes requires this suppression now?

prime-router/src/main/kotlin/azure/service/SubmissionResponseBuilder.kt
```diff
@@ -106,12 +106,11 @@ class SubmissionResponseBuilder(
...
-            val hl7Reader = HL7Reader(ActionLogger())
-            val messages = hl7Reader.getMessages(requestBody)
-            val isBatch = hl7Reader.isBatch(requestBody, messages.size)
+            val messageCount = HL7MessageHelpers.messageCount(requestBody)
+            val isBatch = HL7Reader.isBatch(requestBody, messageCount)
 
-            if (!isBatch && messages.size == 1) {
-                val message = messages.first()
+            if (!isBatch && messageCount == 1) {
+                val message = HL7Reader.parseHL7Message(requestBody)
```
arnejduranovic, line 113: I think we need to be careful here. parseHL7Message can return exceptions, so what does that mean for the submission api in general? If there is a failure here, it will except out of the main `processRequest` function and an HTTP error will be returned in from the caller `submitToWaters`. The DB state will be updated I think by this point, but the queueMessage for the convert step will not be populated. 

I poked around a bit and I think this is fine. This submission method is slated to be deprecated and we should take this issue into consideration when implementing this HL7v2 ACK functionality into the submission microservice.

arnejduranovic (approved):

prime-router/src/main/kotlin/azure/service/SubmissionResponseBuilder.kt
arnejduranovic: I updated the DevNotes section of this ticket linking to this comment: https://github.com/CDCgov/prime-reportstream/issues/15731

PR #16920, opened 2024-12-27
Flexion/jgilmore/update blob storage transport filenames

GilmoreA6:
This PR updates Blob Storage Transport to utilize the file-naming-templates in the same manner as the current SFTP transport. Currently Blob Storage Transport names the file copied into the receiver's storage container with the same name as the file in the RS "ready" folder. This will impact existing receivers utilizing blob storage transport so it may be worth adding a file-naming-template that simply passes along the existing filename. ETOR

Test Steps:
1. run ./prime test

2. run ./gradlew testIntegration

3. Add the attached settings to local organizations.yml file
[golden-copy-test-settings.txt](https://github.com/user-attachments/files/18264051/golden-copy-test-settings.txt)

4. In storage account devstoreaccount1 create a container named "test"

5. Use attached test message to send a message to local reportstream from sender "flexion.simulated-hospital"
[test-blob-name.txt](https://github.com/user-attachments/files/18264103/test-blob-name.txt)

6. Check created "test" container for two files, one named "golden-copy-<UUID>.hl7" and one named "oru_r01-base.yml-<UUID>.hl7

## Changes
- simplify copyBlob function to only return the byteArray of the copied blob
- add logic to utilize the externalFilename in blobstoretransport
- clean tests for copyBlob

### Testing
- [x ] Tested locally?
- [x ] Added tests?


prime-router/src/main/kotlin/azure/BlobAccess.kt
```diff
@@ -404,14 +404,10 @@ class BlobAccess() : Logging {
         /**
          * Copy a blob at [fromBlobUrl] to a blob in [blobConnInfo]
          */
-        fun copyBlob(fromBlobUrl: String, blobConnInfo: BlobContainerMetadata): String {
+        fun copyBlob(fromBlobUrl: String): ByteArray {
```
jack-h-wang, line 407: This change essentially turns `copyBlob` into a copy of `downloadBlobAsByteArray` except with one extra line of logging. Why not call `downloadBlobAsByteArray` directly? If we don't need `copyBlob` after doing that, then we could delete it (and its test) entirely.

I think it's also worth pointing out the `digest` parameter for `downloadBlob` is available (as `BlobUtils.digestToString(header.reportFile.blobDigest)`) though the return type for that function is `String` instead of `ByteArray`. Performing that bit of verification might make sense?

jack-h-wang (changes requested):

prime-router/src/main/kotlin/azure/BlobAccess.kt
GilmoreA6: @jack-h-wang agree with getting rid of the copyBlob, clarification re: the digest parameter, are you thinking that we should compare the blobDigest to the output to make sure they are the same?

GilmoreA6: @jack-h-wang updated to call downloadBlobAsByteArray directly and removed copyBlob and associated tests. We discussed added the compare between the digest and a digest of the downloaded blob but because digestToString has a limit of 40 we thought it added complexity without really verifying much. Happy to still add if you think it's necessary.

jack-h-wang (approved):

prime-router/src/main/kotlin/transport/BlobStoreTransport.kt
```diff
@@ -31,7 +31,11 @@ class BlobStoreTransport : ITransport {
             val receiver = header.receiver ?: error("No receiver defined for report ${header.reportFile.reportId}")
             val bodyUrl = header.reportFile.bodyUrl ?: error("Report ${header.reportFile.reportId} has no blob to copy")
             context.logger.info("About to copy $bodyUrl to $envVar:$storageName")
-            val newUrl = BlobAccess.copyBlob(bodyUrl, BlobAccess.BlobContainerMetadata.build(transportType))
+            val blobConnection = BlobAccess.BlobContainerMetadata.build(transportType)
+            val blobFile = BlobAccess.downloadBlobAsByteArray(bodyUrl)
```
jalbinson, line 35: Outside the scope of this ticket/bugfix but I think a better approach would be to use the blob client to copy the blob within Azure instead of download and then upload. It looks like there are some APIs that could accomplish this:

https://learn.microsoft.com/en-us/java/api/com.azure.storage.blob.specialized.blobclientbase?view=azure-java-stable#com-azure-storage-blob-specialized-blobclientbase-copyfromurl(java-lang-string)

jalbinson (approved):
Alternative approach suggested but not required.

PR #16950, opened 2025-01-03
Update dependabot config to ignore specific dependencies' major versions

### LINKED ISSUE #16949: Update dependabot config to ignore specific dependencies' major versions
By default dependabot, regardless of semver syntax on dependencies, will attempt to update across major versions. Unfortunately this means needing to periodically update the dependabot config to ignore major versions for dependencies that need manual migration (though not ideal as we want to keep edits to the dependabot config to a minimum).

Acceptance Criteria
- [x] Current list of libraries with incompatible major updates added to dependabot frontend ignore list (specifying to ignore major updates only)
- [x] Aforementioned libraries' entries in package.json reverted back to `^` semver syntax rather than `~`

jpandersen87:
Fixes #16949

This PR:

- Manually specifies ignore rules for major version updates to libraries that require manual migration
- Reverts semver syntax in package.json for these libraries to default


devopsmatt (approved):

devopsmatt:
@jpandersen87  Approved, but needs conflicts to be resolved before merging

PR #17012, opened 2025-01-07
upgrading to kotlin 2.0 and fixing tests

### LINKED ISSUE #16908: Resolve blockers to updating to Kotlin 2.0.0 and later
### User Story
As an RS engineer, I want to remain on current and supported versions of dependencies.

### Description/Use Case
RS currently runs Kotlin version 1.9.25. Attempts to update to Kotlin 2.0.0 and later result in compiler failures. There are a number of dependency updates that now require Kotlin 2.0.0 or later. We need to resolve any blockers and update to at minimum Kotlin 2.0.0 in order to update to the current version of many of our dependencies.

See [the list of dependencies that cannot currently be updated](https://github.com/CDCgov/prime-reportstream/pulls?q=is%3Aopen+is%3Apr+label%3Ajava+author%3Aapp%2Fdependabot+status%3Afailure+). Note that not all failures are due to this issue; inspect the output of the Build Router step to determine which failures are due to the Kotlin version. 

### Dev Notes
Update `systemProp.kotlinVersion` in `gradle.properties` to a Kotlin 2.x version. Note that `compileKotlin` will fail due to the presence of several instances of `@Suppress("NEW_INFERENCE_NO_INFORMATION_FOR_PARAMETER")`. Removing these annotations allows compile to succeed, but causes some tests to fail.

See also:
- #15430 
- #15441

### Acceptance Criteria
- [ ] Update to Kotlin 2.0.0 or later without any test failures

kant777:
This PR upgrades kotlin to version `2.1.0` and fixes the broken tests according to the acceptance criteria define here https://github.com/CDCgov/prime-reportstream/issues/16908

Test Steps:
1. ./gradlew quickRun

## Changes
image

### Testing
- [x] Tested locally? Yes
- [ ] Ran `./prime test` or `./gradlew testSmoke` against local Docker ReportStream container? Yes
- [ ] (For Changes to /frontend-react/...) Ran `npm run lint:write`? Yes
- [ ] Added tests? No

## Linked Issues
- Fixes #16908

### DIFF
--- a/prime-router/docs/design/proposals/0023-condition-to-code-mapping/0023-condition-to-code-mapping.md
+++ b/prime-router/docs/design/proposals/0023-condition-to-code-mapping/0023-condition-to-code-mapping.md
@@ -19,7 +19,7 @@ The below items are not covered in this proposal.
 
 Overview Diagram
 
-![Code to Condition Overview Diagram](C:\Users\James.Gilmore\Documents\GitHub\PRIME\prime-reportstream\prime-router\docs\design\proposals\0023-condition-to-code-mapping\code-to-condtion-overview.png)
+![Code to Condition Overview Diagram](https://github.com/CDCgov/prime-reportstream/blob/main/prime-router/docs/design/proposals/0023-condition-to-code-mapping/code-to-condtion-overview.png)
 
 ### Creating Observation Mapping Table
 


prime-router/src/test/kotlin/azure/batch/UniversalBatchFunctionTests.kt
```diff
@@ -448,10 +455,10 @@ class UniversalBatchFunctionTests {
         every { mockTask.schemaName } returns schemaName
         every { mockTask.bodyFormat } returns bodyFormat
         every { engine.generateEmptyReport(any(), any()) } returns Unit
-        every { engine.db.fetchAndLockBatchTasksForOneReceiver(any(), any(), any(), any(), any()) } returns listOf(
+        every { mockDb.fetchAndLockBatchTasksForOneReceiver(any(), any(), any(), any(), any()) } returns listOf(
```
arnejduranovic, line 458: If you had to make this change on this line then that means you are not properly passing mockDb to the engine object? You are passing mockDb to engine on 437, so why doesn't engine.db work here? Even though this test works, I think it's messy and can get cleaned up because it's not clear what function/purpose the object on line 54 has?

auth/build.gradle.kts
```diff
@@ -7,6 +7,10 @@ plugins {
     kotlin("plugin.spring") version "2.1.0"
 }
 
+ktlint {
+    version = "1.5.0"
```
jack-h-wang, line 11: Just wanted to point out a different ktlint version is specified in [reportstream.project-conventions.gradle.kts](https://github.com/CDCgov/prime-reportstream/blob/f48d719b876859169deb0360487f63965d8be5a0/buildSrc/src/main/kotlin/reportstream.project-conventions.gradle.kts#L37). We should try to keep these versions aligned, but also I'd want to look into whether we should have the version specified in only one location.

arnejduranovic (approved):
Top significant changes that I see (I approve of them):

1. moving function declarations to single line when short enough
2. replacing function bodies with body expression when its a single statement function `(standard:function-expression-body)`

For testing, I undid some changes in the PR and ran `./gradlew :prime-router:ktlintCheck` to verify ktlint catches them on my machine. I confirmed there is no local setup needed to get same behavior.

scott-aquia (approved):
I have reviewed this change from a security perspective.

PR #17093, opened 2025-01-17
Tf deplotment changes

bethbeza:
Places tf plan and apply in its corrected place.


.github/workflows/deploy_terraform.yml
```diff
@@ -63,7 +63,16 @@ jobs:
...
           add-args: "-refresh=false"
-
+        -name:  Run Terraform Plan
+          run: |
+            terraform init -input=false
+            terraform validate
+            terraform fmt -recursive
+            terraform plan -out=tf.plan
+        - name: Run Terraform Apply
+          run: |
```
devopsmatt, line 73: I think this apply step should be conditional on the plan step, using `needs: ... `

.github/workflows/cleanup_acr_images.yml
devopsmatt, line 1: I think this file doesn't need modifications, we should exclude it from the PR.

.github/workflows/0
devopsmatt, line 1: I think this was included in the commit by mistake/type, it should not be in this PR.

devopsmatt (approved):

PR #17094, opened 2025-01-17
16815 authz design

### LINKED ISSUE #16815: Design and Document Modern and Secure AuthZ Strategy First Draft
### User Story

As a stakeholder of ReportStream,
I want users and machines to be able to authenticate and authorize to ReportStream securely and simply,
to minimize vulnerabilities and increase maintainability

### Description/Use Case

There was an effort by the Engagement team to [document permission needs from a user perspective](https://app.mural.co/t/primedatainput6410/m/primedatainput6410/1707502140746/6af0ac6188bf57666b601c7563abee95385f5a4f). This needs to be synthesized with our understanding of machine auth and [previous work to understand current authz](https://github.com/CDCgov/prime-reportstream/pull/16769) to come up with a design that serves identified `machine` and `user` AuthZ needs (see linked documents for details).

- Back and forth with engagement and product may delay this ticket

### Dev Notes

An example of a complicated use case: 

As a sender under the CDC_ELIMS organization, I would like to:
- Submit reports under my 'default' sender
- View my submissions page on the ReportStream website
- View the daily data page of all the receivers I am sending to so that I can see what they received

### Acceptance Criteria

- [x] AuthZ Design document under `prime-router/docs/authz/`
- [x] AuthZ Design approved by:
    - [ ] @arnejduranovic and Platform

jalbinson:
This PR documents a new design for authorizing endpoints in ReportStream.

## Linked Issues
- Fixes #16815


jalbinson:
Easy access to see how the page looks rendered
https://github.com/CDCgov/prime-reportstream/blob/platform/jamie/16815-authz-design/prime-router/docs/authz/authz-design.md

prime-router/docs/authz/authz-design.md
```diff
@@ -0,0 +1,190 @@
...
+}
+```
+
+This particular user would be able to read information about `md-phd` and `ca-phd`. They would also be able to submit 
+reports only as `md-phd.full-elr` or for any sender under `ca-phd`. Note that the current implementation uses an 
+`organization` claim. This will continue to be present during the transition but will eventually be phased out.
+
+## Setting up a new user
+
+In this system, setting up a new user would be quite easy. Create the user in Okta and then add it the required
```
arnejduranovic, line 102:
```suggestion
In this system, setting up a new user would be quite easy. Create the user in Okta and then add it to the required 
```

prime-router/docs/authz/authz-design.md
```diff
@@ -0,0 +1,190 @@
...
+) {
+    val application = applicationApi.getApplication(clientId, null)
+    val groups = getApplicationGroups(clientId)
+    application.putprofileItem("groups", groups)
+    application.putprofileItem("appSubmit", submit)
+    applicationApi.replaceApplication(arnejAppId, application)
+}
+```
+
+We will also need to keep these profiles up to date, therefore, we would need some sort of daily job to check that groups
```
arnejduranovic, line 129: I think there may be more thinking to do here when we design this CLI (if we decide to go this way). Perhaps the CLI tool doesn't just update the profiles but we also use it to create/update applications as well (so a more full-featured CLI) and don't do anything in the Okta GUI for applications so as to not confused things? Does this sound reasonable/plausible?

prime-router/docs/authz/authz-design.md
```diff
@@ -0,0 +1,190 @@
...
+rapid group changes.
+
+## Authorization check
+
+There will have to be multiple approaches to doing the actual authorization check given it will work differently across 
+different application frameworks.
+
+### Spring
+
+In Spring we can leverage method security. This allows us to write custom predicates in SpEL (Spring expression language)
```
arnejduranovic, line 140: Note that this only applies to Submission microservice at the moment, right?

prime-router/docs/authz/authz-design.md
```diff
@@ -0,0 +1,190 @@
...
+## Access Token Design
+
+### Scopes
+Scopes refer to actions that can be done on certain endpoints. Scopes are requested during the login process. We 
+will set up rules in Okta that allow scopes to be requested based on organization membership. A requested scope will 
+only be returned given you have the correct group membership.
+
+| Scope       | Actions                                                       | Okta Group              |
+|-------------|---------------------------------------------------------------|-------------------------|
+| super_admin | Anything! (org membership does not matter)                    | ReportStream-SuperAdmin |
```
arnejduranovic, line 18: I'm confused about Okta Group as shown here? So a User (or App?) can be a member of these groups and in addition there will be groups for each Sender/Receiver Organization, like "CA_PHD"? If this is correct, can we update the claims example that's lower down to show how/where these groups would appear?

prime-router/docs/authz/authz-design.md
jalbinson: I think this is reasonable. I'm fairly certain everything you need to make a new application user is available in the API via the Okta Admin SDK.

prime-router/docs/authz/authz-design.md
jalbinson: You can limit which scopes are available to you via group membership so thats how I thought about it. You can also use regex to limit groups that get added to the `org` claim. We currently limit groups added to the `organization` claim with only groups that start with "DH".

My idea is:
- Add user to correct group for scope
- Add user to correct org group
- Optionally update profile for those that need to directly submit

tsurya (approved):
LGTM

prime-router/docs/authz/authz-design.md
arnejduranovic: Okay I think I understand. A follow-up question is: In ### Claims example would the example have "ReportStream-OrgAdmin", or any of the other "scope" groups in the `"org"` attribute?

prime-router/docs/authz/authz-design.md
```diff
@@ -0,0 +1,190 @@
...
+| org:write   | Able to update organizations                                  | ReportStream-OrgAdmin   |
+| org:read    | Able to access read-only information related to organizations | ReportStream-User       |
+| submit      | Able to submit reports                                        | ReportStream-Submit     |
+| elims       | Special `elims` specific actions (see below)                  | ReportStream-Elims      |
```
arnejduranovic, line 22: Hmm I don't really like the idea of this, it's coupled to a particular customer's use case. Instead we should look to try to understand the use case in a "generic" way. For example, what would a generic name for this scope be? I think there wouldn't be one because its a mishmash to serve a weird edge case? 

I think I would prefer we don't support this and instead force users to create new accounts if they want a particular role for a particular organization? So that means:

If a user has a "scope group" those scopes will apply to all "org groups". If we want a user to have different "scope groups" for different "org groups" that will not be supported and instead a new user account should be created. This makes sense to me because from an organizational perspective that IS a different user. 

For example, if I worked for CDC and Palantir, I would have two different accounts, one for each organization, and if the two companies work closely together and communicate via slack, I could log in to slack with either of my two different accounts and have different permissions depending on what account I am signed in under.

prime-router/docs/authz/authz-design.md
jalbinson: My thought would be to exclude them via a fancy regex like so:

Screenshot 2025-01-24 at 1 19 39 PM

prime-router/docs/authz/authz-design.md
jalbinson: > For example, what would a generic name for this scope be? I think there wouldn't be one because its a mishmash to serve a weird edge case?

We could break it out into something like the following if it would be more clear.
```
submissions
daily-data
```

> If a user has a "scope group" those scopes will apply to all "org groups". If we want a user to have different "scope groups" for different "org groups" that will not be supported and instead a new user account should be created. This makes sense to me because from an organizational perspective that IS a different user.

I'm a little confused on this. Okta groups are serving multiple purposes in this design. The `ReportStream-*` groups serve as a way to show what users are able to request certain scopes. The `org` groups are the more standard `md-phd` type showing what organization your scope will apply to.

arnejduranovic: Okay I think I figured out at least part of what's not sitting right with me. I think we are actually trying to combine `user/group permissions (scopes)` with `organization settings`. For example: If we want to give CDC_ELIMS organization permission to view NY's daily data page, then that's not a new scope, but a new organization setting (managed in RS settings database), like so:

```
CA_DPH
    - allow-daily-data-access: [CDC_ELIMS]
 ```

File: auth/docs/setup.md [hunk 6]
```diff
@@ -5,10 +5,25 @@
...
-|----------------------|---------------------------------|
-| OKTA_ADMIN_CLIENT_API_ENCODED_PRIVATE_KEY | Base 64 encoded private key pem |
-| SPRING_SECURITY_OAUTH2_RESOURCESERVER_OPAQUETOKEN_CLIENT_SECRET | Base 64 encoded secret |
+| Environment variable                                            | Value                           |
+|-----------------------------------------------------------------|---------------------------------|
+| OKTA_ADMIN_CLIENT_API_ENCODED_PRIVATE_KEY                       | Base 64 encoded private key pem |
+| SPRING_SECURITY_OAUTH2_RESOURCESERVER_OPAQUETOKEN_CLIENT_SECRET | Shared secret                   |
+
+Setup for `OKTA_ADMIN_CLIENT_API_ENCODED_PRIVATE_KEY`:
+1. Create a new application user
```
arnejduranovic, line 14: How? What is `application user` vs `application`? Please answer in the doc.

prime-router/docs/authz/authz-design.md
arnejduranovic: TODO/UPDATE: As we discussed, add more detail here on how we can set this up so that "syncing" doesn't need to happen

arnejduranovic (changes requested):

prime-router/docs/authz/authz-design.md
jalbinson: Update this section

prime-router/docs/authz/authz-design.md
```diff
@@ -0,0 +1,242 @@
...
+will set up rules in Okta that allow scopes to be requested based on organization membership. A requested scope will 
+only be returned given you have the correct group membership.
+
+| Scope       | Actions                                                               | Okta Group               |
+|-------------|-----------------------------------------------------------------------|--------------------------|
+| super_admin | Anything! (org membership does not matter)                            | ReportStream-SuperAdmin  |
+| org:write   | Able to update organizations                                          | ReportStream-OrgAdmin    |
+| org:read    | Able to access read-only information related to organizations         | ReportStream-User        |
+| submit      | Able to submit reports                                                | ReportStream-Submit      |
+| submissions | Able to access submissions page                                       | ReportStream-submissions |
```
jalbinson, line 22: No longer required since both covered by `org:read`

prime-router/docs/authz/authz-design.md
```diff
@@ -0,0 +1,242 @@
...
+    "daily-data"  
+  ],
+  "org": [
+    "elims"
+  ]
+}
+```
+
+At a code level:
+- ensure the user contains the `daily-data` scope
```
jalbinson, line 132: simplify because daily-data page is only 1 receiver at a time

auth/src/scripts/get_client_access_token.py
```diff
@@ -0,0 +1,155 @@
...
+
+"""
+okta_private_key and client_id is generated within Okta by an RS onboarding person and emailed
+to the client. See https://developer.okta.com/docs/guides/implement-oauth-for-okta-serviceapp/main
+
+See this documentation to understand how to setup an authorization server:
+https://developer.okta.com/docs/guides/customize-authz-server/main/#create-scopes
+"""
+authorization_server_id = "" # set your authorization server id here
+client_id = "" # set you application user's client it
```
arnejduranovic, line 21:
```suggestion
client_id = "" # set your application user's client id
```

prime-router/docs/authz/authz-design.md
```diff
@@ -0,0 +1,279 @@
...
+groups will allow us to make authorization decisions on most endpoints. Organizations are easily added to the token 
+for users via the UI. I will delve into how to set it up for senders later in the document.
+
+<img src="img/groups-claim.png" width="600"/>
+
+### Submit claim
+
+There will also be an optional `userSubmit` or `appSubmit` claim containing specific organizations AND optional sender. These values will 
+be dot seperated. If a sender is not included it will be assumed to be a wildcard and allow all senders under that 
+organization. Unfortunately the claim names have to seperated as to access the user profile and application user profile
```
arnejduranovic, line 49:
```suggestion
organization. Unfortunately the claim names have to be separated as to access the user profile and application user profile 
```

prime-router/docs/authz/authz-design.md
```diff
@@ -0,0 +1,279 @@
...
+for users via the UI. I will delve into how to set it up for senders later in the document.
+
+<img src="img/groups-claim.png" width="600"/>
+
+### Submit claim
+
+There will also be an optional `userSubmit` or `appSubmit` claim containing specific organizations AND optional sender. These values will 
+be dot seperated. If a sender is not included it will be assumed to be a wildcard and allow all senders under that 
+organization. Unfortunately the claim names have to seperated as to access the user profile and application user profile 
+require different expressions.
```
arnejduranovic, line 50:
```suggestion
requires different expressions.
```

arnejduranovic: Also can we elaborate on what "requires different expressions" means or find a way to reword this section?

prime-router/docs/authz/authz-design.md
arnejduranovic: What do you mean by "claim names have to be separated"? You mean userSubmit vs appSubmit?

prime-router/docs/authz/authz-design.md
```diff
@@ -0,0 +1,279 @@
...
+```
+
+This particular user would be able to read information about `md-phd` and `ca-phd`. They would also be able to submit 
+reports only as `md-phd.full-elr` or for any sender under `ca-phd`. Note that the current implementation uses an 
+`organization` claim. This will continue to be present during the transition but will eventually be phased out.
+
+### Daily Data page special considerations
+
+For specific users within specific organizations, we want to allow them to access another organization's daily data 
+page. Since these are organizational level settings that are tightly coupled to Report Stream, we would store these
```
arnejduranovic, line 98:
```suggestion
page. Since these are organizational level settings that are tightly coupled to ReportStream, we would store these 
```

prime-router/docs/authz/authz-design.md
```diff
@@ -0,0 +1,279 @@
...
+
+This particular user would be able to read information about `md-phd` and `ca-phd`. They would also be able to submit 
+reports only as `md-phd.full-elr` or for any sender under `ca-phd`. Note that the current implementation uses an 
+`organization` claim. This will continue to be present during the transition but will eventually be phased out.
+
+### Daily Data page special considerations
+
+For specific users within specific organizations, we want to allow them to access another organization's daily data 
+page. Since these are organizational level settings that are tightly coupled to Report Stream, we would store these 
+settings within the RS Postgres DB.
```
arnejduranovic, line 99: As we said, this is probably fine for version 1.0 but it would be good to think on this and if there is a better way to manage and store org settings. For example, does Okta provide any facilities for this?

prime-router/docs/authz/authz-design.md
```diff
@@ -0,0 +1,279 @@
...
+    application.putprofileItem("appSubmit", submit)
+    applicationApi.replaceApplication(arnejAppId, application)
+}
+```
+
+We could go even further and write various APIs (and potentially CLI calls) that will set up an application user from 
+scratch. This would dramatically simplify the process of setting up a sender since we would not need to handle going
+back and forth between API and Okta Admin UI and could provide the necessary values needed for a new sender.
+We could also reuse our existing authentication and authorization within the project to ensure only specific people are able to 
+hit these endpoints.
```
arnejduranovic, line 163: Maybe just call out that the next/final step could be to create a GUI in ReportStream where engagement can fill out a form and the app user is correctly created/updated.

prime-router/docs/authz/authz-design.md
```diff
@@ -0,0 +1,279 @@
...
+    }
+    
+    return containsSubmitScope && canSubmit
+}
+
+```
+
+## Deprecated approach to passing groups downstream
+
+At the time of writing this section (2/5/25), code still exists which is doing the following:
```
arnejduranovic, line 271: Where is this code? Auth microservice? Also filenames/functions would be helpful.

prime-router/docs/authz/authz-design.md
jalbinson: I meant that a claim name has to be unique and can only have a single expression.

My workaround:
```
appSubmit: app.profile.submit
userSubmit: user.submit
```

I'll clarify in the document.

prime-router/docs/authz/authz-design.md
jalbinson: see above ^

jack-h-wang (approved):

auth/src/main/resources/application.yml
```diff
@@ -7,7 +7,7 @@ spring:
     oauth2:
       resourceserver:
         opaquetoken: # Set client secret in SPRING_SECURITY_OAUTH2_RESOURCESERVER_OPAQUETOKEN_CLIENT_SECRET env variable
-          client-id: 0oaek8tip2lhrhHce1d7
+          client-id: 0oak8tcr0z6h2RC671d7
```
arnejduranovic, line 10: What resource server is this changing from/to?

arnejduranovic (approved):

auth/src/main/resources/application.yml
jack-h-wang: I changed this to a new app integration I created to make certain I was able to set it up from scratch. These will likely change again when we get closer to implementing the transition plan.

PR #17143, opened 2025-01-24
UP Deduplication Design Documentation

### LINKED ISSUE #14103: Deduplication in the UP MVP
## Outcome/Objective  

An MVP version of deduplication detection is implemented in the UP for ELR use cases (ORU-R01). 

 

## Context 

Deduplication was originally built into the CP in response to a specific sender bug. This occurred the beginning of July 22 where a sender(PMG) introduced a bug in their system resulting in tests results not being marked as sent. That meant their system starting sending the same message over and over again. This continued for a few days before we turned off the auth for that sender. In those few days PMG was sender >800,000 messages/day. We worked with PMG and they eventually fixed the bug.  

  

In followup to that that incident, we decided to implement deduplication logic to catch the resends in the event a similar situation occurred in the future. The idea being that duplicate reports would be discarded without the need to completely disable the sender. 

  

Due to a [bug](https://app.zenhub.com/workspaces/platform-6182b02547c1130010f459db/issues/gh/cdcgov/prime-reportstream/14102), in the implementation the duplication feature never worked for the UP and needs to be re-implemented 

  

Furthermore, recently to solve the file limit issue we've decoupled the batching step within the pipeline so that we can process messages one at a time. However, we do allow senders to send batched messages. Thus we will have scenarios where a sender sends a batch, some messages in that batch are good and processed and others error out. When we ask the sender to fix the errors and resend, if they send in batch, they could resend the whole batch including the messages that were already good and processed. To avoid processing the same messages twice we need a way to detect the duplicate message in the batch and ensure the duplicates are not processed, only the messages that needed fixing should get through. 

 

Some receivers, such as CA Department of Public Health have deduplication systems in place, but not all receivers are fortunate to have the same resources. Therefore, ReportStream aims to provide the same opportunity to access to improved data accuracy by implementing deduplication prior to delivering to its receivers. 

 

## Use Cases 

- As a receiver of the UP, I would like to not receive duplicative data – so that I can have a more accurate account of data that I’m receiving. 

- As a sender to the UP, I would like to submit reports to ReportStream in the most convenient method possible, which may sometimes contain previously submitted, duplicate messages. 

 

## Scenarios  

_“Items” refer to the messages contained within a report. An item may contain multiple DiagnosticReports and Observations (including AOEs)._ 
**SCENARIOS SUPPORTED BY THIS DEDUPLICATION MVP**
 >**1. Deduplicating reports**
_This is what was historically supported by the COVID Pipeline._   
Sender submits Report1 on Nov 1 and submits Report1 (with no changes to the content) on Nov 2. The deduplication feature will remove any subsequent submissions of Report1 (in this case, the one submitted on Nov 2).

>**2. Deduplicating items within the same report**  
_This is what was historically supported by the COVID Pipeline._ 
Sender submits Report1 with items A, B, C, A within it. The deduplication feature will remove any subsequent instances of item A.  

>**3. Deduplicating items across reports**  
Sender submits two reports: Report1 with items A, B, C & Report2 with items D, E, A. The deduplication feature will detect that item A is a duplicate across multiple submitted reports and remove any subsequent instances of item A (in this case, the one that was submitted within Report2). Items D and E will continue to be processed. 
 
Example: As of Nov 2024, SimpleReport does not have a deduplication feature. An organization that uses SimpleReport’s _CSV Uploader_ adds new test result data by appending new rows to a previously submitted CSV file (new + old content) and uploads it to SimpleReport. SimpleReport will then submit this report to ReportStream (assigning a new message header for each upload instance). 

**XXX  SCENARIOS NOT SUPPORTED  XXX** 
>**4. Deduplicating items across senders**  
Sender ABC submits Report1 [with items A, B, C]. Sender XYZ  submits Report1 [with items A, B, C]. Both reports from both senders are exactly identical (duplicates). A deduplication feature will detect, not only all the items as duplicates, but also the reports as duplicates. Subsequent instances are removed.

This MVP will only deduplicate on an individual sender basis. 

 >**5. Deduplicating observations across items**  
Sender submits Report1 with items A and B. Item A contains test result observations for COVID, flu, and RSV. Item B contains the same COVID and flu test result observations (for the same patient, test performed, time, etc.) and an additional observation for syphilis. A deduplication feature will detect that the COVID and Flu observations from Item A and B are identical and remove any subsequent instances (in this case, the ones from item B). The syphilis observation from Item B will continue to be processed by the pipeline. 

>**6. Deduplicating across systems**  
ReportStream sends Report1 with items A, B, C to a Public Health Dpt. AIMS sends Report2 with items D, E, A. A deduplication feature (mostly likely on the PHD’s side) will detect item A as a duplicate and remove subsequent instances of it.

## Product Requirements
1. Once a duplicate message is detected, prevent it from being further processed. 

2. All incoming items of a report should be compared to all items submitted in at least the **past year**. This excludes data submitted prior to the implementation of the deduplication feature.

3. Duplication detection can be configured 'on' or 'off' for a specified sender (at the specific sender level, not sender org level). The default for senders is to have their deduplication configuration as 'on', which means deduplication is active. However, if the sender's settings for dedupe is configured as 'off', the message contents will still be hashed in event that the sender configuration will be toggled back to 'on' in the future.  

4. Senders shall not be able to change their deduplication configuration without the intervention of/ request through a ReportStream admin. 

5. If an Item is determined to be a duplicate, ReportStream shall log (application log AND action log) an error that an item was rejected. The log message should be: ERR, UNKNOWN, "Duplicate message was detected and removed." 

6. For the MVP iteration, the deduplication feature will apply only to test results (ORU_R01) data types. However, one must keep in mind that there may be future iterations that implement deduplication for other data types, such as test orders (for ETOR). 
   

## Acceptance criteria   
- [ ] All the requirements above are met in staging and production 
- [ ] Run a series of tests that represent the supported scenarios above 
- [ ] Each test scenario will simulate at least 5 duplicate items/ reports being detected and removed 
- [ ] Confirm that the duplications are being logged correctly (syntax and timing) 
- [ ] Set deduplication feature ‘off’ for all senders under the ELIMS org.  
- [ ] Prior to deployment, notify Deliver and Product Mgrs of potential deduplication dependencies with active pilot sender partners. 
- [ ] Technical design and final implementation are documented 
 

## Technical Design Considerations
- De-duplication in the covid pipeline currently occurs in the Receive function, but that likely does not make sense for the UP.  It's worth considering if the logic in the UP should live somewhere else (i.e. convert step) 

- The current logic for de-duping needs to be reconsidered for the UP; likely the item hash should be generated from the FHIR bundle so that the logic only needs to get written once - the issue here is if mappings change then FHIR bundle will be different even if original message is the same... 

- What if a message fails because of a bug on our side? If they resent, would hashing see it as the same message and filter out. 
- If a value is blank or not provided, consider how the hash will represent those blank fields (so that it can compare other messages with the same blank fields). 
- Consider how to purge stored hashes once they have "expired" after a year.
- Should MessageID uniqueness be enforced here as well? 
- In the case of SimpleReport, some unique IDs (accession number [specimen ID], CLIA number, etc.) are not guaranteed to be unique or validated due to some users of SimpleReport not having mature systems to generate and track them. If that is the case for SR, it could be true for other senders as well, making deduplication strategies less-effective or error prone. To mitigate accidentally detecting and removing non-duplicates, we recommend being more strict (comparing across many more data elements) with the deduplication criteria. This theoretically allows for more duplicates to pass through RS, but lessens the risk of accidentally removing data that is valid.
- Verify that the deduplication criteria works for all types of ELR messages (e.g., ELIMS, RADxMARS, full-ELR) and doesn't accidentally remove non-duplicates.
  

**Criteria for an item (ORU-0R1 message) to be considered a duplicate** 
Combine all the following into a hash. Should the hashes match, remove the corresponding item(s) that correlate to the subsequent instance(s). 
| Data Element                                                           | HL7                             | FHIR                                                                                                       |
|------------------------------------------------------------------------|---------------------------------|------------------------------------------------------------------------------------------------------------|
| Specimen ID                                                            | SPM.2                           | Specimen.Identifier                                                                                        |
| Accession number                                                       | SPM.30 (not v2.5.1 compatible)  | Specimen.accessionIdentifier                                                                               |
| Specimen Collection Date/Time (if different from other date/time)      | SPM.17                          | Specimen.collection.collectedDateTime                                                                      |
| Patient ID                                                             | PID.3                           | Patient.identifier                                                                                         |
| Patient name                                                           | PID.5                           | Patient.name                                                                                               |
| Patient DOB                                                            | PID.7                           | Patient.birthDate birthDate.extension[1].valueDateTime                                                     |
| Results Rpt/Status Chng – Date/Time                                    | OBR.22                          | DiagnosticReport.issued                                                                                    |
| Result Status                                                          | OBR.25                          | DiagnosticReport.status                                                                                    |
| Performing Organization/ Testing facility CLIA                         | OBX.23                          | Observation.Performer -> Organization.Identifier.value                                                     |
| Performing Organization/ Testing facility name                         | OBX.23                          | Observation.Performer -> Organization.name                                                                 |
| Test performed code                                                    | OBX.3.1                         | Observation.resource.code.system                                                                           |
| Test performed code system                                             | OBX.3.3                         | Observation.resource.code.coding                                                                           |
| Date/Time of the Observation (this appears in multiple HL7 locations)  | OBX.14, OBR.7, SPM.17           | Observation.resource.issued <br>DiagnosticReport.effectiveDateTime DiagnosticReport.effectivePeriod.start  |
| Observation Value/ Test result code                                    | OBX.5                           | Observation.resource.valueCodeableConcept.coding.code                                                      |
| Observation Value/ Test result code system                             | OBX.3.3                         | Observation.resource.valueCodeableConcept.coding.system                                                    |

JFisk42:
This PR is for the UP Deduplication design. There are some outstanding research items which could affect minor pieces of the described design.

## Changes
- Added UP Deduplication Design Documentation

## Linked Issues
- Fixes #14103

## Outstanding TODOs 
_Should be resolved before merge_:  
- [x] Database processing efficiency investigation.
- [x] Is sender name collision a concern?
  - A: No, sender names are unique to an organization. 
- [x] Product Q: Are Azure Event logs needed for instances of duplicate detection?
  - A: From a platform perspective this is not needed
- [x] Engagement Q: Are Azure Event logs needed for instances of duplicate detection?
  - A: Engagement would like an Azure Event logged with parent_report_id, child_report_id, and sender org/name.
- [x] Update UP SRD with dedupe requirements/convert step changes as necessary.
  - A: This will be completed as part of the implementation.


prime-router/docs/design/design/deduplication/deduplication.md
```diff
@@ -0,0 +1,197 @@
...
+
+## Introduction
+
+Refer to the [UP Deduplication epic](https://app.zenhub.com/workspaces/platform-6182b02547c1130010f459db/issues/gh/cdcgov/prime-reportstream/14103) for full background information.
+
+#### Technical Overview
+Deduplication shall be performed at the item level (a FHIR Bundle) after conversion to FHIR happens. Each bundle will have key elements (see [Key FHIR Elements table](#key-fhir-elements)) converted to a string which is then hashed deterministically and stored in the database (item_lineage.item_hash). That hash will only be compared against other hashes from the same sender that were created within the last year.
+
+#### Scope for Initial Implementation
+Deduplication will only be implemented for ORU_R01 messages from FULL_ELR topic senders. Covid Pipeline deduplication functionality should remain unchanged.
```
arnejduranovic, line 12:
```suggestion
Deduplication will only be implemented for ORU_R01 messages from FULL_ELR topic senders. Covid Pipeline deduplication functionality shall remain unchanged.
```

prime-router/docs/design/design/deduplication/deduplication.md
```diff
@@ -0,0 +1,197 @@
+
+# Universal Pipeline Deduplication
```
arnejduranovic, line 2: Consider adding a new section at the bottom that has the proposed convert step SRD changes so those can be reviewed as well. If these changes are only to the convert step, then I would copy the convert step requirements, amend them, and label each new or changed req with "(NEW)" tag.

prime-router/docs/design/design/deduplication/deduplication.md
```diff
@@ -0,0 +1,197 @@
...
+## Deduplication Workflow Design
+The Convert Step is where the UP’s deduplication workflow will live. See [Deduplication Workflow Placement](#deduplication-workflow-placement) for it's context within the Convert Step.
+
+### Convert Step
+
+##### Item Hash Generation
+The first part of the Deduplication Workflow will take in the item’s Bundle and use the specified fields ([Key FHIR Elements table](#key-fhir-elements)) to generate a string which is then hashed. The current hashing implementation ([Report.getItemHashForRow()](https://github.com/CDCgov/prime-reportstream/blob/15648395efc2b60322d931bf88e0c2c5b6cc0371/prime-router/src/main/kotlin/Report.kt#L1287-L1289)) uses Java’s [MessageDigest](https://docs.oracle.com/javase/8/docs/api/java/security/MessageDigest.html) to create a secure one-way hash of the string with SHA-256. This class also satisfies the requirements for UP Deduplication.
+
+Technical Considerations:
+- This should be flexible enough that different senders, message types, and topics can implement different key fields.
```
arnejduranovic, line 29: How will this be achieved? I.e. Strategy design pattern? Something similar to how convert step does item [validation](https://github.com/CDCgov/prime-reportstream/blob/25f5e2bba34c67f4033c8ffaa67d2ee8406b83f5/prime-router/src/main/kotlin/validation)?

prime-router/docs/design/design/deduplication/deduplication.md
```diff
@@ -0,0 +1,197 @@
...
+- The ordering of fields within a FHIR Bundle are **not** guaranteed. Before converting to a string to be hashed, the [populated key fields](#key-fhir-elements) should be extracted from the bundle and put in a static order (alphabetical etc.).
+- Sender id/name needs to be incorporated into pre-hashed string. (Dependent upon sql efficiency investigation noted below).
```
arnejduranovic, line 31:
```suggestion
- Sender id/name shall be incorporated into the pre-hashed string. (Dependent upon sql efficiency investigation noted below).
```

prime-router/docs/design/design/deduplication/deduplication.md
```diff
@@ -0,0 +1,197 @@
...
+
+##### Item Hash Comparison
+Hash comparison will be skipped if deduplication is disabled on that sender's settings. The generated item hash shall first be compared with items in the same report. (Note: At this part of the Convert Step, parallelization is not a blocker as the original items are [still within scope](https://github.com/CDCgov/prime-reportstream/blob/15648395efc2b60322d931bf88e0c2c5b6cc0371/prime-router/src/main/kotlin/fhirengine/engine/FHIRConverter.kt#L510)). The item hash will then be compared to existing hashes from the item_lineage table. There is an [existing SQL Query](https://github.com/CDCgov/prime-reportstream/blob/9ec0a59c73d7dad9a319cd321baf9efd71ceab46/prime-router/src/main/kotlin/azure/DatabaseAccess.kt#L166-L183) which performs the hash comparison through a database query. This will need to be enhanced or recreated as the original query only searches within the last 7 days and does not take sender into account. (_This last part will be unnecessary if the sender is incorporated into the hash._)
```
arnejduranovic, line 42: Define the setting name

prime-router/docs/design/design/deduplication/deduplication.md
```diff
@@ -0,0 +1,197 @@
...
+    "errors": [
+        {
+            "scope": "item",
+            "indices": [
+                1
+            ],
+            "trackingIds": [
+                "849b3151-25f7-41f3-b19b-fd8ad47bae18"
+            ],
+            "message": "Report could not processed. Reason: Duplicate report was detected and removed.",
```
arnejduranovic, line 133:
```suggestion
            "message": "Report could not processed. Reason: Duplicate report was detected and removed (no unique test results were found in the report.",
```

or some version of the above that sheds light on what a duplicate report actually is. This might be a good thing for @jsutantio to review.

prime-router/docs/design/design/deduplication/deduplication.md
```diff
@@ -0,0 +1,197 @@
...
+    "warnings": [],
+    "destinationCount": 0,
+}
+```
+
+### Deduplication Workflow Placement
+
+<BR><img alt="Dedupe_Design.jpg" height="310" src="Dedupe_Design.jpg" width="450"/>
+
+- The deduplication workflow will happen at the very end of the convert step processing ([FHIRConverter.kt#L541](https://github.com/CDCgov/prime-reportstream/blob/c942a9a6f6be347d82196939e1cf677512af4a06/prime-router/src/main/kotlin/fhirengine/engine/FHIRConverter.kt#L541)).
```
arnejduranovic, line 146: According to your diagram it does not happen "at the very end"

prime-router/docs/design/design/deduplication/deduplication.md
```diff
@@ -0,0 +1,197 @@
...
+  - Before the Deduplication Workflow, the Convert Step will:
+    - Convert to FHIR, if necessary
+    - De-batched items from the report
```
arnejduranovic, line 149:
```suggestion
    - De-batch items from the report
```

prime-router/docs/design/design/deduplication/deduplication.md
```diff
@@ -0,0 +1,197 @@
...
+    - Parse these items into bundles
```
arnejduranovic, line 150:
```suggestion
    - Convert the items into FHIR bundles
```

prime-router/docs/design/design/deduplication/deduplication.md
```diff
@@ -0,0 +1,197 @@
...
+    - Validate the items
```
arnejduranovic, line 151: This happens before converting, no?

arnejduranovic (commented):
Looks great! Please address TODOs before merging.

prime-router/docs/design/design/deduplication/deduplication.md
JFisk42: I'll update the wording here. This is referring to the FHIRConverter method process() which does the everything listed before deduplication. After that method there are still FHIRConverter actions taken but they're more related to informational actions (updating the db, action logging, azure events, queueing, etc).

prime-router/docs/design/design/deduplication/deduplication.md
JFisk42: Ah thanks for the callout. I should have also pointed out the content of the warning/error messages when asked this morning what @jsutantio should review here. Personally, I don't love what I came up with, but I am somewhat ambiguous on what the right vocab is for the submission history API audience.

prime-router/docs/design/design/deduplication/deduplication.md
JFisk42: Good catch. 
Looking over these lines a bit closer it appears that was the only thing out of order. I'll move it to before the convert.

prime-router/docs/design/design/deduplication/deduplication.md
JFisk42: Added Convert step changes under "Proposed Universal Pipeline SRD Additions", please review those when you have time.

Also noticed that there is some related language under 
```
Submission Step Business Logic Requirements 
    6.4. The submission step shall NOT accept the other custom HTTP Headers currently defined in RequestFunction.kt:
        6.4.4. AllowDuplicate – no clear requirements for this in UP at this time
```

This might also be a @jsutantio question. From my recollection we've been pretty adamant about not allowing senders to avoid this workflow on a per request basis. If that is still true then I would want to update this description to reflect that.

prime-router/docs/design/design/deduplication/deduplication.md
JFisk42: Somehow I previously missed that we should be able to entirely repurpose the covid pipeline sender setting for this. I've updated this document to reflect that, as well as a couple other docs that made reference to it in the context of the UP. https://github.com/CDCgov/prime-reportstream/pull/17143/commits/798d27495092457188697aa05e99a021d1bd3a9b

prime-router/docs/design/design/deduplication/deduplication.md
JFisk42: I've thought about this a couple of times and I keep coming back to yaml files (when you work at the ~~yaml~~hammer factory everything looks like a nail). Ultimately, the key fields are just a list of static fhirpaths. If you're simply looking for more specificity to address this consideration, I could prescribe a class which has the responsibility to reads in yaml files and handles the string concatenation. Then it could later be enhanced to take in a yaml filepath if we wanted to go down that road.

We could use something more OOP like a pattern if there was anything they were responsible for outside of reading in the specific paths. Off the top of my head the only example justification I can think of is something along the lines of "what if a sender/topic/message type needs to violate the core requirements of UP dedup". Ex: For X sender we only look at past month of results. For Y topic we want to compare against all senders.

prime-router/docs/design/design/deduplication/deduplication.md
JFisk42: Spoke with @jsutantio after standup today and she confirmed that in the UP we do not want to enable this HTTP param which could override the sender settings for duplication. I have pushed a commit to add this section to proposed UP SRD changes:
```
6.4. [No Change] The submission step shall NOT accept the other custom HTTP Headers currently defined in RequestFunction.kt: 
    6.4.4. [UPDATED] AllowDuplicate – Not to be used for UP De-duplication
```

prime-router/docs/design/design/deduplication/deduplication.md
JFisk42: TLDR: I've updated Technical Consideration with some technical description and a short mock of the proposed class structure (I believe this is in fact the Strategy design pattern but I am rusty on patterns in Java).

Actually, after mulling over this strategy again I don't like the idea of using yaml files to host fhirpath for the key fields. Storing the business logic for the key fields in Kotlin will avoid having to define yet another yaml schema and having to load/store them for the convert step. The likelihood of needing sender or topic specific key fields seems very low. Additionally, this will reduce steps in this part of the process. (There will no longer need to be a call to a fhirpath util method to extract values before concatenation of each field, instead extraction and concatenation can occur at the same time.

JFisk42:
Key changes since last reviews:
* New DB recommendations based on discovery that only 5% of item_lineage.item_hash values are necessary for deduplication. (See "Other Updates")
* Added proposal for 2-step rollout. Talked with Jessica and having some time to gather item_hash data to analyze potential false positives would be a good safety check.
* Added UP SRD update proposals
* Documented existing sender setting for enabling duplicate detection
* Addition of Azure Events for engagement team
* Mock class design for containing key field business logic

prime-router/docs/design/design/deduplication/deduplication.md
```diff
@@ -0,0 +1,248 @@
...
+Deduplication shall be performed at the item level (a FHIR Bundle) after conversion to FHIR happens. Each bundle will have key elements (see [Key FHIR Elements table](#key-fhir-elements)) converted to a string which is then hashed deterministically and stored in the database (item_lineage.item_hash). If the sender has deduplication enabled, that hash will only be compared against other hashes from the same sender that were created within the last year.
+
+#### Scope for Initial Implementation
+Deduplication will only be implemented for ORU_R01 messages from FULL_ELR topic senders. Covid Pipeline deduplication functionality shall remain unchanged.
+
+#### Deduplication Differences between Covid and Universal Pipeline
+ There are three key differences between the deduplication designs. In the Universal Pipeline,
+- Deduplication will now happen during the Convert Step instead of the Receive Step.
+- Deduplication will only consider a hash of key fields instead of the entire submitted item.
+- Deduplication will only consider hashes from the same sender in the last year.
```
stevewassnava, line 18: question: does this mean the calendar year or a year to date? I think this means to say within the past year to date (trying to think of a more concise way to send) - #2 below

_e.g., hash X sent 2/6/25, do we consider_
1. hashes from 2/6/24- 2/6/25 **OR**
2. hashes from 2024

prime-router/docs/design/design/deduplication/deduplication.md
```diff
@@ -0,0 +1,248 @@
...
+
+Technical Considerations:
+- This should be flexible enough that different senders, message types, or even topics can implement different key fields.
+  - This will be achieved through a new base class which shall own the logic for concatenating strings values in a map. Each set of new key fields would implement its own class which owns the logic knowing how to retrieve the necessary FHIR values. [See: Example Key Fields Class](#example-key-fields-file-structure) 
+  - If deemed necessary, sender specific key field classes could be created and then referenced in the sender settings with a new sender field.
+- The ordering of fields within a FHIR Bundle are **not** guaranteed. Before converting to a string to be hashed, the [populated key fields](#key-fhir-elements) should be extracted from the bundle and put in a static order (alphabetical etc.).
+- Sender id/name shall be incorporated into the pre-hashed string. (Dependent upon sql efficiency investigation noted below).
+- Investigation TODO (_To be removed and appropriate sections updated before merge_): Investigate the time efficiency of the following options:
```
stevewassnava, line 34: is this still the case - to update before the merge of THIS PR, correct?

prime-router/docs/design/design/deduplication/deduplication.md
```diff
@@ -0,0 +1,248 @@
...
+  - Adding the sender id (or name) into the string to be hashed and allowing the column index to be the main SQL query parameter.
+
+##### Item Hash Comparison
+Hash comparison will be skipped if `sender.allowDuplicates` is set to true. Otherwise, the generated item hash shall first be compared with items in the same report. (Note: At this part of the Convert Step, parallelization is not a blocker as the original items are [still within scope](https://github.com/CDCgov/prime-reportstream/blob/15648395efc2b60322d931bf88e0c2c5b6cc0371/prime-router/src/main/kotlin/fhirengine/engine/FHIRConverter.kt#L510)). The item hash will then be compared to existing hashes from the item_lineage table. There is an [existing SQL Query](https://github.com/CDCgov/prime-reportstream/blob/9ec0a59c73d7dad9a319cd321baf9efd71ceab46/prime-router/src/main/kotlin/azure/DatabaseAccess.kt#L166-L183) which performs the hash comparison. This will need to be enhanced or recreated as the original query only searches within the last 7 days and does not take sender into account. (_This last part will be unnecessary if the sender is incorporated into the hash._)
```
stevewassnava, line 39: would it be better to note the existing name of the function /variable/etc. in addition to the link to the specific line numbers? That is perfect for right now, but those references can easily get out of date when files are updated / altered in the future (and may not cause an update for this file)

prime-router/docs/design/design/deduplication/deduplication.md
JFisk42: Correct, Arnej [added an index](https://github.com/CDCgov/prime-reportstream/pull/17279) to one of the columns on this table. Basically, using metabase to determine the db efficiency needs could not be done because all the queries were dying.

prime-router/docs/design/design/deduplication/deduplication.md
JFisk42: I try to use permalinks for things like this (let me know if there are spots where I am not). GitHub can create permalinks to a line on any branch or commit to prevent against this very thing. (Actually... looking now, if you're on a branch the most recent commit will be what is referenced. I want to say that permalinks should also be able to refer to commits that are "wiped out" by things like a rebase/force push where that commit is no longer in history, but now that I tried to google the answer I didn't see an answer either way.)
Screenshot 2025-02-06 at 11 30 43 AM

prime-router/docs/design/design/deduplication/deduplication.md
JFisk42: Yeah this is a good point about the ambiguity. It may be better documented on the epic but I'll update the wording here. The answer is #1 here so I'll update it to be "from the same sender _within_ the last year" or could do something a slightly more verbose if that isn't sufficient either.

prime-router/docs/design/design/deduplication/deduplication.md
stevewassnava: thank you Josh- will have to note this for my future reference - sounds good!

prime-router/docs/design/design/deduplication/deduplication.md
```diff
@@ -0,0 +1,248 @@
...
+- Database Changes: Several changes will need to take place to improve efficiency. **Over 90% of the values stored in item_lineage.item_hash are not relevant to either CP or UP deduplication.**
+  - Immediate change: Modify item_lineage.item_hash to accept NULL values.
+    - Reasoning: 90%+ of item_lineage table rows have no need to store the item_hash. Allowing null values will reduce updates to the index and indexing times for the table. 
+  - Long term change: Modify the index on item_lineage to a partial index, on only those values which are not null.
+    - Caveat: Barring a significant database wipe of out of date information, this would likely result in a significant _decrease_ in database efficiency.
+    - A hybrid plan would be to begin storing NULL values. After older values have been culled, and/or enough NULL values have accumulated in the table, then change the index to a partial index. 
+- Database: Investigation will need to happen on how to best purge out of date records on the item_lineage table. item_lineage.created_at will be used for this. This is outside the scope for this design/implementation.
```
arnejduranovic, line 199: If there is an index on created_at performance should not be affected that much if we just limit our select query by created_at?

prime-router/docs/design/design/deduplication/deduplication.md
```diff
@@ -0,0 +1,248 @@
...
+  - The UniversalPipelineSender class [should default to false](https://github.com/CDCgov/prime-reportstream/blob/f48d719b876859169deb0360487f63965d8be5a0/prime-router/src/main/kotlin/Sender.kt#L196).
+- Receiver Step: There is stubbed code to be removed here: The SumissionReceiver refers to the once theorized area where UP deduplication would be invoked. ([SubmissionReceiver.kt#L284-L292](https://github.com/CDCgov/prime-reportstream/blob/0c5e0b058e35e09786942f2c8b41c1d67a5b1d16/prime-router/src/main/kotlin/SubmissionReceiver.kt#L284-L292))
+- Database Changes: Several changes will need to take place to improve efficiency. **Over 90% of the values stored in item_lineage.item_hash are not relevant to either CP or UP deduplication.**
+  - Immediate change: Modify item_lineage.item_hash to accept NULL values.
+    - Reasoning: 90%+ of item_lineage table rows have no need to store the item_hash. Allowing null values will reduce updates to the index and indexing times for the table.
```
arnejduranovic, line 195: Hmm, I wouldn't say they have NO reason to store these. They may not be relevant to deduplication but they do technically record the item in the DB long after it has been deleted from the blob store. This MAY be useful for debugging purposes but probably not... For example, someone says "I received this item from RS and its wrong" we could hash it and check against the DB if it is indeed what we sent or if it has been since modified

JFisk42: That's a fair point but I think this will have a positive effect on performance around this table, especially long term as more null values are added. I think pulling the report from blob storage would be better for this kind of scenario, but I'm not sure how long those stick around for.

prime-router/docs/design/design/deduplication/deduplication.md
JFisk42: Correct, and this was more aimed at the original notion of periodically purging the tables. I've removed this line as it isn't really relevant to the design here and added an "investigation results" bullet point under technical considerations with a brief summary of the db efficiency investigation and how created_at was relevant to that.

prime-router/docs/design/design/deduplication/deduplication.md
```diff
@@ -0,0 +1,248 @@
...
+  - During the Deduplication Workflow, the Convert Step will:
+    - Generate a hash for the item
+    - Compare this hash against those batched in the same submitted report
+    - Compare this hash with others stored in the database
+    - Create appropriate action logs and Azure Events if any duplicates were detected
+  - After the Deduplication Workflow occurs, the Convert Step will perform necessary pipeline functions:
+    - Apply sender schema transform
+    - Make a report object with item lineage
+      - **This will now include the newly generated (or null) hash**
+    - Create a route event
```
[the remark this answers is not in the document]
JFisk42: Ope, this was referring to the convert step's existing azure events. But it made me realize I hadn't named the new azure event(s). I've added that under the item hash comparison section where the azure event is described.

prime-router/docs/design/design/deduplication/deduplication.md
```diff
@@ -0,0 +1,248 @@
...
+  - Before the Deduplication Workflow, the Convert Step will:
+    - Validate the items
+    - Convert to FHIR, if necessary
+    - De-batch items from the report
+    - Convert these items into FHIR bundles
+    - Stamp Observations with condition codes
+  - During the Deduplication Workflow, the Convert Step will:
+    - Generate a hash for the item
+    - Compare this hash against those batched in the same submitted report
+    - Compare this hash with others stored in the database
```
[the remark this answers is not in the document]
JFisk42: Added.

prime-router/docs/design/design/deduplication/deduplication.md
```diff
@@ -0,0 +1,248 @@
...
+    "warnings": [
+        {
+            "scope": "item",
+            "indices": [
+                1
+            ],
+            "trackingIds": [
+                "849b3151-25f7-41f3-b19b-fd8ad47bae18"
+            ],
+            "message": "Duplicate message was detected and removed.",
```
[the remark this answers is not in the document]
JFisk42: Now that's an interesting question. That should be relatively easy to do but I'll let @jsutantio weigh in on if she wants that or any other details here.

prime-router/docs/design/design/deduplication/deduplication.md
```diff
@@ -0,0 +1,248 @@
...
+
+### Convert Step
+
+##### Item Hash Generation
+The first part of the Deduplication Workflow will take in the item’s Bundle and use the specified fields ([Key FHIR Elements table](#key-fhir-elements)) to generate a string which is then hashed. The current hashing implementation ([Report.getItemHashForRow()](https://github.com/CDCgov/prime-reportstream/blob/15648395efc2b60322d931bf88e0c2c5b6cc0371/prime-router/src/main/kotlin/Report.kt#L1287-L1289)) uses Java’s [MessageDigest](https://docs.oracle.com/javase/8/docs/api/java/security/MessageDigest.html) to create a secure one-way hash of the string with SHA-256. This class also satisfies the requirements for UP Deduplication.
+
+Technical Considerations:
+- This should be flexible enough that different senders, message types, or even topics can implement different key fields.
+  - This will be achieved through a new base class which shall own the logic for concatenating strings values in a map. Each set of new key fields would implement its own class which owns the logic knowing how to retrieve the necessary FHIR values. [See: Example Key Fields Class](#example-key-fields-file-structure)
```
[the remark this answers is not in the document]
JFisk42: Added a suggestion to use `LinkedHashMap` for this reason. I think there's probably some other good ways to get the fields into an ordered string so I won't go more strict than suggesting this.

prime-router/docs/design/design/deduplication/deduplication.md
JFisk42: Added this now.

stevewassnava: confirmed- update looks good- resolving

arnejduranovic (approved):

PR #17186, opened 2025-01-27
Devsecops/matts/workflow fixes

devopsmatt:
This PR reverts ubuntu semantic version pin back to latest.


emvaldes (approved):
rubber stamping this code as it seems pretty trivial and from a trusted source.

victor-chaparro (approved):

PR #17224, opened 2025-01-30
Update prepare_deployment_branch.yaml

devopsmatt:
fixed callable action names


scott-aquia (approved):
I have reviewed this change from a security perspective

PR #17334, opened 2025-02-12
[16142] Implement Updated ReceiverFilter Step Requirements

wcutshall:
This PR confirms the receiver filter step of the ReportStream pipeline.  Additionally, it introduces the filtering mechanism that will allow for the migration away from the current approach for quality, etc. filters.

Test Steps:
1. Identify (or create) a receiver that you will be using for testing.
2. Run the pipeline locally (`../gradlew quickRun`)
3. Start with a clean environment (`../gradlew clearDB`, `../gradlew reloadSettings`, `../gradlew reloadTables`).
4. Create a new message or choose a message from the Postman collection (e.g., _Single HL7 Full ELR_) and post it to the running application.
5. In the **Azure Storage Explorer** find the resulting file(s) in the blob container and note the number in the **translate** report store.
6. Add one element (e.g., "false") to the **routingFilter** for the receiver in question.
7. Clean the environment (`../gradlew clearDB`, `../gradlew reloadSettings`, `../gradlew reloadTables`).
8. Post the message again.
9. In the **Azure Storage Explorer** find the resulting file(s) in the blob container and note the number in the **translate** report store.  It should not have increased in number.

## Changes
- You can view all the file changes [here](https://github.com/CDCgov/prime-reportstream/pull/17334/files).

### Testing
- [x] Tested locally?
- [x] Added tests?

### Process

## Linked Issues
- No linked issues.

## To Be Done
- No TBD items.

    - No new application endpoints, but a new Azure function
        - elr-fhir-receiver-enrichment
    - No.
    - No.
    - No.
    - No.
    - No.


prime-router/src/main/kotlin/fhirengine/engine/FHIRReceiverFilter.kt
```diff
@@ -117,16 +118,56 @@ class FHIRReceiverFilter(
...
+        @JsonProperty
+        val receiverOrg: String,
+        @JsonProperty
+        val receiverName: String,
+        val index: Int = 1,
+    ) : ActionLogDetail {
+        override val scope: ActionLogScope = ActionLogScope.item
+
+        override val message: String = "Item was not routed to $receiverOrg.$receiverName because it did not pass the" +
+            " $filterType. Item failed on: $filter"
```
stevewassnava, line 134: nit: is this split across lines due to screen width or another reason? seems unneeded to create 3 strings here

prime-router/src/main/kotlin/ReportStreamFilter.kt
```diff
@@ -10,9 +10,16 @@ import kotlin.reflect.full.memberProperties
...
 typealias ReportStreamFilter = List<String>
 typealias ReportStreamConditionFilter = List<ConditionFilter>
+typealias ReportStreamReceiverRoutingFilter = List<FHIRExpressionFilter>
 
 fun ReportStreamConditionFilter.codes(): List<String> = this.flatMap { it.codes() }
 
+data class FHIRExpressionFilter(
+    val filterName: String,
+    val filterDescription: String,
+    val filterExpressions: List<String>,
```
stevewassnava, line 20: is this an extra "," ?

prime-router/src/test/kotlin/FileSettingsTests.kt
```diff
@@ -72,6 +79,14 @@ class FileSettingsTests {
         val settings = FileSettings().also { it.loadOrganizations(ByteArrayInputStream(receiversYaml.toByteArray())) }
         val result = settings.findReceiver("phd1.elr")
         assertThat(result?.jurisdictionalFilter).isNotNull().hasSize(1)
+        assertThat(result?.routingFilters).isNotNull()
+        assertThat(result?.routingFilters!!.size).isEqualTo(2)
+        assertThat(result.routingFilters.first().filterName).isEqualTo("My fine filter type 1.")
```
stevewassnava, line 84: may be more of a question for @arnejduranovic  but are there any thoughts on putting the literals on the left side of the `isEqualTo()` ? maybe there are other checks present that avoid the `filterName` (for example) from being `null`?

prime-router/src/main/kotlin/fhirengine/engine/FHIRReceiverFilter.kt
wcutshall: ktlint complains about lines over 120 characters in length.  It's a style thing.

prime-router/src/main/kotlin/ReportStreamFilter.kt
wcutshall: Nope, Kotlin wants you to use the trailing comma.  A nice discussion here:  [Trailing Comma Support in Kotlin](https://www.baeldung.com/kotlin/trailing-comma-support)

prime-router/src/test/kotlin/FileSettingsTests.kt
wcutshall: In this case, I think the Kotlin compiler is smart enough to know that it cannot be null at this point in the code (note the lack of "?" and "!!" at this point.)

prime-router/src/main/kotlin/ReportStreamFilter.kt
stevewassnava: Wow, I definitely didn't know that- thanks for the tip!

prime-router/src/main/kotlin/fhirengine/engine/FHIRReceiverFilter.kt
```diff
@@ -117,16 +118,56 @@ class FHIRReceiverFilter(
         override val errorCode: ErrorCode = ErrorCode.UNKNOWN
     }
 
-    data class FilterDetails(val filters: List<String>, val filterType: ReportStreamFilterType)
+    class AlternativeReceiverItemFilteredActionLogDetail(
```
arnejduranovic, line 121: So it looks like this is required because the new filters we implemented are just `string` instead of type `ReportStreamFilterType` making it so you can't reuse the already existing `ReceiverItemFilteredActionLogDetail`. The issues I see are:
- The name: `AlternativeReceiverItemFilteredActionLogDetail` isn't that descriptive. What's next, `AlternativeAlternativeReceiverItemFilteredActionLogDetail`?
- There is redundancy between the two classes in their implementation

I think the following solution would be cleaner, what do you think?:

0. Delete `AlternativeReceiverItemFilteredActionLogDetail`
1. Change the implementation of the currently existing `ReceiverItemFilteredActionLogDetail` to take a string for the `filterType` param.
2. Update all old instantiations of this class to call `.name` on the `ReportStreamFilterType` that gets passed in, like so: ReportStreamFilterType.CONDITION_FILTER.name

Another option that I like less is to have one class with two constructors, so you don't need to have the weird `AlternativeReceiverItemFilteredActionLogDetail` class

prime-router/src/main/kotlin/fhirengine/engine/FHIRReceiverFilter.kt
```diff
@@ -117,16 +118,56 @@ class FHIRReceiverFilter(
...
+        override val scope: ActionLogScope = ActionLogScope.item
+
+        override val message: String = "Item was not routed to $receiverOrg.$receiverName because it did not pass the" +
+            " $filterType. Item failed on: $filter"
+
+        override val errorCode: ErrorCode = ErrorCode.UNKNOWN
+    }
+
+    interface FilterDetailsInterface {
+        fun theFilterType(): String
```
arnejduranovic, line 140: On this project we do not follow the convention of "the" and "a" and "my". You can just use lower camel case without the prefix.

prime-router/src/main/kotlin/fhirengine/engine/FHIRReceiverFilter.kt
```diff
@@ -117,16 +118,56 @@ class FHIRReceiverFilter(
         override val errorCode: ErrorCode = ErrorCode.UNKNOWN
     }
 
-    data class FilterDetails(val filters: List<String>, val filterType: ReportStreamFilterType)
```
arnejduranovic, line 120: See my other comment regarding treating filterType as a string and handling the enum (when required) on the client/caller side. If you just update this one, then you don't need the three extra classes you created?

prime-router/src/main/kotlin/fhirengine/engine/FHIRReceiverFilter.kt
```diff
@@ -164,9 +205,65 @@ class FHIRReceiverFilter(
...
+            )
+            if (result is FhirExpressionEvaluationResult.Failure) {
+                return ReceiverFilterEvaluationResult.Failure(result.failingFilter)
+            }
+        }
+
         return evaluateObservationConditionFilters(receiver, bundle, actionLogger, trackingId)
     }
 
+    private fun evaluateFhirRoutingFilters(
```
arnejduranovic, line 224: Similar comment to the others I have: Does it make sense to combine this with the already existing `evaluateFhirExpressionFilters`. One of the things that's confusing about this one is the other one is called `evaluateFhirExpressionFilters` but this is the one that actually takes in a FHIRExpressionFilter object.

arnejduranovic (commented):
Good work! Just a few comments to consider.

prime-router/src/test/kotlin/fhirengine/azure/FHIRReceiverFilterIntegrationTests.kt
```diff
@@ -414,7 +414,7 @@ class FHIRReceiverFilterIntegrationTests : Logging {
             assertThat(event.params).isEqualTo(
                 mapOf(
                 ReportStreamEventProperties.FAILING_FILTERS to listOf("%resource.code.coding.code='1234'"),
-                ReportStreamEventProperties.FILTER_TYPE to ReportStreamFilterType.CONDITION_FILTER,
+                ReportStreamEventProperties.FILTER_TYPE to ReportStreamFilterType.CONDITION_FILTER.name,
```
JFisk42, line 417: I think this was what Arnej was referring to in standup, but we should have integration tests for not filtering and filtering things out with the new routing filter setup. Looking over the integration scenarios we appear to have scenarios for everything but the regular routing filter so I guess this works out.

JFisk42 (commented):
Some code comments for now, and will post back about test steps in a little bit.

JFisk42:
Tested this out locally and everything worked as expected 👍 🚀 ✅ 

A couple FYIs for you @wcutshall in case you weren't aware. 
- After posting a report you can check the delivery status of it in the history submission api `http://localhost:7071/api/waters/report/<report-id-from-post-response>/history/`.
- If all you're doing is updating a filter in organizations.yml from true to false the only command you should need to pick up that change is `./gradlew primeCLI --args "multiple-settings set --input='settings/organizations.yml'"`

I did also see a couple interesting things that are probably out of scope for this change:
- The formatting and content of the filtered message feels awkward in a few ways. `"For development.DEV_ENRICHMENT, filter false[] filtered out item 1234d1d1-95fe-462c-8ac6-46728dba581c. \nFilter Type: ROUTING_FILTER Filter Args: [] \nFiltered Observation Details: null "`
- Feels weird that we would specifically have "Filtered Observation Details" in the message for any filter failure.
- `"errorCode": "UNKNOWN"` Actually, this one makes sense because I think either Angela or Jessica was explaining how a lot of our error coding for the history endpoint was unknown.


``` json
{
  "id": "b1414d90-e585-4d82-a2d8-96416435affc",
  "submissionId": 23,
  "overallStatus": "Not Delivering",
  "timestamp": "2025-02-14T22:48:21.586Z",
  "plannedCompletionAt": null,
  "actualCompletionAt": null,
  "sender": "development.dev-elims",
  "reportItemCount": 1,
  "errorCount": 0,
  "warningCount": 6,
  "httpStatus": 201,
  "destinations": [
    {
      "organization": "FOR DEVELOPMENT PURPOSES ONLY",
      "organization_id": "development",
      "service": "DEV_ENRICHMENT",
      "itemCount": 0,
      "filteredReportRows": [
        "For development.DEV_ENRICHMENT, filter false[] filtered out item 1234d1d1-95fe-462c-8ac6-46728dba581c. \nFilter Type: ROUTING_FILTER Filter Args: [] \nFiltered Observation Details: null "
      ],
      "filteredReportItems": [
        {
          "filterType": "ROUTING_FILTER",
          "filterName": "false",
          "filteredTrackingElement": "1234d1d1-95fe-462c-8ac6-46728dba581c",
          "filterArgs": [],
          "message": "For development.DEV_ENRICHMENT, filter false[] filtered out item 1234d1d1-95fe-462c-8ac6-46728dba581c. \nFilter Type: ROUTING_FILTER Filter Args: [] \nFiltered Observation Details: null "
        }
      ],
      "sentReports": [],
      "downloadedReports": []
    }
  ],
  "actionName": "receive",
  "externalName": null,
  "reportId": "b1414d90-e585-4d82-a2d8-96416435affc",
  "topic": "full-elr",
  "bodyFormat": "",
  "errors": [],
  "warnings": [
    {
      "scope": "item",
      "indices": [
        1
      ],
      "trackingIds": [
        "1234d1d1-95fe-462c-8ac6-46728dba581c"
      ],
      "message": "Item was not routed to development.DEV_ENRICHMENT because it did not pass the routing. Item failed on: false",
      "errorCode": "UNKNOWN"
    }
  ],
  "destinationCount": 0,
  "fileName": ""
}
```

prime-router/src/main/kotlin/fhirengine/engine/FHIRReceiverFilter.kt
wcutshall: good points.  Changes made.

prime-router/src/main/kotlin/Receiver.kt
```diff
@@ -100,6 +102,7 @@ open class Receiver(
         mappedConditionFilter: ReportStreamConditionFilter = emptyList(),
         reverseTheQualityFilter: Boolean = false,
         enrichmentSchemaNames: List<String> = emptyList(),
+        routingFilters: ReportStreamReceiverRoutingFilter = emptyList(),
```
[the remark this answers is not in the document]
wcutshall: Your evaluation is correct.

prime-router/src/main/kotlin/fhirengine/engine/FHIRReceiverFilter.kt
wcutshall: Changed/updated.

prime-router/src/main/kotlin/fhirengine/engine/FHIRReceiverFilter.kt
wcutshall: Changed/updated.

arnejduranovic (approved):

JFisk42 (approved):

PR #17346, opened 2025-02-13
bump sleep to 10s in vpn action

devopsmatt:
Testing to see if longer sleep is needed for vpn connection to be usable


scott-aquia (approved):
Approved - DevSecOps

PR #17387, opened 2025-02-20
Adding facility, lab and clia

### LINKED ISSUE #16400: Add Facility Name, Testing Lab Name, and Testing Lab CLIA to Azure Logging observability
### User Story
As a ReportStream Engineer,
I want to be able to query for more information (Facility Name, Testing Lab Name, Testing Lab CLIA) at the item level in Azure logs,
so that I can troubleshoot issues easier.

### Description/Use Case
Sometimes we receive reports that include multiple testing facilities (such as Carbon Health) and having the information of Facility Name, Testing Lab Name, and Testing Lab CLIA will allow us to identify specific messages easier and troubleshoot issues faster.

### Dev Notes
Currently, we track items at the Report level and Item level. The added observability with these fields we would like to be added at the item level.

Here is an example in `FHIRDestinationFilter.kt` where we are sending an Item event, specifically `ITEM_ROUTED`:

```kotlin
    reportEventService.sendItemEvent(
        eventName = ReportStreamEventName.ITEM_ROUTED,
        childReport = report,
        pipelineStepName = TaskAction.destination_filter
    ) {
        parentReportId(queueMessage.reportId)
        params(
            mapOf(
            ReportStreamEventProperties.RECEIVER_NAME to receiver.fullName,
            ReportStreamEventProperties.BUNDLE_DIGEST
                to bundleDigestExtractor.generateDigest(bundle)
        )
        )
        trackingId(bundle)
    }
```

As you can see, `BUNDLE_DIGEST` is passed to the `params` section of the event data, and the object returned by `generateDigest` looks like so:

```kotlin
data class BundleDigestLabResult(
    val observationSummaries: List<ObservationSummary>,
    val patientState: List<String>,
    val performerState: List<String>,
    val orderingFacilityState: List<String>,
    override val eventType: String,
) : BundleDigest
```

Therefore, this work probably revolves around expanding this data class and the factory that creates it: `FhirPathBundleDigestLabResultExtractorStrategy`

I confirmed with Victor that performer/testing lab and order facility data should be correlated. So that means we need a more complicated object to store the info than we use today, like so:

```kotlin
[
{
    performerName: "something"
    performerState: "NA"
    performerCLIA: "Dfds1234"
},
{
    performerName: "something2"
    performerState: "NA"
    performerCLIA: "a234o23"
}
]
```

```kotlin
[
{
    orderingFacilityName: ""
    orderingFacilityState: ""
},
{
    orderingFacilityName: ""
    orderingFacilityState: ""
}
]
```

### Acceptance Criteria
- [ ] A query in Azure Logs for any `Item` events includes the Facility Name, Testing Lab Name, and Testing Lab CLIA and those fields that can also be used in the query to target specific items.
- [ ] Azure Event documentation in LucidChart is updated accordingly: https://lucidgov.app/lucidchart/89edc4c3-695b-4324-be8e-6556dd9fec95/edit?invitationId=inv_a44a9145-fc99-49ca-9053-a873937c7b23
- [ ] An example query for "List of items from Facility Name 'X'" is added to the bottom of the Azure Events document in Github: https://github.com/CDCgov/prime-reportstream/blob/master/prime-router/docs/observability/azure-events.md

kant777:
This PR ...

This PR introduces orderFacilitySummary and performerSummary which includes the desired fields like Facility Name, Testing Lab Name, Testing Lab CLIA and we should be able to see these values in the azure event logs going forward

Test Steps:

1. ./gradlew test
2. ./gradlew testSmoke
3. ./gradlew testEnd2EndUp

## Changes
- This PR introduces `orderFacilitySummary` and `performerSummary` which includes the desired fields like `Facility Name`, `Testing Lab Name`, `Testing Lab CLIA` and we should be able to see these values in the azure event logs going forward
- Fixes all the unit and integrations tests that are required after introducing the change above. In most cases I had to remove  `sender.reference` field—not only because it was redundant with our updated data model, but also to fix test failures by eliminating mismatches between the expected and actual output.

### Testing
- [x] Tested locally? yes
- [x] Ran `./prime test` or `./gradlew testSmoke` against local Docker ReportStream container? yes
- [x] Added tests? yes

image

### Process
- [ ] Are there licensing issues with any new dependencies introduced? N/A
- [ ] Updated the release notes? N/A
- [ ] Database changes are submitted as a separate PR? N/A
- [ ] DevOps team has been notified if PR requires ops support? N/A

## Linked Issues
- Fixes #16400


prime-router/src/test/kotlin/azure/observability/bundleDigest/FhirPathBundleDigestExtractorStrategyTests.kt
```diff
@@ -58,7 +60,12 @@ class FhirPathBundleDigestExtractorStrategyTests {
                             )
                         )
                     ),
-                                listOf("VA"), listOf("MD"), listOf("DC"), "ORU_R01"
+                                listOf("VA"),
+                                listOf(PerformerSummary(performerState = "MD")),
```
JFisk42, line 64: We should update the test data to include the new fields being populated in the events.

prime-router/src/main/kotlin/azure/observability/event/OrderFacilitySummary.kt
```diff
@@ -0,0 +1,27 @@
...
+    val orderingFacilityName: String = UNKNOWN,
+    val orderingFacilityState: String = UNKNOWN,
+) {
+    companion object {
+        const val UNKNOWN = "Unknown"
+
+        /**
+         * Create an instance of [OrderingFacilitySummary] from a [Organization]
+         */
+        fun fromOrganization(requester: Base): OrderingFacilitySummary {
```
JFisk42, line 16: We should have unit tests for these methods with all the different logic going on, especially PerformerSummary.

JFisk42 (changes requested):
This needs some new unit tests, but otherwise seems fine. Also, is there a finalized example azure message documented somewhere so it can be verified locally? Ideally there would be test steps in the PR description too but I've done the azure testing locally so I can figure that part out. However it looks like several parts of the PR description were missed.

prime-router/src/main/kotlin/azure/observability/bundleDigest/FhirPathBundleDigestLabResultExtractorStrategy.kt
```diff
@@ -10,24 +12,26 @@ class FhirPathBundleDigestLabResultExtractorStrategy(private val context: Custom
 
     private val eventCodePath = "Bundle.entry.resource.ofType(MessageHeader).event.display"
     private val patientStatePath = "Bundle.entry.resource.ofType(Patient).address.state"
-    private val performerStatePath =
-        "Bundle.entry.resource.ofType(ServiceRequest)[0].performer.resolve().address.state"
-    private val orderingFacilityStatePath =
-        "Bundle.entry.resource.ofType(ServiceRequest)[0].requester.resolve().organization.resolve().address.state"
+    private val performerPath =
+        "Bundle.entry.resource.ofType(ServiceRequest)[0].performer.resolve()"
```
JFisk42, line 16: Doing manual testing right now but just wanted to report some findings. Apparently we do not map ServiceRequest.performer (and it is not in the mapping inventory, I think. ServiceRequest can be a little tricky with ORU messages). So performer state was never being returned and this won't return anything either.

JFisk42: Tracked this down and it appears this is a field that SimpleReport populates. ServiceRequest.performer is called out in the original message monitoring docs for this, but the issue is that no other senders will have data for that field. We can follow up next week when Victor is back to see if this needs to be expanded to handle logging for more than the SimpleReport use case.

prime-router/src/main/kotlin/azure/observability/event/PerformerSummary.kt
```diff
@@ -0,0 +1,54 @@
...
+                        ?.firstOrNull()
+                        ?.nameAsSingleString ?: UNKNOWN
+
+                    val performerState = performer.address
+                        ?.firstOrNull()
+                        ?.state ?: UNKNOWN
+
+                    // For Practitioner, identifier list may contain CLIA. You need to find it explicitly.
+                    val performerCLIA = performer.identifier
+                        ?.firstOrNull { it.system?.contains("CLIA", ignoreCase = true) == true }
```
JFisk42, line 31: Another issue that may also need to wait until next week when Victor is back in. All the Simple Report fhir examples in our repo do not have CLIA under identifier.system but instead under identifier.type.code, example:
```json
  "identifier": [
    {
      "use": "official",
      "type": {
        "coding": [
          {
            "system": "http://terminology.hl7.org/CodeSystem/v2-0301",
            "code": "CLIA"
          }
        ]
      },
      "value": "11D1111111"
    }
  ],
```

JFisk42 (approved):
Code and tests look good. I have a couple outstanding questions that I documented as comments:
1. Are we okay with only SimpleReport data populating these summaries in azure?
2. Does check for CLIA need to be moved from identifier.system to identifier.type.coding.code?

I was able to manually verify this by posting one of our [sample SimpleReport files](https://github.com/CDCgov/prime-reportstream/blob/main/prime-router/src/testIntegration/resources/datatests/FHIR_to_HL7/sample_SR_1_20230302-0001.fhir) and modifying the performer identifier to have `CLIA` in the identifier.system.

Results:
Screenshot 2025-03-11 at 3 31 22 PM

prime-router/src/main/kotlin/azure/observability/bundleDigest/FhirPathBundleDigestLabResultExtractorStrategy.kt
arnejduranovic, line 16: Should this path be looking at MessageHeader instead?

prime-router/src/main/kotlin/azure/observability/event/PerformerSummary.kt
JFisk42: Today we looked through an example of this in prod and it confirmed that "CLIA" is going to be communicated through `identifier.type.coding.code` and not `identifier.system`.

JFisk42 (approved):
Just a few minor things that need to be taken out, otherwise everything looks good here! I ran through the verification steps again and was able to get all the new data fields populating:

Screenshot 2025-03-25 at 3 50 44 PM

prime-router/src/test/resources/fhirengine/engine/valid_data.fhir
```diff
@@ -1 +1 @@
...1d0f48","meta":{"extension":[{"url":"http://ibm.com/fhir/cdm/StructureDefinition/source-data-model-version","valueString":"2.5.1"}],"tag":[{"system":"http://terminology.hl7.org/CodeSystem/v2-0103","code":"P"}]},"identifier":[{"value":"MCLSH"}],"name":[{"text":"Shaunderrios T McLaughlin","family":"McLaughlin","given":["Shaunderrios","T"],"suffix":["MD"]}]}},{"fullUrl":"Location/1671741861180479600.7d17c735-b974-4eb0-b795-8decf3ea9753","resource":{"resourceType":"Location","id":"1671741861180479600.7d17c735-b974-4eb0-b795-8decf3ea9753","meta":{"extension":[{"url":"http://ibm.com/fhir/cdm/StructureDefinition/source-data-model-version","valueString":"2.5.1"}],"tag":[{"system":"http://terminology.hl7.org/CodeSystem/v2-0103","code":"P"}]},"extension":[{"url":"https://reportstream.cdc.gov/fhir/StructureDefinition/point-of-care","valueString":"F.ER"}],"mode":"instance"}},{"fullUrl":"Endpoint/1666038441953190000.245100f2-cee6-4c1b-9f7c-9635ab87fe74","resource":{"resourceType":"Endpoint","id":"1666038441953190000.245100f2-cee6-4c1b-9f7c-9635ab87fe74","identifier":[{"system":"https://reportstream.cdc.gov/prime-router","value":"co-phd.full-elr-hl7"}],"status":"active","name":"full-elr-hl7"}}]}
```
arnejduranovic, line 1: Can you please comment what you changed here, and why, since the diff is hard to read and this file is used a lot? Ditto for the other updated files.

arnejduranovic (commented):
Changes and tests look good! Before approving, I would request that:
1. The PR description is properly updated (its just the template placeholder text in there right now)
2. Explain what changes were made to the .fhir test files, and why.

PR #17490, opened 2025-03-05
Update queue message config params in host.json

arnejduranovic:
This PR takes a first pass at tuning the Azure Queue Message parameters used by the Azure Function runtime to determine how queue messages are handled. These changes are being made in hopes of remediating the issue described in [this incident report](https://cdc.sharepoint.com/:w:/r/teams/ReportStream/_layouts/15/Doc.aspx?sourcedoc=%7B2E1169CA-B921-4CAA-B0E6-2BEF7443A1D7%7D&file=2025_02_18%20-%20Database%20Connectivity%20Issues.docx&action=default&mobileredirect=true). The following params are changed:
1. modified `batchSize`
2. modified `newBatchThreshold`
3. added `visibilityTimeout`
4. added `maxDequeueCount`

## Changes
1. Updated `batchSize` from 8 to 4 and `newBatchThreshold` from 4 to 2. This means _each queue triggered function_ can process at most 6 queue messages at a time (4+2). The idea here is to cut down on the number of functions running at a time and thus the number of database calls happening to help alleviate the observed connection pool starving under high loads.
5. Added `visibilityTimeout` to override the default of 0 to 5 minutes and . This means that if a queue-triggered function throws an exception, which generally only occurs in the UP if a database exception occurs, the queue message will not get rescheduled for 5 minutes. The idea here is to wait some time before trying functions that had failed DB calls to avoid overloading resources. With the `maxDequeueCount` set to 5, we will now retry these failed functions 5 times over 25 minutes (5 * 5) instead of 5 times over potentially a few seconds.

### DIFF
--- a/prime-router/host.json
+++ b/prime-router/host.json
@@ -6,8 +6,10 @@
   },
   "extensions": {
     "queues": {
-      "batchSize": 8,
-      "newBatchThreshold": 4
+      "batchSize": 4,
+      "newBatchThreshold": 2,
+      "visibilityTimeout": "00:05:00",
+      "maxDequeueCount": 5
     }
   },
   "logging": {


victor-chaparro (approved):
Nice!

PR #17494, opened 2025-03-06
[16288] Adding Submission History API documentation.

wcutshall:
This PR contains just written documentation for the Submission History API (in Markdown format).  Just one new file is included as part of the effort and no application changes were made.


arnejduranovic (approved):
Looks good to me. Added suggestions/corrections, please review before merging. Another overall comment is to use actual org names (or at least org names that are representative of actual org names) instead of names like "FOR DEVELOPMENT PURPOSES ONLY" to make this documentation more human centered/easily digestible.

PR #17515, opened 2025-03-09
Deploy tf

bethbeza:
TF Plan code added


.github/workflows/\.code-workspace
devopsmatt, line 1: We don't need this file.  Please remove from the PR.

.github/workflows/reflog
devopsmatt, line 1: We don't need this file. Please remove from the PR.

.github/workflows/deploy_terraform.yml
```diff
@@ -93,40 +121,46 @@ jobs:
...
           sp-creds: ${{ env.AZURE_CREDENTIALS }}
           tf-auth: true
       - name: Use specific version of Terraform
-        uses: hashicorp/setup-terraform@b9cd54a3c349d3f38e8881555d616ced269862dd
+        uses: hashicorp/setup-terraform@v1
         with:
-          terraform_version: 1.7.4
-          terraform_wrapper: false
-      - name: Run Terraform
+          terraform_version: 1.7.4
```
devopsmatt, line 152: Let's set [terraform_wrapper](https://github.com/hashicorp/setup-terraform) to false

.github/workflows/deploy_terraform.yml
```diff
@@ -14,24 +14,67 @@ env:
 jobs:
   pre_job:
     name: Set Build Environment
-    concurrency:
-      group: ${{ github.workflow }}-${{ needs.pre_job.outputs.env_name }}
+    concurrency: 
+      group: ${{ github.workflow }}-${{ github.run_id }}
```
devopsmatt, line 18: I am not sure run_id will match and cancel as expected.

.github/workflows/deploy_terraform.yml
```diff
@@ -93,40 +121,46 @@ jobs:
...
+        run: terraform init -input=false
+      - name: Run Terraform Sanity Checks
         run: |
-          terraform init -input=false
+          terraform fmt -check -recursive
           terraform validate
-          terraform fmt -recursive
-          terraform plan -out ${{ needs.pre_job.outputs.env_name }}-tf.plan
-          terraform apply -input=false -no-color -lock-timeout=600s -auto-approve ${{ needs.pre_job.outputs.env_name }}-tf.plan
+          terraform workspace select -or-create ${{ needs.pre_job.outputs.env_name }}
```
devopsmatt, line 159: I think it would be better to select/create before format check and validation

.github/workflows/deploy_terraform.yml
devopsmatt: This will work:
group: ${{ github.workflow }}-${{ needs.pre_job.outputs.env_name }}

PR #17583, opened 2025-03-17
Update sonarcloud.yml - remove refs to tj-actions

scott-aquia:
This PR comments out the sections in the SonarCloud workflow that use tj-actions

### DIFF
--- a/.github/workflows/sonarcloud.yml
+++ b/.github/workflows/sonarcloud.yml
@@ -46,19 +46,19 @@ jobs:
         with:
           fetch-depth: 0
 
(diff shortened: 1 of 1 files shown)


buck30wale:
@scott-aquia we should remove the vulnerability completely. We only commented out a duplicate instance of the action but left the first instance active.

buck30wale:
I've updated the PR to completely remove the vulnerable tj-actions/changed-files action as per the security advisory. All checks are passing now. Feel free to review the changes and merge when you're ready.

devopsmatt (approved):

PR #17596, opened 2025-03-18
Add CSV header check for code condition mapping endpoint

the-andrew:
This PR adds a check on the backend for the CSV received in the request body for the code condition mapping endpoint.
The CSV header is checked to make sure the 3 required columns are present.
Expected headers are:
- `test code`
- `test description`
- `coding system`

Test Steps:
1. Send HTTP request to `/api/sender/conditionCode/comparison` with CSV in request body with all 3 required headers
2. Send same request but with one or more headers missing

## Changes
- Slight change on backend to check request body CSV's headers against the list of 3 required headers

### Testing
- [ ] Tested locally?
- [ ] Added tests?

### Process

## Linked Issues
- Fixes #issue

## To Be Done
- #issue

    - new endpoint A
    - new endpoint B


prime-router/src/main/kotlin/azure/SenderFunction.kt
```diff
@@ -56,6 +56,25 @@ class SenderFunction(
...
             val bodyCsvText = request.body ?: ""
+
+            // Check CSV headers in request body
+            val csvHeader = bodyCsvText.lineSequence().firstOrNull()
+            if (csvHeader != null) {
+                val headersToCheck = csvHeader.split(",")
+
+                val missingHeaders = mutableListOf<String>()
+                val requiredHeaders = listOf("test code", "test description", "coding system")
+                for (header in requiredHeaders) {
```
etanb, line 67: @victor-chaparro is this all the validation we need for this CSV? do we want have that list of hard coded header strings to leave here? we can also do this validation on the F/E as well, but either we could mimic this directly `"test code", "test description", "coding system"` or if we expose the required headers from the B/E than we don't have to maintain two sources of truth.

victor-chaparro: There are some other pieces  we should check, but I'm not sure if we're doing that already
Here are some things we should check:

- File does not have any empty rows
- Every row has a code, description and coding system
- Every row has the correct number of columns

I’d rather have all these validations handled on the backend. Since the backend needs to process the file anyway, it makes sense to avoid having the frontend parse through all its contents unnecessarily.

victor-chaparro (approved):
Nice!

Issue #16818, opened 2024-12-16
Document Rollout Strategy for new AuthZ Strategy


arnejduranovic:
### User Story

As a user of RS endpoints,
I want a smooth transition to the new auth,
so that my business operations are not disrupted

### Description/Use Case

This ticket is to create a migration/rollout plan for the new AuthZ design defined in #16815 and #16816. The following questions should be addressed in detail:
    - Which endpoints will need to be updated? How will they be updated/what changes are needed?
    - Will RS support multiple auth methods at first, and how and for how long?
    - How does the Auth MicroService play a role? Does it need to be deployed first?
    - What does this migration look like from the sender/receiver/RSReact side? -- create preliminary user documentation!

### Acceptance Criteria

- Migration/rollout plan created and approved by:
    - [ ] @arnejduranovic 
    - [ ] @chris-kuryak

MichaelEsuruoso:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/CDCgov/prime-reportstream/16818?planning-poker) with Zenhub @adegolier @arnejduranovic @jack-h-wang @jalbinson @JFisk42 @kant777

arnejduranovic:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-6182b02547c1130010f459db/issues/gh/CDCgov/prime-reportstream/16818?planning-poker) with Zenhub @feralearthman

arnejduranovic:
We expect this story to carry over multiple sprints due to back and forth

PR #17619, opened 2025-03-21
17682 update auth and submissions group authz strategy

jack-h-wang:
This PR updates the auth/submission microservices to follow the current authz design and removes deprecated code supporting an older design.

This PR implements a stopgap solution to the need for Okta application users to have their group memberships as profile attributes. This temporary solution simply updates the Okta application user profile whenever it is authenticated so that the profile will always include the current list of groups. This will be removed when the services handling application user profiles is built.

Additionally, this PR includes a plan to transition from the current auth scheme to microservices.

## Linked Issues
- #17682
- #16818


prime-router/docs/authz/authz-design.md
```diff
@@ -168,13 +168,57 @@ fun setApplicationProfile(
...
+* Build a process within the auth project to retrieve an application user's group memberships and update their profile
+  accordingly. 
+  * This allows group management to occur within the Okta Admin Console, but a trigger to perform this step is needed.
+    * We can build a API call (triggerable via frontend and/or CLI) that takes a client ID of a sender as input and does
+      the needed profile updates on demand.
+    * We can configure a schedule trigger to scan all application users and update their profiles based on their Okta
+      group memberships periodically. This removes the potential for group memberships and sender profiles to fall out
+      of sync, but the refresh will happen on a delay and this would require the microservice to perform more work on an
+      ongoing basis.
+    * Using Okta event hooks to send a POST request to the auth microservice whenever an application's group membership
```
JFisk42, line 181: Not really relevant to the review itself, but (from my less-informed opinion) this is the option that makes the most sense to me. It sounds like once it is running it would have the least amount of administrative hassle when onboarding.

prime-router/docs/authz/authz-transition-plan.md
```diff
@@ -5,35 +5,78 @@ This document will detail the steps needed to successfully transition from the c
...
+  * Utilizing groups permits onboarding via Okta admin console, but the group memberships are redundant as it is only
+    the profile attributes that are actively used during the authz process
+  * Storing solely as profile attributes would require the sender management APIs to be built before transition to the
+    auth service can occur
+  * We can change this strategy at any time
 * Create test workflow in staging representative of production environment
   * Update workflow documentation to create new sender
+* Add auth/submissions endpoints to deployment and CI processes
+  * auth and submissions projects currently ignored
+  * end2end testing for endpoints should be created
```
JFisk42, line 18: Not sure if this needs to be called out here but the end2end_up test will have to be adjusted to work with the new auth. It hits the submission and history endpoints.

Edit: Looks like this partially depends on the question of if we will have both auth schemes running at once or not. I see there's also a comment about making a dupe of it for the microservices 👍 🚀

prime-router/docs/authz/authz-transition-plan.md
```diff
@@ -5,35 +5,78 @@ This document will detail the steps needed to successfully transition from the c
...
+    * Coordinating a cutover date with external partners may not be feasible
+  * Can designate a pilot partner before initiating a larger migration
+    * Good idea to have a pilot partner even if not preserving separate endpoints for new API 
 
 * Cons:
   * Transparent cutover not an option; new API must have different path
-  * Will require tests for both APIs to be maintained simultaneously
+  * Will require implementation and tests for both APIs to be maintained simultaneously
+  * Will likely require some level of reonboarding for all users of RS
   * Can both APIs coexist on the same listening port? Would this require the functionapp to act as a passthrough?
```
JFisk42, line 99: Also wanted to leave some musings on the SMART on FHIR question below. 

Context: Two jobs ago I worked on an EHR vendor's FHIR api and a sibling team of ours handled our companies SMART on FHIR implementation.

Overall, from what I recall, SMART on FHIR is irrelevant to our use case because it's mainly for a seamless auth workflow between apps. Ex: If a clinic was using Cerner or Epic's EHR system and wanted to send a report to report stream directly to us without an intermediary. So they would use the EHR to launch a report stream SMART on FHIR app and the authorization handshake would happen at that point. 

There's probably other use cases for it but just wanted to throw out my 2 cents.

JFisk42 (approved):
Synced up with Jack today to verify that the code is doing what it says it is. Left a few comments but nothing needs to be fixed. The code all makes sense and the docs look good!

auth/src/main/resources/application.yml
```diff
@@ -30,12 +28,8 @@ okta:
     clientId: 0oak8383risyLJY9W1d7
     requiredScopes:
       - okta.apps.read
+      - okta.apps.manage
     # set base 64 encoded pem string in OKTA_ADMIN_CLIENT_API_ENCODED_PRIVATE_KEY env variable
```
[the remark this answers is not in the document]
jack-h-wang: The code to retrieve group memberships was not removed and relies on this value to function. It is still exercised in this POC and is likely to be used in the final solution as well.

arnejduranovic (approved):

prime-router/docs/authz/authz-transition-plan.md
```diff
@@ -5,35 +5,48 @@ This document will detail the steps needed to successfully transition from the c
...
-  * Flexible timeline for transitioning senders; don't have to migrate all at once
-  * Can designate a pilot partner
+  * Flexible timeline for transitioning senders; no hard cutoff date, don't have to migrate all at once
+  * Can designate a pilot partner before initiating a larger migration
+    * Good idea to have a pilot partner even if not preserving separate endpoints for new API 
 
 * Cons:
   * Transparent cutover not an option; new API must have different path
   * Will require tests for both APIs to be maintained simultaneously
+  * Will likely require some level of reonboarding for all users of RS
```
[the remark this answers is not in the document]
jack-h-wang: I'm actually just going to remove this line, as reonboarding of senders will occur either way. I was being optimistic about being able to cut over without actually requiring senders to change anything.

prime-router/docs/authz/authz-transition-plan.md
```diff
@@ -5,35 +5,48 @@ This document will detail the steps needed to successfully transition from the c
 * Implement profile groups (API > ReportStream API > Claims > sender_name)
```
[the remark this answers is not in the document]
jack-h-wang: This actually refers to the configuration of claims returned in the bearer token, not any individual sender. Will edit to make that more clear.

prime-router/docs/authz/authz-transition-plan.md
```diff
@@ -1,41 +1,131 @@
...
+Microservices should be updated to provide the endpoints for Okta event hooks to push events to as well as the endpoint
+for application users to acquire bearer tokens via Okta. The ability to update application user profiles should also be
+built at this time, and hooked in to the Okta event hook API. We should also allow application user profile updates via 
+CLI.
+
+### Build passthrough API for functionapp APIs via auth microservice ###
+Calls to the APIs served via the functionapp (reports/waters) should be able to pass through the auth microservice so it
+is no longer necessary to expose the functionapp to the network. In the process we should evaluate the authorization
+requirements of the APIs so that unauthorized requests are immediately rejected rather than passed through. This will be
+needed particularly for endpoints intended for sender use; authorization for this needs to be handled by the
```
arnejduranovic, line 23: My latest understanding is each endpoint would deal with Authorization, but Authentication would be done on the Auth Microservice side? This contradicts what you have written here?

prime-router/docs/authz/authz-transition-plan.md
```diff
@@ -1,41 +1,131 @@
...
+functionapp. We will need to determine if this test should integrate with Okta or if Okta connections should be mocked
+in the absence of an offline test container. We should also aim for the new smoke test to be performed by a GitHub
+action rather than be executed from a developer machine; this will require storing secrets for this purpose.
+
+### Create application users for senders and begin migration outreach ###
+Senders are represented as application users. Creating the application user will produce a client ID and a private key
+that the sender will use to authenticate to the microservices. Begin outreach to at least one sender to have them begin
+using the microservices to submit instead of the functionapp.
+
+### Add group and scope assignments to existing Okta users ###
```
arnejduranovic, line 45: Does this need to happen before ### Create application users for senders and begin migration outreach ### and even before ### Build passthrough API for functionapp APIs via auth microservice ###?

prime-router/docs/authz/authz-transition-plan.md
```diff
@@ -1,41 +1,131 @@
...
+implementation of this API and the related frontend can occur in parallel to the other migration tasks.
+
+# Dev notes #
+These are miscellaneous dev notes that should be considered during the implementation process.
+
 * Migrate all senders in staging
-  * Can existing public keys be transferred to Okta, or will all senders have to be fully reonboarded?
-  * Do we intend to run both auth systems simultaneously?
+    * Can existing public/private keys be transferred to Okta, or will all senders have to be fully reonboarded?
+    * Do we intend to run both auth systems simultaneously?
```
arnejduranovic, line 71: Yes, the migration plan indicates this is so?

prime-router/docs/authz/authz-transition-plan.md
```diff
@@ -1,41 +1,131 @@
...
+        * Document differences between RS auth and proposed Okta auth (endpoints used, etc.)
+
+* Do sender and user migrations need to be performed simultaneously?
```
arnejduranovic, line 74: I don't think so? The "user migrations" are entirely for the RS website, right?

prime-router/docs/authz/authz-transition-plan.md
jack-h-wang: It's likely not feasible to update the authorization logic in the functionapp APIs as we can't introduce breaking changes until the migration to the auth microservice is complete. So we are probably going to have to do some level of authorization within the auth microservice to decide if the client requesting a given API should be allowed to make the request. I'm envisioning that we'll either have to introduce a translation layer that would work with the functionapp API authorization logic, or the auth microservice will have to do all the authorization logic. I updated the dev notes to make this more clear.

prime-router/docs/authz/authz-transition-plan.md
jack-h-wang: I don't think it's necessary to do the migration steps in order (and it's even possible to do them in parallel - there's very little blocking between them). The microservices simply won't be ready for use until the majority of steps are completed.

prime-router/docs/authz/authz-design.md
```diff
@@ -149,11 +149,11 @@ the specific senders. For users this can be done in the Okta UI.
 Senders are set up as application users within Okta to allow the machine-to-machine client credentials OAuth 2.0 flow. The 
 difficulty arises because while Okta allows you to add application users to groups, it does not allow you to easily add 
 those groups to the generated access token JWT. The workaround there is adding the necessary information to the Okta 
-application user's profile. This can only be done via API at the moment though.
+application user's profile. This can only be done via Okta's APIs at the moment though.
```
arnejduranovic, line 152:
```suggestion
application user's profile. This can only be done via Okta's APIs at the moment, though.
```

prime-router/docs/authz/authz-transition-plan.md
```diff
@@ -1,41 +1,136 @@
...
-  * Do we intend to run both auth systems simultaneously?
+    * Can existing public/private keys be transferred to Okta, or will all senders have to be fully reonboarded?
+* Implement sending profile attributes in claims (API > ReportStream API > Claims > create `appSubmit` claim)
+* API changes for new auth flow
+  * Which Azure functions require authorization processes?
+  * Sender onboarded to use auth microservice will have significantly different authorization flow from original API -
+    need strategy for how authorization decisions are made when passing API calls from microservices to functionapp
+    * Could use some sort of translation function to bridge new authorization to old in order to support legacy
+      endpoint before all senders are reonboarded
+    * Alternative would be to handle all authorization on auth microservice and bypass authorization on functionapp
```
arnejduranovic, line 83: Third option? : create v1/v2 alternate endpoints with new authz implementation that the Auth service will do AuthN for prior to forwarding the request. In Auth service, create a "whitelist" of the original endpoints and DO NO authN/authZ for these requests, simply pass them to the original endpoints and let the server2server auth code run.

prime-router/docs/authz/authz-transition-plan.md
```diff
@@ -1,41 +1,136 @@
...
+      end when auth microservice is the caller 
+* Should sender groups be stored as Okta groups or solely as profile attributes?
+    * Utilizing groups permits onboarding via Okta admin console, but the group memberships are redundant as it is only
+      the profile attributes that are actively used during the authz process
+      * Event hook strategy should keep group memberships and profile attributes in sync without much upkeep 
+    * Storing solely as profile attributes would require the sender management APIs to be built before transition to the
```
arnejduranovic, line 89: But if we do both methods then we still need the event hook based updater, right?

prime-router/docs/authz/authz-design.md
```diff
@@ -366,26 +366,88 @@ fun authorizeSubmit(
...
+
+## Server to server authentication via Okta ##
+
+Senders currently connect to a token endpoint provided by the functionapp (which utilizes the database to store keys).
+Going forward, senders will need to access Okta (or the request can be forwarded through the auth microservice) to
+obtain their bearer token. See [get_client_access_token.py](../../../auth/src/scripts/get_client_access_token.py) for an
+example of the bearer token request senders will need to develop. From there, senders can submit reports to the auth
+microservice, from which the submission microservice will handle the request. The submission microservice will utilize
+the submission storage table in Azure rather than the database, then send a request to the `convert` queue to initiate
+processing the report (there will not be a `receive` step for submissions handled by the submissions microservice).
```
arnejduranovic, line 453: This example is specific to submitting reports. Can we also include an example of hitting the history API since that is in the Azure Function app? I feel like we are skimming over what the auth would look like there? Perhaps a flow of current and desired auth (like in the okta user example above) is warranted here?

prime-router/docs/authz/authz-design.md
```diff
@@ -366,26 +366,88 @@ fun authorizeSubmit(
...
+developers. The claims included in the bearer token are determined by the configuration of the OAuth API. The bearer
+token is authenticated and processed within OktaAuthentication.kt utilizing an [AccessTokenVerifier](https://developer.okta.com/docs/guides/validate-access-tokens/java/main/)
+via the [OKTA JWT Verifier library](https://github.com/okta/okta-jwt-verifier-java). This library automatically caches
+the public keys (JWKS) used to validate the JWT at startup and when the keys are rotated.
+
+As part of migrating to a new auth design we will change the API calls requested by the frontend to route them through
+the auth microservice. The auth microservice will authenticate the request, then forward the request to the functionapp.
+The functionapp's handling of Okta authentication can be updated to conform to the new JWT claims design, and the 
+authorization can be switched to a new class processing the JWT claims instead of the current custom scope handling. 
+Note that the functionapp will have to remain available to the public internet until there are no more senders utilizing
```
arnejduranovic, line 407: This says function app needs to remain accessible to the public internet but another statement down below (see other comment) contradicts this?

prime-router/docs/authz/authz-transition-plan.md
```diff
@@ -22,10 +22,8 @@ updates via CLI.
 
 ### Build passthrough API for functionapp APIs via auth microservice ###
 Calls to the APIs served via the functionapp (reports/waters) should be able to pass through the auth microservice so it
-is no longer necessary to expose the functionapp to the network. In the process we should evaluate the authorization
-requirements of the APIs so that unauthorized requests are immediately rejected rather than passed through. This will be
-needed particularly for endpoints intended for sender use; authorization for this needs to be handled by the
-microservice. See [deprecated auth documentation](authz-deprecated-implementation.md) for a list of endpoints.
+is no longer necessary to expose the functionapp to the network. Requests should be passed verbatim, including the
```
arnejduranovic, line 25: This is contradicting a statement above (see other comment?)

prime-router/docs/authz/authz-transition-plan.md
```diff
@@ -52,16 +50,24 @@ Existing Okta users can be assigned to groups and scopes within Okta without aff
...
 the `admin` scope. New users created after beginning this effort will need to be onboarded with both the new and 
-deprecated scopes.
+deprecated scopes. If there are a very large number of users we may want to consider building a script that retrieves 
+the users via the Okta API then programmatically derives and applies the equivalent scopes.
+
+### Update functionapp APIs to utilize new authorization and authentication ###
+All APIs in the functionapp should be updated to use the updated authentication and authorization design. In the process
+we should evaluate the authorization requirements for each endpoint. We should consider splitting this effort as there
+are as many as 58 endpoints to be evaluated and updated. This work will need to be performed in a feature branch in
+order to not introduce breaking changes before the frontend API connections are updated.
```
arnejduranovic, line 60: Specifically for Senders (Submission and History APIs) Can you explain here why we don't need to support both new and old Authn/Authz as we transition senders to the new application auth?

prime-router/docs/authz/authz-transition-plan.md
```diff
@@ -52,16 +50,24 @@ Existing Okta users can be assigned to groups and scopes within Okta without aff
...
-### Update frontend claims authorization handling and API connections ###
-The frontend is changed to process user authorization based on the claim structure outlined in the design. After this is
-accomplished, the deprecated organization scopes can be retired. All references to the RS API are changed to be directed
-to the auth microservice. 
+### Update frontend API connections ###
+Once all Okta users and functionapp APIs have been updated as outlined above, all references to the RS API are changed
+to be directed to the auth microservice. This involves updating the Okta OAuth API to the new bearer token format.
 
 ### Retire API access to the functionapp ###
 Once all senders and users have been provisioned in Okta using the updated structures we can remove network access to
```
arnejduranovic, line 67: Again mentioning outside access to azure function app

prime-router/docs/authz/authz-design.md
jack-h-wang: I don't think the flow needs to be repeated here as it is much the same - we just have to identify which endpoints we expect to see server to server communication for, as the sender claims are structured differently and the authz will have to take that into account.

prime-router/docs/authz/authz-transition-plan.md
jack-h-wang: This is not the case; we will have to support both schemes unless we reimplement all server to server endpoints. We should consider at least finalizing API URIs that better segregate endpoints intended to be used in a server to server context and those intended for the frontend so that those changes can be made at the time senders change endpoints for token/submissions. We should be able to make authorization behave differently if called from a different URI handled by the auth microservice even if we call the same endpoint internally to actually produce the result.

prime-router/docs/authz/authz-design.md
jack-h-wang: I don't believe there is a contradiction, but I've made the timing of when we expect to remove network access to the functionapp more clear.

PR #17756, opened 2025-04-07
[17510] - moving convert queue message call to report function

### LINKED ISSUE #17510: Receive Step: Only Send Queue Message Once Fully Complete
### User Story

As a maintainer of the Universal Pipeline,
I want the receive step to enqueue a Queue Message for the next step only when it has successfully completed ALL operations,
so that the downstream steps do not encounter data integrity issues.

### Description/Use Case

A Convert step Queue Message can be sent from the Receive step before the database has been properly updated. If for some reason the database does not get updated (due to a DB exception), then the convert step will try to update report_lineage with a parentID that does not exist and encounter an exception. 

This happened in prod: Search for Operation ID `c10285c27d4876f6fa9855ca34bb350b` in prod app insights for more info (March 4th). Here is a snapshot of the logging that shows a database write encountering an exception AFTER the convert step has already been queued:

### Dev Notes

This scenario is handled for the UP-specific steps. See `FHIRFunctions.kt` for inspo:

```
    private fun runFhirEngine(
        message: String,
        dequeueCount: Int,
        fhirEngine: FHIREngine,
        actionHistory: ActionHistory,
    ): List<QueueMessage> {
        val messageContent = readMessage(fhirEngine.engineType, message, dequeueCount)

        try {
            val newMessages = databaseAccess.transactReturning { txn ->
                val results = fhirEngine.run(messageContent, actionLogger, actionHistory, txn)
                recordResults(message, actionHistory, txn)
                results
            }
            reportStreamEventService.sendQueuedEvents()
            return newMessages
        }
```

### Acceptance Criteria

- [x] DB writes in Receive step are guaranteed to complete before a QueueMessage for the next step is sent out.

oluwie:
This PR removes the manual call to add the convert step to the queue in favor of using the pre-exising `actionHistory.track` paradigm.

**If you are suggesting a fix for a currently exploitable issue, please disclose the issue to the prime-reportstream team directly outside of GitHub instead of filing a PR, so we may immediately patch the affected systems before a disclosure. See [SECURITY.md/Reporting a Vulnerability](https://github.com/CDCgov/prime-reportstream/blob/main/SECURITY.md#reporting-a-vulnerability) for more information.**

Test Steps:
    - Run ./gradlew quickRun & submit a report to the reports endpoint and make sure the report runs through the pipeline as usual. 

## Changes
 - Replaces manual call to send convert event to the queue in favor of `actionHistory.trackFhirMessage` which keeps track all queued fhir messages and routes them to the appropriate queue. This was done to continue using existing paradigms by having the `ActionHististory` class route messages/events to their respective queues using it's internal logic. The existing `actionHistory.track` call couldn't be used because it's only used to process messages of type `Event`. Typically, a `CONVERT` Event call would have been used to to transition a message from the receive stop to the convert step. However, the convert event type lacks necessary data that the convert azure function requires. The azure convert function requires a message of base type `QueueMessage` and it would have been a heavier and more complicated lift to modify the `Event` type with the extra necessary data.  A new function `actionHistory.trackFhirMessage` was created that expects a message of type `QueueMessage` which is what the convert function expects.

Note: While this fixes the issue of queueing the convert message as the last step of the submission workflow, a more complete solution would be to also ensure all writes in a the submission flow occur in a single transaction. This would prevent all data integrity issues that can occur within the submission flow. A new ticket (#17785 ) has been created to handle this scenario

### Testing
- [x] Tested locally?
- [x] Added tests?

### Process
- [] Updated the release notes?

## Linked Issues
- Fixes  #17510
- #17785 

    - No
   - No
   - No
  - No
   - No
  - No


prime-router/src/main/kotlin/azure/Event.kt
```diff
@@ -53,7 +53,7 @@ abstract class Event(val eventAction: EventAction, val at: OffsetDateTime?) {
                 RECEIVER_ENRICHMENT -> TaskAction.receiver_enrichment
                 RECEIVER_FILTER -> TaskAction.receiver_filter
                 RECEIVE -> TaskAction.receive
-                CONVERT -> TaskAction.convert
```
oluwie, line 56: Instead of updating this event, another option would be to update `toQueueName()` to properly return the correct queue. But made this change to ensure all event names and queue match up even though that tech isn't required

arnejduranovic (commented):
I think this is a good attempt, just needs a bit of clean up. The changes to ActionHistory are simple and fit in with what that class is doing, so its not making it less cohesive. Although, ActionHistory itself probably needs to be overhauled at some point, that's a task for later.

Please make sure to include good detail in the Pull Request description on why this change was necessary (why we couldn't just use trackEvent).

prime-router/src/main/kotlin/azure/Event.kt
oluwie: no longer needed

prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -113,6 +114,8 @@ class ActionHistory(
      */
     val messages = mutableListOf<Event>()
 
+    val fhirMessages = mutableListOf<FhirConvertQueueMessage>()
```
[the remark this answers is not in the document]
oluwie: Updated

prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -770,6 +781,16 @@ class ActionHistory(
         }
     }
 
+    /**
+     *  Sends any FHIR queue messages to the appropriate queue
+     */
+    fun queueFhirMessages(workflowEngine: WorkflowEngine) {
```
JFisk42, line 787: Can we add a happy path test for this? Some parts of this file are a beast but I think it should be pretty easy to add in some messages with `trackFhirMessage` and then verify they get processed here.

JFisk42 (commented):
Everything is looking good and was able to verify locally that there are no hiccups with this change 🚀 . Only thing that would be really nice is if we could add a unit test for the ActionHistory changes.

JFisk42 (approved):
Great looking tests!

prime-router/src/test/kotlin/azure/ActionHistoryTests.kt
```diff
@@ -908,4 +914,135 @@ class ActionHistoryTests {
...
+            Topic.FULL_ELR,
+            ""
+        )
+        )
+
+        assertEquals(2, actionHistory.fhirQueueMessages.size)
+    }
+
+    @Test
+    fun `test queueFhirMessages properly tracks and queues a fhir messages`() {
```
arnejduranovic, line 965:
```suggestion
    fun `test queueFhirMessages properly tracks and queues a fhir message`() {
```

arnejduranovic, line 965: Good job on these tests! `nit:` you could probably remove the two tests that test adding only a single message, I think they are redundant (if you can add two, why wouldn't you be able to add one?)

oluwie: Updated

arnejduranovic (approved):

PR #17765, opened 2025-04-08
fix(ci): cache is looking for a path before yarn is installed

myduong-navapbc:
addresses this error

Screenshot 2025-04-08 at 1 27 40 PM

Post job cleanup.
Error: Path Validation Error: Path(s) specified in the action for caching do(es) not exist, hence no cache is being saved.


scott-aquia:
Can we install yarn without the --immutable flag? 

Some of the dependabot PRs are failing because [doing whatever they need to do] would modify the yarn.lock file, which is not permitted in the default config. Thank you!

myduong-navapbc:
> Can we install yarn without the --immutable flag?
> 
> Some of the dependabot PRs are failing because [doing whatever they need to do] would modify the yarn.lock file, which is not permitted in the default config. Thank you!

We can, i feel that it causes more issues than it solves

scott-aquia:
> We can, i feel that it causes more issues than it solves

Fantastic - thanks, My!

scott-aquia (approved):

PR #17854, opened 2025-04-16
[17763] - Pass root report to azure events

### LINKED ISSUE #17763: Pass root report id to Azure events when available
### User Story

As a maintainer of the Universal Pipeline,
I want to avoid making unnecessary calls to the DB,
so that I can avoid related performance issues

### Description/Use Case

Several UP steps need to call `reportService.getRootReport(message.reportId)` to perform their business logic. These same steps then end up calling `reportEventService.sendItemEvent(...)` which ends up calling `getRootReports` in  `ReportStreamEventService.getItemEventData`. Instead of querying the DB twice in this flow, we should investigate to see if we can optionally pass in the root report so it does not need to be calculated again in `ReportStreamEventService.getItemEventData`.

There is a similar issue with `ReportStreamEventService.getReportEventData` as well that will need to be investigated in this ticket.

Once `ReportStreamEventService.getReportEventData` and `reportEventService.sendItemEvent(...)` have been updated to accept the optional root reports, go through all usages of Azure Events and pass in the root reports if it makes sense to do so for those particular usages.

### Dev Notes

See FHIRTranslator.kt for an example:
1. In `sendTranslated`, we call: `val originalReport = reportService.getRootReport(message.reportId)`
2. Then later in the same method, we call `reportEventService.sendItemEvent(` which will ultimately repeat the DB query to get the root reports

In this case, since we already get the root report, we should just pass that to the event logic so it doesn't have to calculate it again. 

### Acceptance Criteria

- [ ] Updated `sendItemEvent` to accept optional `rootReport(s)`
- [ ] Updated `sendReportEvent` to accept optional `rootReport(s)`
- [ ] Updated usages of the above to pass in the root report IDs or Report objects where appropriate

oluwie:
This PR updates `ReportStreamEventService.getReportEventData` and `ReportStreamEventService.getItemEventData` functions to accept a new parameter `rootRepot(s)` which is either a single or a list of `ReportFile(s)` meant to represent the origin report in the pipeline. If passed in, the `rootReport(s)` is used in these functions instead of querying the db for the original `rootReport(s)`. This is meant as an optimization so we don't need to query the db for data we already have access to.

**If you are suggesting a fix for a currently exploitable issue, please disclose the issue to the prime-reportstream team directly outside of GitHub instead of filing a PR, so we may immediately patch the affected systems before a disclosure. See [SECURITY.md/Reporting a Vulnerability](https://github.com/CDCgov/prime-reportstream/blob/main/SECURITY.md#reporting-a-vulnerability) for more information.**

Test Steps:
1. `./gradlew quickRun` & submit a report to the any endpoint and verify the report is processed as usual. As this is only an optimization, it shouldn't affect current existing behavior.

## Changes
- Added `addRootReport` function to `ReportStreamEventBuilder` builder
- Modified `ReportStreamEventService.getItemEventData`  to accept and use `rootReport` instead of db query if passed in
- Modified `ReportStreamEventService.getReportEventData`  to accept and use `rootReports` instead of db query if passed in
- Updated usages of `reportEventService.sendItemEvent` and reportEventService.sendReportEvent` to add `addRootReport` function where it makes sense

### Testing
- [x] Tested locally?
- [x] Added tests?

### Process

## Linked Issues
- Fixes #17763


prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -636,10 +636,13 @@ class ActionHistory(
         reportFile.blobDigest = blobInfo.digest
         reportFile.bodyUrl = blobInfo.blobUrl
 
+        val rootReports = listOf(reportFile)
```
arnejduranovic, line 639: This is not the "root report", this is just the reportId of the sent report?

prime-router/src/main/kotlin/fhirengine/azure/FHIRFunctions.kt
```diff
@@ -247,6 +247,7 @@ class FHIRFunctions(
                 ReportStreamEventName.PIPELINE_EXCEPTION,
                 report,
                 fhirEngine.taskAction,
+                null,
```
arnejduranovic, line 250: I'm thinking this should not be a required param, but optional param that is set similar to how "params" is set below.

arnejduranovic (changes requested):

prime-router/src/test/kotlin/azure/observability/event/ReportEventServiceTest.kt
```diff
@@ -217,4 +229,135 @@ class ReportEventServiceTest {
             )
         )
     }
+}
+
+class ReportEventStreamUnitTests {
```
oluwie, line 234: These tests should prob live somewhere else

prime-router/src/main/kotlin/azure/observability/event/ReportStreamEventService.kt
```diff
@@ -189,6 +189,7 @@ interface IReportStreamEventService {
         parentReportId: UUID?,
         pipelineStepName: TaskAction,
         topic: Topic?,
+        rootReports: List<ReportFile>,
```
arnejduranovic, line 192: Why is this one not optional but it is optional in `getItemEventData`?

arnejduranovic (changes requested):
Please see comments below. Additionally:

- val submittedReportIds = workflowEngine.reportService.getRootReports(report.reportId) can be moved up so that in can be passed to be `REPORT_LAST_MILE_FAILURE` event.
- submitted report IDs could also probably be passed to the events called in `FHIRConverter.kt`? I recommend tracing how submittedReportId gets calculated for those normally and, if it required a DB call, see if we can't pass it in from the calling code.

PR #17869, opened 2025-04-17
Add timeBetweenBatches delay to space out batch messages

### LINKED ISSUE #17744: Update UP to avoid more than one request at a time to a receiver's endpoint
### User Story

As a ReportStream receiver (CA specifically),
I want my receiving endpoint to only have to process one submission at a time,
so that my endpoint can function properly

### **Description / Use Case**

**Note:** This is an alternate and more scoped down version of #17706

Currently, **ca-dph** is experiencing issues processing reports because ReportStream is sending more than one Post request to their endpoint at a time

- **ca-dph** is configured to receive **400 reports every 10 minutes**.
    
- The batch step sends multiple files containing 400 reports each. If ReportStream receives more than 400 reports within a 10-minute window for CA, multiple files are generated.
    
- The send step transmits all these files simultaneously to the receiver, causing only the first request to return a successful response. Any subsequent API calls return the following error **BUT ALLEGEDLY PROCESS THE REPORTS FINE**:
    
    `org.apache.http.ConnectionClosedException: Premature end of Content-Length delimited message body (expected: 32,032; received: 16,235)`  
    

### Dev Notes

- The root cause of this issue is currently expected to be on the CA side. The following dev notes discuss how RS could help alleviate these symptoms.
- A potential short-term solution is to introduce a **timeout** between each batch file sent when multiple files are queued. This would allow the receiver to process each API request sequentially.
- Existing batch logic can be found inside `BatchDeciderFunction.kt` starting on line 48 `determineQueueMessageCount`. If implementing the proposed solution, this code shall need to be updated to stagger the visibility value of output QueueMessages for a single receiver so that there is enough time for one to get processed before the other is triggered. Updating the visibility time of a QueueMessage can be done by setting the third param (`invisibleDuration`) of `QueueAccess.sendMessage` (See FHIRFunctions.kt for an example usage of this function)
- Think about best way to set `invisibleDuration`. Should it be hardcoded? Or should it be calculated somehow? If calculating it, something like the following formula may be desirable: `number-of-seconds-to-next-batch / number-of-generated-batched-reports`

### Acceptance Criteria
- [ ]  UP at least makes an attempt to avoid having more than one request hit a single receiver's endpoint at a time
- [ ] This functionality should be wrapped in a receiver setting and only enabled for CA in prod
- [ ] Create new ticket to remove the new logic once CA fixes the root cause. Add comment in the code that links to ticket.
- [ ] Communicate to @chris-kuryak that this feature is complete and can be configured

kant777:
This PR adds support for spacing out batch messages by leveraging Azure Storage Queue’s invisible-duration feature. I’ve introduced a new `timeBetweenBatches` field on the Receiver model, which defines the delay between each batch. In the `BatchDeciderFunction`, I now calculate a per-message `delay = timeBetweenBatches.multipliedBy(index)` and pass it into `sendMessageToQueue(event, queueName, delay)`, so that each queued message remains invisible for its configured interval before becoming visible. This ensures that batches are released evenly over time rather than all at once.

## Changes
- Calculate the interval between scheduled runs based on receiver.timing.numberPerDay
- Divide the interval by the number of messages to compute spacingSeconds
- Stamp each BatchEvent’s at timestamp to delay its visibility by spacingSeconds * index
- Preserve existing comments and minimal diff footprint

### Testing
- [ ] Tested locally? don't have a way to test other than running below tests
- [ ] Ran `./prime test` or `./gradlew testSmoke` against local Docker ReportStream container? yes
- [ ] (For Changes to /frontend-react/...) Ran `npm run lint:write`? N/A
- [ ] Added tests? no

### Process
- [ ] Are there licensing issues with any new dependencies introduced? N/A
- [ ] Includes a summary of what a code reviewer should test/verify?  please see the changes above
- [ ] Updated the release notes? N/A
- [ ] Database changes are submitted as a separate PR? N/A
- [ ] DevOps team has been notified if PR requires ops support? N/A

## Linked Issues
- Fixes #17744


prime-router/src/main/kotlin/azure/QueueAccess.kt
```diff
@@ -33,6 +33,15 @@ object QueueAccess {
         sendMessageToQueue(event, queueName)
     }
 
+    fun sendMessageToQueue(
+        event: Event,
+        queueName: String,
+        invisibleDuration: Duration,
+    ) {
+        val base64Message = String(Base64.getEncoder().encode(event.toQueueMessage().toByteArray()))
+        sendMessage(queueName, base64Message, invisibleDuration)
```
jack-h-wang, line 42: Should we also add the existing duration calculation (`Duration.between(now, event.at ?: now)`) here?

prime-router/src/main/kotlin/azure/batch/BatchDeciderFunction.kt
```diff
@@ -41,22 +42,62 @@ class BatchDeciderFunction(private val workflowEngine: WorkflowEngine = Workflow
...
+                        )
+                        val recordsToBatch = workflowEngine.db.fetchNumReportsNeedingBatch(
+                            rec.fullName, backstopTime, txn
+                        )
+                        val maxCount = rec.timing!!.maxReportCount
+                        val fullBatches = recordsToBatch / maxCount
+                        val remainder = recordsToBatch % maxCount
+                        val batchSizes = mutableListOf<Int>().apply {
+                            repeat(fullBatches) { add(maxCount) }
+                            if (remainder > 0) add(remainder)
```
jack-h-wang, line 67: Should we calculate the remainder here rather than before we create the list of batch sizes? I'm not sure how well this works if `recordsToBatch` is more than 2 times the size of `maxCount`.

jack-h-wang (changes requested):
Have some concerns from a quick look. Also, I think we should have a test to exercise the new code.

prime-router/src/main/kotlin/azure/batch/BatchDeciderFunction.kt
```diff
@@ -56,7 +56,12 @@ class BatchDeciderFunction(private val workflowEngine: WorkflowEngine = Workflow
                                 BatchConstants.Queue.COVID_BATCH_QUEUE
                             }
 
-                            workflowEngine.queue.sendMessageToQueue(event, queueName)
+                            val delay = rec.timing!!.timeBetweenBatches
+                            if (delay.isZero) {
+                                workflowEngine.queue.sendMessageToQueue(event, queueName)
```
arnejduranovic, line 61: We can just pass the absolute value (including if 0) and that will reduce a lot of the code?

arnejduranovic (changes requested):

prime-router/src/main/kotlin/azure/batch/BatchDeciderFunction.kt
```diff
@@ -56,7 +57,7 @@ class BatchDeciderFunction(private val workflowEngine: WorkflowEngine = Workflow
                                 BatchConstants.Queue.COVID_BATCH_QUEUE
                             }
 
-                            workflowEngine.queue.sendMessageToQueue(event, queueName)
+                            workflowEngine.queue.sendMessageToQueue(event, queueName, delay)
```
jack-h-wang, line 60: As far as I can tell, this is actually the only call to `sendMessageToQueue` aside from mocks set up during tests. In this case, I think I'd prefer this to just become the sole definition and not have an overload with the old method - see next comment.

prime-router/src/main/kotlin/azure/QueueAccess.kt
```diff
@@ -33,6 +33,31 @@ object QueueAccess {
...
+        event: Event,
+        queueName: String,
+        delay: Duration,
+    ) {
+        if (delay.isZero || delay.isNegative) {
+            return sendMessageToQueue(event, queueName)
+        }
+        val base64 = Base64.getEncoder()
+            .encodeToString(event.toQueueMessage().toByteArray())
+        sendMessage(queueName, base64, delay)
```
jack-h-wang, line 58: I still think that `Duration.between(now, event.at ?: now)` should be used as the base delay instead of being completely ignored. If we add the two values together here, then we will also end up with the equivalent to the previous logic without having to keep the original as an overload.

jack-h-wang (approved):
Just wanted to note there are two `organizations.yml` files, and it's the one in `prime-router/settings` that is currently validated by `organizations.json`. It might be more effective to place a demonstration of the new setting there for that reason.

prime-router/src/main/kotlin/Receiver.kt
```diff
@@ -6,6 +6,7 @@ import gov.cdc.prime.router.common.DateUtilities
 import gov.cdc.prime.router.common.Environment
 import gov.cdc.prime.router.fhirengine.translation.hl7.FhirToHl7Converter
 import gov.cdc.prime.router.fhirengine.translation.hl7.SchemaException
+import java.time.Duration
```
jack-h-wang, line 9: This needs to be removed, causing a linting error

prime-router/src/test/kotlin/azure/batch/BatchDeciderTests.kt
```diff
@@ -177,4 +182,43 @@ class BatchDeciderTests {
...
+    @Test
+    fun `run should enqueue correct number of batch messages with appropriate delays`() {
+        // ensure this receiver is considered valid and due to batch
+        every { timing1.isValid() } returns true
+        // stub the default‐arg overload for batchInPrevious60Seconds()
+        every { timing1.batchInPrevious60Seconds(any<OffsetDateTime>()) } returns true
+
+        every { timing1.numberPerDay } returns 1
+        every { timing1.maxReportCount } returns 2
+        every { timing1.timeBetweenBatches } returns 120
```
arnejduranovic, line 195: Would it be better if instead of mocking this out we set it up in the Receiver to verify it is read properly?

scott-aquia (approved):

PR #17906, opened 2025-04-21
add skip-schema-validation option to build-backend action

buck30wale:
This PR ...

## Changes
 Added `skip-schema-validation` input parameter to the build-backend action with a default value of true, which:

 1. Adds explicit control over schema validation in the build-backend action
2. Removes redundant skip-schema-validation settings from release_to_github.yml

Both workflows will skip schema validation by default since they use the action's default value of true.

## Testing
- The parameter is set to `true` by default in all workflows to unblock deployments
- Schema validation can be re-enabled by setting `skip-schema-validation: false` once the underlying issue is resolved

## Related Issues
- This change is a temporary workaround to unblock deployments while investigating the schema validation error
- A separate investigation is needed to resolve the root cause of the schema validation failure

### DIFF
--- a/.github/actions/build-backend/action.yml
+++ b/.github/actions/build-backend/action.yml
@@ -15,6 +15,9 @@ inputs:
     default: false
   sp-creds:
     description: "Azure Service Principal creds"
+  skip-schema-validation:
+    description: "Skip schema validation step"
+    default: true
 
 runs:
   using: "composite"
@@ -80,7 +83,7 @@ runs:
       uses: ./.github/actions/runner-ip
 
     - name: Validate translation schemas
-      if: inputs.run-integration-tests == 'true' && inputs.sp-creds
+      if: inputs.run-integration-tests == 'true' && inputs.sp-creds && !inputs.skip-schema-validation
       uses: ./.github/actions/retry
       with:
         timeout_minutes: 5


.github/actions/build-backend/action.yml
```diff
@@ -15,6 +15,9 @@ inputs:
     default: false
   sp-creds:
     description: "Azure Service Principal creds"
+  skip-schema-validation:
+    description: "Skip schema validation step"
+    default: false
```
arnejduranovic, line 20: Based on the name "skip-schema-validation" it would make more sense to invert the logic, meaning default: should be TRUE, and then update the logic on line 86 to be `!inputs.skip-schema-validation`

.github/workflows/build_hub.yml
```diff
@@ -60,9 +60,11 @@ jobs:
         uses: actions/checkout@d632683dd7b4114ad314bca15554477dd762a938
 
       - name: Build backend
+        if: ${{ needs.pre_job.outputs.has_router_change == 'true' }}
```
arnejduranovic, line 63: It looks like we are doing more than just skip schema validation in this PR? Please update the PR description to include a description and reason for these changes or remove these changes from this PR?

arnejduranovic (changes requested):

.github/workflows/release_to_azure.yml
```diff
@@ -53,9 +53,11 @@ jobs:
         uses: actions/checkout@d632683dd7b4114ad314bca15554477dd762a938
 
       - name: Build backend
+        if: ${{ needs.wf_vars.outputs.enable_builds == 'true' }}
         uses: ./.github/actions/build-backend
         with:
-          version: ${{ needs.pre_job.outputs.version }}
+          version: ${{ needs.wf_vars.outputs.version }}
```
arnejduranovic, line 59: Is there a reason we included this change? The default of build-backend is to skip validation, so why are we passing true here? Ditto for the release_to_github change.

arnejduranovic (changes requested):

arnejduranovic (approved):

PR #17989, opened 2025-04-30
17343 add queuemessage to azure events

### LINKED ISSUE #17343: Add queueMessage to all Azure Events
### User Story
As an Engagement Engineer, I need additional information about a failed message – so that I may more efficiently identify the issue at hand.

### Description/Use Case
When Engagement Engineers reprocess/ resend messages (for a variety of reasons), they normally have to tediously look through all the messages' traces to find the queue message. To expedite this process, they would like to have the queueMessage property be included in all Azure Events so they can more quickly identify the location of issues when re-processing messages. The following are examples of how they use queueMessage from various Azure events:
- ITEM_NOT_ROUTED - In case we need to adjust a filter to route the messages
- ITEM_TRANSFORMED - If the receiver asks to change their transforms and we need to resend the messages after changing the transform ( This happens a lot)
- ITEM_FILTER_FAILED - If we dropped messages for having incorrect filters and we need to resend the message after fixing the filter
- REPORT_NOT_PROCESSABLE - If there's an error in the Convert step due to either 1) missing mappings in the code to condition table or 2) errors in the sender transforms, we would need to resend the message after those get addressed.

- Are there other events that would benefit the inclusion of the queueMessage property?
- For the scenario when there are missing mappings in the code to condition table, should an entirely new Azure Event be created to better isolate this issue? Or is it sufficient to rely on the REPORT_NOT_PROCESSABLE event?
- Will this hit the character limit? - not if we put it at the root of customDimensions and not inside Params
- Should this be added as a global property to **all events**?
- Is there a risk of PII being included in the message?

### Dev Notes
- Add this property at the root of customDimensions – therefore, character limit will not be a problem.
- The queue message is passed in to all queue triggered azure functions in the app, which includes all steps between convert and translate functions. We want to simply add this to a new queueMessage property at the root of customDimensions.

### Acceptance Criteria
- [ ] queueMessage property is added to all the Azure Events

jack-h-wang:
This PR updates Azure events to emit the Azure queue message when available.

Test Steps:
1. Search the logs in `pdhtest-appinsights` to look for `ITEM_TRANSFORMED` or `ITEM_NOT_ROUTED` events to see samples of queue messages being included.

## Changes
- Structure of Azure events changed to include `queueMessage` as a top level item within `customDimensions`.
- Various pipeline steps adjusted to emit queueMessage when available.
- Adjustments to allow serialization of queue messages.

### Testing
- [x] Tested locally?
- [x] Added tests?

## Linked Issues
- Fixes #17343

### DIFF
--- a/prime-router/src/main/kotlin/azure/ActionHistory.kt
+++ b/prime-router/src/main/kotlin/azure/ActionHistory.kt
@@ -644,8 +644,8 @@ class ActionHistory(
         reportFile.bodyUrl = blobInfo.blobUrl
 
         reportEventService.sendReportEvent(
-            childReport = reportFile,
             eventName = ReportStreamEventName.REPORT_SENT,
+            childReport = reportFile,
             pipelineStepName = TaskAction.send
         ) {
             parentReportId(header.reportFile.reportId)
@@ -722,8 +722,7 @@ class ActionHistory(
                             nextRetryTime?.let {
                                 ReportStreamEventProperties.NEXT_RETRY_TIME to
                                     DateTimeFormatter.ISO_DATE_TIME.format(it)
-                            },
-                            queueMessage?.let { ReportStreamEventProperties.QUEUE_MESSAGE to queueMessage }
+                            }
                         ).toMap()
                     )
                 }
--- a/prime-router/src/main/kotlin/azure/ActionHistory.kt
+++ b/prime-router/src/main/kotlin/azure/ActionHistory.kt
@@ -646,7 +646,7 @@ class ActionHistory(
         reportEventService.sendReportEvent(
             eventName = ReportStreamEventName.REPORT_SENT,
             childReport = reportFile,
-            pipelineStepName = TaskAction.send
+            pipelineStepName = TaskAction.send,
         ) {
             parentReportId(header.reportFile.reportId)
             params(
@@ -710,7 +710,7 @@ class ActionHistory(
                         )
                     )
                 )
-                reportEventService.sendItemEvent(eventName, report, TaskAction.send) {
+                reportEventService.sendItemEvent(eventName, report, TaskAction.send, queueMessage ?: "") {
                     trackingId(bundle)
                     parentReportId(itemLineage.parentReportId)
                     childItemIndex(itemLineage.childIndex)
--- a/prime-router/src/main/kotlin/azure/ActionHistory.kt
+++ b/prime-router/src/main/kotlin/azure/ActionHistory.kt
@@ -608,6 +608,7 @@ class ActionHistory(
         reportService: ReportService,
         transportType: String,
         lineages: List<ItemLineage>?,
+        queueMessage: String,
     ) {
         if (isReportAlreadyTracked(sentReportId)) {
             error(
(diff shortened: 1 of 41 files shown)


prime-router/src/main/kotlin/azure/ActionHistory.kt
```diff
@@ -722,8 +722,7 @@ class ActionHistory(
                             nextRetryTime?.let {
                                 ReportStreamEventProperties.NEXT_RETRY_TIME to
                                     DateTimeFormatter.ISO_DATE_TIME.format(it)
-                            },
-                            queueMessage?.let { ReportStreamEventProperties.QUEUE_MESSAGE to queueMessage }
```
jack-h-wang, line 726: Removing the `QUEUE_MESSAGE` parameter in favor of the new top level element.

prime-router/src/test/kotlin/fhirengine/azure/FHIRConverterIntegrationTests.kt
```diff
@@ -184,21 +184,26 @@ class FHIRConverterIntegrationTests {
...
+        "headers":$headersString
         }
-    """.trimIndent()
+        """.trimIndent().replace("\n", "")
     }
 
+    private fun appendMessageQueueName(
+        queueMessage: String,
+        messageQueueName: String,
+    ): String = queueMessage.substringBeforeLast("}") + ",\"messageQueueName\":\"$messageQueueName\"}"
```
jack-h-wang, line 205: `messageQueueName`, though not included in the original queueMessage, appears as a result of serializing the `ReportPipelineMessage` within the pipeline.

prime-router/src/test/kotlin/fhirengine/azure/FHIRConverterIntegrationTests.kt
```diff
@@ -166,15 +166,15 @@ class FHIRConverterIntegrationTests {
...
+        "type":"convert",
+        "reportId":"${report.id}",
+        "blobURL":"${report.bodyURL}",
+        "digest":"${BlobUtils.digestToString(BlobUtils.sha256Digest(blobContents.toByteArray()))}",
+        "blobSubFolderName":"${sender.fullName}",
+        "topic":"${sender.topic.jsonVal}",
+        "schemaName":"${sender.schemaName}"
         }
-    """.trimIndent()
+        """.trimIndent().replace("\n", "")
```
jack-h-wang, line 177: These formatting changes remove whitespace that was previously introduced unintentionally.

victor-chaparro (approved):
LGTM. Verified all these ITEM custom events contained the queue message in the expected format:
image

adegolier:
Are we expecting some of these to be blank?
image

jack-h-wang:
> Are we expecting some of these to be blank? image

`REPORT_RECEIVED` doesn't appear to be an event thrown by a queue trigger, so I wouldn't expect anything to be included there. For the others, I've tried introducing changes to the transport types to pass this information to the events.

adegolier:
Still not seeing it on `REPORT_SENT` but I do see it on `ITEM_SENT` now, and, notably on `ITEM_SEND_ATTEMPT_FAIL` 
image
If that's expected, I'll approve.

Issue #8222, opened 2023-02-07
Add repo documentation


PatriciaPerozoUSDS:

wcutshall:
As mentioned in #18412, we should be sure to document the usage of SchemaServiceProvider.

PR #18412, opened 2025-08-14
[18380] Remove the Blob Connection dependency from FHIR mapping library.

wcutshall:
This PR is in support of user story [18380](https://app.zenhub.com/workspaces/reportstream-om-67feaaa15ecf34000f4cc206/issues/gh/cdcgov/prime-reportstream/18380). 

The gist of this PR is to move the references to `BlobAccess.BlobContainerMetadata` references out of the `FhirToHl7Converter` and `FhirTransformer` classes and just pass in a schema reference that is created outside of those classes.

Test Steps:
1. Run unit, integration, smoke and e2e tests.  Functionality should remain the same, only the implementation has changed.

## Changes
- See this [list](https://github.com/CDCgov/prime-reportstream/pull/18412/files).

### Testing
- [x] Tested locally?
- [ ] Added tests?

### DIFF
--- a/prime-router/src/main/kotlin/fhirengine/translation/hl7/schema/ConfigSchemaReader.kt
+++ b/prime-router/src/main/kotlin/fhirengine/translation/hl7/schema/ConfigSchemaReader.kt
@@ -137,7 +137,7 @@ object ConfigSchemaReader : Logging {
     ): Schema {
         val schemaServiceProvider = schemaServiceProviders.get(schemaUri.scheme)
         if (schemaServiceProvider == null) {
-            throw SchemaException("Unexpected scheme: ${schemaUri.scheme}")
+            throw SchemaException("No schema service provider found for: ${schemaUri.scheme}")
         }
         val rawSchema =
             readOneYamlSchema(schemaServiceProvider.getInputStream(schemaUri), schemaClass)
--- a/prime-router/src/test/kotlin/fhirengine/translation/hl7/utils/helpers/SchemaReferenceResolverHelperTests.kt
+++ b/prime-router/src/test/kotlin/fhirengine/translation/hl7/utils/helpers/SchemaReferenceResolverHelperTests.kt
@@ -18,7 +18,7 @@ class SchemaReferenceResolverHelperTests {
         assertNotNull(blobInfo)
         assertEquals("metadata", blobInfo.containerName)
         // This is set in build.gradle.kts
-//        assertEquals("test-AzureWebJobsStorage", blobInfo.connectionString)
+        assertEquals("test-AzureWebJobsStorage", blobInfo.connectionString)
     }
 
     @Test


prime-router/src/main/kotlin/fhirengine/translation/hl7/FhirTransformer.kt
```diff
@@ -36,23 +33,6 @@ class FhirTransformer(
     private val valueXRegex = Regex("""value[A-Z][a-z]*""")
```
JFisk42, line 33: Unrelated to this PR, should FhirTransformer be a part of the prime-fhir-converter library? I can see the argument either way but I don't know if there has been any discussion.

prime-router/src/main/kotlin/fhirengine/translation/hl7/schema/providers/SchemaServiceProvider.kt
```diff
@@ -0,0 +1,9 @@
+package gov.cdc.prime.router.fhirengine.translation.hl7.schema.providers
+
+import java.io.InputStream
+import java.net.URI
+
+interface SchemaServiceProvider {
```
JFisk42, line 6: I think this will be one of the main entry points for implementers, we should have documentation aimed at that as a part of [#8222](https://github.com/CDCgov/prime-reportstream/issues/8222) or wherever we end up doing usage documentation. Now that I'm typing it out this is probably something that should be in an .md instead of a javadoc.

prime-router/src/test/kotlin/fhirengine/translation/hl7/FhirToHl7ConverterTests.kt
```diff
@@ -445,8 +446,12 @@ class FhirToHl7ConverterTests {
...
         // Use a file based schema which will fail as we do not have enough data in the bundle
         val transformer = FhirToHl7Converter(
-            "classpath:/fhirengine/translation/hl7/schema/schema-read-test-01/ORU_R01.yml",
-            mockk<BlobAccess.BlobContainerMetadata>(), warnings = mutableListOf(), errors = mutableListOf()
+            SchemaReferenceResolverHelper.retrieveHl7SchemaReference(
+                "classpath:/fhirengine/translation/hl7/schema/schema-read-test-01/ORU_R01.yml",
+                mockk<BlobAccess.BlobContainerMetadata>()
+            ),
+            warnings = mutableListOf(),
+            errors = mutableListOf()
```
JFisk42, line 454: These tests should be refactored to use file or classpath schemas instead. It can happen now or in #18423 (add a note there if so).

prime-router/src/test/kotlin/fhirengine/translation/hl7/schema/ConfigSchemaReaderTests.kt
```diff
@@ -94,7 +93,7 @@ class ConfigSchemaReaderTests {
                 "classpath:/fhir_sender_transforms/sample_schema.yml",
 
                 schemaClass = FhirTransformSchema::class.java,
-                mockk<BlobAccess.BlobContainerMetadata>(),
+                SchemaReferenceResolverHelper.getSchemaServiceProviders(mockk<BlobAccess.BlobContainerMetadata>())
```
JFisk42, line 96: Same note here about refactoring tests.

JFisk42 (approved):
Great job Bill! 🚀 
I'm going to add on to the file spreadsheet to track which files should stay and which are moving out. I did get confused at one point because of my assumption that everything in /hl7/ would be moving out.

prime-router/src/main/kotlin/fhirengine/translation/hl7/schema/providers/SchemaServiceProvider.kt
wcutshall: Comment left in #8222

prime-router/src/test/kotlin/fhirengine/translation/hl7/schema/ConfigSchemaReaderTests.kt
wcutshall: See last comment above.

prime-router/src/main/kotlin/fhirengine/translation/hl7/FhirTransformer.kt
JFisk42: I started tracking that info on a new tab in the sheet victor started: https://docs.google.com/spreadsheets/d/116ju7q4YtTxgRxn-v1hOwVcsEq4m0wPxufjXeEYP_kY/edit?usp=sharing

PR #18493, opened 2025-09-09
[18452] Identify new workaround for hl7-v2-fhir-converter build issues.

wcutshall:
This PR documents a workaround for handling the `io.github.linuxforhealth:hl7v2-fhir-converter` source dependency inside of the IntelliJ IDEA IDE.


prime-router/docs/getting-started/intellij-tips.md
```diff
@@ -1,3 +1,90 @@
+# IntelliJ Tips
```
JFisk42, line 1: It looks like this document isn't linked anywhere yet. Can we toss in a link in the [main doc](https://github.com/CDCgov/prime-reportstream/blob/main/prime-router/docs/getting-started/README.md) (maybe under prerequisites or building the baseline) so that there are some breadcrumbs to what is a required step for IntelliJ use.

JFisk42 (approved):
Documentation look great 👍 , only ask is to link it from our main setup docs.

prime-router/docs/getting-started/intellij-tips.md
wcutshall: Updated the README.

Issue #15849, opened 2024-09-11
Copy of Develop documentation for Okta


chris-kuryak:
## Problem statement
Now that we have identified what documentation we need, we now need to create that documentation per the plan.

## What you need to know
- Okta Documentation Plan - #13621

## Acceptance criteria
Change the following in Github:
- [x] Update Simple Report (localdev) to reference our new instance in [Swagger](https://github.com/CDCgov/prime-reportstream/blob/b5bae0acfe6f73074d5ff047670a0dbf2784faef/prime-router/docs/getting-started/swagger.md)
  - [x] Step 3
  - [x] Step 8
- [x] In [Receivers doc](https://github.com/CDCgov/prime-reportstream/blob/b5bae0acfe6f73074d5ff047670a0dbf2784faef/prime-router/docs/onboarding-users/receivers.md), add a link to [setting up an Okta account](https://github.com/CDCgov/prime-reportstream/blob/b5bae0acfe6f73074d5ff047670a0dbf2784faef/prime-router/docs/onboarding-users/okta-account-creation.md) in the section that says "If the organization has elected for download access, set up an Okta account."

Create the following documentation In Github:
- [ ] How to manage the Okta instance and add/remove permissions for Okta admins
  - [ ] Roles and responsibilities around managing Okta
  - [x] Admin groupings within Okta
- [ ] How is Okta integrated into our front-end
  - [ ] Documenting the application under Okta that we use for FE things
  - [ ] Environment variables that need to be set to point correctly to our FE
  - [ ] Have a link to the Okta documentation with their libraries (we use their official JS libraries for login)
  - [ ] Include a blurb that we use the Okta-specific library with React to authenticate with Okta (we do not use a generic code)in the Front-end docs
- [ ] How is Okta integrated into our back-end (TBD by Platform)
  - [ ] Environment variables that need to be set to point correctly to our BE
  - [ ] If giving us an Okta token is a valid way to authenticate, then we need to ensure there is documentation in the Prime-router area explaining how Okta is an OAuth2 and how they can communicate using that with examples
- [ ] How Okta integrates with Azure and DevOps and the necessary elements needed from Okta to make it work properly (TBD by DevOps)
  - [ ] Mapping names from Okta variables to DevOps variables (may be covered by environment variables documentation above)

## To do

- [x] Chris to review what documentation already exists from Support team
- [x] Chris to align with Support team, Engagement team, and Patrick Stoneburner on roles and responsibilities

PR #18714, opened 2025-12-04
15849 add Okta setup doc

jack-h-wang:
This PR adds a setup documentation detailing Okta use by the backend.

Test Steps:
N/A

## Linked Issues
- #15849


prime-router/docs/okta/setup.md
```diff
@@ -0,0 +1,21 @@
...
+Okta endpoints. As of this writing, the expected values are:
+* `reportstream.okta.com` for production
+* `reportstream.oktapreview.com` for all other environments
+
+## Development notes
+
+* Okta authentication is attempted for nearly all API calls utilizing a bearer token, but as of this writing, most
+  API calls use RS's built in server to server authentication; requests to Okta are currently only expected to succeed 
+  for specific API calls that are requested via the frontend on behalf of a logged in user.
+* The email function appears to use a hardcoded value for the Okta URL instead of the environment variable.
```
JFisk42: > appears to

Might just be a nitpick with the wording here. I glanced at the function and it does appear to be true as far as I understand, if so let's update the wording. If there is still ambiguity we can document that.

